PR #4416, opened 2023-07-09
[cross_file] Improved documentation about ignored parameters in IO module.

andreacioni:
Hello guys, following the discussion in the issue below here I've tried to document all the ignored constructor parameters that are available but never used in the IO module.

- https://github.com/flutter/flutter/issues/87812

## Pre-launch Checklist

- [x] I signed the [CLA].


ValentinVignal:
I guess you can link your PR to https://github.com/flutter/flutter/issues/93379

ditman:
@stuartmorgan do you mind taking a look at this one? I'm going to be OOTO for the next few weeks!

packages/cross_file/lib/src/types/base.dart
```diff
@@ -44,6 +44,8 @@ abstract class XFileBase {
   }
 
   /// For web, it may be necessary for a file to know its MIME type.
+  ///
+  /// For io implementation, this is not intended to be set and will be `null`.
```
stuartmorgan, line 48: This isn't accurate; it doesn't have to be set, but it can be; there's no guarantee that it will be null.

packages/cross_file/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## NEXT
+## 0.3.3+5
 
-* Updates minimum supported SDK version to Flutter 3.7/Dart 2.19.
+* Updates minimum supported SDK version to Flutter 3.3/Dart 2.18.
```
stuartmorgan, line 3: It looks like there was a bad merge here.

packages/cross_file/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
...
+* Improved documentation about ignored parameters in IO module.
```
stuartmorgan, line 4: Improves

(Per the style guide linked from the checklist.)

stuartmorgan, line 4: I'm not sure what "module" means here. Maybe "in the non-web implementation."?

packages/cross_file/lib/src/types/io.dart
```diff
@@ -17,6 +17,15 @@ class XFile extends XFileBase {
   /// [bytes] is ignored; the parameter exists only to match the web version of
   /// the constructor. To construct a dart:io XFile from bytes, use
   /// [XFile.fromData].
+  ///
+  /// [name] is ignored; the parameter exists only to match the web version of
+  /// the constructor. **You should avoid using this parameter because it will
+  /// be likely removed in the future releases.**
```
stuartmorgan, line 23: I don't think we should make claims about what's going to be removed; it's much more likely that we would restructure the API in a broader way than just remove parameters.

stuartmorgan (changes requested):

packages/cross_file/lib/src/types/io.dart
andreacioni: Got it. I'm going the claim "_You should avoid using this parameter because it will be likely removed in the future releases._" from `name`, `length` in constructor and `fromData` factory

packages/cross_file/lib/src/types/io.dart
```diff
@@ -82,9 +92,17 @@ class XFile extends XFileBase {
   @override
   String get path => _file.path;
 
+  /// the name extracted from the last part of [File.path].
```
stuartmorgan, line 95: This should be a properly capitalized sentence.

packages/cross_file/lib/src/types/io.dart
```diff
@@ -82,9 +92,17 @@ class XFile extends XFileBase {
...
+  ///
+  /// **Be aware**, the name here is not the one
+  /// you may have passed in the constructor/factory.
```
stuartmorgan, line 98: The client of this API is often not the person who called the constructor, which makes this comment confusing. I think the comment you've add to the constructor is sufficient.

packages/cross_file/lib/src/types/io.dart
```diff
@@ -82,9 +92,17 @@ class XFile extends XFileBase {
...
   @override
   String get name => _file.path.split(Platform.pathSeparator).last;
 
+  /// the [File.length].
+  ///
+  /// **Be aware**, the length here is not the one
+  /// you may have passed in the constructor/factory.
```
stuartmorgan, line 105: Same two comments here.

stuartmorgan (changes requested):

packages/cross_file/lib/src/types/io.dart
```diff
@@ -82,9 +92,11 @@ class XFile extends XFileBase {
   @override
   String get path => _file.path;
 
+  /// The name extracted from the last part of [File.path].
   @override
   String get name => _file.path.split(Platform.pathSeparator).last;
 
+  /// The [File.length].
   @override
   Future<int> length() {
```
ditman, line 101: `name` and `length` are documented in the base class.

The documentation of the base class for [`name`](https://github.com/flutter/packages/blob/51e74b97508a3f85af2c45d994be0e0a58b496f8/packages/cross_file/lib/src/types/base.dart#L39-L44) and [`length`](https://github.com/flutter/packages/blob/main/packages/cross_file/lib/src/types/base.dart#L51-L54) may be a better place to fix?

andreacioni: Yes it could work too. I placed that there because it make more sense to me to have specificity of an implementation declared as close as possible to the actual class instead of to the abstract class. Anyway, I'm seeing right now that in the base class you are already documenting the return value of some methods based on the implementation it may be. I'll try to do the same for these ones.

ditman: Thanks! I'm not exactly sure what happens to the [generated documentation](https://pub.dev/documentation/cross_file/0.3.3+5/cross_file/XFile/name.html) when a method with `@override` has its own docs (not sure if they get merged, or the base class one is ignored!)

stuartmorgan (approved):
LGTM

stuartmorgan:
@ditman Did you intend to approve this?

ditman:
@stuartmorgan d'oh! I assumed this would land with only one approval, so I applied `autosubmit` directly. I'll approve and autosubmit again!

ditman (approved):
LGTM!

PR #4417, opened 2023-07-10
[flutter_plugin_tools] Reimplements the excerpt system inline in the tool, rather than relying on a separate package.

Hixie:
* Allows excerpts to come from any package, not just examples.
* Fixes a bug in the excerpting logic that was causing a stray `}` to appear in one example.
* Removes the need for `build.excerpt.yaml` files.
* Remove the dependency on build_runner for excerpts.
* Reduces the time to generate the excerpts from about 10 minutes to about 5 seconds.
* Almost certainly fixes https://github.com/flutter/flutter/issues/107180 (untested).

The new logic is not quite backwards compatible; the `path-base` feature now specifies a real path to the actual source directories, rather than a path into the invisible generated `excerpts/` directory with its special structure. Also, a number of features from the previous package that were not actually used in this repository are no longer supported (such as having multiple section names per `#docregion` pragma).

## Pre-launch Checklist

- [x] I signed the [CLA].


Hixie:
If this lands I'll update https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#readme-code

Hixie:
(As far as I can tell this should be exempt from versioning and changelog changes since it's just changing internal repo stuff and not anything that affects downstream users, so I added the labels.)

Hixie:
Proposed new documentation for the wiki:

## README code

All new code samples in `README.md` files must `<?code-excerpt?>` pragmas to manage the code. With `<?code-excerpt?>`, the source of truth for the code is an actual Dart file, which is analyzed, compiled, and tested by our CI. This ensures that it stays updated as the package APIs, Flutter, Dart, and repository analyzer settings change.

### Updating `<?code-excerpt?>`-managed examples

If a code block has a `<?code-excerpt ...?>` tag just before it, it is already using `code-excerpt`. To update it:
1. Find the file referenced in that tag (e.g., `<?code-excerpt "lib/main.dart (AppLifecycle)"?>` comes from `lib/main.dart`; if there is a `<?code-excerpt path-base=""?>` directive earlier in the file, paths are resolved relative to the given `path-base` instead of the markdown file's path).
2. Update the code in that file. If you are fixing a bug, add unit tests or widget tests for the change.
3. Run the `update-excerpts` [repository tool command](https://github.com/flutter/packages/blob/main/script/tool/README.md#update-readmemd-from-example-sources), which will update `README.md`. From the root of the packages repository, the command to run this is: `./script/tool_runner.sh update-excerpts`

### Add or converting a code block

To add a new code block, or fix a legacy code block:
1. Optionally, if this is the first code block to be added, add the following line to the start of the package's `README.md`:

    ```
    <?code-excerpt path-base="example/lib"?>
    ```

    replacing `example/lib` with the directory from which samples will be taken.

2. Find or write the code that you want to use in the example, then annotate it with `#docregion` and `#enddocregion` comments. Ideally the code should be in one of the examples, but you could also put the code in a test. The key is to put the code somewhere that is analyzed and, ideally, tested in CI.

    For example:

    ~~~
    // #docregion purple
    const Color purple = Color(0xFFE6E6FA);
    // #enddocregion purple
    ~~~

3. Reference the code from the `README.md` (or other markdown file where you want the sample to appear):

    ~~~
    <?code-excerpt "{example/lib/-relative filename} ({excerpt name})"?>
    ```dart
    ```
    ~~~

    So if you added `#docregion Foo` to `example/lib/main.dart` and specified `example/lib` as the path base, you would put the following in `README.md`:

    ~~~
    <?code-excerpt "main.dart (Foo)"?>
    ```dart
    ```
    ~~~

4. Follow the steps in the "Updating" section above to automatically fill the new block with the code excerpt.

### Documentation

Samples can come from any file. In XML and HTML files, use `<!--#docregion sectionname-->` and `<!--#enddocregion sectionname-->`. In CSS files, use `/* #docregion sectionname */` and `/* #enddocregion sectionname */`. In YAML files, use `# #docregion sectionname` and `# #enddocregion sectionname`. In C++, Dart, JS, Kotlin, Swift, TypeScript, and other similar languages, use `// #docregion sectionname` and `// #enddocregion sectionname`.

You can extract multiple segments into one sample by just having multiple docregions with the same name. The regions will be concatenated with a language-appropriate comment such as `// ···`. You can control what this looks like by using the `plaster` control, as in:

    <?code-excerpt "main.dart (Foo)" plaster="more..."?>
    ```dart
    ```

To disable this feature, use `plaster="none"`.

The plaster is indented as much as the `#enddocregion` of the region where the splice occurs.

Regions are maximally unindented before being injected into the markdown files (while preserving relative indentation within the block).

stuartmorgan:
> (As far as I can tell this should be exempt from versioning and changelog changes since it's just changing internal repo stuff and not anything that affects downstream users, so I added the labels.)

Yep, the tooling just doesn't have any special-case handling of `<?code-excerpt ...>` tags, so it was flagging it as a README change.

stuartmorgan:
> * Reduces the time to generate the excerpts from about 10 minutes to about 5 seconds.

O.o

Any idea why the package is so slow? Is there something that we should fix upstream to improve the website repo CI?

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
-            f.basename.toLowerCase().endsWith('.md') &&
-            // Exclude CHANGELOG since it should never have excerpts.
-            f.basename != 'CHANGELOG.md')
-        .map((File f) => getRelativePosixPath(f, from: example.directory))
-        .toList();
-    if (relativeMdPaths.isEmpty) {
-      return true;
+  _UpdateResult _updateExcerptsIn(File file) {
+    bool detectedChange = false;
+    List<String>? errors;
```
stuartmorgan, line 102: It looks like making this non-nullable and initialized empty would be simpler.

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -2,18 +2,21 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:io' as io;
-
 import 'package:file/file.dart';
-import 'package:meta/meta.dart';
-import 'package:yaml/yaml.dart';
-import 'package:yaml_edit/yaml_edit.dart';
+import 'package:path/path.dart' as path;
```
stuartmorgan, line 6: In the tooling we use `as p` when we want to import `path`, because the base command that all commands inherit from has a `path` field that is a `path` `Context` that matches the `platform` it's initialized with.

So the logic is tool commands is generally:
- `import 'package:path/path.dart' as p;` if you need it.
- if you just want the current OS's path behavior, use `path.foo`.
- if you want Posix behavior, use `p.posix.foo`

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
-        .writeAsStringSync(pubspecContents);
-
-    // Update the actual pubspec.
-    final YamlEditor editablePubspec = YamlEditor(pubspecContents);
-    const String devDependenciesKey = 'dev_dependencies';
-    final YamlNode root = editablePubspec.parseAt(<String>[]);
-    // Ensure that there's a `dev_dependencies` entry to update.
-    if ((root as YamlMap)[devDependenciesKey] == null) {
-      editablePubspec.update(<String>['dev_dependencies'], YamlMap());
+    String pathBase = path.dirname(file.path);
```
stuartmorgan, line 108: We generally prefer using the `file` package constructs to working with paths manually. So we would generally do `Directory baseDir = file.directory` and then use `childDirectory`, `childFile`, and/or one of the `child*WithSubcomponents` helpers in `file_utils.dart`.

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
+            RegExpMatch? match = _basePattern.firstMatch(line);
+            if (match != null) {
+              pathBase = match.group(1)!;
+            } else {
+              match = _injectPattern.firstMatch(line);
+              if (match != null) {
+                final String excerptPath = match.namedGroup('path')!;
+                final File excerptSourceFile = file.fileSystem.file(
+                    path.normalize(path.join(
+                        path.dirname(file.path), pathBase, excerptPath)));
```
stuartmorgan, line 129: I'm confused how this works; aren't the first two things here *both* `pathBase`? (But this would become `Directory` operations, per the above.)

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
+                final String extension = path.extension(excerptSourceFile.path);
+                switch (extension) {
+                  case '':
+                    language = 'txt';
+                    break;
+                  case '.kt':
+                    language = 'kotlin';
+                    break;
+                  case '.cpp':
```
stuartmorgan, line 138: We also have `.cc` files; the Linux and Windows conventions are, unfortunately, different.

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
-        getRelativePosixPath(repoRoot, from: package.directory);
-    for (final String dependency in submoduleDependencies) {
-      editablePubspec.update(<String>[
-        devDependenciesKey,
-        dependency
-      ], <String, String>{
-        'path': '$relativeRootPath/site-shared/packages/$dependency'
-      });
+    if (detectedChange) {
+      if (errors != null) {
```
stuartmorgan, line 199: With errors being non-null, this would be an `isNotEmpty` check, and we would eliminate the possibility that a later change to the logic above would make `errors` non-null but empty and incorrectly report errors here.

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
+    final List<String> buffer = <String>[];
+    bool extracting = false;
+    int lineNumber = 0;
+    int maxLength = 0;
+    bool found = false;
+    String prefix = '';
+    String suffix = '';
+    String padding = '';
+    switch (language) {
+      case 'c++':
```
stuartmorgan, line 228: And `cc`.

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
+      case 'dart':
+      case 'js':
+      case 'kotlin':
+      case 'rfwtxt':
+      case 'scss':
```
stuartmorgan, line 233: What is this? We don't have any in the repo.

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
+      case 'swift':
+      case 'ts':
```
stuartmorgan, line 235: I really hope nobody is adding typescript files...

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
+        } else {
+          if (trimmedLine == startRegionMarker) {
+            reportError(
+                '${excerptSourceFile.path}:$lineNumber: saw "$startRegionMarker" pragma while already in a "$section" doc region');
+          }
+          if (excerptLine.length > maxLength) {
+            maxLength = excerptLine.length;
+          }
+          if (!excerptLine.contains('#docregion ') &&
+              !excerptLine.contains('#enddocregion ')) {
```
stuartmorgan, line 272: Shouldn't these include `$prefix` technically?

stuartmorgan (changes requested):

script/tool/test/update_excerpts_command_test.dart
```diff
@@ -7,301 +7,87 @@ import 'package:file/file.dart';
...
-
-  setUp(() {
-    fileSystem = MemoryFileSystem();
-    packagesDir = createPackagesDirectory(fileSystem: fileSystem);
-    final MockGitDir gitDir = MockGitDir();
-    when(gitDir.path).thenReturn(packagesDir.parent.path);
-    processRunner = RecordingProcessRunner();
-    final UpdateExcerptsCommand command = UpdateExcerptsCommand(
+Future<void> testInjection(String before, String source, String after, { bool failOnChange = false }) async {
+  final FileSystem fileSystem = MemoryFileSystem();
```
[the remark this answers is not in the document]
reidbaker: +1

stuartmorgan:
> Proposed new documentation for the wiki:

Nice! A few comments, but generally looks good.

> 1. Optionally, if this is the first code block to be added, add the following line to the start of the package's `README.md`:
>    ```
>    <?code-excerpt path-base="example/lib"?>
>    ```
>      
>    replacing `example/lib` with the directory from which samples will be taken.

I think we should say explicitly what it's relative to. It looks from the PR changes like it's the outermost enclosing package?

>    So if you added `#docregion Foo` to `example/lib/main.dart` and specified `example/lib` as the path base, you would put the following in `README.md`:

The earlier step says the path base is optional, so this should say what it's relative to if no base is provided. (I assume the same thing path base is relative to, but it wouldn't hurt to be explicit.)

reidbaker (approved):
Approve % stuarts comments.

script/tool/test/update_excerpts_command_test.dart
```diff
@@ -7,301 +7,87 @@ import 'package:file/file.dart';
...
+  final Directory packagesDir = createPackagesDirectory(fileSystem: fileSystem);
+  final CommandRunner<void> runner = CommandRunner<void>('', '')
+    ..addCommand(UpdateExcerptsCommand(
       packagesDir,
-      processRunner: processRunner,
       platform: MockPlatform(),
```
stuartmorgan, line 23: You should probably run at least some tests with a `MockPlatform(isWindows: true)` now that we have custom path parsing/construction logic in the tool itself, to make sure it's behaving as expected on Windows.

packages/google_identity_services_web/README.md
```diff
@@ -29,7 +29,7 @@ script tag [as recommended](https://developers.google.com/identity/gsi/web/guide
 Place the `script` tag in the `<head>` of your site, next to the script tag that
 loads `flutter.js`, so the browser can downloaded both in parallel:
 
-<?code-excerpt "../../web/index-with-script-tag.html (script-tag)"?>
```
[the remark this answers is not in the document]
Hixie: i mostly tried to do update mechanically, rather than optimize them, because when i changed them i was still trying to just prove to myself that the output wouldn't change. i'm very happy to study these more carefully and make them more useful, i'll do that shortly.

script/tool/lib/src/common/repository_package.dart
```diff
@@ -152,6 +152,17 @@ class RepositoryPackage {
             RepositoryPackage(entity as Directory));
   }
 
+  /// Returns all the subpackages contained in the package, if any, including
+  /// the package itself.
+  Iterable<RepositoryPackage> getSubpackages() {
```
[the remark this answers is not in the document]
Hixie: oops, yes, will remove

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -2,18 +2,21 @@
...
 
 import 'common/core.dart';
 import 'common/package_looping_command.dart';
-import 'common/pub_utils.dart';
 import 'common/repository_package.dart';
 
+class _UpdateResult {
+  const _UpdateResult(this.changed, this.errors);
+  final bool changed;
+  final List<String>? errors;
```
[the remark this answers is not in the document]
Hixie: Good call, fixed.

script/tool/lib/src/update_excerpts_command.dart
Hixie: oh, neat, ok. i'll use the builtin one and remove the import.

script/tool/lib/src/update_excerpts_command.dart
Hixie: the pathBase can be relative (in fact come to think of it it should originally be the empty string, rather than `path.dirname(file.path)`).

I'm happy to convert this to using Directories instead of paths but I'm not quite sure how I would handle relative path bases in that world. Can you elaborate?

script/tool/lib/src/update_excerpts_command.dart
Hixie: added

script/tool/lib/src/update_excerpts_command.dart
stuartmorgan: It's all relative to the `file` parameter, right? So you already have a `File`, and can do everything relative to `file.parent`, unless I'm missing something.

script/tool/lib/src/update_excerpts_command.dart
Hixie: agreed, fixed

script/tool/lib/src/update_excerpts_command.dart
Hixie: added

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -23,232 +26,302 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
...
+      case 'html':
+      case 'xml':
+        prefix = '<!--';
+        suffix = '-->';
+        padding = ' ';
+        break;
+      case 'yaml':
+        prefix = '# ';
+        break;
     }
```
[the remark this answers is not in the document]
Hixie: i mean, if someone tries to grab an excerpt from a binary file, and they can actually get the docregion strings in there... who are we to stop them

script/tool/lib/src/update_excerpts_command.dart
Hixie: dunno, this list was just the list from the original package plus `rfwtxt` (and now `cc`)

script/tool/lib/src/update_excerpts_command.dart
Hixie: sure. added.

script/tool/test/update_excerpts_command_test.dart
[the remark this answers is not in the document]
Hixie: i tried using the old test structure for a while but it drove me batty. i wish we could just entirely remove `setUp` and friends. The twisty non-obvious code execution path with unpredictably state is completely undebuggable IMHO.

That said, if you really want me to try to figure out how to do it that way i can give it a go.

script/tool/test/update_excerpts_command_test.dart
Hixie: added logic to run all the tests with windows as well as the default.

packages/google_identity_services_web/README.md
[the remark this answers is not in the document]
Hixie: i updated the two cases where we used `../` to have more reasonable path bases.

Hixie:
> > * Reduces the time to generate the excerpts from about 10 minutes to about 5 seconds.
> 
> O.o
> 
> Any idea why the package is so slow? Is there something that we should fix upstream to improve the website repo CI?

Well, the old code using build_runner, the Dart compiler, codegen, stored all the extracts on disk, and had two separate processes for extracting and injecting. The new code does none of that. Not running code is faster than running code.

I'm not sure if it makes sense to upstream this because I don't fully understand the website repo's use case so it's hard to say whether this code addresses the same needs.

script/tool/lib/src/update_excerpts_command.dart
Hixie: Oh, I see. It wasn't intuitive to me that `directory.childFile` would take a relative path. As in, `directory.childFile('../../foo/bar.test')`. Seems like `childFile` is a weird name for that. :-)

Hixie: updated code as suggested

Hixie: only remaining use of `path` is for `path.extension`.

Hixie:
> > Proposed new documentation for the wiki:
> 
> Nice! A few comments, but generally looks good.

Added a bit about path-base's default.

Hixie:
updated, PTAL.

domesticmouse (approved):

stuartmorgan (approved):
LGTM, thanks!

I saw from your comments in the `rfw` PR that we may need to iterate more for that specific case, but I think we should land this as-is and then iterate from there in follow-ups since this is already a big quality of life improvement over the current implementation.

Hixie:
I will remove `scss` and `ts` and land it. Thanks for the reviews!

PR #4428, opened 2023-07-10
[ci] Add initial LUCI legacy analysis

stuartmorgan:
Adds new legacy analysis targets to LUCI in bringup mode.

Also updates the repo tooling to use a Flutter->Dart version mapping instead of having to separately maintain a Dart version, now that we already have such a map in the tool anyway.


.ci/targets/analyze.yaml
```diff
@@ -12,4 +12,4 @@ tasks:
     # the changes won't break analysis of other packages in the respository
     # that depend on it.
   - name: analyze - pathified
-    script: .ci/scripts/pathified_analyze.sh
```
stuartmorgan, line 15: I renamed this just so the variants of the `analyze` scripts would be grouped together in file lists, as it makes navigating the scripts easier.

tarrinneal (approved):

PR #4429, opened 2023-07-10
[web_benchmarks] Migrate custom test to `custom_package_tests`

stuartmorgan:
Removes the bespoke CI task for `web_benchmarks`, and sets up a custom test script to drive it, causing it to run in the existing `custom_package_tests` CI task.

For now it's Linux-only to preserve the existing behavior; anyone is welcome to enable it for other packages in the future after ensuring it works in those environments.


yjbanov (approved):
lgtm

PR #4437, opened 2023-07-11
[ci] Remove unused Chromium setup

stuartmorgan:
The tasks that used the Cirrus template and shell script to install a specific version of Chromium have been migrated to LUCI, so this is cruft; I forgot to remove it when finishing that migration.

Also combines two setup steps into one template now that they are only used together.


ditman (approved):
Good riddance!

PR #4443, opened 2023-07-11
[pigeon] Adds package name to method channel.

tarrinneal:
Adds package name to method channel name strings. The name can be added via command line, pigeon configuration, or automatically using `deducePackageName`.

This pr also changes the `generate` function to format by default and adds an `n` command to not format if needed.

## Pre-launch Checklist

- [x] I signed the [CLA].


tarrinneal:
broke up the commits to make reviewing easier.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 10.1.4
+
+* Adds package name to method channel strings.
```
stuartmorgan, line 3: Let's specifically mention "to avoid potential collisions between plugins."

packages/pigeon/example/README.md
```diff
@@ -31,6 +31,7 @@ needed for your project.
   // Set this to a unique prefix for your plugin or application, per Objective-C naming conventions.
   objcOptions: ObjcOptions(prefix: 'PGN'),
   copyrightHeader: 'pigeons/copyright.txt',
+  packageName: 'pigeonExamplePackage',
```
stuartmorgan, line 34: `pigeon_example_package`; since the default is a package name, we should follow package naming conventions for consistency.

packages/pigeon/lib/generator_tools.dart
```diff
@@ -163,8 +165,9 @@ class Indent {
 }
 
 /// Create the generated channel name for a [func] on a [api].
-String makeChannelName(Api api, Method func) {
-  return 'dev.flutter.pigeon.${api.name}.${func.name}';
+String makeChannelName(Api api, Method func, String? packageName) {
```
stuartmorgan, line 168: What are the circumstances where this would be null?

packages/pigeon/lib/pigeon_lib.dart
```diff
@@ -329,6 +335,16 @@ class PigeonOptions {
   PigeonOptions merge(PigeonOptions options) {
     return PigeonOptions.fromMap(mergeMaps(toMap(), options.toMap()));
   }
+
+  /// Returns provided or deduced package name, throws `Exception` if none found.
+  String getPackageNameOrThrow() {
```
stuartmorgan, line 340: Generally Dart exceptions aren't encoded into the name; this can just be `getPackageName`.

packages/pigeon/lib/pigeon_lib.dart
```diff
@@ -329,6 +335,16 @@ class PigeonOptions {
...
+    final String? name = _packageName ?? deducePackageName(input ?? '');
+    if (name == null) {
+      throw Exception(
+          'Unable to deduce package name, and no package name supplied.');
```
stuartmorgan, line 344: Let's explicitly add instructions about using the flag, so that someone seeing this gets clear next steps about how to fix it.

stuartmorgan (changes requested):

packages/pigeon/lib/generator_tools.dart
tarrinneal: Shouldn't happen.

packages/pigeon/lib/swift_generator.dart
```diff
@@ -268,7 +274,8 @@ import FlutterMacOS
         final _SwiftFunctionComponents components =
             _SwiftFunctionComponents.fromMethod(func);
 
-        final String channelName = makeChannelName(api, func);
+        final String channelName =
+            makeChannelName(api, func, generatorOptions.packageName!);
```
stuartmorgan, line 278: Can we make the fields non-nullable so we're not force-unwrapping them everywhere, and instead ensure that they are non-null at the point where we populate them (which is where that logic actually lives)? Currently it's very hard to tell if/why this is safe because the unwrapping is so far from the logic that populates it.

tarrinneal: I'll look at it, in general I'm trying to keep all pigeonOptions nullable to match all other inputs and allow for consistent logic on the `merge` methods.

stuartmorgan: Maybe we should pass the package name as a separate non-nullable argument into the generators then, rather than relying on the options to pass it?

Fundamentally it seems the issue here is that we're using the same object for user-facing options (where it's an optional parameter) and internal implementation (where have some value is required).

tarrinneal: I can change the non user faces options, if that makes the most sense

packages/pigeon/lib/cpp_generator.dart
```diff
@@ -671,6 +750,8 @@ class CppSourceGenerator extends StructuredGenerator<CppOptions> {
                 .map((Class x) => x.name)
                 .contains(field.type.baseName)) {
           return '${hostDatatype.datatype}::FromEncodableList(std::get<EncodableList>($encodable))';
+        } else if (field.type.baseName == 'Object') {
+          return encodable;
```
stuartmorgan, line 754: Isn't this from the other PR?

stuartmorgan (approved):
LGTM (with one question).

I hadn't thought through quite how much boilerplate plumbing my suggestion would involve, sorry :( If we ever have to do something like this again, we should probably add an internal-facing context object to contain these kinds of params.

tarrinneal:
> I hadn't thought through quite how much boilerplate plumbing my suggestion would involve, sorry :( If we ever have to do something like this again, we should probably add an internal-facing context object to contain these kinds of params.

It's not too late to make this change.

tarrinneal:
@stuartmorgan thoughts on the javaOptions.package and kotlinOptions.package params? Should they be removed or deprecated (or nothing)? I didn't want to assume that replacing them with the new system would be good for everyone.

stuartmorgan:
Dart package and Java/Kotlin package are orthogonal, so we should definitely not remove the latter. We may want to rename the parameter for the former `dartPackage` to avoid confusion.

PR #4445, opened 2023-07-12
[file_selector] Avoids using path_provider in web example app.

ditman:
This is a minor change to the `file_selector` example app to prevent it from using `path_provider` when running on the web (because `path_provider` isn't available on the web).

Fixes https://github.com/flutter/flutter/issues/130368

(PS: This should be exempt from publishing requirements, because the modified file is not shown in pub.dev, or anywhere else AFAIK.)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/file_selector/file_selector/example/lib/save_text_page.dart
+++ b/packages/file_selector/file_selector/example/lib/save_text_page.dart
@@ -5,7 +5,7 @@
 import 'package:file_selector/file_selector.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
-import 'package:path_provider/path_provider.dart';
+import 'package:path_provider/path_provider.dart' as path_provider;
 
 /// Page for showing an example of saving with file_selector
 class SaveTextPage extends StatelessWidget {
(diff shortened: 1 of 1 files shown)


stuartmorgan:
> This should be exempt from publishing requirements

Not only is it, but the tooling knows that :)

(It does ask for a changelog for any example change in case that would be interesting to clients even though it's not on the pub.dev page, but I agree this one isn't.)

packages/file_selector/file_selector/example/lib/save_text_page.dart
```diff
@@ -22,8 +22,9 @@ class SaveTextPage extends StatelessWidget {
     // This demonstrates using an initial directory for the prompt, which should
     // only be done in cases where the application can likely predict where the
     // file will be saved. In most cases, this parameter should not be provided.
-    final String initialDirectory =
-        (await getApplicationDocumentsDirectory()).path;
+    final String? initialDirectory = !kIsWeb
```
stuartmorgan, line 25: Optional nit: I find double negation harder to parse; I would prefer this as `kIsWeb ? null : ...`.

stuartmorgan (approved):
LGTM! Thanks for fixing this; I actually noticed this while adjusting something else in here recently, but forgot to follow up with a PR for it.

packages/file_selector/file_selector/example/lib/save_text_page.dart
ditman: Yes, I thought about this too! Let me fix it.

ditman:
@stuartmorgan I removed the `-` root file in this PR.

PR #4446, opened 2023-07-12
[ci] Move snippet checks to LUCI

stuartmorgan:
Moves the check that README snippets using code excerpting are up to date to LUCI. Now that the check has been rewritten to be extremely fast, it's folded into the existing repo checks instead of being a separate task.

Also adjusts the tooling slightly so that it logs the count of snippets checked, for auditability in CI, like the old version of the excerpt update tooling did.

### DIFF
--- a/script/tool/lib/src/update_excerpts_command.dart
+++ b/script/tool/lib/src/update_excerpts_command.dart
@@ -63,7 +63,7 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
       if (result.snippetCount > 0) {
         final String displayPath =
             getRelativePosixPath(file, from: package.directory);
-        print('${indentation}Checked ${result.snippetCount} snippets in '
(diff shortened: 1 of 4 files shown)


script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -165,7 +178,9 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
               errors.add(
                   '${file.path}:$lineNumber: code block was followed by a space character instead of the language (expected "$language")');
               mode = _ExcerptParseMode.injecting;
-            } else if (line != '```$language' && line != '```rfwtxt' && line != '```json') {
+            } else if (line != '```$language' &&
+                line != '```rfwtxt' &&
+                line != '```json') {
```
stuartmorgan, line 183: It turns out we're not enforcing formatting for the `script/tool/` directory; I'll file an issue to fix that.

bparrishMines (approved):
LGTM

PR #4285, opened 2023-06-23
[image_picker] Update to 1.0

stuartmorgan:
The plugin now covers all of our supported platforms, most highly-requested features have been addressed, and the API has been stable for quite some time (in that no breaking changes have been required). Given that, a 1.0 version number reflects the current state of the package than a 0.x version.

As part of the breaking change, the `get*` methods that were deprecated several years ago have now been removed.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ ] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #3683, opened 2023-04-11
[file_selector_web] Listens to file input cancel event.

raju-muliyashiya:
While using package ``file_selector`` making for multiple file request in web platform we are not able to get any data if user clicks cancel button on file selection window. This PR fixes it by watching the ``focus event`` and if ``onChange`` is not fired then return empty array, Empty array will only return if multiple selection is enabled with ``openFiles()``.

*List which issues are fixed by this PR. You must list at least one issue.*
[Issue #121328](https://github.com/flutter/flutter/issues/121328)

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #4453, opened 2023-07-13
[image_picker_web] Listens to file input cancel event.

ditman:
## Changes

This PR listens to the `cancel` event from the `input type=file` used by the web implementation of the image_picker plugin, so apps don't end up endlessly awaiting for a file that will never come **in modern browsers** (Chrome 113, Safari 16.4, or newer). _Same API as https://github.com/flutter/packages/pull/3683._

Additionally, this PR:

* Removes all code and tests mentioning `PickedFile`. (Deprecated years ago, and unused since https://github.com/flutter/packages/pull/4285) **(Breaking change)**
* Updates README to mention `XFile` which is the current return type of the package.
* Updates the dependency on `image_picker_platform_interface` to `^2.9.0`.
  * Implements all non-deprecated methods from the interface, and makes deprecated methods use the fresh ones.
  * Updates tests.

### Issues

* Fixes https://github.com/flutter/flutter/issues/92176

### Testing

* Added integration testing coverage for the 'cancel' event.
* Tested manually in Chrome with the example app running on web.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/image_picker/image_picker_for_web/CHANGELOG.md
+++ b/packages/image_picker/image_picker_for_web/CHANGELOG.md
@@ -1,10 +1,9 @@
-## 2.3.0
+## 3.0.0
 
+* **BREAKING CHANGE:** Removes all code and tests mentioning `PickedFile`.
 * Listens to `cancel` event on file selection. When the selection is canceled:
   * `Future<XFile?>` methods return `null`
   * `Future<List<XFile>>` methods return an empty list.
-* Removes all code and tests mentioning `PickedFile`.
-* Updates README to mention `XFile`.
 
 ## 2.2.0
 
--- a/packages/image_picker/image_picker_for_web/CHANGELOG.md
+++ b/packages/image_picker/image_picker_for_web/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 2.3.0
+
+* Listens to `cancel` event on file selection. When the selection is canceled:
+  * `Future<XFile?>` methods return `null`
+  * `Future<List<XFile>>` methods return an empty list.
+* Removes all code and tests mentioning `PickedFile`.
+* Updates README to mention `XFile`.
+
 ## 2.2.0
 
 * Adds `getMedia` method.
--- a/packages/image_picker/image_picker_for_web/README.md
+++ b/packages/image_picker/image_picker_for_web/README.md
@@ -70,7 +70,7 @@ should add it to your `pubspec.yaml` as usual.
 You should be able to use `package:image_picker` _almost_ as normal.
 
 Once the user has picked a file, the returned `XFile` instance will contain a
-`network`-accessible Blob URL (pointing to a location within the browser).
+`network`-accessible `Blob` URL (pointing to a location within the browser).
 
 The instance will also let you retrieve the bytes of the selected file across all platforms.
 
--- a/packages/image_picker/image_picker_for_web/README.md
+++ b/packages/image_picker/image_picker_for_web/README.md
@@ -4,23 +4,9 @@ A web implementation of [`image_picker`][1].
 
 ## Limitations on the web platform
 
-Since Web Browsers don't offer direct access to their users' file system,
-this plugin provides a `PickedFile` abstraction to make access uniform
-across platforms.
+This plugin uses `XFile` objects to abstract files picked/created by the user.
 
-The web version of the plugin puts network-accessible URIs as the `path`
-in the returned `PickedFile`.
-
-### URL.createObjectURL()
-
-The `PickedFile` object in web is backed by [`URL.createObjectUrl` Web API](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL),
-which is reasonably well supported across all browsers:
-
-![Data on support for the bloburls feature across the major browsers from caniuse.com](https://caniuse.bitsofco.de/image/bloburls.png)
-
-However, the returned `path` attribute of the `PickedFile` points to a `network` resource, and not a
-local path in your users' drive. See **Use the plugin** below for some examples on how to use this
-return value in a cross-platform way.
+Read more about `XFile` on the web in [`package:cross_file`'s README](https://pub.dev/packages/cross_file)
 
 ### input file "accept"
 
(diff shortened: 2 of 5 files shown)


packages/image_picker/image_picker_for_web/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 2.3.0
+
+* Listens to `cancel` event on file selection. When the selection is canceled:
+  * `Future<XFile?>` methods return `null`
+  * `Future<List<XFile>>` methods return an empty list.
+* Removes all code and tests mentioning `PickedFile`.
```
ditman, line 6: This has been only around for a few weeks, but I don't want to postpone the cleanup much longer. Should I make this PR a major version bump, so it's guaranteed that nobody ends up in a state of having an out-of-date core plugin, but a "too new" web impl?)

stuartmorgan: Sorry, I thought I'd responded to this already. Removing deprecated code is a breaking change, so requires a major version bump, otherwise you will actively break people with older versions of the app-facing package.

> but I wouldn't want to postpone this cleanup too much.

We haven't done any of the cleanup in the lower layers; we should do it bottom-up for the non-user-facing parts.

> Or can I assume that people migrated to the new API years ago, with the deprecation?

You can't assume anything that the resolver will allow to happen won't happen :)

packages/image_picker/image_picker_for_web/README.md
```diff
@@ -4,23 +4,9 @@ A web implementation of [`image_picker`][1].
...
-
-The `PickedFile` object in web is backed by [`URL.createObjectUrl` Web API](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL),
-which is reasonably well supported across all browsers:
-
-![Data on support for the bloburls feature across the major browsers from caniuse.com](https://caniuse.bitsofco.de/image/bloburls.png)
-
-However, the returned `path` attribute of the `PickedFile` points to a `network` resource, and not a
-local path in your users' drive. See **Use the plugin** below for some examples on how to use this
-return value in a cross-platform way.
+Read more about `XFile` on the web in [`package:cross_file`'s README](https://pub.dev/packages/cross_file)
```
stuartmorgan, line 9: (These README changes are all fine; we don't need to still README the deprecated implementation APIs.)

stuartmorgan (changes requested):

packages/image_picker/image_picker_for_web/README.md
ditman: > we don't need to still README the deprecated implementation APIs.

Having trouble parse that bit? Please rephrase?

ditman:
I've updated the README to mention the current methods (not deprecated ones), and marked this change as a major version to prevent any issues with package resolution (I'll send a PR updating `image_picker` to v.3 of the web implementation after this lands). PTAL @stuartmorgan!

packages/image_picker/image_picker_for_web/CHANGELOG.md
stuartmorgan: I checked, and we haven't even formally deprecated the old methods in the platform interface package, which makes removing the implementations here worrying since we aren't doing anything to ensure we don't call it anywhere.

I think we should deprecate the methods first in a prequel PR, and then update the min version of the platform interface this package requires to the one that deprecates them. I'll make that PR now, then we can update this one after that lands.

ditman: @stuartmorgan the methods may not be deprecated on the `platform_interface`, but everything that returned a `PickedFile` was marked as deprecated ~2 years ago in the core plugin https://github.com/flutter/plugins/pull/4073, and eventually removed here https://github.com/flutter/packages/pull/4285, right?

AFAIK, nothing that returns a `PickedFile` is being called anymore by the plugin as of version `^1.0` (could it? yes)

(This deprecation started on the front-end, of the package, I guess :P)

As for landing this in sequence with the deprecation in the platform_interface: no problem either! :)

stuartmorgan: > @stuartmorgan the methods may not be deprecated on the `platform_interface`, but everything that returned a `PickedFile` was marked as deprecated ~2 years ago in the core plugin [flutter/plugins#4073](https://github.com/flutter/plugins/pull/4073), and eventually removed here #4285, right?
> 
> AFAIK, nothing that returns a `PickedFile` is being called anymore by the plugin as of version `^1.0` (could it? yes)

Correct, it's not that we are calling this in the app-facing package's implementation, it's that there's nothing structural that would prevent us from calling it in the future (and as I found when I tried to do the deprecation, nothing preventing other locations in our codebase from calling it).

> (This deprecation started on the front-end, of the package, I guess :P)

There are two distinct API layers: the client-facing API, and the platform interface API. Deprecating things in the client-facing API is unrelated in general to deprecating things in the platform interface (although they happen to be more closely linked than usual here).

My concern is that the platform implementations should be implementing the platform interface API as it is expressed formally, rather than tailoring the implementations to how current clients are calling it.

packages/image_picker/image_picker_for_web/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: image_picker_for_web
 description: Web platform implementation of image_picker
 repository: https://github.com/flutter/packages/tree/main/packages/image_picker/image_picker_for_web
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+image_picker%22
-version: 2.2.0
+version: 3.0.0
```
stuartmorgan, line 6: Let's update the `image_picker_platform_interface` to `^2.9.0` so that this can't be used with a version that hasn't deprecated the methods you removed.

packages/image_picker/image_picker_for_web/README.md
```diff
@@ -65,8 +68,8 @@ should add it to your `pubspec.yaml` as usual.
 
 You should be able to use `package:image_picker` _almost_ as normal.
 
-Once the user has picked a file, the returned `PickedFile` instance will contain a
-`network`-accessible URL (pointing to a location within the browser).
+Once the user has picked a file, the returned `XFile` instance will contain a
+`network`-accessible Blob URL (pointing to a location within the browser).
```
stuartmorgan, line 72: Nit: Should that be `Blob`?

stuartmorgan (approved):
LGTM modulo the version constraint.

packages/image_picker/image_picker_for_web/README.md
ditman: Yeah, MDN uses [both](https://developer.mozilla.org/en-US/docs/Web/API/Blob), but the non-monospace version looks like exceptions to the rule.

ditman:
Hey @stuartmorgan, here's the refresh to the `image_picker_web` to implement the latest platform interface! PTAL (and change/merge at will)

stuartmorgan (approved):
StillLGTM

PR #4472, opened 2023-07-14
[webview_flutter] Add a method for getting the user agent

bparrishMines:
Aggregate PR for adding a method for retrieving the user agent. 

Internal: b/272582342

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_android/test/android_webview_controller_test.dart
```diff
@@ -1056,6 +1056,17 @@ void main() {
...
       verify(mockSettings.setUserAgentString('Test Framework')).called(1);
     });
+
+    test('getUserAgent', () async {
+      final MockWebSettings mockSettings = MockWebSettings();
+      final AndroidWebViewController controller = createControllerWithMocks(
+        mockSettings: mockSettings,
+      );
+      when(mockSettings.getUserAgentString())
+          .thenAnswer((_) => Future<String?>.value('str'));
```
stuartmorgan, line 1066: Nit: It would be clearer to use a `const String` local variable for this and the expectation, to make it obvious that what's expected is that the mock value is being returned exactly.

packages/webview_flutter/webview_flutter_wkwebview/test/src/web_kit/web_kit_test.dart
```diff
@@ -833,14 +833,21 @@ void main() {
...
           webViewInstanceId,
           'hello',
         ));
       });
 
+      test('getCustomUserAgent', () {
+        when(
+          mockPlatformHostApi.getCustomUserAgent(webViewInstanceId),
+        ).thenReturn('str');
+        expect(webView.getCustomUserAgent(), completion('str'));
```
stuartmorgan, line 848: See comment in the Android test about using a local constant.

packages/webview_flutter/webview_flutter_wkwebview/test/webkit_webview_controller_test.dart
```diff
@@ -823,6 +823,19 @@ void main() {
...
+      final MockWKWebView mockWebView = MockWKWebView();
+
+      final WebKitWebViewController controller = createControllerWithMocks(
+        createMockWebView: (_, {dynamic observeValue}) => mockWebView,
+      );
+
+      when(mockWebView.getCustomUserAgent()).thenAnswer(
+        (_) => Future<String?>.value('str'),
+      );
+      expect(controller.getUserAgent(), completion('str'));
```
stuartmorgan, line 836: See comment in the Android test about using a local constant.

stuartmorgan, line 836: See comment in the Android test about using a local constant.

packages/webview_flutter/webview_flutter/test/webview_controller_test.dart
```diff
@@ -387,6 +387,20 @@ void main() {
...
+        MockPlatformWebViewController();
+    when(mockPlatformWebViewController.getUserAgent()).thenAnswer(
+      (_) => Future<String?>.value('str'),
+    );
+
+    final WebViewController webViewController = WebViewController.fromPlatform(
+      mockPlatformWebViewController,
+    );
+
+    await expectLater(webViewController.getUserAgent(), completion('str'));
```
stuartmorgan, line 402: See comment in the Android test about using a local constant.

packages/webview_flutter/webview_flutter_android/test/android_webview_test.dart
```diff
@@ -527,6 +527,12 @@ void main() {
           100,
         ));
       });
+
+      test('getUserAgentString', () async {
+        when(mockPlatformHostApi.getUserAgentString(webSettingsInstanceId))
+            .thenReturn('str');
+        expect(await webSettings.getUserAgentString(), 'str');
```
stuartmorgan, line 534: See comment in the other Android test about using a local constant.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebSettingsHostApiImpl.java
```diff
@@ -132,4 +132,11 @@ public void setTextZoom(@NonNull Long instanceId, @NonNull Long textZoom) {
     final WebSettings webSettings = Objects.requireNonNull(instanceManager.getInstance(instanceId));
     webSettings.setTextZoom(textZoom.intValue());
   }
+
+  @Nullable
```
stuartmorgan, line 136: Does the underlying API really return a nullable string?

packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/WebSettingsTest.java
```diff
@@ -117,4 +118,10 @@ public void setTextZoom() {
     testHostApiImpl.setTextZoom(0L, 100L);
     verify(mockWebSettings).setTextZoom(100);
   }
+
+  @Test
+  public void getWebSettings() {
+    when(mockWebSettings.getUserAgentString()).thenReturn("str");
+    assertEquals(testHostApiImpl.getUserAgentString(0L), "str");
```
stuartmorgan, line 125: See comment in the other Android test about using a local constant.

packages/webview_flutter/webview_flutter_wkwebview/example/ios/RunnerTests/FWFWebViewHostApiTests.m
```diff
@@ -480,4 +480,20 @@ - (void)testSetInspectable API_AVAILABLE(ios(16.4), macos(13.3)) {
...
+
+  FWFInstanceManager *instanceManager = [[FWFInstanceManager alloc] init];
+  [instanceManager addDartCreatedInstance:mockWebView withIdentifier:0];
+
+  FWFWebViewHostApiImpl *hostAPI = [[FWFWebViewHostApiImpl alloc]
+      initWithBinaryMessenger:OCMProtocolMock(@protocol(FlutterBinaryMessenger))
+              instanceManager:instanceManager];
+
+  FlutterError *error;
+  XCTAssertEqualObjects([hostAPI userAgentForWebViewWithIdentifier:@0 error:&error], @"str");
```
stuartmorgan, line 496: See comment in the Android test about using a local constant.

packages/webview_flutter/webview_flutter_wkwebview/pigeons/web_kit.dart
```diff
@@ -697,6 +697,9 @@ abstract class WKWebViewHostApi {
 
   @ObjCSelector('setInspectableForWebViewWithIdentifier:inspectable:')
   void setInspectable(int identifier, bool inspectable);
+
+  @ObjCSelector('userAgentForWebViewWithIdentifier:')
```
stuartmorgan, line 701: This should be `customUserAgentForWebViewWithIdentifier`, since it doesn't give you the actual user agent.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
```diff
@@ -558,6 +558,9 @@ class WebKitWebViewController extends PlatformWebViewController {
   Future<void> setInspectable(bool inspectable) {
     return _webView.setInspectable(inspectable);
   }
+
+  @override
+  Future<String?> getUserAgent() => _webView.getCustomUserAgent();
```
stuartmorgan, line 563: These aren't equivalent; the RHS returns null unless a custom user agent has been set, so this implementation would give people the wrong answer on iOS (except in cases where they set it themselves, but they probably don't need to ask then). It looks like we need to fall back to using JS when this returns null.

stuartmorgan: This issue also suggests we need an integration test that validates that we get a non-null value when we haven't set a custom agent.

stuartmorgan (changes requested):

ditman:
I'll keep an eye out for when this lands, the `_web` version of this should be very simple to add!

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebSettingsHostApiImpl.java
bparrishMines: I tested this out and it looks like it is always nonnull. Updated.

stuartmorgan (approved):
LGTM

packages/webview_flutter/webview_flutter_android/test/legacy/webview_android_widget_test.mocks.dart
```diff
@@ -303,6 +303,14 @@ class MockWebSettings extends _i1.Mock implements _i2.WebSettings {
...
 
   @override
+  _i5.Future<String> getUserAgentString() => (super.noSuchMethod(
+        Invocation.method(
+          #getUserAgentString,
+          [],
+        ),
+        returnValue: _i5.Future<String>.value(''),
+      ) as _i5.Future<String>);
+  @override
```
bparrishMines, line 313:
```suggestion
```

PR #4489, opened 2023-07-16
Remove `equatable` and `xml` allowances

stuartmorgan:
The only package that was using the `equatable` package removed it, so we no longer need this allowance.

The only package depending on `xml` was `flutter_migrate`, but it never actually used it, so that can be removed as well.

Part of https://github.com/flutter/flutter/issues/122713


bparrishMines (approved):
LGTM

PR #4494, opened 2023-07-17
[ci] Enable Android emulator-based tests

stuartmorgan:
Adds emulators to the Android platform test tasks, and enable both native integration tests and `flutter test` Dart integration tests. This finally gives us presubmit coverage of Android integration tests for non-team-member PRs, as well as giving us a path to reducing the dependency on FTL for all of our Android testing.

Currently this uses Android 13 emulators; we could expand this to a set of legacy version emulator tests in the future once this is in place.

Some specific notes:
- Camera and Google Maps tests are currently excluded, so that we can enable everything else while we work on the blockers for those.
- `shared_preferences` was missing its debug entitlement, which prevented the debug service from running, and thus broke `flutter test`; we'd just never noticed because FTL tests don't need that mechanism to work.
- IAP was using `test` instead of `testWidget`, which we only check for in `drive-examples`, so hadn't noticed.

Fixes https://github.com/flutter/flutter/issues/93810


packages/shared_preferences/shared_preferences/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## NEXT
 
+* Fixed the example app to be debuggable on Android.
```
tarrinneal, line 3: fixes

tarrinneal (approved):

packages/shared_preferences/shared_preferences/CHANGELOG.md
stuartmorgan: 🤦🏻

reidbaker (approved):

reidbaker (approved):
Are the skipped test being run in the older system or are they not being run?

gmackall (approved):
IAP changes lgtm

stuartmorgan:
> Are the skipped test being run in the older system or are they not being run?

All FTL tests are still being run, this particular change is purely additive. We'll figure out what we want to do with FTL as a follow-up.

PR #4513, opened 2023-07-18
[ci] Roll Flutter to f842ed91

stuartmorgan:
Rolls Flutter `master` to f842ed916514879fe6898b2a5a4053c63c3308fe manually.

The switch to Material 3 by default broke the expectations of the `dynamic_layouts` package; this updates them to handle the values reported starting with that switch.


stuartmorgan:
@Piinks I bisected to the M3 switch commit, so I'm sure that's the cause. I'm assuming that means these changes are expected, but if that's not the case I'd like to spin that out into an issue for follow-up so that I can get the roller unblocked.

packages/dynamic_layouts/example/test/wrap_example_test.dart
```diff
@@ -23,10 +22,21 @@ void main() {
     expect(find.text('Index 3'), findsOneWidget);
     expect(find.text('Index 4'), findsOneWidget);
 
+    // Material 3 changes the expected layout positioning.
+    final bool usesMaterial3 = (app.theme ?? ThemeData.light()).useMaterial3;
```
Piinks, line 26: Yes, I double checked. The slight change in offset is expected.

Piinks (approved):
This LGTM - thank you for fixing it!

packages/dynamic_layouts/example/test/wrap_example_test.dart
stuartmorgan: Perfect, thanks for verifying!

stuartmorgan:
Ugh, looks like web has slightly different offsets. I'll adjust accordingly for web...

Piinks:
Oh man. @HansMuller is there meant to be such discrepancies with Material 3 across platforms? I would expect the pixels to be the same when building the same across platforms. It was the same when this was tested in M2. The disparity here is coming from the text, is there a different visual density across platforms for Material 3?

packages/dynamic_layouts/example/test/wrap_example_test.dart
```diff
@@ -23,10 +23,29 @@ void main() {
...
-    expect(tester.getTopLeft(find.text('Index 4')), const Offset(381.0, 224.0));
+    expect(tester.getTopLeft(find.text('Index 0')), offset0);
+    expect(tester.getTopLeft(find.text('Index 1')), offset1);
+    expect(tester.getTopLeft(find.text('Index 3')), offset3);
+    expect(tester.getTopLeft(find.text('Index 4')), offset4);
   });
 }
+
+double _getExpectedYOffset(double nonWeb) {
+  return kIsWeb ? nonWeb - 0.5 : nonWeb;
```
ditman, line 50: Yikes! Is this a "web is rounding wrong" bug?

ditman: I brought this up to the team chat as well!

stuartmorgan:
I'm going to go ahead and autosubmit this as-is since our roller is many days behind and I want to unblock it; we can definitely iterate on how exactly we handle the web diff (and investigate the framework if that's a concern) after the fact.

packages/dynamic_layouts/example/test/wrap_example_test.dart
yjbanov: I vaguely recall we had a special-case for an off-by-one issue in text for Firefox, but not generally. @mdebbar might know more.

PR #4517, opened 2023-07-19
[go_router_builder] Removes `path_to_regexp` from the dependencies 

ValentinVignal:
Fixes https://github.com/flutter/flutter/issues/130817
Relates to https://github.com/flutter/flutter/issues/122713

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/go_router_builder/lib/src/utils/path_token.dart
+++ b/packages/go_router_builder/lib/src/utils/path_token.dart
@@ -1,0 +1,61 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// The base type of all tokens produced by a path specification.
+abstract class Token {
+  /// The name of the parameter.
+  String get name;
+
+  /// Parses a [path] specification.
+  static List<Token> parse(String path) => _parse(path);
+}
+
+/// Corresponds to a parameter of a path specification.
+class ParameterToken implements Token {
+  /// Creates a parameter token for [name].
+  ParameterToken(this.name);
+
+  /// The parameter name.
+  @override
+  final String name;
+}
+
+/// Corresponds to a non-parameterized section of a path specification.
+class PathToken implements Token {
+  /// Creates a path token with [name].
+  PathToken(this.name);
+
+  /// A substring of the path specification.
+  @override
+  final String name;
+}
+
+/// The regular expression used to extract parameters from a path specification.
+///
+/// Capture groups:
+///   1. The parameter name.
+///   2. An optional pattern.
+final RegExp _parameterRegExp = RegExp(
+  /* (1) */ r':(\w+)'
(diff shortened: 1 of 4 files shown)


packages/go_router_builder/lib/src/utils/path_token.dart
ValentinVignal, line 1: This file is very much inspired by `path_to_regexp` library, I don't know if I'm allowed to do that

ValentinVignal:
I believe this PR can have a no-test exemption. This is a revamp and the current tests should cover it already

packages/go_router_builder/lib/src/utils/path_token.dart
stuartmorgan: What does "inspired by" mean? If you read the source of that package as part of writing this implementation, then no, that would disqualify you from contributing an implementation (in which case this PR should be closed).

ValentinVignal: I did read the package source code to understand what the method was actually happening and how `go_router_builder` was using it.

I guess the PR cannot be merged as it is right now. Before closing it, let me try to clean it out and re-write something else.
But `go_router_builder` already has its methods built around it so it might end up looking similar anyway

ValentinVignal: So I cannot work on this issue anymore? 😕 

What if I find another way to code it that is not similar to the package? Would that work?

ValentinVignal: > That is correct. We need a clean room implementation of the functionality this package needs, and you cannot provide that.

Even if as I suggested the implementation's logic is completely different? What would be the issue with that since it would explicitly not be inspired by the package?

chunhtai:
we already has an in house implementation of this package https://github.com/flutter/packages/blob/main/packages/go_router/lib/src/path_utils.dart you can just copy over

ValentinVignal:
> we already have an implementation of the path util https://github.com/flutter/packages/blob/main/packages/go_router/lib/src/path_utils.dart you can just copy over

This is what I started to do at first but go_router_builder's needs are simpler so I felt it was overkilled. I can do that and stip out `path_utils.dart` to only what we need. But after @stuartmorgan 's comments, I don't know if I'll be allowed to make a PR

packages/go_router_builder/lib/src/utils/path_token.dart
stuartmorgan: > Even if as I suggested the implementation's logic is completely different? What would be the issue with that since it would explicitly not be inspired by the package?

I'm not a lawyer, and I have no interest in unnecessarily involving lawyers in the process of fixing the referenced issue.

ValentinVignal:
@chunhtai @stuartmorgan I've made another PR that should comply with what you want.
I copied `path_utils.dart` and only removed some methods from it or simplify some methods that were doing extra logic I didn't need. 
I explained my steps with some comments on my commits

PR #4519, opened 2023-07-19
[ci] Shard Windows Dart unit tests

stuartmorgan:
The test currently takes about 45 minutes; this starts the process of spliting it into two shards, as on Linux, to bring it more in line with the other repo test durations.


bparrishMines (approved):
LGTM

PR #4520, opened 2023-07-19
[image_picker] Deprecate platform interface methods

stuartmorgan:
Deprecates all of the platform interface methods that have been replaced with newer variants.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
+++ b/packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
@@ -37,8 +37,6 @@ abstract class ImagePickerPlatform extends PlatformInterface {
     _instance = instance;
   }
 
-  // Next version of the API.
-
   /// Returns a [PickedFile] with the image that was picked.
   ///
   /// The `source` argument controls where the image comes from. This can
@@ -67,6 +65,7 @@ abstract class ImagePickerPlatform extends PlatformInterface {
   /// in this call. You can then call [retrieveLostData] when your app relaunches to retrieve the lost data.
   ///
   /// If no images were picked, the return value is null.
+  @Deprecated('Use getImageFromSource instead.')
   Future<PickedFile?> pickImage({
     required ImageSource source,
     double? maxWidth,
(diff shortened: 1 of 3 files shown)


tarrinneal (approved):

Hixie:
test-exempt: roll dependency

packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
```diff
@@ -175,6 +175,7 @@ abstract class ImagePickerPlatform extends PlatformInterface {
   /// in this call. You can then call [getLostData] when your app relaunches to retrieve the lost data.
   ///
   /// If no images were picked, the return value is null.
+  @Deprecated('Use getImageFromSource instead.')
```
ditman, line 178: `getImageFromSource` uses `getImage`, should we flip the default implementation so the deprecated method calls the recommended one internally?

```
  Future<XFile?> getImageFromSource({
    required ImageSource source,
    ImagePickerOptions options = const ImagePickerOptions(),
  }) {
    return getImage(
      source: source,
      maxHeight: options.maxHeight,
      maxWidth: options.maxWidth,
      imageQuality: options.imageQuality,
      preferredCameraDevice: options.preferredCameraDevice,
    );
  }
```

stuartmorgan: > `getImageFromSource` uses `getImage` in this file:

That's the default, not very good implementation, which platform implementations should override with a real one. See the paragraph starting "Note:" in https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-platform-interface-method-parameters

> Should we flip the default implementation so the deprecated method calls the recommended one internally?

That would suddenly make all old implementations throw when used with newer versions of the app-facing package.

ditman: > The delegation of this method will feel backwards, because the implementing the new method in terms of the old method will discard the information from the new parameters. This is correct though, because the goal is to allow the app-facing package to call the new method regardless of whether implementation packages have implemented it.

Yep, exactly.

I'll flip the web implementation so the deprecated methods are based off of the new ones, so we can eventually remove the deprecated methods later!

PR #3258, opened 2023-02-22
[google_maps_flutter_web] Options to disable tilt controls and configure gesture handling

Rexios80:
Transferred from https://github.com/flutter/plugins/pull/4916

tiltControlsEnabled: On web, in satellite view, and at close zoom levels, the map displays tilt controls. This lets you disable them.

gestureHandling: On web, without greedy gesture handling the map can have an overlay saying you can only move the map with two fingers or Cmd+drag. This is not ideal, and greedy gestures lets you make that not happen. I added the other gesture handling options from the [Google Maps documentation](https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions.gestureHandling) for completeness.

*List which issues are fixed by this PR. You must list at least one issue.*

[flutter/flutter/#99044](https://github.com/flutter/flutter/issues/99044)

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter]. (Unlike the flutter/flutter repo, the flutter/plugins repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the plugin surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/plugins/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[...]

PR #4521, opened 2023-07-19
[google_maps_flutter_web] Adds options for gesture handling and tilt controls.

Rexios80:
See #3258


packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -62,9 +62,20 @@ gmaps.MapOptions _configurationAndStyleToGmapsOptions(
 
   if (configuration.scrollGesturesEnabled == false ||
       configuration.zoomGesturesEnabled == false) {
-    options.gestureHandling = 'none';
+    window.console.warn(
+      '`scrollGesturesEnabled` and `zoomGesturesEnabled` are now ignored on the web.'
+      ' Use `webGestureHandling: WebGestureHandling.none` instead.',
```
stuartmorgan, line 67: Won't this warn people who are setting these for mobile? We shouldn't have warnings that are impossible to remove in normal usage.

stuartmorgan: Couldn't we just keep the old behavior if `webGestureHandling` is null?

Rexios80: Wouldn't it be possible to want `scrollGesturesEnabled` or `zoomGesturesEnabled` false for mobile but not want `gestureHandling = 'none'`

Rexios80: Right you said only if it's null

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -62,9 +62,20 @@ gmaps.MapOptions _configurationAndStyleToGmapsOptions(
...
+    );
+  }
+
+  if (configuration.webGestureHandling != null) {
+    options.gestureHandling = configuration.webGestureHandling!.name;
   } else {
```
ditman, line 73:
```suggestion
  } else if (configuration.scrollGesturesEnabled == false ||
      configuration.zoomGesturesEnabled == false) {
      // Old behavior
      options.gestureHandling = 'none';
  } else {
    options.gestureHandling = WebGestureHandling.auto.name;
  }
```

/cc @stuartmorgan is this what you mean? (and remove the warn?)

stuartmorgan: Yes, exactly (unless there's a reason that won't work, it seems like it's the most backwards-compatible behavior).

ditman (approved):
LGTM!

stuartmorgan (approved):
LGTM

PR #4530, opened 2023-07-20
[ci] Simplify Dockerfile

stuartmorgan:
Most of the things installed by the Dockerfile are for tests that have moved to LUCI, so are no longer used. This simlifies down to just the FTL requirements while we work on that final conversion.

### DIFF
--- a/.ci/Dockerfile
+++ b/.ci/Dockerfile
@@ -1,7 +1,6 @@
 # The Flutter version is not important here, since the CI scripts update Flutter
 # before running. What matters is that the base image is pinned to minimize
 # unintended changes when modifying this file.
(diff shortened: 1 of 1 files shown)


.ci/Dockerfile
```diff
@@ -1,7 +1,6 @@
 # The Flutter version is not important here, since the CI scripts update Flutter
 # before running. What matters is that the base image is pinned to minimize
 # unintended changes when modifying this file.
-# This is the hash for the 3.0.0 image.
```
stuartmorgan, line 4: This was very stale; dependabot manages the next line now, not humans.

tarrinneal (approved):

PR #3259, opened 2023-02-22
[video_player] Add optional web options

defuncart:
* Fixes https://github.com/flutter/flutter/issues/62192
* Fixes first part of https://github.com/flutter/flutter/issues/88151 (ability to show controls)
* Fixes https://github.com/flutter/flutter/issues/64397

Can be used as follows:

```dart
_controller = VideoPlayerController.asset(
  'assets/Butterfly-209.mp4',
  videoPlayerOptions: const VideoPlayerOptions(
    webOptions: VideoPlayerWebOptions(
        controls: VideoPlayerWebOptionsControls.enabled(
          allowDownload: false,
          allowFullscreen: false,
          allowPlaybackRate: false,
          allowPictureInPicture: false,
        ),
        allowContextMenu: false,
        allowRemotePlayback: false,
     ),
  ),
);
```

Bildschirmfoto 2023-07-06 um 17 04 59

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [X] I signed the [CLA].
- [X] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [X] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[...]

PR #4551, opened 2023-07-23
[video_player] Add optional web options [web]

defuncart:
Web PR for Video Player Web Options (https://github.com/flutter/packages/pull/3259).

## Pre-launch Checklist

- [X] I signed the [CLA].


defuncart:
@ditman @stuartmorgan Web PR for #3259.

packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -202,9 +203,35 @@ class VideoPlayer {
     return Duration(milliseconds: (_videoElement.currentTime * 1000).round());
   }
 
+  /// Sets options
+  Future<void> setOptions(VideoPlayerWebOptions options) async {
```
ditman, line 207: Can this method be called multiple times on the same _videoElement? Does this method need to support both cases, setting and unsetting options?

Imagine that this gets called with `allowContextMenu=false` once, and then `allowContextMenu=true`

ditman: (Since there's no mechanism in the code that prevents a core plugin from calling `setWebOptions` multiple times, I think this method should work if called multiple times in succession, and be able to set/remove values).

defuncart: As per #3259, the use case is `video_player/lib/video_player.dart` calls `setWebOptions` once in `initialize`, however there is nothing in place to stop this method being called multiple times. I've added a check to de-initialize attributes when required.

packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -202,9 +203,57 @@ class VideoPlayer {
...
+
+    if (!options.allowRemotePlayback) {
+      _videoElement.setAttribute('disableRemotePlayback', true);
+    }
+  }
+
+  void _resetOptions() {
+    _videoElement.controls = false;
+    if (_videoElement.hasAttribute('controlsList')) {
+      _videoElement.setAttribute('controlsList', '');
```
ditman, line 236: Instead of setting these to a weird value, I think it is more correct to `removeAttribute`. I also think you don't need to check if the attribute exists to remove it, so this method could end up much simpler, like this:

```dart
_videoElement.controls = false;
_videoElement.removeAttribute('controlsList');
_videoElement.removeAttribute('disablePictureInPicture');
_videoElement.removeAttribute('disableRemotePlayback');
if (_onContextMenu != null) {
  // Unsubscribe!
}
```

defuncart: Tests seem happy, nice simplification.

packages/video_player/video_player_web/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## NEXT
```
ditman, line 1:
```suggestion
## 2.1.0
```

This needs to be published to be usable, right? If so, we want to bump the version to `2.1.0`, here and in the `pubspec.yaml` of the package!

ditman: (Done)

ditman (approved):
This LGTM, I think this is ready to land. Second opinion, @stuartmorgan?

packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -202,9 +203,51 @@ class VideoPlayer {
     return Duration(milliseconds: (_videoElement.currentTime * 1000).round());
   }
 
+  /// Sets options
+  Future<void> setOptions(VideoPlayerWebOptions options) async {
+    // incase called multiple times, reset options
```
stuartmorgan, line 208:
```suggestion
    // In case this method is called multiple times, reset options.
```

stuartmorgan (approved):
LGTM with two nits.

defuncart:
@stuartmorgan @ditman Requested changes pushed :)

packages/video_player/video_player_web/CHANGELOG.md
stuartmorgan: I can't see the history since this PR has been force-pushed, but it looks like you undid an existing change.

stuartmorgan (changes requested):

packages/video_player/video_player_web/CHANGELOG.md
defuncart: @stuartmorgan Apologies, added back https://github.com/flutter/packages/pull/4551/commits/688077fb194b5803e9180afef1b26c6b71245384

stuartmorgan: The actual package version bump is still reverted, as indicated by CI.

stuartmorgan (changes requested):
Updating review state per my comment above; this is failing CI because it doesn't actually contain a version bump.

defuncart:
@ditman @stuartmorgan Please re-reviewed and let me know if anything needs to be changed.

stuartmorgan (approved):
I've updated the PR to bump the version, since that was still missing.

LGTM with that change.

stuartmorgan:
Sorry, looks like I just missed adding the label to this.

PR #4562, opened 2023-07-25
[tool/ci] Add iOS/macOS and Dart support to `fetch-deps`

stuartmorgan:
Adds `fetch-deps` support for:
- iOS/macOS dependencies, using `pod install`
- Dart package dependencies, using `pub get`

To make avoid doing extra work in the Dart dependencies step when using this with `*_platform_tests` CI, also adds flags for all of the other platforms, and adds a flag that allows skipping Dart dependencies for any package that doesn't have an example supporting any requested platform. This means that we can pass, e.g., `--windows --supporting-target-platforms-only` to only fetch Dart packages for packages with examples that will be build during the build-and-drive Windows tests.

Adds this as a new step in every platform tests CI task, and in the standard analyze step, so that we will pre-fetch Dart packages (and for iOS/macOS, pods). This won't yet fully eliminate later network access (see https://github.com/flutter/flutter/issues/131204), but will give us early warning on any major failures, such as pub being entirely unreachable from the bots.
- These are marked as an infrastructure step; we'll have to see if this ends up being confusing in practice. If `pub` resolution fails for legitimate reasons, such as a PR that tries to require a version of a package that doesn't exist or that has conflicts, this will cause a failure that is marked as infra. My assumption is that the much more common case is going to be that it is actually an infra failure.

Fixes https://github.com/flutter/flutter/issues/130280

### DIFF
--- a/.ci/targets/android_platform_tests.yaml
+++ b/.ci/targets/android_platform_tests.yaml
@@ -2,10 +2,10 @@ tasks:
   - name: prepare tool
     script: .ci/scripts/prepare_tool.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
-  - name: download android deps
+  - name: download Dart and android deps
     script: script/tool_runner.sh
     infra_step: true
-    args: ["fetch-deps"]
+    args: ["fetch-deps", "--android", "--supporting-target-platforms-only"]
   - name: build examples
     script: script/tool_runner.sh
     args: ["build-examples", "--apk"]
--- a/.ci/targets/windows_build_and_platform_tests.yaml
+++ b/.ci/targets/windows_build_and_platform_tests.yaml
@@ -2,6 +2,10 @@ tasks:
   - name: prepare tool
     script: .ci/scripts/prepare_tool.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
+  - name: download Dart deps
+    script: script/tool_runner.sh
+    args: ["fetch-deps", "--windows", "--supporting-target-platforms-only"]
+    infra_step: true
   - name: build examples (Win32)
     script: .ci/scripts/build_examples_win32.sh
   - name: native unit tests (Win32)
--- a/script/tool/lib/src/fetch_deps_command.dart
+++ b/script/tool/lib/src/fetch_deps_command.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:file/file.dart';
 
 import 'common/core.dart';
 import 'common/gradle.dart';
@@ -10,12 +11,14 @@ import 'common/package_looping_command.dart';
 import 'common/plugin_utils.dart';
 import 'common/repository_package.dart';
 
-/// Download dependencies for the following platforms {android}.
+const int _exitNothingRequested = 3;
+
+/// Download dependencies, both Dart and native.
 ///
 /// Specficially each platform runs:
 ///   Android: 'gradlew dependencies'.
-///   Dart: TBD (flutter/flutter/issues/130279)
-///   iOS: TBD (flutter/flutter/issues/130280)
+///   Dart: 'flutter pub get'.
+///   iOS/macOS: 'pod install'.
 ///
 /// See https://docs.gradle.org/6.4/userguide/core_dependency_management.html#sec:dependency-mgmt-in-gradle.
 class FetchDepsCommand extends PackageLoopingCommand {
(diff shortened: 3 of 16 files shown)


script/tool/lib/src/fetch_deps_command.dart
```diff
@@ -10,12 +11,14 @@ import 'common/package_looping_command.dart';
...
+/// Download dependencies, both Dart and native.
 ///
 /// Specficially each platform runs:
 ///   Android: 'gradlew dependencies'.
-///   Dart: TBD (flutter/flutter/issues/130279)
-///   iOS: TBD (flutter/flutter/issues/130280)
+///   Dart: 'flutter pub get'.
+///   iOS/macOS: 'pod install'.
 ///
 /// See https://docs.gradle.org/6.4/userguide/core_dependency_management.html#sec:dependency-mgmt-in-gradle.
```
reidbaker, line 23: nit: Consider linking to ios and dart documentation or removing android documentation.

.ci/targets/windows_build_and_platform_tests.yaml
```diff
@@ -2,6 +2,10 @@ tasks:
   - name: prepare tool
     script: .ci/scripts/prepare_tool.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
+  - name: download Dart deps
+    script: script/tool_runner.sh
```
reidbaker, line 6: This name is kind of awkward but I genuinely can't think of something better. I thought filter might be better but this is the opposite of a filter is is an allow list.

script/tool/lib/src/fetch_deps_command.dart
```diff
@@ -63,12 +169,81 @@ class FetchDepsCommand extends PackageLoopingCommand {
...
     }
 
     return PackageResult.success();
   }
+
+  Future<PackageResult> _fetchDarwinDeps(
+      RepositoryPackage package, final String platform) async {
+    if (!pluginSupportsPlatform(platform, package,
+        requiredMode: PlatformSupport.inline)) {
+      final String displayPlatform = platform.replaceFirst('os', 'OS');
```
reidbaker, line 186: consider adding a comment for why this replacement is required.

script/tool/lib/src/fetch_deps_command.dart
```diff
@@ -63,12 +169,81 @@ class FetchDepsCommand extends PackageLoopingCommand {
...
+
+      final File generatedXCConfig = platform == platformMacOS
+          ? platformDir
+              .childDirectory('Flutter')
+              .childDirectory('ephemeral')
+              .childFile('Flutter-Generated.xcconfig')
+          : platformDir
+              .childDirectory('Flutter')
+              .childFile('Generated.xcconfig');
+      // Running `pod install` requires `flutter pub get` or `flutter build` to
```
reidbaker, line 203: FWIW I think you can --config-only for ios builds. That said I think pub get is probably the more stable command to run here.

reidbaker (approved):

script/tool/lib/src/fetch_deps_command.dart
```diff
@@ -24,25 +27,128 @@ class FetchDepsCommand extends PackageLoopingCommand {
     super.packagesDir, {
     super.processRunner,
     super.platform,
-  });
+  }) {
+    argParser.addFlag(_dartFlag, defaultsTo: true, help: 'Run "pub get"');
+    argParser.addFlag(_supportingTargetPlatformsOnlyFlag,
+        help: 'Restricted "pub get" runs to packages that have at least one '
```
tarrinneal, line 33: Restricted or Restricts?

.ci/targets/android_platform_tests.yaml
```diff
@@ -2,10 +2,10 @@ tasks:
   - name: prepare tool
     script: .ci/scripts/prepare_tool.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
-  - name: download android deps
+  - name: download Dart and android deps
```
tarrinneal, line 5: Seems like the `android` in these names should be capitalized.

tarrinneal (approved):
Nothing outstandingly wrong with this one.

stuartmorgan:
Looks like I'll need to plumb `environment` through our process wrapper to pass `LANG` to resolve the CI failures; I checked `flutter`, and it passes `LANG` for `pod install` as suggested by the error message.

.ci/targets/android_platform_tests.yaml
stuartmorgan: Fixed.

.ci/targets/windows_build_and_platform_tests.yaml
stuartmorgan: The tool unfortunately collects these useful-but-hard-to-describe-succinctly flags. This was my third or fourth attempt, and it was the best I came up with as well.

script/tool/lib/src/fetch_deps_command.dart
stuartmorgan: Since the others are just running commands that are run as part of building I couldn't think of any good docs to link to. I left Android because I think it's useful given that this isn't just us pre-running a command that's being directly run by `flutter`.

script/tool/lib/src/fetch_deps_command.dart
stuartmorgan: Fixed.

script/tool/lib/src/fetch_deps_command.dart
stuartmorgan: Done.

script/tool/lib/src/fetch_deps_command.dart
stuartmorgan: Added.

stuartmorgan:
> Nothing outstandingly wrong with this one.

Maybe we should replace LGTM with NOWWTO :)

tarrinneal:
> > Nothing outstandingly wrong with this one.
> 
> Maybe we should replace LGTM with NOWWTO :)

It's got a nice ring to it. Only two sillables too.

stuartmorgan:
Plus then you can say "NOWWTO land it!"

PR #4571, opened 2023-07-26
[ci] Add LUCI version of Android FTL tests in bringup mode

stuartmorgan:
Adds new LUCI versions of the Android FTL tests. These are new tasks, rather than being part of android_platform_tests, since those are already running the emulator version of the tests, and the combination is far too slow (unless we roughly double the number of shards anyway), and this lets us control the behavior separately (e.g., not running them on `stable`, and potentially making them post-submit only in the future).

Adjusts the repo tooling to make the `gcloud auth` step optional, since that's handled automatically for us in the LUCI environment, and to make it less tightly coupled to the Cirrus configuration by removing those defaults from the tool.

See also https://github.com/flutter/flutter/issues/131429


.ci.yaml
```diff
@@ -464,6 +464,70 @@ targets:
           {"dependency": "android_virtual_device", "version": "33"}
         ]
 
+  # Device versions of Android integration tests, run via FTL.
+  # TODO(stuartmorgan): Revisit whether we want these once we have more data
+  # about emulator tests; see https://github.com/flutter/flutter/issues/131429.
```
reidbaker, line 469: Nit: I think we should be more specific for our future selves. "Revisit if physical device tests are redundant once we have more data about emulator tests;

reidbaker (approved):

.ci.yaml
stuartmorgan: Done.

PR #4580, opened 2023-07-28
[pigeon] primitive enums

tarrinneal:
Adds support for enums as parameters and return types.

Dart, C++, Java, Kotlin, Objective-C, and swift:
Primitive synchronous non-null
Primitive synchronous nullable
Primitive asynchronous non-null
Primitive asynchronous nullable
Primitive Flutter api non-null
Primitive flutter api nullable

Objective-C:
Class property nullable 

Also fixes an issue with nullable enums in classes on objc. This fix required a breaking change that nested all nullable enums in a wrapper class to allow nullability. 

Also adds the ability to format files before tests run (to make my life better)

Also replaces https://github.com/flutter/packages/pull/4756

Also adds objc prefixes to enums

fixes: https://github.com/flutter/flutter/issues/87307
fixes: https://github.com/flutter/flutter/issues/118733


packages/pigeon/lib/dart_generator.dart
```diff
@@ -442,10 +442,15 @@ $resultAt != null
                   } else {
                     indent.writeln('final $returnType output = $call;');
                   }
+
                   const String returnExpression = 'output';
+                  final String nullability =
+                      func.returnType.isNullable ? '?' : '';
+                  final String enumOutput =
+                      isEnum(root, func.returnType) ? '$nullability.index' : '';
```
stuartmorgan, line 450: It's weird to call this `enumOutput` when it's usually not a enum. Maybe `valueExtraction`?

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -425,8 +431,15 @@ class KotlinGenerator extends StructuredGenerator<KotlinOptions> {
             });
           } else {
             indent.addScoped('{', '}', () {
-              indent.writeln(
-                  'val result = ${_cast('it', type: func.returnType)}');
+              // nullable enums require special handling.
```
stuartmorgan, line 434: Nit: Nullable

packages/pigeon/lib/java_generator.dart
```diff
@@ -452,6 +452,17 @@ class JavaGenerator extends StructuredGenerator<JavaOptions> {
         }
       });
 
+      /// Returns an argument name that can be used in a context where it is possible to collide
+      /// and append `.index` to enums.
+      String getEnumSafeArgumentName(int count, NamedType argument) {
```
stuartmorgan, line 457: Nit: s/Name/Expression/, since it's not always a name.

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -736,6 +761,17 @@ String _getCodecName(Api api) => '${api.name}Codec';
 String _getArgumentName(int count, NamedType argument) =>
     argument.name.isEmpty ? 'arg$count' : argument.name;
 
+/// Returns an argument name that can be used in a context where it is possible to collide
+/// and append `.index` to enums.
+String _getEnumSafeArgumentName(Root root, int count, NamedType argument) {
```
stuartmorgan, line 766: Same Name/Expression change here too.

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -828,6 +864,14 @@ String _cast(String variable, {required TypeDeclaration type}) {
   if (typeString == 'Int' || typeString == 'Long') {
     return _castInt(type.isNullable);
   }
+  if (isEnum(root, type)) {
+    if (type.isNullable) {
+      return '(list[13] as Int?)?.let {\n'
```
stuartmorgan, line 869: I suspect this `list[13]` isn't going to work for all input files ;)

packages/pigeon/lib/swift_generator.dart
```diff
@@ -275,6 +275,17 @@ import FlutterMacOS
     required String dartPackageName,
   }) {
     assert(api.location == ApiLocation.flutter);
+
+    /// Returns an argument name that can be used in a context where it is possible to collide.
+    String getEnumSafeArgumentName(Root root, int count, NamedType argument) {
```
stuartmorgan, line 280: Same name change here.

stuartmorgan (changes requested):
Nice! Various nits and suggestions, but it's really great to see comprehensive enum support.

packages/pigeon/lib/objc_generator.dart
```diff
@@ -156,6 +156,21 @@ class ObjcHeaderGenerator extends StructuredGenerator<ObjcOptions> {
...
+
+  void _writeEnumWrapper(
+    Indent indent,
+    String enumName,
+  ) {
+    indent.newln();
+    indent.writeln('@class ${enumName}Wrapper;');
+    indent.newln();
+    indent.writeln(
+        '/// Wrapper for $enumName to manage primitive and nullable use.');
```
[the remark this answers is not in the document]
tarrinneal: That's outdated, I was originally wrapping all primitives.

packages/pigeon/lib/objc_generator.dart
```diff
@@ -156,6 +156,21 @@ class ObjcHeaderGenerator extends StructuredGenerator<ObjcOptions> {
...
+    indent.writeln('@interface ${enumName}Wrapper : NSObject');
+    indent.writeln('@property(nonatomic, assign) $enumName value;');
```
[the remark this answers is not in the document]
tarrinneal: 🤦

packages/pigeon/lib/objc_generator.dart
```diff
@@ -420,6 +458,26 @@ class ObjcSourceGenerator extends StructuredGenerator<ObjcOptions> {
...
+  }) {
+    final String enumName = _className(generatorOptions.prefix, anEnum.name);
+    addDocumentationComments(
+        indent, anEnum.documentationComments, _docCommentSpec);
+
+    indent.writeln('@interface ${enumName}Wrapper ()');
+    indent.writeln('@end');
+    indent.newln();
+    indent.writeln('@implementation ${enumName}Wrapper');
+    indent.writeln('@end');
```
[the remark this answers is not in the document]
tarrinneal: We don't follow the `if (self)` pattern anywhere else in this package, is it necessary here?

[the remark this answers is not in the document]
stuartmorgan: It is always best practice to do some form of the check (another variant is `if ((self = [super init])) {`), because `init` is not guaranteed to return a non-nil value.

Are we ever going to end up in a situation where that happens in a class that inherits from `NSObject`? Almost certainly not, but I'm not aware of any argument for **not** doing it, and consistently applying the pattern means not forgetting it in cases where it actually could be necessary in practice.

packages/pigeon/lib/objc_generator.dart
```diff
@@ -670,16 +763,34 @@ class ObjcSourceGenerator extends StructuredGenerator<ObjcOptions> {
         }
       } else {
         const String callback = 'callback(wrapResult(output, error));';
+        String returnTypeString = '${returnType.withPtr}_Nullable output';
+        final String enumConvert = func.returnType.isNullable
+            ? 'NSNumber *output = enumValue == nil ? nil : [NSNumber numberWithInteger:enumValue.value];'
+            : 'NSNumber *output = [NSNumber numberWithInteger:enumValue];';
```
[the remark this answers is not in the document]
tarrinneal: Just for my learning, would this be worse than that?

```
        const String numberOutput = 'NSNumber *output =';
        final String enumConvert = func.returnType.isNullable
            ? '$numberOutput enumValue == nil ? nil : [NSNumber numberWithInteger:enumValue.value];'
            : '$numberOutput [NSNumber numberWithInteger:enumValue];';```

File: packages/pigeon/lib/cpp_generator.dart [hunk 11]
```diff
@@ -888,11 +888,26 @@ class CppSourceGenerator extends StructuredGenerator<CppOptions> {
...
+            );
+          }
+          if (returnType.isEnum && returnType.isNullable) {
+            indent.writeScoped('if (!encodable_return_value.IsNull()) {', '}',
+                () {
+              indent.writeln('on_success(&$successCallbackArgument);');
+            }, addTrailingNewline: false);
+            indent.addScoped(' else {', '}', () {
+              indent.writeln('on_success(nullptr);');
+            });
```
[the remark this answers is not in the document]
tarrinneal: I tried something like that and had issues, but that might have been pointer related (I was trying to return return_value directly)

packages/pigeon/lib/dart_generator.dart
tarrinneal: I agree, but it is an exclusively enum string, and the name makes it clear that it is. enumExtraction might be better though. I'm open to whatever makes sense though.

packages/pigeon/lib/kotlin_generator.dart
tarrinneal: that's just a restriction of the new feature. I'll add documentation...

packages/pigeon/lib/dart_generator.dart
stuartmorgan: I missed that it's empty in the non-enum case; I was thinking it was wrapping and replacing, rather than additive. Your naming is fine given that, sorry for the noise.

packages/pigeon/lib/objc_generator.dart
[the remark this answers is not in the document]
stuartmorgan: Subjectively, I would say yes. In the version I wrote, it's extremely clear that there's a single expression declaring `output`, and just the RHS varies. In your new version, you still have to read the strings and refer back to an earlier variable to see that both are declaring the same variable.

packages/pigeon/lib/cpp_generator.dart
```diff
@@ -1355,20 +1345,16 @@ ${prefix}reply(EncodableValue(std::move(wrapped)));''';
...
+          indent.writeln(
+              'const int64_t $valueVarName = $encodableArgName.IsNull() ? 0 : $encodableArgName.LongValue();');
           if (apiType == ApiType.flutter) {
-            indent.writeln('${hostType.datatype} $argName;');
+            indent.writeln(
+                'const auto* $argName = $encodableArgName.IsNull() ? nullptr : &(${hostType.datatype})$valueVarName;');
           } else {
-            indent.writeln('std::optional<${hostType.datatype}> $argName;');
+            indent.writeln(
+                'const std::optional<${hostType.datatype}> $argName = $encodableArgName.IsNull() ? std::optional<${hostType.datatype}>() : std::make_optional<${hostType.datatype}>(static_cast<${hostType.datatype}>(${argName}_value));');
```
stuartmorgan, line 1356: Can `std::optional<${hostType.datatype}>()` just be `std::nullopt`? I think that works in ternaries.

stuartmorgan (approved):
LGTM!

(The more I see `FooEnumBox` the less I like that suggestion, but I still don't have a better idea. I'll just take comfort in the fact that nullable enums will probably be rare, and ObjC generation is likely decreasing, so the intersection should be rare and shrinking.)

packages/pigeon/lib/cpp_generator.dart
tarrinneal: It didn't work when I tried it in this context, but it's possible I had a different issue.

tarrinneal:
> (The more I see `FooEnumBox` the less I like that suggestion, but I still don't have a better idea. I'll just take comfort in the fact that nullable enums will probably be rare, and ObjC generation is likely decreasing, so the intersection should be rare and shrinking.)

better or worse than wrapper?

PR #4583, opened 2023-07-28
[ci] Switch FTL to LUCI

stuartmorgan:
Enables the new LUCI targets, and removes the Cirrus version.

Now that Cirrus is only used for a single task that only ever runs against Flutter `master`, also simplifies the remaining Cirrus setup logic to only handle `master`.

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -470,7 +470,6 @@ targets:
   # https://github.com/flutter/flutter/issues/131429.
 
   - name: Linux_android android_device_tests_shard_1 master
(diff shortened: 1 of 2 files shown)


stuartmorgan:
(The new LUCI configs probably haven't propagated yet, so scheduling may stall, but I wanted to go ahead and post it for review so it's ready to land once the CI infrastructure is ready.)

reidbaker (approved):

tarrinneal (approved):
NOWWTO!

PR #4586, opened 2023-07-29
[camera] add heif support image iOS

Mairramer:
Add support to heic format on IOS on take picture.

Based on https://github.com/flutter/flutter/issues/119795

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Thanks for the contribution! Please see https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins for the process for a PR like this; the missing steps described there are why many of the CI steps are failing.

This will also need to follow https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-platform-interface-method-parameters, since this isn't something that could be landed as written.

I'm going to mark it as a draft for now, please mark it as ready for review once the PR has been updated accordingly. Thanks!

Mairramer:
@stuartmorgan Do you think we can configure avfoundation_camera to always return HEIF/HEVC format? I implemented the above but I don't think it was good, as this only applies to iOS.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -254,16 +259,27 @@ - (void)updateOrientation:(UIDeviceOrientation)orientation
...
   if (_resolutionPreset == FLTResolutionPresetMax) {
     [settings setHighResolutionPhotoEnabled:YES];
   }
 
+  if (_outputFormat == OutputFormatHEIF) {
+    if (@available(iOS 11.0, *)) {
+      if ([self.capturePhotoOutput.availablePhotoCodecTypes containsObject:AVVideoCodecTypeHEVC]) {
+        settings = [AVCapturePhotoSettings
+            photoSettingsWithFormat:@{AVVideoCodecKey : AVVideoCodecTypeHEVC}];
+        extension = @"heif";
```
hellohuanlin, line 272: rather than setting it to `jpg` and then overwrite it to `heif`, can you write `if/else`?

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -254,16 +259,27 @@ - (void)updateOrientation:(UIDeviceOrientation)orientation
 
 - (void)captureToFile:(FLTThreadSafeFlutterResult *)result {
   AVCapturePhotoSettings *settings = [AVCapturePhotoSettings photoSettings];
+  NSString *extension = @"jpg";
   if (_resolutionPreset == FLTResolutionPresetMax) {
     [settings setHighResolutionPhotoEnabled:YES];
   }
 
+  if (_outputFormat == OutputFormatHEIF) {
+    if (@available(iOS 11.0, *)) {
```
hellohuanlin, line 268: I could be wrong but did we already stopped supporting versions before iOS 11

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -254,16 +259,27 @@ - (void)updateOrientation:(UIDeviceOrientation)orientation
...
+      if ([self.capturePhotoOutput.availablePhotoCodecTypes containsObject:AVVideoCodecTypeHEVC]) {
```
hellohuanlin, line 269: do you know when will this return false? what should we do if it's false? if we default to jpg, we probably wanna document that behavior too.

Mairramer: https://github.com/flutter/packages/pull/4586#discussion_r1364741704

stuartmorgan:
> @stuartmorgan Do you think we can configure avfoundation_camera to always return HEIF/HEVC format?

That would be a question for @bparrishMines and @hellohuanlin to decide. If the PR ends up being a breaking change though, that will make it much more difficult to adopt.

packages/camera/camera_avfoundation/ios/Classes/CameraProperties.h
```diff
@@ -127,4 +127,19 @@ extern FLTResolutionPreset FLTGetFLTResolutionPresetForString(NSString *preset);
  */
 extern OSType FLTGetVideoFormatFromString(NSString *videoFormatString);
 
+/**
+ * Represents image format. Mirrors OutputFormat in camera.dart.
+ */
+typedef NS_ENUM(NSInteger, OutputFormat) {
```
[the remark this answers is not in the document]
stuartmorgan: > `FLT-` prefix

New code should use `FCP`; see https://github.com/flutter/flutter/issues/102601

packages/camera/camera_avfoundation/ios/Classes/CameraProperties.m
```diff
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 #import "CameraProperties.h"
+#import <FLTCam.h>
```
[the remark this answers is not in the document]
Mairramer: A mistake

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
hellohuanlin: what prevents us from writing

```
if supported {
  extension = "heif"
else {
  extension = "jpg"
}
```

hellohuanlin:
> Do you think we can configure avfoundation_camera to always return HEIF/HEVC format?

Not sure if i get this question - are you saying you want to make heif the default? What if there's no heif support?

Mairramer:
> > Do you think we can configure avfoundation_camera to always return HEIF/HEVC format?
> 
> Not sure if i get this question - are you saying you want to make heif the default? What if there's no heif support?

It was my first idea. But as you said, there may not be heif support and I also know some developers who prefer and use jpeg as the default in their services.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
Mairramer: I rewrote it to use if/else

cyanglaz:
@Mairramer I see that you have updated the code since the last review. Is this ready for another review?

Mairramer:
> I see that you have updated the code since the last review. Is this ready for another review?

Yes, I made some small changes that were requested.

packages/camera/camera_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## NEXT
+
+* Add support to HEIF format
```
hellohuanlin, line 3: nit: Adds

packages/camera/camera_avfoundation/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
```diff
@@ -94,4 +94,86 @@ - (void)testCaptureToFile_mustReportPathToResultIfSavePhotoDelegateCompletionsWi
   [self waitForExpectationsWithTimeout:1 handler:nil];
 }
 
+- (void)testCaptureToFile_mustSetExtensionToHeicIfAvailablePhotoCodecTypesContainsHEVC {
+  XCTestExpectation *expectation =
+      [self expectationWithDescription:
+                @"Test must set extension to heif if availablePhotoCodecTypes contains HEVC."];
+  dispatch_queue_t captureSessionQueue = dispatch_queue_create("capture_session_queue", NULL);
+  dispatch_queue_set_specific(captureSessionQueue, FLTCaptureSessionQueueSpecific,
```
hellohuanlin, line 102: is this used anywhere?

packages/camera/camera_avfoundation/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
```diff
@@ -94,4 +94,86 @@ - (void)testCaptureToFile_mustReportPathToResultIfSavePhotoDelegateCompletionsWi
...
+  AVCapturePhotoSettings *settings =
+      [AVCapturePhotoSettings photoSettingsWithFormat:@{AVVideoCodecKey : AVVideoCodecTypeHEVC}];
+
+  id mockSettings = OCMClassMock([AVCapturePhotoSettings class]);
+  OCMStub([mockSettings photoSettingsWithFormat:OCMOCK_ANY]).andReturn(settings);
+
+  NSString *filePath = @"test";
+  id mockResult = OCMClassMock([FLTThreadSafeFlutterResult class]);
+  OCMStub([mockResult sendSuccessWithData:filePath]).andDo(^(NSInvocation *invocation) {
+    [expectation fulfill];
```
hellohuanlin, line 116: the expectation is `"Test must set extension to heif if availablePhotoCodecTypes contains HEVC."`, but here it fulfilled without verifying the extension.

packages/camera/camera_avfoundation/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
```diff
@@ -94,4 +94,86 @@ - (void)testCaptureToFile_mustReportPathToResultIfSavePhotoDelegateCompletionsWi
...
+  id mockOutput = OCMClassMock([AVCapturePhotoOutput class]);
+  // Set availablePhotoCodecTypes to HEVC
+  NSArray *codecTypes = @[ AVVideoCodecTypeHEVC ];
+  OCMStub([mockOutput availablePhotoCodecTypes]).andReturn(codecTypes);
+
+  OCMStub([mockOutput capturePhotoWithSettings:OCMOCK_ANY delegate:OCMOCK_ANY])
+      .andDo(^(NSInvocation *invocation) {
+        FLTSavePhotoDelegate *delegate = cam.inProgressSavePhotoDelegates[@(settings.uniqueID)];
+        // Completion runs on IO queue.
+        dispatch_queue_t ioQueue = dispatch_queue_create("io_queue", NULL);
```
hellohuanlin, line 128: Labeling it "io_queue" doesn't make it the same as the actual IO queue of FLTCam. Can you either get the queue from FLTCam, or inject this IO queue into FLTCam?

hellohuanlin (changes requested):

packages/camera/camera_avfoundation/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
hellohuanlin: Sorry never mind - the actual FLTCam's implementation that relies on IO queue is mocked out. So any queue here is fine.

packages/camera/camera_avfoundation/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
```diff
@@ -94,4 +94,86 @@ - (void)testCaptureToFile_mustReportPathToResultIfSavePhotoDelegateCompletionsWi
   [self waitForExpectationsWithTimeout:1 handler:nil];
 }
 
+- (void)testCaptureToFile_mustSetExtensionToHeicIfAvailablePhotoCodecTypesContainsHEVC {
```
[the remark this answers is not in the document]
Mairramer: If the user does not define a value, it will default to jpg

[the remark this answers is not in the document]
hellohuanlin: Then the function name is inaccurate?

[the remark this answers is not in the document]
Mairramer: in a way yes

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -254,16 +259,28 @@ - (void)updateOrientation:(UIDeviceOrientation)orientation
...
 
+  BOOL isHEVCCodecAvailable =
+      [self.capturePhotoOutput.availablePhotoCodecTypes containsObject:AVVideoCodecTypeHEVC];
+
+  if (_fileFormat == FCPFileFormatHEIF && isHEVCCodecAvailable) {
+    settings =
+        [AVCapturePhotoSettings photoSettingsWithFormat:@{AVVideoCodecKey : AVVideoCodecTypeHEVC}];
+    _fileExtension = @"heif";
+  } else {
+    _fileExtension = @"jpg";
```
hellohuanlin, line 275: The `fileExtension` should be just local variables.

packages/camera/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## NEXT
```
hellohuanlin, line 1: do we need version bump to the interface package? @stuartmorgan

packages/camera/camera_avfoundation/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
```diff
@@ -110,10 +110,10 @@ - (void)testCaptureToFile_mustReportFileExtensionWithHeifWhenHEVCIsAvailableAndF
   id mockSettings = OCMClassMock([AVCapturePhotoSettings class]);
   OCMStub([mockSettings photoSettingsWithFormat:OCMOCK_ANY]).andReturn(settings);
 
-  NSString *filePath = @"test";
+  NSString *filePath = @"test.heif";
   id mockResult = OCMClassMock([FLTThreadSafeFlutterResult class]);
   OCMStub([mockResult sendSuccessWithData:filePath]).andDo(^(NSInvocation *invocation) {
-    XCTAssertEqualObjects(cam.fileExtension, @"heif");
+    XCTAssertEqualObjects([filePath pathExtension], @"heif");
```
hellohuanlin, line 116: This is odd - filePath is a local variable hard coded to be "test.heif". 

Can you do something like: 

```
OCMStub([mockResult sendSuccessWithData: OCARG_ANY]).andDo(^(NSInvocation *invocation)) {
  NSString *pathArgument;
  [invocation getArgument: &pathArgument at index...];
  XCTAssert( pathArgument contains heif )
});
```

packages/camera/camera/CHANGELOG.md
stuartmorgan: This doesn't match pubspec.yaml, and is failing CI accordingly.

packages/camera/camera_avfoundation/example/ios/RunnerTests/FLTCamPhotoCaptureTests.m
Mairramer: I'm not sure if this will work. Because here I will need to pass the path to  delegate.completionHandler;

Mairramer: I changed this part for testing and it worked as expected:

```
  FLTSavePhotoDelegate *savePhotoDelegate = [[FLTSavePhotoDelegate alloc]
           initWithPath:path
                ioQueue:self.photoIOQueue
      //changed here to use path directly only to test,because completionHandler needs to path: 
       completionHandler:^(NSString *_Nullable path2, NSError *_Nullable error) {
        typeof(self) strongSelf = weakSelf;
        if (!strongSelf) return;
        dispatch_async(strongSelf.captureSessionQueue, ^{
          // cannot use the outter `strongSelf`
          typeof(self) strongSelf = weakSelf;
          if (!strongSelf) return;
          [strongSelf.inProgressSavePhotoDelegates removeObjectForKey:@(settings.uniqueID)];
        });

        if (error) {
          [result sendError:error];
        } else {
          NSAssert(path, @"Path must not be nil if no error.");
          [result sendSuccessWithData:path];
        }
      }];
```
Maybe I should change the FLTSavePhotoDelegate or is there some way I'm not seeing?

packages/camera/camera_platform_interface/lib/src/types/file_format.dart
```diff
@@ -0,0 +1,30 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// The format in which images should be returned from the camera.
+enum FileFormat {
```
bparrishMines, line 6: This seems redundant to `ImageFormatGroup`: https://github.com/flutter/packages/blob/main/packages/camera/camera_platform_interface/lib/src/types/image_format_group.dart.

Should `heif` just bee added to this?

Mairramer: Correct me if I'm wrong, but these are for pixel definitions, right?

Mairramer:
> Setting the file format for images seems like a feature that would be supported on every platform. For us to accept this PR, we would probably need to review an implementation for each platform. This is to ensure the `platform_interface` change will work for every platform.

In this case, just the dart code implementation layer?

hellohuanlin (approved):
the objc part looks good

packages/camera/camera_platform_interface/lib/src/types/file_format.dart
bparrishMines: Yea, I think you're right. The issue is that we already have jpeg in the other one and it seems we are adding it in two places. But I do agree we should probably make this distinction.

I would change the name of this to `ImageFileFormat`.

packages/camera/camera_platform_interface/lib/src/types/file_format.dart
```diff
@@ -0,0 +1,30 @@
...
+  jpeg,
+
+  /// It will return the image in HEIF format.
+  /// HEIF is a file format name that refers to High Efficiency Image Format (HEIF).
+  /// Will works only iOS 11+.
+  heif,
+}
+
+/// Extension on [FileFormat] to stringify the enum
+extension FileFormatName on FileFormat {
```
bparrishMines, line 18: This extension should no longer be necessary. You should be able to use `FileFormat.jpeg.name`.

bparrishMines:
> > Setting the file format for images seems like a feature that would be supported on every platform. For us to accept this PR, we would probably need to review an implementation for each platform. This is to ensure the `platform_interface` change will work for every platform.
> 
> In this case, just the dart code implementation layer?

No, as in Android, web, and Windows should also support setting the file format. Therefore, this PR would also need to update the `camera_android`, `camera_web`, etc... as well. We don't typically add support for just one platform unless a feature is only supported for a single platform.

Mairramer:
> > > Setting the file format for images seems like a feature that would be supported on every platform. For us to accept this PR, we would probably need to review an implementation for each platform. This is to ensure the `platform_interface` change will work for every platform.
> > 
> > 
> > In this case, just the dart code implementation layer?
> 
> No, as in Android, web, and Windows should also support setting the file format. Therefore, this PR would also need to update the `camera_android`, `camera_web`, etc... as well. We don't typically add support for just one platform unless a feature is only supported for a single platform.

The idea was just to add support for iOS, but I think it's worth adding for other platforms. Although it only has one value for these platforms for now.

packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
```diff
@@ -298,4 +298,9 @@ abstract class CameraPlatform extends PlatformInterface {
   Future<void> dispose(int cameraId) {
     throw UnimplementedError('dispose() is not implemented.');
   }
+
+  /// Sets the output format for the selected camera.
+  Future<void> setFileFormat(int cameraId, ImageFileFormat format) {
```
bparrishMines, line 303: I think we should change this to `setImageFileFormat` to distinguish it from video and image streaming formats.

bparrishMines (commented):
Following [Platform Support](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#platform-support) in the contributing doc, I think we can add support for the feature on iOS, but we won't add it to the app facing (`camera`) package until the other platforms land support.

You can remove your changes to the `camera` package and I'll do a final review on the interface changes.

packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
Mairramer: I will add support for other platforms. Only on Android it will still take some time, because I will look at the best way to do this.

Mairramer:
@bparrishMines I will add support for other platforms. Only on Android it will still take some time, because I will look at the best way to do this.

packages/camera/camera/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## NEXT
 
-* Updates minimum supported SDK version to Flutter 3.10/Dart 3.0.
+* Updates minimum supported SDK version to Flutter 3.10/Dart 3.
```
bparrishMines, line 3: It looks like there are still some changes in the `camera` package:

```suggestion
* Updates minimum supported SDK version to Flutter 3.10/Dart 3.0.
```

packages/camera/camera/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -169,7 +169,7 @@
 		97C146E61CF9000F007C117D /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastUpgradeCheck = 1300;
+				LastUpgradeCheck = 1430;
```
bparrishMines, line 172: And here

```suggestion
				LastUpgradeCheck = 1300;
```

packages/camera/camera/example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "1300"
+   LastUpgradeVersion = "1430"
```
bparrishMines, line 3: And here

```suggestion
   LastUpgradeVersion = "1300"
```

packages/camera/camera/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: A Flutter plugin for controlling the camera. Supports previewing
   Dart.
 repository: https://github.com/flutter/packages/tree/main/packages/camera/camera
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+camera%22
-version: 0.10.5+5
+version:  0.10.5+5
```
bparrishMines, line 7: And here

```suggestion
version: 0.10.5+5
```

packages/camera/camera/pubspec.yaml
```diff
@@ -22,8 +22,8 @@ flutter:
 
 dependencies:
   camera_android: ^0.10.7
-  camera_avfoundation: ^0.9.13
-  camera_platform_interface: ^2.5.0
+  camera_avfoundation: ^0.9.13+2
+  camera_platform_interface: ^2.5.1
```
bparrishMines, line 26: And here

```suggestion
  camera_avfoundation: ^0.9.13
  camera_platform_interface: ^2.5.0
```

packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
```diff
@@ -298,4 +298,9 @@ abstract class CameraPlatform extends PlatformInterface {
   Future<void> dispose(int cameraId) {
     throw UnimplementedError('dispose() is not implemented.');
   }
+
+  /// Sets the output format for the selected camera.
+  Future<void> setImageFileFormat(int cameraId, ImageFileFormat format) {
+    throw UnimplementedError('setFileFormat() is not implemented.');
```
bparrishMines, line 304: nit:

```suggestion
    throw UnimplementedError('setImageFileFormat() is not implemented.');
```

packages/camera/camera_platform_interface/lib/src/types/image_file_format.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// The format in which images should be returned from the camera.
+enum ImageFileFormat {
+  /// This is the default value. It will return the image in JPEG format.
+  /// Is the widely used format for images.
```
bparrishMines, line 8: This should follow: https://dart.dev/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph. Same for `heif` below.

Also, I don't think this should mention that this is the default value. You should add a comment to `setImageFileFormat` that JPEG is the default.

packages/camera/camera_platform_interface/test/types/image_file_format_test.dart
bparrishMines, line 1: Since you are now using the `Enum.name` feature, this file and test shouldn't be necessary and can be removed.

packages/camera/camera_platform_interface/lib/src/method_channel/method_channel_camera.dart
```diff
@@ -529,6 +529,17 @@ class MethodChannelCamera extends CameraPlatform {
     );
   }
 
+  @override
+  Future<void> setImageFileFormat(int cameraId, ImageFileFormat format) {
```
bparrishMines, line 533: This needs a test

bparrishMines (commented):
This platform interface is still

bparrishMines (approved):
LGTM for the interface. I went ahead and updated some of the documentation as well. You can now create a new PR with just the platform interface changes and we can review and land that first.

jmagman:
https://github.com/flutter/packages/pull/5593 merged, looks like there are conflicts with the CHANGELOG and Xcode project.  Can you rebase?
@bparrishMines is there anything else needed to get this merged?

PR #4616, opened 2023-07-31
[ci] Add emulator tests on older Android version

stuartmorgan:
Adds new tasks to run integration tests on legacy Android emulators (API 22, since we don't currently have API 21 emulators available), so that we are getting more complete coverage of our supported OS versions.

For now, excludes several plugins with TODOs and issues filed, so that the suite can be brought online in parallel with investigation of those issues.

Most of https://github.com/flutter/flutter/issues/130010


script/configs/exclude_integration_android_legacy_emulator.yaml
```diff
@@ -0,0 +1,13 @@
+# Tihs plugin no-ops below API 25, which is newer than the legacy emulator.
```
reidbaker, line 1: this not tihs

.ci.yaml
```diff
@@ -529,6 +529,90 @@ targets:
...
+    timeout: 60
+    bringup: true # New target
+    properties:
+      target_file: android_legacy_emulator_tests.yaml
+      channel: master
+      version_file: flutter_master.version
+      package_sharding: "--shardIndex 3 --shardCount 6"
+      dependencies: >-
+        [
+          {"dependency": "android_virtual_device", "version": "22"}
```
reidbaker, line 585: Out of scope for this pr but I wish we could use variables for these shards since they cant actually have different dependencies.

reidbaker (approved):

.ci.yaml
stuartmorgan: I could make a new config for legacy emulator Android tests, but without a way to have one `platform_properties` entry inherit from another one that sort of just moves the problem (since I'd have to instead duplicate all the core Android dependencies, which can't really be different either). Would you prefer that? I can see an argument for it being less error-prone.

reidbaker (approved):

.ci.yaml
stuartmorgan: Since this is a whole new set of shards with configs, it seems in scope to have it start the way you'd prefer it; I'll leave the existing ones alone for now though (especially so I don't mess up Camille's WIP PR).

stuartmorgan: Let me know what you think of this updated version :)

reidbaker (approved):
I like this change!

PR #3682, opened 2023-04-10
[google_maps_flutter] Cloud-based map styling support

amuramoto:
Recreates https://github.com/flutter/plugins/pull/6553 form flutter/plugins which had approvals in-progress. 

Fixes https://github.com/flutter/flutter/issues/67631

PR #4638, opened 2023-08-03
[google_maps_flutter] cloud-based map styling implementation

jokerttu:
This PR is sub-PR splitted out from the https://github.com/flutter/packages/pull/3682
containing only following packages:
 - google_maps_flutter_web
 - google_maps_flutter_android
 - google_maps_flutter_ios

Related to issue https://github.com/flutter/flutter/issues/67631

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -740,6 +740,7 @@ Map<String, Object> _jsonForMapConfiguration(MapConfiguration config) {
     if (config.trafficEnabled != null) 'trafficEnabled': config.trafficEnabled!,
     if (config.buildingsEnabled != null)
       'buildingsEnabled': config.buildingsEnabled!,
+    if (config.cloudMapId != null) 'cloudMapId': config.cloudMapId!,
```
stuartmorgan, line 743: We should have Dart unit tests that validate that this is being passed.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
```diff
@@ -630,6 +630,7 @@ Map<String, Object> _jsonForMapConfiguration(MapConfiguration config) {
     if (config.trafficEnabled != null) 'trafficEnabled': config.trafficEnabled!,
     if (config.buildingsEnabled != null)
       'buildingsEnabled': config.buildingsEnabled!,
+    if (config.cloudMapId != null) 'cloudMapId': config.cloudMapId!,
```
stuartmorgan, line 633: Same here.

stuartmorgan (changes requested):
I left a couple of comments about some small testing improvements that should be easy to add.

Unfortunately the overall testing of this feature is quite thin; the Dart integration test doesn't actually test that anything works beyond not crashing (i.e., they would pass even if we reverted all of the native changes), but given that we don't have real API testing working in general I don't see a way to do address that without blocking this feature indefinitely, so we'll have to live with that for now.

(Ideally we'd also do testing of the native code, but we don't have the native test scaffolding set up for that; I'll file an issue for that rather than blocking this.)

ditman (approved):
Web bits LGTM.

I agree with Stuart that the testing is very bare bones, but this is the downside of a cloud-based feature, the plugin is completely unaware of what the endpoints end up doing.

Every testing solution I can think of is pretty brittle; for example: taking a screenshot and comparing against a reference image, or (in the web) inspecting the inner DOM of the map and see if we can find the mapId somewhere, or traces of WebGL ;)

stuartmorgan:
Yes, I think the best we'll be able to do is just test that we're passing it through both the Dart->native (what I requested above) and native->SDK layers (filed as https://github.com/flutter/flutter/issues/131851 for follow-up). Once we have that, the only plausible breakages I can think of on our side would be if that was insufficient (e.g., if at some point there were a requirement that some other data be passed with the cloud ID in order for it to work, in which case our tests would still pass but the feature wouldn't), but that's relatively low risk.

ditman:
Yep, there's not much more we can do. The only breaking scenario that I can think of is if `mapId` makes some of the other properties that we accept incompatible/obsolete (imagine a map that happens to have "cloud styling" applied suddenly starts rejecting the local mapStyle property with an error)

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
jokerttu: Added dart unit test to check that cloudMapId is passed through the PlatformView creation message.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
jokerttu: Added dart unit test to check that cloudMapId is passed through the PlatformView creation message.

jokerttu:
I have added a few Dart unit tests as requested. These tests verify that the given cloudMapId is passed to the platform view via the method channel upon its creation. 

Since the Maps SDK doesn't provide a means to retrieve the passed cloudMapId after a map instance is initialized, or even a status to indicate if the map has loaded with the applied cloud-based styling, I don't see any alternatives for effectively checking whether the cloud-based map styling is really applied, other than visual test comparison using screenshots. Having a non-public cloudMapId on test cases to which not all developers have access could result in failed tests outside the CI pipeline. At this moment, I personally do not know of any other way to test if the cloudMapId is genuinely functioning with the Maps SDK, aside from manual testing.

reidbaker (commented):
Leaving final approval for stuart but android and dart bits LGTM.

stuartmorgan (approved):
Sorry for the delay, LGTM. Thanks for adding the tests!

stuartmorgan:
@cedvdb Please report issues with landed code in the issue tracker; changes can't be made here as this is already merged.

cedvdb:
@stuartmorgan will do, I was too lazy to make a repro but I can probably find the strength to do it today.

kornha:
I cannot get this working for me; I have a mapId, and a linked style. When I try to set the ID the code:
 ```
 static const CameraPosition _kGooglePlex = CameraPosition(
    target: LatLng(37.42796133580664, -122.085749655962),
    zoom: 14.4746,
  );

    return const SizedBox(
      width: 250, // some padding for map
      height: 250,
      child: GoogleMap(
        cloudMapId: "MY ID HERE",
        initialCameraPosition: _kGooglePlex,
      ),
    );
```
I'm running in iOS. Futhermore, I cannot find instructions anywhere how to use this in Flutter. Can anyone assist me?

PR #4639, opened 2023-08-03
[ci] Enable Android maps tests for emulators

stuartmorgan:
With the recent flaky test disabling, this enables the maps tests in emulators.

Fixes https://github.com/flutter/flutter/issues/130986


stuartmorgan:
A different test was disabled, so this is still broken.

PR #4483, opened 2023-07-16
[path_provider] Add getApplicationCachePath()

jpnurmi:
Provides a suitable place for caching application-specific files on all supported platforms.

| Platform | Location |
|---|---|
| Android | `<app>/cache` |
| iOS | `<app>/Library/Caches` |
| Linux | `$XDG_CACHE_HOME/<app>` or `~/.cache/<app>` |
| macOS | `~/Library/Caches/<app>` or `~/Library/Containers/<app>/Data/Library/Caches/<app>` |
| Windows | `%LOCALAPPDATA%/<app>` |

Fixes: flutter/flutter#105386

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[...]

PR #4647, opened 2023-08-04
[tool] Skip pathified analysis on resolver errors

stuartmorgan:
If adding a pathified dependency creates a resolver error, then skip it instead of failing when running pathified analysis. The purpose of pathified analysis it to pre-detect failures that would happen on publishing, and if there's a resolver error that means the publishing even won't affect the package anyway.

See https://github.com/flutter/packages/pull/4483#issuecomment-1664468621 for an example case where we need this.

(In theory we could get delayed OOB errors that this will miss—e.g., in the case above if the PR would actually break rfw/example/wasm, then if at some later date `wasm` updated to use a newer `ffi`, eliminating the resolver conflict, then suddenly rfw/example/wasm would pick up the PR and break. That seems *extremely* unlikely, however, so I'm not concerned that this will be a problem in practice. We can revisit if that changes.)

### DIFF
--- a/script/tool/test/analyze_command_test.dart
+++ b/script/tool/test/analyze_command_test.dart
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:io';
-
 import 'package:args/command_runner.dart';
 import 'package:file/file.dart';
 import 'package:file/memory.dart';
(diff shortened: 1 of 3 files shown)


script/tool/test/analyze_command_test.dart
```diff
@@ -315,6 +315,39 @@ void main() {
     });
   });
 
+  test('skips if requseted if "pub get" fails in the resolver', () async {
```
tarrinneal, line 318: requested

tarrinneal (approved):
one typo but NOWWTO

PR #3973, opened 2023-05-12
[url_launcher] Set broadcast reciever visability as required by target api 34

reidbaker:
- Update the version of androidx used.
- Explicitly set Exported on previously exported broadcast reciever
[126460](https://github.com/flutter/flutter/issues/126460)

This change is a no-op in behavior. I am open to the argument that instead we make this configurable. 

https://developer.android.com/guide/components/broadcasts#context-registered-receivers

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] All existing and new tests are passing.

PR #4660, opened 2023-08-07
[url_launcher] Android API 34 support

reidbaker:
fixed [#flutter/flutter/issues/126460 ](https://github.com/flutter/flutter/issues/126460)
Reland of https://github.com/flutter/packages/pull/3973 which incorrectly used application context when registering and not when unregistering. This pr uses activity context for both which is aligned with what the behavior was before. 

- Register on the same context as  unregister
- Add integration test for android url launch and close

Tested with manual test and new integration test. 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
```diff
@@ -20,4 +20,27 @@ void main() {
...
+        await launcher.launch('https://flutter.dev',
+            useSafariVC: true,
+            useWebView: true,
+            enableJavaScript: false,
+            enableDomStorage: false,
+            universalLinksOnly: false,
+            headers: <String, String>{'my_header_key': 'my_header_value'}),
+        true);
+    // Allow time for page to load.
+    await Future.delayed(Duration(seconds: 3));
```
reidbaker, line 38: @bparrishMines @stuartmorgan open to suggestions for how I can make this less hacky. For example some flutter event I can wait on during a normal shutdown. If I modify the android code to cause a crash and I dont await here the tests pass when they should not.

reidbaker:
also fixes https://b.corp.google.com/issues/294558650

packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
```diff
@@ -20,4 +20,27 @@ void main() {
...
+            enableJavaScript: false,
+            enableDomStorage: false,
+            universalLinksOnly: false,
+            headers: <String, String>{'my_header_key': 'my_header_value'}),
+        true);
+    // Allow time for page to load.
+    await Future<void>.delayed(const Duration(seconds: 3));
+    await launcher.closeWebView();
+    // Delay required to catch android side crashes in onDestroy
+    await Future<void>.delayed(const Duration(seconds: 3));
```
stuartmorgan, line 41: I think we would need https://github.com/flutter/flutter/issues/20034 to remove this; a TODO referencing that issue wouldn't hurt.

In the meantime, do we really need 3 seconds to reliably catch closing the page? That seems crazy long.

packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
```diff
@@ -20,4 +20,27 @@ void main() {
     // sms:, tel:, and mailto: links may not be openable on every device, so
     // aren't tested here.
   });
+
+  testWidgets('launch and close', (WidgetTester _) async {
+    final UrlLauncherPlatform launcher = UrlLauncherPlatform.instance;
+
+    // Launch a url then close.
+    expect(
+        await launcher.launch('https://flutter.dev',
```
stuartmorgan, line 29: We shouldn't actually launch real external URLs when we can avoid it. See https://github.com/flutter/packages/blob/ac4137624a13c87f20781795b9a56891cdadff11/packages/webview_flutter/webview_flutter/example/integration_test/webview_flutter_test.dart#L25-L43 for a better option for integration tests.

stuartmorgan (changes requested):

packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
reidbaker: Oh thanks I can convert to this.

packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
reidbaker: 3s was the shortest I could make it and have the test reliability catch a crash that can happen in onDestroy on my physical phone. I ended up verifying on an emulator and actually needed to bump the delay to 5s. 3 would not reliably catch the potential crash and 4s was flakey in catching the crash. 

I think we need to wait the full 5s otherwise this test does not actually catch a regression in the code that was fixed.

packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
```diff
@@ -20,4 +20,27 @@ void main() {
...
+            useSafariVC: true,
+            useWebView: true,
+            enableJavaScript: false,
+            enableDomStorage: false,
+            universalLinksOnly: false,
+            headers: <String, String>{'my_header_key': 'my_header_value'}),
+        true);
+    // Allow time for page to load.
+    await Future<void>.delayed(const Duration(seconds: 3));
```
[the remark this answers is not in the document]
reidbaker: I was able to remove the load delay entirely.

packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
```diff
@@ -20,4 +23,44 @@ void main() {
...
+
+    // Launch a url then close.
+    expect(
+        await launcher.launch(primaryUrl,
+            useSafariVC: true,
+            useWebView: true,
+            enableJavaScript: false,
+            enableDomStorage: false,
+            universalLinksOnly: false,
+            headers: <String, String>{'my_header_key': 'my_header_value'}),
```
stuartmorgan, line 52: The headers map can be empty since they aren't related to the test.

packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
```diff
@@ -20,4 +23,44 @@ void main() {
...
+    // Delay required to catch android side crashes in onDestroy.
+    //
+    // If this test flakes with an android crash during this delay the test
+    // should be considered failing because this integration test can have a
+    // false positive pass if the test closes before an onDestroy crash.
+    // See https://github.com/flutter/flutter/issues/126460 for more info.
+    await Future<void>.delayed(const Duration(seconds: 5));
+
+    // sms:, tel:, and mailto: links may not be openable on every device, so
+    // aren't tested here.
```
stuartmorgan, line 64: This comment is copypasta.

stuartmorgan (approved):
LGTM with nits.

Thanks for investigating more on the delay. It's not ideal, but at least this is an integration test where taking longer is expected.

packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
reidbaker: done

packages/url_launcher/url_launcher_android/example/integration_test/url_launcher_test.dart
reidbaker: facepalm, fixed

PR #4677, opened 2023-08-10
[dynamic_layouts] relax layout test to fix tree

tarrinneal:
A current roll from the engine to flutter/flutter to packages has broken this test value by a small amount. 

I wrote a small function to manage tolerances.

Let me know if this is a reasonable way to manage this for now.


Piinks (approved):
LGTM, I will file an issue to follow up on these tests. Thy appear to have become rather brittle to text changes.

Piinks:
Filed https://github.com/flutter/flutter/issues/132321

PR #4682, opened 2023-08-11
[pigeon] skipGen

tarrinneal:
adds ability to run pigeon tests without regenerating code (to make iterating and testing easier).

### DIFF
--- a/packages/pigeon/tool/shared/test_runner.dart
+++ b/packages/pigeon/tool/shared/test_runner.dart
@@ -14,11 +14,11 @@ import 'test_suites.dart';
 /// Runs the given tests, printing status and exiting with failure if any of
 /// them fails.
 Future<void> runTests(List<String> testsToRun, {bool skipGen = false}) async {
-  // Pre-generate the necessary common output files.
-  // TODO(stuartmorgan): Consider making this conditional on the specific
-  // tests being run, as not all of them need these files.
-  final String baseDir = p.dirname(p.dirname(Platform.script.toFilePath()));
   if (!skipGen) {
+    // Pre-generate the necessary common output files.
+    // TODO(stuartmorgan): Consider making this conditional on the specific
+    // tests being run, as not all of them need these files.
+    final String baseDir = p.dirname(p.dirname(Platform.script.toFilePath()));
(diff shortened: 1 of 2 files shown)


packages/pigeon/tool/shared/test_runner.dart
```diff
@@ -13,17 +13,19 @@ import 'test_suites.dart';
 
 /// Runs the given tests, printing status and exiting with failure if any of
 /// them fails.
-Future<void> runTests(List<String> testsToRun) async {
+Future<void> runTests(List<String> testsToRun, {bool skipGen = false}) async {
```
stuartmorgan, line 16: Prefer bools that are affirmative to avoid double-negatives. So this should be `runGeneration = true`, and the condition below reversed.

packages/pigeon/tool/test.dart
```diff
@@ -105,5 +107,5 @@ ${parser.usage}''');
     }
   }
 
-  await runTests(testsToRun);
+  await runTests(testsToRun, skipGen: true);
```
stuartmorgan, line 110: Presumably this was supposed to use the flag?

packages/pigeon/tool/test.dart
```diff
@@ -18,11 +18,13 @@ import 'shared/test_runner.dart';
 import 'shared/test_suites.dart';
 
 const String _testFlag = 'test';
+const String _genFlag = 'skip-generation';
 const String _listFlag = 'list';
 
 Future<void> main(List<String> args) async {
   final ArgParser parser = ArgParser()
     ..addMultiOption(_testFlag, abbr: 't', help: 'Only run specified tests.')
+    ..addFlag(_genFlag, abbr: 'g', help: 'Skips the generation step.')
```
stuartmorgan, line 27: You probably want to make this non-negatable (or referring or whatever it's called, where by default you can do `-no-whateverflag`).

stuartmorgan (changes requested):

packages/pigeon/tool/test.dart
tarrinneal: lol

stuartmorgan (approved):

PR #4317, opened 2023-06-26
[path_provider] added android support for getDownloadsDirectory

zopagaduanjr:
Followed step 3. of this approach https://github.com/flutter/plugins/pull/4559#pullrequestreview-819322914

Fixes flutter/flutter#93198

originally from #4167 , but I renamed the branch hence it was closed

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #4708, opened 2023-08-15
[path_provider] Fix93198: Added getDownloadsDirectory() for Android

talhakhan1297:
This PR adds the Android implementation for the getDownloadsDirectory() function by making getDownloadsDirectory redirect at the Dart level to getExternalStorageDirectories with the downloads directory constant.

Fixes [#93198](https://github.com/flutter/flutter/issues/93198)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/path_provider/path_provider_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.1.1
+
+- Added implementation of getDownloadsDirectory for Android
```
stuartmorgan, line 3: Please see https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style discussion of verbs and punctuation.

Also, please use the same list bullet format as the rest of the file.

packages/path_provider/path_provider_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.1.1
```
stuartmorgan, line 1: Since this is new functionality, it should be 2.2.0.

stuartmorgan (changes requested):
Thanks for the contribution!

talhakhan1297:
@stuartmorgan  I have made the changes, please have a look 🙂.

packages/path_provider/path_provider_android/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 2.2.0
+
+* Adds implementation of getDownloadsDirectory for Android.
 ## NEXT
```
stuartmorgan, line 4: Please see https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#updating-a-changelog-that-has-a-next

packages/path_provider/path_provider_android/test/path_provider_android_test.dart
```diff
@@ -98,13 +98,9 @@ void main() {
...
     } // end of for-loop
 
-    test('getDownloadsPath fails', () async {
-      try {
-        await pathProvider.getDownloadsPath();
-        fail('should throw UnsupportedError');
-      } catch (e) {
-        expect(e, isUnsupportedError);
-      }
+    test('getDownloadsPath succeeds', () async {
```
stuartmorgan, line 101: I missed this before; you're only testing one of three different codepaths. Please test both the null and empty platform channel conditions that the code is intended to handle.

See discussion in https://github.com/flutter/packages/pull/4317, which is another PR intended to add this functionality, but which hasn't been completed.

stuartmorgan (changes requested):

packages/path_provider/path_provider_android/test/path_provider_android_test.dart
talhakhan1297: Added the tests, please have a look. @stuartmorgan

packages/path_provider/path_provider_android/example/lib/main.dart
```diff
@@ -199,6 +206,15 @@ class _MyHomePageState extends State<MyHomePage> {
...
                 future: _externalCacheDirectories, builder: _buildDirectories),
+            Padding(
+              padding: const EdgeInsets.all(16.0),
+              child: ElevatedButton(
+                onPressed: _requestDownloadsDirectory,
+                child: const Text('Get Downloads Directory'),
+              ),
+            ),
+            FutureBuilder<String?>(
+                future: _externalDownloadsDirectory, builder: _buildDirectory),
```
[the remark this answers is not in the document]
stuartmorgan: No, that will need to be a follow-up PR. (This is why going forward we'll be using https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#api-support-queries)

packages/path_provider/path_provider_android/pigeons/messages.dart
```diff
@@ -42,5 +42,5 @@ abstract class PathProviderApi {
   @TaskQueue(type: TaskQueueType.serialBackgroundThread)
   List<String?> getExternalCachePaths();
   @TaskQueue(type: TaskQueueType.serialBackgroundThread)
-  List<String?> getExternalStoragePaths(StorageDirectory directory);
+  List<String?>? getExternalStoragePaths(StorageDirectory directory);
```
stuartmorgan, line 45: Why was this changed? We generally do not use nullable collections, as they are error-prone.

packages/path_provider/path_provider_android/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 2.2.0
 
+* Adds implementation of getDownloadsDirectory.
```
stuartmorgan, line 3: Nit: backticks around `getDownloadsDirectory` please.

packages/path_provider/path_provider_android/test/path_provider_android_test.dart
```diff
@@ -98,13 +115,27 @@ void main() {
...
+      final String? path = await pathProvider.getDownloadsPath();
+      expect(path, kExternalStoragePaths);
+    });
+
+    test('getDownloadsPath fails with null', () async {
```
stuartmorgan, line 123: I would not describe these as "fails"; they are returning empty lists as expected for those cases.

stuartmorgan (changes requested):

stuartmorgan (changes requested):
Oops, cleared my own review state on accident.

packages/path_provider/path_provider_android/pigeons/messages.dart
talhakhan1297: Because, according to `PathProviderPlatform` class the `getExternalStoragePaths` returns a nullable List<String>, this is one of the cases that I had to test. But the generated class `TestPathProviderApi`'s function `getExternalStoragePaths` returns List<String?> that is not nullable. If this is also not nullable I can't return a null from this function to mock that case.

stuartmorgan: We don't want to make production code worse to artificially create an extra codepath just so that we test it. If it's impossible to actually get a null result, then the solution is to extract a helper that returns a non-nullable list which both the new method and `getExternalStoragePaths` can delegate to, such that `getDownloadsDirectory` doesn't need to have dead code that handles a case that can't actually happen.

talhakhan1297: Got it 👍. I will fix this.

talhakhan1297: @stuartmorgan I have reverted the changes and extracted that function to return a non-nullable list. Please have a look.

packages/path_provider/path_provider_android/lib/path_provider_android.dart
```diff
@@ -81,12 +81,20 @@ class PathProviderAndroid extends PathProviderPlatform {
...
+    return _getExternalStoragePaths(type: type);
   }
 
   @override
-  Future<String?> getDownloadsPath() {
-    throw UnsupportedError('getDownloadsPath is not supported on Android');
+  Future<String?> getDownloadsPath() async {
+    final List<String> paths =
+        await _getExternalStoragePaths(type: StorageDirectory.downloads);
+    return paths.isNotEmpty ? paths.first : null;
```
stuartmorgan, line 91: Prefer conditions without negation when when using `?:` or if/else, to avoid the implicit double-negation for the "else" branch (i.e., currently this is "if paths is not empty, return the first path; if paths is not not empty, return null). So `return paths.isEmpty ? null : paths.first;`

packages/path_provider/path_provider_android/test/path_provider_android_test.dart
```diff
@@ -17,7 +17,16 @@ const String kExternalCachePaths = 'externalCachePaths';
...
 
+enum ExternalStoragePathsValueState { isEmpty, isNotEmpty }
+
 class _Api implements TestPathProviderApi {
+  _Api({
+    this.externalStoragePathsValueState =
+        ExternalStoragePathsValueState.isNotEmpty,
+  });
+
+  final ExternalStoragePathsValueState externalStoragePathsValueState;
```
stuartmorgan, line 28: This can just be `final bool returnsExternalStoragePaths;` now, and the enum can be removed.

stuartmorgan (changes requested):
Thanks, just a couple of minor nits.

packages/path_provider/path_provider_android/test/path_provider_android_test.dart
```diff
@@ -34,8 +38,13 @@ class _Api implements TestPathProviderApi {
...
 
   @override
-  List<String?> getExternalStoragePaths(messages.StorageDirectory directory) =>
-      <String>[kExternalStoragePaths];
+  List<String?> getExternalStoragePaths(messages.StorageDirectory directory) {
+    if (returnsExternalStoragePaths) {
+      return <String?>[kExternalStoragePaths];
+    } else {
+      return <String?>[];
+    }
```
stuartmorgan, line 46: Optional nit: in modern Dart this whole implementation can be condensed to:
```dart
return <String?>[ if (returnsExternalStoragePaths) kExternalStoragePaths ];
```

stuartmorgan, line 46: Optional nit: in modern Dart this whole implementation can be condensed to:
```dart
return <String?>[ if (returnsExternalStoragePaths) kExternalStoragePaths ];
```

stuartmorgan (approved):
LGTM

@camsim99 for second approval

camsim99 (approved):
LGTM!

nirmalraj2000:
Looks like `getDownloadsDirectory()` functionality is added to `path_provider` for `android`, but still mentioned as not supported in the documentation. 

I have tested in android emulator and I was able to get the path. Is that mentioned by mistake in documentation.

Screenshot 2024-07-16 at 4 23 53 PM

stuartmorgan:
@nirmalraj2000 That was an oversight; I've filed https://github.com/flutter/flutter/issues/151823 to track it. In the future, please use the issue tracker to report issues rather than commenting on old PRs.

PR #4728, opened 2023-08-17
 getting 403 disallowed user agent

pese-git:
Hello!

I would like to present to you a fix that I have made to the `url_launcher_android` package, addressing an issue related to user authorization through Google OAuth.

Problem Description:
When attempting to authorize users via Google OAuth, a `403 disallow_useragent` error was occurring. After investigating the issue, I discovered that the problem was due to the absence of a User-Agent string being set for the `WebViewActivity`. In my fix, I added the following code to the `onCreate` method of the `WebViewActivity` class:

```java
webview.getSettings().setUserAgentString(System.getProperty("http.agent"));
```

### DIFF
--- a/packages/url_launcher/url_launcher_android/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.0.39
+
+* The fix addresses a "403 disallow_useragent" error during Google OAuth user authorization
+
 ## 6.0.38
 
 * Updates android implementation to support api 34 broadcast receiver requirements.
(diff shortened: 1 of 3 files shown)


stuartmorgan:
Thanks for the contribution, but this PR is not something we would accept.

> I discovered that the problem was due to the absence of a User-Agent string being set for the `WebViewActivity`.

The web view has a user agent; the problem is that the content you are accessing is blocking that user agent. This is "fixing" the issue by breaking that detection, due to giving the webview the wrong user agent. See https://github.com/flutter/flutter/issues/18589#issuecomment-1681686862 for relevant discussion.

PR #4738, opened 2023-08-18
[image_picker] Copy exif tags in categories II and III

RobertOdrowaz:
Added all exif tags from Category II and Category III (excluding ones that don't have consts in `ExifInterface` WaterDepth etc.). Those tags include data not related to the structure of the file itself so resizing definitely doesn't invalidate them and thus they should be copied over.

I've also switched from list of raw strings to consts provided be the `ExifInterface`

I'm not sure show to handle `TAG_ISO_SPEED_RATINGS` deprecation in this case. The const is deprecated and the tag itself is deprecated in the standard however it would be reasonable to copy it as well to handle legacy files which include it. It was also being copied in the previous versions of this package.

https://github.com/flutter/flutter/issues/132827

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan (changes requested):
Thanks for the contribution!

> * [x]  I read and followed the [relevant style guides](https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style) and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)

The CI format check is failing; please ensure that the Java is auto-formatted as well.

> * [x]  I added new tests to check the change I am making, or this PR is [test-exempt](https://github.com/flutter/flutter/wiki/Tree-hygiene#tests).

This change is not exempt, so will need tests before we can proceed with the review.

gmackall:
Thanks for adding tests! I'll take a look this afternoon 🙂

gmackall (commented):
Thanks for the contribution! This change mostly LGTM, but do you have a link to documentation on where Exif tags are broken into categories? I couldn't find anything clear just by searching. 

Also, would you mind adding documentation to `copyExif` explaining why we select this subset of Exif tags? It probably should have some documentation covering that, and this change seems like a good time to add it.

RobertOdrowaz:
Thanks for the review. https://www.cipa.jp/std/documents/download_e.html?DC-008-Translation-2023-E here you can download the standard. On page 234 there is a table explaining categories and on page 238 is a table matching tags to categories.

I will add the explanation regarding tags chosen later today.

RobertOdrowaz:
FYI couple of tags are missing eg. WaterDepth because `ExifInterface` is based on older 2.3 exif standard and those were introduced in 2.31. I believe we could copy them as well but we would need to go back to using raw string values

gmackall (approved):
This LGTM! @stuartmorgan do you want to do secondary review or shall I find another reviewer?

stuartmorgan:
@tarrinneal is the ecosystem team codeowner for this plugin, so would probably be the best person to do the secondary review.

tarrinneal (approved):
lgtm, thanks for putting this together!

stuartmorgan:
It looks like this just needs conflicts resolved, then we can land it.

stuartmorgan (approved):

PR #4739, opened 2023-08-18
[url_launcher_android] Add support for Custom Tabs

rajveermalviya:
Implement support for [Android Custom Tabs](https://developer.chrome.com/docs/android/custom-tabs/).

Custom Tabs will only be used if *__all__* of the following conditions are true:
- `launchMode` == `LaunchMode.inAppWebView` (or `LaunchMode.platformDefault`; only if url is web url)
- `WebViewConfiguration.headers` == `{}` (or if it only contains [CORS-safelisted headers](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header))

Fixes flutter/flutter#18589

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/url_launcher/url_launcher/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.1.14
+
+* Updates documentation to mention support for Android Custom Tabs.
+
 ## 6.1.13
 
 * Adds pub topics to package metadata.
--- a/packages/url_launcher/url_launcher/example/lib/main.dart
+++ b/packages/url_launcher/url_launcher/example/lib/main.dart
@@ -63,6 +63,12 @@ class _MyHomePageState extends State<MyHomePage> {
   }
 
   Future<void> _launchInWebViewOrVC(Uri url) async {
+    if (!await launchUrl(url, mode: LaunchMode.inAppWebView)) {
+      throw Exception('Could not launch $url');
+    }
+  }
+
+  Future<void> _launchAsInAppWebViewWithCustomHeaders(Uri url) async {
     if (!await launchUrl(
       url,
       mode: LaunchMode.inAppWebView,
@@ -170,6 +176,12 @@ class _MyHomePageState extends State<MyHomePage> {
                   _launched = _launchInWebViewOrVC(toLaunch);
                 }),
                 child: const Text('Launch in app'),
+              ),
+              ElevatedButton(
+                onPressed: () => setState(() {
+                  _launched = _launchAsInAppWebViewWithCustomHeaders(toLaunch);
+                }),
+                child: const Text('Launch in app (Custom Headers)'),
               ),
               ElevatedButton(
                 onPressed: () => setState(() {
--- a/packages/url_launcher/url_launcher_android/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.1.0
+
+* Adds support for Android Custom Tabs.
+
 ## 6.0.39
 
 * Adds pub topics to package metadata.
--- a/packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
+++ b/packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
@@ -16,8 +16,10 @@ import android.util.Log;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
+import androidx.browser.customtabs.CustomTabsIntent;
 import io.flutter.plugins.urllauncher.Messages.UrlLauncherApi;
 import io.flutter.plugins.urllauncher.Messages.WebViewOptions;
+import java.util.Locale;
 import java.util.Map;
 
 /** Implements the Pigeon-defined interface for calls from Dart. */
@@ -97,13 +99,24 @@ final class UrlLauncher implements UrlLauncherApi {
     ensureActivity();
     assert activity != null;
 
+    Bundle headersBundle = extractBundle(options.getHeaders());
+
+    // Try to launch using Custom Tabs if they have the necessary functionality.
+    if (!containsRestrictedHeader(options.getHeaders())) {
+      Uri uri = Uri.parse(url);
+      if (openCustomTab(activity, uri, headersBundle)) {
+        return true;
+      }
+    }
+
+    // Fall back to a web view if necessary.
     Intent launchIntent =
         WebViewActivity.createIntent(
             activity,
             url,
             options.getEnableJavaScript(),
             options.getEnableDomStorage(),
-            extractBundle(options.getHeaders()));
+            headersBundle);
     try {
       activity.startActivity(launchIntent);
     } catch (ActivityNotFoundException e) {
@@ -116,6 +129,35 @@ final class UrlLauncher implements UrlLauncherApi {
   @Override
   public void closeWebView() {
     applicationContext.sendBroadcast(new Intent(WebViewActivity.ACTION_CLOSE));
+  }
+
+  private static boolean openCustomTab(
+      @NonNull Context context, @NonNull Uri uri, @NonNull Bundle headersBundle) {
+    CustomTabsIntent customTabsIntent = new CustomTabsIntent.Builder().build();
+    customTabsIntent.intent.putExtra(Browser.EXTRA_HEADERS, headersBundle);
+    try {
+      customTabsIntent.launchUrl(context, uri);
+    } catch (ActivityNotFoundException ex) {
+      return false;
+    }
+    return true;
+  }
+
+  // Checks if headers contains a CORS restricted header.
+  //  https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header
+  private static boolean containsRestrictedHeader(Map<String, String> headersMap) {
+    for (String key : headersMap.keySet()) {
+      switch (key.toLowerCase(Locale.US)) {
+        case "accept":
+        case "accept-language":
+        case "content-language":
+        case "content-type":
+          continue;
+        default:
+          return true;
+      }
+    }
+    return false;
   }
 
   private static @NonNull Bundle extractBundle(Map<String, String> headersMap) {
--- a/packages/url_launcher/url_launcher_android/android/src/test/java/io/flutter/plugins/urllauncher/UrlLauncherTest.java
+++ b/packages/url_launcher/url_launcher_android/android/src/test/java/io/flutter/plugins/urllauncher/UrlLauncherTest.java
@@ -6,9 +6,11 @@ package io.flutter.plugins.urllauncher;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.isNull;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
@@ -128,13 +130,15 @@ public class UrlLauncherTest {
   }
 
   @Test
-  public void openWebView_opensUrl() {
+  public void openWebView_opensUrl_inWebView() {
     Activity activity = mock(Activity.class);
     UrlLauncher api = new UrlLauncher(ApplicationProvider.getApplicationContext());
     api.setActivity(activity);
     String url = "https://flutter.dev";
     boolean enableJavaScript = false;
     boolean enableDomStorage = false;
+    HashMap<String, String> headers = new HashMap<>();
+    headers.put("key", "value");
 
     boolean result =
         api.openUrlInWebView(
@@ -142,7 +146,7 @@ public class UrlLauncherTest {
             new Messages.WebViewOptions.Builder()
                 .setEnableJavaScript(enableJavaScript)
                 .setEnableDomStorage(enableDomStorage)
-                .setHeaders(new HashMap<>())
+                .setHeaders(headers)
                 .build());
 
     final ArgumentCaptor<Intent> intentCaptor = ArgumentCaptor.forClass(Intent.class);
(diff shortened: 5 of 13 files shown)


rajveermalviya:
About tests, the `integration_tests` pass but the native tests fail for me because of an unrelated reason:
```shell
    java.lang.NoSuchMethodError: 'java.util.List android.content.pm.PackageManager.queryIntentActivities(android.content.Intent, android.content.pm.PackageManager$ResolveInfoFlags)'
        at io.flutter.plugins.urllauncher.UrlLauncher.launchNativeBeforeApi30(UrlLauncher.java:157)
        at io.flutter.plugins.urllauncher.UrlLauncher.launchUrl(UrlLauncher.java:93)
        at io.flutter.plugins.urllauncher.UrlLauncherTest.launch_returnsTrue(UrlLauncherTest.java:125)
```

But if the older version of the function is used - [`queryIntentActivities(android.content.Intent, int)`](https://developer.android.com/reference/android/content/pm/PackageManager#queryIntentActivities(android.content.Intent,%20int)) then there are deprecation errors.

The first error may be because of robolectric/robolectric#8215. So, I am not really sure what needs to be done here.

rajveermalviya:
Updated to use older api and suppressing the `deprecated` warning; which seems normal for this repo (looked at neighboring package implementations.)

rajveermalviya:
(separated tests for custom tabs & legacy code paths)

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
```diff
@@ -118,6 +156,87 @@ public void closeWebView() {
     applicationContext.sendBroadcast(new Intent(WebViewActivity.ACTION_CLOSE));
   }
 
+  @TargetApi(Build.VERSION_CODES.R)
+  private static @NonNull Boolean launchNativeApi30(@NonNull Context context, @NonNull Uri uri) {
+    Intent nativeAppIntent =
+        new Intent(Intent.ACTION_VIEW, uri)
+            .addCategory(Intent.CATEGORY_BROWSABLE)
+            .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_REQUIRE_NON_BROWSER);
```
gnprice, line 164: This part is useful functionality, but should be controlled by the launch mode — this is implementing `LaunchMode.externalNonBrowserApplication`.

An implementation of that would resolve https://github.com/flutter/flutter/issues/66721 and would also be quite welcome. It's probably simplest to handle that as a separate PR, though. In order to make it properly conditional on the launch mode, I think it may also require changes in `url_launcher_platform_interface`.

gnprice (changes requested):
Welcome @rajveermalviya, and thanks for the contribution!

This version introduces Chrome custom tabs to replace launching the URL in an external app, for `LaunchMode.externalApplication`. The thread https://github.com/flutter/flutter/issues/18589 is a bit confusing, because that is the approach requested in the original issue description. But I think what we actually would like to do is to instead have Chrome custom tabs replace the custom WebView-based screen, for `LaunchMode.inAppWebView`.

See @stuartmorgan's comment at https://github.com/flutter/flutter/issues/18589#issuecomment-1627721520 suggesting that, and my two comments starting at https://github.com/flutter/flutter/issues/18589#issuecomment-1681590394 about why the existing custom WebView screen isn't a good UX.

Conversely, I think it wouldn't be appropriate for `LaunchMode.externalApplication` to open the URL in a browser that stays within the app, which is what a Chrome custom tab means from a user perspective.

The FLAG_ACTIVITY_REQUIRE_NON_BROWSER part is also useful but should probably be left for a separate PR; see below.

rajveermalviya:
Hello @gnprice :wave: 

> I think what we actually would like to do is to instead have Chrome custom tabs replace the custom WebView-based screen, for LaunchMode.inAppWebView.

Yeah, I went with the updating `LaunchMode.externalApplication` mode because that's what Custom Tabs actually feels like - links are still opened by an external application (a browser in this case) but with a compact look, a _custom tab_.
If the WebView mode is replaced then that will indeed be a breaking change. Also Custom Tabs won't be able to completely replace them anyway, because WebView mode allows passing headers not present in [CORS-safelisted headers list](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header); Custom Tabs can't support that. 
One way would be to break the users and tell them to migrate to using `webview` package for their exotic needs.

Also I wouldn't call them Chrome Custom Tabs, since it is supported by many browsers on Android, and the [doc website](https://developer.chrome.com/docs/android/custom-tabs/) also calls them Android Custom Tabs now. It's API is also part of androidx:browser jetpack library - nothing chrome specific _anymore_ ^^.

Firefox | Chrome
:-: | :-:
firefox | chrome

> The FLAG_ACTIVITY_REQUIRE_NON_BROWSER part is also useful but should probably be left for a separate PR; see below.

That was added to match the behavior of legacy launch (`Intent(Intent.ACTION_VIEW).setData(uri)`). So that there is no breaking change, only changed behavior is that if the link is open-able in a browser it will now open in a Custom Tab and if not it will be handed over to the corresponding native app, if any, the former part remains the same.

stuartmorgan:
Thanks for the contribution, but along the lines of @gnprice's comments above this is not a change we would accept as written.

> Yeah, I went with the updating `LaunchMode.externalApplication` mode because that's what Custom Tabs actually feels like - links are still opened by an external application (a browser in this case) but with a compact look, a _custom tab_.

That is not the way the custom tabs feature is described:

>> Custom Tabs are a feature in Android browsers that gives app developers a way to add a customized browser experience **directly within their app**.

>> Custom Tabs offer a better user experience than simply opening an external browser. **They allow users to remain within the app while browsing**

And importantly, it's also not the way they actually behave from a system perspective. If you open a custom tab and then pull up the app switching UI you will be shown as in the same app, not in your browser. It is closely analogous to the Safari View Controller used for launch-in-app mode on iOS.

> If the WebView mode is replaced then that will indeed be a breaking change.

As currently written, this PR **removes the ability to launch in a the user's browser**, which is absolutely a breaking change. And one that violates the documented and intended behavior of that mode.

> WebView mode allows passing headers not present in [CORS-safelisted headers list](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header); Custom Tabs can't support that.

We need to keep the webview for the foreseeable future anyway, to handle devices that don't have Custom Tab support; we can fall back to that codepath if any unsupported headers are passed (and add a comment to that effect on the header property) so that nothing regresses.

I'm going to mark this as a draft for now; if you are interested in returning this to match the behavior we want for the plugin please mark it as ready for review once those changes are made.

rajveermalviya:
Thanks for clearing the doubts @stuartmorgan - I'll update the PR to use Custom Tabs when `LaunchMode.inAppWebView` 
is used and fallback to actual WebView implementation if Custom Tabs are unavailable or if any headers passed are not present in [CORS-safelisted headers list](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header).

rajveermalviya:
(also updated the PR description to reflect current implementation)

rajveermalviya:
Thanks @stuartmorgan, updated the description & implementation to ignore them.

packages/url_launcher/url_launcher/example/lib/main.dart
```diff
@@ -63,6 +63,12 @@ class _MyHomePageState extends State<MyHomePage> {
   }
 
   Future<void> _launchInWebViewOrVC(Uri url) async {
+    if (!await launchUrl(url, mode: LaunchMode.inAppWebView)) {
+      throw Exception('Could not launch $url');
+    }
+  }
+
+  Future<void> _launchInWebViewOrVCWithCustomHeaders(Uri url) async {
```
stuartmorgan, line 71: `_launchAsInAppWebViewWithCustomHeaders` (The `WebViewOrVC` is a legacy of the old API surface.)

packages/url_launcher/url_launcher/example/lib/main.dart
```diff
@@ -194,7 +206,7 @@ class _MyHomePageState extends State<MyHomePage> {
               const Padding(padding: EdgeInsets.all(16.0)),
               ElevatedButton(
                 onPressed: () => setState(() {
-                  _launched = _launchInWebViewOrVC(toLaunch);
+                  _launched = _launchInWebViewOrVCWithCustomHeaders(toLaunch);
```
stuartmorgan, line 209: We can use the non-header version here.

packages/url_launcher/url_launcher_android/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 6.1.0
 
+* Adds support for Android Custom Tabs
```
stuartmorgan, line 3: This is missing the `.` at the end.

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
```diff
@@ -46,12 +48,12 @@ interface IntentResolver {
   }
 
   UrlLauncher(@NonNull Context context) {
-    this(
-        context,
+    this.applicationContext = context;
+    this.intentResolver =
```
stuartmorgan, line 52: Why are you changing this to remove the constructor delegation?

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
```diff
@@ -97,13 +99,21 @@ void setActivity(@Nullable Activity activity) {
     ensureActivity();
     assert activity != null;
 
+    Uri uri = Uri.parse(url);
```
stuartmorgan, line 102: This line can move into the `if` block.

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
```diff
@@ -97,13 +99,21 @@ void setActivity(@Nullable Activity activity) {
...
+    Bundle headersBundle = extractBundle(options.getHeaders());
+    if (!containsRestrictedHeader(options.getHeaders())) {
+      if (openCustomTab(activity, uri, headersBundle)) {
+        return true;
+      }
+    }
```
stuartmorgan, line 108: Now that there are two distinct paths through this method, please add comments before each of the two blocks explaining what's going on at a high level. E.g.,:
```java
// Try to launch using Custom Tabs if they have the necessary functionality.
...

// Fall back to a web view if necessary.
...
```

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
```diff
@@ -118,6 +128,38 @@ public void closeWebView() {
...
+      customTabsIntent.launchUrl(context, uri);
+    } catch (ActivityNotFoundException ex) {
+      return false;
+    }
+    return true;
+  }
+
+  // Checks if headers contains a CORS restricted header.
+  //  https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header
+  private static @NonNull Boolean containsRestrictedHeader(Map<String, String> headersMap) {
```
stuartmorgan, line 145: `@NonNull Boolean` -> `boolean`

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
```diff
@@ -118,6 +128,38 @@ public void closeWebView() {
     applicationContext.sendBroadcast(new Intent(WebViewActivity.ACTION_CLOSE));
   }
 
+  private static @NonNull Boolean openCustomTab(
```
stuartmorgan, line 131: `@NonNull Boolean` -> `boolean`

The methods above use the other pattern only because it's what Pigeon generates.

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
```diff
@@ -118,6 +128,38 @@ public void closeWebView() {
...
+    return true;
+  }
+
+  // Checks if headers contains a CORS restricted header.
+  //  https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header
+  private static @NonNull Boolean containsRestrictedHeader(Map<String, String> headersMap) {
+    for (String key : headersMap.keySet()) {
+      switch (key.toLowerCase(Locale.US)) {
+        case "accept":
+          continue;
```
stuartmorgan, line 149: Java supports multi-case blocks (just like Dart), so you don't need a separate statement for each supported header.

packages/url_launcher/url_launcher_android/android/src/test/java/io/flutter/plugins/urllauncher/UrlLauncherTest.java
```diff
@@ -157,19 +161,73 @@ public void openWebView_opensUrl() {
...
+    final ArgumentCaptor<Intent> intentCaptor = ArgumentCaptor.forClass(Intent.class);
+    verify(activity).startActivity(intentCaptor.capture(), isNull());
+    assertTrue(result);
+    assertEquals(Intent.ACTION_VIEW, intentCaptor.getValue().getAction());
+    assertNull(intentCaptor.getValue().getComponent());
+    final Bundle passedHeaders =
+        intentCaptor.getValue().getExtras().getBundle(Browser.EXTRA_HEADERS);
+    assertEquals(headers.get(headerKey), passedHeaders.getString(headerKey));
+  }
+
```
stuartmorgan, line 215: There needs to be a test for falling back to the webview if launching custom tabs is attempted but fails.

stuartmorgan (changes requested):
Structurally this looks great, just some comments inline on the details.

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
rajveermalviya: Can't do multiple cases for single block because of source version being Java 8.
At best can do the scary, not so scary:
```
  case "accept":
        case "accept-language":
        case "content-language":
        case "content-type":
          continue;
```

stuartmorgan: > At best can do the scary, not so scary:
> 
> ```
> case "accept":
> case "accept-language":
> case "content-language":
> case "content-type":
>   continue;
> ```

This is the construction I was referring to. I'm not sure why this would be scary; it's an extremely common pattern for switch statements in many languages.

rajveermalviya: I prefer to avoid silent fallthroughs, but yeah pretty common

stuartmorgan: Google's style guides ([including Java](https://google.github.io/styleguide/javaguide.html#s4.8.4-switch)) distinguish between stacked case labels (which are common, and don't have a history of being problematic) and what people generally think of as fallthrough, where there is both code and a fallthrough (and does have a history of being problematic, so is generally avoided).

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
rajveermalviya: reverted

rajveermalviya:
Thanks for the review @stuartmorgan, pushed a new revision with the suggested changes.

packages/url_launcher/url_launcher/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 6.1.14
+
+* Updates documentation mention support for Android Custom Tabs.
```
stuartmorgan, line 3: Nit: "to mention"

stuartmorgan (approved):
LGTM! (With one changelog nit.)

@gnprice or @gmackall can review for the second approval

packages/url_launcher/url_launcher/CHANGELOG.md
rajveermalviya: addressed

gmackall:
Thanks for the contribution, and for the fast iteration on reviews! I'll take a look this afternoon 🙂

gmackall:
Re running the checks, as it looks like the failures are due to infra

gmackall (commented):
This mostly LGTM, I just have one question I want to clarify before approving!

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/UrlLauncher.java
```diff
@@ -118,6 +131,35 @@ public void closeWebView() {
...
+      customTabsIntent.launchUrl(context, uri);
+    } catch (ActivityNotFoundException ex) {
+      return false;
+    }
+    return true;
+  }
+
+  // Checks if headers contains a CORS restricted header.
+  //  https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header
+  private static boolean containsRestrictedHeader(Map<String, String> headersMap) {
```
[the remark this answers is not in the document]
rajveermalviya: Digging into this a bit more, Chromium has a lot more exhaustive list of headers and checks for the values. Should the same be done here?
(Disclaimer; BSD licensed) - https://source.chromium.org/chromium/chromium/src/+/main:services/network/public/cpp/cors/cors.cc;l=278-425;drc=86339296c23927c8e8abb6cd81c8cb8d4bb3700e

Problem with not doing this correctly, that is check everything header+values, chrome & firefox will just ignore the restricted headers and the page will be opened in the custom tabs without those headers, it doesn't fail to open custom tab. So we can't fallback to webview if that happens, unless we do the check ourselves.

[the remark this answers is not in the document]
stuartmorgan: My preference would be to not add a bunch of complexity without knowing if anyone actually needs it. I would vote for keeping the current basic check, and see if we get feedback about it. Anyone affected can trivially work around it in the meantime by just adding an arbitrary other header to the request (e.g., add `'do-not-use-custom-tabs': true` to the header map).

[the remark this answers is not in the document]
gmackall: > Anyone affected can trivially work around it in the meantime by just adding an arbitrary other header to the request (e.g., add 'do-not-use-custom-tabs': true to the header map).

Good point, waiting and seeing if we get feedback SGTM with that in mind

gmackall (approved):

stuartmorgan:
This is now published. Thanks for the great contribution!

PR #4747, opened 2023-08-19
Replace ScaleTransition.scale with MatrixTransition.animation.

gmilou:
This is cleanup for flutter/flutter#130946 to deprecate the ScaleTransition.scale getter since it can be replaced with the parent's MatrixTransition.animation getter.


gmilou:
packages need to remain compatible with multiple stable releases, so PR is about a year premature.

PR #4753, opened 2023-08-22
[url_launcher] migrating objc plugin to swift

chrisdlangham:
This PR converts the iOS portion of the url_launcher plugin from objc to swift. 

Fixes https://github.com/flutter/flutter/issues/119102

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan (changes requested):
Thanks for the contribution! Could you split out a new PR that contains just the migration of the tests to Swift, so we can land that first in a separate PR? Rewriting tests and implementation in the same PR is much, much riskier than doing each one at a time.

hellohuanlin:
@chrisdlangham do we plan to rebase and update this PR?

chrisdlangham:
@hellohuanlin Yes this is ready now

packages/url_launcher/url_launcher_ios/ios/Classes/messages.g.swift
```diff
@@ -0,0 +1,136 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+// Autogenerated from Pigeon (v9.2.5), do not edit directly.
```
stuartmorgan, line 4: The PR should update to Pigeon 11; there were bug fixes for Swift in 10.x.

packages/url_launcher/url_launcher_ios/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 6.1.6
+
+* Migrates plugin from objc to swift
```
stuartmorgan, line 3: "Objective-C to Swift."

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,116 @@
...
+    registrar.publish(plugin)
+  }
+
+  private var currentSession: URLLaunchSession?
+  private let launcher: Launcher
+
+  init(launcher: Launcher = UIApplicationLauncher(), binaryMessenger: FlutterBinaryMessenger) {
+    self.launcher = launcher
+    super.init()
+    UrlLauncherApiSetup.setUp(binaryMessenger: binaryMessenger, api: self)
```
stuartmorgan, line 20: Why did you move this into `init` (making the test file more complicated)?

chrisdlangham: Yeah, it is not necessary. I moved that out

stuartmorgan:
Is this ready for another review?

chrisdlangham:
Yes this is ready

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
```diff
@@ -9,57 +9,42 @@ import XCTest
...
-      XCTAssertFalse(result?.boolValue ?? true)
-      XCTAssertNil(error)
-    } else {
-      XCTAssertNil(result)
-      XCTAssertNotNil(error)
-      XCTAssertEqual(error?.code, "argument_error")
-      XCTAssertEqual(error?.message, "Unable to parse URL")
-      XCTAssertEqual(error?.details as? String, "Provided URL: urls can't have spaces")
-    }
+    let result = createPlugin().canLaunchUrl(url: "urls can't have spaces")
```
hellohuanlin, line 34: why is this behavior changed?

hellohuanlin (commented):
I thought the test part is already in a separate PR?

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
stuartmorgan: This seems like a merge mistake; changing this would re-break the tests for Xcode 15.

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,115 @@
...
+  func openUrlInSafariViewController(
+    url: String, completion: @escaping (Result<Bool, Error>) -> Void
+  ) {
+    guard let url = URL(string: url) else {
+      completion(Result.failure(invalidURLError(for: url)))
+      return
+    }
+
+    currentSession = URLLaunchSession(url: url, completion: completion)
+    guard let session = currentSession else { return }
```
stuartmorgan, line 49: This should not be possible, and if it were it would cause a bug since this wouldn't be calling completion. Please make the line above assign to `session` and then do `currentSession = session` after, so that you don't need this check.

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,115 @@
...
+///
+/// It supports the following scenarios:
+///
+/// - The view controller is presenting another view.
+/// - The view controller is a UINavigationController.
+/// - The view controller is a UITabBarController.
+///
+/// @return The top most view controller.
+extension UIViewController {
+  var topViewController: UIViewController {
```
stuartmorgan, line 90: How does Swift handle name conflicts in extensions? This would be extremely dangerous in Obj-C code; I'm not sure if that's true in Swift as well.

Specifically:
- What would happen if Apple added a `UIViewController.topViewController` in a future update?
- What would happen if another plugin had a `UIViewController.topViewController` extension?

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,115 @@
...
+      return tabBarController.selectedViewController?.topViewController ?? tabBarController
+    }
+    if let presented = presentedViewController {
+      return presented.topViewController
+    }
+    return self
+  }
+}
+
+class GeneralError: Error {
```
stuartmorgan, line 105: The information in this error class will be discarded by Pigeon.

@tarrinneal Can you look into https://github.com/flutter/flutter/issues/123851#issuecomment-1500567081 and see whether we can in fact use the extension approach? If not, we'll need to add Pigeon functionality before we could convert these errors this way.

stuartmorgan (changes requested):

stuartmorgan:
> I thought the test part is already in a separate PR?

The error handling conventions are different in Obj-C vs Swift, so the generated Swift signatures aren't quite the same as the old signatures. Given that, tweaks to the tests are going to be required here unfortunately, but they should all be minor (not structural like the accidental revert of the invalid URL branching).

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
hellohuanlin: In this example, the definition in current module shadows definitions from other modules. So if Apple or another frameworks with the same definitions are both imported, the current module still takes precedence. 

However, if we were to make this extension `public` (rather than `internal`) and use it in other modules, it would be problematic when multiple modules with the same extension are imported, because there would be no way to indicate which one to use. In that case, we can use a prefix to disambiguate.

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
hellohuanlin: I don't think we support this kind of behavior in pigeon. But IIRC `FlutterError` should already be supported. 

I think I should update `FlutterError` in the engine so it conforms to `Error` automatically (e.g. to make it a subclass of `NSError`), but for now, you may have to write an empty extension here: 

```
extension FlutterError: Error {}
```

hellohuanlin (commented):
Not a complete review since I haven't looked closely at the ObjC code.

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
stuartmorgan: Thanks!

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
hellohuanlin: Filed an issue here: https://github.com/flutter/flutter/issues/136081

stuartmorgan: > but for now, you may have to write an empty extension here

The comment I linked to above is about validating whether we can do that without inter-plugin conflicts before we start shipping plugins that do it.

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
chrisdlangham: I think I see the issue. The pigeon protocol replaced the objective-c error argument with the swift `throws` keyword. But Swift will allow for implementation not to have the throws keyword. So I made this method throw an exception. Now the test have do/catch blocks. This is a bit closer to the original behavior.

chrisdlangham:
I think I have addressed everything so far. Thanks for the feedback.
@stuartmorgan @hellohuanlin

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,107 @@
...
+      return tabBarController.selectedViewController?.topViewController ?? tabBarController
+    }
+    if let presented = presentedViewController {
+      return presented.topViewController
+    }
+    return self
+  }
+}
+
+extension FlutterError: Error {}
```
hellohuanlin, line 107: Nice! 

After we make `FlutterError` a subclass of `NSError` in the engine, this will be a compile error complaining about redundant conformance. We will just need to delete this.

stuartmorgan: That's not a feature.

When that happens, how will `url_launcher` support both `master` and `stable`, as we require for our packages?

Packages don't work like the engine or the framework, in that they have to be compatible with more than exactly one version of all of the engine code, and they have to be designed differently as a result. If collisions are a compile error, then we shouldn't do this.

stuartmorgan: Relatedly, https://github.com/flutter/packages/pull/4753#discussion_r1349219420 still hasn't been addressed.

stuartmorgan:
Given the unresolved issues with `FlutterError`, and the fact that we only have a couple of errors which don't have any interesting native-side details, I think we should just change the Pigeon APIs here to enum returns and not use errors at all. We can then (temporarily, per https://github.com/flutter/flutter/issues/127665) convert them to equivalent `PlatformException`s on the Dart side to avoid any client-facing changes.

See https://github.com/flutter/packages/blob/9804d951b435896c031365b924624580bf239342/packages/local_auth/local_auth_ios/lib/local_auth_ios.dart#L44-L69 for an example of this pattern.

chrisdlangham:
@stuartmorgan I believe I have followed your suggestion on changing the pigeon API. It has changed a lot, so let me know what you think. I can revert it if I went in the wrong direction

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
stuartmorgan: Resolving since I checked, and it's safe to use the extension if we want to go that route.

packages/url_launcher/url_launcher_ios/pigeons/messages.dart
```diff
@@ -9,22 +9,55 @@ import 'package:pigeon/pigeon.dart';
...
+  failure,
+
+  /// The URL was not launched because it is not invalid URL
+  invalidUrl,
+
+  /// The URL did not load successfully in the SFSafariViewController.
+  failedToLoad,
+}
+
+class LaunchResultDetails {
```
stuartmorgan, line 28: I don't think we need this class; there isn't any information I see in the native error messages that comes from the native side, so we can just put those strings in the Dart code.

packages/url_launcher/url_launcher_ios/pigeons/messages.dart
```diff
@@ -9,22 +9,55 @@ import 'package:pigeon/pigeon.dart';
...
+  /// A system-provided error details, if any.
+  final String? errorDetails;
+}
+
 @HostApi()
 abstract class UrlLauncherApi {
   /// Returns true if the URL can definitely be launched.
   @ObjCSelector('canLaunchURL:')
-  bool canLaunchUrl(String url);
+  LaunchResultDetails canLaunchUrl(String url);
```
stuartmorgan, line 49: We should have two different enums, since the `failedToLoad` is specific to Safari view controller, so having it be an enum option here and in `launchUrl` is confusing.

packages/url_launcher/url_launcher_ios/lib/url_launcher_ios.dart
```diff
@@ -45,11 +47,39 @@ class UrlLauncherIOS extends UrlLauncherPlatform {
...
     }
   }
+
+  bool _mapLaunchResults({required final LaunchResultDetails results}) {
+    switch (results.result) {
+      case LaunchResult.success:
+        return true;
+      case LaunchResult.failure:
+        return false;
+      case LaunchResult.invalidUrl:
```
stuartmorgan, line 71: This would need explanatory comments with a link to https://github.com/flutter/flutter/issues/127665

stuartmorgan (changes requested):
I left some minor comments on the Pigeon API changes, but in general this matches what I was envisioning, thanks!

That said, given https://github.com/flutter/flutter/issues/123851#issuecomment-1758454011 if you wanted to revert that and keep the previous structure (but using `FlutterError` and the extension to make it work in place of the custom error class), I'm fine with that as well. I do think we'll want to move the plugin to this enum approach eventually regardless, but minimizing changes in this PR now that we have a viable option for doing so would be totally reasonable.

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
```diff
@@ -123,31 +120,34 @@ final class URLLauncherTests: XCTestCase {
...
+  func canOpenURL(_ url: URL) -> Bool {
+    url.scheme == "good"
   }
 
-  func open(
-    _ url: URL, options: [UIApplication.OpenExternalURLOptionsKey: Any] = [:],
-    completionHandler: ((Bool) -> Void)? = nil
+  func openURL(
+    _ url: URL, options: [UIApplication.OpenExternalURLOptionsKey: Any],
+    completionHandler completion: ((Bool) -> Void)?
```
hellohuanlin, line 148: curious if this line wrap is done by swift-formatter? it's odd to layout 2 params in 1 line and another in a separate line. swift developers typically write: 

```
func openURL(
  _ url: URL,
  options: [UIAppli..., 
  completionHandler...
)
```

packages/url_launcher/url_launcher_ios/ios/Classes/Launcher.swift
```diff
@@ -0,0 +1,27 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Protocol for UIApplication methods relating to launching URLs.
+///
+/// This protocol exists to allow injecting an alternate implementation for testing.
+protocol Launcher {
+  func canOpenURL(_ url: URL) -> Bool
+  func openURL(
```
hellohuanlin, line 10: wanna document these functions

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,109 @@
...
+    currentSession?.close()
+  }
+
+  /**
+    * Creates an error for an invalid URL string.
+    *
+    * @param url The invalid URL string
+    * @return The error to return
+    */
+  func invalidURLError(for url: String) -> LaunchResultDetails {
```
hellohuanlin, line 78: should this be private?

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,109 @@
...
+  public static func register(with registrar: FlutterPluginRegistrar) {
+    let plugin = URLLauncherPlugin()
+    UrlLauncherApiSetup.setUp(binaryMessenger: registrar.messenger(), api: plugin)
+    registrar.publish(plugin)
+  }
+
+  private var currentSession: URLLaunchSession?
+  private let launcher: Launcher
+
+  var topViewController: UIViewController? {
```
hellohuanlin, line 18: should this be private?

hellohuanlin: hmmm, what's the diff between this func, and the one in UIViewController extension below?

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
hellohuanlin: Just looked up [our style guide](https://google.github.io/swift/#line-wrapping): 

> Comma-delimited lists are only laid out in one direction: horizontally or vertically. In other words, all elements must fit on the same line, or each element must be on its own line. 

Can you update the format for this function and a few others as well?

chrisdlangham:
I think I have addressed everything so far. Thanks for the feedback.
@stuartmorgan @hellohuanlin

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
```diff
@@ -81,22 +68,15 @@ final class URLLauncherTests: XCTestCase {
...
-        XCTAssertNotNil(result)
-        XCTAssertFalse(result?.boolValue ?? true)
-        XCTAssertNil(error)
-      } else {
-        XCTAssertNil(result)
-        XCTAssertNotNil(error)
-        XCTAssertEqual(error?.code, "argument_error")
-        XCTAssertEqual(error?.message, "Unable to parse URL")
-        XCTAssertEqual(error?.details as? String, "Provided URL: urls can't have spaces")
+        XCTAssertTrue(details == .failedToLoad || details == .invalidUrl)
```
hellohuanlin, line 76: which case will it actually be?

packages/url_launcher/url_launcher_ios/ios/Classes/Launcher.swift
```diff
@@ -0,0 +1,32 @@
...
+
+  /// Attempts to asynchronously open the resource at the specified URL.
+  func openURL(
+    _ url: URL,
+    options: [UIApplication.OpenExternalURLOptionsKey: Any],
+    completionHandler completion: ((Bool) -> Void)?)
+}
+
+/// Default implementation of Launcher, using UIApplication.
+final class UIApplicationLauncher: Launcher {
```
hellohuanlin, line 20: can this simply be an empty extension? 
```
extension UIApplication: Launcher {}
```

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,96 @@
...
+  func canLaunchUrl(url: String) -> LaunchResult {
+    guard let url = URL(string: url) else {
+      return .failedToLoad
+    }
+    let canOpen = launcher.canOpenURL(url)
+    return canOpen ? .success : .failedToLoad
+  }
+
+  func launchUrl(
+    url: String, universalLinksOnly: Bool,
```
hellohuanlin, line 37: line break

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,96 @@
...
+      return
+    }
+    let options = [UIApplication.OpenExternalURLOptionsKey.universalLinksOnly: universalLinksOnly]
+    launcher.openURL(url, options: options) { result in
+      completion(.success(result ? .success : .failedToLoad))
+    }
+  }
+
+  func openUrlInSafariViewController(
+    url: String, completion: @escaping (Result<LaunchResult, Error>) -> Void
```
hellohuanlin, line 51: line break. please go over all signatures and make sure comma separated items are in separate lines.

packages/url_launcher/url_launcher_ios/ios/Classes/URLLauncherPlugin.swift
```diff
@@ -0,0 +1,96 @@
...
+    let session = URLLaunchSession(url: url, completion: completion)
+    currentSession = session
+
+    session.didFinish = { [weak self] in
+      self?.currentSession = nil
+    }
+    topViewController?.present(session.safariViewController, animated: true, completion: nil)
+  }
+
+  func closeSafariViewController() throws {
```
hellohuanlin, line 67: why is this throwing?

stuartmorgan (dismissed):

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
chrisdlangham: It depends on the iOS version the tests are being run on. I think @stuartmorgan can explain it better, but on iOS 17 we should get .failedToLoad, and on older versions, we should get .invalidUrl

hellohuanlin: Can you do "if iOS 17 else".

chrisdlangham:
I've noticed quite a few merge conflicts. I've been contemplating whether to continue this, and this has helped me make a decision. Unfortunately, I'm now facing a shortage of both time and energy for this. Please feel free to either close this or relocate it as needed.

stuartmorgan:
@chrisdlangham Thanks for letting us know. This is basically done, and I definitely don't want to lose all the great work you've done on this, so I'll resolve the conflicts with my recent PR, and do any last adjustments to get it landed. We really appreciate the contribution!

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
stuartmorgan: I realized there's a better option; I've added a utility method in the test file that checks the NSURL behavior directly, and then switches based on that in the tests.

stuartmorgan:
I think I've addressed the previous comments, so this should be ready for another review @hellohuanlin

hellohuanlin (approved):
LGTM!

stuartmorgan (approved):

PR #4757, opened 2023-08-22
[tool] Add Android dependency (gradle) option to update dependencies command

camsim99:
Adds an `android-dependency` option to the `update-dependency` command such that you can update Android dependencies provided the dependency and a version across relevant plugins. This PR specifically adds support for the Gradle dependency, relevant to plugin example apps.

Running the command looks like:
```
dart run script/tool/bin/flutter_plugin_tools.dart update-dependency --android-dependency gradle --version 1.2.3
```

## Pre-launch Checklist

- [x] I signed the [CLA].


script/tool/lib/src/update_dependency_command.dart
```diff
@@ -102,6 +117,32 @@ ${response.httpResponse.body}
...
+    _targetAndroidDependency = getNullableStringArg(_androidDependency);
+    if (_targetAndroidDependency != null) {
+      final String? version = getNullableStringArg(_versionFlag);
+      if (version == null) {
+        printError('A version must be provided to update this dependency.');
+        throw ToolExit(_exitNoTargetVersion);
+      } else if (_targetAndroidDependency == 'gradle') {
+        final RegExp validGradleVersionPattern = RegExp(r'(\d(\.\d){1,2})');
+        final bool isValidGradleVersion =
+            validGradleVersionPattern.stringMatch(version) == version;
```
camsim99, line 134: I am not well versed in regular expressions, so please let me know if there's a better way to validate the Gradle version to be either `#.#` or `#.#.#`.

script/tool/lib/src/update_dependency_command.dart
```diff
@@ -181,6 +226,58 @@ ${response.httpResponse.body}
...
+                .childFile('gradle-wrapper.properties');
+
+        final String gradleWrapperPropertiesContents =
+            gradleWrapperPropertiesFile.readAsStringSync();
+        final RegExp validGradleDistributionUrl =
+            RegExp(r'^\s*distributionUrl\s*=\s*.*\.zip', multiLine: true);
+        if (!validGradleDistributionUrl
+            .hasMatch(gradleWrapperPropertiesContents)) {
+          return PackageResult.fail(<String>[
+            'Failed to update Gradle version in the example app for ${package.displayName}'
```
stuartmorgan, line 262: It would be good to explain why. E.g., 'Unable to find a "distributionUrl" entry to update."

script/tool/lib/src/update_dependency_command.dart
```diff
@@ -181,6 +226,58 @@ ${response.httpResponse.body}
...
+          ]);
+        }
+
+        final String newGradleWrapperPropertiesContents =
+            gradleWrapperPropertiesContents.replaceFirst(
+                validGradleDistributionUrl,
+                // ignore: unnecessary_string_escapes
```
stuartmorgan, line 269: What you have below is the Dart string `...https://...` with the `:` unnecessarily escaped; the warning you are ignoring is correct. What you actually want to write is `https\\://` which creates the string literal `https\://` (and won't cause the warning).

If you run this locally you should see that it's generating the wrong output.

script/tool/lib/src/update_dependency_command.dart
```diff
@@ -181,6 +226,58 @@ ${response.httpResponse.body}
...
+                'distributionUrl=https\://services.gradle.org/distributions/gradle-$_targetVersion-all.zip');
+        gradleWrapperPropertiesFile
+            .writeAsStringSync(newGradleWrapperPropertiesContents);
```
stuartmorgan, line 272: We try to be relatively verbose in this tooling so it's easy for people running it to see what the commands are actually doing; we should print something for each entry we change (along the lines of line 201 above).

stuartmorgan, line 272: Can we add a TODO here with an issue link for validating that the AGP version is compatible with the new Gradle version?

script/tool/test/update_dependency_command_test.dart
```diff
@@ -584,4 +609,197 @@ dev_dependencies:
...
+        package.displayName,
+        '--android-dependency',
+        'gradle',
+        '--version',
+        newGradleVersion,
+      ]);
+
+      final String updatedGradleWrapperPropertiesContents =
+          gradleWrapperPropertiesFile.readAsStringSync();
+      expect(updatedGradleWrapperPropertiesContents.contains(newGradleVersion),
```
stuartmorgan, line 762: This (and the next test) should look for the whole line, to catch the bug with the `\:`. You can make it extra robust by not using double-escaping, but instead do string concatenation with part of it as a raw string, so that the code generating the string and the code matching the output can't just copy/paste the same bug:

```
...updatedGradleWrapperPropertiesContents, contains(r'distributionUrl=https\://services.gradle.org/distributions/'
    'gradle-$newGradleVersion-all.zip`);

Comment (stuartmorgan) [reply script/tool/test/update_dependency_command_test.dart:762]: (There's a matcher for `contains`, so you can use that directly in `expect` rather than the String `contains` + `isTrue`.)

File: script/tool/test/update_dependency_command_test.dart [hunk 6]
```diff
@@ -584,4 +609,197 @@ dev_dependencies:
...
+          isTrue);
```
stuartmorgan, line 763: These should also expect the status output requested in my comment above.

stuartmorgan (changes requested):
Nice!

script/tool/lib/src/update_dependency_command.dart
```diff
@@ -181,6 +229,65 @@ ${response.httpResponse.body}
...
+        final RegExp validGradleDistributionUrl =
+            RegExp(r'^\s*distributionUrl\s*=\s*.*\.zip', multiLine: true);
+        if (!validGradleDistributionUrl
+            .hasMatch(gradleWrapperPropertiesContents)) {
+          return PackageResult.fail(<String>[
+            'Unable to find a "distributionUrl" entry to update for ${package.displayName}.'
+          ]);
+        }
+
+        print('${indentation}Updating to "$_targetVersion"');
```
stuartmorgan, line 272: Since this is in a loop, we should say which example:
`print('${indentation}Updating ${getRelativePosixPath(example.directory,
          from: package.directory)} to "$_targetVersion"');`

script/tool/lib/src/update_dependency_command.dart
```diff
@@ -181,6 +229,65 @@ ${response.httpResponse.body}
...
+                validGradleDistributionUrl,
+                'distributionUrl=https\\://services.gradle.org/distributions/gradle-$_targetVersion-all.zip');
+        // TODO(camsim99): Validate current AGP version against target Gradle
+        // version: https://github.com/flutter/flutter/issues/133887.
+        gradleWrapperPropertiesFile
+            .writeAsStringSync(newGradleWrapperPropertiesContents);
+      }
+      return updateRanForExamples
+          ? PackageResult.success()
+          : PackageResult.skip('Example app does not run on Android.');
```
stuartmorgan, line 284: 'No example apps run on Android.'

script/tool/lib/src/update_dependency_command.dart
```diff
@@ -181,6 +229,65 @@ ${response.httpResponse.body}
     return PackageResult.success();
   }
 
+  /// Handles all of the updates for [package] when the target dependency is
+  /// an Android dependency.
+  Future<PackageResult> _runForAndroidDependency(
+      RepositoryPackage package) async {
+    if (_targetAndroidDependency == 'gradle') {
+      print('${indentation}Updating examples running on Android...');
```
stuartmorgan, line 237: Optional: I would remove this print since we have specific details below; I think this will end up being noise in the output.

stuartmorgan (approved):
LGTM with minor comments

PR #4768, opened 2023-08-24
[google_maps_flutter_android] Fix for testToggleInfoWindow persistently flaky 

mossmana:
Fixes issue: https://github.com/flutter/flutter/issues/131783

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
@@ -938,7 +938,7 @@ void googleMapsTests() {
 
     await tester.pumpAndSettle();
 
-    // TODO (mossmana): Adding this delay addresses 
+    // TODO(mossmana): Adding this delay addresses 
     // https://github.com/flutter/flutter/issues/131783. It may be related
     // to https://github.com/flutter/flutter/issues/54758 and should be
     // re-evaluated when that issue is fixed.
--- a/packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
@@ -938,7 +938,7 @@ void googleMapsTests() {
 
     await tester.pumpAndSettle();
 
-    // TODO(mossmana): Adding this delay addresses 
+    // TODO(mossmana): Adding this delay addresses
     // https://github.com/flutter/flutter/issues/131783. It may be related
     // to https://github.com/flutter/flutter/issues/54758 and should be
     // re-evaluated when that issue is fixed.
--- a/packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
@@ -936,26 +936,26 @@ void googleMapsTests() {
     final ExampleGoogleMapController controller =
         await controllerCompleter.future;
 
+    await tester.pumpAndSettle();
+
+    // TODO (mossmana): Adding this delay addresses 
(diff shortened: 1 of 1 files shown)


mossmana:
I have a request out to @Hixie to get a test exemption for this PR.

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -936,26 +936,26 @@ void googleMapsTests() {
     final ExampleGoogleMapController controller =
         await controllerCompleter.future;
 
+    await tester.pumpAndSettle();
+
+    // TODO(mossmana): Adding this delay addresses
+    // https://github.com/flutter/flutter/issues/131783. It may be related
+    // to https://github.com/flutter/flutter/issues/54758 and should be
+    // re-evaluated when that issue is fixed.
+    await Future<void>.delayed(const Duration(seconds: 1));
```
Hixie, line 945: this violates our style guide: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#never-check-if-a-port-is-available-before-using-it-never-add-timeouts-and-other-race-conditions

Hixie:
test-exempt: is a test

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
mossmana: @Hixie I entirely agree, which is why I posted the following comment https://github.com/flutter/flutter/issues/54758#issuecomment-1684226391. It is unfortunately an existing pattern in this particular test and I would like to fix it the proper way, but I believe there's additional prerequisite work to be done first.

stuartmorgan:
Good find, we should rename this file. I just checked the repo, and it looks like it's the only one that made this mistake, so we can just add renaming it to this PR.

mossmana:
> Good find, we should rename this file. I just checked the repo, and it looks like it's the only one that made this mistake, so we can just add renaming it to this PR.

I'll try my best, but not 100% certain I can preserve the git history of this file (since it's a delete/add). Would that be a problem?

stuartmorgan:
> > Good find, we should rename this file. I just checked the repo, and it looks like it's the only one that made this mistake, so we can just add renaming it to this PR.
> 
> I'll try my best, but not 100% certain I can preserve the git history of this file (since it's a delete/add). Would that be a problem?

If it doesn't preserve it that's fine, but I would be extremely surprised if it doesn't detect this as a rename given the small scope of the changes relative to the overall file. (git doesn't care how you actually do the move; move detection is heuristic in git.)

stuartmorgan:
> In this particular case, I think `google_maps_tests ` was imported into `legacy_renderer_test` and `latest_renderer_test` tests and is not able to be run standalone.

Oops, I missed that. Sorry for the waste of time there :(

So the bot was right to flag it (since it wouldn't automatically be run), and doing a manual exception was the right solution after all.

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
mossmana: I tried the suggestion and `onCameraIdle` is never called.

mossmana: @stuartmorgan I documented what I tried in https://docs.google.com/document/d/1BPQ4OdlNkNEq5xUt8cp72F4qrZrOVTja4vIxqIvq3X4/edit?usp=sharing. Any other suggestions?

stuartmorgan (approved):
LGTM with one comment.

The delay is definitely not ideal, but this seems better than the status quo of having no coverage, and we have a TODO for improving it once the necessary feature exists to do so.

PR #4809, opened 2023-08-29
[local_auth] Update to pigeon 11 and remove enum wrappers

tarrinneal:
updates pigeon version to 11 to remove need for enum wrappers - removes enum wrappers.

Some work toward https://github.com/flutter/flutter/issues/133455


packages/local_auth/local_auth_android/test/local_auth_test.dart
```diff
@@ -79,7 +79,7 @@ void main() {
       ]);
     });
 
-    test('handles emtpy', () async {
```
stuartmorgan, line 82: Are we sure we don't need test coverage of "emtpy" handling?

stuartmorgan (approved):
LGTM! It's great to see all that manual boilerplate go away.

packages/local_auth/local_auth_android/test/local_auth_test.dart
tarrinneal: I think it's a risk we're going to have to take unfortunately. If we see a flare up of emtpy in the future we can revisit this at that time.

PR #4820, opened 2023-08-30
[Android] Introduces API 34 emulators for subset of tests

camsim99:
Introduces API 34 emulators for subset of tests not causing test failures in https://github.com/flutter/packages/pull/4804. Also changes `linux_android` version to 34v1.

## Pre-launch Checklist

- [x] I signed the [CLA].


reidbaker:
I know I asked for you to do this but today when talking to @stuartmorgan he said the shards are not stable we cant actually update one at a time.

stuartmorgan:
Do we want to re-open this to use https://github.com/flutter/packages/pull/4904 ?

If so, I'm happy to help with adjusting the scripts to partition the tests.

.ci.yaml
```diff
@@ -306,7 +306,7 @@ targets:
     recipe: packages/packages
     timeout: 60
     properties:
-      target_file: android_platform_tests.yaml
+      target_file: android_platform_tests_api_33.yaml
```
reidbaker, line 309: Can you add a comment that says all the shards have to have the exact same dependency list since they are not actually independent?

.ci/targets/android_platform_tests_api_33.yaml
```diff
@@ -0,0 +1,29 @@
...
+  # these steps.
+  - name: native unit tests
+    script: script/tool_runner.sh
+    args: ["native-test", "--android", "--no-integration", "--exclude=script/configs/exclude_native_unit_android.yaml"]
+  - name: native integration tests
+    script: script/tool_runner.sh
+    args: ["native-test", "--android", "--no-unit"]
+  - name: drive examples
+    script: script/tool_runner.sh
+    args: ["drive-examples", "--android", "--exclude=script/configs/exclude_integration_android.yaml,script/configs/exclude_integration_android_emulator.yaml", "--filter-packages-to=script/configs/still_requires_api_33_avd.yaml"]
```
reidbaker, line 29: Do you need this exclude if you are using the filter-packages-to flag?

reidbaker (approved):

packages/quick_actions/quick_actions_android/example/android/app/src/androidTest/java/io/flutter/plugins/quickactionsexample/QuickActionsTest.java
```diff
@@ -99,7 +99,7 @@ public void appShortcutLaunchActivityAfterStarting() {
             .filter(s -> s.getId().equals(firstShortcut.getId()))
             .findFirst()
             .get();
-    Intent dynamicShortcutIntent = dynamicShortcut.getIntent();
+    Intent dynamicShortcutIntent = dynamicShortcut.getIntent().addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
reidbaker, line 102: This line is too long and needs to be formatted.

reidbaker (approved):

packages/quick_actions/quick_actions_android/example/android/app/src/androidTest/java/io/flutter/plugins/quickactionsexample/QuickActionsTest.java
camsim99: Removed this fix attempt and excluded `quick_actions` for now since the test is still flaky.

camsim99:
@stuartmorgan Could you take a look at the way I attempted to filter the failing tests out? It appears to be working minus one build that keeps getting canceled that is not one I filtered out. For my own notes, here's an example build of that: https://ci.chromium.org/ui/p/flutter/builders/try/Linux_android%20android_platform_tests_shard_1%20master/1905/overview

.ci.yaml
```diff
@@ -302,11 +302,13 @@ targets:
...
+  # dependency list, despite some running on Android 33 AVDs versus 34.
   - name: Linux_android android_platform_tests_shard_1 master
     recipe: packages/packages
     timeout: 60
     properties:
-      target_file: android_platform_tests.yaml
+      target_file: android_platform_tests_api_33.yaml
       channel: master
       version_file: flutter_master.version
       package_sharding: "--shardIndex 0 --shardCount 6"
```
stuartmorgan, line 314: You'll need to changed the shard flags (the names of the tasks will be slightly confusing during the transition, but changing names is a huge pain with LUCI), so that each of the two partitions has their own shards, otherwise we may end up with a plugin assigned to a specific shard, but then excluded in that shard (since sharding is pre-exclusion, for slightly complicated reasons).

So I believe the setup we need is something like:
```
- name: ..._shard_1
  // set up for 34
  package_sharding: "--shardIndex 0 --shardCount 4"

- name: ..._shard_2
  // set up for 34
  package_sharding: "--shardIndex 1 --shardCount 4"

- name: ..._shard_3
  // set up for 34
  package_sharding: "--shardIndex 2 --shardCount 4"

- name: ..._shard_4
  // set up for 34
  package_sharding: "--shardIndex 3 --shardCount 4"

- name: ..._shard_5
  // set up for 33
  package_sharding: "--shardIndex 0 --shardCount 1"

- name: ..._shard_6
  // set up for 33
  package_sharding: "--shardIndex 1 --shardCount 2"
```

stuartmorgan: The other option we could consider if we think we'll be in this state for a while is to bring up a new task or two for 33 tests, and leave the existing 6 shards as-is but with 34, so the names make more sense.

.ci/targets/android_platform_tests_api_33.yaml
```diff
@@ -0,0 +1,29 @@
...
+  # these steps.
+  - name: native unit tests
+    script: script/tool_runner.sh
+    args: ["native-test", "--android", "--no-integration", "--exclude=script/configs/exclude_native_unit_android.yaml"]
+  - name: native integration tests
+    script: script/tool_runner.sh
+    args: ["native-test", "--android", "--no-unit", "--filter-packages-to=script/configs/still_requires_api_33_avd_integration.yaml"]
+  - name: drive examples
+    script: script/tool_runner.sh
+    args: ["drive-examples", "--android", "--filter-packages-to=script/configs/still_requires_api_33_avd_drive_examples.yaml"]
```
stuartmorgan, line 29: I would suggest having just one config file covering everything that needs an exclusion for any reason, and then passing that config file to *every* task. That way there's a clear and consistent separation between 33 and 34. Without that, it'll be tricky to avoid double-running some tests.

.ci.yaml
camsim99: I'm going to be actively working on getting the tests failing on 34 working, but two of the failures have been really tricky, so could be a matter of months worst case. I'll start changing the shard flags but let me know if you think it makes more sense to add new tasks instead given that projection.

stuartmorgan: I think on a potential timeline of months it's probably better to bring up 1-2 new shards for 33.

camsim99: Ok hoping to be faster but I think that's probably safest.

.ci.yaml
```diff
@@ -374,7 +381,38 @@ targets:
...
       channel: master
       version_file: flutter_master.version
+      # set up for 34
       package_sharding: "--shardIndex 5 --shardCount 6"
+      dependencies: >-
+        [
+          {"dependency": "android_virtual_device", "version": "34"}
+        ]
+
+  - name: Linux_android android_platform_tests_api_33_shard_1 master
```
stuartmorgan, line 391: The addition of these new tasks should be done in a new prequel PR; if we land it this way, we'll lose test coverage on some critical plugins until things propagate and you can remove the `bringup` flag. We can instead do a PR that just adds the new tasks, and then once they propagate, this full PR will just take them out of bringup so we are moving tests between active tasks rather than moving them into tasks that won't show up in most of our CI.

.ci/targets/android_platform_tests_api_33.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - name: download Dart and Android deps
+    script: script/tool_runner.sh
+    infra_step: true
+    args: ["fetch-deps", "--android", "--supporting-target-platforms-only"]
+  - name: build examples
+    script: script/tool_runner.sh
+    args: ["build-examples", "--apk"]
+  - name: lint
+    script: script/tool_runner.sh
+    args: ["lint-android"]
```
stuartmorgan, line 16: Since these tasks aren't being passed either of the include/exclude flags, they will all run twice: once on the 34 shards and once on the 33 shards, even though there's no difference.

We should either pass the flags to everything, to keep all the tasks for a given plugin together, and running only in one place, or we should remove these two tasks from the 33 version, so all builds and lints run on the 34 shards.

I would vote for the former, for consistency, but either should work.

.ci.yaml
```diff
@@ -461,7 +499,7 @@ targets:
       package_sharding: "--shardIndex 5 --shardCount 6"
       dependencies: >-
         [
-          {"dependency": "android_virtual_device", "version": "33"}
+          {"dependency": "android_virtual_device", "version": "34"}
         ]
```
stuartmorgan, line 504: You'll need to add matching `stable` shards, or we'll lose coverage on stable since they are running the same scripts. That can be done in the same prequel PR.

stuartmorgan (changes requested):

camsim99:
This PR is quickly becoming outdated and I have been doing some experimental changes to fix the path provider flake (such as https://ci.chromium.org/ui/p/flutter/builders/try/Linux_android%20android_platform_tests_shard_4%20master/2375/overview). Will fix path provider (https://github.com/flutter/packages/pull/5098) and then open a PR for landing this change.

PR #4822, opened 2023-08-31
[ci] Convert version presubmit check to LUCI

stuartmorgan:
Adds a LUCI version of the version presubmit check, using the new label passthrough functionality, and removes it from Cirrus. Since that was the last Cirrus task, this removes the entire Cirrus config, and the associated Dockerfile.

Part of https://github.com/flutter/flutter/issues/130076


stuartmorgan:
I've validated that the labels are working as expected in this configuration.

yusuf-goog (approved):

PR #4826, opened 2023-08-31
[tools,pigeon] Update tooling to handle Windows build output changes

stuartmorgan:
Updates the tooling that builds and runs Windows unit tests to handle the build output path changes in https://github.com/flutter/flutter/pull/131843


stuartmorgan:
This is now ready for review. I tested it with a roll to make sure it would work (which is how I found the Pigeon part, which I'd forgotten would be an issue), but since it also needs to support the `stable` behavior I've removed the roll, and it can land without a roll so that tomorrow's roll won't fail.

packages/pigeon/tool/shared/test_suites.dart
```diff
@@ -313,9 +314,23 @@ Future<int> _runWindowsUnitTests() async {
...
+  // Depending on the Flutter version, the build output path is different.
+  // Check for the new location first, and fall back to the old location, to
+  // support running against both stable and master.
+  // TODO(stuartmorgan): Remove all this when these tests no longer need to
+  // support Flutter 3.13, and just construct the version of the path with the
+  // architecture.
+  const String buildDirBase = '$examplePath/build/windows';
+  const String buildRelativeBinaryPath =
+      'plugins/test_plugin/Debug/test_plugin_test.exe';
+  final String arch = Abi.current() == Abi.windowsArm64 ? 'arm64' : 'x64';
```
loic-sharma, line 326: FYI, the tool does not support targeting Windows Arm64 yet. Currently, building on Windows Arm64 will create an x64 executable that will run using emulation. Here's the issue that tracks creating Windows Arm64 executables: https://github.com/flutter/flutter/issues/129807

stuartmorgan: Thanks, I misremembered what the roadmap ordering was on the arm64 support.

To avoid having a mini fire-drill when https://github.com/flutter/flutter/issues/129807 lands, I've made the scripts messier for now but future-proof, so that they will work with the new structure with only x64, but then use the new structure with arm64 when it starts working.

tarrinneal:
> which is how I found the Pigeon part

Are there greater implications to pigeon than just the testing? Or is pigeon too high level for it to matter?

loic-sharma (approved):
LGTM

tarrinneal (approved):
SF, there are a few todo's in here that don't actually reference an issue. Not sure if they need to or not though...

stuartmorgan:
All that changes is where build products go; Pigeon is only affected because it has hand-rolled test scripts, so is sensitive to changes in where the test binaries end up.

PR #4836, opened 2023-09-02
[image_picker_android] check if data from result is empty when picking a single img or video

aldee:
This PR solves the issue when the image_picker plugin is being used on Android 10 devices or below, where when we use the Android photo picker, it crashes due to invalid Uri data being fetched from the Intent.

This issue is reproducible on my company's app when trying to pick image from the Android photo picker on Android 10 devices and Android 10 emulators with Google Play

https://github.com/flutter/flutter/issues/104168

## Pre-launch Checklist

- [ x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [ x] I signed the [CLA].
- [ x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ x] I listed at least one issue that this PR fixes in the description above.
- [ x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x ] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [ x] I updated/added relevant documentation (doc comments with `///`).
- [ x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ x] All existing and new tests are passing.


stuartmorgan:
Thanks for the contribution! You’ve checked boxes in the PR checklist above that are not reflected in this PR, so I’m assuming this is a work in progress and am marking it as a Draft. Please review the checklist, updating the PR as appropriate, and when the checklist reflects the state of the PR as posted please feel free to mark it as ready for review.

aldee:
Hi, thanks for the feedback, I didn't realize all of them were relevant to my PR. But I will work on it now

stuartmorgan:
> I didn't realize all of them were relevant to my PR.

If there's something about the template that suggests that the checklist wouldn't apply to some PRs, please let us know and we can adjust the wording.

aldee:
@stuartmorgan Hi, it was my fault for not looking at the checklist thoroughly. But I do have a question now, since I think I've followed all the checklist, some automated checks were not successful, and I don't understand why. Can you give a guidance on that?

Thanks!

aldee:
never mind, I just found out that the problem is code formatting, but I had to look at several pages to see the actual error. Looking at the Details page of the check only tells me there an error with formatting, but not telling me where.

gmackall:
Hey @aldee, thanks for the contribution! This PR will still need tests before it is ready for review, though (per the comment from the flutter-dashboard bot). 

Let me know once you've had a chance to add tests to the PR and I can get started reviewing!

packages/image_picker/image_picker_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.8.7+6
+
+* Fix NullPointerException on Android 10 devices when using Android Photo Picker
```
stuartmorgan, line 3: Please follow the CHANGELOG style guide linked in the checklist.

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -630,7 +630,14 @@ public boolean onActivityResult(
 
   private void handleChooseImageResult(int resultCode, Intent data) {
     if (resultCode == Activity.RESULT_OK && data != null) {
-      String path = fileUtils.getPathFromUri(activity, data.getData());
+      String path;
+      // On Android 10 devices using Android Photo Picker, the Uri from getData() could be null.
+      if (data.getData() != null) {
+        path = fileUtils.getPathFromUri(activity, data.getData());
```
stuartmorgan, line 636: This is unnecessarily duplicating `fileUtils.getPathFromUri`; only getting the URI needs to be branched.

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -630,7 +630,14 @@ public boolean onActivityResult(
...
+      } else {
+        Uri uri = data.getClipData().getItemAt(0).getUri();
```
stuartmorgan, line 638: This isn't null-checking `getClipData` before using it.

packages/image_picker/image_picker_android/android/src/test/java/io/flutter/plugins/imagepicker/ImagePickerDelegateTest.java
```diff
@@ -129,6 +130,12 @@ public void setUp() throws IOException {
 
     Uri mockUri = mock(Uri.class);
     when(mockIntent.getData()).thenReturn(mockUri);
+
+    ClipData mockClipData = mock(ClipData.class);
+    ClipData.Item mockItem = mock(ClipData.Item.class);
+    when(mockItem.getUri()).thenReturn(mockUri);
+    when(mockClipData.getItemAt(0)).thenReturn(mockItem);
+    when(mockIntent.getClipData()).thenReturn(mockClipData);
```
stuartmorgan, line 138: Why are you changing the mock for every test instead of just the new test?

stuartmorgan (changes requested):

packages/image_picker/image_picker_android/CHANGELOG.md
aldee: I have changed Fix to Fixes, and have added . at the end of the sentence

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
aldee: Have changed this to check for null getClipData and check if the ClipData contains 1 item, as this should only be called when the client uses ImagePicker().pickImage() which only retrieve 1 image

aldee:
Hi @stuartmorgan I've updated the PR as per your feedback. Let me know if you have more from my update. Thanks!

aldee:
Hi @gmackall, I've updated the PR as per the feedback shared by @stuartmorgan , can you please review the MR and approve it if it's all good? thanks!

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -630,7 +630,18 @@ public boolean onActivityResult(
 
   private void handleChooseImageResult(int resultCode, Intent data) {
     if (resultCode == Activity.RESULT_OK && data != null) {
-      String path = fileUtils.getPathFromUri(activity, data.getData());
+      Uri uri;
+      // On Android 10 devices using Android Photo Picker, the Uri from getData() could be null.
+      if (data.getData() != null) {
+        uri = data.getData();
+      } else if (data.getClipData() != null && data.getClipData().getItemCount() == 1) {
```
gmackall, line 637: The documentation for ClipData.Item `getUri` isn't perfectly clear, but the documentation for other methods (ex [here](https://developer.android.com/reference/android/content/ClipData.Item#getUri())) seems to indicate it can return null. Can you add logic to the check here for that case?

packages/image_picker/image_picker_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.8.7+6
+
+* Fixes NullPointerException on Android 10 devices when using Android Photo Picker.
```
gmackall, line 3: This note and another comment later indicate that this PR is targeted at fixing this issue for Android 10. On the other hand the linked issue (https://github.com/flutter/flutter/issues/104168) seems to occur on a range of versions (with the initial report indicating Android 12). 

If you are linking it as fixed (and therefore saying that this fix isn't specific to Android 10) then the CHANGELOG and comments here should probably not indicate that, correct?

gmackall (changes requested):
Thanks for the contribution! I left a couple of comments.

With your reproduction, would you also be willing to see if the same issue occurs for video picking? The code makes the same `data.getData()` call, and it would be a good time to make the same fix if the same problem exists.

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
aldee: Good point, I'm adding it now

packages/image_picker/image_picker_android/CHANGELOG.md
aldee: Yes the issue does seem to happen in several pre-Android 13 devices, but in my app's case, most of the crashes are from Android 10 devices. but will change the changelog note in the next commit

aldee:
> Thanks for the contribution! I left a couple of comments.
> 
> With your reproduction, would you also be willing to see if the same issue occurs for video picking? The code makes the same `data.getData()` call, and it would be a good time to make the same fix if the same problem exists.

I'm unable to reproduce this on my Android 10 emulator, as the emulator keeps crashing. So I won't be checking that case for now.

aldee:
> Thanks for the contribution! I left a couple of comments.
> 
> With your reproduction, would you also be willing to see if the same issue occurs for video picking? The code makes the same `data.getData()` call, and it would be a good time to make the same fix if the same problem exists.

Hi, I've been able to reproduce the issue with video picking after downgrading the Android Emulator. So I will try to fix it on this PR as well

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -630,7 +630,20 @@ public boolean onActivityResult(
 
   private void handleChooseImageResult(int resultCode, Intent data) {
     if (resultCode == Activity.RESULT_OK && data != null) {
-      String path = fileUtils.getPathFromUri(activity, data.getData());
+      Uri uri = null;
+      // On several pre-Android 13 devices using Android Photo Picker, the Uri from getData() could be null.
+      if (data.getData() != null) {
+        uri = data.getData();
```
stuartmorgan, line 636: Why is this being called twice, instead of:

```java
      Uri uri = data.getData();
      // On several pre-Android 13 devices using Android Photo Picker, the Uri from getData() could be null.
      if (uri == null && data.getClipData() != null && data.getClipData().getItemCount() == 1) {
        ...
```

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -630,7 +630,20 @@ public boolean onActivityResult(
...
+      } else if (data.getClipData() != null && data.getClipData().getItemCount() == 1) {
+        uri = data.getClipData().getItemAt(0).getUri();
+
+        if (uri == null) {
```
stuartmorgan, line 640: This should not be nested here; as written there is a path through this method where the URI couldn't be extracted either way, but no error is returned here and `getPathFromUri` is still called.

stuartmorgan: (Which also means there's a missing unit test for that case.)

packages/image_picker/image_picker_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.8.9
```
stuartmorgan, line 1: This is just a bugfix, so should be 0.8.8+1.

stuartmorgan (changes requested):

aldee:
@stuartmorgan @gmackall it has been more than 2 weeks since I opened this PR, and there has been too much back and forth for such a simple fix.

I really wish you were this strict when you reviewed the previous person who wrote this part of the code, as they missed a simple null check.

I'm just trying to fix a simple bug with a simple solution, and I believe it is all good enough for a such a simple null chevk, but it seems there's always a reason to improve the code, which has been taking too much of my time and patience since not everyday we all have time to go back and forth to this PR.

I don't believe it is worth all this effort, since I believe the bugfix is critical since it will always happen in many Android devices that will return null when you try to retrieve the Uri using Intent.getData(), so it should be made available to the public as long as the it passes the checks enforced

Thanks

stuartmorgan:
> @stuartmorgan @gmackall it has been more than 2 weeks since I opened this PR

And it's been reviewed several times during those two weeks. We try to set clear expectations about review latency in our contributing docs; this is well within the bounds there. Obviously it would be ideal if all reviews could be immediate, but we're balancing a lot of priorities.

> and there has been too much back and forth for such a simple fix.

I'm not sure what "too much" is in relation to here, but there's no set amount of reviewing for a PR. It depends on the specifics of the code.

> I really wish you were this strict when you reviewed the previous person who wrote this part of the code

The goal of reviews isn't to be "strict", it's to improve the quality of the code over time. Catching bugs is one useful outcome of reviews, but catching every bug in review isn't a realistic expectation unfortunately.

> as they missed a simple null check.

As does the current version of your PR. I'm not clear as to why you believe that your PR has had too much review when it currently has exactly the same kind of bug that you are saying you wish we would catch in review.

> I'm just trying to fix a simple bug with a simple solution

We appreciate the PR, and share the goal of fixing the bug. In addition, we want to ensure that we have test coverage to prevent this bug from recurring when people edit this code in the future, and also that the changes don't introduce any new bugs (such as new codepaths where null values are unhandled).

> and I believe it is all good enough for a such a simple null chevk

I can't agree that code with an obvious path through it that will create an unhandled null is good enough for a null check.

> it should be made available to the public as long as the it passes the checks enforced

Agreed. But having approvals from two project members *is* one of the checks enforced for Flutter PRs, as explained in our contributing guide.

(And of course, if people want to distribute code changes without going through our processes, there are ways to do so, such as maintaining a fork.)

> I don't believe it is worth all this effort

It is of course entirely your choice whether or not to continue with this PR.

Is there a specific action you are looking for from us at this point?

aldee:
@stuartmorgan the short answer is I'd rather have you, who already get the general idea of the bug that I'm trying to fix here and the solution, or someone else to continue or create a new PR to fix this issue.

I'm really not looking for a credit here, I'm just trying to fix something that should've been prevented as it was a very simple null check that has been affecting several of my app's users and I'm sure others have encountered the same issue.

I really think there shouldn't be anymore issue with the recent version of my changes, that's why I felt there's too much back and forth here. 

To be clear, I'm not expecting you or whoever is going to review the code to respond quickly, as I know it's not the reviewer's full time job here. But when you expect perfection with a simple code changes that would benefit a lot of people, the ROI is just isn't there. I don't think it's worth it to go into the small details (which I've been doing until the last part, and I still think all the possible cases in regards to retrieving the Uri have been handled) everywhere.

Since I've found this bug and found the solution to fix it, I've been using a fork as I need the fix immediately at the time, but having the fix included in the main repo would be much easier to maintain in the future rather than having a fork. That's why I created this PR to save me some pain the in the future.

aldee:
I've even added another bugfix that was requested by @gmackall that didn't concern my use case at all...

stuartmorgan:
> @stuartmorgan the short answer is I'd rather have you, who already get the general idea of the bug that I'm trying to fix here and the solution, or someone else to continue or create a new PR to fix this issue.

If you aren't interested in continuing this PR, we can certainly close it and leave a comment in the issue indicating there was a WIP PR, and wait for someone else to continue it.

I'm confused though, because you've also posted a new version and marked it for review. It would be helpful if you could give a clear indication of whether you want to continue with the PR so we know how to proceed.

> I really think there shouldn't be anymore issue with the recent version of my changes, that's why I felt there's too much back and forth here.

I'm not sure what this means exactly. Are you saying that [this comment](https://github.com/flutter/packages/pull/4836#discussion_r1328896377) was factually incorrect, or that you don't think that crashing due to unchecked nulls is an issue?

> But when you expect perfection with a simple code changes that would benefit a lot of people

In both my rounds of code review I have pointed out that you are using values without null checking them. I don't agree with the characterization that preventing a PR whose entire purpose is to fix use-of-null crashes from introducing **new* use-of-null crashes is requiring perfection, and I review code accordingly.

> the ROI is just isn't there. I don't think it's worth it to go into the small details

As I said above, that's of course your decision to make. But to be clear, there are two outcomes I can see here:
1. We continue the review process until it's approved, and then land the PR, or
2. we close the PR, and the issue can remain open until someone else continues it.

If you are trying to argue for an option 3 where we land the PR before reviewers approve it because you personally disagree with the quality bar we use for reviews, that's not how our PR process works.

gmackall (approved):
This LGTM at this point. I'm sorry to hear the review process was frustrating, but thanks for the contribution regardless! 

cc @tarrinneal for secondary review (or @stuartmorgan, if you wish to continue reviewing).

gmackall:
Also might be good as a follow up when this lands to refactor the changes into a common `getUriFromIntent`

tarrinneal:
I can give a review, but it still can't land until @stuartmorgan approves.

stuartmorgan (approved):

aldee:
Thanks, everyone @gmackall @stuartmorgan @tarrinneal ! Sorry for the rather rough comments from me and the late follow-up now, I wasn't feeling well for the past 2 weeks.

PR #4847, opened 2023-09-05
[ios_platform_images]  migrate objC to swift

Mairramer:
Migrate ios_platform_images to Swift.

Fixes https://github.com/flutter/flutter/issues/119101

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/ios_platform_images/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.0
```
stuartmorgan, line 1: What part of this change is client-breaking?

packages/ios_platform_images/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+* Migrate objC to Swift.
```
stuartmorgan, line 3: > Migrates to a Swift implementation.

packages/ios_platform_images/example/ios/Podfile
```diff
@@ -41,4 +38,4 @@ post_install do |installer|
   installer.pods_project.targets.each do |target|
     flutter_additional_ios_build_settings(target)
   end
-end
+end
```
stuartmorgan, line 41: Please do not remove the trailing newline.

packages/ios_platform_images/example/ios/RunnerTests/IosPlatformImagesTests.swift
```diff
@@ -0,0 +1,107 @@
...
+// found in the LICENSE file.
+
+import Flutter
+import XCTest
+
+@testable import ios_platform_images
+
+class IosPlatformImagesTests: XCTestCase {
+
+  var mockChannel: MockMethodChannel!
```
stuartmorgan, line 12: Why do this and setUp exist? You're already initializing the mock in every test anyway (which is the preferred approach).

packages/ios_platform_images/example/ios/RunnerTests/IosPlatformImagesTests.swift
```diff
@@ -0,0 +1,107 @@
...
+
+  override func setUp() {
+    super.setUp()
+    mockChannel = MockMethodChannel()
+  }
+
+  func testHandleMethodCall_loadImage() {
+    let name = "flutter"
```
stuartmorgan, line 20: Please give this a more descriptive name. `assetName` or `imageName` maybe?

packages/ios_platform_images/example/ios/RunnerTests/IosPlatformImagesTests.swift
```diff
@@ -0,0 +1,107 @@
...
+
```
stuartmorgan, line 21: Why is there so much vertical whitespace in these files? Setup can be a single logical block.

packages/ios_platform_images/example/ios/RunnerTests/IosPlatformImagesTests.swift
```diff
@@ -0,0 +1,107 @@
...
+      let scale = result?["scale"] as? CGFloat
+      let data = result?["data"] as? FlutterStandardTypedData
+
+      XCTAssertNotNil(scale)
+      XCTAssertNotNil(data)
+
+      resultExpectation.fulfill()
+    }
+
+    waitForExpectations(timeout: 2, handler: nil)
```
stuartmorgan, line 43: Flutter style prefers not using timeouts in tests, as they contribute to flake. This should just wait until the expectation is fulfilled.

Same in all of the tests below.

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
```diff
@@ -0,0 +1,69 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Flutter
+import Foundation
+
+public final class IosPlatformImagesPlugin: NSObject, FlutterPlugin {
+
+  private let channel: MethodChannel
```
stuartmorgan, line 10: This channel does not need to be an ivar; it is never needed outside of `register`. That also means we don't need a mock implementation of it for tests.

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
```diff
@@ -0,0 +1,69 @@
...
+    case "resolveURL":
+      resolveURL(call, result)
+    default:
+      result(FlutterMethodNotImplemented)
+    }
+  }
+
+  private func loadImage(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
+    guard let arguments = call.arguments as? [Any],
+      let name = arguments.first as? String,
```
stuartmorgan, line 38: Missing arguments should not silently fail; it's fine to force-unwrap in this case (especially since we'll convert to Pigeon shortly.

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
```diff
@@ -0,0 +1,69 @@
...
+      loadImage(call, result)
+    case "resolveURL":
+      resolveURL(call, result)
+    default:
+      result(FlutterMethodNotImplemented)
+    }
+  }
+
+  private func loadImage(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
+    guard let arguments = call.arguments as? [Any],
```
stuartmorgan, line 37: The arguments aren't an array. Have you successfully run the example with your changes and seen the image load?

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
```diff
@@ -0,0 +1,69 @@
...
+    else {
+      result(nil)
+      return
+    }
+
+    let data = image.pngData()
+
+    let imageResult: [String: Any] = [
+      "scale": image.scale,
+      "data": FlutterStandardTypedData(bytes: data!),
```
stuartmorgan, line 49: This will crash if creating the PNG data failed for any reason, which is a regression from the Obj-C implementation (which would return nil).

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
```diff
@@ -0,0 +1,69 @@
...
+    ]
+
+    result(imageResult)
```
stuartmorgan, line 52: What's the purpose of the intermediate `imageResult` variable, rather than inlining the result?

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  private func resolveURL(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
+    if let args = call.arguments as? [String?] {
+      let name = args[0]
+      let extensionName = args.count == 2 ? args[1] : nil
```
stuartmorgan, line 58: This is not the same logic as the Obj-C version, and is not correct. There will always be two arguments, but the second may be `NSNull`.

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
```diff
@@ -0,0 +1,69 @@
...
+
+      if let url = Bundle.main.url(forResource: name, withExtension: extensionName) {
```
stuartmorgan, line 60: Why the explicit branching, rather than optional chaining like the Obj-C version? The more `result` calls there are, the harder it is to make sure the logic is correctly calling it exactly one time.

packages/ios_platform_images/ios/Classes/MethodChannel.swift
```diff
@@ -0,0 +1,16 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Flutter
+
+/// A channel for platform code to communicate with the Dart code.
+protocol MethodChannel {
```
stuartmorgan, line 8: This class will also not be necessary.

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
```diff
@@ -0,0 +1,33 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Flutter
+import Foundation
+import UIKit
+
+extension UIImage {
```
stuartmorgan, line 9: Extending framework classes is more potential pain than it is worth; we should take the opportunity of this rewrite to make this a helper function rather than a static class extension.

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
```diff
@@ -0,0 +1,33 @@
...
+  static func flutterImage(withName name: String) -> UIImage? {
+    let imageName = (name as NSString).lastPathComponent
+    let path = (name as NSString).deletingLastPathComponent
```
stuartmorgan, line 12: Why are these being cast to NSString?

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
```diff
@@ -0,0 +1,33 @@
...
+    let screenScale = UIScreen.main.scale
+
+    var scaledImage: UIImage?
+
+    // Search for a scaled image in the bundle
```
stuartmorgan, line 17: Comments need to be properly punctuated sentences, per the style guide.

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
```diff
@@ -0,0 +1,33 @@
...
+    // The image name may be suffixed with a scale factor, e.g. @2x
+    // If a scaled image is not found, search for standard image
+
+    for scale in stride(from: Int(screenScale), through: 1, by: -1) {
```
stuartmorgan, line 21: Why is `screenScale` being initially created with the wrong type and then cast here? (Relatedly, why is it a separate variable at all, instead of inline like before?)

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
```diff
@@ -0,0 +1,33 @@
...
+      let scaledName = imageName.replacingOccurrences(of: ".", with: "@\(scale)x.")
```
stuartmorgan, line 22: Why have you changed the logic and dropped the `lookupKeyForAsset` calls? This is a completely different implementation.

If you believe there's an issue with the current implementation, that needs to be resolved in a separate PR. Swift conversions PRs should not be combining conversion with logic changes, as those are completely orthogonal things.

stuartmorgan (changes requested):
Thanks for the contribution!

It looks like we don't have any end-to-end tests for this plugin at the moment, so I added them in https://github.com/flutter/packages/pull/4899. Once that lands this can be synced with main; that will give us coverage to ensure that the PR isn't breaking anything (as I believe the current version does, per my comments below.)

packages/ios_platform_images/example/ios/RunnerTests/IosPlatformImagesTests.swift
Mairramer: I forgot remove him

packages/ios_platform_images/CHANGELOG.md
Mairramer: No change will break for the client.

stuartmorgan: Okay, this should be 0.2.3 as a new version then (per the second-to-last paragraph [here](https://dart.dev/tools/pub/versioning#semantic-versions)).

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
```diff
@@ -0,0 +1,33 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Flutter
+import Foundation
+import UIKit
+
+extension UIImage {
+  static func flutterImage(withName name: String) -> UIImage? {
```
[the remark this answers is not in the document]
Mairramer: I will add back

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
Mairramer: I test with lookupKeyForAsset in old and in new implementation and he dont works,at least it didn't work for me. I dont know why exactly, i will made some tests and try back again

Mairramer: I need fix this, when i pass only image name he works.

Mairramer: @stuartmorgan I reverted the implementation to how it should be and it works normally without changing the original logic. My test gave a false positive

stuartmorgan:
Please merge in a version of main newer than https://github.com/flutter/packages/pull/4899 so that we can ensure that the new integration tests still pass.

stuartmorgan:
Sorry, it turns out they aren't actually run in CI yet; https://github.com/flutter/packages/pull/4920 will need to land and then be merged in here as well.

In the meantime though, you can run them locally using [the repository tooling](https://github.com/flutter/packages/blob/main/script/tool/README.md) with the arguments: `drive-examples --ios --packages=ios_platform_images` and see that this breaks the integration tests (as expected since the argument parsing issue I flagged hasn't been fixed yet).

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
```diff
@@ -22,6 +22,13 @@
...
 
-@end
+func flutterImage(withName imageName: String) -> UIImage? {
+    for scale in stride(from: Int(UIScreen.main.scale), through: 1, by: -1) {
+      let scaledImage = imageName.replacingOccurrences(of: ".", with: "@\(scale)x.")
+      return UIImage(named: scaledImage, in: Bundle.main, compatibleWith: nil)
+    }
+
+    return UIImage(named: imageName, in: Bundle.main, compatibleWith: nil)
+  }
```
hellohuanlin, line 33: The indentation is a bit off in this file. Can you run the formatter locally?
Refer to [this wiki section](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#swift-migration-for-1p-plugins) on installing the formatter on your local machine.

Mairramer: @hellohuanlin @stuartmorgan  I have a question about the code I migrated, specifically about the image. Reading about the documentation [UIImage](https://developer.apple.com/documentation/uikit/uiimage) , I imagined that UIImage itself could take care of the image scale.
After @stuartmorgan  comment I came back with the previous code. Below is the implementation but it doesn't work as expected, I'm probably doing something wrong. I don't know what to do in this situation.

func flutterImage(withName name: String) -> UIImage? {
  let components = name.components(separatedBy: "/")
  let filename = components.last
  let path = components.dropLast().joined(separator: "/")

  for screenScale in stride(from: Int(UIScreen.main.scale), to: 1, by: -1) {
    let key = FlutterDartProject.lookupKey(forAsset: "\(path)\(screenScale).0x/\(filename!)")
    if let image = UIImage(named: key, in: Bundle.main, compatibleWith: nil) {
      return image
    }
  }

  let key = FlutterDartProject.lookupKey(forAsset: name)
  return UIImage(named: key, in: Bundle.main, compatibleWith: nil)
}

hellohuanlin: Can you set a breakpoint there and see if Swift has the same `key` value as the ObjC?

hellohuanlin: Oh you are missing a `/` in the path

hellohuanlin: Can you add a unit test there? It would've been an easy-to-find mistake if we had unit test before.

Mairramer: > Oh you are missing a `/` in the path: `"\(path)\(screenScale).0x/\(filename!)"`

I removed it because it was //

Mairramer: > Can you set a breakpoint there and see if Swift has the same `key` value as the ObjC?
on scale -> key	String	"Frameworks/App.framework/flutter_assets/2.0x/flutter"	
normal ->    key	String	"Frameworks/App.framework/flutter_assets/flutter"

Mairramer: @hellohuanlin I get key of ObjC an give me :` key	__NSCFString *	@"Frameworks/App.framework/flutter_assets/flutter"	0x0000600001430230`

hellohuanlin: I meant, the ObjC version has a `/` between path and screenScale: 

`[NSString stringWithFormat:@"%@/%d.0x/%@", path, screenScale, filename]]`

and the Swift version is missing this `/`: 

`"\(path)\(screenScale).0x/\(filename!)"`

Mairramer: @hellohuanlin Ok, but it was supposed to load the default image for this reason but it didn't happen.
The image scaled should be `Frameworks/App.framework/flutter_assets//2.0x/flutter`?

hellohuanlin: No there shouldn't be `//` (double slash).

hellohuanlin: Can you set a break point there, take a screenshot of Xcode window (including your code and the debug window). I can help to take a look.

Mairramer: Captura de Tela 2023-09-13 às 20 52 12

stuartmorgan: >   let components = name.components(separatedBy: "/")
>   let filename = components.last
>   let path = components.dropLast().joined(separator: "/")

Is there some reason you rewrote this logic?

Mairramer: Before I was writing it as if it was ObjC, but now I can write it more clearly.

hellohuanlin: From the screenshot, you added the `/`, and the path seems to be correct. Is there still an issue?

Mairramer: Yes, returns nil

stuartmorgan: > Before I was writing it as if it was ObjC, but now I can write it more clearly. Not committed yet.

I'm not sure what this means, but what you have in both the PR and the screenshot is different logic than the working Obj-C version; rather than converting the working logic to Swift, you've written different code. As I explained above logic changes aren't something we want in a conversion PR. The fact that your version doesn't work is an example of why changing the logic while porting to Swift problematic (especially since in cases like the Dart argument parsing you seem to have changed the logic without a clear understanding of what the original code was doing).

hellohuanlin: > Yes, returns nil.

Does the objc version have exactly the same arguments passed in? And what does it return?

Mairramer: @hellohuanlin I wrote another ObjC test and the scaled key is:
`key	__NSCFString *	@"Frameworks/App.framework/flutter_assets//2.0x/flutter"	0x00006000018a0c80`, with double slash

Mairramer: > > Before I was writing it as if it was ObjC, but now I can write it more clearly. Not committed yet.
> 
> I'm not sure what this means, but what you have in both the PR and the screenshot is different logic than the working Obj-C version; rather than converting the working logic to Swift, you've written different code. As I explained above logic changes aren't something we want in a conversion PR. The fact that your version doesn't work is an example of why changing the logic while porting to Swift problematic (especially since in cases like the Dart argument parsing you seem to have changed the logic without a clear understanding of what the original code was doing).

I came back with the original logic to test, but it still doesn't work. I think it must have some connection with the UIImage, I still haven't figured out why.
In this recent logic that I sent, I cast String to NSString to have the same result as I have in ObjC

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
```diff
@@ -0,0 +1,73 @@
...
+    case "resolveURL":
+      resolveURL(call, result)
+    default:
+      result(FlutterMethodNotImplemented)
+    }
+  }
+
+  private func loadImage(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
+    if let name = call.arguments as? [String], let firstItem = name.first {
+      guard let image = flutterImageWithName(withName: firstItem) else {
```
stuartmorgan, line 38: Why did you replace [a call to `[UIImage imageNamed:]` with a call to `flutterImageWithName`? Those aren't the same method, and aren't intended to do the same thing at all.

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
stuartmorgan: Sorry, I forgot that this is one of the rare cases of a method intended for use outside of the plugin, so changing its structure would be a breaking change.

We should do a precursor PR to add Obj-C unit tests of this method, with a comment explaining that they are testing a public API and must continue to work unchanged, and then preserve the Obj-C tests even in the Swift conversion to be sure we don't accidentally change the signature.

stuartmorgan: > We should do a precursor PR to add Obj-C unit tests of this method, with a comment explaining that they are testing a public API and must continue to work unchanged, and then preserve the Obj-C tests even in the Swift conversion to be sure we don't accidentally change the signature.

https://github.com/flutter/packages/pull/4925

stuartmorgan:
Since no mocking is necessary in the unit tests, please create a new PR that extracts just the unit tests you've added here, so that we can land that separately in advance of this PR. There have been bugs in this PR that would have been caught that way, such as the incorrect argument parsing, so doing it in two parts will help you find issues in the porting, isolated to specific unit tests.

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
hellohuanlin: Hmmm, I'm surprised that ObjC path uses double slash and it still works. I may need to look into this plugin for more details, but as Stuart said, it's better not to modify any behavior in a migration PR. 

To reduce risk of regression, we should backfill some unit tests. For example, adding tests for this function is a good starting point, since we are already making a mistake here. 

Backfilling tests also helps to better understand the existing ObjC behavior, which actually makes migration easier for us.

stuartmorgan: > To reduce risk of regression, we should backfill some unit tests. Adding tests for this function is a good starting point, since we are already making a mistake here.
> 
> Backfilling tests also helps to better understand the existing ObjC behavior, which actually makes migration easier for us.

See https://github.com/flutter/packages/pull/4847#discussion_r1326012844

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
Mairramer: I change to avoid  `Fatal error: Index out of range`

stuartmorgan: In what context? [This](https://github.com/flutter/packages/blob/59d7b6d86c77ac971dd14b8c92bad75837515b22/packages/ios_platform_images/lib/ios_platform_images.dart#L147) is the call, and it passes two arguments.

If you are referring to the unit tests you added, the solution is to make the unit tests match actual usage. Having unit tests that call the method in a way that production code never will, and then changing the implementation to work in the unit test but break production, makes the unit tests actively harmful.

Mairramer: I saw here in another test the reason for your comment, I will make the appropriate change

packages/ios_platform_images/ios/Classes/IosPlatformImagesPlugin.swift
Mairramer: I added it back as it was in the ObjC code

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
Mairramer: @hellohuanlin In my free time I will investigate the behavior of the ObjC code and compare it with the Swift code.

stuartmorgan: > Hmmm, I'm surprised that ObjC path uses double slash and it still works.

POSIX requires multiple consecutive separators in a path to be treated the same as a single separator, so it's not really surprising.

stuartmorgan:
Merging in `main` will now pick up all of the new tests for this plugin (including in CI).

stuartmorgan:
(I didn't intend to close this, I just hit the wrong button.)

stuartmorgan:
FYI, https://github.com/flutter/packages/pull/4945 will conflict with this, but it should be straightforward to resolve, and will simplify the new native unit tests.

stuartmorgan:
The Pigeon conversion landed, so this can be updated whenever you are ready to continue. (I don't anticipate any more changes to this package that would cause new conflicts.)

Mairramer:
@stuartmorgan @hellohuanlin  I managed to get back to working on the code migration after a few weeks and now everything works as expected.

packages/ios_platform_images/example/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
```
hellohuanlin, line 5: What triggers this change?

packages/ios_platform_images/example/ios/RunnerTests/UIImageIosPlatformImagesTests.swift
```diff
@@ -0,0 +1,39 @@
...
+import Flutter
+import XCTest
+
+@testable import ios_platform_images
+
+final class UIImageIosPlatformImagesTests: XCTestCase {
+
+  func testMultiResolutionImageUsesBest() {
+    if let image = flutterImageWithName(withName: "assets/multisize.png") {
+      XCTAssertNotNil(image)
```
hellohuanlin, line 14: the `image` inside `if let` will always be non null. You should do something like: 

```
What if image is nil? should we fail the test instead? 

```
func testMultiResolutionImageUsesBest() {
  guard let image = flutterImageWithName(...) else {
    XCTFail("Image not found)
  }
  // continue here
}
```

File: packages/ios_platform_images/example/ios/RunnerTests/UIImageIosPlatformImagesTests.swift [hunk 26]
```diff
@@ -0,0 +1,39 @@
...
+      } else {
+        XCTAssertEqual(image.size.height, height1x, accuracy: 0.00001)
+      }
+    }
+  }
+
+  func testSingleResolutionFindsImage() {
+    // When there is no resolution-aware asset, the main asset should be used.
+    if let image = flutterImageWithName(withName: "assets/monosize.png") {
+      XCTAssertNotNil(image)
```
hellohuanlin, line 31: same here

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
```diff
@@ -2,9 +2,9 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#import <UIKit/UIKit.h>
-
-@interface UIImage (ios_platform_images)
+import Flutter
+import Foundation
+import SwiftUI
```
hellohuanlin, line 7: why are we importing swiftui?

packages/ios_platform_images/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: ios_platform_images
 description: A plugin to share images between Flutter and iOS in add-to-app setups.
 repository: https://github.com/flutter/packages/tree/main/packages/ios_platform_images
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+ios_platform_images%22
-version: 0.2.2+3
+version:  0.2.3
```
hellohuanlin, line 5: extra space

hellohuanlin (changes requested):
Not a complete review (I haven't looked closely at the objc code yet), but main issue is the test part.

packages/ios_platform_images/example/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
stuartmorgan: This is likely from running with a newer version of Xcode; this kind of thing happens pretty much every time there's a new major version, and sometimes ever for minor versions IIRC. It's fine to land these as incidental changes.

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
Mairramer: i was testing another thing. Removed now.

packages/ios_platform_images/example/ios/RunnerTests/UIImage+ios_platform_imagesTests.m
```diff
@@ -1,43 +0,0 @@
...
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE file.
-
-@import ios_platform_images;
-@import XCTest;
-
-// The tests test the UIImage extension which is a public API intended for use from native code
-// outside of the plugin (see package README). Any change that requires changing existing tests
-// in this file (unless it's just to reflect changes to the test assets) is a BREAKING CHANGE for
-// the package.
```
hellohuanlin, line 11: We may want to keep this ObjC test untouched, since it's testing public API and we want to make sure no accidental breaking changes in ObjC.

packages/ios_platform_images/example/ios/Podfile
```diff
@@ -28,9 +28,6 @@ require File.expand_path(File.join('packages', 'flutter_tools', 'bin', 'podhelpe
 flutter_ios_podfile_setup
 
 target 'Runner' do
-  use_frameworks!
-  use_modular_headers!
```
stuartmorgan, line 32: Why are these being changed?

packages/ios_platform_images/example/ios/RunnerTests/UIImage+ios_platform_imagesTests.m
stuartmorgan: Yes, this needs to stay; see https://github.com/flutter/packages/pull/4847#discussion_r1325857344

This conversion is masking the fact that this PR is breaking the public API in the way described in that comment.

stuartmorgan (changes requested):

packages/ios_platform_images/example/ios/Podfile
Mairramer: It was unintentional.

packages/ios_platform_images/example/ios/RunnerTests/UIImage+ios_platform_imagesTests.m
Mairramer: I returned the ObjC test and removed the test with Swift. If you find it necessary, I will also add the test with Swift.

hellohuanlin: Just objc test should be fine as they are testing the same thing. Adding a swift version can cause maintenance overhead of keeping 2 files in sync.

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
```diff
@@ -0,0 +1,41 @@
...
+  /// For example, if your Flutter project's `pubspec.yaml` lists "assets/foo.png"
+  /// and "assets/2.0x/foo.png", calling
+  /// `[UIImage flutterImageWithName:@"assets/foo.png"]` will load
+  /// "assets/2.0x/foo.png".
+  ///
+  /// See also https://flutter.dev/docs/development/ui/assets-and-images
+  ///
+  /// Note: We don't yet support images from package dependencies (ex.
+  /// `AssetImage('icons/heart.png', package: 'my_icons')`).
+
```
hellohuanlin, line 26: extra blank line

packages/ios_platform_images/example/ios/RunnerTests/IosPlatformImagesTests.swift
```diff
@@ -0,0 +1,50 @@
...
+    XCTAssertNil(imageData)
+  }
+
+  func testResolveURL() {
+    let resourceName = "textfile"
+    let extensionName: String? = nil
+    do {
+      let url = try plugin.resolveUrl(resourceName: resourceName, extension: extensionName)
+      XCTAssertNotNil(url)
+      XCTAssertTrue(url!.contains(resourceName))
```
hellohuanlin, line 34: `XCTAssertTrue(url?.contains(resourceName) ?? false)`

packages/ios_platform_images/example/ios/RunnerTests/IosPlatformImagesTests.swift
```diff
@@ -0,0 +1,50 @@
...
+    } catch {
+      XCTFail("Error while resolving URL: \(error)")
+    }
+  }
+
+  func testResolveURLNotFound() {
+    let resourceName = "notFound"
+    let extensionName: String? = nil
```
hellohuanlin, line 42: same here

packages/ios_platform_images/example/ios/RunnerTests/IosPlatformImagesTests.swift
```diff
@@ -0,0 +1,50 @@
...
+  func testLoadImageNotFound() {
+    let assetName = "notFound"
+    let imageData = plugin.loadImage(name: assetName)
+
+    XCTAssertNil(imageData)
+  }
+
+  func testResolveURL() {
+    let resourceName = "textfile"
+    let extensionName: String? = nil
```
hellohuanlin, line 30: can just inline these 2 values

hellohuanlin (approved):
Looks clean with some nits

packages/ios_platform_images/ios/Classes/UIImageIosPlatformImages.swift
hellohuanlin: @Mairramer Does the Swift version still use double-slash? If so, can you add a TODO comment there to fix it?

Mairramer: Yes, it continues. Should I assign it to someone?

hellohuanlin: feel free to put my name there

packages/ios_platform_images/example/ios/RunnerTests/IosPlatformImagesTests.swift
```diff
@@ -26,25 +26,22 @@ class IosPlatformImagesTests: XCTestCase {
...
 
   func testResolveURL() {
-    let resourceName = "textfile"
-    let extensionName: String? = nil
     do {
-      let url = try plugin.resolveUrl(resourceName: resourceName, extension: extensionName)
+      let url = try plugin.resolveUrl(resourceName: "textfile", extension: nil)
       XCTAssertNotNil(url)
-      XCTAssertTrue(url!.contains(resourceName))
+      XCTAssertTrue(url?.contains("textfile") ?? false)
```
hellohuanlin, line 32: oh this string is used in 2 places. so shouldn't inline it

packages/ios_platform_images/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -49,26 +50,27 @@
 		0EF1CD9A3A3064B5289EF22E /* Pods-Runner.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Runner.release.xcconfig"; path = "Target Support Files/Pods-Runner/Pods-Runner.release.xcconfig"; sourceTree = "<group>"; };
 		1498D2321E8E86230040F4C2 /* GeneratedPluginRegistrant.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = GeneratedPluginRegistrant.h; sourceTree = "<group>"; };
 		1498D2331E8E89220040F4C2 /* GeneratedPluginRegistrant.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = GeneratedPluginRegistrant.m; sourceTree = "<group>"; };
-		33B0E32B2AB348FF0006F0C1 /* UIImage+ios_platform_imagesTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = "UIImage+ios_platform_imagesTests.m"; sourceTree = "<group>"; };
+		3131C9A22ADEFDA8000703FA /* UIImage+ios_platform_imagesTests.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; name = "UIImage+ios_platform_imagesTests.m"; path = "../../../../../../../Downloads/UIImage+ios_platform_imagesTests.m"; sourceTree = "<group>"; };
```
stuartmorgan, line 53: You'll need to revert all of the project changes related to this file; you've pointed it to a location outside of the repo on your local machine.

stuartmorgan (changes requested):

packages/ios_platform_images/example/ios/Runner.xcodeproj/project.pbxproj
hellohuanlin: Good observation! 
This probably happens when you drag files into xcode, and picked "reference only" instead of "copy files"

packages/ios_platform_images/ios/ios_platform_images.podspec
```diff
@@ -15,11 +15,13 @@ Downloaded by pub (not CocoaPods).
   s.author           = { 'Flutter Dev Team' => 'flutter-dev@googlegroups.com' }
   s.source           = { :http => 'https://github.com/flutter/packages/tree/main/packages/ios_platform_images' }
   s.documentation_url = 'https://pub.dev/packages/ios_platform_images'
-  s.source_files = 'Classes/**/*'
+  s.source_files = 'Classes/**/*.swift'
   s.dependency 'Flutter'
   s.platform = :ios, '11.0'
 
-  # Flutter.framework does not contain a i386 slice. Only x86_64 simulators are supported.
-  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES' }
```
stuartmorgan, line 23: This shouldn't be removed; I'll put a commit to re-add it (and remove the bridging header import in the unit test that was only necessary because of its removal).

stuartmorgan (approved):
I made a few trivial edits for style; LGTM with those.

PR #4731, opened 2023-08-17
[ci] Update minimums for 3.13 stable

stuartmorgan:
- Updates the N-1 and N-2 analysis bot configs to 3.7 and 3.10
- Updates the minimum allowed SDK to 3.7, since that's now the earliest we have any testing for.
- Updates all packages to Flutter 3.7/Dart 2.19 as a minimum.
  - In packages where we hadn't released since the last minimum version bump, I removed the previous line from the `#NEXT` section to make it less confusing when we do release.

Per https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version this does not update any package versions.

PR #4851, opened 2023-09-05
[flutter_markdown] Fix changelog regarding minimum supported SDK version

nilsreichardt:
The changelog of the `flutter_markdown` package is a bit confusing. First it says:

> Updates minimum supported SDK version to Flutter 3.10/Dart 3.0

and in the latest update is says

> Updates minimum supported SDK version to Flutter 3.7/Dart 2.19

Just from reading this I would expect that the minimum SDK lowered with the latest update. However, this is not the case.

After a bit of researching, I found this was caused by https://github.com/flutter/packages/pull/4731. The mentioned PR only updates the minimum SDK of the example but not of the package, see: https://github.com/flutter/packages/pull/4731/files#diff-e780ee3aec5fdc10512b452b5464878a11caf22fb23d28f725324308ad243d17R6

Therefore, I clarified the changelog with this PR.

telegram-cloud-photo-size-2-5456132896298487892-y

image

Fixes https://github.com/flutter/flutter/issues/134056

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/flutter_markdown/CHANGELOG.md
+++ b/packages/flutter_markdown/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.6.17+2
 
 * Adds pub topics to package metadata.
-* Updates minimum supported SDK version to Flutter 3.7/Dart 2.19.
+* Updates minimum supported SDK version to Flutter 3.7/Dart 2.19 of the example.
 
 ## 0.6.17+1
 
--- a/packages/flutter_markdown/CHANGELOG.md
+++ b/packages/flutter_markdown/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## 0.6.17+2
 
 * Adds pub topics to package metadata.
-* Updates minimum supported SDK version to Flutter 3.7/Dart 2.19 of the example.
 
 ## 0.6.17+1
 
--- a/packages/flutter_markdown/CHANGELOG.md
+++ b/packages/flutter_markdown/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.6.17+3
+
+* Fixes an incorrect note about SDK versions in the 0.6.17+2 CHANGELOG.md entry.
+
 ## 0.6.17+2
 
 * Adds pub topics to package metadata.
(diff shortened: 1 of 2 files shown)


domesticmouse (approved):

nilsreichardt:
Not sure, if a change like this still needs a version change. If yes, I can bump the version in the `pubspec.yaml` and add a note in the changelog for changing the changelog.

packages/flutter_markdown/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## 0.6.17+2
 
 * Adds pub topics to package metadata.
-* Updates minimum supported SDK version to Flutter 3.7/Dart 2.19.
+* Updates minimum supported SDK version to Flutter 3.7/Dart 2.19 of the example.
```
stuartmorgan, line 4: This line can just be removed; it was only added by accident since I didn't notice that it was example-only.

stuartmorgan:
> Not sure, if a change like this still needs a version change.

It depends whether we want the changelog to continue to be confusing on pub.dev until there happens to be some other change to the package or not. This is minor enough that I don't have a strong feeling either way.

nilsreichardt:
> It depends whether we want the changelog to continue to be confusing on pub.dev until there happens to be some other change to the package or not. This is minor enough that I don't have a strong feeling either way.

If you don't mind, let's add this to the changelog so that we can publish a new version to update the changelog on pub.dev 👍 

Should I just take the following sentences as change note? Feel free to suggest a different one.

> Fixes a change note in the `CHANGELOG.md`

stuartmorgan:
CHANGELOG entries should be specific about what they are fixing, otherwise they don't convey much information to the reader:

> Fixes an incorrect note about SDK versions in the 0.6.17+2 CHANGLOG.md entry.

nilsreichardt:
Thanks, I updated the version and the changelog.

stuartmorgan (approved):
LGTM, thanks!

PR #4862, opened 2023-09-06
[rfw, ci] Regenerate goldens, manually roll #4835

ditman:
This PR updates the goldens for `package:rfw`.

This PR also:

* Modifies `run_tests.sh` so it forwards all its command-line parameters to the inner `test_coverage.dart` bin. (This enables passing `--update-goldens` without much fuss.)
* Updates the documentation on how to update goldens in the package.

## Issues

This PR fixes #4835, and should enable the auto-roller to resume.

## Pre-launch Checklist

- [x] I read and followed the [relevant style guides] ~~and ran the auto-formatter.~~ (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].


ditman:
This PR should be text-exempt. It actually fixes tests so `flutter/flutter` may resume rolling into `flutter/packages`.

packages/rfw/README.md
```diff
@@ -991,3 +991,34 @@ exclude that line from the coverage calculation.
 Golden tests are only run against the Flutter master channel and only
 run on Linux, since minor rendering differences are expected on
 different platforms and on different versions of Flutter.
+
```
stuartmorgan, line 994: We should really move this whole Contributing section to a CONTRIBUTING.md, so it's not showing up on the pub.dev README, but that's out of scope here.

stuartmorgan (approved):
LGTM

packages/rfw/README.md
ditman: Ah yes, that's a good call. I was tempted to move all the testing information to a README.md inside the `test` directory as well.

stuartmorgan:
test-exempt: is a test

ditman:
(PS, I've just noticed the png diffing in github, it's pretty handy!)

packages/rfw/README.md
ditman: Yikes, updating this README.md forces me to update the CHANGELOG. I'll add a `NEXT` entry, since these additional docs are more interesting for us than for the general public and might be moved.

ditman:
@stuartmorgan do I still need an `override: no versioning needed` after adding a `NEXT` entry in the changelog?

ditman:
Nah, I'm going to do the CONTRIBUTING.md change, and version rfw normally, it doesn't make much sense to skip publishing the readme changes.

packages/rfw/README.md
ditman: > We should really move this whole Contributing section to a CONTRIBUTING.md, so it's not showing up on the pub.dev README

Done, this way we have a good reason to properly publish the package as a z version.

stuartmorgan:
> @stuartmorgan do I still need an `override: no versioning needed` after adding a `NEXT` entry in the changelog?

For future reference, yes. If we didn't have the explicit check, people would use `NEXT` all the time via cargo-culting and we'd forget to check that it was actually right because CI wouldn't warn us. With the label, it's impossible to not think about it.

ditman:
The last check has passed here: https://ci.chromium.org/ui/p/flutter/builders/try/Linux_android%20android_platform_tests_shard_2%20master/1663/overview

But the GH status hasn't updated. I'm tempted to manually merge this one (otherwise I'll push an empty commit and let CI try again)

stuartmorgan:
It's fine to merge manually in cases like this where we can see that the tests did in fact pass and it's just a report-back-to-GitHub problem.

ditman:
Will land manually to unblock the roll, and keep an eye on the CI on main.

PR #4882, opened 2023-09-10
[go_router] Fixed so that it works even if Scaffold.showBottomSheet is closed.

lynrin:
Fixed an issue where using Scaffold.showBottomSheet would prevent screen transitions using the AppBar's back button.

- [flutter/issue#134370](/flutter/flutter/issues/134370)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/go_router/lib/src/builder.dart
+++ b/packages/go_router/lib/src/builder.dart
@@ -594,7 +594,14 @@ class _PagePopContext {
     final Page<Object?> page = route.settings as Page<Object?>;
 
     final RouteMatch match = _routeMatchesLookUp[page]!.last;
-    _routeMatchesLookUp[page]!.removeLast();
+
+    // If a popup is opened using something like Scaffold.showBottomSheet,
+    // it will be registered in LocalHistoryRoute._localHistory,
+    // so it should be deleted from _routeMatchesLookUp only if
+    // LocalHistoryRoute.willHandlePopInternally is false.
+    if (route is LocalHistoryRoute && route.willHandlePopInternally == false) {
+      _routeMatchesLookUp[page]!.removeLast();
+    }
 
     return onPopPageWithRouteMatch(route, result, match);
   }
--- a/packages/go_router/CHANGELOG.md
+++ b/packages/go_router/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Fixed so that it works even if Scaffold.showBottomSheet is closed.
+
 ## 10.1.2
 
 * Adds pub topics to package metadata.
--- a/packages/go_router/pubspec.yaml
+++ b/packages/go_router/pubspec.yaml
@@ -1,7 +1,7 @@
 name: go_router
 description: A declarative router for Flutter based on Navigation 2 supporting
   deep linking, data-driven routes and more
-version: 10.1.2
+version: 10.1.3
 repository: https://github.com/flutter/packages/tree/main/packages/go_router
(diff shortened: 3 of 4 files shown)


chunhtai:
I can't reproduce the issue, the back button works for me. unless I have pressed `Go to Home page`. It seems to me that the go to home page is the issue here.

packages/go_router/lib/src/builder.dart
```diff
@@ -594,7 +594,14 @@ class _PagePopContext {
     final Page<Object?> page = route.settings as Page<Object?>;
 
     final RouteMatch match = _routeMatchesLookUp[page]!.last;
-    _routeMatchesLookUp[page]!.removeLast();
+
+    // If a popup is opened using something like Scaffold.showBottomSheet,
+    // it will be registered in LocalHistoryRoute._localHistory,
+    // so it should be deleted from _routeMatchesLookUp only if
+    // LocalHistoryRoute.willHandlePopInternally is false.
+    if (route is LocalHistoryRoute && route.willHandlePopInternally == false) {
```
chunhtai, line 602: I am a bit confused, if willHandlePopInternally is true, it should not call this function at the first place

chunhtai (commented):
See my reply

packages/go_router/lib/src/builder.dart
lynrin: It says "it should not call this function at the first place", but if you close the Bottom Sheet displayed with showBottomSheet by calling "Navigator.pop(context);", go_router will end up here.

lynrin:
Thank you @chunhtai for checking.

I was not aware of the existing issue flutter/flutter#98736.
The cause seems to be the way LocalHistoryEntry is handled.

I received a comment to builder.dart saying "I am a bit confused, if willHandlePopInternally is true, it should not call this function at the first place", but in order to close the BottomSheet,
```dart
// A call from a Widget displayed on the bottom sheet.
Navigator.pop(context);
```
Since it is only a call, you will need to consider measures to prevent it from reaching this point as an internal process of go_router.

paul-appliquette:
In a similar manner, I also see this issue occurring when trying to close a `Scaffold`'s NavigationDrawer by calling `Navigator.pop(context)` in `NavigationDrawer.onDestinationSelected`.

Result:
```
======== Exception caught by gesture ===============================================================
The following StateError was thrown while handling a gesture:
Bad state: No element

#0      List.last (dart:core-patch/growable_array.dart:348:5)
#1      _PagePopContext.onPopPage (package:go_router/src/builder.dart:596:57)
#2      NavigatorState.pop (package:flutter/src/widgets/navigator.dart:5045:28)
#3      Navigator.pop (package:flutter/src/widgets/navigator.dart:2507:27)
#4      _MainNavigationDrawerState.build.<anonymous closure> (package:my_app/widgets/common/main_navigation_drawer.dart:67:23)
#5      NavigationDrawer.build.wrapChild.<anonymous closure>.<anonymous closure> (package:flutter/src/material/navigation_drawer.dart:153:39)
#6      _InkResponseState.handleTap (package:flutter/src/material/ink_well.dart:1154:21)
#7      GestureRecognizer.invokeCallback (package:flutter/src/gestures/recognizer.dart:275:24)
```

The comment from @lynrin that _"The cause seems to be the way LocalHistoryEntry is handled."_ makes sense to me.

lynrin:
It seems that this has not occurred since the release of v11.0.0.
This PR will be closed.

PR #4898, opened 2023-09-11
Migrate Material curves to new names

guidezpl:
Cleanup following https://github.com/flutter/flutter/pull/129942 and preceding deprecation of 3 curves with new names.

Part of https://github.com/flutter/flutter/issues/116525

### DIFF
--- a/packages/animations/lib/src/fade_scale_transition.dart
+++ b/packages/animations/lib/src/fade_scale_transition.dart
@@ -134,7 +134,7 @@ class FadeScaleTransition extends StatelessWidget {
   static final Animatable<double> _scaleInTransition = Tween<double>(
     begin: 0.80,
     end: 1.00,
-  ).chain(CurveTween(curve: decelerateEasing));
+  ).chain(CurveTween(curve: Easing.legacyDecelerate));
   static final Animatable<double> _fadeOutTransition = Tween<double>(
     begin: 1.0,
     end: 0.0,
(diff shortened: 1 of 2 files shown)


goderbauer (approved):
LGTM

goderbauer:
Oh, but looks like this hasn't landed in stable yet (see failing checks)?

Hixie:
test-exempt: refactor with no semantic change

guidezpl:
Yes, and after reading up on https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#supported-flutter-versions, I'll hold off on landing this until the next stable release

guidezpl:
Ah thank you, I missed this instruction in https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#supported-flutter-versions

stuartmorgan:
@guidezpl Please do not ever merge a PR that has a failing test in this repository unless the PR is fixing tree breakage.

guidezpl:
Can the repository rules be updated to make this clear? Otherwise, this check just looks advisory

image

stuartmorgan:
> Can the repository rules be updated to make this clear?

As far as I know the configuration of this repository is the same as the other Flutter repositories. I don't have the ability to check as I'm not an admin, so if you want to see repo configuration change you would need to file an infra issue. (I'm not sure that setting would allow for fixing tree breakage though?)

> Otherwise, this check just looks advisory

Our [documented process](https://github.com/flutter/flutter/wiki/Tree-hygiene#overview) for all Flutter repos is to use `autosubmit`, which would not have landed your PR, and would have left a comment saying it couldn't be landed with that failure. If you are bypassing the documented process, you are responsible for following documented policies (in this case "Make sure your PR passes all the pre-commit tests.")

guidezpl:
Fair enough. Regarding repository rules, I'll file a separate issue. I like the way GitHub lets you bypass the rules for cases such as fixing the tree

image

PR #4903, opened 2023-09-11
[camerax] Implements torch mode

camsim99:
Implements the torch flash mode. Also wraps classes necessary for the implementation (a method in `Camera`, `CameraControl`).

Fixes https://github.com/flutter/flutter/issues/120715.
Fixes https://github.com/flutter/flutter/issues/115846.
Part of https://github.com/flutter/flutter/issues/115847.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/CameraAndroidCameraxPlugin.java
```diff
@@ -81,7 +82,8 @@ public void setUp(
     GeneratedCameraXLibrary.LiveDataHostApi.setup(binaryMessenger, liveDataHostApiImpl);
     GeneratedCameraXLibrary.ObserverHostApi.setup(
         binaryMessenger, new ObserverHostApiImpl(binaryMessenger, instanceManager));
-    imageAnalysisHostApiImpl = new ImageAnalysisHostApiImpl(binaryMessenger, instanceManager);
+    imageAnalysisHostApiImpl =
+        new ImageAnalysisHostApiImpl(binaryMessenger, instanceManager, context);
```
camsim99, line 86: This PR also makes a fix to this implementation by adding a `Context` to the constructor to remove an unnecessary call to update the context for the Host API implementations after running this method also setting the `Context` for those implementations (see line 131).

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/ImageAnalysisHostApiImpl.java
camsim99, line 1: See https://github.com/flutter/packages/pull/4903/files#r1323473452

camsim99:
After revisiting https://developer.android.com/training/camerax/configuration#torch, I realize I could get the torch state as well as whether or not the device has a flash unit and use that info in this PR versus tracking the torch state myself and not throwing an error/warning when torch state is not possible. If folks have strong opinions, I can fix in this PR; otherwise, I'll file an issue and fix this after I finish feature parity.

packages/camera/camera_android_camerax/pigeons/camerax_library.dart
```diff
@@ -417,3 +419,14 @@ abstract class FallbackStrategyHostApi {
   void create(int identifier, VideoQuality quality,
       VideoResolutionFallbackRule fallbackRule);
 }
+
+@HostApi(dartHostTestHandler: 'TestCameraControlHostApi')
+abstract class CameraControlHostApi {
+  @async
+  void enableTorch(int identifier, bool torch);
```
camsim99, line 426: @bparrishMines If the `Result` this returns ends up with an error, would I be able to catch this from the Dart side and throw an error there?

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -491,6 +496,14 @@ class AndroidCameraCameraX extends CameraPlatform {
   /// Sets the flash mode for the selected camera.
   @override
   Future<void> setFlashMode(int cameraId, FlashMode mode) async {
+    CameraControl? cameraControl;
+    // Turn off torch mode if it is enabled and not being redundantly set.
+    if (mode != FlashMode.torch && torchEnabled) {
```
gmackall, line 501: As currently written, we would have different behavior for 
```
...
setFlashMode(camera_id, FlashMode.auto)
setFlashMode(camera_id, FlashMode.torch)
takePicture(camera_id)
// results in taking a picture with both underlying camerax auto flash mode, and torch mode on
```
and
```
...
setFlashMode(camera_id, FlashMode.torch)
setFlashMode(camera_id, FlashMode.auto)
takePicture(camera_id)
// results in taking a picture with torch mode off and auto flash mode on
```
which could be confusing (I don't think this would be confusing if it were always just the case that the mode set last is the enabled one, but in this case the result is that you either get both set, or only one, depending on order).

I'm not sure if the best way to fix it is just to make the documentation very clear, or to do a similar thing where we unset the flash modes when setting the torch mode (like we unset the torch mode when setting the flash). What do you think?

All of the awkwardness seems to come from the fact that camerax behaves differently here, where flash and torch are independent configurations that can both be enabled or disabled at the same time. But it seems like we can't do that without adding a `FlashMode.torchOn` and `FlashMode.torchOff` to the type.

gmackall: Also most camerax examples seem to guard the setting of the flashmode with a [`hasFlashUnit`](https://developer.android.com/reference/androidx/camera/core/CameraInfo#hasFlashUnit()) call on the corresponding cameraInfo. 

It looks like both setting the torch and flash mode will just take no effect if there is no flash unit, so it probably isn't crucial to include, but it would be good to note that behavior in the documentation here

gmackall (commented):
Mostly LGTM, just have one question about how flash modes and torch modes interact with each other.

packages/camera/camera_android_camerax/pigeons/camerax_library.dart
bparrishMines: Yea, an exception on the Java side should cause a `PlatformException` error on the Dart side.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: To your first point about the behavior differences, I don't think that is true because when non-torch modes are set, it is saved and used when taking a picture. Setting the torch sets that saved mode to null, so setting auto then torch will clear auto as the saved setting.

As to the `hasFlashUnit` gating that seems to happen, I'l document that behavior and leave a todo to come back and use it for better error handling.

packages/camera/camera_android_camerax/pigeons/camerax_library.dart
camsim99: I'm going to modify this PR to catch `PlatformException`s that may be cause by CameraX failing to turn on torch mode.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: Oh wait I see what you're saying on the behavior difference in terms of CameraX. Yeah I'll update this PR to clear the flash mode when the mode isn't changed because that is undesired behavior.

gmackall: Yeah it's actually not true exactly as written, but if I change it to 

```
...
setFlashMode(camera_id, FlashMode.auto)
takePicture(camera_id)
setFlashMode(camera_id, FlashMode.torch)
takePicture(camera_id)
// results in taking a picture with both underlying camerax auto flash mode, and torch mode on
```
and
```
...
setFlashMode(camera_id, FlashMode.torch)
takePicture(camera_id)
setFlashMode(camera_id, FlashMode.auto)
takePicture(camera_id)
// results in taking a picture with torch mode off and auto flash mode on
```

then it should be true, right? Because we [don't update the flash mode](https://github.com/flutter/packages/blob/bb0e9d76b998fb3b858b9304fcd12becd098c051/packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart#L489) used by the image capture after we set the `_currentFlashMode` to null.

camsim99: Right yeah I'm going to make sure we update the flash mode to be off if the torch is on and the saved mode is null.

camsim99:
@gmackall Made the fix on your comment and added a test for it and also added error handling for failure to change torch state in the latest commit.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -491,6 +500,14 @@ class AndroidCameraCameraX extends CameraPlatform {
   /// Sets the flash mode for the selected camera.
```
gmackall, line 500: nit: Might be worth adding to the documentation here that this method combines the notion of flash and torch from

https://developer.android.com/reference/androidx/camera/core/ImageCapture and
https://developer.android.com/reference/androidx/camera/core/CameraControl#enableTorch(boolean).

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -491,6 +500,14 @@ class AndroidCameraCameraX extends CameraPlatform {
...
   @override
   Future<void> setFlashMode(int cameraId, FlashMode mode) async {
+    CameraControl? cameraControl;
+    // Turn off torch mode if it is enabled and not being redundantly set.
+    if (mode != FlashMode.torch && torchEnabled) {
+      cameraControl = await camera!.getCameraControl();
+      await cameraControl.enableTorch(false);
```
gmackall, line 507: it still might be worth adding a note to the documentation here that points out the difference between this method and the underlying CameraX apis (specifically just that torch and flash can't be set independently in this plugin, like they can with the CameraX apis).

So something like
```
setFlashMode(id, FlashMode.torch)
takePicture(id)
setFlashMode(id, FlashMode.auto)
takePicture(id)
```
would behave differently in the plugin (torch disabled, flash mode on auto) from what someone familiar with the native apis would expect (**torch enabled**, flash mode on auto).

gmackall (approved):
LGTM, minus some requests for explanation in the documentation of `setFlashMode`!

PR #4904, opened 2023-09-11
[tool] Add a package inclusion filter

stuartmorgan:
Creates a package filter flag that is an opt-in mirror to the `exclude` flag, for use with the automatic package selection options used in CI. Like `exclude`, it allows for YAML files as input.

This allows for creating split test runs in CI (e.g., during incremental migrations), where the same file is an inclusion filter for one run and an exclusion filter for the other, guaranteeing that tests are in one or the other without the possibility of some tests falling through the cracks.

### DIFF
--- a/script/tool/lib/src/common/package_command.dart
+++ b/script/tool/lib/src/common/package_command.dart
@@ -98,6 +98,14 @@ abstract class PackageCommand extends Command<void> {
             'Cannot be combined with $_packagesArg.\n\n'
             'This is intended for use in CI.\n',
         hide: true);
+    argParser.addMultiOption(_filterPackagesArg,
+        help: 'Filters any selected packages to only those included in this '
+            'list. This is intended for use in CI with flags such as '
+            '--$_packagesForBranchArg.\n\n'
+            'Entries can be package names or YAML files that contain a list '
+            'of package names.',
+        defaultsTo: <String>[],
+        hide: true);
     argParser.addFlag(_currentPackageArg,
         negatable: false,
         help:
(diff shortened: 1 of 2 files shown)


stuartmorgan:
This is intended as a generic solution to the problem of https://github.com/flutter/packages/pull/4820, where we want tests run in one of two modes, while still making it very hard to accidentally lose any tests (it would require eliminating one of the runs without removing the flag from the other, which I think we can effectively guard against with comments on the scripts).

So for that PR you could make, e.g., a `still_requires_api_33_emulators.yaml`, and then the shards with 33 would run a script that passes that with `--filter-packages-to`, while the 34 shards would instead pass the same file to `--exclude`.

reidbaker (approved):

reidbaker:
Test failure is clearly unrelated. 

Potentially flakey "loadUrl with headers" test 
https://firebase.corp.google.com/project/flutter-infra-staging/testlab/histories/bh.555af5d3de4d333b/matrices/7947959860904394775/details?stepId=bs.52b034256ab1b4b8&testCaseId=6&tabId=logs&query=et.f3a383dc00dd0737&deeplink=true
```
java.lang.Exception: ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞═════════════════
The following TestFailure was thrown running a test:
Expected: 'http://0.0.0.0:37207/headers'
Actual: 'http://0.0.0.0:37207/hello.txt'
Which: is different.
Expected: ... 0:37207/headers
Actual: ... 0:37207/hello.txt
```

PR #4912, opened 2023-09-12
Replace collection type lints with more general lint

kevmoo:
See https://dart.dev/tools/linter-rules/collection_methods_unrelated_type


stuartmorgan (approved):
LGTM

PR #4920, opened 2023-09-13
[ci] Enable ios_platform_images tests

stuartmorgan:
https://github.com/flutter/packages/pull/4899 added integration tests, but I forgot that they were being excluded in CI. This enables them.

Fixes https://github.com/flutter/flutter/issues/82208


tarrinneal (approved):
I guess this might be important

stuartmorgan:
Note to self: I actually checked that the test is running in CI this time.

PR #4923, opened 2023-09-14
[video_player] isCompleted event.

tarrinneal:
Adds `isCompleted` event to `VideoPlayerEvent`.

fixes https://github.com/flutter/flutter/issues/21929

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -113,6 +113,9 @@ class VideoPlayerValue {
   final String? errorDescription;
 
   /// True if video has finished playing to end.
+  ///
+  /// Reverts to false if video position changes, or video begins playing.
+  /// Does not update if video is looping.
   final bool isCompleted;
 
   /// The [size] of the currently loaded video.
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -547,7 +547,7 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   /// Sets whether or not the video should loop after playing once. See also
   /// [VideoPlayerValue.isLooping].
   Future<void> setLooping(bool looping) async {
-    value = value.copyWith(isLooping: looping, isCompleted: false);
+    value = value.copyWith(isLooping: looping);
     await _applyLooping();
   }
 
@@ -756,7 +756,7 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     value = value.copyWith(
       position: position,
       caption: _getCaptionAt(position),
-      isCompleted: false,
+      isCompleted: position == value.duration,
     );
   }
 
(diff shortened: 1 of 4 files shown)


packages/video_player/video_player/lib/video_player.dart
```diff
@@ -737,6 +756,7 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     value = value.copyWith(
       position: position,
       caption: _getCaptionAt(position),
+      isCompleted: false,
```
tarrinneal, line 759: I'm not sure whether this is needed/wanted. I do worry that it may cause inconsistencies since the logic that sets `isCompleted` to true also seeks to the end.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -528,7 +547,7 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   /// Sets whether or not the video should loop after playing once. See also
   /// [VideoPlayerValue.isLooping].
   Future<void> setLooping(bool looping) async {
-    value = value.copyWith(isLooping: looping);
+    value = value.copyWith(isLooping: looping, isCompleted: false);
```
stuartmorgan, line 550: Does enabling looping start playback? I would have assumed not, and if it's sitting paused and completed, enabling looping for the future doesn't seem like it should clean `isCompleted`.

packages/video_player/video_player/lib/video_player.dart
stuartmorgan: Can't we pass nil here if `position == duration` to avoid that logic clearing the flag?

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -1204,6 +1205,64 @@ void main() {
...
+        if (controller.value.isCompleted) {
+          isCompletedTest();
+        }
+      }
+    });
+
+    fakeVideoEventStream.add(VideoEvent(eventType: VideoEventType.completed));
+  });
+
+  test('isCompleted updates on video end', () async {
```
stuartmorgan, line 1236: This has the same name as the test above.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -1204,6 +1205,64 @@ void main() {
...
+              .add(VideoEvent(eventType: VideoEventType.isPlayingStateUpdate));
+        } else {
+          isNoLongerCompletedTest();
+        }
+      }
+    });
+
+    fakeVideoEventStream.add(VideoEvent(eventType: VideoEventType.completed));
+  });
 }
```
stuartmorgan, line 1267: There should be a test for each case that is supposed to set `isCompleted` back to `false`.

stuartmorgan (changes requested):

packages/video_player/video_player/lib/video_player.dart
tarrinneal: I set this to just be set to position == duration.

packages/video_player/video_player/test/video_player_test.dart
tarrinneal: oops

tarrinneal:
> LGTM
> 
> I suspect we'll need to tune edge cases for this state based on feedback from real-world use, but this looks good as a baseline for that.

Agreed, I'll keep an eye on it.

PR #4961, opened 2023-09-21
[video_player] Ensures autoplay is false on the web.

ditman:
### Description

This PR changes the initialization of the `video` element used by the web implementation of the video_player plugin to **ensure the `autoplay` attribute is set to _**false**_**, as expected by the code.

### Issues

* Fixes flutter/flutter#135194
* ~~May be the root cause of flutter/flutter#130147~~ (it isn't, I've seen the integration test failing again)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player_web/example/integration_test/video_player_test.dart
+++ b/packages/video_player/video_player_web/example/integration_test/video_player_test.dart
@@ -31,8 +31,11 @@ void main() {
 
       expect(video.controls, isFalse,
           reason: 'Video is controlled through code');
-      expect(video.getAttribute('autoplay'), 'false',
-          reason: 'Cannot autoplay on the web');
+      expect(video.autoplay, isFalse,
+          reason: 'autoplay attribute on HTMLVideoElement MUST be false');
+      // see: https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML
+      expect(video.getAttribute('autoplay'), isNull,
+          reason: 'autoplay attribute on video tag must NOT be set');
       expect(video.getAttribute('playsinline'), 'true',
           reason: 'Needed by safari iOS');
     });
--- a/packages/video_player/video_player_web/lib/src/video_player.dart
+++ b/packages/video_player/video_player_web/lib/src/video_player.dart
@@ -63,11 +63,12 @@ class VideoPlayer {
       ..autoplay = false
       ..controls = false;
 
-    // Allows Safari iOS to play the video inline
+    // Allows Safari iOS to play the video inline.
+    //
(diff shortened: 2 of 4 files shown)


packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -66,8 +66,8 @@ class VideoPlayer {
     // Allows Safari iOS to play the video inline
     _videoElement.setAttribute('playsinline', 'true');
 
-    // Set autoplay to false since most browsers won't autoplay a video unless it is muted
-    _videoElement.setAttribute('autoplay', 'false');
+    // Ensure autoplay is not set through the DOM!
```
stuartmorgan, line 69: Does this mean we should also remove line 63, or is that doing something different? I'm not clear on what exactly the HTML wrapper class properties do.

packages/video_player/video_player_web/example/integration_test/video_player_test.dart
```diff
@@ -31,8 +31,11 @@ void main() {
 
       expect(video.controls, isFalse,
           reason: 'Video is controlled through code');
-      expect(video.getAttribute('autoplay'), 'false',
-          reason: 'Cannot autoplay on the web');
+      expect(video.autoplay, isFalse,
+          reason: 'autoplay attribute on HTMLVideoElement MUST be false');
+      // see: https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML
```
stuartmorgan, line 36: Ugh, I had forgotten about this.

packages/video_player/video_player_web/lib/src/video_player.dart
ditman: Let me vanilla-JS this bit, I'm not sure if the DOM attribute is synced with the internal state of the element.

ditman: At least in chrome, they're kept in sync (I'm going to assume this is standard behavior), so we could do either:

Screenshot 2023-09-21 at 11 39 42 AM

ditman: I removed the `removeAttribute` call, which is redundant with line 63. Left the check in the tests, which should pass anyway. I also documented why we need to do setAttribute on the "playsinline" property.

packages/video_player/video_player_web/example/integration_test/video_player_test.dart
ditman: Yeah, this is a doozy of a feature :)

stuartmorgan (approved):
LGTM!

PR #4982, opened 2023-09-22
[video_player] Add macOS support

stuartmorgan:
Adds macOS support to `video_player`, sharing almost all of the code with iOS.

Notes about changes at a high level:
- macOS does not have `CADisplayLink` (prior to 14, and even there without all the functionality we need), so this adds macOS compilation branches that use the lower-level `CVDisplayLink` instead. Per the TODO, this code should be extracted later to reduce `ifdef`s in what is already a complicated file.
- Adds KVO unregistration on `dealloc` if it wasn't done in `dispose`, since unit tests were crashing on macOS with that.
- Temporarily ifdef's out `publish:` for macOS, with a TODO to re-enable it after the next stable.

Most of https://github.com/flutter/flutter/issues/41688
Once this lands, the app-facing package will be updated to endorse it for macOS.

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
The podspec check failure is because the Flutter Pod we lint against is one that we manually publish occasionally just for this, and it hasn't been updated since those APIs were added. I'll need to dig up the instructions on updating it.

stuartmorgan:
> Includes a `pigeon` update since the version currently being used wasn't generating Obj-C files that compile on macOS.
> * This could easily be moved to a prequel PR if that would be preferred.

I realized it would be better to do [a prequel PR for prefixes](https://github.com/flutter/packages/pull/4994), so this will be removed once that lands.

stuartmorgan:
Tests should all be green now, so this is fully ready.

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
```diff
@@ -53,18 +80,33 @@ @interface FVPVideoPlayer : NSObject <FlutterTexture, FlutterStreamHandler>
 // An invisible AVPlayerLayer is used to overwrite the protection of pixel buffers in those streams
 // for issue #1, and restore the correct width and height for issue #2.
 @property(readonly, nonatomic) AVPlayerLayer *playerLayer;
```
tarrinneal, line 82: could this go under the endif?

tarrinneal (approved):
TPMITIDF,BIDKETFTH.

Seems fine.

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
stuartmorgan: I'll move the ifdef'd properties to the end of the property list, similar to what we do with includes.

stuartmorgan:
> Hm, I thought that script change would fix the check_podspecs failure. It looks like I'll need to find another way to force it to get the right version.

And now it's fine even without the cache clearing 🤷🏻  Maybe just a general CDN propagation issue.

cbracken (approved):
This looks reasonable. This is a bit on the border of `ifdef`s where it might be worth considering macOS/iOS subclasses for the conditionally compiled logic but it's still a relatively small/innocuous amount overall, so seems fine as is.

AlexV525:
> And now it's fine even without the cache clearing 🤷🏻 Maybe just a general CDN propagation issue.

@stuartmorgan I ran into this in CIs, it was trying to find `ios/video_player_avfoundation.podspec` but got nothing.
https://github.com/fluttercandies/flutter_photo_manager/actions/runs/6374503740/job/17299430662?pr=997

PR #4994, opened 2023-09-25
[video_player] Update iOS prefixes

stuartmorgan:
Renames all classes and stand-alone functions to use the `FVP` prefix instead of `FLT`.

Because https://github.com/ollyde/flutter_macos_video_player/issues/9 was never fixed, there are projects in the wild using macOS classes with all of the same names as our classes, but with different implementation/behavior. If we add macOS support to the plugin as-is, developers using that package will end up with two copies of classes with the same name, which has undefined behavior at runtime in Obj-C (and in this case would very likely result in exceptions and/or crashes).

While a third-party should never have published code using the Flutter team's code prefixes, since we wanted to change prefixes anyway we can do so before adding macOS support and just avoid the issue entirely.

Since this involved regenerating Pigeon anyway, this also includes a Pigeon update.

Part of https://github.com/flutter/flutter/issues/102601

## Pre-launch Checklist

- [x] I signed the [CLA].


tarrinneal (approved):
LASOL

PR #4995, opened 2023-09-25
Auto rerun once for top flaky packages targets

keyonghan:
Part of https://github.com/flutter/flutter/issues/135449

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -323,6 +323,7 @@ targets:
       channel: master
       version_file: flutter_master.version
       package_sharding: "--shardIndex 1 --shardCount 6"
+      presubmit_max_attempts: "2"
       dependencies: >-
         [
           {"dependency": "android_virtual_device", "version": "33"}
@@ -403,6 +404,7 @@ targets:
       channel: stable
       version_file: flutter_stable.version
       package_sharding: "--shardIndex 1 --shardCount 6"
+      presubmit_max_attempts: "2"
       dependencies: >-
         [
           {"dependency": "android_virtual_device", "version": "33"}


.ci.yaml
```diff
@@ -403,6 +404,7 @@ targets:
       channel: stable
       version_file: flutter_stable.version
       package_sharding: "--shardIndex 1 --shardCount 6"
+      presubmit_max_attempts: "2"
```
godofredoc, line 407: Can we add a tracking bug?

stuartmorgan (commented):
This will not really work as intended; the shards aren't stable since the packages run on a given shard depend on the packages changed in the PR.

Also, the kind of test that is flaking here already has its own internal retry logic. The problem is that a test in this shard has recently become extremely flaky; we're trying to isolate which one to disable it so that we don't have to turn off the entire plugin's tests. I don't think adding a second (less efficient) layer of retries is a good idea here.

keyonghan:
> This will not really work as intended; the shards aren't stable since the packages run on a given shard depend on the packages changed in the PR.
> 
> Also, the kind of test that is flaking here already has its own internal retry logic. The problem is that a test in this shard (when running all packages) has recently become extremely flaky; we're trying to isolate which one to disable it so that we don't have to turn off the entire plugin's tests. I don't think adding a second (less efficient) layer of retries is a good idea here.

Proposing this change mainly due to the annoying flake that happens on the `dependbot` dependency bumping PRs. For example https://github.com/flutter/packages/pull/4988. Infra team has to manually keep rerunning the failed/flaky targets and adding the autosubmit label until they pass.

Is there any bug tracking the fix/skip of the specific flaky tests in these flaky shards?

stuartmorgan:
> Is there any bug tracking the fix/skip of the specific flaky tests in these flaky shards?

https://github.com/flutter/flutter/issues/130986 is tracking the current bad flake.

PR #4951, opened 2023-09-19
[camera_android] Set buildconfig to true for compatibility with AGP 8.0+

gmackall:
Fixes https://github.com/flutter/flutter/issues/134818

See https://github.com/flutter/packages/pull/4535 for an example of when this was done for a different plugin.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #4997, opened 2023-09-25
[camera_android] Downgrade to AGP 7.3.0 to fix build_alll_packages test failures

gmackall:
Fixes the current failures of the build_all_packages tests that were caused in https://github.com/flutter/packages/pull/4951/files. 

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_android/CHANGELOG.md
+++ b/packages/camera/camera_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.8+11
+
+* Downgrade AGP version for compatibility with legacy projects.
+
 ## 0.10.8+10
 
 * Sets android.defaults.buildfeatures.buildconfig to true for compatibility with AGP 8.0+.
--- a/packages/camera/camera_android/CHANGELOG.md
+++ b/packages/camera/camera_android/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.10.8+11
 
-* Downgrade AGP version for compatibility with legacy projects.
+* Downgrades AGP version for compatibility with legacy projects.
 
 ## 0.10.8+10
 
--- a/packages/camera/camera_android/android/build.gradle
+++ b/packages/camera/camera_android/android/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.1.1'
+        classpath 'com.android.tools.build:gradle:7.3.0'
     }
 }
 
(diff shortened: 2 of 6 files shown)


camsim99 (approved):
Ah yes I ran into this on a PR and wondered if it was this. Unfortunate, trying to investigate the underlying issue.

gmackall:
> Ah yes I ran into this on a PR and wondered if it was this. Unfortunate, trying to investigate the underlying issue.

Yeah, the test in question that is failing is using AGP 4.1/ gradle 6.7/ java 11 - one of those probably isn't mixing well with using agp 8.1.1 in the plugin itself, though it would be good to know exactly which one and why

camsim99:
> Yeah, the test in question that is failing is using AGP 4.1/ gradle 6.7/ java 11 - one of those probably isn't mixing well with using agp 8.1.1 in the plugin itself, though it would be good to know exactly which one and why

Ah I see okay, thank you for this context!

gmackall:
I think the tests are still failing because we explicitly exclude `camera_android`: https://github.com/flutter/packages/blob/619af75f7966f90498dfce3a754d37422b972c6d/script/configs/exclude_all_packages_app.yaml#L17

I believe this effectively turns the build_all_packages test into a post submit test? Because we don't consume the changes in the PR, and they are then only included _after_ the pr gets merged

stuartmorgan:
> I believe this effectively turns the build_all_packages test into a post submit test for `camera_android` alone? Because we don't consume the changes in the PR, and they are then only included _after_ a pr that affects `camera_android` gets merged (and therefore a new version is published).

Oof. Yes, that sounds right; good find. I hadn't consider that when we set up the test this way to include `camerax` :(

What we should be able to do instead is remove that exclusion, and instead special-case the script to remove one of the two Android camera variants [here](https://github.com/flutter/packages/blob/main/script/tool/lib/src/create_all_packages_app_command.dart#L290) but not [here](https://github.com/flutter/packages/blob/main/script/tool/lib/src/create_all_packages_app_command.dart#L294). Special casing the script isn't ideal, but I think it's the best option since the generic exclusion has this problem.

gmackall:
> > I believe this effectively turns the build_all_packages test into a post submit test for `camera_android` alone? Because we don't consume the changes in the PR, and they are then only included _after_ a pr that affects `camera_android` gets merged (and therefore a new version is published).
> 
> Oof. Yes, that sounds right; good find. I hadn't consider that when we set up the test this way to include `camerax` :(
> 
> What we should be able to do instead is remove that exclusion, and instead special-case the script to remove one of the two Android camera variants [here](https://github.com/flutter/packages/blob/main/script/tool/lib/src/create_all_packages_app_command.dart#L290) but not [here](https://github.com/flutter/packages/blob/main/script/tool/lib/src/create_all_packages_app_command.dart#L294). Special casing the script isn't ideal, but I think it's the best option since the generic exclusion has this problem.

Got it, thanks - should I land this in the meantime, or do you think it's better to add that fix to this PR?

stuartmorgan:
I'll take a crack at adding it quickly now, so that we can verify in presubmit that everything is working as expected with that change.

stuartmorgan:
Build-all tasks have passed, so it looks like that worked.

PR #5005, opened 2023-09-26
Migrate Autopublishing to GCB

sealesj:
Initial testing configuration for autopublishing to use Google Cloud Build.

This change requires the pub.dev publishing to migrate into a Google Cloud Build configuration file which is triggered by a flutter GCP trigger. The tagging in the release.yml will continue to execute, and the tagging push will trigger the GCB yaml. 

In this state, the current autopublishing flow would still be enabled, but we could move forward with providing access to the flutter-infra GCP project to impersonate a service account for releasing

Addresses: https://github.com/flutter/flutter/issues/126827

## Pre-launch Checklist

- [x] I signed the [CLA].


sealesj:
@godofredoc would flutter-infra be an appropriate project to re-use for configuring a service account to use this yaml?

godofredoc:
> @godofredoc would flutter-infra be an appropriate project to re-use for configuring a service account to use this yaml?

No, we need to create a new project for this.

sealesj:
> No, we need to create a new project for this.
Ok, the changes should be pretty straightforward on the project, but I don't have the necessary permissions to make a new GCP project in the flutter parent organization

cloudbuild.yml
```diff
@@ -0,0 +1,24 @@
...
+  - PROJECT_ID=$PROJECT_ID
+- id: Publish to pub.dev
+  name: dart
+  volumes:
+  - name: temporary-secrets
+    path: /secrets
+  # TODO @jseales - replace --dry-run on dart command with --force
+  script: | 
+    cat /secrets/temporary-pub-token.txt | dart pub token add https://pub.dev
+    dart pub publish --dry-run
```
stuartmorgan, line 24: Can you describe the chain of events involved for this flow, in particular where this script is run (not our CI I assume?) and how the environment on that machine is configured? What would cause this to be in the correct directory?

sealesj: Thank you for those questions. I was using this starting cloudbuild configuration to test the connection with the existing github action. My thought is that the existing github action will continue executing all the steps up to the dart publish, and that the tag push will trigger the GCB yaml.

From then I am thinking in the new GCB yaml configuring a new service account with permission to publish to pub which is authenticated in the yaml would allow pub publishing. 

This yaml would only need the gcloud and dart cloud builders to run on google cloud. 
However, I am not sure for the directory question and need to continue testing

sealesj: I've tested using a new GCP project. The connection is working, and it seems like the `pub publish` command in the tool workspace is working as well. The remaining step would be to now allow publishing on the pub side for each package. 

The `dry-run` seems to be working as expected, also showing failures since these release were not tagged which is expected, and this workflow would trigger only after the tag has already been pushed by the existing `release.yml` workflow.
Screenshot 2023-10-12 at 3 56 53 PM

From documentation I see that invoking this tool to publish is deprecated. Would using this tool to execute pub publish be acceptable here, or would you prefer a refactoring of the publishing so that it is invoked in a different way?

stuartmorgan: Now that you have it wired up, could you walk me through the flow (i.e., my comment above)? Once we tag a plugin, what is the exact chain of events? Currently I know how everything works in our release system (e.g., we are directly controlling the version of Dart being used, our tooling is controlling the checkout, directory, etc. management), and I'd like to have an understanding (and ideally documentation in the wiki) of exactly what is happening in the new system, both so that I can reason about its correctness (it's important that the checkout have the hash corresponding to the tag, for instance; what manages that?) and so that we know what to do if something goes wrong since it will no longer be directly under repo control.

stuartmorgan: > From documentation I see that invoking this tool to publish is deprecated. Would using this tool to execute pub publish be acceptable here, or would you prefer a refactoring of the publishing so that it is invoked in a different way?

In this context using `pub publish` directly should be fine, since our tooling will be controlling the tagging process.

stuartmorgan: Although actually... we may need a wrapper still. We need to catch warnings, and in some cases bypass them. In which case I'll need to add some CI-only flags to the repo tooling to allow using that.

sealesj: > I'm not sure I follow this part; running `pub publish` in `script/tool` wouldn't do anything. It would have to be run in each relevant package.

Ah I'm sorry, I have been using [this tool](https://github.com/sealesj/packages/tree/release-cloudbuild/script/tool#publish-a-release) to publish in the workflow rather than calling `pub publish` directly, and passing it flags to skip the tagging. In the output it seemed like it is running for all packages which have changes to be released. Am I understanding that tool correctly? Or would a different process be needed to execute the `pub publish` for each package?

stuartmorgan: > Ah I'm sorry, I have been using [this tool](https://github.com/sealesj/packages/tree/release-cloudbuild/script/tool#publish-a-release) to publish in the workflow rather than calling `pub publish` directly, and passing it flags to skip the tagging.

Oh, I see the updated version of the script now; it was calling `pub publish` directly last time I had looked.

We don't have a flag for that in the current version (not with actual publishing), so I'll need to re-add it, but that is indeed how I would want that to work.

> In the output it seemed like it is running for all packages which have changes to be released. Am I understanding that tool correctly? Or would a different process be needed to execute the `pub publish` for each package?

Ideally we wouldn't use change detection, as that adds an additional layer of complexity; the change detection would already have been done with the script that did tagging. Do we have access to the list of triggering packages in the context of that script, so that we can pass them to `--packages`?

stuartmorgan: > that would assume that the triggering tag would always follow a pattern to parse out like `[PACKAGE]-v*.*`. Is that a safe assumption to make?

Yes, that's safe. We control the tag naming, and that's explicitly the format we use; we wouldn't change it without a compelling reason (and we can make a note that doing so would require changes to autopublish).

It sounds like passing a `--packages` argument should be pretty straightforward then. I'll just need to think about how to handle federated plugins since there are subtleties around how that flag interacts with federation. We may need a slightly different flag for this use case, but that's doable.

stuartmorgan: > For the federated plugins, would the change need to be made within the publishing tool? Or is this something that could be handled with flags passed to that tool in the case of a federated plugin?

I need to adjust the repo tooling; there isn't currently a flag that will do what we need. Currently passing something like `--packages=webview_flutter` would try to publish `webview_flutter/webview_flutter*`, and we need to selectively disable that behavior for this use case.

stuartmorgan: So to summarize, I think the repo tooling changes we need are:
- a flag for `publish` to tag without publishing
- a flag for `publish` to publish without tagging (maybe validating that the tag exists already, and for the current hash)
- a flag to disable group-name support for `--package`

sealesj: The run results are shown on commits which are tagged (such as this: https://github.com/sealesj/packages/runs/17791663333) but yes, that would mean separating the tagging and pushing

stuartmorgan: > The run results are shown on commits which are tagged

Where is that surfaced though? E.g., if I'm the gardener for this repo when publishing fails, where/how will I be alerted to that?

sealesj: I don't believe it's currently being surfaced anywhere besides being shown in the github commit checks, but we could create a simple google cloud alert with a policy tied to this trigger.

Since it is a postsubmit action not running on github CI, I'm not sure if there's another github-integrated option to alert.

sealesj: > How will we get notification of run results of the publish step? It occurs to me that an unavoidable downside to this flow is that if we fail a server-side `pub publish` check it will only happen in that final `cloudbuild.yaml` step (which will be after tagging, which will be unfortunate when it happens...)

It's possible we could remove both tagging and publishing logic from the github action and keep it to wait on tests before triggering the Google Cloud Build using a method other than tagging. Would that be an approach you'd be interested in?

stuartmorgan: Having tagging be the trigger for the cloud build solves a problem we have now that the only way to retroactively fix a PR that had unrelated OOB breakage is for someone to release manually, which we want to eliminate entirely, so the potential for post-tag failure is the lesser of two evils.

sealesj: Ok that's great to hear that it solves that problem! I wasn't aware of that context.

I'm going to test an email notification for this trigger. After that is working, I'll look at moving forward with a non-testing project

sealesj: Would an alert like this configured to the appropriate email address be sufficient to monitor failures?

Screenshot 2023-10-26 at 11 35 59 AM

If so, I can go ahead with creating the new Google Cloud project.

stuartmorgan: Having that emailed to flutter-ecosystem-team sounds good.

What's the resolution mechanism for transient publish failures (e.g., a temporary pub.dev server outage); can we just re-try the run like we would a flaky test?

sealesj: Yes we could keep allow the build to accept manual triggers as well, and then re-trigger manually in the case of an exception

sealesj: > So to summarize, I think the repo tooling changes we need are:
> 
> * a flag for `publish` to tag without publishing

Could using `pub-publish-flags=--dry-run` with the current `publish` be a solution using existing tooling? https://github.com/flutter/packages/blob/753bc920133466b24d45e1e956a975f6e67d0ee0/.github/workflows/release.yml#L63 it looks like that achieves tag without publishing https://github.com/flutter/packages/blob/753bc920133466b24d45e1e956a975f6e67d0ee0/script/tool/lib/src/publish_command.dart#L81 

> * a flag for `publish` to publish without tagging (maybe validating that the tag exists already, and for the current hash)

Why would publish without tagging be necessary? My understanding was that the ability to publish from anything other than cloudbuild would need to be removed.

stuartmorgan: > Could using `pub-publish-flags=--dry-run` with the current `publish` be a solution using existing tooling?

It would work, but it's pretty wasteful since we already run `publish --dry-run` as part of our standard test suite, so this would pointlessly duplicate work in CI.

> Why would publish without tagging be necessary?

See https://github.com/flutter/packages/pull/5005#discussion_r1358762032

sealesj: @stuartmorgan Thank you for such great conversation through this issue! Could I help contribute to the necessary repo tooling changes? 

From the GCP side, the configuration is good to go with the exception of security hardening after the workflow is set up. Those changes are: 
- requiring approval before a build executes if it is triggered manually
- disable ability to trigger with exceptions for something like members of an access on demand group

And there is also the pending task of allowing the GCP service account to be a publisher to each of the pub.dev packages.

stuartmorgan: > Could I help contribute to the necessary repo tooling changes?

Sure; they are on my to-do list, but if you have cycles to get to any of it before I do that would be great, and I'm happy to review.

stuartmorgan:
Update from triage: this is blocked on repo tooling work. I hope to spend some time on that this month.

cloudbuild.yml
stuartmorgan: > So to summarize, I think the repo tooling changes we need are:
> 
> * a flag for `publish` to tag without publishing
> * a flag for `publish` to publish without tagging (maybe validating that the tag exists already, and for the current hash)
> * a flag to disable group-name support for `--package`

https://github.com/flutter/packages/pull/6218 will add all of these.

sealesj: Thank you for this update!

stuartmorgan:
@sealesj What's the next step here now that the tool support has landed?

sealesj:
@stuartmorgan Thank you for following up. I've added this to my to-do list and will be working on it soon as a P2 task

stuartmorgan:
From triage: Is this still in development, or was it impacted by team changes?

christopherfujino:
> From triage: Is this still in development, or was it impacted by team changes?

TIL about this. Gonna close this as I don't plan to implement it. Feel free to re-open if you think there's value and something we should work on this year.

PR #5008, opened 2023-09-26
[pigeon] FlutterApi error handling

tarrinneal:
* Adds error handling on Flutter Api methods.
* **Breaking Change** Removes `Reply` class from all Java method returns.
* **Breaking Change** Adds `NullableResult` class for all nullable Java method returns.
* **Breaking Change** Nests all enum returns in Objective-c to allow for `nil` response on error.
* **Breaking Change** Renames `Setup` to `SetUp` in Objective-c.

fixes https://github.com/flutter/flutter/issues/118243
fixes https://github.com/flutter/flutter/issues/118245
fixes https://github.com/flutter/flutter/issues/124268
fixes https://github.com/flutter/flutter/issues/135176

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Since we have breaking changes to Java in this PR we should do the `setup`->`setUp` fix in Java too.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 12.0.0
+
+* Adds error handling on Flutter Api methods.
+* **Breaking Change** Removes `Reply` class from all Java method returns.
```
stuartmorgan, line 4: Did we stop using the `[language]` tags on release notes on purpose? I only just noticed we haven't been doing it recently; it seems convenient for people skimming for things that do or don't apply to their use cases.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
...
+* **Breaking Change** Adds `NullableResult` class for all nullable Java method returns.
```
stuartmorgan, line 5: We should explain the transition here too.

packages/pigeon/example/app/android/app/src/main/java/io/flutter/plugins/Messages.java
```diff
@@ -181,8 +182,13 @@ ArrayList<Object> toList() {
   }
 
   public interface Result<T> {
```
stuartmorgan, line 184: Let's add doc comments to these two result classes and their methods (in the generator, obviously).

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 12.0.0
+
+* Adds error handling on Flutter Api methods.
```
stuartmorgan, line 3: Nit: API

stuartmorgan, line 3: Aren't there breaking changes to at least Swift and Kotlin here that aren't mentioned?

packages/pigeon/example/app/macos/Runner/messages.g.m
```diff
@@ -219,10 +219,32 @@ - (void)flutterMethodAString:(nullable NSString *)arg_aString
...
+                  completion(nil, [FlutterError errorWithCode:reply[0]
+                                                      message:reply[1]
+                                                      details:reply[2]]);
+                } else if (reply[0] == nil) {
+                  completion(
+                      nil,
+                      [FlutterError
+                          errorWithCode:@"null-error"
+                                message:
+                                    @"Flutter api returned null value for non-null return value."
```
stuartmorgan, line 236: This case can't happen without a bug in our Dart generator (with any reasonably current version of Dart), so we don't need runtime handling for it.

packages/pigeon/example/app/macos/Runner/messages.g.m
```diff
@@ -219,10 +219,32 @@ - (void)flutterMethodAString:(nullable NSString *)arg_aString
...
                        codec:PGNMessageFlutterApiGetCodec()];
-  [channel sendMessage:@[ arg_aString ?: [NSNull null] ]
-                 reply:^(id reply) {
-                   NSString *output = reply;
-                   completion(output, nil);
-                 }];
+  [channel
+      sendMessage:@[ arg_aString ?: [NSNull null] ]
+            reply:^(NSArray<id> *reply) {
+              if (reply != nil) {
```
stuartmorgan, line 225: Can we extract this entire block to a static helper method (which we write conditionally if any FlutterAPI methods exist, to avoid unused method compilation issues)?

packages/pigeon/example/app/windows/runner/messages.g.cpp
```diff
@@ -271,17 +271,31 @@ void MessageFlutterApi::FlutterMethod(
...
-      [on_success = std::move(on_success), on_error = std::move(on_error)](
-          const uint8_t* reply, size_t reply_size) {
-        std::unique_ptr<EncodableValue> response =
-            GetCodec().DecodeMessage(reply, reply_size);
-        const auto& encodable_return_value = *response;
+  channel->Send(encoded_api_arguments, [on_success = std::move(on_success),
+                                        on_error = std::move(on_error)](
+                                           const uint8_t* reply,
+                                           size_t reply_size) {
+    std::unique_ptr<EncodableValue> response =
```
stuartmorgan, line 278: I think we could do the same (extract the entire implementation of the lambda to a static helper method) if we make it templated in the return value type and then add template specializations for special cases like enums. But we can leave that to a follow-up.

packages/pigeon/example/app/windows/runner/messages.g.cpp
```diff
@@ -271,17 +271,31 @@ void MessageFlutterApi::FlutterMethod(
...
+        GetCodec().DecodeMessage(reply, reply_size);
+    const auto& encodable_return_value = *response;
+    const auto& list_return_value =
```
stuartmorgan, line 281: This is a pointer, so should be `const auto*`.

packages/pigeon/example/app/macos/Runner/messages.g.m
stuartmorgan: Oh, the extraction part gets complicated if we do that though, right? In which case we can punt on that.

packages/pigeon/lib/java_generator.dart
```diff
@@ -508,30 +500,52 @@ class JavaGenerator extends StructuredGenerator<JavaOptions> {
...
+                        indent.writeln(
+                            '$returnType $output = listReply.get(0) == null ? null : $returnType.values()[(int) listReply.get(0)];');
+                      } else {
+                        indent.writeln(
+                            '$returnType $output = $returnType.values()[(int) listReply.get(0)];');
+                      }
+                    } else {
+                      indent.writeln(
+                          '$returnType $output = ${_cast('listReply.get(0)', javaType: returnType)};');
+                    }
```
stuartmorgan, line 539: It's hard to see/verify that every case here is writing a `$returnType $output =` expression here; I would recommend:

```
final String outputExpression;
// all the same logic here, but instead of `indent.writeln('$returnType $output = ...');`
// do `outputExpression = '...';
// then after all the if/else-if chains:
indent.writeln('$returnType $output = $outputExpression');
```

Then it's easier to see what is the same vs different about the output in each case, and it also won't compile if anyone ever accidentally introduces a path that doesn't set `outputExpression`.

packages/pigeon/lib/objc_generator.dart
```diff
@@ -413,7 +413,7 @@ class ObjcHeaderGenerator extends StructuredGenerator<ObjcOptions> {
     indent.writeln('@end');
     indent.newln();
     indent.writeln(
-        'extern void ${apiName}Setup(id<FlutterBinaryMessenger> binaryMessenger, NSObject<$apiName> *_Nullable api);');
+        'extern void ${apiName}SetUp(id<FlutterBinaryMessenger> binaryMessenger, NSObject<$apiName> *_Nullable api);');
```
stuartmorgan, line 416: It would actually be better as `SetUp${apiName}`, since the function should read as a verb phrase. Sorry, I didn't think about that when I mentioned it the other day.

packages/pigeon/lib/objc_generator.dart
```diff
@@ -1104,26 +1094,42 @@ static id GetNullableObjectAtIndex(NSArray *array, NSInteger key) {
...
+                'reply[0] == [NSNull null] ? nil : reply[0]';
+            if (func.returnType.isVoid) {
+              indent.writeln('completion(nil);');
             } else {
-              indent.writeln(
-                  '${returnType.baseName} output = [reply integerValue];');
+              if (isEnum(root, func.returnType)) {
+                indent.writeln('NSNumber *outputAsNumber = $nullCheck;');
+                indent.writeln(
+                    '${_enumName(returnType.baseName, suffix: ' *', prefix: languageOptions.prefix, box: true)}output = outputAsNumber == nil ? nil : [[${_enumName(returnType.baseName, prefix: languageOptions.prefix, box: true)} alloc] initWithValue: [outputAsNumber integerValue]];');
```
stuartmorgan, line 1121: Nit: remove the space after the : in `initWithValue:`.

stuartmorgan, line 1121: This line is a lot; maybe extract the enum name to a local? That would also let you re-use it, and add the * in the string, rather than call twice.

stuartmorgan, line 1121: Doesn't the `[outputAsNumber integerValue]` need to be cast to the enum type to avoid warnings?

packages/pigeon/lib/swift_generator.dart
```diff
@@ -319,15 +319,15 @@ import FlutterMacOS
...
+            ? 'Void'
             : _nullsafeSwiftTypeForDartType(func.returnType);
         String sendArgument;
         addDocumentationComments(
             indent, func.documentationComments, _docCommentSpec);
 
         if (func.arguments.isEmpty) {
           indent.write(
-              'func ${func.name}(completion: @escaping ($returnType) -> Void) ');
+              'func ${func.name}(completion: @escaping (Result<$returnType, Error>) -> Void) ');
```
stuartmorgan, line 330: Shouldn't it be `FlutterError` rather than `Error`?

stuartmorgan (changes requested):
All the new generated boilerplate makes me sad, but we can tackle that later. It great to have this plumbed correctly everywhere!

packages/pigeon/lib/objc_generator.dart
tarrinneal: I don't think so, they're freely convertible from my understanding. 🤷

packages/pigeon/CHANGELOG.md
```diff
@@ -1,13 +1,26 @@
+## 12.0.0
+
+* Adds error handling on Flutter API methods.
+* **Breaking Change** [kotlin] Flutter API methods now return `Result<return-type>`.
+* **Breaking Change** [swift] Flutter API methods now return `Result<return-type, FlutterError>`.
+* **Breaking Change** [java] Removes `Reply` class from all method returns and replaces it with `Result`.
+  Changes required: Replace all `Reply` callbacks with `Result` classes that contain both `success` and `failure` methods.
```
stuartmorgan, line 7: If there are intended to be formatted as nested, as they are in the raw text, I believe they will need to be sub-bullets.

packages/pigeon/lib/java_generator.dart
```diff
@@ -508,30 +500,54 @@ class JavaGenerator extends StructuredGenerator<JavaOptions> {
...
+                  }, addTrailingNewline: false);
+                }
+                indent.addScoped('else {', '}', () {
+                  if (func.returnType.isVoid) {
+                    indent.writeln('result.success(null);');
                   } else {
-                    indent.writeln(
-                        '$returnType $output = $returnType.values()[(int) channelReply];');
+                    const String output = 'output';
+                    String outputExpression;
```
stuartmorgan, line 524: Can't this be final? I believe Dart allows it as long as every codepaths sets it exactly once.

stuartmorgan (approved):
LGTM!

packages/pigeon/lib/java_generator.dart
tarrinneal: 🤯

PR #5009, opened 2023-09-26
[image_picker_android] Refactor getting of paths from intent to single helper

gmackall:
Fixes https://github.com/flutter/flutter/issues/104168

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -629,24 +629,48 @@ public boolean onActivityResult(
...
+  private ArrayList<MediaPath> getPathsFromIntent(Intent data, boolean includeMimeType) {
+    ArrayList<MediaPath> paths = new ArrayList<>();
+
+    Uri uri = data.getData();
+    // On several pre-Android 13 devices using Android Photo Picker, the Uri from getData() could
+    // be null.
+    if (uri == null) {
+      ClipData clipData = data.getClipData();
+
+      // If data.getData() and data.getClipData() are both null, we are in an error state. By
```
gmackall, line 641: Open to comments as to whether this is a reasonable way to handle this error. I chose to return null here and then report the error from the respective handler so that the error message could correspond to the action, but it may be better to make them all just one error.

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -716,22 +733,14 @@ private void handleChooseMultiImageResult(int resultCode, Intent intent) {
...
       // If there's no valid Uri, return an error
-      if (uri == null) {
+      if (paths == null || paths.size() < 1) {
         finishWithError("no_valid_video_uri", "Cannot find the selected video.");
         return;
       }
 
-      String path = fileUtils.getPathFromUri(activity, uri);
-      handleVideoResult(path);
+      handleVideoResult(paths.get(0).path);
```
gmackall, line 743: This part is the one piece that is a little awkward from forcing the path getting into one common method. It could also probably be fixed by making changes to `handleVideoResult`, though

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
aldee: I think it's better this way, so that the caller of this method will handle the error that will be thrown to the user, since the intents of each callers are different (single asset/media pick vs multiple ones should have different error)

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -629,24 +629,48 @@ public boolean onActivityResult(
...
+      // handler.
+      if (clipData == null) {
+        return null;
+      }
+
+      for (int i = 0; i < data.getClipData().getItemCount(); i++) {
+        uri = data.getClipData().getItemAt(i).getUri();
+        // Same error state as above.
+        if (uri == null) {
+          return null;
```
aldee, line 652: I don't think this should return null here. what if there are multiple items inside of the ClipData, and the first few have valid Uri, but at a certain index it doesn't? In that case, the paths variable should already have values inside it

gmackall: I was on the fence because of that case - but at the same time I wouldn't want to silently return just the strict subset of images that we successfully got uris for. 

cc @tarrinneal as the overall owner of the plugin, do you have any preferences here on the behavior of the plugin for the case of failing to get the uri for one of the images in a multi-image choice action? I think currently it would crash if that were to happen

tarrinneal: I think it would make more sense to throw and handle errors across these methods. Perhaps @stuartmorgan would also like to weigh in.

stuartmorgan: I would also vote for throwing an error in that case rather than silently failing (or silently dropping some paths).

gmackall:
Closing for now, I still want to land this but haven't been able to work on it for a bit and won't be able to for a bit as well

stepanov-standy:
@gmackall when this fix will be available?

gmackall:
I came back to this as I remembered I want to reland this - looking back over the comments, I'm not sure why I closed it. The consensus seems to be for throwing errors across the methods, which is what the current behavior does.

I'm going to re-open and request review based on that.

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
stuartmorgan: It should actually be trivial to avoid this, because I already standardized on returning lists for everything during the Pigeon migration. `handleVideoResult` can be removed, and you can just call `finishWithListSuccess` here, and change the one other use to `finishWithSuccess` (which is all `handleVideoResult` does now).

stuartmorgan (approved):
Minor comments, but LGTM

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -707,17 +732,14 @@ private void handleChooseMediaResult(int resultCode, Intent intent) {
...
-                  fileUtils.getPathFromUri(activity, intent.getClipData().getItemAt(i).getUri()),
-                  null));
-        }
-      } else {
-        paths.add(new MediaPath(fileUtils.getPathFromUri(activity, intent.getData()), null));
+      ArrayList<MediaPath> paths = getPathsFromIntent(intent, false);
+      // If there's no valid Uri, return an error
+      if (paths == null) {
+        finishWithError(
+            "missing_valid_image_uri", "Cannot find at least one " + "of the selected images.");
```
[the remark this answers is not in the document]
gmackall: Yeah I think I put a line break but then the formatter disagreed with me so it got removed.

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
gmackall: `finishWithListSuccess` takes a list of strings, and we have a list of `MediaPath`s, so I just called `finishWithSuccess` directly in both places, and then removed `handleVideoResult`.

PR #4477, opened 2023-07-14
[url_launcher_android] Enable Android WebView to open PDFs and non-HTTP URLs (`mailto:`, `tel:`, `sms:`, etc.)

michaelmiles1:
This PR adds the ability of the launched WebView in Android to follow and open non-HTTP links by starting new Intents, such as opening a PDF viewer or your preferred email application.

Issues fixed: 
[#106331](https://github.com/flutter/flutter/issues/106331) - this fix was inspired by the attempted fix [here](https://github.com/flutter/plugins/pull/6006), but I had to extend the `WebViewClient` to make it testable
[#27309](https://github.com/flutter/flutter/issues/27309) - opening PDFs wasn't causing a crash for me, but was not opening PDFs on any Android device

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [ ] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[...]

PR #5017, opened 2023-09-27
[url_launcher_android] Enable Android WebView to open PDFs and non-HTTP URLs (mailto:, tel:, sms:, etc.)

mobileappmike:
This PR adds the ability of the launched WebView in Android to follow and open non-HTTP links by starting new Intents, such as opening a PDF viewer or your preferred email application.

Issues fixed:
https://github.com/flutter/flutter/issues/106331 - this fix was inspired by the attempted fix https://github.com/flutter/plugins/pull/6006, but I had to extend the WebViewClient to make it testable
https://github.com/flutter/flutter/issues/27309 - opening PDFs wasn't causing a crash for me, but was not opening PDFs on any Android device

Had issues with the last PR because of the CLA. Here's the last PR: https://github.com/flutter/packages/pull/4477

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/url_launcher/url_launcher_android/example/lib/main.dart
```diff
@@ -182,45 +189,52 @@ class _MyHomePageState extends State<MyHomePage> {
...
-                child: Text(toLaunch),
+              Padding(
+                padding: const EdgeInsets.all(16.0),
+                child: TextField(
+                  controller: _urlTextController,
+                  decoration: const InputDecoration(
+                    labelText: 'URL to Launch',
+                    hintText: 'Input the URL to launch',
+                  ),
+                ),
```
michaelmiles1, line 200: The test page now launches any URL the user inputs

packages/url_launcher/url_launcher_android/android/src/test/java/io/flutter/plugins/urllauncher/WebViewActivityTest.java
```diff
@@ -5,13 +5,124 @@
...
+  }
+
+  @Test
+  public void onLoadResource_loadsPdf() {
+    String url = "https://www.flutter.dev/test.pdf";
+    webView.getWebViewClient().onLoadResource(webView, url);
+
+    verify(context, times(1)).startActivity(any(Intent.class));
+    assertTrue(FlutterWebViewClient.resourceShouldOpenDocument(webView, url));
+  }
```
michaelmiles1, line 127: Unit tests to make sure that the URL launcher parses the URL properly based on scenario (.pdf, :mailto, etc)

michaelmiles1:
Will look in to and fix build issues tomorrow 👍

gmackall:
Sorry for the delay on reviewing, I was traveling for the last week. I'll make sure to get this reviewed either today or tomorrow!

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/FlutterWebViewClient.java
```diff
@@ -0,0 +1,79 @@
...
+      view.getContext().startActivity(intent);
+      return true;
+    }
+    return false;
+  }
+
+  public static boolean urlShouldRunActivity(@NonNull WebView view, @NonNull String url) {
+    // Check if URL is not an HTTP(S) request
+    // Handles mailto:, sms:, etc.
+    if (!url.startsWith("http://") && !url.startsWith("https://")) {
```
gmackall, line 70: Can you change these checks, and the one above for `.pdf` to be non case sensitive?

gmackall (changes requested):
Mostly LGTM, just a couple of comments!

stuartmorgan:
@gmackall Are you okay with unilaterally enabling users to leave the app even when the plugin client has requested the in-app webview mode? This seems like something that clients might want to be able to control, either with an opt-in or an opt-out.

michaelmiles1:
Resolved issues, thanks!

gmackall:
> @gmackall Are you okay with unilaterally enabling users to leave the app even when the plugin client has requested the in-app webview mode? This seems like something that clients might want to be able to control, either with an opt-in or an opt-out.

I hadn't considered that, but I agree it does seem likely clients would want control over that. Have the other platform implementations solved into a similar problem?

stuartmorgan:
> Have the other platform implementations solved into a similar problem?

So far only iOS supports a somewhat similar mode, and it's more akin to Android Custom Tabs where we don't have much control over how things work. We could make this a flag in [`WebViewConfiguration`](https://github.com/flutter/packages/blob/main/packages/url_launcher/url_launcher/lib/src/types.dart#L27-L54)

stuartmorgan:
(The other option is that we could land this as-is, and then add configuration only if people request it. In general, we don't want to support every possible configuration option otherwise `url_launcher` will be as complicated as `webview_flutter`; this is just potentially a special case since it used to be restricted and now won't be.)

gmackall:
> (The other option is that we could land this as-is, and then add configuration only if people request it. In general, we don't want to support every possible configuration option otherwise `url_launcher` will be as complicated as `webview_flutter`; this is just potentially a special case since it used to be restricted and now won't be.)

I'm okay with landing this as is and gauging feedback. Do you want to do secondary review as the overall url_launcher codeowner?

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/FlutterWebViewClient.java
```diff
@@ -0,0 +1,82 @@
...
+public class FlutterWebViewClient extends WebViewClient {
+  @Override
+  public void onLoadResource(WebView view, String url) {
+    if (!resourceShouldOpenDocument(view, url)) {
+      super.onLoadResource(view, url);
+    }
+  }
+
+  /*
+   * This method is deprecated in API 24. Still overridden to support
```
gmackall, line 26: nit: 
```suggestion
   * This method is deprecated in API 24. It is still overridden to support
```

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/FlutterWebViewClient.java
```diff
@@ -0,0 +1,82 @@
...
+      intent.setDataAndType(Uri.parse(url), "application/pdf");
+      view.getContext().startActivity(intent);
+      return true;
+    }
+    return false;
+  }
+
+  @VisibleForTesting
+  public static boolean urlShouldRunActivity(@NonNull WebView view, @NonNull String url) {
+    // Check if URL is not an HTTP(S) request
```
gmackall, line 71: nit: end with period

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/FlutterWebViewClient.java
```diff
@@ -0,0 +1,82 @@
...
+    // Handles mailto:, sms:, etc.
+    if (!url.toLowerCase().startsWith("http://") && !url.toLowerCase().startsWith("https://")) {
+      Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+      view.getContext().startActivity(intent);
+      return true;
+    }
+
+    // Otherwise, let WebView load URL
```
gmackall, line 79: nit: end with period

gmackall (approved):

michaelmiles1:
[ ] upperCase and lowerCase methods must pass in locale (lint)

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/FlutterWebViewClient.java
```diff
@@ -0,0 +1,84 @@
...
+   * This method is deprecated in API 24. It is still overridden to support
+   * earlier Android versions.
+   */
+  @SuppressWarnings("deprecation")
+  @Override
+  public boolean shouldOverrideUrlLoading(WebView view, String url) {
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+      if (urlShouldRunActivity(view, url)) {
+        return true;
+      } else {
```
stuartmorgan, line 36: `else` isn't necessary after an early return.

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/FlutterWebViewClient.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @RequiresApi(Build.VERSION_CODES.N)
+  @Override
+  public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+      String url = request.getUrl().toString();
+      if (urlShouldRunActivity(view, url)) {
+        return true;
+      } else {
```
stuartmorgan, line 51: Same.

packages/url_launcher/url_launcher_android/android/src/main/java/io/flutter/plugins/urllauncher/FlutterWebViewClient.java
```diff
@@ -0,0 +1,84 @@
...
+        view.loadUrl(url);
+      }
+    }
+    return false;
+  }
+
+  @VisibleForTesting
+  public static boolean resourceShouldOpenDocument(@NonNull WebView view, @NonNull String url) {
+    // Check if URL is PDF
+    if (url.toLowerCase(Locale.US).endsWith(".pdf")) {
```
stuartmorgan, line 61: Why is this using the URL string rather than a content type? A PDF URL doesn't need to end in `pdf`.

stuartmorgan:
Are you still planning on updating this PR to address the review comments above?

gmackall:
Thanks for your contribution! Since I haven't seen a reply, I'm going to reluctantly close this pull request for now (to get it off our review queue), but if you get time to address the comments, please feel free to re-open this PR!

PR #5018, opened 2023-09-27
[Debug] Run tests on either macOS 12 or macOS 13

vashworth:
Running tests against different ci.yaml configurations to ensure they pass


.ci/targets/ios_platform_tests.yaml
```diff
@@ -21,7 +21,7 @@ tasks:
     args: ["xcode-analyze", "--ios", "--ios-min-version=13.0"]
   - name: native test
     script: script/tool_runner.sh
-    args: ["native-test", "--ios", "--ios-destination", "platform=iOS Simulator,name=iPhone 14,OS=latest"]
+    args: ["native-test", "--ios", "--ios-destination", "platform=iOS Simulator,name=Flutter-iPhone,OS=16.4"]
```
vashworth, line 24: Due to the changes in the way runtimes are handled in CI for macOS 13, we need to be more specific when targeting a simulator device.

.ci/scripts/create_simulator.sh
```diff
@@ -11,5 +11,19 @@ readonly DEVICE_NAME=Flutter-iPhone
 readonly DEVICE=com.apple.CoreSimulator.SimDeviceType.iPhone-14
 readonly OS=com.apple.CoreSimulator.SimRuntime.iOS-16-4
 
+# Delete any existing devices named Flutter-iPhone. Having more than one may
+# cause issues when builds target the device.
```
vashworth, line 15: For reference, here's an example of this problem in action: https://ci.chromium.org/ui/p/flutter/builders/try/Mac_arm64%20custom_package_tests%20master/7040/overview

Unclear if it's a problem new to macOS 13

PR #5032, opened 2023-09-28
Bump AGP & Gradle versions in legacy project used for testing

camsim99:
The [legacy project](https://github.com/flutter/packages/tree/main/.ci/legacy_project) used in this repo for testing has started to interact poorly with the plugins as they are using much higher AGP & Gradle versions. This PR bumps the versions just high enough to fix test failures caused by trying to upgrade the AGP versions of plugins to 8+, for example [this build failure](https://ci.chromium.org/ui/p/flutter/builders/try/Linux_android%20android_build_all_packages%20master/2593/overview) from https://github.com/flutter/packages/pull/4781.

A follow up to @gmackall's findings: https://github.com/flutter/packages/pull/4997#issuecomment-1734613754. Also, just FYI I tested this on https://github.com/flutter/packages/pull/4784.

## Pre-launch Checklist

- [x] I signed the [CLA].


gmackall:
For some timing info (because that will be relevant to if this upgrade is deemed acceptable), these versions were released in April (Gradle) and July (AGP) of 2021.

Related-ish to https://github.com/flutter/flutter/issues/125653

stuartmorgan:
> This PR bumps the versions just high enough to fix test failures caused by trying to upgrade the AGP versions of plugins to 8+, for example [this build failure](https://ci.chromium.org/ui/p/flutter/builders/try/Linux_android%20android_build_all_packages%20master/2593/overview) from #4781.

Unfortunately that error message:
> Could not get unknown property 'android' for project ':path_provider_android' of type org.gradle.api.Project.

is extremely unhelpful to someone without a non-trivial amount of Gradle knowledge, and/or very specific context about what changed in the plugin to trigger the error (which we have in the context of a PR, but plugin clients will not when all the did is something like `flutter pub upgrade`, or added a new plugin)

The legacy project test is just a test of what happens if someone who ran `flutter create` a while ago and hasn't manually updated their Gradle and AGP versions (which many people don't know they need to do). What do we expect clients of our plugins to do when suddenly their project doesn't build and that's the only error message they have? That seems like a very poor experience.

> For some timing info (because that will be relevant to if this upgrade is deemed acceptable), these versions were released in April (Gradle) and July (AGP) of 2021.

The problem is that the important thing isn't when the tools themselves were released, but the state of people's projects. If someone has never heard of AGP because it's never been a problem for them before, and suddenly they have a mysterious error message, the fact that they could in theory have updated that part of their project several years ago won't really matter.

> Related-ish to [flutter/flutter#125653](https://github.com/flutter/flutter/issues/125653).

I think it's much more "related" than "-ish" :) In particular my comment there:

>> We should probably have a soft limit, that warns prominently but continues, which we advance ahead of the hard limit, so that people's first warning that they need to update AGP isn't when they update Flutter and suddenly can't build.

In general, my view is that we should not land any change to the legacy project that is not clearly described by the error message that results from not having that change. I.e., I don't think we should land this until there is a framework change (on stable) that would mean that the error message from a PR like `https://github.com/flutter/packages/pull/4781` is not just:

> Could not get unknown property 'android' for project ':path_provider_android' of type org.gradle.api.Project.

but something like:

>> The plugin 'path_provider_android' requires a newer version of the Android Gradle Plugin that your project is currently using. Please see https://flutter.dev/some_helpful_page to learn how to update AGP and Gradle.

stuartmorgan:
@camsim99 Is there a tracking issue for having some kind of good error message for this case that we can reference here to know when this unblocked?

camsim99:
> @camsim99 Is there a tracking issue for having some kind of good error message for this case that we can reference here to know when this unblocked?

Ah yes this was on my todo list for today :) https://github.com/flutter/flutter/issues/137181

Converting this to a draft for now as I'm not currently working on this.

PR #5034, opened 2023-09-29
[camera] Remove `@throw` from iOS implementation

stuartmorgan:
Using `@throw` in iOS code violates the style guide, so it shouldn't be done in the plugin as mechanism for communicating errors. More importantly, `NSError` is not intended to be used with `@throw`/`@catch`, and is causing issues when compiled with the iOS 17 SDK.

This removes all use of `@throw`, and all `@catch (NSError* e)`, in favor of other methods of communicating errors. It explicitly does not try to fix all the other strange things about this code (having an `NSError` out-param in an init method, using Cocoa and NSURL error domains and codes for some reason), and instead preserves existing behavior as much as possible. In practice, none of these codepaths should ever actually happen (they indicate programming errors within the plugin, not unexpected runtime behavior), and all of this code will go away when converting to Pigeon anyway, so there's not much value in trying to unwind this structure further.

Fixes https://github.com/flutter/flutter/issues/135195

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_avfoundation/CHANGELOG.md
+++ b/packages/camera/camera_avfoundation/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.9.13+6
+
+* Fixes incorrect use of `NSError` that could cause crashes on launch.
+
 ## 0.9.13+5
 
 * Ignores audio samples until the first video sample arrives.
--- a/packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraPropertiesTests.m
+++ b/packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraPropertiesTests.m
@@ -19,7 +19,7 @@
   XCTAssertEqual(FLTFlashModeAuto, FLTGetFLTFlashModeForString(@"auto"));
   XCTAssertEqual(FLTFlashModeAlways, FLTGetFLTFlashModeForString(@"always"));
   XCTAssertEqual(FLTFlashModeTorch, FLTGetFLTFlashModeForString(@"torch"));
-  XCTAssertThrows(FLTGetFLTFlashModeForString(@"unkwown"));
+  XCTAssertEqual(FLTFlashModeInvalid, FLTGetFLTFlashModeForString(@"unknown"));
 }
 
 - (void)testFLTGetAVCaptureFlashModeForFLTFlashMode {
@@ -34,13 +34,13 @@
 - (void)testFLTGetStringForFLTExposureMode {
   XCTAssertEqualObjects(@"auto", FLTGetStringForFLTExposureMode(FLTExposureModeAuto));
   XCTAssertEqualObjects(@"locked", FLTGetStringForFLTExposureMode(FLTExposureModeLocked));
-  XCTAssertThrows(FLTGetStringForFLTExposureMode(-1));
+  XCTAssertNil(FLTGetStringForFLTExposureMode(-1));
 }
 
 - (void)testFLTGetFLTExposureModeForString {
   XCTAssertEqual(FLTExposureModeAuto, FLTGetFLTExposureModeForString(@"auto"));
   XCTAssertEqual(FLTExposureModeLocked, FLTGetFLTExposureModeForString(@"locked"));
-  XCTAssertThrows(FLTGetFLTExposureModeForString(@"unknown"));
+  XCTAssertEqual(FLTExposureModeInvalid, FLTGetFLTExposureModeForString(@"unknown"));
 }
 
 #pragma mark - focus mode tests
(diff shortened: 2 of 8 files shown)


packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -118,10 +118,16 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
...
+  _resolutionPreset = FLTGetFLTResolutionPresetForString(resolutionPreset);
+  if (_resolutionPreset == FLTResolutionPresetInvalid) {
+    *error = [NSError
+        errorWithDomain:NSCocoaErrorDomain
+                   code:NSURLErrorUnknown
+               userInfo:@{
+                 NSLocalizedDescriptionKey :
+                     [NSString stringWithFormat:@"Unknown resolution preset %@", resolutionPreset]
+               }];
+    return nil;
```
stuartmorgan, line 130: This is technically a behavioral change since not returning `nil` on error was clearly wrong, but also this should never be able to happen anyway.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -162,7 +168,9 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
   _motionManager = [[CMMotionManager alloc] init];
   [_motionManager startAccelerometerUpdates];
 
-  [self setCaptureSessionPreset:_resolutionPreset];
+  if (![self setCaptureSessionPreset:_resolutionPreset withError:error]) {
```
stuartmorgan, line 171: This also technically a behavioral change in that now the error actually gets returned instead of us having an unhandled `@throw` of an `NSError`, which... I don't even actually know what that would do. Crash the app probably?

But as above, this should not actually be a thing that can happen as far as I can tell.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -382,17 +390,17 @@ - (void)setCaptureSessionPreset:(FLTResolutionPreset)resolutionPreset {
...
       } else {
-        NSError *error =
-            [NSError errorWithDomain:NSCocoaErrorDomain
-                                code:NSURLErrorUnknown
-                            userInfo:@{
-                              NSLocalizedDescriptionKey :
-                                  @"No capture session available for current capture session."
-                            }];
-        @throw error;
+        *error = [NSError errorWithDomain:NSCocoaErrorDomain
```
stuartmorgan, line 393: All of these NSErrors are directly moved from the code that was `@throw`ing them, to preserve behavior. As noted in the PR description, I'm deliberately not changing the weird domain and code used in these errors for now.

This was some deeply strange Obj-C code.

packages/camera/camera_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.9.13+6
+
+* Fixes incorrect use of `NSError` that could cause crashes on launch.
```
hellohuanlin, line 3: Nit: ... cause crashes on launch *on iOS 17*

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraPropertiesTests.m
```diff
@@ -19,7 +19,7 @@ - (void)testFLTGetFLTFlashModeForString {
   XCTAssertEqual(FLTFlashModeAuto, FLTGetFLTFlashModeForString(@"auto"));
   XCTAssertEqual(FLTFlashModeAlways, FLTGetFLTFlashModeForString(@"always"));
   XCTAssertEqual(FLTFlashModeTorch, FLTGetFLTFlashModeForString(@"torch"));
-  XCTAssertThrows(FLTGetFLTFlashModeForString(@"unkwown"));
+  XCTAssertEqual(FLTFlashModeInvalid, FLTGetFLTFlashModeForString(@"unknown"));
```
hellohuanlin, line 22: Nit: should it be `FLTFlashModeUnknown`? in case we have new flash mode in the future.

hellohuanlin (approved):
I saw these last year when dry-running a swift migration on camera plugin. Thanks for fixing - it will make future migration easier.

packages/camera/camera_avfoundation/CHANGELOG.md
stuartmorgan: I unfortunately don't know if it's "on iOS 17" or "when compiled against the iOS 17 SDK" or some other slight variant of that, which is why I left it vague.

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraPropertiesTests.m
stuartmorgan: If we have a new flash mode in the future then we need to update both the Dart and Obj-C code at the same time. This code is purely for manual platform channel conversions, which means the only way to get a value we don't recognize is if someone adds [serialization for that mode on the Dart side](https://github.com/flutter/packages/blob/d0e9a0e1b3ac1f8afeba2a30da6d30fa06d96a48/packages/camera/camera_avfoundation/lib/src/avfoundation_camera.dart#L531) but not the corresponding deserialization code on the native side, which would make no sense.

If we have flash modes that the Dart side of `camera_avfoundation` has been updated to send over the method channel but the Obj-C side doesn't know about, then someone has made a serious mistake in adding a new flash mode. We shouldn't explicitly allow for that case with an "unknown", implying that's an okay state to be in.

stuartmorgan: (This is why Pigeon will moot all of this; in Pigeon it's impossible for an enum value defined in the interface to exist on one side but not the other.)

PR #5038, opened 2023-09-29
[two_dimensional_scrollables] Fix paint bug when rows/columns are pinned and axes are reversed

Piinks:
Fixes https://github.com/flutter/flutter/issues/135386

This adds one more golden to verify the painting, but once public mock_canvas rolls to stable in flutter_test, we can remove all of the goldens and just verify using the `paints` method.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #5039, opened 2023-09-29
[two_dimensional_scrollables] Add TableSpanPadding

Piinks:
- [x] Waiting on https://github.com/flutter/packages/pull/5038 first

---

Fixes https://github.com/flutter/flutter/issues/134453
Also related: https://github.com/flutter/flutter/issues/134655

This adds padding to TableSpans with TableSpanPadding. This affords folks a leading and trailing offset to pad rows and columns by.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/two_dimensional_scrollables/lib/src/table_view/table.dart
```diff
@@ -667,12 +667,14 @@ class RenderTableViewport extends RenderTwoDimensionalViewport {
   }) {
     // TODO(Piinks): Assert here or somewhere else merged cells cannot span
     // pinned and unpinned cells (for merged cell follow-up), https://github.com/flutter/flutter/issues/131224
-    double yPaintOffset = -offset.dy;
+    late double yPaintOffset = -offset.dy;
     for (int row = start.row; row <= end.row; row += 1) {
       double xPaintOffset = -offset.dx;
       final double rowHeight = _rowMetrics[row]!.extent;
+      yPaintOffset += _rowMetrics[row]!.configuration.padding.leading;
```
goderbauer, line 674: nit: maybe get the rowMetric only once to avoid extra null checks:

```suggestion
      final rowMetric = _rowMetrics[row]!;
      final double rowHeight = rowMetric.extent;
      yPaintOffset += rowMetric.configuration.padding.leading;
```

goderbauer: Same for columnMetrics and the other rowMetric access below.

Quijx:
The paddings for the `TableSpans` are useful for sure the way they are implemented here, but I think having a normal `padding` on the `TableView` that only applies to the outer sides of the table in addition would also be useful. For example in the app I showed in https://github.com/flutter/flutter/issues/134453, I have a global padding defined in a theme extension that needs to be applied to every screen filling scrollable on each page. Or I could imagine a case where someone wants to use `MediaQuery.paddingOf(context)` to make room for system UI while still seeing the cells scroll below it. Integrating such a padding into the  `TableSpans` would be a bit verbose:
```dart
@override
    final padding = MediaQuery.paddingOf(context);
    const columnCount = 100;
    const rowCount = 100;
    return TableView.builder(
      cellBuilder: (context, vicinity) {
        return const Text('foo');
      },
      columnCount: columnCount,
      columnBuilder: (int column) => TableSpan(
        extent: const FixedTableSpanExtent(64),
        padding: TableSpanPadding(
          leading: column == 0 ? padding.left : 0.0,
          trailing: column == columnCount - 1 ? padding.right : 0.0,
        ),
      ),
      rowCount: rowCount,
      rowBuilder: (int row) => TableSpan(
        extent: FixedTableSpanExtent(64),
        padding: TableSpanPadding(
          leading: row == 0 ? padding.top : 0.0,
          trailing: row == rowCount - 1 ? padding.bottom : 0.0,
        ),
      ),
    );
  }
```

----

Also the `TableSpanExtent` should probably clarify in its doc that the colums/rows width/height that it describes does not include the padding.

packages/two_dimensional_scrollables/lib/src/table_view/table_span.dart
```diff
@@ -10,6 +10,37 @@ import 'package:flutter/widgets.dart';
 
 import 'table.dart';
 
+/// Defines the leading and trailing padding values of a [TableSpan].
```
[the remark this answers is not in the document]
Piinks: I didn't want to enumerate multiple `*Padding` properties on TableSpan and just nicely encapsulate it like we would in Padding. I am anticipating the number of configurations on TableSpan growing over time, so I'd like to try to keep it as clean as possible. RawScrollbar is an example, the list or params has grown so much over time.

packages/two_dimensional_scrollables/lib/src/table_view/table.dart
```diff
@@ -689,9 +691,11 @@ class RenderTableViewport extends RenderTwoDimensionalViewport {
           cell.layout(cellConstraints);
           cellParentData.layoutOffset = Offset(xPaintOffset, yPaintOffset);
         }
-        xPaintOffset += columnWidth;
+        xPaintOffset += columnWidth +
+            _columnMetrics[column]!.configuration.padding.trailing;
       }
-      yPaintOffset += rowHeight;
+      yPaintOffset +=
```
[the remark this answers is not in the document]
Piinks: Ah yes! Thank you, when I whipped this up I thought - wait it can't be this simple. Thanks for identifying that case. I will update and add a test that verifies we don't lay out unnecessary rows/columns on account of padding. :D

Piinks:
> a normal padding on the TableView that only applies to the outer sides of the table in addition would also be useful

That can be achieved here by applying the appropriate padding to the first/last columns/rows. Additionally, you can always wrap the TableView with a padding widget. In general, we avoid enumerating padding properties on widgets when we have a Padding widget itself. :)

> I could imagine a case where someone wants to use MediaQuery.paddingOf(context) to make room for system UI while still seeing the cells scroll below it

This sounds like a case for SafeArea!

> Also the TableSpanExtent should probably clarify in its doc that the columns/rows width/height that it describes does not include the padding.

Good point! Thanks for the feedback!

Quijx:
> That can be achieved here by applying the appropriate padding to the first/last columns/rows.

Sure. This wouldn't make anything possible that wasn't possible before. It would just be a convenience feature and would make it consistent with other scrollable widgets such as `ListView` or `GridView` which also have a `padding`.

> Additionally, you can always wrap the TableView with a padding widget.
> This sounds like a case for SafeArea!

But wouldn't both a `SafeArea` or a wrapping `Padding` widget behave different from a build in padding the same way it is different in for example a `ListView`?

`Padding` widget around `ListView`:

https://github.com/flutter/packages/assets/48153628/1780514e-b770-48bf-b825-be4db785b8ce

`padding` attribute on `ListView`:

https://github.com/flutter/packages/assets/48153628/545e308f-aaed-49a1-88ff-a91be57c3e3e

packages/two_dimensional_scrollables/test/table_view/table_test.dart
```diff
@@ -950,13 +1255,17 @@ void main() {
         (WidgetTester tester) async {
       // TODO(Piinks): Rewrite this to remove golden files from this repo when
       //  mock_canvas is public - https://github.com/flutter/flutter/pull/131631
-      // foreground, background, and precedence per mainAxis
+      //  * foreground, background, and precedence per mainAxis
```
Piinks, line 1258: I realize I am packing a lot into this single golden file test, but I don't want to add more images to the repo when very soon I will be able to rewrite this into tests that use the mock canvas API when it reaches stable.

packages/two_dimensional_scrollables/lib/src/table_view/table.dart
```diff
@@ -667,12 +674,17 @@ class RenderTableViewport extends RenderTwoDimensionalViewport {
   }) {
```
goderbauer, line 674: Could you explain to me how the padding is taken into account when calculating the `start` and `end` (input to this method) cells?

goderbauer: _Span.trailingOffset is used to calculate the start and end, which includes the padding. There are even tests for this case. Awesome!

goderbauer (approved):
LGTM

packages/two_dimensional_scrollables/lib/src/table_view/table.dart
```diff
@@ -308,6 +308,13 @@ class RenderTableViewport extends RenderTwoDimensionalViewport {
     );
   }
 
+  // TODO(Piinks): Pinned rows/cols do not account for what is visible on the
+  //  screen. Ostensibly, we would not want to have pinned rows/columns that
+  //  extend beyond the viewport, we would never see them as they would never
+  //  scroll into view. So this currently implementation is fairly assuming
+  //  we will never have rows/cols that are outside of the viewport. We should
+  //  maybe add an assertion for this during layout.
```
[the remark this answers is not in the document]
Piinks: We discussed this offline and agreed we will pursue a non fatal warning here in the same way we do the render overflow error.

packages/two_dimensional_scrollables/lib/src/table_view/table_span.dart
```diff
@@ -20,20 +56,26 @@ class TableSpan {
   /// The [extent] argument must be provided.
   const TableSpan({
     required this.extent,
+    TableSpanPadding? padding,
```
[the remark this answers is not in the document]
Piinks: I had it this way originally, then when writing the sample code, I found it unfriendly. For example:

```dart
// Non-nullable
TableSpan(
  padding: someCondition ? TableSpanPadding(leading: 20) : TableSpanPadding(),
  // ...
)
```

versus

```dart
// Nullable, resolves to non nullable in field member assignment
TableSpan(
  padding: someCondition ? TableSpanPadding(leading: 20) : null,
  // ...
)
```

PR #5071, opened 2023-10-04
[webview_flutter] Fix XCUITest in Xcode 15

stuartmorgan:
For some reason tapping the "Done" button in the XCUITest stopped working in Xcode 15, but tapping it via coordinates works, so switch to that as a workaround.

Fixes https://github.com/flutter/flutter/issues/134971


stuartmorgan:
Normally I would file an issue against Apple and reference it in the comment, but since this is test-only and trivial to work around I don't think it's worth the effort of making a reduced test case.

bparrishMines (approved):
LGTM

PR #5074, opened 2023-10-04
[image_picker] Fix XCTests in Xcode 15

stuartmorgan:
When compiled with Xcode 15, images that can't be loaded by UIImage will cause the codepaths in some XCTest tests to throw exceptions. To address this:
- Replace the test ICO file with one that UIImage is able to open.
- Remove the SVG test, since UIImage cannot directly load SVGs.

Fixes https://github.com/flutter/flutter/issues/134973


vashworth (approved):
So SVGs don't work with image_picker at all? If SVGs don't work at all, we should probably file an issue to investigate supporting them / making sure we have proper error messaging.

Also, just curious, do you know what was wrong with the ICO image?

Other than that, LGTM. Thanks for fixing this!

stuartmorgan:
> So SVGs don't work with image_picker at all? If SVGs don't work at all, we should probably file an issue to investigate supporting them / making sure we have proper error messaging.

I don't know if they work or not, but this test isn't actually testing them because the item provider returns `nil` since `UIImage` doesn't support SVGs directly. From some looking around, SVG support on iOS seems pretty limited, so I'm not sure how easy it would be to validate them. I figured it's obscure enough for the `image_picker` use case that we could leave any further investigation of SVG support to be custom-issue-report driven.

> Also, just curious, do you know what was wrong with the ICO image?

Nope, I looked around a bit and couldn't figure out what the requirements are for macOS to handle an ICO. I don't think the ICO itself was invalid, I think Apple's libraries (it's not just UIImage; Finder wouldn't show a preview for it and Preview wouldn't load it) just can't handle some subset of the format.

stuartmorgan:
Changelog/version override: these changes are test-only, just in a way the tooling doesn't detect.

guptavirendra:
I am facing this issue on Xcode 15  
Please help to resolve this issue. It would be great help for me. Thanks in advance.



webview_bug

PR #5075, opened 2023-10-04
[path_provider] Fix integration tests

stuartmorgan:
- Fix the `throwsA` test to properly await.
- Updates the `getDownloadsDirectory` test to not assume that Android will throw, since that is no longer true.


cyanglaz (approved):
LGTM

PR #5078, opened 2023-10-05
[video_player] Improve macOS frame management

stuartmorgan:
Fixes some logic around getting new frames on macOS:
- The logic to copy a frame had a race condition; the macOS code checks that a frame is available for the current playback time before informing the engine that a frame is available (it's not clear why the iOS code doesn't do this; apparently the engine tolerates returning NULL frames on iOS?), but failed to account for the fact that the current playback time when the (async) request from the engine comes in can be different. This fixes it to remember the last time that was known to be available, and uses that when the current time isn't available. This fixes flickering on playback (since returning NULL on macOS causes the video to vanish until a new frame is available).
- Fixes seek to temporarily restart the display link if the video is paused, rather than telling the engine that a frame is available, because it might not be. This is changed for both macOS and iOS since I don't see any reason this bug couldn't affect iOS as well (although in practice I'm only aware of it being reproducible on macOS).

This extracts the display link code for macOS and iOS into an abstraction, eliminating most of the ifdefing, in order to support the latter (since more code needs to be able to play/pause the display link), which also resolves a TODO from the initial implementation.

There is also some non-trivial addition of factory injection in order to make the code more testable. This is definitely not complete, but it incrementally moves the code toward being more testable than it was before, and allows for testing the display link behavior.

Lastly, this moves some code used by tests to the existing `_Test.h` header, removing redeclarations from unit test files, since we already have a test header and that's our preferred approach for accessing private details in ObjC tests. (Longer term the multi-class mega-file should be broken up more to reduce the need for that.)

Fixes https://github.com/flutter/flutter/issues/136027
Improves https://github.com/flutter/flutter/issues/135999

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
I can pull https://github.com/flutter/packages/pull/5078/commits/a1d46cbab6723fda88586bedf9b8e2c82c8886b3 out into a prequel PR if it would help with review, but hopefully just having it as a separate commit here makes it easy enough.

stuartmorgan:
(I can also do some of the testability refactoring, like moving things to the Test header and adjusting the AV* factory injection, in a prequel PR if desired.)

hellohuanlin:
CC: @hellohuanlin

packages/video_player/video_player_avfoundation/darwin/Classes/ios/FVPDisplayLink.m
```diff
@@ -0,0 +1,65 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "../FVPDisplayLink.h"
+
+#import <Foundation/Foundation.h>
+#import <QuartzCore/QuartzCore.h>
+
+/** A proxy object to act as a CADisplayLink target, to avoid retain loops. */
```
hellohuanlin, line 10: can you explain more about the retain cycle situation and how this helps preventing it?

packages/video_player/video_player_avfoundation/darwin/Classes/ios/FVPDisplayLink.m
```diff
@@ -0,0 +1,65 @@
...
+  }
+  return self;
+}
+
+- (void)dealloc {
+  [_displayLink invalidate];
+}
+
+- (void)setRunning:(BOOL)running {
+  self.displayLink.paused = !running;
```
hellohuanlin, line 62: also `_running = running`?

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
```diff
@@ -20,66 +21,72 @@ @interface FVPFrameUpdater : NSObject
...
+  // Only report a new frame if one is actually available, or the check is being skipped.
+  BOOL reportFrame = NO;
+  if (self.skipBufferAvailabilityCheck) {
+    reportFrame = YES;
+  } else {
+    CMTime outputItemTime = [self.videoOutput itemTimeForHostTime:CACurrentMediaTime()];
+    if ([self.videoOutput hasNewPixelBufferForItemTime:outputItemTime]) {
+      _lastKnownAvailableTime = outputItemTime;
+      reportFrame = YES;
+    }
```
hellohuanlin, line 53: Reset in else? 

```
else {
  _lastKnownAvailableTime = kCMTimeInvalid;
  reportFrame = NO;
}
```

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
```diff
@@ -558,12 +551,29 @@ - (void)setPlaybackSpeed:(double)speed {
...
-    return NULL;
+    // If the current time isn't available yet, use the time that was checked when informing the
+    // engine that a frame was available (if any).
+    CMTime lastAvailableTime = self.frameUpdater.lastKnownAvailableTime;
+    if (CMTIME_IS_VALID(lastAvailableTime)) {
+      buffer = [_videoOutput copyPixelBufferForItemTime:lastAvailableTime itemTimeForDisplay:NULL];
+    }
+  }
+
+  if (self.waitingForFrame && buffer) {
```
hellohuanlin, line 567: When will buffer be nil (seeking to the end?)
 
should this be just `if (self.waitingForFrame)` so that displayLink is guaranteed to shutdown?

packages/video_player/video_player_avfoundation/darwin/Classes/ios/FVPDisplayLink.m
stuartmorgan: Added more to the comment.

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
stuartmorgan: `reportFrame` is a local initialized to `NO`, so there's no need to set it to `NO`.

Why would we reset `_lastKnownAvailableTime`? If we don't have a new time, then the last known time is still the last known time.

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
stuartmorgan: > When will buffer be nil (seeking to the end?)

Seeking to the end should work. It could be nil on iOS since iOS just fires the display link all the time and relies on the engine (apparently) allowing null frames. There may also be edge cases on macOS where we still can get nulls.

> should this be just `if (self.waitingForFrame)` so that displayLink is guaranteed to shutdown?

But if we're waiting for a frame and we didn't get one yet, then we need to wait longer.

Consider iOS seeking while paused, for instance, and then firing the display link indiscriminately before the video frame is actually ready. In that case, we don't want to stop the display link or the frame will be stuck on the old location. Instead we need to wait until the video decoding catches up and there's a frame to show.

hellohuanlin (approved):
LGTM!

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
```diff
@@ -20,66 +21,72 @@ @interface FVPFrameUpdater : NSObject
...
-#if TARGET_OS_IOS
-- (void)onDisplayLink:(CADisplayLink *)link;
-#endif
+// The last time that has been validated as avaliable according to hasNewPixelBufferForItemTime:.
+@property(nonatomic, assign) CMTime lastKnownAvailableTime;
+// If YES, the engine is informed that a new texture is available any time the display link
+// callback is fired, regardless of the videoOutput state.
+//
+// TODO(stuartmorgan): Investigate removing this; it exists only to preserve existing iOS behavior
+// while implementing macOS, but iOS should very likely be doing the check as well.
```
cbracken, line 30: Add an issue link here.

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
cbracken: It may be worth documenting the macOS edge cases here in a comment.

cbracken (approved):
lgtm modulo a couple tiny doc nits.

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
stuartmorgan: To clarify, I don't actually know what the macOS edge cases are. Recent comments in https://github.com/flutter/flutter/issues/135999 suggest they still exist, but I don't yet know when/why.

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
stuartmorgan: Filed and added!

stuartmorgan:
Hm, looks like one of the new tests may be flaky. I'll investigate that.

stuartmorgan:
I couldn't get it to flake locally in many attempts, and there's no info about what failed on the bot, so I'm going to land it and see if it was a fluke, or if it shows up again (hopefully with details about what actually failed).

PR #5082, opened 2023-10-06
[webview_flutter_wkwebview] skip other navigation requests on iOS

dmeijboom:
In flutters webview package you can set a navigation delegate which is triggered when the user clicks on a link in the webview. This can be used to load these pages in an in-app browser or external browser for example but I can imagine it has other use-cases as well.

On Android this seems to work well but on iOS the delegate is triggered onload without any user action involved. It seems like this is an implementation issue. Currently, we the `decidePolicyForNavigationAction` handler is used to trigger the navigation delegate in flutter. This is fine but it seems like this handler is more generic on iOS and is also triggered whenever an URL is loaded. The fix has been discussed already:

See: https://github.com/flutter/flutter/issues/73242#issuecomment-1358461716

Now that https://github.com/flutter/plugins/pull/6863 has added support for `WKNavigationAction.navigationType` we can actually solve this issue by excluding `other` navigation types. This is what this PR does, it excludes the `other` navigation type so that both Android and iOS behave similarly.

Issue: https://github.com/flutter/flutter/issues/73242

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
+++ b/packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
@@ -888,7 +888,7 @@ class WebKitNavigationDelegate extends PlatformNavigationDelegate {
         WKWebView webView,
         WKNavigationAction action,
       ) async {
-        if (weakThis.target?._onNavigationRequest != null) {
+        if (weakThis.target?._onNavigationRequest != null && action.navigationType != WKNavigationType.other) {
           final NavigationDecision decision =
               await weakThis.target!._onNavigationRequest!(NavigationRequest(
             url: action.request.url,


bparrishMines:
I was considering doing this, but I decided against it to avoid requiring a breaking change for the long term solution.

iOS provides the [WKNavigationType](https://developer.apple.com/documentation/webkit/wknavigationtype?language=objc) and Android provides [WebResourceRequest.hasGesture]. I think it would be better to update the platform interface to provide a `NavigationType` in `NavigationRequest`.

If you would be interested in implementing this, you can follow https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins. Otherwise we can close this PR.

dmeijboom:
> I was considering doing this, but I decided against it to avoid requiring a breaking change for the long term solution.
> 
> iOS provides the [WKNavigationType](https://developer.apple.com/documentation/webkit/wknavigationtype?language=objc) and Android provides [WebResourceRequest.hasGesture]. I think it would be better to update the platform interface to provide a `NavigationType` in `NavigationRequest`.
> 
> If you would be interested in implementing this, you can follow https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins. Otherwise we can close this PR.

I guess it makes sense. Even though it's probably 'fine' for most apps it would be a breaking change which can be easily avoided by passing the navigation type. We will make sure this gets implemented but I don't have a concrete timeline yet.

MitchellGoodwin:
Marking this as a draft pending the changes discussed above.

stuartmorgan:
Since this is marked as a draft and hasn't been updated in several months I'm going to close it to clean out our review queue. Please don't hesitate to submit a new PR if you decide to revisit this. Thanks!

PR #5089, opened 2023-10-09
Run tests on either macOS 12 or macOS 13

vashworth:
In preparation for migrating our fleet to macOS 13, we're updating tests to run on either macOS 12 or macOS 13.

## Pre-launch Checklist

- [x] I signed the [CLA].


.ci.yaml
```diff
@@ -77,7 +77,7 @@ platform_properties:
         [
           {"dependency": "gems", "version": "v3.3.14"}
         ]
-      os: Mac-12
+      os: Mac-13
```
vashworth, line 80: Temporarily only running on macOS 13 to show that tests pass

vashworth: @stuartmorgan After you take a look, I'll update to do either 12 or 13

.ci/targets/ios_platform_tests.yaml
```diff
@@ -21,7 +21,11 @@ tasks:
     args: ["xcode-analyze", "--ios", "--ios-min-version=13.0"]
   - name: native test
     script: script/tool_runner.sh
-    args: ["native-test", "--ios", "--ios-destination", "platform=iOS Simulator,name=iPhone 14,OS=latest"]
+    args: ["native-test", "--ios", "--ios-destination", "platform=iOS Simulator,name=Flutter-iPhone,OS=16.4"]
```
vashworth, line 24: As we upgrade, there's no guarantee that there will be an iPhone 14 simulator. Better to use the simulator we create in the "create simulator" step (same for the pigeon test_suites.dart).

.ci/scripts/boot_simulator.sh
```diff
@@ -0,0 +1,12 @@
...
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+set -e
+
+# The name here must match create_simulator.sh
+readonly DEVICE_NAME=Flutter-iPhone
+
+xcrun simctl boot "$DEVICE_NAME" || :
+echo -e ""
+xcrun simctl list
```
stuartmorgan, line 12: Nit: please add the missing newline.

stuartmorgan, line 12: Do we want to keep this echo?

.ci/targets/ios_platform_tests.yaml
stuartmorgan: Pigeon's `test_suites` is a weird case; we need it to run locally as well as in CI. So we don't want to hard-code it to a simulator that only exists in CI.

.ci/targets/ios_platform_tests.yaml
```diff
@@ -21,7 +21,11 @@ tasks:
...
+  - name: boot simulator
+    # Ensure simulator is still booted
```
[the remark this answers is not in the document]
stuartmorgan: That's really weird. Is there any indication of why it's shutting down?

.ci/scripts/boot_simulator.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Copyright 2013 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+set -e
+
+# The name here must match create_simulator.sh
+readonly DEVICE_NAME=Flutter-iPhone
+
+xcrun simctl boot "$DEVICE_NAME" || :
```
stuartmorgan, line 10: This could use a comment (i.e., is the `|| :` because this returns non-zero if it's already booted?)

.ci/scripts/boot_simulator.sh
vashworth: It's not needed, but I liked it - it puts a newline between the `boot` output and `list` output. Just gives it a little separation

.ci/targets/ios_platform_tests.yaml
vashworth: Okay, so I updated Pigeon's `test_suites` to create and delete it's own simulator so that it'll work in CI and locally

FYI couple reasons I think that's the best approach: 
1.  Due to new way Simulator runtimes are handled, when you use `OS=latest` or you omit the OS (which defaults to `latest`), it may use a newer simulator runtime that's not the default for the Xcode version. For example, if iOS 17 runtime is mounted, it'll use that instead of iOS 16.4 even if you're using Xcode 14.3. So even if there's an "iPhone 14" device with OS 16.4, it won't match against it since it's looking for iOS 17.
2. There's no guarantee to be a Simulator named "iPhone 14" unless we create it beforehand (both in CI and locally)

vashworth: Also btw, I ran `dart run tool/run_tests.dart` in `packages/packages/pigeon` locally to make sure it passed locally

.ci/scripts/create_simulator.sh
```diff
@@ -11,5 +11,19 @@ readonly DEVICE_NAME=Flutter-iPhone
...
+RESULT=0
+while [[ $RESULT == 0 ]]; do
+    xcrun simctl delete "$DEVICE_NAME" || RESULT=1
+    if [ $RESULT == 0 ]; then
+        echo -e "Deleted $DEVICE_NAME"
+    fi
+done
+echo -e ""
+
 xcrun simctl list
```
stuartmorgan, line 26: Seems like we could remove this now that we're running it at the end.

.ci/targets/ios_platform_tests.yaml
stuartmorgan, line 24: This should get a comment mentioning that the name should match the name in create_simulator.sh

stuartmorgan (approved):
LGTM with nit (and with the .ci.yaml `os` changed to 12-or-13 per your comment).

tarrinneal (approved):

PR #5099, opened 2023-10-10
[quick_actions] Convert android to pigeon

tarrinneal:
Converts from direct use of method channels to using Pigeon.

Part of https://github.com/flutter/flutter/issues/117844
part of https://github.com/flutter/flutter/issues/117913

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
```diff
@@ -0,0 +1,184 @@
...
+                        } else {
+                          result.error(
+                              new FlutterError(
+                                  "quick_action_setshortcutitems_failure",
+                                  "Exception thrown when setting dynamic shortcuts",
+                                  null));
+                        }
+                      }
+                    })
+                .start();
```
tarrinneal, line 91: This used to be run on the main thread. I'm not 100% sure this is the correct way to handle moving it to the background thread. Would appreciate some feedback.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
```diff
@@ -1,180 +0,0 @@
...
-                  shortcutManager.setDynamicShortcuts(shortcuts);
-                  dynamicShortcutsSet = true;
-                } catch (Exception e) {
-                  // Leave dynamicShortcutsSet as false
-                }
-
-                final boolean didSucceed = dynamicShortcutsSet;
-
-                // TODO(camsim99): Move re-dispatch below to background thread when Flutter 2.8+ is
-                // stable.
```
tarrinneal, line 78: todo referencing moving to bg thread

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/MethodCallHandlerImpl.java
```diff
@@ -1,180 +0,0 @@
...
-                    () -> {
-                      if (didSucceed) {
-                        result.success(null);
-                      } else {
-                        result.error(
-                            "quick_action_setshortcutitems_failure",
-                            "Exception thrown when setting dynamic shortcuts",
-                            null);
-                      }
-                    });
```
tarrinneal, line 89: todo referencing moving to bg thread

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActionsPlugin.java
```diff
@@ -97,26 +115,20 @@ public boolean onNewIntent(@NonNull Intent intent) {
...
+    if (intent.hasExtra(QuickActions.EXTRA_ACTION)) {
+      Context context = this.quickActions.getActivity().getApplicationContext();
       ShortcutManager shortcutManager =
           (ShortcutManager) context.getSystemService(Context.SHORTCUT_SERVICE);
-      String shortcutId = intent.getStringExtra(MethodCallHandlerImpl.EXTRA_ACTION);
-      channel.invokeMethod("launch", shortcutId);
+      String shortcutId = intent.getStringExtra(QuickActions.EXTRA_ACTION);
+      quickActionsFlutterApi.launchAction(
+          shortcutId,
+          new AndroidQuickActionsFlutterApi.Reply<Void>() {
```
stuartmorgan, line 125: Since this is a single method, can't we replace the whole thing with a lambda? (If so, AS will offer to do it for you automagically via a tip.)

packages/quick_actions/quick_actions_android/android/src/test/java/io/flutter/plugins/quickactions/QuickActionsTest.java
```diff
@@ -75,11 +69,13 @@ public void onAttachedToActivity_buildVersionSupported_invokesLaunchMethod()
     final QuickActionsPlugin plugin =
         new QuickActionsPlugin((version) -> SUPPORTED_BUILD >= version);
     setUpMessengerAndFlutterPluginBinding(testBinaryMessenger, plugin);
-    Field handler = plugin.getClass().getDeclaredField("handler");
-    handler.setAccessible(true);
-    handler.set(plugin, mock(MethodCallHandlerImpl.class));
+    //     Field handler = plugin.getClass().getDeclaredField("quickActions");
+    //     final QuickActions mockQuickActions = mock(QuickActions.class);
+    //     handler.setAccessible(true);
+    //     handler.set(plugin, mockQuickActions);
```
stuartmorgan, line 75: I assume this and the below can be removed?

packages/quick_actions/quick_actions_android/pigeons/messages.dart
```diff
@@ -0,0 +1,51 @@
...
+    this.icon,
+  );
+
+  /// The identifier of this item; should be unique within the app.
+  String type;
+
+  /// Localized title of the item.
+  String localizedTitle;
+
+  /// Name of native resource (xcassets etc; NOT a Flutter asset) to be
```
stuartmorgan, line 31: Since this is Android, probably not an xcasset :)

packages/quick_actions/quick_actions_android/pigeons/messages.dart
```diff
@@ -0,0 +1,51 @@
...
+  /// displayed as the icon for this item.
+  String? icon;
+}
+
+@HostApi()
+abstract class AndroidQuickActionsApi {
+  @async
+  String? getLaunchAction();
```
stuartmorgan, line 39: These should all get documentation comments saying what they are supposed to do (as a guide for implementation on the native side).

packages/quick_actions/quick_actions_android/pigeons/messages.dart
```diff
@@ -0,0 +1,51 @@
...
+  @async
+  void setShortcutItems(List<ShortcutItemMessage> itemsList);
+
+  @async
+  void clearShortcutItems();
+}
+
+@FlutterApi()
+abstract class AndroidQuickActionsFlutterApi {
+  void launchAction(String action);
```
stuartmorgan, line 50: Same here, as a guide for the Dart implementation.

packages/quick_actions/quick_actions_android/test/quick_actions_android_test.dart
```diff
@@ -4,169 +4,98 @@
...
-/// with `!` and `?` on the stable branch.
-T? _ambiguate<T>(T? value) => value;
+class _FakeQuickActionsApi implements AndroidQuickActionsApi {
+  List<ShortcutItem> items = <ShortcutItem>[];
+  String? launchAction;
+
+  @override
+  Future<void> clearShortcutItems() {
+    items = <ShortcutItem>[];
+    return Future<void>.value();
```
stuartmorgan, line 82: You an just make the method `async` and not need this line.

packages/quick_actions/quick_actions_android/test/quick_actions_android_test.dart
```diff
@@ -4,169 +4,98 @@
...
+  }
+
+  @override
+  Future<String?> getLaunchAction() {
+    launchAction = LAUNCH_ACTION_STRING;
+    return Future<String?>.value(launchAction);
```
stuartmorgan, line 88: Similarly, `async` means you can just `return launchAction;`

packages/quick_actions/quick_actions_android/test/quick_actions_android_test.dart
```diff
@@ -4,169 +4,98 @@
...
+
+  @override
+  Future<void> clearShortcutItems() {
+    items = <ShortcutItem>[];
+    return Future<void>.value();
+  }
+
+  @override
+  Future<String?> getLaunchAction() {
+    launchAction = LAUNCH_ACTION_STRING;
```
stuartmorgan, line 87: This seems like an odd side effect; if the goal is to detect the method being called it can just be a bool tracking that (e.g., `getLaunchActionCalled`).

packages/quick_actions/quick_actions_android/test/quick_actions_android_test.dart
```diff
@@ -4,169 +4,98 @@
...
+    return Future<String?>.value(launchAction);
+  }
+
+  @override
+  Future<void> setShortcutItems(List<ShortcutItemMessage?> itemsList) async {
+    await clearShortcutItems();
+    for (final ShortcutItemMessage? element in itemsList) {
+      if (element != null) {
```
stuartmorgan, line 95: Wouldn't a null here indicate a bug in the code being tested? I would think we would want this to crash, not silently skip.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
stuartmorgan: I believe that TODO meant that we could just remove this whole redispatch, because the engine added support for calling `result` on threads other than the platform thread.

I would suggest we leave the UI redispatch code as-is for this migration and change the threading in a different PR though, to minimize churn.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  public Activity getActivity() {
+    return this.activity;
+  }
+
+  @ChecksSdkIntAtLeast(api = Build.VERSION_CODES.N_MR1)
+  boolean isVersionAllowed() {
+    // We already know that this functionality does not work for anything
+    // lower than API 25 so we chose not to return error. Instead we do nothing.
```
stuartmorgan, line 49: This comment doesn't make sense here with the restructuring; it belongs on the `if` that calls it.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
```diff
@@ -0,0 +1,183 @@
...
+    return Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1;
+  }
+
+  @Override
+  public void setShortcutItems(
+      @NonNull List<ShortcutItemMessage> itemsList, @NonNull Result<Void> result) {
+    if (isVersionAllowed()) {
```
stuartmorgan, line 56: Early return guards (i.e., the pattern that was used in the old code) are often clearer than nesting all of the logic.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
```diff
@@ -0,0 +1,183 @@
...
+                              new FlutterError(
+                                  "quick_action_setshortcutitems_failure",
+                                  "Exception thrown when setting dynamic shortcuts",
+                                  null));
+                        }
+                      }
+                    })
+                .start();
+          });
+    }
```
stuartmorgan, line 92: You have a codepath that never calls `result`, which is an error.

stuartmorgan: Seems like we should add unit tests that the no-op for old versions actually returns.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
```diff
@@ -0,0 +1,183 @@
...
+    return;
```
stuartmorgan, line 93: You don't need to explicitly `return` at the end of a method.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  @Override
+  public void clearShortcutItems(@NonNull Result<Void> result) {
+    if (isVersionAllowed()) {
+      ShortcutManager shortcutManager =
+          (ShortcutManager) context.getSystemService(Context.SHORTCUT_SERVICE);
+      shortcutManager.removeAllDynamicShortcuts();
+      result.success(null);
+    }
```
stuartmorgan, line 103: This has the same never-replying bug. You can avoid it even being possible by making this method non-asnyc in the pigeon definition file. (That's a main reason we have the non-async version; it's just way less error-prone.)

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
```diff
@@ -0,0 +1,183 @@
...
+      }
+      final Intent intent = activity.getIntent();
+      final String launchAction = intent.getStringExtra(EXTRA_ACTION);
+      if (launchAction != null && !launchAction.isEmpty()) {
+        shortcutManager.reportShortcutUsed(launchAction);
+        intent.removeExtra(EXTRA_ACTION);
+      }
+      result.success(launchAction);
+    }
+    return;
```
stuartmorgan, line 127: Same here; looks like this can also be non-async.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  @TargetApi(Build.VERSION_CODES.N_MR1)
+  private List<ShortcutInfo> ShortcutItemMessageToShortcutInfo(
```
stuartmorgan, line 131: Nit: camel case for Java methods.

stuartmorgan (changes requested):

packages/quick_actions/quick_actions_android/pigeons/messages.dart
tarrinneal: Can't blame me for this one, I just moved it to the other file, lol

stuartmorgan: That was probably me, duplicating code during conversion of plugins to fully federated.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActionsPlugin.java
tarrinneal: it isn't suggesting anything, so 🤷

packages/quick_actions/quick_actions_android/pigeons/messages.dart
tarrinneal: actually, I'm to blame. This is copied from the platform interface.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
stuartmorgan: This method is inherently asynchronous due to the thread bounce, so it still needs to be declared as async. The version you have now will return success immediately no matter what, and the later `throw` would not work correctly (maybe crash the app? I forget how unhandled throws work in Java).

Since the thread bounce here is [to keep `setDynamicShortcuts` off of the main thread](https://github.com/flutter/plugins/pull/4502#discussion_r750511289) we could simplify this a lot by removing both of them and using background channels for the plugin (the linked discussion saying it would break the platform interface is no longer true due to the change to in-package method channels, so it should be easy), but in the spirit of minimal changes I would make that part of the TODO below (i.e., change it to "TODO(camsim99): Investigate removing all of the executor logic in favor of background channels.") and keep the current implementation for now, with this method still `async`.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
```diff
@@ -0,0 +1,192 @@
...
+
+  @ChecksSdkIntAtLeast(api = Build.VERSION_CODES.N_MR1)
+  boolean isVersionAllowed() {
+    return Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1;
+  }
+
+  @Override
+  public void setShortcutItems(@NonNull List<ShortcutItemMessage> itemsList) {
+    // We already know that this functionality does not work for anything
+    // lower than API 25 so we chose not to return error. Instead we do nothing.
```
stuartmorgan, line 57: Actually, I think we could rephrase this and move it to the helper method, rather than repeating it everywhere. E.g.,

```
Returns true when running on a version of Android that supports quick actions.

When this returns false, methods should silently no-op, per the documented behavior (see README.md).
```

packages/quick_actions/quick_actions_android/lib/quick_actions_android.dart
```diff
@@ -3,54 +3,73 @@
...
-        await channel.invokeMethod<String?>('getLaunchAction');
+    _handler = handler;
+    final String? action = await _hostApi.getLaunchAction();
     if (action != null) {
-      handler(action);
+      _handler(action);
     }
+    final _QuickActionHandlerApi quickActionsHandlerApi =
+        _QuickActionHandlerApi();
+    AndroidQuickActionsFlutterApi.setup(quickActionsHandlerApi);
```
stuartmorgan, line 37: Nit: I would keep the original order (set up the channel, then do the logic to get any queued action) since it seems more logical to do setup tasks first.

packages/quick_actions/quick_actions_android/lib/quick_actions_android.dart
```diff
@@ -3,54 +3,73 @@
...
+  ShortcutItemMessage _shortcutItemToShortcutItemMessage(ShortcutItem item) {
+    return ShortcutItemMessage(
+      type: item.type,
+      localizedTitle: item.localizedTitle,
+      icon: item.icon,
+    );
+  }
+
+  /// Conversion tool to change [ShortcutItemMessage] back to [ShortcutItem]
+  @visibleForTesting
```
stuartmorgan, line 59: It looks like this is only called in tests, so it could just be in the test file.

packages/quick_actions/quick_actions_android/test/quick_actions_android_test.dart
```diff
@@ -4,169 +4,100 @@
...
+        items.add(
+            QuickActionsAndroid.shortcutItemMessageToShortcutItem(element));
+      } else {
+        throw Exception('ShortcutItemMessage unexpectedly null');
```
stuartmorgan, line 99: Optional: you could skip the if/else logic and just do `element!`, since this exception is basically saying the same thing that a failed force-unwrap exception would say.

packages/quick_actions/quick_actions_android/android/src/main/java/io/flutter/plugins/quickactions/QuickActions.java
tarrinneal: Is this still an issue? I'm not certain where this was a problem originally, as I force pushed over the original code. It seems to me that all paths reply currently.

tarrinneal: oh, this was part of the change that I reverted already.

stuartmorgan (approved):
LGTM!

PR #5123, opened 2023-10-12
[gis_web] Adds FedCM toggle.

ditman:
This PR updates the `IdConfiguration` object of the Google Identity Services SDK with new values added since this was last updated.

The main change is adding `use_fedcm_for_prompt` so [FedCM can be enabled](https://developers.google.com/identity/gsi/web/guides/fedcm-migration) on Flutter Web apps.

### Issues

Pre-requirement for: https://github.com/flutter/flutter/issues/133703

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_identity_services_web/example/integration_test/js_interop_id_test.dart
```diff
@@ -31,6 +31,27 @@ void main() async {
     });
   });
 
+  group('initialize', () {
+    testWidgets('can receive `use_fedcm_for_prompt` boolean', (_) async {
+      utils.expectFedCMConfig();
+
+      expect(() {
+        id.initialize(IdConfiguration(client_id: 'testing_1-2-3'));
+      }, throwsA(isA<String>()));
```
ditman, line 40: This is not testing the real behavior of `initialize`, but that our mock pukes if we don't pass the [`use_fedcm_for_prompt`](https://developers.google.com/identity/gsi/web/reference/js-reference#use_fedcm_for_prompt) configuration value.

(Another test might have seen to just extract `use_fedcm_for_prompt` from a freshly created `IdConfiguration` object, without changing the behavior of the mock (?)

stuartmorgan: Can you just add a comment to this `expect` explaining that in the code, since it's a bit mysterious otherwise?

packages/google_identity_services_web/lib/src/js_interop/google_accounts_id.dart
```diff
@@ -230,12 +230,40 @@ abstract class IdConfiguration {
...
+    /// If your application knows in advance which user should be signed-in, it
+    /// can provide a login hint to Google.
+    ///
+    /// When successful, account selection is skipped. Accepted values are:
+    ///   * an email address or
+    ///   * an ID token sub field value.
+    ///
+    /// For more information, see:
+    ///   * https://developers.google.com/identity/protocols/oauth2/openid-connect#authenticationuriparameters
+    String? login_hint,
```
stuartmorgan, line 249: The ship has clearly already sailed on this, but out of curiosity why are we using snake case instead of camel case for all of these, given that it's Dart code?

packages/google_identity_services_web/lib/src/js_interop/google_accounts_id.dart
```diff
@@ -230,12 +230,40 @@ abstract class IdConfiguration {
...
+    ///
+    /// When successful, user accounts displayed during account selection are
+    /// limited to the provided domain.
+    ///
+    /// A wildcard value: `*` offers only Workspace accounts to the user and
+    /// excludes consumer accounts (user@gmail.com) during account selection.
+    ///
+    /// For more information, see:
+    ///   * https://developers.google.com/identity/protocols/oauth2/openid-connect#authenticationuriparameters
+    String? hd,
```
stuartmorgan, line 262: Shouldn't we have some test (or at least example) code using `login_hint` and `hd` somewhere? If we removed the fields it doesn't look like anything would break.

stuartmorgan (changes requested):

packages/google_identity_services_web/lib/src/js_interop/google_accounts_id.dart
ditman: > why are we using snake case instead of camel case for all of these, given that it's Dart code?

Originally, my idea was that it'd be easier for users to find documentation for the parameters if they were called exactly the same as in the underlying JS SDK (see also some of the [exotic names of the WebGL API in package:web](https://github.com/dart-lang/web/blob/main/lib/src/dom/ext_blend_minmax.dart)).

If we prefer `camelCase` in the Dart side of things, it can definitely be done:

```dart
@JS('login_hint')
String? loginHint,
```

And it wouldn't be a hard migration, after all I suspect it's only [us using this package directly](https://pub.dev/packages?q=dependency%3Agoogle_identity_services_web+show%3Aunlisted).

Let me know if you want this camelCase'd, I can create a task and go for it later :)

packages/google_identity_services_web/lib/src/js_interop/google_accounts_id.dart
ditman: > If we removed the fields it doesn't look like anything would break.

This is true of many of the other fields available in this object. Let me create a simplified test so I can assert that all the possible fields are passed through, and have the expected type and/or value.

packages/google_identity_services_web/example/integration_test/js_interop_id_test.dart
ditman: I will change this test a little bit, so we can assert that constructing a `IdConfiguration` results in a JS Object that has the expected properties, so I can test all of them, rather than doing this weird thing.

ditman:
Please PTAL @stuartmorgan when you find it convenient. Thanks!

packages/google_identity_services_web/lib/src/js_interop/google_accounts_id.dart
stuartmorgan: I don't know that it's worth doing a breaking change for, it just struck me as odd so I was curious.

stuartmorgan (approved):
LGTM (once the bots are happy). Thanks for the new tests!

ditman:
`warning - example/integration_test/js_interop_id_test.dart:11:63 - The name allowInterop is shown, but isn't used. Try removing the name from the list of shown members. - unused_shown_name`

woopsie!

PR #5155, opened 2023-10-16
[url_launcher] Add an `inAppBrowserView` mode

stuartmorgan:
`url_launcher_android` recently switched from an in-app webview to an Android Custom Tab (when supported), which was intended to be an in-place upgrade. However, this broke `closeInAppWebView`, and that couldn't be fixed directly because Android Custom Tab has no mechanism for programatic close. To address the regression, this adds a new `inAppBrowserView` launch mode which is distinct from `inAppWebView`, so that use cases that require programatic close can specifically request `inAppWebView` instead. The default for web links is the new `inAppBrowserView` since that gives better results in most cases.

Since whether `closeInAppWebView` will work in any given case is now non-trivial (on iOS, both in-app modes are supported, but on Android it's only the web view mode), this adds a new support API to query it, in keeping with the relatively new guidance of https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#api-support-queries. It also adds API to query for support for being able to use specific launch modes, since there wasn't a good way to understand which modes worked in general on different platforms.

Since there are new APIs, this adds support for those APIs to all of our implementations to ensure that they give accurate responses.

Fixes https://github.com/flutter/flutter/issues/134208

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/url_launcher/url_launcher/lib/src/url_launcher_uri.dart
```diff
@@ -11,25 +11,13 @@ import 'type_conversion.dart';
...
-///   - [LaunchMode.externalNonBrowserApplication] is supported on iOS 10+.
-///     This setting is used to require universal links to open in a non-browser
-///     application.
+/// [mode] support varies significantly by platform. Clients can use
+/// [supportsLaunchMode] to query for support, but platforms will fall back to
+/// other modes if the requested mode is not supported, so checking is not
+/// required. The default behavior of [LaunchMode.platformDefault] is up to each
+/// platform, and its behavior for a given platform may change over time as new
+/// modes are supported, so clients that want a specific mode should request it
+/// rather than rely on any currently observed default behavior.
```
stuartmorgan, line 20: This makes the documentation federation-friendly and evergreen. If we find that people want docs about specific platform behaviors, we can add it to the platform package READMEs later.

packages/url_launcher/url_launcher_android/lib/url_launcher_android.dart
```diff
@@ -62,16 +60,57 @@ class UrlLauncherAndroid extends UrlLauncherPlatform {
...
+                ? PreferredLaunchMode.inAppWebView
+                : PreferredLaunchMode.externalApplication,
+            webViewConfiguration: InAppWebViewConfiguration(
+                enableDomStorage: enableDomStorage,
+                enableJavaScript: enableJavaScript,
+                headers: headers)));
+  }
+
+  @override
+  Future<bool> launchUrl(String url, LaunchOptions options) async {
```
stuartmorgan, line 76: The reason there's so much change here is that we never went back and implemented `launchUrl`, so we only had the legacy `launch` platform interface method. This adds the entire method so that we can support `inAppBrowserView`.

(That could be done as a prequel without the new mode, but it didn't seem worth splitting out just that small piece; if anyone would prefer that let me know.)

packages/url_launcher/url_launcher_android/test/url_launcher_android_test.dart
```diff
@@ -197,12 +198,198 @@ void main() {
...
+    });
+
+    test('uses external for other schemes', () async {
+      final UrlLauncherAndroid launcher = UrlLauncherAndroid(api: api);
+      final bool launched = await launcher.launchUrl(
+          'supportedcustomscheme://example.com/', const LaunchOptions());
+      expect(launched, true);
+      expect(api.usedWebView, false);
+    });
+  });
```
stuartmorgan, line 334: Everything from here up is pretty much copies of the existing legacy (`launch`) tests, but using `launchUrl`.

packages/url_launcher/url_launcher_ios/lib/url_launcher_ios.dart
```diff
@@ -45,11 +45,79 @@ class UrlLauncherIOS extends UrlLauncherPlatform {
...
+        LaunchOptions(
+            mode: mode,
+            webViewConfiguration: InAppWebViewConfiguration(
+                enableDomStorage: enableDomStorage,
+                enableJavaScript: enableJavaScript,
+                headers: headers)));
+  }
+
+  @override
+  Future<bool> launchUrl(String url, LaunchOptions options) async {
```
stuartmorgan, line 68: Technically this isn't required for this PR, but it seemed like a good opportunity to add `launchUrl` since it's very similar to the Android version. I can make this a separate PR later if preferred.

packages/url_launcher/url_launcher_ios/test/url_launcher_ios_test.dart
```diff
@@ -134,11 +135,171 @@ void main() {
...
+
+    test('uses standard external for other schemes', () async {
+      final UrlLauncherIOS launcher = UrlLauncherIOS(api: api);
+      final bool launched = await launcher.launchUrl(
+          'supportedcustomscheme://example.com/', const LaunchOptions());
+      expect(launched, true);
+      expect(api.usedSafariViewController, false);
+      expect(api.passedUniversalLinksOnly, false);
+    });
+  });
```
stuartmorgan, line 240: As with Android tests, the above is copies of the existing tests but with the non-deprecated API.

packages/url_launcher/url_launcher_linux/lib/url_launcher_linux.dart
```diff
@@ -51,4 +51,16 @@ class UrlLauncherLinux extends UrlLauncherPlatform {
       },
     ).then((bool? value) => value ?? false);
   }
+
```
stuartmorgan, line 54: I didn't add `launchUrl` to the desktop platforms in this PR because they don't currently support any mode switches, so it didn't seem as useful. (Someday I'll update them just so we can deprecate `launch`.)

packages/url_launcher/url_launcher_web/lib/url_launcher_web.dart
```diff
@@ -90,4 +90,17 @@ class UrlLauncherPlugin extends UrlLauncherPlatform {
     openNewWindow(url, webOnlyWindowName: webOnlyWindowName);
     return true;
   }
+
+  @override
+  Future<bool> supportsMode(PreferredLaunchMode mode) async {
+    // Web doesn't allow any control over the destination beyond
+    // webOnlyWindowName, so don't claim support for any mode beyond default.
+    return mode == PreferredLaunchMode.platformDefault;
```
ditman, line 98: These are good defaults for the web platform. This maybe could be expanded more (if you use window.open, and keep a reference to the window that you opened, it can be [`close`](https://developer.mozilla.org/en-US/docs/Web/API/Window/close)d, for example), but so far I have never heard from anybody wanting the "close" feature on the web.

ditman (approved):
Web bits LGTM!

packages/url_launcher/url_launcher/lib/src/url_launcher_uri.dart
ditman: Per-platform docs are the way, agreed!

packages/url_launcher/url_launcher_linux/example/pubspec.yaml
```diff
@@ -26,3 +26,8 @@ dev_dependencies:
 
 flutter:
   uses-material-design: true
+
+# FOR TESTING AND INITIAL REVIEW ONLY. DO NOT MERGE.
+# See https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins
+dependency_overrides:
+   {url_launcher_platform_interface: {path: ../../../url_launcher/url_launcher_platform_interface}}
```
cbracken, line 33: Here and elsewhere, I assume this will be backed out?

cbracken (approved):
lgtm modulo removal of `FOR TESTING AND INITIAL REVIEW` bits.

packages/url_launcher/url_launcher_linux/example/pubspec.yaml
stuartmorgan: Yes, this is the combo PR. Once it's approved I'll break it into three and land then in order, and these become constraint updates on the new version.

stuartmorgan:
https://github.com/flutter/packages/pull/5205 splits out the first part for landing.

stuartmorgan:
The sub-PRs have landed and published, and this has been updated to set the implementation package constraints to require versions that have the implementations of the new methods.

PR #5159, opened 2023-10-17
[quick_actions] convert to pigeon

tarrinneal:
Converts from direct use of method channels to using Pigeon.

Part of https://github.com/flutter/flutter/issues/117844
fixes https://github.com/flutter/flutter/issues/117913
fixes https://github.com/flutter/flutter/issues/117112
## Pre-launch Checklist

- [x] I signed the [CLA].


packages/quick_actions/quick_actions_ios/test/quick_actions_ios_test.dart
```diff
@@ -2,171 +2,89 @@
...
-          <Matcher>[
-            isMethodCall('getLaunchAction', arguments: null),
-          ],
-        );
-        log.clear();
-
-        expect(quickActionsHandler.future, completion(isTrue));
-      });
-    });
+  test('setShortCutItems', () async {
```
stuartmorgan, line 29: Nit: we can fix the capitalization here, since at this layer it's correctly `Shortcut`.

packages/quick_actions/quick_actions_ios/lib/messages.g.dart
```diff
@@ -0,0 +1,160 @@
...
+    switch (type) {
+      case 128:
+        return ShortcutItemMessage.decode(readValue(buffer)!);
+      default:
+        return super.readValueOfType(type, buffer);
+    }
+  }
+}
+
+class IosQuickActionsApi {
```
stuartmorgan, line 72: Nit: IOS; see https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#capitalize-identifiers-consistent-with-their-spelling

packages/quick_actions/quick_actions_ios/ios/Classes/ShortcutItemParser.swift
stuartmorgan, line 1: We should probably eliminate this whole class and its mock, since it's just doing a trivial conversion at this point. I'm fine with that being done in some later PR though.

packages/quick_actions/quick_actions_ios/example/ios/RunnerTests/Mocks/MockBinaryMessenger.swift
```diff
@@ -0,0 +1,40 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import Flutter
+
+@testable import quick_actions_ios
+
+class MockBinaryMessenger: NSObject, FlutterBinaryMessenger {
```
stuartmorgan, line 9: The need for this is exposing a gap in the Swift generator; could you leave TODO to remove this when https://github.com/flutter/flutter/issues/136811 is implemented?

packages/quick_actions/quick_actions_ios/example/ios/RunnerTests/QuickActionsPluginTests.swift
```diff
@@ -287,8 +228,6 @@ class QuickActionsPluginTests: XCTestCase {
 
     plugin.applicationDidBecomeActive(UIApplication.shared)
     waitForExpectations(timeout: 1)
-
-    XCTAssertEqual(invokeMehtodCount, 1, "shortcut should only be handled once per launch.")
```
stuartmorgan, line 291: Wasn't this a critical part of the test?

packages/quick_actions/quick_actions_ios/example/ios/RunnerTests/QuickActionsPluginTests.swift
```diff
@@ -143,28 +90,28 @@ class QuickActionsPluginTests: XCTestCase {
       userInfo: nil)
 
     let invokeMethodExpectation = expectation(description: "invokeMethod must be called.")
-    mockChannel.invokeMethodStub = { method, arguments in
-      XCTAssertEqual(method, "launch")
-      XCTAssertEqual(arguments as? String, item.type)
```
stuartmorgan, line 148: Here and later, it seems like we're losing important test coverage; shouldn't we still be asserting that the right type is provided?

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
```diff
@@ -80,12 +72,14 @@ public final class QuickActionsPlugin: NSObject, FlutterPlugin {
...
       handleShortcut(shortcutType)
-      launchingShortcutType = nil
+      self.launchingShortcutType = nil
     }
   }
 
-  private func handleShortcut(_ shortcut: String) {
-    channel.invokeMethod("launch", arguments: shortcut)
+  func handleShortcut(_ shortcut: String) {
+    self.testStub?()
```
stuartmorgan, line 80: Is this to avoid relying on the Pigeon serialization details in the mock binary messenger?

If so, maybe we could instead do that for now since the TODO I requested above would address that in a better way long term, and we wouldn't need this test hack.

stuartmorgan (changes requested):

packages/quick_actions/quick_actions_ios/example/ios/RunnerTests/QuickActionsPluginTests.swift
```diff
@@ -58,80 +55,30 @@ class QuickActionsPluginTests: XCTestCase {
...
-    let resultExpectation = expectation(description: "result block must be called.")
-    plugin.handle(call) { result in
-      XCTAssertNil(result, "result block must be called with nil.")
-      resultExpectation.fulfill()
-    }
-
-    waitForExpectations(timeout: 1)
-  }
-
-  func testHandleMethodCall_nonExistMethods() {
```
hellohuanlin, line 104: why are we removing these tests? can you run a coverage report in xcode? IIRC it used to be 100%

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
```diff
@@ -4,48 +4,40 @@
...
-    case "clearShortcutItems":
-      shortcutItemProvider.shortcutItems = []
-      result(nil)
-    case "getLaunchAction":
-      result(nil)
-    case _:
-      result(FlutterMethodNotImplemented)
-    }
+  /// Method to allow for async testing.
+  var testStub: (() -> Void)? = nil
```
hellohuanlin, line 33: why do we have this in non test code?

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
```diff
@@ -4,48 +4,40 @@
...
+
+  func setShortcutItems(itemsList: [ShortcutItemMessage]) throws {
+    shortcutItemProvider.shortcutItems = shortcutItemParser.parseShortcutItems(itemsList)
+  }
+
+  func clearShortcutItems() throws {
```
hellohuanlin, line 39: why do these 2 functions throw?

packages/quick_actions/quick_actions_ios/ios/Classes/ShortcutItemParser.swift
hellohuanlin: Yeah the actual parsing now happens in pigeon, so this becomes trivial.

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
hellohuanlin: Oh IC `IosQuickActionsFlutterApi` is a class rather than protocol. 

A workaround can be to wrap it with your own types: 

```
protocol IosQuickActionsFlutterApiWrapper {
  func launchAction(action: ...)
}

final class DefaultIosQuickActionsFlutterApiWrapper: IosQuickActionsFlutterApiWrapper {
  private let api: IosQuickActionsFlutterApi
  func launchAction(action: ...) {
    api.launchAction(action: action)
  }
}
```

Drawback is that this trivial wrapper itself is not unit testable in swift, but it should already be covered in dart tests.

packages/quick_actions/quick_actions_ios/example/ios/RunnerTests/QuickActionsPluginTests.swift
stuartmorgan: You can't call a method that doesn't exist via Pigeon APIs, so the test is no longer meaningful.

packages/quick_actions/quick_actions_ios/example/ios/RunnerTests/DefaultShortcutItemParserTests.swift
```diff
@@ -44,24 +43,4 @@ class DefaultShortcutItemParserTests: XCTestCase {
     let parser = DefaultShortcutItemParser()
     XCTAssertEqual(parser.parseShortcutItems([rawItem]), [expectedItem])
   }
-
-  func testParseShortcutItems_noType() {
```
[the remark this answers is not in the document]
tarrinneal: Shortcut items can't have no type

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
```diff
@@ -4,48 +4,45 @@
...
-      result(FlutterMethodNotImplemented)
-    }
+  convenience init(
+    messenger: FlutterBinaryMessenger,
+    shortcutItemProvider: ShortcutItemProviding = UIApplication.shared
+  ) {
+    let shortcutItemProvider = shortcutItemProvider
+    let flutterApi = IOSQuickActionsFlutterApi(binaryMessenger: messenger)
+    self.init(flutterApi: flutterApi, shortcutItemProvider: shortcutItemProvider)
+  }
```
tarrinneal, line 38: Not sure if this is the best way to do this, but I was having a hard time figuring out a better way.

tarrinneal: Xcode insists this needs to be a convenience init

packages/quick_actions/quick_actions_ios/example/ios/RunnerTests/Mocks/MockBinaryMessenger.swift
stuartmorgan, line 1: This file can be removed now, right?

packages/quick_actions/quick_actions_ios/example/ios/RunnerTests/QuickActionsPluginTests.swift
```diff
@@ -7,47 +7,43 @@ import XCTest
 
 @testable import quick_actions_ios
 
+class MockFlutterApi: IOSQuickActionsFlutterApiProtocol {
+  /// Method to allow for async testing.
+  var testStub: ((String) -> Void)? = nil
```
stuartmorgan, line 12: How about `launchActionCallback` or something similar, so it's more descriptive?

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
```diff
@@ -4,48 +4,45 @@
...
-    case "getLaunchAction":
-      result(nil)
-    case _:
-      result(FlutterMethodNotImplemented)
-    }
+  convenience init(
+    messenger: FlutterBinaryMessenger,
+    shortcutItemProvider: ShortcutItemProviding = UIApplication.shared
+  ) {
+    let shortcutItemProvider = shortcutItemProvider
```
stuartmorgan, line 35: What is the purpose of this line?

stuartmorgan (approved):
LGTM with nits. Thanks for doing that Pigeon PR! The tests are much cleaner now, and that was definitely a good example of why it's valuable for Pigeon development for us to adopt it :)

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
tarrinneal: This is the most important line in the whole file, clearly

tarrinneal:
@hellohuanlin I would love to get an approval from you, since I'm still not 100% on Swift etiquette.

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
```diff
@@ -4,48 +4,44 @@
...
 
-  private let channel: MethodChannel
   private let shortcutItemProvider: ShortcutItemProviding
-  private let shortcutItemParser: ShortcutItemParser
+  private let shortcutItemParser: ShortcutItemParser = DefaultShortcutItemParser()
+  private let flutterApi: IOSQuickActionsFlutterApiProtocol
   /// The type of the shortcut item selected when launching the app.
   private var launchingShortcutType: String? = nil
 
+  // This init is meant for unit testing only.
```
hellohuanlin, line 22: A comment isn't good enough to prevent it from being misused in non-unit test scenario. I suggest just keep this init function, and remove the convenient init below, and then inject `IOSQuickActionsFlutterApi(binaryMessenger: messenger)` from `static func register(with:)`.

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
```diff
@@ -80,12 +76,13 @@ public final class QuickActionsPlugin: NSObject, FlutterPlugin {
   public func applicationDidBecomeActive(_ application: UIApplication) {
     if let shortcutType = launchingShortcutType {
       handleShortcut(shortcutType)
-      launchingShortcutType = nil
+      self.launchingShortcutType = nil
```
hellohuanlin, line 79: let's keep it consistent - either always use `self.`, or always omit `self.`

packages/quick_actions/quick_actions_ios/ios/Classes/ShortcutItemParser.swift
```diff
@@ -12,26 +12,24 @@ protocol ShortcutItemParser {
   /// - Parameter items an array of raw shortcut items to be parsed.
   /// - Returns an array of parsed shortcut items to be set.
   ///
-  func parseShortcutItems(_ items: [[String: Any]]) -> [UIApplicationShortcutItem]
+  func parseShortcutItems(_ items: [ShortcutItemMessage]) -> [UIApplicationShortcutItem]
```
hellohuanlin, line 15: can you add a TODO to remove this parser since the parsing logic is handled by pigeon now.

tarrinneal: It isn't entirely, since this converts the `ShortcutItem` to a `UIApplicationShortcutItem`.

tarrinneal: I removed it instead. Probably worth one more look through

packages/quick_actions/quick_actions_ios/ios/Classes/QuickActionsPlugin.swift
```diff
@@ -75,14 +66,38 @@ public final class QuickActionsPlugin: NSObject, FlutterPlugin, IOSQuickActionsA
...
   }
 
   func handleShortcut(_ shortcut: String) {
-    flutterApi.launchAction(action: shortcut) { _ in
+    self.flutterApi.launchAction(action: shortcut) { _ in
       // noop
     }
   }
+
+  static func parseShortcutItems(_ items: [ShortcutItemMessage]) -> [UIApplicationShortcutItem] {
```
hellohuanlin: this doesn't seem to belong to a static member of `QuickActionsPlugin`. I suggest 2 alternatives - 

1. make them private non-static members of `QuickActionsPlugin` (so they become implementation details `QuickActionsPlugin`). 

2. A more swifty way I would do is to create extension of either `UIApplicationShortcutItem` or `ShortcutItemMessage`.

tarrinneal: Sounds good, thanks for the feedback!

hellohuanlin: Oh also avoid the naming "parse" since it's not parsing anymore. Something like: 

```
extension ShortcutItemMessage {
  var item: UIApplicationShortcutItem {
    return ...
  }
}
```

tarrinneal: I don't think we need the tests that I was using this for any longer, since this is now an implementation detail and my understanding is that the swifty thing to do is consider the converter tested, since it is used in code that is already tested.

hellohuanlin: Yes if it's implementation detail, it can only be tested via non-private members. You can delete those tests if it's already covered elsewhere

hellohuanlin (approved):

PR #5180, opened 2023-10-19
[url_launcher_web] Disallows launching "javascript:" URLs.

ditman:
This PR updates the URL launcher to:

* Implement the new `launchUrl` method.
* Disallow `launch` and `launchUrl` for URLs with the `javascript:` scheme. 
* Prevent _Tabnabbing_.

### Issues

Fixes https://github.com/flutter/flutter/issues/136657

### Tests

Integration tests (and mocks) updated.

## Pre-launch Checklist

- [x] I signed the [CLA].


mdebbar (commented):
The code itself looks good to me, though I'm hesitant about blanket-adding `noopener,noreferrer`. I'm okay with making this the default behavior. But should we give developers the option to allow communication back with the main window? There could be a legit use case when the app is opening its own popup (not a user-provided URL).

stuartmorgan:
> I'm hesitant about blanket-adding `noopener,noreferrer`. I'm okay with making this the default behavior. But should we give developers the option to allow communication back with the main window? There could be a legit use case when the app is opening its own popup (not a user-provided URL).

I had the same thought initially, but then wasn't convinced that `url_launcher` should be the mechanism for that use case. It's not intended to be a window management utility, and doesn't work that way on any other platform, so I think there's a good argument that if what you want to open is a page that's communicating with the main window, `url_launcher` isn't the plugin for you. Maybe `Link`'s use of `url_launcher` changes that dynamic though? Should `Link` be less tightly bound to allow use of `Link` in those cases without using `openUrl`, maybe?

ditman:
> But should we give developers the option to allow communication back with the main window? There could be a legit use case when the app is opening its own popup (not a user-provided URL).

@mdebbar it's easy for us to pass a new `LaunchOptions` property to disable the "noreferrer, noopener" in case this breaks any user of the feature.

My two opinions, hoever:

* if you want to call `window.open` without the noopener rule, you can write your own code to do so (for example: `google_sign_in` handles its own popup).
* if you need to talk to a popup you opened, you should be using `postMessage`, so you can have a protocol that you can validate.

(Do you really think this change is a big breaking change?)

> Should Link be less tightly bound to allow use of Link in those cases without using openUrl, maybe?

@stuartmorgan hmmm, I think `Link` does not use `openUrl` on the web? (but @mdebbar can confirm!) ([code](https://github.com/flutter/packages/blob/353e9a2f951d10a2c2443a3ce0c62904ed62edc8/packages/url_launcher/url_launcher_web/lib/src/link.dart#L190-L212))

stuartmorgan (approved):
> @stuartmorgan hmmm, I think `Link` does not use `openUrl` on the web

[Doh!](https://github.com/flutter/packages/blob/353e9a2f951d10a2c2443a3ce0c62904ed62edc8/packages/url_launcher/url_launcher/lib/src/link.dart#L83)

This LGTM then. I would at this point vote for only adding options to remove the attributes if we get significant feedback that there's a compelling case for them in the context of the intended use of `url_launcher`.

ditman:
> I would at this point vote for only adding options to remove the attributes if we get significant feedback that there's a compelling case for them in the context of the intended use of `url_launcher`.

I agree, but will wait to see how strongly @mdebbar feels about it!

mdebbar (approved):
LGTM. Keeping an eye for user feedback sounds like a good strategy!

ditman:
Applying `autosubmit`

PR #5182, opened 2023-10-19
[tool] Fix `filter-packages-to` when everything is changed

stuartmorgan:
`filter-packages-to` didn't correctly the handle the case where the set of target packages is empty, meaning that all packages should be tested. This broke it for cases such as changing a CI configuration file, making the filter not take effect.

### DIFF
--- a/script/tool/lib/src/common/package_command.dart
+++ b/script/tool/lib/src/common/package_command.dart
@@ -436,16 +436,26 @@ abstract class PackageCommand extends Command<void> {
       packages = <String>{currentPackageName};
     }
 
-    Set<String> excludedPackageNames = getExcludedPackageNames();
-
-    final Set<String> filter =
-        _expandYamlInPackageList(getStringListArg(_filterPackagesArg));
-    if (filter.isNotEmpty) {
-      final List<String> sortedList = filter.toList()..sort();
+    final Set<String> excludedPackageNames = getExcludedPackageNames();
+    final bool hasFilter = argResults?.wasParsed(_filterPackagesArg) ?? false;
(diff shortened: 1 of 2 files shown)


script/tool/lib/src/common/package_command.dart
```diff
@@ -436,16 +436,26 @@ abstract class PackageCommand extends Command<void> {
...
+    if (excludeAllButPackageNames != null &&
+        excludeAllButPackageNames.isNotEmpty) {
+      final List<String> sortedList = excludeAllButPackageNames.toList()
+        ..sort();
       print('--$_filterPackagesArg is excluding packages that are not '
           'included in: ${sortedList.join(',')}');
-      excludedPackageNames =
-          excludedPackageNames.union(packages.difference(filter));
+    }
+    // Utility to return true if a package that could be identified by any of
```
tarrinneal, line 451: This is funky, but not a deal breaker imo

tarrinneal (approved):

script/tool/lib/src/common/package_command.dart
stuartmorgan: I don't tend to use local functions very often, but they can be nice for small local code sharing.

tarrinneal: I meant the comment itself, sorry. I thought my highlight would come through. 

> Utility to return true

ricardoamador (approved):

script/tool/lib/src/common/package_command.dart
stuartmorgan: Oh, I see. Agreed; changed to just "Returns true if [...]".

PR #5210, opened 2023-10-24
[shared_preferences] full api redesign with DataStore and cache-less interface

tarrinneal:
Full rework for shared_preferences, allows user to decide whether to use cache or be fully async and pull direct from the platform.

fixes https://github.com/flutter/flutter/issues/123078
fixes https://github.com/flutter/flutter/issues/133098
fixes https://github.com/flutter/flutter/issues/65145
fixes https://github.com/flutter/flutter/issues/64739
fixes https://github.com/flutter/flutter/issues/151036


packages/shared_preferences/shared_preferences/lib/better_shared_preference.dart
```diff
@@ -0,0 +1,276 @@
...
+  /// Reads a value from the cache, throwing an exception if it's not a
+  /// double.
+  double? getDouble(String key) => get(key) as double?;
+
+  /// Reads a value from the cache, throwing an exception if it's not a
+  /// String.
+  String? getString(String key) => get(key) as String?;
+
+  /// Reads a set of string values from the cache, throwing an
+  /// exception if it's not a string set.
```
tarrinneal, line 217: nit to self: this is a list of strings, not a set

tmpfs:
Can I please add to this that support for preferences separated by namespace would be *very useful*. My app supports multiple accounts and I want the preferences to be distinct for each account.

I could achieve this be prefixing every key but that is awkward and error prone. I then looked into using `setPrefix()` but I can't overwrite the prefix after `getInstance()` had been called which I need for my app.

If calling `setPrefix()` instead rebuilt (or switched) the cache then I think I would be able to use `shared_preferences`.

As it stands I have no choice but to roll my own preference handling which is a shame when `shared_preferences` is very nearly there for my requirements.

stuartmorgan:
> Can I please add to this that support for preferences separated by namespace would be _very useful_. My app supports multiple accounts and I want the preferences to be distinct for each account.

Please file a new issue to propose new functionality, so that discussion here can remain focused on the review.

tmpfs:
> > Can I please add to this that support for preferences separated by namespace would be _very useful_. My app supports multiple accounts and I want the preferences to be distinct for each account.
> 
> Please file a new issue to propose new functionality, so that discussion here can remain focused on the review.

@stuartmorgan issues are not enabled for this repository, should I open the issue on the main flutter repository?

stuartmorgan:
> @stuartmorgan issues are not enabled for this repository, should I open the issue on the main flutter repository?

https://github.com/flutter/packages#issues

packages/shared_preferences/shared_preferences/lib/better_shared_preference.dart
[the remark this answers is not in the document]
tarrinneal: That's just a placeholder for now.

packages/shared_preferences/shared_preferences/test/better_shared_preferences_test.dart
```diff
@@ -0,0 +1,444 @@
...
+
+  setUp(() async {
+    store = FakeSharedPreferencesStore(testValues);
+    SharedPreferencesStorePlatform.instance = store;
+    preferences = await SharedPreferences.getInstance();
+    store.log.clear();
+  });
+
+  test('reading', () async {
+    expect(preferences.get('String'), testString);
```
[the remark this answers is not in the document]
tarrinneal: Sorry, I didn't realize I had added this file, it's just a copy of the current test file. I haven't updated it yet.

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/com/flutter/plugins/shared_preferences/BetterMessages.kt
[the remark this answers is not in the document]
tarrinneal: The java file is no longer needed, I just hadn't removed it yet.

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,131 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
hellohuanlin, line 1: I'd call this `CachelessSharedPreferencePlugin`. it's hard to tell what`Better` means

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,131 @@
...
+
+  private func getUserDefaults(options: SharedPreferencesPigeonOptions) -> UserDefaults {
+    return UserDefaults(suiteName: options.suiteName)
+  }
+
+  func getKeys(allowList: [String]?, options: SharedPreferencesPigeonOptions) -> [String] {
+    return getAll(allowList: allowList, options: options).getKeys()
+  }
+
+  func getAll(allowList: [String]?, options: SharedPreferencesPigeonOptions) -> [String: Any] {
```
hellohuanlin, line 33: should this be private?

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,131 @@
...
+
+  func setDouble(key: String, value: Double, options: SharedPreferencesPigeonOptions) {
+    getUserDefaults(options: options).set(value, forKey: key)
+  }
+
+  func setValue(key: String, value: Any, options: SharedPreferencesPigeonOptions) {
+    getUserDefaults(options: options).set(value, forKey: key)
+  }
+
+  func getString(key: String, options: SharedPreferencesPigeonOptions) throws -> String? {
```
hellohuanlin, line 49: where does it throw?

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,131 @@
...
+  
+  func getBool(key: String, options: SharedPreferencesPigeonOptions) throws -> Bool? {
+    getUserDefaults(options: options).bool(forKey: key)
+  }
+  
+  func getDouble(key: String, options: SharedPreferencesPigeonOptions) throws -> Double? {
+    getUserDefaults(options: options).double(forKey: key)
+  }
+  
+  func getInt(key: String, options: SharedPreferencesPigeonOptions) throws -> Int64? {
```
hellohuanlin, line 61: should it be `getInt64`?

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,131 @@
...
+  func remove(key: String, options: SharedPreferencesPigeonOptions) {
+    getUserDefaults(options: options).removeObject(forKey: key)
+  }
+
+  func clear(allowList: [String]?, options: SharedPreferencesPigeonOptions) -> Bool {
+    let defaults = getUserDefaults(options: options)
+    for (key, _) in getAllPrefs(allowList: allowList) {
+      defaults.removeObject(forKey: key)
+    }
+    return true
```
hellohuanlin, line 78: what's the return value used for? It's always returning true here.

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,131 @@
...
+
+  /// Returns all preferences stored with specified prefix.
+  /// If [allowList] is included, only items included will be returned.
+  /// If no [allowList], returns supported types only.
+  func getAllPrefs(allowList: [String]?, options: SharedPreferencesPigeonOptions) -> [String: Any] {
+    var filteredPrefs: [String: Any] = [:]
+    var allowSet: Set<String>?
+    if let allowList = allowList {
+      allowSet = Set(allowList)
+    }
```
hellohuanlin, line 89: nit:  `let allowSet = arrowList.map { Set($0) }`

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,131 @@
...
+    {
+      if allowList == nil {
+        for (key, value) in prefs where (isTypeCompatible(value: value)) {
+          filteredPrefs[key] = value
+        }
+      } else {
+        for (key, value) in prefs where (allowSet!.contains(key)) {
+          filteredPrefs[key] = value
+        }
+      }
```
hellohuanlin, line 101: can avoid the force unwrapping by 
```
if let allowSet = allowSet {
} else {
}
```

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,131 @@
...
+    }
+    if let value = list as? [Object] {
+      let compatible = true
+      for (listItem) in list {
+        compatible = isTypeCompatible(value: listItem)
+        if !compatible {
+          return compatible
+        }
+      }
+      return compatible
```
hellohuanlin, line 127: can do `allSatisfy` instead of the loop

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
```diff
@@ -0,0 +1,131 @@
...
+      return true
+    }
+    if let value = double as? Double {
+      return true
+    }
+    if let value = string as? String {
+      return true
+    }
+    if let value = int as? Int {
+      return true
```
hellohuanlin, line 117: nit: can check the runtime type

```
switch type(of: value) {
  ...
}
```

packages/shared_preferences/shared_preferences_platform_interface/lib/better_shared_preferences_platform_interface.dart
```diff
@@ -0,0 +1,252 @@
...
+import 'better_types.dart';
+
+/// The interface that implementations of shared_preferences must implement.
+///
+/// Platform implementations should extend this class rather than implement it as `shared_preferences`
+/// does not consider newly added methods to be breaking changes. Extending this class
+/// (using `extends`) ensures that the subclass will get the default implementation, while
+/// platform implementations that `implements` this interface will be broken by newly added
+/// [BetterSharedPreferencesPlatform] methods.
+abstract base class BetterSharedPreferencesPlatform {
```
stuartmorgan, line 17: Do we need a completely new platform interface, or can we just add new methods to the old one? What are the pros/cons of a full split?

stuartmorgan (changes requested):

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
tarrinneal: This is called by the pigeon code

tarrinneal:
> We should probably start brainstorming actual names in parallel with the rest of the review. Maybe AsyncSharedPreferences and CachedSharedPreferences?

CachedSharedPreferences seems incorrect. SharedPreferencesWithCache? I'm fine with either.

packages/shared_preferences/shared_preferences/lib/better_shared_preference.dart
```diff
@@ -0,0 +1,269 @@
...
+  Future<bool> setDouble(String key, double value) =>
+      _platform.setDouble(key, value, options);
+
+  /// Saves a string [value] to the platform.
+  ///
+  /// Note: Due to limitations in Android's SharedPreferences,
+  /// values cannot start with any one of the following:
+  ///
+  /// - 'VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu'
+  /// - 'VGhpcyBpcyB0aGUgcHJlZml4IGZvciBEb3VibGUu'
```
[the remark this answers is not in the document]
tarrinneal: Should I remove the list of strings that aren't allowed?

packages/shared_preferences/shared_preferences/lib/better_shared_preference.dart
```diff
@@ -0,0 +1,269 @@
...
+  /// platform specific extensions of the [SharedPreferencesOptions] class.
+  final SharedPreferencesOptions sharedPreferencesOptions;
+
+  /// Options that define cache behavior.
+  final SharedPreferencesWithCacheOptions cacheOptions;
+
+  /// Async access directly to the platform.
+  ///
+  /// Methods called through [directAccess] will NOT update the cache.
+  final BetterSharedPreferences directAccess;
```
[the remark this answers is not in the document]
tarrinneal: Maybe I will replace it with a `getAsyncAndRefreshCache` type method? I think managing two separate `options` and prefs classes might be a bit more error prone than having direct access baked in.

packages/shared_preferences/shared_preferences_platform_interface/lib/better_shared_preferences_platform_interface.dart
tarrinneal: I'm hoping to keep them separate so that deprecation is simpler. I guess it may not matter in the end. I wrote most of the code as it's own file (here and on the platforms) so that people could get a better feeling for the new systems as they are without the old code dirtying the water.

I'm open to merging files if you think that's best.

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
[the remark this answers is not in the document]
tarrinneal: The new prefs not using `prefix` actually makes most of the code distinct.

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/BetterSharedPreferencesPlugin.swift
tarrinneal: I've been trying this to no avail, can you expand further?

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,274 @@
...
+  /// On platforms that do not support storing doubles,
+  /// the value will be stored as a float instead.
+  Future<bool> setDouble(String key, double value) async {
+    _cache[key] = value;
+    return directAccess.setDouble(key, value);
+  }
+
+  /// Saves a string [value] to the cache and platform.
+  ///
+  /// Note: Due to limitations on some platform's,
```
stuartmorgan, line 241: platforms

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,274 @@
...
+
+  /// Removes an entry from cache and platform.
+  Future<bool> remove(String key) async {
+    _cache.remove(key);
+    return directAccess.remove(key);
+  }
+
+  /// Clears cache and platform preferences that match filter options.
+  Future<bool> clear() async {
+    if (_cacheOptions.filter.allowList != null) {
```
stuartmorgan, line 265: Nit: consider swapping the blocks and making this `== null` so that the `else` isn't a double-negation.

packages/shared_preferences/shared_preferences/test/shared_preferences_async_test.dart
```diff
@@ -0,0 +1,617 @@
...
+  const List<String> testList = <String>['foo', 'bar'];
+
+  group('Async', () {
+    late FakeSharedPreferencesAsync store;
+    late SharedPreferencesAsync preferences;
+
+    setUp(() async {
+      store = FakeSharedPreferencesAsync();
+      SharedPreferencesAsyncPlatform.instance = store;
+      preferences =
```
stuartmorgan, line 33: See comments elsewhere about potentially using a helper returning a record to avoid the stateful test fixture.

packages/shared_preferences/shared_preferences/test/shared_preferences_async_test.dart
```diff
@@ -0,0 +1,617 @@
...
+          SharedPreferencesAsync(options: const SharedPreferencesOptions());
+      store.log.clear();
```
stuartmorgan, line 35: Surely a new store has a clear log already?

packages/shared_preferences/shared_preferences/test/shared_preferences_test.dart
```diff
@@ -47,7 +47,7 @@ void main() {
     store.log.clear();
   });
 
-  test('reading', () async {
+  testWidgets('reading', (WidgetTester _) async {
```
stuartmorgan, line 50: Same comment as in another file: why testWidgets in a unit test?

stuartmorgan (commented):
Still to do: Android, Linux, Web. But some comments apply generally, so posting these for now.

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/SharedPreferencesPlugin.swift
```diff
@@ -54,18 +54,144 @@ public class SharedPreferencesPlugin: NSObject, FlutterPlugin, UserDefaultsApi {
...
+    try SharedPreferencesPlugin.getUserDefaults(options: options).value(forKey: key) as! Bool?
+  }
+
+  func getDouble(key: String, options: SharedPreferencesPigeonOptions) throws -> Double? {
+    try SharedPreferencesPlugin.getUserDefaults(options: options).value(forKey: key) as! Double?
+  }
+
+  func getInt(key: String, options: SharedPreferencesPigeonOptions) throws -> Int64? {
+    try SharedPreferencesPlugin.getUserDefaults(options: options).value(forKey: key) as! Int64?
+  }
```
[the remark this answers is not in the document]
tarrinneal: depends, this way allows us to send back the typed value. And in the future, if needed, alter the individual methods.

packages/shared_preferences/shared_preferences_foundation/test/shared_preferences_foundation_test.dart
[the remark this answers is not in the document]
tarrinneal: how's that?

packages/shared_preferences/shared_preferences_windows/test/shared_preferences_windows_test.dart
```diff
@@ -76,13 +77,13 @@ void main() {
     return prefs;
   }
 
-  test('registered instance', () {
+  testWidgets('registered instance', (WidgetTester _) async {
```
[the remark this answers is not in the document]
tarrinneal: The tools yell at me that `test` is no good and only `testWidgets` is good enough. Maybe I misinterpreted it though

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,274 @@
...
+  /// platform specific extensions of the [SharedPreferencesOptions] class.
+  final SharedPreferencesOptions sharedPreferencesOptions;
+
+  /// Options that define cache behavior.
+  final SharedPreferencesWithCacheOptions _cacheOptions;
+
+  /// Async access directly to the platform.
+  ///
+  /// Methods called through [directAccess] will NOT update the cache.
+  final SharedPreferencesAsync directAccess;
```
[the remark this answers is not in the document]
tarrinneal: Conversation was started here: https://github.com/flutter/packages/pull/5210#discussion_r1412262733

Never was resolved though.

packages/shared_preferences/shared_preferences/test/shared_preferences_async_test.dart
tarrinneal: you never know these days

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/SharedPreferencesPlugin.swift
```diff
@@ -54,18 +54,144 @@ public class SharedPreferencesPlugin: NSObject, FlutterPlugin, UserDefaultsApi {
...
+      let messenger = registrar.messenger
+    #endif
+    UserDefaultsApiSetup.setUp(binaryMessenger: messenger, api: instance)
+    DeprecatedSharedPreferencesPlugin.register(with: registrar)
+  }
+
+  static private func getUserDefaults(options: SharedPreferencesPigeonOptions) throws
+    -> UserDefaults
+  {
+    let prefs = UserDefaults(suiteName: options.suiteName)
```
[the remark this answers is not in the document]
tarrinneal: so macOS can have any suite name?

packages/shared_preferences/shared_preferences_foundation/lib/shared_preferences_foundation.dart
```diff
@@ -2,106 +2,214 @@
...
+    @visibleForTesting UserDefaultsApi? api,
+  }) : _api = api ?? UserDefaultsApi();
+
+  final UserDefaultsApi _api;
+
+  /// Registers this class as the default instance of [SharedPreferencesAsyncPlatform].
   static void registerWith() {
-    SharedPreferencesStorePlatform.instance = SharedPreferencesFoundation();
+    SharedPreferencesAsyncPlatform.instance = SharedPreferencesFoundation();
+    DeprecatedSharedPreferencesFoundation.registerWith();
```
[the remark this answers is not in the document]
tarrinneal: do we want to make this possible using the normal tooling? Or is this too niche to be worth making possible?

packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_test.dart
```diff
@@ -19,147 +21,502 @@ void main() {
...
-      await preferences.setString(key, testString);
-      await preferences.setBool(key, testBool);
-      await preferences.setInt(key, testInt);
-      await preferences.setDouble(key, testDouble);
-      await preferences.setStringList(key, testList);
-      await preferences.remove(key);
-      expect(preferences.get('testKey'), isNull);
-    });
+    void runAllTests() {
+      testWidgets('reading', (WidgetTester _) async {
```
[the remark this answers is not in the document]
tarrinneal: old code, but I'll update it

packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_test.dart
```diff
@@ -19,147 +21,502 @@ void main() {
...
-      expect(preferences.getInt('int'), writeCount);
-    });
-  }
+      testWidgets('removing', (WidgetTester _) async {
+        const String key = 'testKey';
+        await preferences.setString(key, testString);
+        await preferences.setBool(key, testBool);
+        await preferences.setInt(key, testInt);
+        await preferences.setDouble(key, testDouble);
+        await preferences.setStringList(key, testList);
```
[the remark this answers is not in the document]
tarrinneal: dunno, this is old code, I'll remove them.

packages/shared_preferences/shared_preferences_foundation/darwin/Classes/SharedPreferencesPlugin.swift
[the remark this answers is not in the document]
stuartmorgan: Correct, the new Required Reason API requirement docs have a list of platforms it applies to, and that doesn't list macOS.

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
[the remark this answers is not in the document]
tarrinneal: Do you think getting one (or an allowlist of) item(s) direct from the platform, and refreshing them in the cache is a worthwhile tool? I think it is. A lot simpler (and less error prone) than manually managing the cache/platform manually. And more efficient than a full refresh, for those people who think it's a db.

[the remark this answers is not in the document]
stuartmorgan: But this says it doesn't do that?

If we want partial syncs, we could rename `reloadCache` something like `syncCache` and allow passing a filter object to it.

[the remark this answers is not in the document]
tarrinneal: I was considering adding a `getAsyncAndRefreshItemsInCache` type method alongside reload. So someone who wants to use the cache 90% of the time can still grab something critical and refresh that one (or more) item.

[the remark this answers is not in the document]
tarrinneal: method name in progress obviously

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
stuartmorgan, line 1: This file should be called shared_preferences_async.dart (it's missing an s).

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///
+  /// If no [parameters] are provided, and [SharedPreferencesAsync] has no filter,
+  /// all preferences will be removed.
+  Future<bool> clear(ClearPreferencesParameters parameters) {
+    return _platform.clear(parameters, _options);
+  }
+}
+
+/// Options necessary to create a [SharedPreferencesWithCache].
+class SharedPreferencesWithCacheOptions {
```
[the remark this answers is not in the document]
stuartmorgan: Mark `@immutable`?

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,313 @@
...
+    _cache.clear();
+    _cache.addAll(await _directAccess
+        .getAll(GetPreferencesParameters(filter: _cacheOptions.filter)));
+  }
+
+  /// Returns true if cache contains the given [key].
+  bool containsKey(String key) => _cache.containsKey(key);
+
+  /// Returns all keys in the cache.
+  Set<String> getKeys() {
```
stuartmorgan, line 217: Since this should be cheap, we could make it `Set<String> get keys`.

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,313 @@
...
+    _cache.remove(key);
+    return _directAccess.remove(key);
+  }
+
+  /// Clears cache and platform preferences that match filter options.
+  Future<bool> clear() async {
+    if (_cacheOptions.filter.allowList == null) {
+      _cache.clear();
+    } else {
+      _cache.removeWhere(
```
stuartmorgan, line 307: Since the options are immutable, can't we always `_cache.clear()`, since the only things in the cache should be things that match the filter?

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
```diff
@@ -125,12 +125,12 @@ class SharedPreferences {
...
   /// Reads a set of string values from persistent storage, throwing an
-  /// exception if it's not a string set.
+  /// exception if it's not a string list.
   List<String>? getStringList(String key) {
     List<dynamic>? list = _preferenceCache[key] as List<dynamic>?;
-    if (list != null && list is! List<String>) {
-      list = list.cast<String>().toList();
-      _preferenceCache[key] = list;
+    list = list?.cast<String>();
+    if (list is List<String>) {
```
stuartmorgan, line 132: Isn't this always true?

packages/shared_preferences/shared_preferences/test/shared_preferences_async_test.dart
```diff
@@ -0,0 +1,912 @@
...
+
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:shared_preferences/shared_preference_async.dart';
+import 'package:shared_preferences_platform_interface/in_memory_shared_preferences_async.dart';
+import 'package:shared_preferences_platform_interface/shared_preferences_async_platform_interface.dart';
+import 'package:shared_preferences_platform_interface/types.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
```
stuartmorgan, line 13: I think we don't actually need this since we aren't using method channels in this package?

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.kt
```diff
@@ -0,0 +1,306 @@
...
+  /** Adds property to data store of type bool. */
+  override fun setBool(
+      key: String,
+      value: Boolean,
+      options: SharedPreferencesPigeonOptions
+  ): Boolean {
+    return runBlocking { dataStoreSetBool(key, value) }
+  }
+
+  private suspend fun dataStoreSetBool(key: String, value: Boolean): Boolean {
```
stuartmorgan, line 77: n00b Kotlin question: we can't just inline these into the `runBlocking` lambdas?

packages/shared_preferences/shared_preferences_android/android/src/test/java/io/flutter/plugins/sharedpreferences/DeprecatedSharedPreferencesTest.java
```diff
@@ -303,27 +304,27 @@ private static class FakeSharedPreferences implements SharedPreferences {
...
     }
 
     @Override
-    public @NonNull int getInt(@NonNull String key, @NonNull int defValue) {
+    public int getInt(@NonNull String key, int defValue) {
       throw new UnsupportedOperationException("This method is not implemented for testing");
     }
 
     @Override
-    public @NonNull long getLong(@NonNull String key, @NonNull long defValue) {
```
stuartmorgan, line 326: Were all of these annotations wrong?

packages/shared_preferences/shared_preferences_foundation/lib/shared_preferences_foundation.dart
```diff
@@ -2,106 +2,351 @@
...
+
+/// Options for the Foundation specific SharedPreferences plugin.
+class SharedPreferencesFoundationOptions extends SharedPreferencesOptions {
+  /// Creates a new instance with the given options.
+  SharedPreferencesFoundationOptions({
+    this.suiteName,
+  }) {
+    if (Platform.isIOS && !(suiteName?.startsWith('group.') ?? true)) {
+      throw ArgumentError(
+          'iOS suite name must follow a specific naming convention as outlined in ');
```
stuartmorgan, line 331: In what? The suspense is killing me :)

packages/shared_preferences/shared_preferences_foundation/lib/shared_preferences_foundation.dart
```diff
@@ -2,106 +2,351 @@
...
+        rethrow;
+      }
+    }
+  }
+}
+
+/// Options for the Foundation specific SharedPreferences plugin.
+class SharedPreferencesFoundationOptions extends SharedPreferencesOptions {
+  /// Creates a new instance with the given options.
+  SharedPreferencesFoundationOptions({
```
stuartmorgan, line 326: `const`, and `@immutable` for the class?

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,313 @@
...
+
+  /// Removes an entry from the platform.
+  Future<bool> remove(String key) {
+    return _platform.clear(
+        ClearPreferencesParameters(
+            filter: PreferencesFilters(allowList: <String>{key})),
+        _options);
+  }
+
+  /// Completes with true once the user preferences for the app have been cleared.
```
stuartmorgan, line 121: When does it return false?

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  /// Returns true if the the platform contains the given [key].
+  Future<bool> containsKey(String key) async =>
+      (await getKeys(GetPreferencesParameters(
+              filter: PreferencesFilters(allowList: <String>{key}))))
+          .isNotEmpty;
+
+  /// Saves a boolean [value] to the platform.
+  Future<bool> setBool(String key, bool value) =>
```
stuartmorgan, line 88: What do all of the `bool` return values mean? It should be documented.

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,313 @@
...
+
+  /// Options that define cache behavior.
+  final SharedPreferencesWithCacheOptions _cacheOptions;
+
+  /// Async access directly to the platform.
+  ///
+  /// Methods called through [_directAccess] will NOT update the cache.
+  final SharedPreferencesAsync _directAccess;
+
+  Future<void> _init() async {
```
stuartmorgan, line 197: Why the wrapper, vs just calling `reloadCache()` in `create`?

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
```diff
@@ -0,0 +1,313 @@
...
+    List<dynamic>? list = _cache[key] as List<dynamic>?;
+    list = list?.cast<String>();
+    if (list is List<String>) {
+      _cache[key] = list.toList();
+    }
+    // Make a copy of the list so that later mutations won't propagate
+    return list?.toList() as List<String>?;
+  }
+
+  /// Saves a boolean [value] to the cache and platform.
```
stuartmorgan, line 257: Same note about documenting (or changing) the return value for all of these.

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
stuartmorgan: (Alternatively, if we don't think it's something that should happen in practice, they could be `void` and throw exceptions for any crazy errors.)

packages/shared_preferences/shared_preferences_platform_interface/lib/shared_preferences_async_platform_interface.dart
```diff
@@ -0,0 +1,99 @@
...
+/// The interface that implementations of shared_preferences_async must implement.
+abstract base class SharedPreferencesAsyncPlatform {
+  /// Constructs a SharedPreferencesAsyncPlatform.
+  SharedPreferencesAsyncPlatform();
+
+  /// The instance of [SharedPreferencesAsyncPlatform] to use.
+  static SharedPreferencesAsyncPlatform? instance;
+
+  /// Stores the String [value] associated with the [key].
+  Future<bool> setString(
```
stuartmorgan, line 18: Same thing for all the bools here

packages/shared_preferences/shared_preferences_platform_interface/lib/types.dart
```diff
@@ -2,9 +2,52 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+/// Basic options for creating SharedPreferences classes.
+///
+/// This class exists to provide extension to platform specific options as
+/// there are currently no general options that are not platform specific.
+class SharedPreferencesOptions {
```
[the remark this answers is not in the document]
stuartmorgan: `@immutable` for all of them?

stuartmorgan (commented):
I didn't do a super detailed re-review of each implementation, but I'm down to just minor comments. There's one API-level question about whether we want the setters to return `bool` or `void`, but otherwise I think this is pretty much ready (modulo ideally getting a review of the Kotlin code, and maybe me re-reviewing the Linux/Windows version in more detail, but those aren't going to affect anything big-picture).

Looks like this is in good shape to add things like README updates with discussions of the changes, migration, etc. Great work!

packages/shared_preferences/shared_preferences_foundation/lib/shared_preferences_foundation.dart
tarrinneal: That one is up to the imagination

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
tarrinneal: You're only allowed one preference with the new system

packages/shared_preferences/shared_preferences/lib/shared_preference_async.dart
tarrinneal: Style points.

packages/shared_preferences/shared_preferences/lib/shared_preferences.dart
tarrinneal: could be null

packages/shared_preferences/shared_preferences_android/android/src/test/java/io/flutter/plugins/sharedpreferences/DeprecatedSharedPreferencesTest.java
tarrinneal: android studio was yelling at me about it not being necessary for these primitives

packages/shared_preferences/shared_preferences_foundation/lib/shared_preferences_foundation.dart
tarrinneal: const is already immutable isn't it?

tarrinneal: can't be const and have throws in the constructor anyway

packages/shared_preferences/shared_preferences_android/android/src/test/java/io/flutter/plugins/sharedpreferences/DeprecatedSharedPreferencesTest.java
stuartmorgan: Ah, right. I forget when I'm not actively writing Java code that it has primitive and object versions of each numeric type.

goderbauer:
(triage): I spoke to @tarrinneal last week and he said this is still on his radar.

packages/shared_preferences/shared_preferences/lib/shared_preferences_async.dart
```diff
@@ -0,0 +1,370 @@
...
+  /// String.
+  String? getString(String key) {
+    if (!_isValidKey(key)) {
+      throw ArgumentError(
+          '$key is not included in the PreferencesFilter allowlist');
+    }
+    return get(key) as String?;
+  }
+
+  /// Reads a set of string values from the cache, throwing an
```
stuartmorgan, line 275: "a list of string values"

packages/shared_preferences/shared_preferences/lib/shared_preferences_async.dart
```diff
@@ -0,0 +1,370 @@
...
+  /// exception if it's not a string list.
+  List<String>? getStringList(String key) {
+    if (!_isValidKey(key)) {
+      throw ArgumentError(
+          '$key is not included in the PreferencesFilter allowlist');
+    }
+    List<dynamic>? list = _cache[key] as List<dynamic>?;
```
stuartmorgan, line 282: We should be able to just make this `as List<String>?` and remove the next four lines, unless I'm missing something.

packages/shared_preferences/shared_preferences/lib/shared_preferences_async.dart
```diff
@@ -0,0 +1,370 @@
...
+
+  /// Returns all keys in the cache.
+  Set<String> get keys {
+    return _cache.keys.toSet();
+  }
+
+  /// Reads a value of any type from the cache.
+  Object? get(String key) {
+    if (!_isValidKey(key)) {
+      throw ArgumentError(
```
stuartmorgan, line 229: I know it'll be repetitive, but we should add:
```dart
///
/// Throws an [ArgumentError] if [key] is not in this instance's filter.
```

to all the public methods that do this.

(While we could just put it once in the class description, doing it this way means it'll be in pop-up help in IDEs.)

packages/shared_preferences/shared_preferences_android/android/build.gradle
```diff
@@ -41,6 +44,15 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
 
+    kotlinOptions {
+        jvmTarget = '1.8'
```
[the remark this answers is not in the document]
stuartmorgan: From offline discussion: It's still 8 at the tooling level for now. Given that, we should hold off, and update to 11/17 after the tool (on stable) has done so.

packages/shared_preferences/shared_preferences_platform_interface/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 3.0.0
```
stuartmorgan, line 1: No breaking version change here.

packages/shared_preferences/shared_preferences/lib/shared_preferences_async.dart
```diff
@@ -0,0 +1,370 @@
...
+          '$key is not included in the PreferencesFilter allowlist');
+    }
+    return _cache[key];
+  }
+
+  /// Reads a value from the cache, throwing a [TypeError] if the value is not a
```
stuartmorgan, line 235: If we want to specifically say it's a `[TypeError]`, then we need to actually make that a documented part of the contract of the platform interface API. Otherwise it could easily be a `PlatformException`, and if we don't want that (which we don't per [this](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#platform-exception-handling)) we need to tell implementors that it's a requirement to transform that internally.

stuartmorgan: And we should add platform implementation package integration tests on all the platforms that we actually do that correctly, if we don't already have that in this PR (I haven't checked), by having a test for each type that sets one type, then reads another.

stuartmorgan (commented):
I'll finish reviewing tomorrow, but I wanted to get the first batch of things out now.

packages/shared_preferences/shared_preferences_foundation/test/shared_preferences_foundation_test.dart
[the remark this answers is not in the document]
tarrinneal: I'm not sure what you mean by this, can you clarify?

packages/shared_preferences/shared_preferences/lib/shared_preferences_async.dart
tarrinneal: This is only reading from the dart cache, it shouldn't be reliant on platform error handling.

packages/shared_preferences/shared_preferences_foundation/test/shared_preferences_foundation_test.dart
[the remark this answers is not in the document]
stuartmorgan: Github shows `100644 → 100755` in the header for this file, which I believe is showing the file permissions. That would indicate that this file has had the equivalent of `chmod a+x` run on it, marking it as an executable file.

packages/shared_preferences/shared_preferences/lib/shared_preferences_async.dart
stuartmorgan: Oops, let's pretend this comment is on line 54 😁

packages/shared_preferences/shared_preferences/example/lib/readme_excerpts.dart
```diff
@@ -4,6 +4,7 @@
 
 // ignore_for_file: public_member_api_docs, unused_local_variable, invalid_use_of_visible_for_testing_member
 import 'package:shared_preferences/shared_preferences.dart';
+import 'package:shared_preferences_platform_interface/types.dart';
```
stuartmorgan, line 7: This shouldn't be necessary; it's a sign that there are missing exports.

packages/shared_preferences/shared_preferences_foundation/pigeons/messages.dart
```diff
@@ -12,15 +12,64 @@ import 'package:pigeon/pigeon.dart';
...
-abstract class UserDefaultsApi {
+abstract class DeprecatedUserDefaultsApi {
   void remove(String key);
-  // TODO(stuartmorgan): Give these setters better Swift signatures (_,forKey:)
-  // once https://github.com/flutter/flutter/issues/105932 is fixed.
   void setBool(String key, bool value);
   void setDouble(String key, double value);
   void setValue(String key, Object value);
-  // TODO(stuartmorgan): Make these non-nullable once
-  // https://github.com/flutter/flutter/issues/97848 is fixed.
```
stuartmorgan, line 23: This comment is still applicable.

packages/shared_preferences/shared_preferences_linux/lib/shared_preferences_linux.dart
```diff
@@ -148,13 +117,296 @@ class SharedPreferencesLinux extends SharedPreferencesStorePlatform {
...
+    return (await getPreferences(parameters, options)).keys.toSet();
+  }
+
+  @override
+  Future<void> setString(
+    String key,
+    String value,
+    SharedPreferencesOptions options,
+  ) =>
+      _setValue(key, value, options);
```
stuartmorgan, line 175: Same note in this file about the style guide for `=>`.

packages/shared_preferences/shared_preferences_linux/lib/shared_preferences_linux.dart
```diff
@@ -148,13 +117,296 @@ class SharedPreferencesLinux extends SharedPreferencesStorePlatform {
...
+  }
+
+  @override
+  Future<List<String>?> getStringList(
+    String key,
+    SharedPreferencesOptions options,
+  ) async {
+    final Map<String, Object> data =
+        await _readAllFromFile(<String>{key}, options);
+    return data[key] as List<String>?;
```
stuartmorgan, line 256: Per other comments, this is where we'll need to do a copy.

packages/shared_preferences/shared_preferences_linux/lib/shared_preferences_linux.dart
```diff
@@ -148,13 +117,296 @@ class SharedPreferencesLinux extends SharedPreferencesStorePlatform {
...
+
+  @override
+  Future<Map<String, Object>> getPreferences(
+    GetPreferencesParameters parameters,
+    SharedPreferencesOptions options,
+  ) async {
+    return _readAllFromFile(parameters.filter.allowList, options);
+  }
+
+  Future<Map<String, Object>> _readAllFromFile(
```
stuartmorgan, line 285: We should rename this; I thought it was reading directly from the file, but it actually uses the cache if the cache has been made.

packages/shared_preferences/shared_preferences_platform_interface/lib/shared_preferences_async_platform_interface.dart
```diff
@@ -0,0 +1,99 @@
...
+  );
+
+  /// Returns all key/value pairs persisting in this store that match the given [parameters].
+  Future<Map<String, Object>> getPreferences(
+    GetPreferencesParameters parameters,
+    SharedPreferencesOptions options,
+  );
+
+  /// Returns all keys persisting in this store that match the given [parameters].
+  Future<Set<String?>> getKeys(
```
stuartmorgan, line 95: This should be `Future<Set<String>>` shouldn't it? Null keys shouldn't be allowed.

packages/shared_preferences/shared_preferences_platform_interface/lib/types.dart
```diff
@@ -2,9 +2,55 @@
...
+
+/// Filter options used to get and clear preferences on shared_preferences_async.
+class PreferencesFilters {
+  /// Creates a new instance with the given options.
+  const PreferencesFilters({
+    this.allowList,
+  });
+
+  /// A list of preference keys that will limit getting and clearing to only
+  /// items included in this list.
```
stuartmorgan, line 27: Let's comment about the difference between null and empty.

packages/shared_preferences/shared_preferences_web/lib/shared_preferences_web.dart
```diff
@@ -88,28 +90,184 @@ class SharedPreferencesPlugin extends SharedPreferencesStorePlatform {
...
+      SharedPreferencesOptions options) async {
+    return (await getPreferences(parameters, options)).keys.toSet();
+  }
+
+  @override
+  Future<void> setString(
+    String key,
+    String value,
+    SharedPreferencesOptions options,
+  ) =>
```
stuartmorgan, line 151: Same style issue.

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -154,13 +117,297 @@ class SharedPreferencesWindows extends SharedPreferencesStorePlatform {
...
+  ) async {
+    return (await getPreferences(parameters, options)).keys.toSet();
+  }
+
+  @override
+  Future<void> setString(
+    String key,
+    String value,
+    SharedPreferencesOptions options,
+  ) =>
```
stuartmorgan, line 175: `=>` style.

packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart
```diff
@@ -154,13 +117,297 @@ class SharedPreferencesWindows extends SharedPreferencesStorePlatform {
...
+  }
+
+  @override
+  Future<List<String>?> getStringList(
+    String key,
+    SharedPreferencesOptions options,
+  ) async {
+    final Map<String, Object> data =
+        await _readAllFromFile(<String>{key}, options);
+    return data[key] as List<String>?;
```
stuartmorgan, line 257: Copy here.

stuartmorgan (changes requested):

packages/shared_preferences/shared_preferences/README.md
```diff
@@ -60,32 +97,70 @@ final List<String>? items = prefs.getStringList('items');
...
 
 ### Migration and Prefixes
 
+#### Migrating from SharedPreferences to SharedPreferencesAsync/WithCache
+
+Currently, migration from the older [SharedPreferences] API to the newer 
+[SharedPreferencesAsync] or [SharedPreferencesWithCache] will need to be done manually.
+
+A simple form of this could be fetching all preferences with [SharedPreferences] and adding 
+them back using [SharedPreferencesAsync].
```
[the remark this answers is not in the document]
tarrinneal: That's fair, I just assumed people were smart enough to realize that. Shouldn't assume things...

packages/shared_preferences/shared_preferences_platform_interface/lib/types.dart
tarrinneal: an empty list would break functionality, I think we should have them act the same.

packages/shared_preferences/shared_preferences_foundation/lib/shared_preferences_foundation.dart
```diff
@@ -2,106 +2,345 @@
...
+      ))
+          .toSet();
+    } on PlatformException catch (err) {
+      if (err.code == _argumentErrorCode) {
+        throw ArgumentError(
+            'shared_preferences_foundation getString argument error${err.message ?? ''}');
+      } else {
+        rethrow;
+      }
+    }
```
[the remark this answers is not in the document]
tarrinneal: That would be cleaner, but would make the errors less clear also. I think error clarity is more important, but I'm willing to make this change if you think it's nbd.

packages/shared_preferences/shared_preferences_platform_interface/lib/types.dart
stuartmorgan: We generally don't want two ways of expressing the same thing, collection-wise, so either we should only allow one (usually we do that by making the collection non-nullable, but here I think throwing an argument error on empty might be clearer), or we should make them do different things and just allow the somewhat odd case of empty. It *might* be useful in some circumstances where the set being passed is dynamically generated?

tarrinneal: That's what I was thinking, if you want to call "nothing" you do want to get everything.

packages/shared_preferences/shared_preferences_foundation/lib/shared_preferences_foundation.dart
```diff
@@ -2,106 +2,345 @@
...
+      return (await _api.getKeys(
+        filter.allowList?.toList(),
+        _convertOptionsToPigeonOptions(options),
+      ))
+          .toSet();
+    } on PlatformException catch (err) {
+      if (err.code == _argumentErrorCode) {
+        throw ArgumentError(
+            'shared_preferences_foundation getString argument error${err.message ?? ''}');
+      } else {
```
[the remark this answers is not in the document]
tarrinneal: Does NSUserDefaults return nil for type issues? not sure if this is possible.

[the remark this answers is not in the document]
tarrinneal: or maybe I need to move to setValue for string and list<String>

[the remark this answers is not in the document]
stuartmorgan: It looks like it depends on the API; some return `nil`, and some attempt to do sketchy type coercion. E.g., bool:

>> This method automatically coerces certain ”truthy” values—such as the strings "true", "YES", and "1", and the numbers 1 and 1.0—to the Boolean value YES. The same is true for certain ”falsy” values—such as the strings "false", "NO", and "0", and the numbers 0 and 0.0—which are automatically coerced to the Boolean value NO.

You could have all the getters use `objectForKey:` and check the type of the object, returning a specific error if it's wrong.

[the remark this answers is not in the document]
tarrinneal: what a mess.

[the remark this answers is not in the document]
tarrinneal: Or I could pass the object back to dart and let dart sort it out, creates a natural TypeError in that case.

packages/shared_preferences/shared_preferences_platform_interface/lib/in_memory_shared_preferences_async.dart
```diff
@@ -0,0 +1,158 @@
...
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'shared_preferences_async_platform_interface.dart';
+import 'types.dart';
+
+/// Stores data in memory.
+///
+/// Data does not persist across application restarts. This is useful in unit-tests.
```
stuartmorgan, line 12: unit tests

packages/shared_preferences/shared_preferences_platform_interface/lib/types.dart
```diff
@@ -2,9 +2,59 @@
...
+/// This class exists to provide extension to platform specific options as
+/// there are currently no general options that are not platform specific.
+@immutable
+class SharedPreferencesOptions {
+  /// Constructor for SharedPreferencesOptions.
+  const SharedPreferencesOptions();
+}
+
+/// Filter options used to get and clear preferences on shared_preferences_async.
+class PreferencesFilters {
```
stuartmorgan, line 20: `@immutable`?

packages/shared_preferences/shared_preferences_platform_interface/lib/types.dart
```diff
@@ -2,9 +2,59 @@
...
+
+  /// A list of preference keys that will limit getting and clearing to only
+  /// items included in this list.
+  ///
+  /// An empty set will create a filter that allows no items to be set/get.
+  final Set<String>? allowList;
+}
+
+/// Parameters for use in [get] methods on shared_preferences_async.
+class GetPreferencesParameters {
```
stuartmorgan, line 34: `@immutable`?

packages/shared_preferences/shared_preferences_platform_interface/lib/types.dart
```diff
@@ -2,9 +2,59 @@
...
+  /// Creates a new instance with the given options.
+  const GetPreferencesParameters({required this.filter});
+
+  /// Filter to limit which preferences are returned.
+  final PreferencesFilters filter;
+}
+
+/// Parameters for use in [clear] methods on shared_preferences_async.
+class ClearPreferencesParameters {
```
stuartmorgan, line 43: `@immutable`?

stuartmorgan (approved):
It looks like there was a force-push since my last review, so GitHub won't show me the incremental diffs. I tried to re-review where I think thought there were likely to be changes, as I don't think re-reviewing every file that's changed would be productive. With that caveat, it looks good with some minor comments!

tarrinneal:
> It looks like there was a force-push since my last review, so GitHub won't show me the incremental diffs. 

There hasn't, probably just a github bug. I think the things that needed review were pretty clear though, since you'd already covered everything

stuartmorgan:
I was able to coax it into showing me some of the recent diffs if I selected them by hand.

tarrinneal:
@stuartmorgan last chance if you want to do another pass on this now that it's just the front end.

packages/shared_preferences/shared_preferences/README.md
```diff
@@ -17,9 +17,48 @@ Supported data types are `int`, `double`, `bool`, `String` and `List<String>`.
 
 ## Usage
 
-### Examples
+## SharedPreferences vs SharedPreferencesAsync vs SharedPreferencesWithCache
+
+As of the 2.3.0 update there are three available APIs that can be used in this package.
```
stuartmorgan, line 22: Optional nit: I think this would be a bit more evergreen as "Starting with version 2.3.0 there are [...]"; referring to the "update" will seem a little weird when someone is reading this for the first time six months from now.

packages/shared_preferences/shared_preferences/README.md
```diff
@@ -17,9 +17,48 @@ Supported data types are `int`, `double`, `bool`, `String` and `List<String>`.
...
+[SharedPreferences] is a legacy API that will be deprecated in the future. We highly encourage
+any new users of the plugin to use the newer [SharedPreferencesAsync] or [SharedPreferencesWithCache] 
+APIs instead.
+
+Consider migrating existing code to one of the new APIs. See  [below](#migrating-from-sharedpreferences-to-sharedpreferencesasyncwithcache)
```
stuartmorgan, line 27: Nit: there's an extra space after "See".

packages/shared_preferences/shared_preferences/README.md
```diff
@@ -17,9 +17,48 @@ Supported data types are `int`, `double`, `bool`, `String` and `List<String>`.
...
+[SharedPreferences] is a legacy API that will be deprecated in the future. We highly encourage
+any new users of the plugin to use the newer [SharedPreferencesAsync] or [SharedPreferencesWithCache] 
+APIs instead.
+
+Consider migrating existing code to one of the new APIs. See  [below](#migrating-from-sharedpreferences-to-sharedpreferencesasyncwithcache) 
+for more information. 
+
+### Cache and async or sync getters
+
+[SharedPreferences] and [SharedPreferencesWithCache] both utilize a local cache to store preferences.
```
stuartmorgan, line 32: Nit: utilize->use

packages/shared_preferences/shared_preferences/lib/src/shared_preferences_async.dart
```diff
@@ -0,0 +1,404 @@
...
+    this.allowList,
+  });
+
+  /// Information about what data will be fetched during `get` and `init`
+  /// methods, what data can be `set`, as well as what data will be removed by `clear`.
+  ///
+  /// A `null` allowList will prevent filtering, allowing all items to be cached.
+  /// An empty allowList will prevent all caching as well as getting and setting.
+  ///
+  /// It is highly recommended to provide an allowList to prevent getting and
```
stuartmorgan, line 160: "Setting an `allowList` is strongly recommended, to prevent [...]"

packages/shared_preferences/shared_preferences/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Flutter plugin for reading and writing simple key-value pairs.
   Wraps NSUserDefaults on iOS and SharedPreferences on Android.
 repository: https://github.com/flutter/packages/tree/main/packages/shared_preferences/shared_preferences
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+shared_preferences%22
-version: 2.2.3
+version: 2.3.0
 
 environment:
   sdk: ^3.2.0
```
stuartmorgan, line 9: Looks like this will need a bump to 3.19 due to one of the dependencies.

packages/shared_preferences/shared_preferences_platform_interface/README.md
```diff
@@ -1,6 +1,6 @@
 # shared_preferences_platform_interface
 
-A common platform interface for the [`shared_preferences`][1] plugin.
+A common platform interface for the [`shared_preferences`][1] plugins.
```
stuartmorgan, line 3: Revert this; it's one plugin.

stuartmorgan (approved):

tarrinneal:
Looks like tests are passing (besides mac flake), going to mark as auto merge!

xVemu:
`setMockInitialValues` is missing for SharedPreferencesWithCache

tarrinneal:
> `setMockInitialValues` is missing for SharedPreferencesWithCache

You can set the cache manually when instantiating `SharedPreferencesWithCache` so there is no need for that method.

PR #5225, opened 2023-10-25
[google_sign_in] Enable FedCM for web. Use token expiration.

ditman:
* Enables [**FedCM API**](https://developer.mozilla.org/en-US/docs/Web/API/FedCM_API) on compatible browsers.
  * The GIS JS SDK falls-back to the JS implementation on browsers that don't support the new standard. See [migration instructions](https://developers.google.com/identity/gsi/web/guides/fedcm-migration)).
* Uses the supplied token expiration information to **more accurately compute `isSignedIn()` and `canAccessScopes(scopes)`**.
  * This does not handle the case where users sign in/out in another tab or from outside the web app, that's still something that needs to be checked server-side.
* **Deprecates the `signIn()` method on the web.**
  * Users should migrate to a combination of `renderButton()` and `silentSignIn()`, as described [here](https://pub.dev/packages/google_sign_in_web#migrating-to-v011-and-v012-google-identity-services).

### Issues

* FedCM:
  * Fixes https://github.com/flutter/flutter/issues/133703 (once rebuilt/redeployed)
  * Fixes `b/301259123`
* Token expiration:
  * Unblocks `b/245740319`

### Testing

* Added a few unit tests
* Manually verified token expiration: https://dit-gis-test.web.app

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
+++ b/packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
@@ -66,6 +66,8 @@ class GisSdkClient {
 
     _tokenResponses.stream.listen((TokenResponse response) {
       _lastTokenResponse = response;
+      _lastTokenResponseExpiration =
+          DateTime.now().add(Duration(seconds: response.expires_in));
     }, onError: (Object error) {
       _logIfEnabled('Error on TokenResponse:', <Object>[error.toString()]);
       _lastTokenResponse = null;
@@ -280,6 +282,8 @@ class GisSdkClient {
   // This function returns the currently signed-in [GoogleSignInUserData].
   //
   // It'll do a request to the People API (if needed).
+  //
+  // @Deprecated
   Future<GoogleSignInUserData?> _computeUserDataForLastToken() async {
     // If the user hasn't authenticated, request their basic profile info
     // from the People API.
@@ -317,9 +321,17 @@ class GisSdkClient {
     await signOut();
   }
 
-  /// Returns true if the client has recognized this user before.
+  /// Returns true if the client has recognized this user before, and the last-seen
+  /// credential is not expired.
   Future<bool> isSignedIn() async {
-    return _lastCredentialResponse != null || _requestedUserData != null;
+    bool isSignedIn = false;
+    if (_lastCredentialResponse != null) {
+      final DateTime? expiration = utils
+          .getCredentialResponseExpirationTimestamp(_lastCredentialResponse);
+      isSignedIn = expiration?.isAfter(DateTime.now()) ?? false;
+    }
+
+    return isSignedIn || _requestedUserData != null;
   }
 
   /// Clears all the cached results from authentication and authorization.
@@ -353,12 +365,15 @@ class GisSdkClient {
   /// Checks if the passed-in `accessToken` can access all `scopes`.
   ///
   /// This validates that the `accessToken` is the same as the last seen
-  /// token response, and uses that response to check if permissions are
-  /// still granted.
+  /// token response, that the token is not expired, then uses that response to
+  /// check if permissions are still granted.
   Future<bool> canAccessScopes(List<String> scopes, String? accessToken) async {
     if (accessToken != null && _lastTokenResponse != null) {
       if (accessToken == _lastTokenResponse!.access_token) {
-        return oauth2.hasGrantedAllScopes(_lastTokenResponse!, scopes);
+        final bool isTokenValid =
+            _lastTokenResponseExpiration?.isAfter(DateTime.now()) ?? false;
+        return isTokenValid &&
+            oauth2.hasGrantedAllScopes(_lastTokenResponse!, scopes);
       }
     }
     return false;
@@ -383,6 +398,8 @@ class GisSdkClient {
   // The last-seen credential and token responses
   CredentialResponse? _lastCredentialResponse;
   TokenResponse? _lastTokenResponse;
+  // Expiration timestamp for the lastTokenResponse, which only has an `expires_in` field.
+  DateTime? _lastTokenResponseExpiration;
 
   /// The StreamController onto which the GIS Client propagates user authentication events.
   ///
@@ -394,5 +411,7 @@ class GisSdkClient {
   // (if needed)
   //
   // (This is a synthetic _lastCredentialResponse)
+  //
+  // @Deprecated
   GoogleSignInUserData? _requestedUserData;
 }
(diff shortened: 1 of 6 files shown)


ditman:
I'm awaiting confirmation from the GIS team to see if they want Flutter users to opt-in to this new API (would require a platform interface + core plugin change), or is it OK that we opt them in.

ditman:
Internally confirmed that it's OK that we flip this switch on for our users. The more, the merrier.

packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
```diff
@@ -265,6 +282,8 @@ class GisSdkClient {
   // This function returns the currently signed-in [GoogleSignInUserData].
   //
   // It'll do a request to the People API (if needed).
+  //
+  // @Deprecated
```
stuartmorgan, line 286: Why is a private method deprecated?

packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
```diff
@@ -302,9 +321,17 @@ class GisSdkClient {
...
-  /// Returns true if the client has recognized this user before.
+  /// Returns true if the client has recognized this user before, and the last-seen
+  /// credential is not expired.
   Future<bool> isSignedIn() async {
-    return _lastCredentialResponse != null || _requestedUserData != null;
+    bool isSignedIn = false;
+    if (_lastCredentialResponse != null) {
+      final DateTime? expiration = utils
+          .getCredentialResponseExpirationTimestamp(_lastCredentialResponse);
+      isSignedIn = expiration?.isAfter(DateTime.now()) ?? false;
```
stuartmorgan, line 331: So a missing `expiration` field would indicate that it's already expired? Naively (not knowing the API surface) I would expect that if expiration is optional, then not having it should yield a `true` result here, so maybe a comment would be helpful?

packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
```diff
@@ -379,5 +411,7 @@ class GisSdkClient {
   // (if needed)
   //
   // (This is a synthetic _lastCredentialResponse)
+  //
+  // @Deprecated
```
stuartmorgan, line 415: Same question here about what deprecation of a private field means.

packages/google_sign_in/google_sign_in_web/lib/src/utils.dart
```diff
@@ -52,19 +52,22 @@ Map<String, Object?>? decodeJwtPayload(String? payload) {
   return null;
 }
 
+/// Returns the payload of a [CredentialResponse].
+Map<String, Object?>? getResponsePayload(CredentialResponse? response) {
+  if (response == null || response.credential == null) {
```
stuartmorgan, line 57: Optional nit: `if (response?.credential == null)`

packages/google_sign_in/google_sign_in_web/lib/src/utils.dart
```diff
@@ -74,10 +77,24 @@ GoogleSignInUserData? gisResponsesToUserData(
     id: payload['sub']! as String,
     displayName: payload['name'] as String?,
     photoUrl: payload['picture'] as String?,
-    idToken: credentialResponse.credential,
+    idToken: credentialResponse!.credential,
```
stuartmorgan, line 80: This is conceptually removed enough from the code that ensures that the `!` is safe (it's in another method) that I worry about someone changing that code later and this becoming a crash. Couldn't it just be `credentialResponse?.credential` so it's safe, since it's a nullable field?

packages/google_sign_in/google_sign_in_web/lib/src/utils.dart
```diff
@@ -74,10 +77,24 @@ GoogleSignInUserData? gisResponsesToUserData(
...
 }
 
+/// Returns the expiration timestamp ('exp') of a [CredentialResponse].
+///
+/// May return `null` if the `credentialResponse` is null, its `credential`
+/// cannot be decoded, or the `exp` field is not set on the JWT payload.
+DateTime? getCredentialResponseExpirationTimestamp(
+    CredentialResponse? credentialResponse) {
+  final Map<String, Object?>? payload = getResponsePayload(credentialResponse);
+  if (payload == null || payload['exp'] == null) {
```
stuartmorgan, line 91: How about:

```dart
final int? expiration = playload == null ? null : payload['exp'] as int?;
if (expiration == null) {
  return null;
}
```

so we don't need the force-unwrap?

stuartmorgan (approved):
Small comments, but otherwise LGTM

ditman:
Thanks for the review! I'll address the comments shortly!

packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
ditman: I added these because I don't want to forget about removing these when actually removing the `signIn` method, but this should probably be an issue + proper TODO.

I'll create the issue right away and link it properly.

ditman: Linked these breadcrumbs to this issue:

* https://github.com/flutter/flutter/issues/137727

packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
ditman: I'm going to document this a little bit further, but in this case, if the `CredentialResponse` payload cannot be decoded, is null, or doesn't contain an "exp" claim, we assume that the token is bad, and thus the user is *not* authenticated.

In a valid Google ID Token, the "exp" claim should **always** be provided:

https://developers.google.com/identity/openid-connect/openid-connect#an-id-tokens-payload

packages/google_sign_in/google_sign_in_web/lib/src/utils.dart
ditman: Yep, good call!

packages/google_sign_in/google_sign_in_web/lib/src/utils.dart
ditman: Yes, in this case `payload != null` implies `credentialResponse != null` but as you said, the connection is pretty vague.

I think what I really want here is an assertion that (credentialResponse?.credential != null) before returning the GoogleSignInUserData, that way I'm expressing in code that the credential must NOT be null before returning a GoogleSignInUserData.

packages/google_sign_in/google_sign_in_web/lib/src/utils.dart
ditman: Good call, I also did a ternary for the `return`. Much shorter!

ditman:
I'll land this now, and talk to the Cocoon ppl.

PR #5226, opened 2023-10-25
Run one build per platform in recipes cq.

godofredoc:
Flakiness in packages build block recipe changes very frequently and slows them down. Given that the same recipe is used in all the builds we only need to run one build per platform in the cq.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -112,7 +112,6 @@ targets:
     recipe: packages/packages
     timeout: 30
     properties:
-      add_recipes_cq: "true"
       target_file: repo_tools_tests.yaml
       channel: master
       version_file: flutter_master.version
@@ -135,7 +134,6 @@ targets:
     recipe: packages/packages
     timeout: 60
     properties:
-      add_recipes_cq: "true"
       target_file: dart_unit_tests.yaml
       channel: master
       version_file: flutter_master.version
@@ -145,7 +143,6 @@ targets:
     recipe: packages/packages
     timeout: 60
     properties:
-      add_recipes_cq: "true"
       target_file: dart_unit_tests.yaml
       channel: master
       version_file: flutter_master.version
@@ -173,7 +170,6 @@ targets:
     recipe: packages/packages
     timeout: 60
     properties:
-      add_recipes_cq: "true"
       target_file: web_dart_unit_tests.yaml
       channel: master
       version_file: flutter_master.version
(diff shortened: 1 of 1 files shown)


stuartmorgan:
> Given that the same recipe is used in all the builds we only need to run one build per platform in the cq.

How does that follow? Not every build uses every feature of the recipes. For instance, we need to add secrets support for the maps API keys, and none of the builds chosen here would make use of them. None of these use emulator support either, which IIRC has a recipe component.

Shouldn't we have one of each kind of build, and just remove shards beyond shard 1?

stuartmorgan:
@godofredoc Ping on the question above.

stuartmorgan:
@godofredoc Should this still be open?

godofredoc:
> > Given that the same recipe is used in all the builds we only need to run one build per platform in the cq.
> 
> How does that follow? Not every build uses every feature of the recipes. For instance, we need to add secrets support for the maps API keys, and none of the builds chosen here would make use of them. None of these use emulator support either, which IIRC has a recipe component.
> 
> Shouldn't we have one of each kind of build, and just remove shards beyond shard 1?

Recipes CQ is intended to validate changes on recipes and/or recipe modules. Just enough builds to test the packages recipe paths is desirable. Also inspecting the packages recipe seems really flat not sure why we need tens of builds to validate it.

The main reason I created this PR is because some of these builds are very slow and fail frequently blocking recipes CLs for several hours. If we would like to continue running all these builds can we at least remove the slow and flaky ones?

stuartmorgan:
> Recipes CQ is intended to validate changes on recipes and/or recipe modules. Just enough builds to test the packages recipe paths is desirable.

Then we need at least one build using everything the recipe does. E.g., emulators, as I mentioned above. Just one per platform isn't sufficient to test all the recipe functionality.

> The main reason I created this PR is because some of these builds are very slow and fail frequently blocking recipes CLs for several hours.

I'm fine with trimming the list down, but it needs to do so in a way that doesn't remove coverage of critical functionality.

> If we would like to continue running all these builds can we at least remove the slow and flaky ones?

Emulators are only used for integration tests, so there is no way to ensure that emulator support isn't broken without one such test being included.

stuartmorgan:
@godofredoc How do you want to proceed with this? Do you need me to identify a more minimal set without losing key coverage?

stuartmorgan:
From triage: I'll update this to a smaller (but less minimal) set.

PR #5240, opened 2023-10-26
[google_sign_in_ios] Upgrade GoogleSignIn iOS SDK to 7.0

vashworth:
Fixes https://github.com/flutter/flutter/issues/137140.

See https://developers.google.com/identity/sign-in/ios/quick-migration-guide and https://developers.google.com/identity/sign-in/ios/release for info about changes made to the SDK.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.h
```diff
@@ -6,5 +6,13 @@
...
 #import "messages.g.h"
 
+#import <GoogleSignIn/GoogleSignIn.h>
+
 @interface FLTGoogleSignInPlugin : NSObject <FlutterPlugin, FSIGoogleSignInApi>
+
+// Configuration wrapping Google Cloud Console, Google Apps, OpenID,
+// and other initialization metadata.
+// @property(strong) GIDConfiguration *configuration;
+@property(strong, nonatomic) GIDConfiguration *configuration;
```
vashworth, line 16: I moved this here so that I could access it in tests to validate it was set correctly. Let me know if there's a better way.

packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.m
```diff
@@ -116,18 +112,14 @@ - (void)initializeSignInWithParameters:(nonnull FSIInitParams *)params
   if (configuration != nil) {
     self.requestedScopes = [NSSet setWithArray:params.scopes];
     self.configuration = configuration;
-  } else {
-    *error = [FlutterError errorWithCode:@"missing-config"
-                                 message:@"GoogleService-Info.plist file not found and clientId "
-                                         @"was not provided programmatically."
-                                 details:nil];
```
vashworth, line 123: The Google Sign In SDK now gets the configuration settings within the SDK itself from the Info.plist and that's now the preferred way to do it.

packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.m
```diff
@@ -116,18 +112,14 @@ - (void)initializeSignInWithParameters:(nonnull FSIInitParams *)params
...
   }
 }
 
 - (void)signInSilentlyWithCompletion:(nonnull void (^)(FSIUserData *_Nullable,
                                                        FlutterError *_Nullable))completion {
-  [self.signIn restorePreviousSignInWithCallback:^(GIDGoogleUser *user, NSError *error) {
-    [self didSignInForUser:user withCompletion:completion error:error];
+  [self.signIn restorePreviousSignInWithCompletion:^(GIDGoogleUser *_Nullable user,
+                                                     NSError *_Nullable error) {
+    [self didSignInForUser:user serverAuthCode:nil withCompletion:completion error:error];
```
vashworth, line 122: You may notice here I set `serverAuthCode` to `nil`. This is something I was unsure about. In previous version, the `serverAuthCode` was a property of the user. However, in the new version, it's now a property of `GIDSignInResult`, which you get when you sign in. I did confirm that in the old version on restore, the `serverAuthCode` was included.

However, I don't think this is a dealbreaker because I think the new way where it doesn't include the `serverAuthCode` is more inline with how OAuth works. Usually that code is a temporary one-time use code and then after that tokens are use to validate (see diagram in https://developers.google.com/identity/sign-in/web/server-side-flow).

packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.m
```diff
@@ -190,31 +199,38 @@ - (void)requestScopes:(nonnull NSArray<NSString *> *)scopes
...
+                        BOOL granted = NO;
+                        FlutterError *error = nil;
+
+                        if ([addedScopeError.domain isEqualToString:kGIDSignInErrorDomain] &&
+                            addedScopeError.code == kGIDSignInErrorCodeMismatchWithCurrentUser) {
+                          error =
+                              [FlutterError errorWithCode:@"request_scopes"
+                                                  message:@"There is an operation on a previous "
+                                                          @"user. Try signing in again."
+                                                  details:nil];
```
vashworth, line 221: FYI, they removed `kGIDSignInErrorCodeNoCurrentUser` as an error, I'm assuming because the method is now part of the `GIDGoogleUser` class. They also added a new error `kGIDSignInErrorCodeMismatchWithCurrentUser`, which I believe happens when GIDSignIn.currentUser doesn't match the user that makes this request. This shouldn't ever happen since we get the user from `self.signIn.currentUser`, but I added a new error just in case.

packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.h
```diff
@@ -6,5 +6,12 @@
 
 #import "messages.g.h"
 
+#import <GoogleSignIn/GoogleSignIn.h>
+
 @interface FLTGoogleSignInPlugin : NSObject <FlutterPlugin, FSIGoogleSignInApi>
+
+// Configuration wrapping Google Cloud Console, Google Apps, OpenID,
+// and other initialization metadata.
+@property(strong) GIDConfiguration *configuration;
```
vashworth, line 15: I moved this here so that I could access it in tests to validate it was set correctly. Let me know if there's a better way.

stuartmorgan:
> The addScopes: method has been moved to GIDGoogleUser. Instead of requesting additional authorization scopes from GIDSignIn, you should now [request them from GIDGoogleUser after authentication](https://developers.google.com/identity/sign-in/ios/api-access#2_request_additional_scopes) has completed.

Does this mean we are now stuck with the two-stage described [here](https://github.com/flutter/flutter/issues/86436#issuecomment-883774462)?

vashworth:
> > The addScopes: method has been moved to GIDGoogleUser. Instead of requesting additional authorization scopes from GIDSignIn, you should now [request them from GIDGoogleUser after authentication](https://developers.google.com/identity/sign-in/ios/api-access#2_request_additional_scopes) has completed.
> 
> Does this mean we are now stuck with the two-stage UI flow described [here](https://github.com/flutter/flutter/issues/86436#issuecomment-883774462)?

I don't believe so, you can also request the scopes on initial sign in
https://github.com/flutter/packages/blob/815fd8f4a7e8ceb6bafd82c2dc531fa90b75d0ab/packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.m#L147-L149

I think that was fixed in 6.2 (https://github.com/flutter/plugins/pull/5708)

stuartmorgan:
> I don't believe so, you can also request the scopes on initial sign in

😌 

> I think that was fixed in 6.2 ([flutter/plugins#5708](https://github.com/flutter/plugins/pull/5708))

Right, I was afraid from that migration description that they had un-reversed that decision in 7 and were re-committing to separating authn and authz.

packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.m
```diff
@@ -190,31 +199,38 @@ - (void)requestScopes:(nonnull NSArray<NSString *> *)scopes
...
-                        }];
+                      completion:^(GIDSignInResult *_Nullable signInResult,
+                                   NSError *_Nullable addedScopeError) {
+                        BOOL granted = NO;
+                        FlutterError *error = nil;
+
+                        if ([addedScopeError.domain isEqualToString:kGIDSignInErrorDomain] &&
+                            addedScopeError.code == kGIDSignInErrorCodeMismatchWithCurrentUser) {
+                          error =
+                              [FlutterError errorWithCode:@"request_scopes"
```
stuartmorgan, line 218: We may want to give this a new code; currently string matching on codes is kind of the best option clients have for `PlatformException` handling (until we [fix it long term](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#platform-exception-handling)), so when we have a distinct case it's useful to give it its own code.

stuartmorgan (changes requested):

packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.m
```diff
@@ -281,7 +298,7 @@ - (void)didSignInForUser:(GIDGoogleUser *)user
                                           email:user.profile.email
                                          userId:user.userID
                                        photoUrl:[photoUrl absoluteString]
-                                 serverAuthCode:user.serverAuthCode],
+                                 serverAuthCode:serverAuthCode],
```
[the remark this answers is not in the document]
vashworth: `idToken` was the only one from what I could tell.

packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.m
vashworth: FYI I moved `GoogleService-Info.plist` from the Runner target to the RunnerTests target, so we can still use it for the example tests.

packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.h
```diff
@@ -6,5 +6,8 @@
 
 #import "messages.g.h"
 
+#import <GoogleSignIn/GoogleSignIn.h>
```
stuartmorgan, line 9: Do we need this?

packages/google_sign_in/google_sign_in_ios/ios/Classes/FLTGoogleSignInPlugin.m
```diff
@@ -139,19 +121,36 @@ - (nullable NSNumber *)isSignedInWithError:
...
+                                           GIDGoogleUser *user;
+                                           NSString *serverAuthCode;
+                                           if (signInResult) {
+                                             user = signInResult.user;
+                                             serverAuthCode = signInResult.serverAuthCode;
+                                           }
+
+                                           [self didSignInForUser:user
+                                                   serverAuthCode:serverAuthCode
+                                                   withCompletion:completion
```
stuartmorgan, line 151: nit: the `with` should go on the first additional argument, so `didSignInForUser:withServerAuthCode:completion:error:`.

stuartmorgan (approved):
LGTM with nits. Thanks for doing this update!

PR #5243, opened 2023-10-27
Bump play-service-maps dependency from version 18.1.0 to 18.2.0

johnmccutchan:
Google Maps team has requested we update our client library version to the latest.


packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.5.3
+
+* Update `com.google.android.gms:play-services-maps` to 18.2.0.
```
gmackall, line 3: nit: this should be "Updates...", per https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style

johnmccutchan: Done.

gmackall (approved):

gmackall:
Looks like there are some actual test failures here, the types used in the constructors of `Polyline` and `Polygon` seem to have changed. The tests seem to pass after just changing the type of the mock used for these tests to fit

https://github.com/flutter/packages/blob/21a585f176a5e20028977ddfd49dafed0e87eecb/packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/PolylineControllerTest.java#L15
https://github.com/flutter/packages/blob/21a585f176a5e20028977ddfd49dafed0e87eecb/packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/PolygonControllerTest.java

PR #5245, opened 2023-10-27
[camera_avfoundation] Set highest available resolution for ResolutionPreset.Max

sergeidesenko:
The current implementation of the camera plugin for iOS iterates though all other resolution presets when set to FLTResolutionPresetMax. This results in a resolution of 3840x2160 at most, while native camera is able to produce 4032x3024

This change should partially address these issues at least.
- https://github.com/flutter/flutter/issues/58163
- https://github.com/flutter/flutter/issues/78247
- https://github.com/flutter/flutter/issues/45906

P.S. I'm not really sure about tests - it seems that resolution presets are not covered by any. Any feedback is appreciated!

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -348,6 +348,29 @@ - (NSString *)getTemporaryFilePathWithExtension:(NSString *)extension
...
   switch (resolutionPreset) {
     case FLTResolutionPresetMax:
+      {
+          AVCaptureDeviceFormat *bestFormat = [self getHighestResolutionFormatFor:_captureDevice];
+          if ( bestFormat ) {
+              _videoCaptureSession.sessionPreset = AVCaptureSessionPresetInputPriority;
+              if ( [_captureDevice lockForConfiguration:NULL] == YES ) {
+
+                  // set best device format and finish device configuration
+                  _captureDevice.activeFormat = bestFormat;
```
hellohuanlin, line 358: is this only required in best format but not other resolutions?

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -403,6 +426,23 @@ - (BOOL)setCaptureSessionPreset:(FLTResolutionPreset)resolutionPreset withError:
...
+        CMVideoDimensions res = CMVideoFormatDescriptionGetDimensions(format.formatDescription);
+        NSUInteger height = res.height;
+        NSUInteger width = res.width;
+        NSUInteger pixelCount = height * width;
+        if ( pixelCount > maxPixelCount ) {
+          maxPixelCount = pixelCount;
+          bestFormat = format;
+        }
+    }
+    return bestFormat;
```
hellohuanlin, line 443: what if `[_captureDevice formats]` is empty? do we want to handle nil case?

hellohuanlin (changes requested):
Can you add some unit test?

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
sergeidesenko: We do handle it actually, this method returns nil if there is no bestFormat and switch case from setCaptureSessionPreset doesn't break and falls through to other cases

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
sergeidesenko: In all other cases device is configured with sessionPreset through capture session, like this:
 _videoCaptureSession.sessionPreset = AVCaptureSessionPreset3840x2160;
 
 The max resolution case is different because there is no predetermined option

sergeidesenko:
> Can you add some unit test?
@hellohuanlin 
As I said in original pr message, I'm not really sure about tests - it seems that resolution presets are not covered by any. Could you tell me a little bit more about what you are expecting? Happy to add some tests, just feel a little bit lost

hellohuanlin:
@sergeidesenko For testing, at a high level, we will want verify if each of different configs correctly updates the capture session. If you take a look at the FLTCam constructor, it takes both audio and video sessions, so we can inject them in the test. Let me know if it makes sense.

stuartmorgan:
@sergeidesenko Are you planning on updating this PR with tests as discussed above?

sergeidesenko:
@stuartmorgan Hi! Didn't have enough bandwidth at the end of the year to look into this. Yes, I will add tests before the end of January, does this work?

hellohuanlin:
I'm converting this to draft state. Feel free to add tests and re-request for review.

sergeidesenko:
@hellohuanlin Hi! Added tests that check whether setting resolutionPreset actually updates captureSession. Would be happy to hear what you think!

jmagman:
@hellohuanlin can you take another look?

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -403,6 +424,23 @@ - (BOOL)setCaptureSessionPreset:(FLTResolutionPreset)resolutionPreset withError:
   return YES;
 }
 
+/// Finds the highest available resolution in terms of pixel count for the given device
+- (AVCaptureDeviceFormat *)getHighestResolutionFormatFor:(AVCaptureDevice *)captureDevice {
```
hellohuanlin, line 428: nit: `getHighestResolutionFormatForCaptureDevice`

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraSessionPresetsTest.m
```diff
@@ -0,0 +1,53 @@
...
+                                       [value isEqualToString:expectedMaxPreset];
+                              }]]);
+
+  FLTCreateCamWithVideoCaptureSession(videoSessionMock, @"max");
+
+  OCMVerifyAll(videoSessionMock);
+}
+
+- (void)
+    testResolutionPresetWithCanSetSessionPresetUltraHigh_mustUpdateCaptureSessionPreset {  // TODO
```
hellohuanlin, line 38: is this still TODO?

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraSessionPresetsTest.m
```diff
@@ -0,0 +1,53 @@
...
+  NSString *expectedUltraHighPreset = AVCaptureSessionPreset3840x2160;
+  NSString *expectedMaxPreset = AVCaptureSessionPresetInputPriority;
+
+  id videoSessionMock = OCMClassMock([AVCaptureSession class]);
+  OCMStub([videoSessionMock addInputWithNoConnections:[OCMArg any]]);  // no-op
+  OCMStub([videoSessionMock canSetSessionPreset:[OCMArg any]]).andReturn(YES);
+  OCMExpect([videoSessionMock setSessionPreset:[OCMArg checkWithBlock:^BOOL(id value) {
+                                // Return YES if the value is one of the expected presets
+                                return [value isEqualToString:expectedUltraHighPreset] ||
+                                       [value isEqualToString:expectedMaxPreset];
```
hellohuanlin, line 29: can you explicitly mock the values and verify each of them?

packages/camera/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.5+6
+
+* Fixes bug where max resolution preset does not produce highest available resolution on iOS.
```
hellohuanlin, line 3: this doesn't need a bump

sergeidesenko: fixed

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraSessionPresetsTest.m
sergeidesenko: fixed

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
sergeidesenko: fixed

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraSessionPresetsTest.m
sergeidesenko: Thanks for the review! 
I'm not sure I understood your message here. If you are suggesting two different tests (one for `expectedUltraHighPreset` and one for `expectedMaxPreset`), then that's not really the expected behaviour. 
The test is for the case where camera receives `FLTResolutionPresetMax`, and in that case `sessionPreset` can be one of these values and both are acceptable.
If I missed your point - could you please clarify what you meant?

hellohuanlin: >  in that case sessionPreset can be one of these two values and both are acceptable.

Can you explain when will it be one value and when the other value? It's unclear which one will be true when running the test.

sergeidesenko: If you take a look in the file [FLTCam.m](https://github.com/flutter/packages/pull/5245/files#diff-1af831ad95b26582a2afa02afb68f6de8596c2d118f39d246523cf0a25f95aa4) at line 372, there are two options on how to handle FLTResolutionPresetMax. Either we can find `bestFormat`, in which case `sessionPreset` is set to `AVCaptureSessionPresetInputPriority`, or the `bestFormat` is nil, in which case we try to set `sessionPreset` to `AVCaptureSessionPreset3840x2160`. 

Since method `canSetSessionPreset` of our mock object always returns true, we can expect that if we try to set `AVCaptureSessionPreset3840x2160`, we will always succeed. That's how we get only to possible outcomes. 


> It's unclear which one will be true when running the test.
About that - we could mock further and provide our videoSessionMock with availableFormats, split the test in two different tests - one that covers case with the best format and one that covers case with no formats and `AVCaptureSessionPreset3840x2160`. But since both of those cases are an acceptable response to setting FLTResolutionPresetMax, I didn't see the need to go this far

stuartmorgan: > Since method `canSetSessionPreset` of our mock object always returns true, we can expect that if we try to set `AVCaptureSessionPreset3840x2160`, we will always succeed. That's how we get only to possible outcomes.

If we have two codepaths, we should be testing both of them.

> But since both of those cases are an acceptable response to setting FLTResolutionPresetMax (as I explained above), I didn't see the need to go this far

Why would we want no test coverage that one of the possible outcomes actually works?

sergeidesenko: > If we have two codepaths, we should be testing both of them.

Thank you! I'll add another test


> Why would we want no test coverage that one of the possible outcomes actually works?

I'm not sure, since this whole behavior of resolution presets wasn't covered by any test at all :)

stuartmorgan: https://github.com/flutter/flutter/wiki/Plugin-Tests#faq-do-i-need-to-add-tests-even-if-the-part-of-the-code-im-changing-isnt-already-tested

stuartmorgan: https://github.com/flutter/flutter/wiki/Plugin-Tests#faq-do-i-need-to-add-tests-even-if-the-part-of-the-code-im-changing-isnt-already-tested

sergeidesenko:
@stuartmorgan separated tests to cover both codepaths. It got a bit messy - since `activeDevice` property  is set and `setCaptureSessionPreset` is called during initialization of `FLTCam`, I couldn't think of a way to reach a required conditions for a test without swizzling. If you see a better approach - please let me know!

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -422,6 +444,24 @@ - (BOOL)setCaptureSessionPreset:(FLTResolutionPreset)resolutionPreset withError:
   return YES;
 }
 
+/// Finds the highest available resolution in terms of pixel count for the given device
+- (AVCaptureDeviceFormat *)getHighestResolutionFormatForCaptureDevice:
```
stuartmorgan, line 448: Drive-by nit: Obj-C methods should not start with "get".

stuartmorgan:
> It got a bit messy - since `activeDevice` property is set and `setCaptureSessionPreset` is called during initialization of `FLTCam`, I couldn't think of a way to reach a required conditions for a test without swizzling. If you see a better approach - please let me know!

If you need a mock `AVCaptureDevice` during initialization, then the device should be provided via dependency injection to a test-visible initializer, and the public initializer should call that initializer with a real device.

sergeidesenko:
> If you need a mock `AVCaptureDevice` during initialization, then the device should be provided via dependency injection to a test-visible initializer, and the public initializer should call that initializer with a real device.

That's right, but then I need to mock `formats` property of this device, so I need a mocked object of  `AVCaptureDeviceFormat`. And this object, when passed to `CMVideoFormatDescriptionGetDimensions`, should return valid `CMVideoDimensions`. No idea how to do the last part, the method is from Core Media without any visibility into how it works under the hood

stuartmorgan:
If you have a black-box system function whose output the code under test depends on, you can create a trivial wrapper around it that is DI'ed (as a block) using the same approach. The real version of the block would be nothing but a direct passthrough to the black-box function, and the test would instead inject a version that checks that the inputs are as expected, and gives dummy output.

sergeidesenko:
@stuartmorgan Thank you for your help! I changed the tests like you suggested

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraTestUtils.m
```diff
@@ -29,6 +29,51 @@
...
 }
 
+FLTCam *FLTCreateCamWithVideoCaptureSession(AVCaptureSession *captureSession,
+                                            NSString *resolutionPreset) {
+  id inputMock = OCMClassMock([AVCaptureDeviceInput class]);
+  OCMStub([inputMock deviceInputWithDevice:[OCMArg any] error:[OCMArg setTo:nil]])
+      .andReturn(inputMock);
+
+  id audioSessionMock = OCMClassMock([AVCaptureSession class]);
+  OCMStub([audioSessionMock addInputWithNoConnections:[OCMArg any]]);  // no-op
```
stuartmorgan, line 39: Same question here and below. (If this is meant to say that this is stubbing it out with a no-op, the comment should be more explicit.)

packages/camera/camera_avfoundation/ios/Classes/FLTCam.h
```diff
@@ -14,6 +14,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef CMVideoDimensions (^VideoDimensionsForFormatBlock)(AVCaptureDeviceFormat *);
```
stuartmorgan, line 17: Needs a declaration comment.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.h
```diff
@@ -45,6 +47,19 @@ NS_ASSUME_NONNULL_BEGIN
                        orientation:(UIDeviceOrientation)orientation
                captureSessionQueue:(dispatch_queue_t)captureSessionQueue
                              error:(NSError **)error;
+
+- (instancetype)initWithCameraName:(NSString *)cameraName
```
stuartmorgan, line 51: Needs a declaration comment.

Also, does this really need to be here instead of `FLTCam_Test.h`?

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -86,6 +86,7 @@ @interface FLTCam () <AVCaptureVideoDataOutputSampleBufferDelegate,
 /// Videos are written to disk by `videoAdaptor` on an internal queue managed by AVFoundation.
 @property(strong, nonatomic) dispatch_queue_t photoIOQueue;
 @property(assign, nonatomic) UIDeviceOrientation deviceOrientation;
+@property(nonatomic, copy) VideoDimensionsForFormatBlock videoDimensionsForFormatBlock;
```
stuartmorgan, line 89: Needs a declaration comment.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -99,13 +100,15 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
...
-              captureSessionQueue:captureSessionQueue
-                            error:error];
+                   resolutionPreset:resolutionPreset
+                        enableAudio:enableAudio
+                        orientation:orientation
+                videoCaptureSession:[[AVCaptureSession alloc] init]
+                audioCaptureSession:[[AVCaptureSession alloc] init]
+                captureSessionQueue:captureSessionQueue
+                      captureDevice:nil
+      videoDimensionsForFormatBlock:nil
```
stuartmorgan, line 110: This should be passing real values, not `nil`; DI behavioral difference should come from the DI itself, not logic in the code.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -116,6 +119,29 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
...
                              error:(NSError **)error {
+  return [self initWithCameraName:cameraName
+                   resolutionPreset:resolutionPreset
+                        enableAudio:enableAudio
+                        orientation:orientation
+                videoCaptureSession:videoCaptureSession
+                audioCaptureSession:videoCaptureSession
+                captureSessionQueue:captureSessionQueue
+                      captureDevice:nil
+      videoDimensionsForFormatBlock:nil
```
stuartmorgan, line 130: Same.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -136,7 +162,12 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
   _photoIOQueue = dispatch_queue_create("io.flutter.camera.photoIOQueue", NULL);
   _videoCaptureSession = videoCaptureSession;
   _audioCaptureSession = audioCaptureSession;
-  _captureDevice = [AVCaptureDevice deviceWithUniqueID:cameraName];
+  if (captureDevice) {
+    _captureDevice = captureDevice;
+  } else {
+    _captureDevice = [AVCaptureDevice deviceWithUniqueID:cameraName];
```
stuartmorgan, line 168: This would move up to the convenience inits (at which point it looks like the `cameraName` parameter can be removed from this inner init method).

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -422,6 +475,32 @@ - (BOOL)setCaptureSessionPreset:(FLTResolutionPreset)resolutionPreset withError:
...
+  }
+  return bestFormat;
+}
+
+/// Wrapper for CMVideoFormatDescriptionGetDimensions to use custom code for testing
+- (CMVideoDimensions)videoDimensionsForFormat:(AVCaptureDeviceFormat *)format {
+  if (self.videoDimensionsForFormatBlock) {
+    return self.videoDimensionsForFormatBlock(format);
+  }
+  return CMVideoFormatDescriptionGetDimensions(format.formatDescription);
```
stuartmorgan, line 501: This also goes in the inits, and this whole wrapper method goes away.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
sergeidesenko: Since captureDevice is also set in the `setDescriptionWhileRecording` method, I went with another block for creating captureDevice. Got rid of `cameraName`

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraSessionPresetsTest.m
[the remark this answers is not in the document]
sergeidesenko: fixed

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraSessionPresetsTest.m
```diff
@@ -0,0 +1,77 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import camera_avfoundation;
+@import camera_avfoundation.Test;
+@import AVFoundation;
+@import XCTest;
+#import <OCMock/OCMock.h>
+#import "CameraTestUtils.h"
```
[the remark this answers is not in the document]
sergeidesenko: fixed

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraSessionPresetsTest.m
```diff
@@ -0,0 +1,77 @@
...
+  FLTCreateCamWithVideoCaptureSession(videoSessionMock, @"max");
+
+  OCMVerifyAll(videoSessionMock);
+}
+
+- (void)testResolutionPresetWithCanSetSessionPresetUltraHigh_mustUpdateCaptureSessionPreset {
+  NSString *expectedPreset = AVCaptureSessionPreset3840x2160;
+
+  id videoSessionMock = OCMClassMock([AVCaptureSession class]);
+  OCMStub([videoSessionMock addInputWithNoConnections:[OCMArg any]]);  // no-op
```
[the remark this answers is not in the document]
sergeidesenko: Removed no-op comments for this line, seems unnecessary

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraSessionPresetsTest.m
```diff
@@ -0,0 +1,77 @@
...
+  OCMVerifyAll(videoSessionMock);
+}
+
+- (void)testResolutionPresetWithCanSetSessionPresetMax_mustUpdateCaptureSessionPreset {
+  NSString *expectedPreset = AVCaptureSessionPreset3840x2160;
+
+  id videoSessionMock = OCMClassMock([AVCaptureSession class]);
+  OCMStub([videoSessionMock addInputWithNoConnections:[OCMArg any]]);
+
+  // make sure that setting resolution preset for session always succeeds
```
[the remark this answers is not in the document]
sergeidesenko: I think I got all of them now

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraTestUtils.m
sergeidesenko: removed

packages/camera/camera_avfoundation/ios/Classes/FLTCam.h
sergeidesenko: fixed

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
sergeidesenko: fixed

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
sergeidesenko: fixed

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
sergeidesenko: fixed

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
sergeidesenko: fixed

sergeidesenko:
@stuartmorgan Ready for re-review

packages/camera/camera_avfoundation/ios/Classes/FLTCam_Test.h
```diff
@@ -5,6 +5,37 @@
 #import "FLTCam.h"
 #import "FLTSavePhotoDelegate.h"
 
+/**
+ A block type definition named VideoDimensionsForFormatBlock.
```
stuartmorgan, line 9: See previous comments; this is a restatement of the line of code, and thus doesn't add any information.

packages/camera/camera_avfoundation/ios/Classes/FLTCam_Test.h
```diff
@@ -5,6 +5,37 @@
...
+
+ This block is intended to be used for determining the video dimensions (width and height) for a
+ given capture device format. It accepts an AVCaptureDeviceFormat object as its input and returns a
+ CMVideoDimensions struct representing the video dimensions associated with that format.
```
stuartmorgan, line 13: See previous comments. The goal is to explain to someone who knows Obj-C but not this specific code what this is, not to explain to someone who doesn't know anything about the language what this is.

packages/camera/camera_avfoundation/ios/Classes/FLTCam_Test.h
```diff
@@ -5,6 +5,37 @@
...
+
+ Parameters:
+ - format: An AVCaptureDeviceFormat object representing the format for which the video dimensions
+ are being queried.
+
+ Returns:
+ A CMVideoDimensions struct containing the width and height (in pixels) of the video associated with
+ the provided format.
```
stuartmorgan, line 21: If you are going to do formal parameter and return documentation (which is optional), it needs to use Doxygen format for consistency.

packages/camera/camera_avfoundation/ios/Classes/FLTCam_Test.h
```diff
@@ -5,6 +5,37 @@
...
+
+ This block is intended for use when you need to obtain an instance of AVCaptureDevice without
+ requiring any input parameters. It is particularly useful in scenarios where the desired capture
+ device is predefined or does not depend on dynamic conditions.
+
+ Returns:
+ An AVCaptureDevice instance. The specific instance returned by the block can be predefined within
+ the block's implementation, allowing for consistent access to a particular device, such as the
+ default camera or microphone, without the need for specifying its unique ID each time.
+ */
```
stuartmorgan, line 36: All the same comments apply here.

packages/camera/camera_avfoundation/ios/Classes/FLTCam_Test.h
```diff
@@ -5,6 +5,37 @@
 #import "FLTCam.h"
 #import "FLTSavePhotoDelegate.h"
 
+/**
+ A block type definition named VideoDimensionsForFormatBlock.
+
+ This block is intended to be used for determining the video dimensions (width and height) for a
```
stuartmorgan, line 11: "This block is intended to be used for" adds a lot of unnecessary verbosity; people understand that the purpose of a declaration comment is to explain what something is used for.

packages/camera/camera_avfoundation/ios/Classes/FLTCam_Test.h
```diff
@@ -55,6 +86,37 @@
                captureSessionQueue:(dispatch_queue_t)captureSessionQueue
                              error:(NSError **)error;
 
+/**
+ Initializes an instance for testing with specified resolution, audio preference, orientation, and
+ direct access to capture sessions and blocks.
+
+ This initializer allows the explicit configuration of media capture settings and direct
+ manipulation of internal components like capture sessions and device selection blocks, facilitating
+ thorough testing scenarios.
```
stuartmorgan, line 95: Readers can be expected to understand the purpose of declarations that are in a `_Test.h` file; you don't have to explain that part.

stuartmorgan (changes requested):

packages/camera/camera_avfoundation/ios/Classes/FLTCam_Test.h
```diff
@@ -5,6 +5,37 @@
 #import "FLTCam.h"
 #import "FLTSavePhotoDelegate.h"
 
+/**
```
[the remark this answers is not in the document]
sergeidesenko: Screenshot 2024-02-21 at 19 37 08
Screenshot 2024-02-21 at 19 39 11

Sorry, I am not a guru of Obj-c styling conventions. That's from FLTCam.h. Am I missing some important distinction between use cases?

sergeidesenko:
@stuartmorgan 
updated comments. Just one thing - could you take a look at my question about styling? There is a mix of different styles as far as I can see from the existing code

sergeidesenko:
@hellohuanlin 
I believe that all changes you've requested are now in this pr. Could you confirm?

packages/camera/camera_avfoundation/ios/Classes/FLTCam_Test.h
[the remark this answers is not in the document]
stuartmorgan: I didn't realize the package was internally inconsistent already. It should all be using one style or the other.

Given that, please be consistent within each file (or when a file is already inconsistent, which it looks like there's some of, consistent with whatever is dominant in the file).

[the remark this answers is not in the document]
sergeidesenko: No problem! I went through the files - I think my changes should be consistent now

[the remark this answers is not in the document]
stuartmorgan: Thanks. I filed https://github.com/flutter/flutter/issues/143868 to hopefully avoid this in the future.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -86,6 +86,11 @@ @interface FLTCam () <AVCaptureVideoDataOutputSampleBufferDelegate,
 /// Videos are written to disk by `videoAdaptor` on an internal queue managed by AVFoundation.
 @property(strong, nonatomic) dispatch_queue_t photoIOQueue;
 @property(assign, nonatomic) UIDeviceOrientation deviceOrientation;
+/// A wrapper for CMVideoFormatDescriptionGetDimensions.
+/// Allows for alternate implementations in tests.
+@property(nonatomic, copy) VideoDimensionsForFormatBlock videoDimensionsForFormat;
+/// A wrapper for AVCaptureDevice creation to allow for dependency injection in tests.
+@property(nonatomic, copy) CaptureDeviceBlock captureDeviceFactory;
```
hellohuanlin, line 93: can we call this type `CaptureDeviceFactory`?

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -86,6 +86,11 @@ @interface FLTCam () <AVCaptureVideoDataOutputSampleBufferDelegate,
 /// Videos are written to disk by `videoAdaptor` on an internal queue managed by AVFoundation.
 @property(strong, nonatomic) dispatch_queue_t photoIOQueue;
 @property(assign, nonatomic) UIDeviceOrientation deviceOrientation;
+/// A wrapper for CMVideoFormatDescriptionGetDimensions.
+/// Allows for alternate implementations in tests.
+@property(nonatomic, copy) VideoDimensionsForFormatBlock videoDimensionsForFormat;
```
hellohuanlin, line 91: Nit: can we make the type name and variable name consistent?

hellohuanlin (approved):

sergeidesenko:
@stuartmorgan Same question - is there anything else you would like to see changed?

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -366,7 +397,29 @@ - (NSString *)getTemporaryFilePathWithExtension:(NSString *)extension
...
+          [_captureDevice unlockForConfiguration];
+
+          // Set the preview size based on values from the current capture device.
+          _previewSize =
+              CGSizeMake(_captureDevice.activeFormat.highResolutionStillImageDimensions.width,
+                         _captureDevice.activeFormat.highResolutionStillImageDimensions.height);
+          break;
+        }
+      }
+    }
```
stuartmorgan, line 417: Nit: This should be after the code below, enclosing the entire case implementation, otherwise the indentation is very confusing.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -366,7 +397,29 @@ - (NSString *)getTemporaryFilePathWithExtension:(NSString *)extension
 
 - (BOOL)setCaptureSessionPreset:(FLTResolutionPreset)resolutionPreset withError:(NSError **)error {
   switch (resolutionPreset) {
-    case FLTResolutionPresetMax:
+    case FLTResolutionPresetMax: {
+      AVCaptureDeviceFormat *bestFormat =
+          [self highestResolutionFormatForCaptureDevice:_captureDevice];
+      if (bestFormat) {
+        _videoCaptureSession.sessionPreset = AVCaptureSessionPresetInputPriority;
+        if ([_captureDevice lockForConfiguration:NULL] == YES) {
```
stuartmorgan, line 405: Nit: Never compare to `YES`; see go/objc-style#bool-expressions-and-conversions

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -422,6 +475,24 @@ - (BOOL)setCaptureSessionPreset:(FLTResolutionPreset)resolutionPreset withError:
   return YES;
 }
 
+/// Finds the highest available resolution in terms of pixel count for the given device.
+- (AVCaptureDeviceFormat *)highestResolutionFormatForCaptureDevice:
+    (AVCaptureDevice *)captureDevice {
+  AVCaptureDeviceFormat *bestFormat = nil;
+  NSUInteger maxPixelCount = 0;
+  for (AVCaptureDeviceFormat *format in [_captureDevice formats]) {
```
stuartmorgan, line 483: Nit: `in _captureDevice.formats`

stuartmorgan (approved):
A few more nits for things I didn't catch before, but otherwise LGTM. Thanks!

sergeidesenko:
@stuartmorgan @hellohuanlin (sorry, not sure who would be responsible for this)
Is this pr ready for an autosubmit now or are there any additional steps?

stuartmorgan (changes requested):
The next step is to fix the issues I raised above. I had set my status bit to approval to make it easier to land once that was done, but it seems like that created confusion, so updating it to clarify.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
sergeidesenko: fixed everything
In this case I decided to just remove everything related to AVCaptureSessionPreset3840x2160, since it was just a repetition of the next case (FLTResolutionPresetUltraHigh). Now it's more in line with other cases

Out of curiosity - what's your definition or nit in this context? I thought these are [not compulsory](https://google.github.io/eng-practices/review/reviewer/standard.html)

stuartmorgan: I generally use it to indicate that something is a very small change. I wasn't aware there was a doc that equated it with being optional; I'll stop using it given that.

stuartmorgan (approved):

hellohuanlin:
@sergeidesenko FYI This may have caused a regression based on discussion here: https://github.com/flutter/flutter/issues/162376#issuecomment-2710455240

(Sorry for pinging in the PR since I wasn't able to tag you in the issue)

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -935,7 +1001,7 @@ - (void)setDescriptionWhileRecording:(NSString *)cameraName
     return;
   }
 
-  _captureDevice = [AVCaptureDevice deviceWithUniqueID:cameraName];
+  _captureDevice = self.captureDeviceFactory();
```
sergeidesenko, line 1005: @hellohuanlin you are probably right, this line is causing camera not updating. 
Reverting this one line would fix it, I think, and wouldn't break the tests that depend on captureDeviceFactory to work. Unfortunately, I don't think I have enough bandwidth right now to go through with a new pr, which again will require writing new tests and going through a code review. 

Would you consider just reverting this pr altogether?

stuartmorgan: It shouldn't be reverted, the `cameraName` should just be a parameter to the factory method. I missed in review that it was no longer using the local name.

PR #5254, opened 2023-10-28
[google_maps_flutter_web] Migrate to package:web

navaronbracke:
This PR migrates `google_maps_flutter_web` from `dart:html` to `package:web`.

The `google_maps` package does provide a beta version with support for `package:web`, which is currently what this PR is depending on. Before landing this change, we should change the dependency to the stable release version of that package.

Fixes https://github.com/flutter/flutter/issues/137374

N/A

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/marker_test.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/marker_test.dart
@@ -7,9 +7,8 @@ import 'dart:async';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:google_maps/google_maps.dart' as gmaps;
 import 'package:google_maps_flutter_web/google_maps_flutter_web.dart';
+import 'package:google_maps_flutter_web/src/utils.dart';
 import 'package:integration_test/integration_test.dart';
-
-import 'test_utils.dart';
 
 /// Test Markers
 void main() {
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/marker_test.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/marker_test.dart
@@ -7,6 +7,7 @@ import 'dart:async';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:google_maps/google_maps.dart' as gmaps;
 import 'package:google_maps_flutter_web/google_maps_flutter_web.dart';
+// ignore: implementation_imports
 import 'package:google_maps_flutter_web/src/utils.dart';
 import 'package:integration_test/integration_test.dart';
 
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/marker_test.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/marker_test.dart
@@ -3,12 +3,13 @@
 // found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:html' as html;
 
 import 'package:flutter_test/flutter_test.dart';
 import 'package:google_maps/google_maps.dart' as gmaps;
 import 'package:google_maps_flutter_web/google_maps_flutter_web.dart';
 import 'package:integration_test/integration_test.dart';
+
+import 'test_utils.dart';
 
 /// Test Markers
 void main() {
@@ -123,7 +124,7 @@ void main() {
 
     testWidgets('showInfoWindow', (WidgetTester tester) async {
       final gmaps.InfoWindow infoWindow = gmaps.InfoWindow();
-      final gmaps.GMap map = gmaps.GMap(html.DivElement());
+      final gmaps.GMap map = gmaps.GMap(createDivElement());
       marker.set('map', map);
       final MarkerController controller = MarkerController(
         marker: marker,
@@ -138,7 +139,7 @@ void main() {
 
     testWidgets('hideInfoWindow', (WidgetTester tester) async {
       final gmaps.InfoWindow infoWindow = gmaps.InfoWindow();
-      final gmaps.GMap map = gmaps.GMap(html.DivElement());
+      final gmaps.GMap map = gmaps.GMap(createDivElement());
       marker.set('map', map);
       final MarkerController controller = MarkerController(
         marker: marker,
@@ -156,7 +157,7 @@ void main() {
 
       setUp(() {
         final gmaps.InfoWindow infoWindow = gmaps.InfoWindow();
-        final gmaps.GMap map = gmaps.GMap(html.DivElement());
+        final gmaps.GMap map = gmaps.GMap(createDivElement());
         marker.set('map', map);
         controller = MarkerController(marker: marker, infoWindow: infoWindow);
       });
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/lib/main.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/lib/main.dart
@@ -20,6 +20,9 @@ class MyApp extends StatefulWidget {
 class _MyAppState extends State<MyApp> {
   @override
   Widget build(BuildContext context) {
-    return const Text('Testing... Look at the console output for results!');
+    return const Directionality(
+      textDirection: TextDirection.ltr,
+      child: Text('Testing... Look at the console output for results!'),
+    );
   }
 }
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
@@ -18,7 +18,7 @@ dev_dependencies:
   build_runner: ^2.1.1
   flutter_test:
     sdk: flutter
-  google_maps: 7.0.0-beta.1 # TODO: migrate to stable 7.0.0 once published
+  google_maps: ^7.0.0
   google_maps_flutter: ^2.2.0 # Needed for projection_test.dart
   http: ">=0.13.0 <2.0.0"
   integration_test:
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
@@ -12,13 +12,13 @@ dependencies:
   google_maps_flutter_platform_interface: ^2.4.0
   google_maps_flutter_web:
     path: ../
-  web: "^0.5.0"
+  web: ^0.5.0
 
 dev_dependencies:
   build_runner: ^2.1.1
   flutter_test:
     sdk: flutter
-  google_maps: ^7.0.0
+  google_maps: ^7.1.0
   google_maps_flutter: ^2.2.0 # Needed for projection_test.dart
   http: ">=0.13.0 <2.0.0"
   integration_test:
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
@@ -3,7 +3,7 @@ publish_to: none
 
 # Tests require flutter beta or greater to run.
 environment:
-  sdk: "^3.3.0"
+  sdk: ^3.3.0
   flutter: ">=3.19.0"
 
 dependencies:
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
@@ -12,13 +12,13 @@ dependencies:
   google_maps_flutter_platform_interface: ^2.4.0
   google_maps_flutter_web:
     path: ../
-  web: ^0.3.0
+  web: ^0.4.0
 
 dev_dependencies:
   build_runner: ^2.1.1
   flutter_test:
     sdk: flutter
-  google_maps: ^6.1.0
+  google_maps: 7.0.0-beta.1 # TODO: migrate to stable 7.0.0 once published
   google_maps_flutter: ^2.2.0 # Needed for projection_test.dart
   http: ">=0.13.0 <2.0.0"
   integration_test:
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
@@ -3,8 +3,8 @@ publish_to: none
 
 # Tests require flutter beta or greater to run.
 environment:
-  sdk: ^3.1.0
-  flutter: ">=3.13.0"
+  sdk: "^3.3.0"
+  flutter: ">=3.19.0"
 
 dependencies:
   flutter:
@@ -12,7 +12,7 @@ dependencies:
   google_maps_flutter_platform_interface: ^2.4.0
   google_maps_flutter_web:
     path: ../
-  web: ^0.4.0
+  web: "^0.5.0"
 
 dev_dependencies:
   build_runner: ^2.1.1
--- a/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
+++ b/packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
@@ -12,6 +12,7 @@ dependencies:
   google_maps_flutter_platform_interface: ^2.4.0
   google_maps_flutter_web:
     path: ../
+  web: ^0.3.0
 
 dev_dependencies:
   build_runner: ^2.1.1
--- a/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
@@ -23,6 +23,7 @@ import 'package:web/web.dart';
 
 import 'src/dom_window_extension.dart';
 import 'src/google_maps_inspector_web.dart';
+import 'src/map_styler.dart';
 import 'src/third_party/to_screen_location/to_screen_location.dart';
 import 'src/types.dart';
 import 'src/utils.dart';
--- a/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
@@ -22,6 +22,7 @@ import 'package:sanitize_html/sanitize_html.dart';
 import 'package:stream_transform/stream_transform.dart';
 import 'package:web/web.dart';
 
+import 'src/dom_window_extension.dart';
 import 'src/google_maps_inspector_web.dart';
 import 'src/third_party/to_screen_location/to_screen_location.dart';
 import 'src/types.dart';
--- a/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
@@ -6,6 +6,7 @@ library google_maps_flutter_web;
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:js_interop';
 import 'dart:js_util';
 import 'dart:ui_web' as ui_web;
 
@@ -24,6 +25,7 @@ import 'package:web/web.dart';
 import 'src/google_maps_inspector_web.dart';
 import 'src/third_party/to_screen_location/to_screen_location.dart';
 import 'src/types.dart';
+import 'src/utils.dart';
 
 part 'src/circle.dart';
 part 'src/circles.dart';
--- a/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
@@ -6,7 +6,6 @@ library google_maps_flutter_web;
 
 import 'dart:async';
 import 'dart:convert';
-import 'dart:html' hide VoidCallback;
 import 'dart:js_util';
 import 'dart:ui_web' as ui_web;
 
@@ -20,6 +19,7 @@ import 'package:google_maps/google_maps.dart' as gmaps;
 import 'package:google_maps_flutter_platform_interface/google_maps_flutter_platform_interface.dart';
 import 'package:sanitize_html/sanitize_html.dart';
 import 'package:stream_transform/stream_transform.dart';
+import 'package:web/web.dart';
 
 import 'src/google_maps_inspector_web.dart';
 import 'src/third_party/to_screen_location/to_screen_location.dart';
--- a/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
@@ -7,7 +7,6 @@ library google_maps_flutter_web;
 import 'dart:async';
 import 'dart:convert';
 import 'dart:js_interop';
-import 'dart:js_util';
 import 'dart:ui_web' as ui_web;
 
 import 'package:collection/collection.dart';
--- a/packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
@@ -221,7 +221,7 @@ gmaps.InfoWindowOptions? _infoWindowOptionsFromMarker(Marker marker) {
       // See: https://pub.dev/documentation/sanitize_html/latest/sanitize_html/sanitizeHtml.html
       // See: b/159137885, b/159598165
       // ignore: unsafe_html
-      ..innerHTML = sanitizeHtml(markerSnippet);
+      ..innerHTML = _sanitizeHtml(markerSnippet);
     container.appendChild(snippet);
   }
 
@@ -524,4 +524,28 @@ gmaps.LatLng _pixelToLatLng(gmaps.GMap map, int x, int y) {
       gmaps.Point((x / scale) + bottomLeft.x!, (y / scale) + topRight.y!);
 
   return projection.fromPointToLatLng!(point)!;
+}
+
+String _sanitizeHtml(String htmlString) {
+  TrustedTypePolicy? trustedTypePolicy;
+
+  try {
+    trustedTypePolicy = TrustedTypePolicyFactory().createPolicy(
+      'google_maps_flutter_sanitize',
+      TrustedTypePolicyOptions(
+        createHTML: (String html) {
+          return sanitizeHtml(html);
+        }.toJS,
+      ),
+    );
+  } catch (_) {
+    // Firefox and Safari don't support Trusted Types yet.
+    // See https://developer.mozilla.org/en-US/docs/Web/API/TrustedTypePolicyFactory#browser_compatibility
+  }
+
+  if (trustedTypePolicy == null) {
+    return sanitizeHtml(htmlString);
+  }
+
+  return trustedTypePolicy.createHTML(htmlString, null).toString();
 }
--- a/packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
@@ -203,30 +203,25 @@ gmaps.InfoWindowOptions? _infoWindowOptionsFromMarker(Marker marker) {
 
   // Add an outer wrapper to the contents of the infowindow, we need it to listen
   // to click events...
-  final HtmlElement container = DivElement()
+  final HTMLElement container = createDivElement()
     ..id = 'gmaps-marker-${marker.markerId.value}-infowindow';
 
   if (markerTitle.isNotEmpty) {
-    final HtmlElement title = HeadingElement.h3()
+    final HTMLHeadingElement title = (document.createElement('h3') as HTMLHeadingElement)
       ..className = 'infowindow-title'
       ..innerText = markerTitle;
-    container.children.add(title);
+    container.appendChild(title);
   }
   if (markerSnippet.isNotEmpty) {
-    final HtmlElement snippet = DivElement()
+    final HTMLElement snippet = createDivElement()
       ..className = 'infowindow-snippet'
       // `sanitizeHtml` is used to clean the (potential) user input from (potential)
       // XSS attacks through the contents of the marker InfoWindow.
       // See: https://pub.dev/documentation/sanitize_html/latest/sanitize_html/sanitizeHtml.html
       // See: b/159137885, b/159598165
-      // The NodeTreeSanitizer.trusted just tells setInnerHtml to leave the output
-      // of `sanitizeHtml` untouched.
       // ignore: unsafe_html
-      ..setInnerHtml(
-        sanitizeHtml(markerSnippet),
-        treeSanitizer: NodeTreeSanitizer.trusted,
-      );
-    container.children.add(snippet);
+      ..innerHTML = sanitizeHtml(markerSnippet);
+    container.appendChild(snippet);
   }
 
   return gmaps.InfoWindowOptions()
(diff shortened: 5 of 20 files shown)


packages/google_maps_flutter/google_maps_flutter_web/lib/src/utils.dart
```diff
@@ -0,0 +1,6 @@
+import 'package:web/web.dart';
+
+/// Convenience method to create a new [HTMLDivElement] element.
+HTMLDivElement createDivElement() {
```
navaronbracke, line 4: I didn't find a way to have this utility for both `lib/src` and the example app, while keeping it private to the package. So I kept this copy internal only and added a second version for the example app's integration tests.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -203,30 +203,25 @@ gmaps.InfoWindowOptions? _infoWindowOptionsFromMarker(Marker marker) {
...
       // See: b/159137885, b/159598165
-      // The NodeTreeSanitizer.trusted just tells setInnerHtml to leave the output
-      // of `sanitizeHtml` untouched.
       // ignore: unsafe_html
-      ..setInnerHtml(
-        sanitizeHtml(markerSnippet),
-        treeSanitizer: NodeTreeSanitizer.trusted,
-      );
-    container.children.add(snippet);
+      ..innerHTML = sanitizeHtml(markerSnippet);
```
navaronbracke, line 223: There is no `NodeSanitizer` anymore, not sure of the implications.

ditman:
There's progress of the dart-google-maps package migration, here:

https://github.com/a14n/dart-google-maps/issues/122#issuecomment-1865216323

navaronbracke:
I've updated the branch to use the new beta version of `google_maps`, and made some last adjustments.
Next, I'll have to run the tests to see if any fail. Once the tests pass, we can adjust the version of `google_maps` and get this into a reviewable state.

navaronbracke:
The `google_maps_flutter_web` integration tests now pass locally, but I do see a different error in the console still?

<details>

<summary>Logs</summary>

```
js_primitives.dart:30 ══╡ EXCEPTION CAUGHT BY GESTURES LIBRARY ╞══════════════════════════════════════════════════════════
js_primitives.dart:30 The following StateError was thrown while handling a pointer data packet:
js_primitives.dart:30 Bad state: No element
js_primitives.dart:30 
js_primitives.dart:30 When the exception was thrown, this was the stack:
js_primitives.dart:30 dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 297:3       throw_
js_primitives.dart:30 dart-sdk/lib/core/iterable.dart 702:5                                             firstWhere
js_primitives.dart:30 packages/flutter_test/src/test_text_input_key_handler.dart.js 11893:63            handlePointerEvent
js_primitives.dart:30 packages/flutter/src/gestures/binding.dart.js 370:14                              [_flushPointerEventQueue]
js_primitives.dart:30 packages/flutter/src/gestures/binding.dart.js 345:40                              [_handlePointerDataPacket]
js_primitives.dart:30 lib/_engine/engine/platform_dispatcher.dart 1361:5                                invoke1
js_primitives.dart:30 lib/_engine/engine/platform_dispatcher.dart 286:5                                 invokeOnPointerDataPacket
js_primitives.dart:30 lib/_engine/engine/pointer_binding.dart 398:30                                    [_sendToFramework]
js_primitives.dart:30 lib/_engine/engine/pointer_binding.dart 224:7                                     onPointerData
js_primitives.dart:30 lib/_engine/engine/pointer_binding.dart 978:16                                    <fn>
js_primitives.dart:30 lib/_engine/engine/pointer_binding.dart 930:7                                     <fn>
js_primitives.dart:30 lib/_engine/engine/pointer_binding.dart 531:9                                     loggedHandler
js_primitives.dart:30 dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 550:37  _checkAndCall
js_primitives.dart:30 dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 555:39  dcall
js_primitives.dart:30 ════════════════════════════════════════════════════════════════════════════════════════════════════
js_primitives.dart:30 00:00 +0: TileOverlaysController addTileOverlays
js_primitives.dart:30 00:00 +1: TileOverlaysController changeTileOverlays
js_primitives.dart:30 00:00 +2: TileOverlaysController updating the z index of a hidden layer does not make it visible
js_primitives.dart:30 00:00 +3: TileOverlaysController removeTileOverlays
js_primitives.dart:30 00:00 +4: TileOverlaysController clearTileCache
js_primitives.dart:30 00:01 +5: (tearDownAll)
js_primitives.dart:30 00:01 +6: All tests passed!
js_primitives.dart:30 No widgets found at Offset(1448.7, 217.7).
```

</details>

packages/google_maps_flutter/google_maps_flutter_web/example/lib/main.dart
```diff
@@ -20,6 +20,9 @@ class MyApp extends StatefulWidget {
 class _MyAppState extends State<MyApp> {
   @override
   Widget build(BuildContext context) {
-    return const Text('Testing... Look at the console output for results!');
+    return const Directionality(
```
navaronbracke, line 23: I accidentally ran the example app once and it did throw an error about a missing Directionality widget

navaronbracke:
I have three failing integration tests, which all relate to the bitmap of the test Tile. Not sure what the resolution is? (this refactor should not have affected the loading of the bitmap)

- The `markers with custom bitmap icon work` test in markers_test.dart fails because the list of bytes does not match the expected list
- The `produces image tiles` test in overlay_test.dart fails due to the image element in the test triggering its `onerror` handler, even though we only try to load a transparent 16x16 bitmap which we don't get over the network (we decode it ourselves)?
- The `update` test in overlay_test.dart fails the same way as the `produces image tiles` test

ditman:
We have a thread in discord about the test failures here: 

* https://discord.com/channels/608014603317936148/608020293944082452/1199091259365662882

navaronbracke:
@ditman The test failure was resolved on my end. Turns out you were right about JS converting stuff to strings, as mentioned in https://github.com/dart-lang/web/issues/91#issuecomment-1789465192 See the last commit for that fix.

Not sure why there's only 4 CI tasks now? Is that due to the failing ci.yaml configuration?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/utils.dart
ditman: One thing you can do with this is import from `package:google_maps_flutter_web/lib/src/utils.dart` (a "private" file), which should be allowed from tests (?)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: I think the "safe" way to solve this now is to :

* https://developer.mozilla.org/en-US/docs/Web/API/TrustedHTML

Then define a setter for `HTMLElement` that lets us set the `innerHTML` property with a `TrustedHTML` instance, rather than the current `Node` or `HTMLElement` type.

The `sanitizeHtml` method could be the implementation of the Trusted Type policy, like described in the MDN example:

```js
const escapeHTMLPolicy = trustedTypes.createPolicy("flutter-sanitize-maps", {
  createHTML: (String html) => sanitizeHtml(html), // Returns an instance of TrustedHTML
});

let el = document.getElementById("myDiv");
const escaped = escapeHTMLPolicy.createHTML(markerSnippet);
el.safeInnerHTML = escaped; // safeInnerHTML only accepts TrustedHTML objects.
```

packages/google_maps_flutter/google_maps_flutter_web/example/lib/main.dart
ditman: This is fine, the skeleton app is only used to have a Flutter context where things run, but IIRC this particular test does not render a map or anything.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -274,8 +270,15 @@ gmaps.Icon? _gmIconFromBitmapDescriptor(BitmapDescriptor bitmapDescriptor) {
...
     final List<int> bytes = iconConfig[1]! as List<int>;
     // Create a Blob from bytes, but let the browser figure out the encoding
-    final Blob blob = Blob(<dynamic>[bytes]);
-    icon = gmaps.Icon()..url = Url.createObjectUrlFromBlob(blob);
+    final Blob blob;
+
+    if (bytes is Uint8List) {
+      blob = Blob(<JSUint8Array>[bytes.toJS].toJS);
+    } else {
+      blob = Blob(<JSUint8Array>[Uint8List.fromList(bytes).toJS].toJS);
```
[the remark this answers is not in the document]
ditman: Have you found a case where `bytes` is *not* a `Uint8List`? (is the else ever going to run?)

If this is just to cover the "else", just in case, I'd prefer to do:

```dart
assert(bytes is Uint8List, 'bytes must be an Uint8List (or something helpful)');
// Create the blob
```

And then see if anybody trips in that assertion :)

ditman:
@navaronbracke thanks for keeping this PR fresh, if you don't mind I may be able to push stuff to this branch to fix the annoying bits of CI?

navaronbracke:
Feel free to do so, I'll give it some attention regarding your review comments as well.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
navaronbracke: The support for `TrustedHTML` is a bit lacking though, neither Safari nor Firefox doesn't support it.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -521,3 +525,27 @@ gmaps.LatLng _pixelToLatLng(gmaps.GMap map, int x, int y) {
 
   return projection.fromPointToLatLng!(point)!;
 }
+
+String _sanitizeHtml(String htmlString) {
```
navaronbracke, line 529: This is a best effort attempt, since Firefox and Safari will fail to create a policy.

ditman: Yes, this is unfortunately the case. I forgot linking this before, but here's what I did the last time I did something like this:

* https://github.com/flutter/packages/blob/main/packages/google_identity_services_web/lib/src/js_loader.dart#L27-L53

And here's the tweaks to add the special getters/setters to make things slightly easier:

* https://github.com/flutter/packages/blob/main/packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart (for example: making window.trustedTypes nullable, or setting a trustedType as the src of a Script element)

navaronbracke: Yeah, I recognize that pattern, I had to do something similar in a different package for yet another API that wasn't supported. I will add this adjustment as well.

navaronbracke:
@ditman What version bump should this be for `google_maps_flutter_web` ? I have currently set it to version `0.5.5`, as the version is still pre-1.0. However, it would force users to upgrade to the latest stable. Do we pick 0.6.0 instead then?

packages/google_maps_flutter/google_maps_flutter_web/pubspec.yaml
```diff
@@ -22,10 +22,11 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  google_maps: ^6.1.0
+  google_maps: ^7.0.0
   google_maps_flutter_platform_interface: ^2.4.0
   sanitize_html: ^2.0.0
   stream_transform: ^2.0.0
+  web: ">=0.3.0 <0.5.0"
```
kevmoo, line 29:
```suggestion
  web: ^0.5.0
```

We should now move to the latest and greater here!

navaronbracke: Well, `google_maps` will have to upgrade as well then. cc @a14n 

```
Resolving dependencies... (1.2s)
Because google_maps 7.0.0 depends on web ^0.4.0 and no versions of google_maps match >7.0.0 <8.0.0, google_maps ^7.0.0 requires web ^0.4.0.
So, because google_maps_flutter_web_integration_tests depends on both web ^0.5.0 and google_maps ^7.0.0, version solving failed.
```

I can send a patch out for that, though.

kevmoo: thank you!

navaronbracke: This is unblocked in google_maps, so I'll update this PR tomorrow with the new constraint.

navaronbracke:
@kevmoo With `package:web 0.5.0` the `TrustedTypes` binding was removed, however, I needed it for a small part here.

Could you look into landing https://github.com/dart-lang/web/commit/975e55cbe4ac683c33d5256a91f4133fa97b4c34 officially as 0.5.1 ?

navaronbracke:
> @navaronbracke – you can always just add your own extensions or helpers, too. Copy-paste here is fine.

I'm aware of that, I did this for the nullable TrustedTypes in this PR. However, the `TrustedTypes` definition was removed between `0.4.2` and `0.5.0`. I guess I'll just copy over the definition.

packages/google_maps_flutter/google_maps_flutter_web/pubspec.yaml
navaronbracke: Done!

ditman:
`Because every version of google_sign_in_web from path depends on web >=0.3.0 <0.5.0 and every version of google_maps_flutter_web from path depends on web ^0.5.0, google_sign_in_web from path is incompatible with google_maps_flutter_web from path.
So, because all_packages depends on both google_maps_flutter_web from path and google_sign_in_web from path, version solving failed.
Failed to generate native build files via 'flutter pub get'`

Hmm, I think I've updated all the packages in the repo to `web: ^0.5.0`, we need to update this branch with the latest `main` (here's [google_maps_flutter_web's pubspec](https://github.com/flutter/packages/blob/fe6111ba54651ea793879070f90b8df9a10feb58/packages/google_sign_in/google_sign_in_web/pubspec.yaml#L28))

(It's a little bit late now, but I can tackle the update next week, haven't had the time to touch this PR this week!)

navaronbracke:
> `Because every version of google_sign_in_web from path depends on web >=0.3.0 <0.5.0 and every version of google_maps_flutter_web from path depends on web ^0.5.0, google_sign_in_web from path is incompatible with google_maps_flutter_web from path. So, because all_packages depends on both google_maps_flutter_web from path and google_sign_in_web from path, version solving failed. Failed to generate native build files via 'flutter pub get'`
> 
> Hmm, I think we've updated all the packages in the repo to `web: ^0.5.0`. This branch needs to be rebased/merged with the latest `main` (here's [google_maps_flutter_web's pubspec](https://github.com/flutter/packages/blob/fe6111ba54651ea793879070f90b8df9a10feb58/packages/google_sign_in/google_sign_in_web/pubspec.yaml#L28))
> 
> (It's a little bit late now, but I can tackle the update next week, haven't had the time to touch this PR this week!)

Don't worry, I've rebased it for you :)

One thing I did do, since `package:web` now uses extension types (and we also do in that stub for Trusted Types), I've bumped the Dart SDK / Flutter SDK requirement to the latest stable. I don't think all packages do that right now?
For example `google_identity_services_web` uses `package:web ^0.5.0` but specifies Dart 3.2.0 / Flutter 3.16 as the minimum in the pubspec.yaml (so people on Flutter 3.16 will not be able to use it, since they don't have extension types in their Dart SDK)

ditman:
> package:web ^0.5.0 but specifies Dart 3.2.0 / Flutter 3.16 as the minimum in the pubspec.yaml

Ah yes, you're right, I must have missed that sdk/flutter dependency.

We need to use Dart `^3.3.0` and Flutter `>=3.19` (I don't remember the style exactly, but one wants `^` and the other `>=`)

kevmoo:
Dart is `^3.3.0` and flutter is `>=3.19` !!

kevmoo:
@ditman – we good to go here? Should we do a rebase?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -274,8 +270,16 @@ gmaps.Icon? _gmIconFromBitmapDescriptor(BitmapDescriptor bitmapDescriptor) {
...
-    final Blob blob = Blob(<dynamic>[bytes]);
-    icon = gmaps.Icon()..url = Url.createObjectUrlFromBlob(blob);
+    final Blob blob;
+
+    assert(
+      bytes is Uint8List,
+      'The bytes for a BitmapDescriptor icon must be a Uint8List',
+    );
+
+    blob = Blob(<JSUint8Array>[(bytes as Uint8List).toJS].toJS);
```
ditman, line 280: Oof, this could use a package:web helper? @srujzs? +1 to [this](https://github.com/dart-lang/web/issues/180)?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/dom_window_extension.dart
```diff
@@ -0,0 +1,59 @@
...
+
+import 'package:web/web.dart';
+
+/// This extension type exists to handle unsupported features by certain browsers.
+@JS()
+extension type WindowWithTrustedTypes(Window _) implements JSObject {
+  /// Get the `trustedTypes` object from the window, if it is supported.
+  @JS('trustedTypes')
+  external TrustedTypePolicyFactory? get trustedTypesNullable;
+}
```
ditman, line 15: Another way of doing this:

* https://github.com/flutter/packages/blob/91d11d696196880fe3cd4817c08362e59d7b5c56/packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart#L11-L30

That way when using window from `package:web`, you get the nullable value directly, without having to wrap it in `WindowWithTrustedTypes(window)` to check.

There's also a `has` property in `dart:js_interop_unsafe` that can be used to do: `window.hasProperty('trustedTypes')` (but I have never done so). See: https://github.com/dart-lang/web/issues/181

packages/google_maps_flutter/google_maps_flutter_web/lib/src/dom_window_extension.dart
```diff
@@ -0,0 +1,59 @@
...
+
+// TODO(ditman): remove this extension type when we depend on package:web 0.5.1
+/// This extension exists as a stop gap until `package:web 0.5.1` is released.
+/// That version provides the `TrustedTypes` API.
+@JS()
+extension type TrustedTypePolicyFactory._(JSObject _) implements JSObject {
```
ditman, line 21: @srujzs if the `extension type` has the same name as the one in `package:web`, would we get an error because of the duplicate Dart name? Should this be named something else that is not an exact copy of the one defined in `package:web`?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -521,3 +525,29 @@ gmaps.LatLng _pixelToLatLng(gmaps.GMap map, int x, int y) {
...
+            );
+  } catch (_) {
+    // Fallback to not using the trusted types policy.
+  }
+
+  if (trustedTypePolicy == null) {
+    return sanitizeHtml(htmlString);
+  }
+
+  return trustedTypePolicy.createHTML(htmlString, null).toString();
```
ditman, line 552: Please don't convert the result of calling `createHTML` to String, or we'll be using `unsafe_html` APIs anyway.

If trustedTypes are available, we need to have a `safeInnerHtml` setter on the HTML Object that accepts the `TrustedHTML` object as its contents. For other browsers, we just use the regular innerHtml setter.

See this example:

* https://github.com/flutter/packages/blob/91d11d696196880fe3cd4817c08362e59d7b5c56/packages/google_identity_services_web/lib/src/js_loader.dart#L49-L53

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -521,3 +525,29 @@ gmaps.LatLng _pixelToLatLng(gmaps.GMap map, int x, int y) {
...
 }
+
+String _sanitizeHtml(String htmlString) {
+  TrustedTypePolicy? trustedTypePolicy;
+
+  try {
+    // Firefox and Safari don't support Trusted Types yet.
+    // See https://developer.mozilla.org/en-US/docs/Web/API/TrustedTypePolicyFactory#browser_compatibility
+    trustedTypePolicy =
+        WindowWithTrustedTypes(window).trustedTypesNullable?.createPolicy(
```
ditman, line 536: This is re-creating the `trustedTypesPolicy` for every call to `_sanitizeHtml`, but there's no need to do that, you can just create this policy the first time it's needed, and then reuse it on every call to `_sanitizeHtml`.

ditman (commented):
I think the only blocking comment I have is about using `dart:js_util` instead of `js_interop` in one of the files.

There's a small improvement that can be made to how TrustedTypes are handled, I linked an example of the last time I had to fiddle with those.

(Other than that, I tested the package **in JS mode** with all these changes locally and It Just Works™)

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: > neither Safari nor Firefox support it.

And probably never will, but that's what the `unsafe_html` lint is actually looking at 😱

ditman:
> What version bump should this be for google_maps_flutter_web ? I have currently set it to version 0.5.5, as the version is still pre-1.0. However, it would force users to upgrade to the latest stable? 

@navaronbracke I've been doing 'Y' bumps for all the other package:web migrations, and so far nobody has complained (that I've seen?) 0.5.5 (or 0.5.6, if something else lands first) should be enough.

My operating assumption here is that for those using an older version of flutter, pub get would say: hey, there's a new version 0.5.6 but that's incompatible with your SDK, so you get the latest good, which is 0.5.4 (until they update flutter to the next stable)

packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/marker_test.dart
```diff
@@ -3,11 +3,11 @@
 // found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:html' as html;
 
 import 'package:flutter_test/flutter_test.dart';
 import 'package:google_maps/google_maps.dart' as gmaps;
 import 'package:google_maps_flutter_web/google_maps_flutter_web.dart';
+import 'package:google_maps_flutter_web/src/utils.dart';
```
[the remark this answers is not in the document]
navaronbracke: Done

packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
```diff
@@ -3,21 +3,22 @@ publish_to: none
 
 # Tests require flutter beta or greater to run.
 environment:
-  sdk: ^3.1.0
-  flutter: ">=3.13.0"
+  sdk: "^3.3.0"
```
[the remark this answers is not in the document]
navaronbracke: Done

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
navaronbracke: Added a TODO for later

packages/google_maps_flutter/google_maps_flutter_web/lib/src/dom_window_extension.dart
navaronbracke: Good catch. I kept only the nullable version, since the code already handles the existence (or not) of trusted types.

packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
```diff
@@ -3,21 +3,22 @@ publish_to: none
...
   google_maps_flutter_web:
     path: ../
+  web: ^0.5.0
 
 dev_dependencies:
   build_runner: ^2.1.1
   flutter_test:
     sdk: flutter
-  google_maps: ^6.1.0
+  google_maps: ^7.1.0
```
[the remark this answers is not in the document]
navaronbracke: Do we file a tracking issue for google_maps WASM support?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/dom_window_extension.dart
navaronbracke: Just for the sake of name collisions, I prefixed these stubs with `GoogleMaps`. These will go away in the short term anyway.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
navaronbracke: Added a setter for innerHTML that only accepts `TrustedHTML`.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/map_styler.dart
```diff
@@ -0,0 +1,40 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:js_interop';
+
+/// The interop type for a Google Maps Map Styler.
+///
+/// See: https://developers.google.com/maps/documentation/javascript/style-reference#stylers
```
navaronbracke, line 9: Due to the removal of the `dart:js_util` import, I added this type to be able to remove the `jsify()` call. See the documentation in the link above.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -139,10 +139,11 @@ List<gmaps.MapTypeStyle> _mapStyles(String? mapStyleJson) {
       styles =
           (json.decode(mapStyleJson, reviver: (Object? key, Object? value) {
         if (value is Map && _isJsonMapStyle(value as Map<String, Object?>)) {
-          List<Object?> stylers = <Object?>[];
+          List<MapStyler> stylers = <MapStyler>[];
           if (value['stylers'] != null) {
             stylers = (value['stylers']! as List<Object?>)
-                .map<Object?>((Object? e) => e != null ? jsify(e) : null)
+                .whereType<Map<String, Object?>>()
```
navaronbracke, line 145: In the old version, we did pass `null`s to the styler objects?

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -203,30 +204,45 @@ gmaps.InfoWindowOptions? _infoWindowOptionsFromMarker(Marker marker) {
...
   }
   if (markerSnippet.isNotEmpty) {
-    final HtmlElement snippet = DivElement()
-      ..className = 'infowindow-snippet'
+    final HTMLElement snippet = createDivElement()
+      ..className = 'infowindow-snippet';
+
+    // Firefox and Safari don't support Trusted Types yet.
+    // See https://developer.mozilla.org/en-US/docs/Web/API/TrustedTypePolicyFactory#browser_compatibility
+    if (window.nullableTrustedTypes != null) {
```
navaronbracke, line 223: I updated this code path to use the TrustedHTML setter, if possible.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/dom_window_extension.dart
```diff
@@ -0,0 +1,92 @@
...
+  ///
+  /// See: https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API
+  @JS('trustedTypes')
+  external GoogleMapsTrustedTypePolicyFactory? get nullableTrustedTypes;
+}
+
+// TODO(ditman): remove this extension type when we depend on package:web 0.5.1
+/// This extension exists as a stop gap until `package:web 0.5.1` is released.
+/// That version provides the `TrustedTypes` API.
+@JS('TrustedTypePolicyFactory')
```
navaronbracke, line 24: These types are prefixed to avoid name collisions with `package:web` types.

packages/google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
```diff
@@ -6,7 +6,7 @@ library google_maps_flutter_web;
 
 import 'dart:async';
 import 'dart:convert';
-import 'dart:html' hide VoidCallback;
+import 'dart:js_interop';
 import 'dart:js_util';
```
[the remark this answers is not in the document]
navaronbracke: Done

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
navaronbracke: Done

packages/google_maps_flutter/google_maps_flutter_web/lib/src/dom_window_extension.dart
srujzs: You'll see an error just because of the collision in Dart names, so if there's code that's importing this and `package:web` and uses the symbol, you'll see a collision. Namespacing the import or renaming via the `@JS` annotation are both valid alternatives.

packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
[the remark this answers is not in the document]
ditman: I think we should, compiling to `--wasm` is the main goal of this migration.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: Tested with the demo app's "night mode" and this seems to work. There's also some tests incoming for this feature in https://github.com/flutter/packages/pull/6205

packages/google_maps_flutter/google_maps_flutter_web/lib/src/dom_window_extension.dart
ditman: Thanks!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
ditman: Ah no need! Just noticed the `getTrustedTypesPolicy` extension.

ditman:
This seems to work fine, and I can't find anything else to say! Let's land this! Thanks for your patience @navaronbracke!

ditman (approved):
Let's go!

kevmoo (approved):

packages/google_maps_flutter/google_maps_flutter_web/example/pubspec.yaml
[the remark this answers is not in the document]
kevmoo: Ah 💩 . Ugh...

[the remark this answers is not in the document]
kevmoo: https://github.com/a14n/dart-google-maps/issues/125

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
srujzs: A (slower) alternative could be `[bytes].jsify() as JSArray<BlobPart>`, but I'm not sure how to make this much more concise while still respecting recursive conversions - food for though for myself.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/map_styler.dart
```diff
@@ -0,0 +1,40 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
srujzs, line 3: I believe you'll need a

```dart
@JS()
library;
```

declaration here for https://github.com/dart-lang/sdk/issues/55039.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/map_styler.dart
```diff
@@ -0,0 +1,40 @@
...
+
+import 'dart:js_interop';
+
+/// The interop type for a Google Maps Map Styler.
+///
+/// See: https://developers.google.com/maps/documentation/javascript/style-reference#stylers
+@JS()
+@staticInterop
```
srujzs, line 11: nits: You don't need `@staticInterop` or `@anonymous` anywhere with extension types.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/map_styler.dart
srujzs: Might also need it in whatever other files have an object literal factory in an extension type.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/map_styler.dart
srujzs: Handled in https://github.com/flutter/packages/pull/6216.

PR #5270, opened 2023-10-30
[video_player] Update iOS/macOS to Pigeon 13

stuartmorgan:
Picks up several breaking changes, including the switch to unwrapped BOOL properties.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
```diff
@@ -512,7 +512,7 @@ - (int64_t)duration {
   return FVPCMTimeToMillis([[[_player currentItem] asset] duration]);
 }
 
-- (void)seekTo:(int)location completionHandler:(void (^)(BOOL))completionHandler {
+- (void)seekTo:(int64_t)location completionHandler:(void (^)(BOOL))completionHandler {
```
stuartmorgan, line 515: We were calling this with `input.position.intValue` previously, which when I removed `intValue` to adjust for the lack of boxing turned into a warning about loss of precision passing an `NSInteger` as an `int`. But this is just used for `CMTimeMake`, which takes an `int64_t`, so it was better to change this type than to pointless cast at the call site and (in theory, but probably never in practice) truncate the value.

packages/video_player/video_player_avfoundation/pigeons/messages.dart
```diff
@@ -7,8 +7,8 @@ import 'package:pigeon/pigeon.dart';
 @ConfigurePigeon(PigeonOptions(
   dartOut: 'lib/src/messages.g.dart',
   dartTestOut: 'test/test_api.g.dart',
-  objcHeaderOut: 'ios/Classes/messages.g.h',
-  objcSourceOut: 'ios/Classes/messages.g.m',
+  objcHeaderOut: 'darwin/Classes/messages.g.h',
+  objcSourceOut: 'darwin/Classes/messages.g.m',
```
stuartmorgan, line 11: I missed this when I move the files to add macOS support.

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
```diff
@@ -709,7 +709,7 @@ - (FVPTextureMessage *)onPlayerSetup:(FVPVideoPlayer *)player
   [eventChannel setStreamHandler:player];
   player.eventChannel = eventChannel;
   self.playersByTextureId[@(textureId)] = player;
-  FVPTextureMessage *result = [FVPTextureMessage makeWithTextureId:@(textureId)];
+  FVPTextureMessage *result = [FVPTextureMessage makeWithTextureId:textureId];
```
hellohuanlin, line 712: is it pigeon's change? the new pigeon supports primitives and the old version does not?

hellohuanlin (approved):

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
stuartmorgan: Yes, I changed Pigeon's Obj-C generator to stop boxing non-nullable values, to make them more idiomatic and less error-prone (given that it turns out Xcode is less reliable about warning for buggy implicit `NSNumber*`->`BOOL` conversions than I thought).

packages/video_player/video_player_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.5.1
+
+* Updates to  Pigeon 13.
```
tarrinneal, line 3: nit '  ' -> ' '.

I can't believe this got merged...

stuartmorgan: Doh! Luckily it shouldn't show up that way on pub.dev.

PR #5273, opened 2023-10-31
[pigeon] Channel name in flutter error.

tarrinneal:
Adds the channel name in the connection error message on all platforms host and flutter api methods.

Also fixes a bug with error handling for swift flutter api void methods.

Also adds method to all generators to help clean up duplicated code.

fixes https://github.com/flutter/flutter/issues/136277

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 14.0.0
+
+* **Breaking Change** [swift] Fixes Flutter Api void return error handling.
+  * This shouldn't be breaking for anyone, but if you were incorrectly getting 
+    success responses, you may now be failing (correctly).
```
tarrinneal, line 5: I believe this is technically breaking, but I'm open to making this 13.1.0 if you think that is more correct.

stuartmorgan: I don't think we need to consider this breaking; if we were just calling the wrong callback method before and now we're not, that's just a bugfix.

stuartmorgan (changes requested):
Sorry I didn't think about this before: hard-coding all of these strings in the generated code will unnecessarily bloat the size of the binary, because each unique string will end up as data in the compiled output. For all languages, we should be using string interpolation so that we aren't creating N new strings constants.

And while compilers would probably handle de-duping for us if we inline them separately, it would be better not to rely on that and avoid duplicating even separate constants, so ideally we should have the generation output something like (example in Dart, but similar for the other languages):
```dart
const String channelName = "big string constant here, once";
final BasicMessageChannel<Object?> channel = BasicMessageChannel<Object?>(
        channelName,
        codec,
        binaryMessenger: _binaryMessenger);
...
      throw PlatformException(
        code: 'channel-error',
        message:
            'Unable to establish connection on channel: "$channelName".',
      );
```

tarrinneal:
> Sorry I didn't think about this before: hard-coding all of these strings in the generated code will unnecessarily bloat the size of the binary, because each unique string will end up as data in the compiled output. For all languages, we should be using string interpolation so that we aren't creating N new strings constants.
> 
> And while compilers would probably handle de-duping for us if we inline them separately, it would be better not to rely on that and avoid duplicating even separate constants, so ideally we should have the generation output something like (example in Dart, but similar for the other languages):
> 
> ```dart
> const String channelName = "big string constant here, once";
> final BasicMessageChannel<Object?> channel = BasicMessageChannel<Object?>(
>         channelName,
>         codec,
>         binaryMessenger: _binaryMessenger);
> ...
>       throw PlatformException(
>         code: 'channel-error',
>         message:
>             'Unable to establish connection on channel: "$channelName".',
>       );
> ```

Makes sense, I made sure to do that in the generators themselves, but I didn't even think about it for the generateds.

This is also a general issue with the duplication of code in the generated files. Something that I'm hoping to tackle some day...

packages/pigeon/example/app/android/app/src/main/java/io/flutter/plugins/Messages.java
```diff
@@ -55,6 +39,28 @@ protected static ArrayList<Object> wrapError(@NonNull Throwable exception) {
...
   }
 
+  @NonNull
+  protected static FlutterError createConnectionError(@NonNull String channelName) {
+    return new FlutterError(
+        "channel-error", "Unable to establish connection on channel: " + channelName + ".", "");
+  }
+
+  /** Error class for passing custom error details to Flutter via a thrown PlatformException. */
+  public static class FlutterError extends RuntimeException {
```
stuartmorgan, line 49: Why did this move to after methods? The old ordering seems clearer.

(Not a big deal if it's hard to change for structural generator reasons.)

packages/pigeon/example/app/android/app/src/main/java/io/flutter/plugins/Messages.java
```diff
@@ -55,6 +39,28 @@ protected static ArrayList<Object> wrapError(@NonNull Throwable exception) {
     return errorList;
   }
 
+  @NonNull
+  protected static FlutterError createConnectionError(@NonNull String channelName) {
```
stuartmorgan, line 43: Why protected rather than private?

packages/pigeon/example/app/windows/runner/messages.g.h
```diff
@@ -147,6 +147,7 @@ class MessageFlutterApi {
   void FlutterMethod(const std::string* a_string,
                      std::function<void(const std::string&)>&& on_success,
                      std::function<void(const FlutterError&)>&& on_error);
+  static FlutterError CreateConnectionError(const std::string channel_name);
```
stuartmorgan, line 150: This should be private, not public.

In fact ideally it should be just be a helper function in the implementation file with internal linkage, so not part of the class and either `static` or in an anonymous namespace. I think we can do that pretty easily by just putting the generation [here](https://github.com/flutter/packages/blob/main/packages/pigeon/lib/cpp_generator.dart#L640).

packages/pigeon/lib/dart_generator.dart
```diff
@@ -705,11 +703,30 @@ if (replyList == null) {
...
+    for (final Api api in root.apis) {
+      if (api.methods.isNotEmpty) {
+        if (api.location == ApiLocation.host) {
+          hasHostApi = true;
+        }
+        if (api.location == ApiLocation.flutter) {
+          hasFlutterApi = true;
+        }
+      }
+    }
```
stuartmorgan, line 717: Optional:
```dart
final bool hasHostApi = root.apis.any((Api api) api.methods.isNotEmpy && api.location == ApiLocation.host);
final bool hasFlutterApi = root.apis.any((Api api) api.methods.isNotEmpy && api.location == ApiLocation.flutter);
```

stuartmorgan: (Same in other generators)

stuartmorgan (approved):
LGTM

packages/pigeon/platform_tests/test_plugin/windows/pigeon/core_tests.gen.cpp
```diff
@@ -25,6 +25,13 @@ using flutter::EncodableList;
 using flutter::EncodableMap;
 using flutter::EncodableValue;
 
+FlutterError CreateConnectionError(const std::string channel_name) {
```
stuartmorgan, line 28: This should either be `static` or in a `namespace { ... }` (i.e., an anonymous namespace).

PR #5285, opened 2023-10-31
[camerax] Implement `lockCaptureOrientation` & `unlockCaptureOrientation`

camsim99:
Implements `lockCaptureOrientation` & `unlockCaptureOrientation` for all camera `UseCase`s. 

Also fixes small bug concerning not initially setting the target rotation of the `UseCase`s to the requested sensor orientation when `createCamera` is called.

Fixes https://github.com/flutter/flutter/issues/125915.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/SystemServicesHostApiImpl.java
camsim99, line 1: Part of DeviceOrientation refactor!

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/PreviewHostApiImpl.java
camsim99, line 1: Adds functionality to dynamically set target rotation of Preview instance.

P.S. Support for setting initial target rotation already supported :)

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/ImageCaptureHostApiImpl.java
camsim99, line 1: Adds support  for setting an initial target rotation of `ImageCapture` instance since this was missed in previous PRs + functionality to dynamically set target rotation.

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/CameraAndroidCameraxPlugin.java
camsim99, line 1: **Note to reviewer:**
Refactored [`DeviceOrientationManager.java`](https://github.com/flutter/packages/blob/14093051d0d1adab80e476aa690c0ea213fe6cb6/packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManager.java) functionality into its own class (previously a part of `SystemServices` [host](https://github.com/flutter/packages/blob/14093051d0d1adab80e476aa690c0ea213fe6cb6/packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/SystemServicesHostApiImpl.java#L95)/[Flutter](https://github.com/flutter/packages/blob/14093051d0d1adab80e476aa690c0ea213fe6cb6/packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/SystemServicesFlutterApiImpl.java#L16) API implementations) for clarity. I will mark* changes related to this throughout this PR to make it clear that these are simply a refactor and no changes in logic occurred.

*Marked with comment: "Part of DeviceOrientation refactor!"

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManagerFlutterApiImpl.java
camsim99, line 1: Part of DeviceOrientation refactor!

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManagerHostApiImpl.java
camsim99, line 1: Part of DeviceOrientation refactor! **But** two new added methods: `getPhotoOrientation`, `getVideoOrientation`.

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/ImageAnalysisHostApiImpl.java
camsim99, line 1: Adds support  for setting an initial target rotation of `ImageAnalysis` instance since this was missed in previous PRs + functionality to dynamically set target rotation.

packages/camera/camera_android_camerax/android/src/test/java/io/flutter/plugins/camerax/DeviceOrientationManagerWrapperTest.java
camsim99, line 1: Part of DeviceOrientation refactor! **But** there are two new tests: `getPhotoOrientation_returnsExpectedOrientation` (testing new `getPhotoOrientation`), `getVideoOrientation_returnsExpectedOrientation` (testing new `getVideoOrientation`).

packages/camera/camera_android_camerax/android/src/test/java/io/flutter/plugins/camerax/ImageCaptureTest.java
camsim99, line 1: Adds testing for setting initial target rotation + dynamically setting target rotation.

packages/camera/camera_android_camerax/android/src/test/java/io/flutter/plugins/camerax/PreviewTest.java
camsim99, line 1: Adds test for dynamically setting target rotation.

P.S. Testing for initially setting target rotation already exists :)

packages/camera/camera_android_camerax/android/src/test/java/io/flutter/plugins/camerax/SystemServicesTest.java
camsim99, line 1: Part of DeviceOrientation refactor!

packages/camera/camera_android_camerax/android/src/test/java/io/flutter/plugins/camerax/VideoCaptureTest.java
camsim99, line 1: Adds test for dynamically setting target rotation.

P.S. Testing for initially setting target rotation not added since I did not implement this functionality.

packages/camera/camera_android_camerax/example/lib/main.dart
camsim99, line 1: Example app changes: Adding back using the `lockCaptureOrientation` and `unlockCaptureOrientation` features and enabling audio since I don't know why I left it disabled :/

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99, line 1: Has all use cases set target rotation based on specified sensor orientation + implements `lockCaptureOrientation` & `unlockCaptureOrientation` + remnants of the DeviceOrientation refactor.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax_flutter_api_impls.dart
camsim99, line 1: Part of DeviceOrientation refactor!

packages/camera/camera_android_camerax/lib/src/camerax_proxy.dart
camsim99, line 1: Updates proxy to support setting initial target rotations + also handle calls to retrieve current photo & video orientation from `DeviceOrientationManager`. For more context on this proxy, see https://github.com/flutter/packages/pull/5423.

packages/camera/camera_android_camerax/lib/src/device_orientation_manager.dart
camsim99, line 1: Part of DeviceOrientation refactor! **But** two new added methods: `getPhotoOrientation`, `getVideoOrientation`.

packages/camera/camera_android_camerax/lib/src/image_analysis.dart
camsim99, line 1: Adds support for setting an initial target rotation of `ImageAnalysis` instance since this was missed in previous PRs + functionality to dynamically set target rotation.

packages/camera/camera_android_camerax/lib/src/image_capture.dart
camsim99, line 1: Adds support for setting an initial target rotation of `ImageCapture` instance since this was missed in previous PRs + functionality to dynamically set target rotation.

packages/camera/camera_android_camerax/lib/src/preview.dart
```diff
@@ -17,6 +17,9 @@ import 'use_case.dart';
 @immutable
```
camsim99, line 17: Adds functionality to dynamically set target rotation of Preview instance.

P.S. Support for setting initial target rotation already supported :)

packages/camera/camera_android_camerax/lib/src/system_services.dart
camsim99, line 1: Part of DeviceOrientation refactor!

packages/camera/camera_android_camerax/lib/src/video_capture.dart
camsim99, line 1: Adds functionality to dynamically set target rotation of `VideoCapture` instance.

P.S. Setting initial target rotation of `VideoCapture` instances would require a refactor away from the current way we are constructing them. If we land this PR as is, I can file an issue to follow up on this later on since it's not required to implement the `lockCaptureOrientation` and `unlockCaptureOrientation` features.

packages/camera/camera_android_camerax/test/android_camera_camerax_test.dart
camsim99, line 1: Adds tests for setting initial target rotation, `lockCaptureOrientation` and `unLockCaptureOrientation` features. Also makes any edits necessary from modifying the proxy.

packages/camera/camera_android_camerax/test/device_orientation_manager_test.dart
camsim99, line 1: Part of DeviceOrientation refactor! But also adds tests for new methods `getPhotoOrientation`, `getVideoOrientation`

packages/camera/camera_android_camerax/test/image_analysis_test.dart
```diff
@@ -30,6 +31,24 @@ void main() {
       TestImageAnalysisHostApi.setup(null);
     });
 
+    test('detached create does not call create on the Java side', () {
```
camsim99, line 34: Added this since it was missing :)

packages/camera/camera_android_camerax/test/system_services_test.dart
camsim99, line 1: Part of the DeviceOrientation refactor!

camsim99 (commented):
Added comments to aid the reviewer 😅

camsim99:
@gmackall As I mentioned offline, this turned into a small monster 😅 I added annotations throughout to make it clear what each file was changing (except for the Dart wrapper tests for the most part which I think are straightforward). Please let me know if I can aid in the review whether it be talking through it or splitting this up into separate wrapper and feature logic PRs.

gmackall (commented):
Only a review of the wrapping code, as discussed offline. But it mostly LGTM, just a couple small comments!

camsim99:
@gmackall Thanks for the review!! Just wanted to update you and say that I now have a better handle of what this is intended to do, so I will follow up in the linked issue soon with an explanation. I also have found that this will require some significant changes, so I will update this in one commit when I finish so it's easy to review the differences :) Apologies for this PR being subpar before review, I really appreciate your time.

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/CameraAndroidCameraxPlugin.java
```diff
@@ -145,6 +148,7 @@ public void onAttachedToActivity(@NonNull ActivityPluginBinding activityPluginBi
     systemServicesHostApiImpl.setActivity(activity);
     systemServicesHostApiImpl.setPermissionsRegistry(
         activityPluginBinding::addRequestPermissionsResultListener);
+    deviceOrientationManagerHostApiImpl.setActivity(activity);
```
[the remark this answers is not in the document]
camsim99: I think so after reading the docs so thanks for pointing this out! I don't want this change to get lost in this PR since it seems like a legitimate bug, so I filed https://github.com/flutter/flutter/issues/139937 and will follow up with a PR.

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManager.java
```diff
@@ -85,120 +84,6 @@ public void stop() {
     broadcastReceiver = null;
   }
 
-  /**
```
camsim99, line 88: Note to reviewer: I removed all unused functionality.

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/PendingRecordingHostApiImpl.java
```diff
@@ -75,7 +75,13 @@ public void handleVideoRecordEvent(@NonNull VideoRecordEvent event) {
...
       VideoRecordEvent.Finalize castedEvent = (VideoRecordEvent.Finalize) event;
       if (castedEvent.hasError()) {
-        systemServicesFlutterApi.sendCameraError(castedEvent.getCause().toString(), reply -> {});
+        String cameraErrorMessage;
+        if (castedEvent.getCause() != null) {
+          cameraErrorMessage = castedEvent.getCause().toString();
+        } else {
+          cameraErrorMessage = "Error code " + castedEvent.getError() + ": An error occurred while recording video.";
+        }
+        systemServicesFlutterApi.sendCameraError(cameraErrorMessage, reply -> {});
```
camsim99, line 84: The [docs](https://developer.android.com/reference/androidx/camera/video/VideoRecordEvent.Finalize) suggest that if `castedEvent.hasError()` is true, then `castedEvent.getCause()` won't be null, but I've seen `castedEvent.getCause().toString()` throw a `NullPointerException` several times, so I added this logic.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -708,7 +779,7 @@ class AndroidCameraCameraX extends CameraPlatform {
           width: imageProxy.width);
 
       weakThis.target!.cameraImageDataStreamController!.add(cameraImageData);
-      unawaited(imageProxy.close());
+      await imageProxy.close();
```
camsim99, line 782: Part of https://github.com/flutter/flutter/issues/132499

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -728,12 +799,6 @@ class AndroidCameraCameraX extends CameraPlatform {
 
   // Methods for configuring image streaming:
 
-  /// The [onListen] callback for the stream controller used for image
-  /// streaming.
-  Future<void> _onFrameStreamListen(int cameraId) async {
-    await _configureImageAnalysis(cameraId);
-  }
-
```
camsim99, line 736: Just changed `_onFrameStreamListen` reference to `_configureImageAnalysis`.

packages/camera/camera_android_camerax/lib/src/image_analysis.dart
```diff
@@ -49,18 +52,33 @@ class ImageAnalysis extends UseCase {
 
   late final _ImageAnalysisHostApiImpl _api;
 
+  /// Initial target rotation of the camera used for the preview stream.
+  ///
+  /// Should be specified in terms of one of the [Surface]
+  /// rotation constants that represents the counter-clockwise degrees of
+  /// rotation relative to [DeviceOrientation.portraitUp].
+  final int? initialTargetRotation;
```
camsim99, line 60: Reflecting on this now and wondering if we should even save the target rotation for `UseCase`s and instead expose methods like https://developer.android.com/reference/androidx/camera/core/ImageCapture#getTargetRotation() that allow us to retrieve the target rotation for each of them. (this is a side effect of these wrapper classes needing to be immutable)

@gmackall lmk what you think. I would probably do this in a follow up PR because this has a lot going on already. Running into a similar thing in the last wrappings I need to do and want to take a similar path.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -370,6 +389,37 @@ class AndroidCameraCameraX extends CameraPlatform {
...
+    if (!shouldSetDefaultRotation) {
+      shouldSetDefaultRotation = true;
+    }
+    captureOrientationLocked = true;
+
+    // Get target rotation based on locked orientation.
+    final int targetLockedRotation =
+        _getRotationConstantFromDeviceOrientation(orientation);
+
+    // Update UseCases to use target device orientation.
```
gmackall, line 410: The Preview usecase also has a method to set the target rotation, should we be setting that as well? 
https://developer.android.com/reference/kotlin/androidx/camera/core/Preview#setTargetRotation(int)

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManager.java
```diff
@@ -265,57 +150,18 @@ PlatformChannel.DeviceOrientation getUIOrientation() {
...
-    int rotation = getDisplay().getRotation();
-    if (((rotation == Surface.ROTATION_0 || rotation == Surface.ROTATION_180)
-            && config.orientation == Configuration.ORIENTATION_LANDSCAPE)
-        || ((rotation == Surface.ROTATION_90 || rotation == Surface.ROTATION_270)
-            && config.orientation == Configuration.ORIENTATION_PORTRAIT)) {
-      return Configuration.ORIENTATION_LANDSCAPE;
-    } else {
-      return Configuration.ORIENTATION_PORTRAIT;
-    }
+  int getDefaultRotation() {
```
gmackall, line 163: nit: I believe this method is named `getDefaultRotation` because it is the rotation that CameraX use cases (at least ImageCapture) default to right? I think we might want to name it something else, though, because without that context from [the comment here](https://developer.android.com/reference/androidx/camera/core/ImageCapture#setTargetRotation(int)), it mostly gets used in places where it looks like we are getting a default rotation of the camera proxy, (i.e. `proxy.getDefaultRotation()`) which at first glance doesn't seem like something that would be a function of the current rotation of the device, but rather a persistent default rotation. 

Basically it just feels more like a `getCurrentRotation` than a `getDefaultRotation`, to me at least. Not a big deal, though.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -493,13 +544,21 @@ class AndroidCameraCameraX extends CameraPlatform {
...
     } else if (torchEnabled) {
       // Ensure any previously set flash modes are unset when torch mode has
       // been enabled.
       await imageCapture!.setFlashMode(ImageCapture.flashModeOff);
     }
+
+    // Set target rotation to current photo orientation only if capture
+    // orientation not locked.
+    if (!captureOrientationLocked && shouldSetDefaultRotation) {
+      await imageCapture!.setTargetRotation(await proxy.getDefaultRotation());
```
gmackall, line 559: In testing, does this achieve the same behavior between "plugin-managed" unlocked case (i.e. lock->unluck), and the "camerax" managed unlocked case? I'm curious because after reading the documentation on `ImageCapture`'s [`setTargetRotation`](https://developer.android.com/reference/androidx/camera/core/ImageCapture#setTargetRotation(int)), particularly:
> If no target rotation is set by the application, it is set to the value of [getRotation](https://developer.android.com/reference/android/view/Display.html#getRotation()) of the default display at the time the use case is bound.

To me this sounds like we would be doing "rotation of default display at time of photo", while camerax is doing "rotation of default display at time of use case creation", which sounds like it would be different if the phone was rotated after creation. Am I misinterpreting the docs, or do the docs not line up with what happens when we test the use cases?

(this comment applies to how we manage the rotation for all use cases, I just put it here as an example)

packages/camera/camera_android_camerax/lib/src/image_analysis.dart
gmackall: I support replacing with the native method, but am also fine with it being a follow up!

packages/camera/camera_android_camerax/lib/src/image_capture.dart
```diff
@@ -28,13 +29,16 @@ class ImageCapture extends UseCase {
         ) {
     _api = ImageCaptureHostApiImpl(
         binaryMessenger: binaryMessenger, instanceManager: instanceManager);
-    _api.createFromInstance(this, targetFlashMode, resolutionSelector);
+    _api.createFromInstance(
+        this, initialTargetRotation, targetFlashMode, resolutionSelector);
   }
 
-  /// Constructs a [ImageCapture] that is not automatically attached to a native object.
+  /// Constructs a [ImageCapture] that is not automatically attached to a
```
gmackall, line 36:
```suggestion
  /// Constructs an [ImageCapture] that is not automatically attached to a
```

gmackall (commented):
Added some comments! My main question is making sure we are managing the "unlocked" state for a given UseCase in the same way that it behaves if it is created and never set/unset.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: I don't believe so since this is only supposed to impact the capture orientation. The preview rotation should be managed by the Flutter app using the plugin like, for example, [our example app](https://github.com/flutter/packages/blob/main/packages/camera/camera_android_camerax/example/lib/camera_preview.dart).

packages/camera/camera_android_camerax/lib/src/image_analysis.dart
camsim99: Ok great! Filed https://github.com/flutter/flutter/issues/140664 to follow up.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: My understanding is that the default display does not change anywhere in the plugin and the rotation of that display always will return the rotation needed to return to its "natural" orientation (https://developer.android.com/reference/android/view/Display.html#getRotation()). From my testing, this yields the desired result of matching `camera_android` behavior.

This makes sense to me because my interpretation of "To return to the default value, set the value to `context.getSystemService(WindowManager.class).getDefaultDisplay().getRotation();`" was that you should be able to do that at anytime to have the `UseCase` return to its default rotation by CameraX standards. I think there would be a concern here if the default display were changing between binding `UseCase`s and setting the target rotation, but like I mentioned, I don't believe it should be a concern of ours.

gmackall (approved):
LGTM!

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManager.java
camsim99: To address this, I renamed the wrapped method to `getDefaultDisplayRotation()` and left this since this really a part of the `DeviceOrientationManager` and would like to keep it consistent with `camera_android` for now.

IsuruRanasinghe1987:
Still this issue exists with the latest flutter version 3.22.0, I don't think is issue was fixed @camsim99

PR #5296, opened 2023-11-02
Updated documentation to decrease issue influx for setup of Google Sign In on Android

Konstanius:
See https://github.com/flutter/flutter/issues/137614


stuartmorgan:
Thanks for the contribution, but extensive documentation about specific UI flows on sites that the Flutter team doesn't control aren't something we want in the README; it makes it very likely to become incorrect frequently due to changes outside our control, when those UI elements are changed on the external pages.

- If there are better official documentation pages to link to, we're happy to review a PR that makes those changes.
- If the pages themselves need better instructions, please pass that feedback directly to the maintainers of those pages.

You might also want to consider another location for these steps, such as a StackOverflow answer.

PR #5347, opened 2023-11-07
[cross_file] adopt code excerpts in README

mike-v2:
Updates the README to use a compiled excerpt source for its example of instantiating an `XFile`.

Part of [flutter/flutter#102679](https://github.com/flutter/flutter/issues/102679)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/cross_file/example/lib/readme_excerpts.dart
+++ b/packages/cross_file/example/lib/readme_excerpts.dart
@@ -17,7 +17,7 @@ Future<XFile> instantiateXFile() async {
   print('- MIME type: ${file.mimeType}');
 
   final String fileContent = await file.readAsString();
-  print('Content of the file: $fileContent'); // e.g. "Moto G (4)"
+  print('Content of the file: $fileContent');
   // #enddocregion Instantiate
 
   return file;
--- a/packages/cross_file/example/lib/readme_excerpts.dart
+++ b/packages/cross_file/example/lib/readme_excerpts.dart
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+// ignore_for_file: avoid_print
+
 import 'package:cross_file/cross_file.dart';
 
 /// Demonstrate instantiating an XFile for the README
@@ -15,7 +17,7 @@ Future<XFile> instantiateXFile() async {
   print('- MIME type: ${file.mimeType}');
 
   final String fileContent = await file.readAsString();
-  print('Content of the file: ${fileContent}'); // e.g. "Moto G (4)"
+  print('Content of the file: $fileContent'); // e.g. "Moto G (4)"
   // #enddocregion Instantiate
 
   return file;
--- a/packages/cross_file/example/lib/readme_excerpts.dart
+++ b/packages/cross_file/example/lib/readme_excerpts.dart
@@ -1,0 +1,22 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:cross_file/cross_file.dart';
+
+/// Demonstrate instantiating an XFile for the README
+Future<XFile> instantiateXFile() async {
+  // #docregion Instantiate
+  final XFile file = XFile('assets/hello.txt');
+
+  print('File information:');
+  print('- Path: ${file.path}');
+  print('- Name: ${file.name}');
+  print('- MIME type: ${file.mimeType}');
+
+  final String fileContent = await file.readAsString();
+  print('Content of the file: ${fileContent}'); // e.g. "Moto G (4)"
+  // #enddocregion Instantiate
+
+  return file;
+}
--- a/packages/cross_file/example/lib/readme_excerpts.dart
+++ b/packages/cross_file/example/lib/readme_excerpts.dart
@@ -6,7 +6,7 @@
 
 import 'package:cross_file/cross_file.dart';
 
-/// Demonstrate instantiating an XFile for the README
+/// Demonstrate instantiating an XFile for the README.
 Future<XFile> instantiateXFile() async {
   // #docregion Instantiate
   final XFile file = XFile('assets/hello.txt');
--- a/packages/cross_file/example/pubspec.yaml
+++ b/packages/cross_file/example/pubspec.yaml
@@ -15,5 +15,4 @@ dependencies:
     path: ../
 
 dev_dependencies:
-  async: ^2.5.0
   test: ^1.24.0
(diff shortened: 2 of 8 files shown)


mike-v2:
The test I made fails because the code in the excerpt file tries to load a text file from the cross_file/tests folder, but I'm not sure how to access that file from the cross_file/example/lib folder.

packages/cross_file/example/lib/readme_excerpts.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:cross_file/cross_file.dart';
```
ditman, line 5: Ignores `avoid_print` lint for this file, since it's used to illustrate an example:

```suggestion

// ignore_for_file: avoid_print

import 'package:cross_file/cross_file.dart';
```

packages/cross_file/example/lib/readme_excerpts.dart
```diff
@@ -0,0 +1,22 @@
...
+  // #docregion Instantiate
+  final XFile file = XFile('assets/hello.txt');
+
+  print('File information:');
+  print('- Path: ${file.path}');
+  print('- Name: ${file.name}');
+  print('- MIME type: ${file.mimeType}');
+
+  final String fileContent = await file.readAsString();
+  print('Content of the file: ${fileContent}'); // e.g. "Moto G (4)"
```
ditman, line 18: Braces not needed in these unambiguous variable names:

```suggestion
  print('Content of the file: $fileContent'); // e.g. "Moto G (4)"
```

packages/cross_file/example/pubspec.yaml
```diff
@@ -0,0 +1,25 @@
...
+    path: ../
+  flutter:
+    sdk: flutter
+
+dev_dependencies:
+  async: ^2.5.0
+  flutter_test:
+    sdk: flutter
+  integration_test:
+    sdk: flutter
```
ditman, line 25: Adds empty line at the end of the file:

```suggestion
    sdk: flutter

```

packages/cross_file/example/test/readme_excerpts_test.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:cross_file/cross_file.dart';
+import 'package:cross_file_example/readme_excerpts.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
```
ditman, line 9: Instructs `flutter test` to only run this in "vm" mode, not in the browser:

```suggestion
import 'package:flutter_test/flutter_test.dart';

@TestOn('vm')

void main() {
```

(Not sure if `@TestOn` can be used from `package:flutter_test` or it's a `package:test` thing, please, adjust the dependency as needed!)

ditman (dismissed):
Hey @mike-v2, thanks for tackling this!

A couple of quick comments:

`flutter analyze` is complaining about a couple of things, here's the excerpt from the error logs:

```
   info - example/lib/readme_excerpts.dart:12:3 - Don't invoke 'print' in production code. Try using a logging framework. - avoid_print
   info - example/lib/readme_excerpts.dart:13:3 - Don't invoke 'print' in production code. Try using a logging framework. - avoid_print
   info - example/lib/readme_excerpts.dart:14:3 - Don't invoke 'print' in production code. Try using a logging framework. - avoid_print
   info - example/lib/readme_excerpts.dart:15:3 - Don't invoke 'print' in production code. Try using a logging framework. - avoid_print
   info - example/lib/readme_excerpts.dart:18:3 - Don't invoke 'print' in production code. Try using a logging framework. - avoid_print
   info - example/lib/readme_excerpts.dart:18:31 - Unnecessary braces in a string interpolation. Try removing the braces. - unnecessary_brace_in_string_interps
```

Also, there's a failure on the "web" test runner; the problem is that on the web `path` doesn't mean anything, so a (real) XFile on the web can only be created from Bytes.

I've suggested a couple of fixes to `readme_excerpts.dart` and its test that might allow this to pass (but these are "off the top of my head" suggestions; please do test locally if they work as intended :P)

mike-v2:
Hey sorry for the easy mistakes and thanks for the detailed reply, it really helps a noob like me! 

- I was having trouble with using 'TestOn' (it never seemed to work, even when I imported from package:test, and my IDE was telling me that it can only be used on libraries), so I used 'kIsWeb' from 'flutter/foundation'. I hope this is ok.
- Also, I wasn't able to get my test to pass unless I copied the assets/hello.txt file to the cross_file/example folder. This doesn't seem like the most elegant solution but I couldn't figure out how to make my test pass without it.

ditman:
(Force-pushed to rebase with latest `main`)

ditman:
I think my `TestOn` change made CI very unhappy with "no tests running"

```
Running command: "dart run test --platform=chrome" in /b/s/w/ir/x/w/packages/packages/cross_file/example
No tests ran.
No tests were found.

[packages/cross_file/example completed in 0m 12s]
```

I guess I'll revert my change and go back to `skip: kIsWeb` so some test "runs" :)

mike-v2:
Oh that's what was causing the 5 failing checks? Interesting. Waiting for all those tests to go green is like watching the longest spin of the Price is Right wheel.
Anyway I definitely learned a lot, and I appreciate your help. I'll try to do some more of these when I get the time!

ditman:
@mike-v2 I think that was causing only the "web" tests to fail. The others were most likely flaky tests that had nothing to do with your change, in fact :)

mike-v2:
Yeah I have access to the failure logs, which is awesome because it allows me to fix many of those simple mistakes. That's why I had a few commits in a row awhile back - I kept watching it fail and then looking through the logs to find what went wrong. Definitely learned a lot by doing that!

ditman:
I think this should be ready to land now. Adding @stuartmorgan as 2nd team reviewer, for when he has a while.

In this PR, we're attempting to test the code of the file that is used to generate the excerpts, I think this is a 1st!

ditman (approved):
LGTM!

(It might be a little odd testing the file that is used to generate the excerpts, but it looks good to me! The weird part is not using `integration_test` in the `example`, but since `cross_file` is a dart package, I didn't want to add any flutter deps, and the CI seems to Just Work™)

stuartmorgan (approved):
LGTM with some nits. Thanks!

> Stuart, in this PR, we're attempting to test the code of the file that is used to generate the excerpts, I think this is a 1st!

Not quite a first, but definitely something we aren't doing very much of yet. It's great to see more of it!

ditman:
This landed and got published, thanks @mike-v2 for the contribution!

PR #5408, opened 2023-11-15
Fix Google Maps rendering issues in TLHC mode when using LATEST renderer

johnmccutchan:
The Google Maps LATEST renderer always uses a TextureView. We can use a signal from this TextureView to perform view invalidation that fixes the rendering glitches (missing updates) in TLHC mode.

NOTE: We have an internal bug 311013682 requesting an official way of achieving this functionality but if the bug is ever acted on it will take many months/years before we can rely on this functionality.

In the meantime, chain the internal SurfaceTextureListener with our own and piggyback on the OnSurfaceTextureUpdated callback to invalidate the MapView.

Fixes https://github.com/flutter/flutter/issues/103686

Tested on an emulator and a physical device (Pixel 6 Pro).

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
@@ -1,5 +1,7 @@
-## NEXT
+## 2.6.0
 
+* Fixes missing updates in TLHC mode.
+* Switched default display mode to TLHC mode.
 * Updates minimum supported SDK version to Flutter 3.10/Dart 3.0.
 
 ## 2.5.3
--- a/packages/google_maps_flutter/google_maps_flutter_android/README.md
+++ b/packages/google_maps_flutter/google_maps_flutter_android/README.md
@@ -30,26 +30,26 @@ void main() {
   final GoogleMapsFlutterPlatform mapsImplementation =
       GoogleMapsFlutterPlatform.instance;
   if (mapsImplementation is GoogleMapsFlutterAndroid) {
+    // Force Hybrid Composition mode.
     mapsImplementation.useAndroidViewSurface = true;
   }
   // ···
 }
 ```
 
-### Hybrid Composition
-
-This is the current default mode, and corresponds to
-`useAndroidViewSurface = true`. It ensures that the map display will work as
-expected, at the cost of some performance.
-
 ### Texture Layer Hybrid Composition
 
-This is a new display mode used by most plugins starting with Flutter 3.0, and
-corresponds to `useAndroidViewSurface = false`. This is more performant than
-Hybrid Composition, but currently [misses certain map updates][4].
+This is the the current default mode and corresponds to `useAndroidViewSurface = false`.
+This mode is more performant than Hybrid Composition and we recommend that you use this mode.
 
-This mode will likely become the default in future versions if/when the
-missed updates issue can be resolved.
+### Hybrid Composition
+
+This mode is available for backwards compatability and corresponds to `useAndroidViewSurface = true`.
+We do not recommend its use as it is less performant than Texture Layer Hybrid Composition and
+certain flutter rendering effects are not supported. 
+
+If you require this mode for correctness, please file a bug so we can investigate and fix
+the issue in the TLHC mode.
 
 ## Map renderer
 
@@ -70,8 +70,13 @@ AndroidMapRenderer mapRenderer = AndroidMapRenderer.platformDefault;
   }
 ```
 
-Available values are `AndroidMapRenderer.latest`, `AndroidMapRenderer.legacy`, `AndroidMapRenderer.platformDefault`.
-Note that getting the requested renderer as a response is not guaranteed.
+`AndroidMapRenderer.platformDefault` corresponds to `AndroidMapRenderer.latest`.
+
+You are not guaranteed to get the requested renderer. For example, on emulators without
+Google Play the latest renderer will not be available and the legacy renderer will always be used.
+
+WARNING: `AndroidMapRenderer.legacy` is known to crash apps and is no longer supported by the Google Maps team
+and therefore cannot be supported by the Flutter team.
 
 [1]: https://pub.dev/packages/google_maps_flutter
 [2]: https://flutter.dev/docs/development/packages-and-plugins/developing-packages#endorsed-federated-plugin
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
@@ -10,10 +10,13 @@ import android.content.Context;
 import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
 import android.graphics.Point;
+import android.graphics.SurfaceTexture;
 import android.os.Bundle;
 import android.util.Log;
-import android.view.Choreographer;
+import android.view.TextureView;
+import android.view.TextureView.SurfaceTextureListener;
 import android.view.View;
+import android.view.ViewGroup;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
@@ -135,61 +138,13 @@ final class GoogleMapController
     return trackCameraPosition ? googleMap.getCameraPosition() : null;
   }
 
-  private boolean loadedCallbackPending = false;
-
-  /**
-   * Invalidates the map view after the map has finished rendering.
-   *
-   * <p>gmscore GL renderer uses a {@link android.view.TextureView}. Android platform views that are
-   * displayed as a texture after Flutter v3.0.0. require that the view hierarchy is notified after
-   * all drawing operations have been flushed.
-   *
-   * <p>Since the GL renderer doesn't use standard Android views, and instead uses GL directly, we
-   * notify the view hierarchy by invalidating the view.
-   *
-   * <p>Unfortunately, when {@link GoogleMap.OnMapLoadedCallback} is fired, the texture may not have
-   * been updated yet.
-   *
-   * <p>To workaround this limitation, wait two frames. This ensures that at least the frame budget
-   * (16.66ms at 60hz) have passed since the drawing operation was issued.
-   */
-  private void invalidateMapIfNeeded() {
-    if (googleMap == null || loadedCallbackPending) {
-      return;
-    }
-    loadedCallbackPending = true;
-    googleMap.setOnMapLoadedCallback(
-        () -> {
-          loadedCallbackPending = false;
-          postFrameCallback(
-              () -> {
-                postFrameCallback(
-                    () -> {
-                      if (mapView != null) {
-                        mapView.invalidate();
-                      }
-                    });
-              });
-        });
-  }
-
-  private static void postFrameCallback(Runnable f) {
-    Choreographer.getInstance()
-        .postFrameCallback(
-            new Choreographer.FrameCallback() {
-              @Override
-              public void doFrame(long frameTimeNanos) {
-                f.run();
-              }
-            });
-  }
-
   @Override
   public void onMapReady(GoogleMap googleMap) {
     this.googleMap = googleMap;
     this.googleMap.setIndoorEnabled(this.indoorEnabled);
     this.googleMap.setTrafficEnabled(this.trafficEnabled);
     this.googleMap.setBuildingsEnabled(this.buildingsEnabled);
+    installInvalidator();
     googleMap.setOnInfoWindowClickListener(this);
     if (mapReadyResult != null) {
       mapReadyResult.success(null);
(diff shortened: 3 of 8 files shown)


stuartmorgan (changes requested):
We should update https://pub.dev/packages/google_maps_flutter_android#texture-layer-hybrid-composition with new guidance. Although if possible, we should also switch the default if this resolves the issue (and leave the switch only as an escape hatch if there are surprise issues). I *think* this would address all cases:
- LATEST would always get this codepath (and work)
- LEGACY would presumably either get this codepath (and work) or hit the SurfaceView codepath in Maps and then automatically fall out of TLHC anyway due to the engine fallback logic (so work).

Does that sound right?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -216,6 +221,67 @@ public void onMapReady(GoogleMap googleMap) {
...
+    return null;
+  }
+
+  private void installInvalidator() {
+    if (mapView == null) {
+      return;
+    }
+    TextureView textureView = findTextureView(mapView);
+    if (textureView == null) {
+      Log.i(TAG, "No TextureView found. Likely using the LEGACY renderer.");
```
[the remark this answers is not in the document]
johnmccutchan: There is no guarantee but the way the code is written it seems very unlikely that they would switch to a SurfaceView. I'm also working with the Maps folks on a way to explicitly request a TextureView is used.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -190,6 +194,7 @@ public void onMapReady(GoogleMap googleMap) {
     this.googleMap.setIndoorEnabled(this.indoorEnabled);
     this.googleMap.setTrafficEnabled(this.trafficEnabled);
     this.googleMap.setBuildingsEnabled(this.buildingsEnabled);
+    installInvalidator();
```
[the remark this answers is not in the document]
johnmccutchan: Does invalidateMapIfNeeded impact the legacy renderer at all? The legacy renderer seems to always use a SurfaceView which ends up in a VD/HC mode.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -216,6 +221,67 @@ public void onMapReady(GoogleMap googleMap) {
...
+        if (r != null) {
+          return r;
+        }
+      }
+    }
+    return null;
+  }
+
+  private void installInvalidator() {
+    if (mapView == null) {
```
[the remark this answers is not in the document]
johnmccutchan: It seems like it can happen in tests and there are already a number of similar checks in the file.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -216,6 +221,67 @@ public void onMapReady(GoogleMap googleMap) {
...
+    TextureView textureView = findTextureView(mapView);
+    if (textureView == null) {
+      Log.i(TAG, "No TextureView found. Likely using the LEGACY renderer.");
+      return;
+    }
+    Log.i(TAG, "Installing custom TextureView driven invalidator.");
+    SurfaceTextureListener internalListener = textureView.getSurfaceTextureListener();
+    // Override the Maps internal SurfaceTextureListener with our own. Our listener
+    // mostly just invokes the internal listener callbacks but in onSurfaceTextureUpdated
+    // we invalidate the mapView which ensures we don't miss display updates.
```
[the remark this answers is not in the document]
johnmccutchan: Done.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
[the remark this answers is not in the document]
johnmccutchan: Internal bug for explicitly requesting a TextureView 311393634

johnmccutchan:
> We should update https://pub.dev/packages/google_maps_flutter_android#texture-layer-hybrid-composition with new guidance. Although if possible, we should also switch the default if this resolves the issue (and leave the switch only as an escape hatch if there are surprise issues). I _think_ this would address all cases:
> 
> * LATEST would always get this codepath (and work)
> * LEGACY would presumably either get this codepath (and work) or hit the SurfaceView codepath in Maps and then automatically fall out of TLHC anyway due to the engine fallback logic (so work).
> 
> Does that sound right?

I've updated the docs, PTAL.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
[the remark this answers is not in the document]
stuartmorgan: It was added specifically for the TLHC problem, so it shouldn't be needed for fallback.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.5.4
+
+* Fix missing updates in TLHC mode.
```
stuartmorgan, line 3: Nit: Fixes

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+* Switch default display mode to TLHC mode.
```
stuartmorgan, line 4: Nit: Switches

packages/google_maps_flutter/google_maps_flutter_android/README.md
```diff
@@ -70,8 +70,14 @@ AndroidMapRenderer mapRenderer = AndroidMapRenderer.platformDefault;
   }
 ```
 
-Available values are `AndroidMapRenderer.latest`, `AndroidMapRenderer.legacy`, `AndroidMapRenderer.platformDefault`.
-Note that getting the requested renderer as a response is not guaranteed.
+NOTE: `AndroidMapRenderer.platformDefault` corresponds to `AndroidMapRenderer.latest`.
+
+WARNING: `AndroidMapRenderer.legacy` is known to crash apps and is no longer supported.
```
stuartmorgan, line 75: Maybe "no longer supported in the Google Maps SDK"? Or do we mean specifically that we won't investigate issues related to it?

packages/google_maps_flutter/google_maps_flutter_android/README.md
```diff
@@ -70,8 +70,14 @@ AndroidMapRenderer mapRenderer = AndroidMapRenderer.platformDefault;
   }
 ```
 
-Available values are `AndroidMapRenderer.latest`, `AndroidMapRenderer.legacy`, `AndroidMapRenderer.platformDefault`.
-Note that getting the requested renderer as a response is not guaranteed.
+NOTE: `AndroidMapRenderer.platformDefault` corresponds to `AndroidMapRenderer.latest`.
```
stuartmorgan, line 73: Doesn't this depend on the device and server behavior? I wouldn't think we could say this at the plugin level.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -504,7 +504,7 @@ class GoogleMapsFlutterAndroid extends GoogleMapsFlutterPlatform {
   /// for more information.
   ///
   /// Currently defaults to true, but the default is subject to change.
-  bool useAndroidViewSurface = true;
+  bool useAndroidViewSurface = false;
```
stuartmorgan, line 508: It doesn't need to be part of this PR, but we should consider doing a follow-up that switches from using `AndroidView` to calling `initSurfaceAndroidView` in the `false` codepath, so that fallback is to HC rather than VD. IIRC there are significant accessibility problems with maps under VD, so HC is probably the better experience as a fallback (and it would be consistent with how the plugin has worked for the last year or so, so it's well-tested).

packages/google_maps_flutter/google_maps_flutter_android/README.md
johnmccutchan: Google Maps won't fix any bugs in it and is working towards completely removing it. We should not spend time on it either.

packages/google_maps_flutter/google_maps_flutter_android/README.md
johnmccutchan: No, the client library (that we dep in via gradle) controls this logic and default is now interpreted as latest.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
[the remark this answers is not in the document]
johnmccutchan: Removed.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.5.4
```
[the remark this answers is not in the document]
johnmccutchan: Done.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
johnmccutchan: Done.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
johnmccutchan: Done.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -504,7 +504,7 @@ class GoogleMapsFlutterAndroid extends GoogleMapsFlutterPlatform {
   /// for more information.
   ///
   /// Currently defaults to true, but the default is subject to change.
```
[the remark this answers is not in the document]
johnmccutchan: Done.

packages/google_maps_flutter/google_maps_flutter_android/README.md
```diff
@@ -30,26 +30,26 @@ void main() {
...
 ```
 
-### Hybrid Composition
+### Texture Layer Hybrid Composition
 
-This is the current default mode, and corresponds to
-`useAndroidViewSurface = true`. It ensures that the map display will work as
-expected, at the cost of some performance.
+This is the the current default mode and corresponds to `useAndroidViewSurface = false`.
+This mode is more performant than Hybrid Composition and we recommend that you use this mode.
```
reidbaker, line 43: If there are notable exceptions or cases where someone should not use TLHC can you include them here.

packages/google_maps_flutter/google_maps_flutter_android/README.md
```diff
@@ -30,26 +30,26 @@ void main() {
...
-This is a new display mode used by most plugins starting with Flutter 3.0, and
-corresponds to `useAndroidViewSurface = false`. This is more performant than
-Hybrid Composition, but currently [misses certain map updates][4].
+This mode is available for backwards compatability and corresponds to `useAndroidViewSurface = true`.
+We do not recommend its use as it is less performant than Texture Layer Hybrid Composition and
+certain flutter rendering effects are not supported. 
 
-This mode will likely become the default in future versions if/when the
-missed updates issue can be resolved.
+If you require this mode for correctness, please file a bug so we can investigate and fix
```
reidbaker, line 51: Do you want them to add a label if it is an TLHC bug?

packages/google_maps_flutter/google_maps_flutter_android/README.md
reidbaker, line 73: nit: Note is specfifically called out as an anti pattern https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-empty-prose

packages/google_maps_flutter/google_maps_flutter_android/README.md
reidbaker, line 75: it might be valuable to say who is not supporting this mode. Specifically the google maps team is no longer supporting AndroidMapRenderer.legacy and therefor the flutter team will not investigate issues either.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -216,6 +171,70 @@ public void onMapReady(GoogleMap googleMap) {
     }
   }
 
+  private static TextureView findTextureView(ViewGroup group) {
```
reidbaker, line 174: It might be valuable to document that this method returns the first texture view it finds.

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_android/README.md
johnmccutchan: There are none that I'm aware of.

packages/google_maps_flutter/google_maps_flutter_android/README.md
johnmccutchan: I'd rather just let our triage process route the bugs correctly than trying to document our current process here.

packages/google_maps_flutter/google_maps_flutter_android/README.md
johnmccutchan: Done.

packages/google_maps_flutter/google_maps_flutter_android/README.md
johnmccutchan: Added more context here.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
johnmccutchan: Done.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
-## NEXT
+## 2.6.0
 
+* Fixes missing updates in TLHC mode.
+* Switched default display mode to TLHC mode.
```
stuartmorgan, line 4: Nit: Switches

(Per our style guide for changelogs.)

stuartmorgan (approved):
LGTM with nit

PR #5416, opened 2023-11-17
Update release step to 3.16

stuartmorgan:
The last commit failed to publish because it requires 3.16, and `release` is still using 3.13.


stuartmorgan:
I'm fixing the tree by publishing manually, but this should land before https://github.com/flutter/packages/pull/5407 so it doesn't happen again.

kevmoo (approved):

PR #5430, opened 2023-11-20
contribute a command to reconcile the analysis_options.yaml file with standardized lint sets

devoncarew:
- contribute a command to reconcile the analysis_options.yaml file with standardized lint sets
- update the `analysis_options.yaml` file as per the output of the `update-analysis-options` command
- the base set of lints is read transitively from the `package:dart_flutter_team_lints` dep on the script/tool package

Starting as a draft as this does not have tests, and, doesn't update the code in the repo wrt the changed set of lints.

cc @stuartmorgan 

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,229 +1,246 @@
 # Specify analysis options.
 #
-# This file is a copy of analysis_options.yaml from flutter repo
-# as of 2022-07-27, but with some modifications marked with
-# "DIFFERENT FROM FLUTTER/FLUTTER" below. The file is expected to
-# be kept in sync with the master file from the flutter repo.
+# This file is a copy of analysis_options.yaml from flutter repo as of
+# 2022-07-27, but with some modifications marked with "DIFFERENT FROM
+# FLUTTER/FLUTTER" below. The file is expected to be kept in sync with the
+# main file from the flutter repo.
 
 analyzer:
   language:
     strict-casts: true
     strict-raw-types: true
   errors:
-    # allow self-reference to deprecated members (we do this because otherwise we have
-    # to annotate every member in every test, assert, etc, when we deprecate something)
+    # Allow self-reference to deprecated members (we do this because otherwise
+    # we have to annotate every member in every test, assert, etc, when we
+    # deprecate something).
     deprecated_member_use_from_same_package: ignore
-  exclude: # DIFFERENT FROM FLUTTER/FLUTTER
-    # Ignore generated files
+  exclude:
+    # DIFFERENT FROM FLUTTER/FLUTTER: Ignore generated files
     - '**/*.g.dart'
     - '**/*.mocks.dart' # Mockito @GenerateMocks
 
 linter:
   rules:
-    # This list is derived from the list of all available lints located at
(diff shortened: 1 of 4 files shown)


devoncarew:
cc @goderbauer as well as this technique would lend itself to having a customized, documented set of lints (while still having them based on the core / recommended / dart_flutter_team_lints sets).

analysis_options.yaml
```diff
@@ -1,229 +1,246 @@
 # Specify analysis options.
 #
-# This file is a copy of analysis_options.yaml from flutter repo
-# as of 2022-07-27, but with some modifications marked with
-# "DIFFERENT FROM FLUTTER/FLUTTER" below. The file is expected to
-# be kept in sync with the master file from the flutter repo.
+# This file is a copy of analysis_options.yaml from flutter repo as of
+# 2022-07-27, but with some modifications marked with "DIFFERENT FROM
+# FLUTTER/FLUTTER" below. The file is expected to be kept in sync with the
+# main file from the flutter repo.
```
goderbauer, line 6: Will this make it a lot harder to sync over the lint settings from flutter/flutter next time we want to do that? Or does the tooling help with that?

analysis_options.yaml
```diff
@@ -1,229 +1,246 @@
...
+    - unawaited_futures
+    # - unnecessary_lambdas # has false positives: https://github.com/dart-lang/linter/issues/498
+    - unnecessary_library_directive
+    - unnecessary_parenthesis
+    - unnecessary_statements
+    - unreachable_from_main
+    - use_is_even_rather_than_modulo
+
+    # Additional customizations (full list at
+    # https://github.com/dart-lang/linter/blob/master/example/all.yaml)
```
goderbauer, line 144: Does the tooling help with keeping this list updated with `all.yaml`? Otherwise, this seems to become pretty painful now? Previously it was just copy over that file, revert the lines that previously had comments, and comment out anything new.

goderbauer (commented):
Overall, I am not sure that this is solving any problems we currently have? But I am also not super familiar with the pain points around this in the flutter/packages repository. 

I'd much rather see work towards making it easier to keep the lint setup in flutter/flutter, flutter/packages, and flutter/engine consistent. Previously, we had talked about maybe defining a flutter_team_lints package that defines the set and is used by all three (it could be based on dart_flutter_team_lints). However, we need to do this in a way that doesn't inhibit trying out new lints and enabling them in the flutter repository.

stuartmorgan:
> I'd much rather see work towards making it easier to keep the lint setup in flutter/flutter, flutter/packages, and flutter/engine consistent. Previously, we had talked about maybe defining a flutter_team_lints package that defines the set and is used by all three (it could be based on dart_flutter_team_lints).

The issue this approach is trying to solve is that actually doing an import of a shared set is painful in a monorepo. There are currently 209 packages in this repo (counting examples and such), and if the analysis options used by all of them have an import, then *every single one of them* needs to have a dev dependency on that package. And if we ever change the import, we have to change 209 packages.

devoncarew:
Sorry - I lost track of this PR. I'm going to close for now (thought my circle back at some point in the future).

PR #4140, opened 2023-06-05
[webview_flutter] Support for handling basic authentication requests

andreisas06:
### Description
This pull request introduces the handling of HTTP Basic Authentication in our WebView implementation. Specifically, it includes the following changes:

1. Added setHttpAuthCredentials() function: This function allows developers to programmatically set credentials (username and password) for a given host and realm.
2. Implemented onReceivedHttpAuthRequest() callback: This callback is invoked when the WebView encounters a website that requires HTTP Basic Authentication. The host and realm information from the authentication request is utilized to match and use the stored credentials.

### Issues fixed by PR:
Closes flutter/flutter#83556

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #5455, opened 2023-11-21
[webview_flutter] Support for handling basic authentication requests (iOS)

JeroenWeener:
Adds the iOS implementation for basic http authentication.

This PR is part of a series of PRs that aim to close https://github.com/flutter/flutter/issues/83556.
The PR that contains all changes can be found at https://github.com/flutter/packages/pull/4140.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/webview_flutter/webview_flutter_wkwebview/example/pubspec.yaml
```diff
@@ -32,3 +32,8 @@ flutter:
     - assets/sample_video.mp4
     - assets/www/index.html
     - assets/www/styles/style.css
+
+# FOR TESTING AND INITIAL REVIEW ONLY. DO NOT MERGE.
+# See https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins
+dependency_overrides:
+   {webview_flutter_platform_interface: {path: ../../../webview_flutter/webview_flutter_platform_interface}}
```
bparrishMines, line 39: Remove this and update the minimum `webview_flutter_platform_interface` to `^2.7.0`

```suggestion
```

packages/webview_flutter/webview_flutter_wkwebview/pubspec.yaml
```diff
@@ -33,3 +33,8 @@ topics:
   - html
   - webview
   - webview-flutter
+
+# FOR TESTING AND INITIAL REVIEW ONLY. DO NOT MERGE.
+# See https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins
+dependency_overrides:
+   {webview_flutter_platform_interface: {path: ../../webview_flutter/webview_flutter_platform_interface}}
```
bparrishMines, line 40: Same for this

```suggestion
```

bparrishMines (changes requested):
Everything looks good. Just need to update the pubspecs

bparrishMines (approved):
LGTM

@stuartmorgan for secondary review

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFNavigationDelegateHostApi.m
```diff
@@ -179,6 +220,40 @@ - (void)webViewWebContentProcessDidTerminate:(WKWebView *)webView {
...
+    didReceiveAuthenticationChallenge:(NSURLAuthenticationChallenge *)challenge
+                    completionHandler:(void (^)(NSURLSessionAuthChallengeDisposition,
+                                                NSURLCredential *_Nullable))completionHandler {
+  [self.navigationDelegateAPI
+      didReceiveAuthenticationChallengeForDelegate:self
+                                           webView:webView
+                                         challenge:challenge
+                                        completion:^(FWFAuthenticationChallengeResponse *response,
+                                                     FlutterError *error) {
+                                          NSAssert(!error, @"%@", error);
```
stuartmorgan, line 234: Why is this an assert? Couldn't a plugin client have an implementation on the Dart side that throws an exception, leading to a `FlutterError` here?

bparrishMines: This is a pattern I started with all the FlutterApi methods. IIRC when there is an exception thrown on the Dart side during a FlutterApi method, this error is returned and it wouldn't be printed anywhere. Essentially failing silently. This would catch the error and print it to the console when the app is not in release mode or disabling asserts.

At the very least, this used to be the case on Android.

stuartmorgan: Right, but in most cases I thought it was just calling our own code, so we could reasonably assert that an error didn't exist.

I guess the question here is: if a plugin *client* throws an exception in their own Dart code that implements the handler, do we want the entire application to crash?

stuartmorgan: @bparrishMines Should we adjust this, or land as-is and revisit later if necessary?

bparrishMines: I created https://github.com/flutter/flutter/issues/140410 to track the discussion. I think we could probably find a better alternative, but I haven't seen any issues raised by users.

PR #5233, opened 2023-10-25
[pointer_interceptor] Add iOS implementation and platform interface

LouiseHsu:
Addresses https://github.com/flutter/flutter/issues/30143 by adding an iOS implementation 

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #5499, opened 2023-11-28
[pointer_interceptor] Add platform interface

LouiseHsu:
Addresses https://github.com/flutter/flutter/issues/30143 by adding an iOS implementation 

This PR is Part 1 of https://github.com/flutter/packages/pull/5233

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/pointer_interceptor/pointer_interceptor_platform_interface/AUTHORS
+++ b/packages/pointer_interceptor/pointer_interceptor_platform_interface/AUTHORS
@@ -1,0 +1,6 @@
+# Below is a list of people and organizations that have contributed
+# to the project. Names should be added to the list like so:
+#
+#   Name/Organization <email address>
+
+Google Inc.
--- a/packages/pointer_interceptor/pointer_interceptor_platform_interface/lib/src/pointer_interceptor_platform.dart
+++ b/packages/pointer_interceptor/pointer_interceptor_platform_interface/lib/src/pointer_interceptor_platform.dart
@@ -1,0 +1,45 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/widgets.dart';
+import 'package:plugin_platform_interface/plugin_platform_interface.dart';
+
+import 'default_pointer_interceptor.dart';
+
+/// Platform-specific implementations should set this with their own
+/// platform-specific class that extends [PointerInterceptorPlatform] when
+/// they register themselves.
+abstract class PointerInterceptorPlatform extends PlatformInterface {
+  /// Constructs a PointerInterceptorPlatform.
+  PointerInterceptorPlatform() : super(token: _token);
+
+  static final Object _token = Object();
+
+  static PointerInterceptorPlatform _instance = DefaultPointerInterceptor();
+
+  static set instance(PointerInterceptorPlatform? instance) {
+    if (instance == null) {
+      throw AssertionError(
+          'Platform interfaces can only be set to a non-null instance');
+    }
+
+    PlatformInterface.verify(instance, _token);
+    _instance = instance;
+  }
+
+  /// The default instance of [PointerInterceptorPlatform] to use.
+  ///
+  /// Defaults to [DefaultPointerInterceptor], which does not do anything
+  static PointerInterceptorPlatform get instance => _instance;
+
+  /// Platform-specific implementations should override this function their own
+  /// implementation of a pointer interceptor widget.
+  Widget buildWidget({
+    required Widget child,
+    bool debug = false,
+    Key? key,
+  }) {
+    throw UnimplementedError('buildWidget() has not been implemented.');
+  }
+}
--- a/packages/pointer_interceptor/pointer_interceptor_platform_interface/pubspec.yaml
+++ b/packages/pointer_interceptor/pointer_interceptor_platform_interface/pubspec.yaml
@@ -1,7 +1,9 @@
 name: pointer_interceptor_platform_interface
 description: "A common platform interface for the pointer_interceptor plugin."
-version: 1.0.0
-publish_to: none
+repository: https://github.com/flutter/packages/tree/main/packages/pointer_interceptor/pointer_interceptor_platform_interface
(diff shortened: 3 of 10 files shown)


packages/pointer_interceptor/pointer_interceptor_platform_interface/lib/src/pointer_interceptor_platform.dart
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  /// The default instance of [PointerInterceptorPlatform] to use.
+  ///
+  /// Defaults to [DefaultPointerInterceptor], which does not do anything
+  static PointerInterceptorPlatform get instance => _instance;
+
+  /// Platform-specific implementations should override this function their own
+  /// implementation of a pointer interceptor widget.
+  Widget buildWidget({
```
stuartmorgan, line 38: We should have a Dart unit test in this PR that calls this and asserts that it throws UnimplementedError; see [here](https://github.com/flutter/packages/blob/fcb1272be272348f32456e5f3e5346b071f8f7c1/packages/webview_flutter/webview_flutter_platform_interface/test/webview_platform_test.dart#L106-L116) for an example. (This is partly because we actually care that the default is to throw, and partially to make sure we have unit tests set up so future changes can more easily add tests.)

packages/pointer_interceptor/pointer_interceptor_platform_interface/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
+name: pointer_interceptor_platform_interface
+description: "A common platform interface for the pointer_interceptor plugin."
+version: 1.0.0
```
stuartmorgan, line 3: 0.10.0

packages/pointer_interceptor/pointer_interceptor_platform_interface/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
...
+publish_to: none
```
stuartmorgan, line 4: This needs to be removed. That will then trigger more checks in CI; you'll need things like the issues and repository links that we have in all the pubspecs.

packages/pointer_interceptor/pointer_interceptor_platform_interface/lib/src/pointer_interceptor_platform.dart
LouiseHsu: By default it returns it's child because thats what the no-op pointer interceptor does, which is set as the default instance

stuartmorgan: Well, we should test that too in a Dart unit test of the default implementation class. But this code you would test in by making a test subclass with no overrides (e.g., `class UnimplementedPointerInterceptorPlatform extends PointerInterceptorPlatform {}`, and then testing that.

Alternately, since this is a new class, you could make it abstract and not provide an implementation of this method at all. The `Unimplemented` pattern is something we do a lot because any method added after publishing must have a default implementation to avoid breaking compilation of existing code, but you don't have that problem here.

LouiseHsu: Ohhh, my bad, I see what you mean about the default class now!

LouiseHsu:
@stuartmorgan this is ready to review again!

packages/pointer_interceptor/pointer_interceptor_platform_interface/test/pointer_interceptor_platform_test.dart
```diff
@@ -0,0 +1,27 @@
...
+      'Default implementation of PointerInterceptorPlatform should throw unimplemented error',
+      () {
+    final PointerInterceptorPlatform unimplementedPointerInterceptorPlatform =
+        UnimplementedPointerInterceptorPlatform();
+
+    final Container testChild = Container();
+    expect(
+        () => unimplementedPointerInterceptorPlatform.buildWidget(
+            child: testChild),
+        throwsUnimplementedError);
```
ditman, line 22: I'd add another test file: `default_pointer_interceptor_test.dart` that contains pretty much the same test as this one, but expects that the return from `buildWidget` is the same as `child` (which is the pass-through feature we really care about!)

ditman (approved):
LGTM! Happy to see the pointer_interceptor grow :P

packages/pointer_interceptor/pointer_interceptor_platform_interface/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
...
+    sdk: flutter
+  plugin_platform_interface: ^2.1.0
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+
+topics:
+  - widgets
+  - platform views
```
stuartmorgan, line 23: Nit: We have a policy for our federated plugins of using the plugin name (with _ converted to - to satisfy pub requirements) as one of the topics to allow finding all the related packages. So we should have `- pointer-interceptor` in the list as well.

stuartmorgan (approved):
LGTM with one nit

stuartmorgan:
I filed https://github.com/flutter/flutter/issues/139305 to catch this earlier next time, and https://github.com/flutter/packages/pull/5525 is open to re-land.

ditman:
> apparently it's a server-side check so our dry-run publish check didn't fail.

Another +1 for: https://github.com/dart-lang/pub/issues/3537

packages/pointer_interceptor/pointer_interceptor_platform_interface/AUTHORS
```diff
@@ -0,0 +1,6 @@
+# Below is a list of people and organizations that have contributed
+# to the project. Names should be added to the list like so:
+#
+#   Name/Organization <email address>
```
hellohuanlin, line 4: probably wanna put your name there?

stuartmorgan: This is the format example.

The actual entries are below, and Google-employed contributors are covered by "Google".

PR #5528, opened 2023-11-30
[google_maps_flutter] Add support for version 8 of the iOS SDK

stuartmorgan:
- Increases the max allowed SDK version to 8.x
- Adds a new iOS 14 example app, per [the documented test structure/plan](https://github.com/flutter/packages/blob/main/packages/google_maps_flutter/google_maps_flutter_ios/example/README.md)
- Moves recently added XCUITests to the iOS 11 example app, where they should have been (per doc linked above). I missed this when reviewing the addition, and noticed that it was incorrectly in the iOS 13 example while duplicating the iOS 13 example for the iOS 14 version.

Fixes https://github.com/flutter/flutter/issues/138972

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
It's probably easiest to review this as individual commits; the big commit called "Add iOS 14" is just an exact copy of `example/ios13` to `example/ios4` so shouldn't need detailed review.

jmagman (approved):
Cross-listing https://github.com/flutter/flutter/issues/86820 and https://github.com/flutter/packages/pull/3757

stuartmorgan:
I hit some deprecation warning failures in CI that I didn't find locally, and it turns out that it's because locally I only resolved to 8.2 (presumably for cache reasons), but 8.3 added a new API and deprecated the ones it replaced. If we have OOB failures from that happening in the future we could restrict the range in the podspec to only allow minor versions we have tested, rather than just major ones, but AFAIK it's never hit us before so for now I'll err on the side of keeping a lenient range.

arthurmonteiroalvesmelo:
Hey @stuartmorgan 

in file (packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/README.md)

The actual text is:
```
## Versions

This example requires iOS 13, so will select a 7.x GoogleMaps SDK version.
```

But, the right is:
```
## Versions

This example requires iOS 14, so will select a 8.x GoogleMaps SDK version.
```

right?

stuartmorgan:
Yes, that's a copy/paste mistake.

PR #5537, opened 2023-12-01
[gsi_web] Do not initialize CodeClient if scopes are empty.

ditman:
When initializing a `CodeClient`, it is mandatory to pass at least one `scope` (or the JS initialization will crash).

This PR ensures that the `CodeClient` is not created unless `initialScopes.isNotEmpty`, and lets the user know when attempting to use said client that it hasn't been initialized properly.

## Issues

* Fixes: https://github.com/flutter/flutter/issues/139382
* Fixes: https://github.com/flutter/flutter/issues/62474
* Closes: https://github.com/flutter/cocoon/pull/3304 (makes this PR unneeded)

## Tests

* Tested locally with `cocoon`.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_sign_in/google_sign_in_web/CHANGELOG.md
+++ b/packages/google_sign_in/google_sign_in_web/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 0.12.2
+## 0.12.3
+
+* Re-publishes 0.12.2 with a small fix to the auth code initialization.
+
+## 0.12.2 (withdrawn)
 
 * Adds server auth code retrieval to google_sign_in_web.
 * Adds `web_only` library to access web-only methods more easily.
--- a/packages/google_sign_in/google_sign_in_web/CHANGELOG.md
+++ b/packages/google_sign_in/google_sign_in_web/CHANGELOG.md
@@ -1,6 +1,6 @@
-## 0.12.3
+## 0.12.2+1
 
-* Re-publishes 0.12.2 with a small fix to the auth code initialization.
+* Re-publishes `0.12.2` with a small fix to the CodeClient initialization.
 
 ## 0.12.2 (withdrawn)
 
--- a/packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
+++ b/packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
@@ -52,13 +52,15 @@ class GisSdkClient {
       onError: _onTokenError,
     );
 
-    _codeClient = _initializeCodeClient(
-      clientId,
-      hostedDomain: hostedDomain,
-      onResponse: _onCodeResponse,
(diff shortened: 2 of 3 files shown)


packages/google_sign_in/google_sign_in_web/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-## 0.12.2
+## 0.12.3
```
stuartmorgan, line 1: `0.12.2+1`?

stuartmorgan (commented):
I haven't paged the whole flow back in here, but what if someone creates with no scopes, but then adds scopes later. Shouldn't we create the client then? Or does that already happen?

ditman:
> I haven't paged the whole flow back in here, but what if someone creates with no scopes, but then adds scopes later. Shouldn't we create the client then? Or does that already happen?

This seems doable with the "TokenClient"; there's an ["OverridableTokenClientConfig" object](https://developers.google.com/identity/oauth2/web/reference/js-reference#OverridableTokenClientConfig) that lets you redefine scopes later, but for the CodeClient flow there's not such a thing, the [`requestCode` method of the CodeClient](https://developers.google.com/identity/oauth2/web/reference/js-reference#CodeClient) doesn't support options.

(However, there's a new `include_granted_scopes` property in the CodeClient initialization)

I suspect the way the CodeClient is intended to be used is to re-create the client with `include_granted_scopes` in order to add more scopes to what has already been authorized, but that doesn't jive well with our "init" method.

Maybe the plugin should re-create the code-client as needed with options right before calling `getCode`, but for now, users will have to request all `scopes` beforehand.

packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
```diff
@@ -291,7 +293,12 @@ class GisSdkClient {
   /// Requests a server auth code per:
   /// https://developers.google.com/identity/oauth2/web/guides/use-code-model#initialize_a_code_client
   Future<String?> requestServerAuthCode() async {
-    _codeClient.requestCode();
+    assert(_codeClient != null,
+        'CodeClient not initialized correctly. Ensure the `scopes` list passed to `init()` or `initWithParams()` is not empty!');
```
stuartmorgan, line 297: Can we have a TODO with an issue link here to track re-creating `_codeClient` when adding scopes?

stuartmorgan (approved):
LGTM with TODO nit.

ditman:
@stuartmorgan if all the tests pass locally I'm going to land this with the tree on red. I don't think it's going to fix the tree, but... it might kick a lucky build and fix. That way we can continue working on google_sign_in_web without worrying about publishing the bad version (the migration to `package:web`, for example).

PR #5544, opened 2023-12-03
[pigeon] Adds ProxyApi code generation for Dart

bparrishMines:
Part of https://github.com/flutter/flutter/issues/134777

* Splits `Api` into `AstHostApi`/`AstFlutterApi`/`AstProxyApi`. I prefixed it with `Ast` since the annotations already had the names `HostApi`, etc...;
* `Api.location` was moved to `Method.location` since `AstProxyApi` has methods implemented on the host and Flutter side.
* Replaces a few of the methods in `lib/functional.dart` with the `package:collection` equivalent.
* `DartGenerator.writeProxyApi` uses `code_builder` and `dart_style` which I added to the allowed deps config file. 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/lib/ast.dart
```diff
@@ -78,29 +93,185 @@ class Method extends Node {
...
+      );
+
+  @override
+  String toString() {
+    return '(ProxyApi name:$name methods:$methods documentationComments:$documentationComments superClassName:$superClassName interfacesNames:$interfacesNames)';
+  }
+}
+
+/// Represents a constructor for an API.
+class Constructor extends Node {
```
stuartmorgan, line 187: Why does a constructor have to be a special type? It seems like all the properties are standard method properties.

packages/pigeon/lib/dart_generator.dart
```diff
@@ -521,92 +436,1264 @@ final BinaryMessenger? ${_varNamePrefix}binaryMessenger;
...
+
+    const String proxyApiBaseClassName = '${classNamePrefix}ProxyApiBaseClass';
+    final Iterable<String> apiHandlerSetups =
+        root.apis.whereType<AstProxyApi>().map(
+      (AstProxyApi api) {
+        return '${api.name}.${classMemberNamePrefix}setUpMessageHandlers(${classMemberNamePrefix}instanceManager: instanceManager);';
+      },
+    );
+    indent.format('''
+/// Maintains instances used to communicate with the native objects they
```
stuartmorgan, line 468: I think we should probably put giant strings like this into a helper file, wrapped in a method that takes all the necessary variables, but without any logic ideally.

packages/pigeon/lib/dart_generator.dart
```diff
@@ -521,92 +436,1264 @@ final BinaryMessenger? ${_varNamePrefix}binaryMessenger;
...
+                    for (final Method method in flutterMethods)
+                      method.name: cb.refer(method.name),
+                  },
+                )
+                .returned
+                .statement,
+          ]),
+      ),
+    ];
+  }
```
stuartmorgan, line 1652: This is quite a method.

Maybe we could have a separate file for all the proxy-specific code? Then it would be easier to do things like extract helpers for different sections without overwhelming the rest of this file.

packages/pigeon/lib/dart_generator.dart
```diff
@@ -736,6 +1824,249 @@ PlatformException _createConnectionError(String channelName) {
...
+\t\tmessage: 'Host platform returned null value for non-null return value.',
+\t);''');
+    }
+    indent.format('''
+} else {
+\t$returnStatement
+}''');
+  }
+
+  void _writeFlutterMethodMessageHandler(
```
stuartmorgan, line 1935: Some of these incidental refactors could ideally be done in a prequel PR that would be trivial to review.

stuartmorgan, line 1935: Some of these incidental refactors could ideally be done in a prequel PR that would be trivial to review.

packages/pigeon/lib/pigeon_lib.dart
```diff
@@ -1335,6 +1767,88 @@ class _RootBuilder extends dart_ast_visitor.RecursiveAstVisitor<Object?> {
             message: 'Expected a named type but found "$node".',
             lineNumber: _calculateLineNumber(source, node.offset)));
       }
+    } else if (_currentApi is AstProxyApi) {
```
stuartmorgan, line 1770: Can we extract all of this block to a helper?

packages/pigeon/pigeons/core_tests.dart
```diff
@@ -699,3 +699,460 @@ class TestMessage {
   // ignore: always_specify_types, strict_raw_type
   List? testList;
 }
+
+/// The core interface that each host language plugin must implement in
```
stuartmorgan, line 703: This may be a case where we do actually want to make a second file instead of putting everything into the one giant file?

stuartmorgan (commented):
I left some high-level comments; @tarrinneal should be the main reviewer for most of the details.

I didn't review the generated code since I'm assuming it's not substantially changed from the webview_flutter version. If there's anything there you do want me to look at, let me know!

One thing that wasn't immediately obvious to me without tracing all the logic: are the big blocks of generated code conditional on using them? I.e., would this be basically a no-op for people not using proxy APIs, or would they suddenly get thousands of new lines of generated code they didn't need?

packages/pigeon/lib/ast.dart
bparrishMines: I think this is similar to our discussion about `parameter` vs `field`. They overlap but some of the parameters don't apply (e.g. `static`/`isAsynchronous`). However, if I consider the `Method` as the message call and not the AST class member, then I suppose they are essentially the same thing. I still need to distinguish the difference between `Constructors` for the Dart code generation, so I think I will keep the class and have it extend `Method`.

bparrishMines:
I'm going to mark this as a draft and split it up into simpler PRs. I will still keep this PR open since it is still useful to demonstrate all the changes put together.

packages/pigeon/lib/pigeon_lib.dart
bparrishMines: Done in https://github.com/flutter/packages/pull/5861

bparrishMines:
> One thing that wasn't immediately obvious to me without tracing all the logic: are the big blocks of generated code conditional on using them? I.e., would this be basically a no-op for people not using proxy APIs, or would they suddenly get thousands of new lines of generated code they didn't need?

Yes, the generation of the `InstanceManager`/`InstanceManagerApi`/`ProxyApiCode` are all dependent on whether there is a `ProxyApi`. I will include a test for this in the PR that implements the code generation.

bparrishMines:
closing if favor of https://github.com/flutter/packages/pull/6043

PR #5565, opened 2023-12-04
[tools] Validate pubspec topic format

LouiseHsu:
Fixes https://github.com/flutter/flutter/issues/139305
## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/pubspec_check_command.dart
+++ b/script/tool/lib/src/pubspec_check_command.dart
@@ -356,16 +356,18 @@ class PubspecCheckCommand extends PackageLoopingCommand {
             'a topic. Add "$topicName" to the "topics" section.';
       }
     }
-    for (final String topic in topics) {
-      // Validates topic names according to https://dart.dev/tools/pub/pubspec#topics
-      final RegExp expectedTopicFormat =
-          RegExp(r'^[a-z](?:-?[a-z0-9]+)*$');
-      if (!expectedTopicFormat.hasMatch(topic) || topic.length < 2 || topic.length > 32 ) {
-        return 'Invalid topic value "$topic" in "topics" section. '
-            'Topics must consist of lowercase alphanumerical characters or dash (but no double dash), '
-            'start with a-z and ending with a-z or 0-9, have a minimum of 2 characters '
-            'and have a maximum of 32 characters.';
-      }
+    
+    // Validates topic names according to https://dart.dev/tools/pub/pubspec#topics
+    final RegExp expectedTopicFormat = RegExp(r'^[a-z](?:-?[a-z0-9]+)*$');
+    final Iterable<String> invalidTopics = topics.where((String topic) =>
+        !expectedTopicFormat.hasMatch(topic) ||
+        topic.length < 2 ||
+        topic.length > 32);
+    if (invalidTopics.isNotEmpty) {
+      return 'Invalid topic(s): ${invalidTopics.join(', ')} in "topics" section. '
+          'Topics must consist of lowercase alphanumerical characters or dash (but no double dash), '
+          'start with a-z and ending with a-z or 0-9, have a minimum of 2 characters '
+          'and have a maximum of 32 characters.';
     }
     return null;
   }
--- a/script/tool/lib/src/pubspec_check_command.dart
+++ b/script/tool/lib/src/pubspec_check_command.dart
@@ -332,7 +332,7 @@ class PubspecCheckCommand extends PackageLoopingCommand {
         false;
   }
 
-  // Validates the "implements" keyword for a plugin, returning an error
+  // Validates the "topics" keyword for a plugin, returning an error
   // string if there are any issues.
   String? _checkTopics(
     Pubspec pubspec, {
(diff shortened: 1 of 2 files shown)


script/tool/lib/src/pubspec_check_command.dart
```diff
@@ -352,6 +352,15 @@ class PubspecCheckCommand extends PackageLoopingCommand {
             'a topic. Add "$topicName" to the "topics" section.';
       }
     }
+    for (final String topic in topics) {
+      final RegExp expectedTopicFormat =
+          RegExp(r'^[a-z](?:[a-z0-9]*[a-z0-9])?(-[a-z0-9]+)*$');
+      if (!expectedTopicFormat.hasMatch(topic)) {
```
stuartmorgan, line 358: We should also enforce the 32-character max and two-character min.

script/tool/lib/src/pubspec_check_command.dart
```diff
@@ -352,6 +352,15 @@ class PubspecCheckCommand extends PackageLoopingCommand {
             'a topic. Add "$topicName" to the "topics" section.';
       }
     }
+    for (final String topic in topics) {
+      final RegExp expectedTopicFormat =
+          RegExp(r'^[a-z](?:[a-z0-9]*[a-z0-9])?(-[a-z0-9]+)*$');
```
stuartmorgan, line 357: Couldn't we make the `-` optional rather than need the middle group? So `r'[a-z](?:-?[a-z0-9]+)*$'`?

script/tool/lib/src/pubspec_check_command.dart
```diff
@@ -352,6 +352,15 @@ class PubspecCheckCommand extends PackageLoopingCommand {
...
+      if (!expectedTopicFormat.hasMatch(topic)) {
+        return 'Invalid topic value "$topic" in "topics" section. '
+            'Topics must consist of lowercase alphanumerical characters or dash (but no double dash), '
+            'starting with a-z and ending with a-z or 0-9.';
+      }
+    }
```
stuartmorgan, line 363: When possible we want to return all errors at once, to avoid cases where someone fixes one error but doesn't realize there's a second one, so we could do something like:
```dart
final List<String> invalidTopics = topics.where((String topic) => !expectedTopicFormat.hasMatch(topic) || topic.length < 2 || topic.length > 32);
if (invalidTopics.isNotEmpty) {
  return 'Invalid topic(s): ${invalidTopics.join(', ')} in ...
}
```

script/tool/lib/src/pubspec_check_command.dart
```diff
@@ -352,6 +352,15 @@ class PubspecCheckCommand extends PackageLoopingCommand {
             'a topic. Add "$topicName" to the "topics" section.';
       }
     }
+    for (final String topic in topics) {
+      final RegExp expectedTopicFormat =
+          RegExp(r'^[a-z](?:[a-z0-9]*[a-z0-9])?(-[a-z0-9]+)*$');
+      if (!expectedTopicFormat.hasMatch(topic)) {
+        return 'Invalid topic value "$topic" in "topics" section. '
+            'Topics must consist of lowercase alphanumerical characters or dash (but no double dash), '
+            'starting with a-z and ending with a-z or 0-9.';
```
stuartmorgan, line 361: We should mention the length requirements as well, and/or give a link to the pubspec docs page from my comment above.

script/tool/test/pubspec_check_command_test.dart
```diff
@@ -633,6 +633,126 @@ ${_topicsSection()}
...
+      expect(
+        output,
+        containsAllInOrder(<Matcher>[
+          contains('Invalid topic value "plugin-A" in "topics" section. '
+              'Topics must consist of lowercase alphanumerical characters or dash (but no double dash), '
+              'starting with a-z and ending with a-z or 0-9.'),
+        ]),
+      );
+    });
+
```
stuartmorgan, line 755: These tests are great! Let's also add one for a topic ending in "-".

stuartmorgan: (And then new ones for the new length checks, so a 1-char and a 33-char).

script/tool/test/pubspec_check_command_test.dart
```diff
@@ -633,6 +633,126 @@ ${_topicsSection()}
...
+        commandError = e;
+      });
+
+      expect(commandError, isA<ToolExit>());
+      expect(
+        output,
+        containsAllInOrder(<Matcher>[
+          contains('Invalid topic value "plugin a" in "topics" section. '
+              'Topics must consist of lowercase alphanumerical characters or dash (but no double dash), '
+              'starting with a-z and ending with a-z or 0-9.'),
```
stuartmorgan, line 661: You could just have one test for "invalid topic has the expected failure message" with the full text we want to check for, and then the other tests can only check for `contains` the first sentence of the error.

stuartmorgan (commented):
Thanks for the quick follow-up on this!

script/tool/lib/src/pubspec_check_command.dart
LouiseHsu: Oh you're right lol. I may have over thought this a little bit. I think your example with a '^' appended in the beginning is functionally equivalent haha.

LouiseHsu:
@stuartmorgan btw this is ready for review again!

PR #5567, opened 2023-12-04
[tool] Add support for `.java`, `.gradle`, `.sh`, and `.m` files…

mike-v2:
… to `update-excerpts` tool

Need this tool to recognize `.java`, `.gradle`, and `.sh` files in order to complete [flutter/flutter#102679](https://github.com/flutter/flutter/issues/102679). The `README.md` for the `espresso` package contains code from `.java`, `.gradle`, and `.sh` files.

Support for `.m` files was requested by @tarrinneal.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/update_excerpts_command.dart
+++ b/script/tool/lib/src/update_excerpts_command.dart
@@ -142,6 +142,7 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
                   case '.kt':
                     language = 'kotlin';
                     break;
+                  case '.cc':
                   case '.cpp':
                     language = 'c++';
                     break;
--- a/script/tool/lib/src/update_excerpts_command.dart
+++ b/script/tool/lib/src/update_excerpts_command.dart
@@ -142,7 +142,6 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
                   case '.kt':
                     language = 'kotlin';
                     break;
-                  case '.cc':
                   case '.cpp':
                     language = 'c++';
                     break;
--- a/script/tool/lib/src/update_excerpts_command.dart
+++ b/script/tool/lib/src/update_excerpts_command.dart
@@ -146,6 +146,12 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
                   case '.cpp':
                     language = 'c++';
                     break;
+                  case '.m':
+                    language = 'objectivec';
+                    break;
+                  case '.gradle':
+                    language = 'groovy';
+                    break;
                   default:
                     language = extension.substring(1);
                     break;
@@ -239,6 +245,10 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
       case 'js':
       case 'kotlin':
       case 'rfwtxt':
+      case 'java':
+      case 'groovy':
+      case 'objectivec':
+      case 'gradle':
       case 'swift':
         prefix = '// ';
         break;
@@ -253,6 +263,9 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
         padding = ' ';
         break;
       case 'yaml':
+        prefix = '# ';
+        break;
+      case 'sh':
         prefix = '# ';
         break;
     }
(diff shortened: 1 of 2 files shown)


mike-v2:
Ah sorry, I should've figured that out. Thanks for pointing me in the right direction!

script/tool/test/update_excerpts_command_test.dart
```diff
@@ -447,6 +443,74 @@ FAIL
...
+
+<?code-excerpt "$fileName (SomeSection)"?>
+```$language
+A B C
+```
+''', fileName);
+      });
+    }
+
+    // ignore: prefer_foreach
```
mike-v2, line 509: Two conflicting lint rules here: [avoid_function_literals_in_foreach_calls](https://dart.dev/tools/linter-rules/avoid_function_literals_in_foreach_calls) and [prefer_foreach](https://dart.dev/tools/linter-rules/prefer_foreach). The latter aims to make the code more terse but the formatter uses the same number of lines with `forEach` as it does with `for...in`.

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -142,10 +142,15 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
                   case '.kt':
                     language = 'kotlin';
                     break;
-                  case '.cc':
```
stuartmorgan, line 145: Why was this removed?

script/tool/test/update_excerpts_command_test.dart
```diff
@@ -33,12 +33,13 @@ void runAllTests(MockPlatform platform) {
       ));
   });
 
-  Future<void> testInjection(String before, String source, String after,
+  Future<void> testInjection(
+      String before, String source, String after, String fileName,
```
stuartmorgan, line 37: This is a lot of positional arguments with the same type; ideally they should all be named, but lets at least name the new one.

stuartmorgan, line 37: Nit throughout the PR: "filename" is generally a single word in programming, so the variable would just be `filename`.

script/tool/test/update_excerpts_command_test.dart
stuartmorgan: There's not a conflict, the linter just gives a misleading suggestion for what you probably wrote as your other version. The version of this that's compliant with all of our lints is `testCases.forEach(runTest);`.

stuartmorgan (changes requested):

script/tool/lib/src/update_excerpts_command.dart
mike-v2: Shoot, there was a problem with my test case for `.cc` files. My mistake.

stuartmorgan (approved):
LGTM!

script/tool/lib/src/update_excerpts_command.dart
```diff
@@ -146,6 +146,12 @@ class UpdateExcerptsCommand extends PackageLoopingCommand {
                   case '.cpp':
                     language = 'c++';
                     break;
+                  case '.m':
+                    language = 'objectivec';
+                    break;
```
tarrinneal, line 151: @stuartmorgan Seems like we don't have support for header files for either language. I'm not sure there's a simple solution though.

script/tool/test/update_excerpts_command_test.dart
```diff
@@ -33,12 +33,16 @@ void runAllTests(MockPlatform platform) {
...
   });
 
-  Future<void> testInjection(String before, String source, String after,
-      {bool failOnChange = false}) async {
+  Future<void> testInjection(
+      {required String before,
+      required String source,
+      required String after,
+      required String filename,
+      bool failOnChange = false}) async {
```
tarrinneal, line 41: nit: add a trailing comma `false,})`

tarrinneal (changes requested):
Do you think there's a good way to handle .h files for c++ and objc?

tarrinneal (approved):
Changed my mind, merging now

script/tool/lib/src/update_excerpts_command.dart
stuartmorgan: Good point. That would need either more special casing [here](https://github.com/flutter/packages/blob/dde5ebf5a4bad528882837eeb9fab1cf978947d8/script/tool/lib/src/update_excerpts_command.dart#L187-L191), or a more generalized solution (like a mapping from extensions to the set of allowed languages)).

PR #5574, opened 2023-12-05
Migrate away from deprecated `js_util` in web_benchmarks

kenzieschmoll:


stuartmorgan:

packages/web_benchmarks/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: web_benchmarks
 description: A benchmark harness for performance-testing Flutter apps in Chrome.
 repository: https://github.com/flutter/packages/tree/main/packages/web_benchmarks
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+web_benchmarks%22
-version: 0.1.0+9
+version: 0.1.0+10
 
 environment:
   sdk: ">=3.0.0 <4.0.0"
```
stuartmorgan, line 8: You'll need to bump these up to match `stable` so the legacy analysis tests don't run and fail.

yjbanov: cc @mdebbar Do you recall in which stable Flutter release the deprecation happened?

packages/web_benchmarks/lib/src/recorder.dart
```diff
@@ -4,10 +4,9 @@
 
 import 'dart:async';
 import 'dart:html' as html;
```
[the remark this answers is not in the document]
kenzieschmoll: Filed https://github.com/flutter/flutter/issues/139577 to track this work separately.

packages/web_benchmarks/pubspec.yaml
mdebbar: This is the commit: https://github.com/flutter/engine/commit/73671d4ade13e020ef0a86ebc1f961af714658d9. According to the version tags, the earliest stable release that contains that commit was `3.16.0` (I think?).

kenzieschmoll:
Closing in favor of https://github.com/flutter/packages/pull/5577 so these two changes can be released together.

PR #5577, opened 2023-12-05
Fix benchmark reload bug and migrate away from deprecated `js_util` APIs

kenzieschmoll:
This PR:
- Ensures the benchmark client reloads with the proper `initialPage`. Without this change, the benchmark app may not reload properly, which causes a hang on the server.
- Migrates away from `js_util` APIs to `dart:ui_web`. This was causing spurious deprecation warnings.

The changes in this PR are test covered by existing tests.


yjbanov (approved):
lgtm

PR #5581, opened 2023-12-06
[gis_web] Migrate to package:web.

ditman:
* Migrates package to use `package:web` so it can be compiled with WASM.
* **Breaking change**: Addresses API mismatches with the latest JS SDK changes.
* Adds some tests to client configuration objects.

(See CHANGELOG.md for a more thorough list of changes)

## Issues

* Fixes: https://github.com/flutter/flutter/issues/138202
* Fixes: https://github.com/flutter/flutter/issues/139167
  * Part of: https://github.com/flutter/flutter/issues/139170

## Tests

* Ran unit/integration tests locally, added a couple more.
* Tested example apps with `flutter run`, both seem to work.
* Tested example apps with `--wasm`, both seem to work.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_identity_services_web/test/js_loader_tt_forbidden_test.dart
```diff
@@ -5,9 +5,8 @@
 @TestOn('browser') // Uses package:js
```
kevmoo, line 5:
```suggestion
@TestOn('browser') // Uses package:web
```

packages/google_identity_services_web/test/js_loader_tt_custom_test.dart
```diff
@@ -5,9 +5,8 @@
 @TestOn('browser') // Uses package:js
```
kevmoo, line 5:
```suggestion
@TestOn('browser') // Uses package:web
```

packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
```diff
@@ -0,0 +1,35 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Provides some useful tweaks to `package:web`.
+library package_web_tweaks;
```
kevmoo, line 6: CC @sigurdm re our helpers

kevmoo (approved):

packages/google_identity_services_web/test/js_loader_tt_forbidden_test.dart
ditman: Very true!

packages/google_identity_services_web/example/integration_test/js_interop_id_test.dart
```diff
@@ -37,45 +40,46 @@ void main() async {
       final IdConfiguration config = IdConfiguration(
         client_id: 'testing_1-2-3',
         auto_select: false,
-        callback: allowInterop((_) {}),
+        callback: (_) {},
```
srujzs, line 43: Wouldn't this and below come across the issue of the parameter being `dynamic` and therefore not supported by `toJS`? Maybe this is handled elsewhere and if so, ignore me. :)

packages/google_identity_services_web/example/integration_test/js_interop_id_test.dart
```diff
@@ -37,45 +40,46 @@ void main() async {
...
-        expect(js_util.getProperty(config, name), matcher, reason: name);
-      }
+      final utils.ExpectConfigValueFn expectConfigValue =
+          utils.createExpectConfigValue(config as JSObject);
 
-      expectConfigValue('allowed_parent_origin', hasLength(2));
+      expectConfigValue('client_id', 'testing_1-2-3');
       expectConfigValue('auto_select', isFalse);
-      expectConfigValue('callback', isA<Function>());
+      expectConfigValue('callback', isA<JSFunction>());
```
srujzs, line 65: Side note: `is` is dependent on the compiler since it's a runtime check. This is fine because you only use JS types for the `is` check, and dart2js/ddc are more restrictive than dart2wasm there. So if it passes on dart2js/ddc, this is fine, but something to be aware.

packages/google_identity_services_web/example/integration_test/utils.dart
```diff
@@ -2,31 +2,47 @@
...
+/// Use [createExpectConfigValue] to create one of this functions associated with
+/// a specific [JSObject].
+typedef ExpectConfigValueFn = void Function(String name, Object? matcher);
 
-@JS('window')
-external Object get domWindow;
+/// Creates a [ExpectConfigValueFn] for the `config` [JSObject].
+ExpectConfigValueFn createExpectConfigValue(JSObject config) {
+  return (String name, Object? matcher) {
+    expect(config.getProperty(name.toJS), matcher, reason: name);
```
srujzs, line 26: Alternatively: `config[name]` here and elsewhere where `getProperty` is used with a `String` :)

packages/google_identity_services_web/lib/src/js_interop/google_accounts_id.dart
```diff
@@ -85,16 +84,40 @@ extension GoogleAccountsIdExtension on GoogleAccountsId {
...
   /// Method: google.accounts.id.renderButton
   /// https://developers.google.com/identity/gsi/web/reference/js-reference#google.accounts.id.renderButton
-  external void renderButton(
+  void renderButton(
     Object parent, [
-    GsiButtonConfiguration options,
-  ]);
+    GsiButtonConfiguration? options,
+  ]) {
+    assert(parent is JSObject,
```
srujzs, line 107: Might be cleaner to leverage type promotion here e.g.

```dart
if (parent is JSObject) {
  if (options == null) ...
  return ...
} else {
  assert(...);
}
```

packages/google_identity_services_web/lib/src/js_interop/google_accounts_id.dart
```diff
@@ -132,12 +166,19 @@ extension GoogleAccountsIdExtension on GoogleAccountsId {
...
-  external void revoke(String hint, [RevocationResponseHandlerFn callback]);
+  void revoke(String hint, [RevocationResponseHandlerFn? callback]) {
+    if (callback == null) {
+      return _revoke(hint.toJS);
+    }
+    return _revokeWithCallback(hint.toJS, callback.toJS);
+  }
+
+  @JS('revoke')
+  external void _revoke(JSString hint);
```
srujzs, line 179: FYI - you don't need `JSString`, you can use `String` here and elsewhere if you want and we'll do the conversion for you. Of course, if you prefer it or want to minimize round-trip conversions (e.g. some other function returns a `JSString` that you pass here), feel free to keep it as is. Same with other primitives like `bool`.

srujzs (approved):

packages/google_identity_services_web/example/integration_test/js_interop_id_test.dart
ditman: @srujzs, having several of this working but not the other did confuse me. I convinced myself that this works because the parameter is a JSObject that I've defined in the JS-interop layer (but so is "Event"?)

If you can't explain me why this works but the other doesn't, there's a problem! :P

kevmoo: I just replaced `(_) {...` with `(JSAny? _) {...` and it worked

packages/google_identity_services_web/example/integration_test/js_interop_id_test.dart
ditman: Would this pass in dart2wasm? I eventually want to check there too. Is there any way of checking `typeof obj === "function"` that works across all compilers?

packages/google_identity_services_web/example/integration_test/utils.dart
ditman: Oh my, of course I couldn't find any `.get`, never tried to look at the `operator []` 😅

packages/google_identity_services_web/lib/src/js_interop/google_accounts_id.dart
ditman: I find it confusing that some things promoting automatically vs others not, so I ended up migrating everything, ROBOT STYLE, to assume that nothing promotes.

When everything promotes, I'll come back and remove half the code :P

packages/google_identity_services_web/example/integration_test/js_interop_id_test.dart
srujzs: Yes, but I think the the right way is to use `typeofEquals` e.g. `Expect.isTrue(callback.typeofEquals('function'), true);` since you want expectations on the JS value itself and not our representation of it (which may differ).  I have a proposal to add a `isA` member that makes this less annoying to write: https://github.com/dart-lang/sdk/issues/54138.

packages/google_identity_services_web/lib/src/js_interop/google_accounts_id.dart
ditman: I might have to revisit this if people start encountering issues in Prod where `assert`s are removed. I don't want to have an `if` forever there.

However, since `JSObject` comes from the dart SDK, I might be able to change the type of the parent to force people to pass a JSObject (regardless of how it was obtained).

packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
ditman: Thanks @sigmundch for the `has` suggestion! I'm going to try that, rather than having my own nullable getter!

However I'm going to file an issue because I think that `trustedTypes` should be `nullable` on our API. It's only available [in Chromium browsers](https://caniuse.com/mdn-api_trustedtypes), and undefined/null elsewhere.

PR #5592, opened 2023-12-06
[web_benchmarks] migrate to pkg:web

kevmoo:
Drop dart:html
Enables usage via WebAssembly

Fixes https://github.com/flutter/flutter/issues/139577

### DIFF
--- a/packages/web_benchmarks/lib/client.dart
+++ b/packages/web_benchmarks/lib/client.dart
@@ -155,7 +155,9 @@ void _printResultsToScreen(Profile profile) {
   profile.scoreData.forEach((String scoreKey, Timeseries timeseries) {
     body.appendHtml('<h2>$scoreKey</h2>');
     body.appendHtml('<pre>${timeseries.computeStats()}</pre>');
-    body.append(TimeseriesVisualization(timeseries).render() as JSObject);
+    // TODO(kevmoo): remove `NodeClue` cast when we no longer need to support
+    // pkg:web 0.3.0
+    html.NodeGlue(body).append(TimeseriesVisualization(timeseries).render());
   });
 }
 
--- a/packages/web_benchmarks/lib/client.dart
+++ b/packages/web_benchmarks/lib/client.dart
@@ -7,7 +7,7 @@ import 'dart:convert' show json;
 import 'dart:js_interop';
 import 'dart:math' as math;
 
-import 'package:web/helpers.dart' as html;
+import 'package:web/helpers.dart';
 
 import 'src/common.dart';
 import 'src/recorder.dart';
@@ -50,7 +50,7 @@ Future<void> runBenchmarks(
 
   await _runBenchmark(nextBenchmark);
 
-  final Uri currentUri = Uri.parse(html.window.location.href);
+  final Uri currentUri = Uri.parse(window.location.href);
   // Create a new URI with the current 'page' value set to [initialPage] to
   // ensure the benchmark app is reloaded at the proper location.
   final Uri newUri = Uri(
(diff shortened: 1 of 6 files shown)


packages/web_benchmarks/lib/client.dart
```diff
@@ -4,11 +4,14 @@
 
 import 'dart:async';
 import 'dart:convert' show json;
-import 'dart:html' as html;
+import 'dart:js_interop';
 import 'dart:math' as math;
 
+import 'package:web/helpers.dart' as html;
```
kenzieschmoll, line 10: nit can we drop to html prefix? seems misleading since this isn't dart:html

packages/web_benchmarks/lib/client.dart
```diff
@@ -127,33 +130,32 @@ void _fallbackToManual(String error) {
         ${_benchmarks.keys.map((String name) => '<li><button id="$name">$name</button></li>').join('\n')}
       </ul>
     </div>
-  ''',
-      validator: html.NodeValidatorBuilder()
-        ..allowHtml5()
-        ..allowInlineStyles());
+  ''');
```
kenzieschmoll, line 134: is this safe to remove?

packages/web_benchmarks/lib/client.dart
kevmoo: done!

packages/web_benchmarks/lib/client.dart
kevmoo: Given that we generate 100% of the HTML embedded, I believe yes!

Hixie:

packages/web_benchmarks/lib/client.dart
```diff
@@ -127,33 +130,33 @@ void _fallbackToManual(String error) {
...
+  final HTMLBodyElement body = document.body! as HTMLBodyElement;
 
-  body.innerHtml = '<h2>${profile.name}</h2>';
+  body.innerHTML = '<h2>${profile.name}</h2>';
 
   profile.scoreData.forEach((String scoreKey, Timeseries timeseries) {
     body.appendHtml('<h2>$scoreKey</h2>');
     body.appendHtml('<pre>${timeseries.computeStats()}</pre>');
-    body.append(TimeseriesVisualization(timeseries).render());
+    // TODO(kevmoo): remove `NodeClue` cast when we no longer need to support
```
yjbanov, line 157: nit: s/Clue/Glue/

packages/web_benchmarks/pubspec.yaml
```diff
@@ -20,6 +20,7 @@ dependencies:
   shelf: ^1.2.0
   shelf_static: ^1.1.0
   test: ^1.19.5
+  web: '>=0.3.0 <0.5.0'
```
yjbanov, line 23: Why the `<0.5.0` constraint?

yjbanov (approved):

packages/web_benchmarks/lib/client.dart
kevmoo: Points for using the suggestion feature on github!

kevmoo:
```suggestion
    // TODO(kevmoo): remove `NodeGlue` cast when we no longer need to support
```

packages/web_benchmarks/pubspec.yaml
yjbanov: `<1.0.0` would also capture `0.4.0`, right? And it would also be less sensitive to version constraints in adjacent packages.

kevmoo: But we're not sure about what we're doing in `0.5.0+` This is a "good" range for things we've tested.

PR #5611, opened 2023-12-07
Add the ability to run web_benchmarks with Wasm

kenzieschmoll:
This uses the same run flags that the flutter benchmark tests use: https://github.com/flutter/flutter/blob/master/dev/devicelab/lib/tasks/web_benchmarks.dart#L36-L40

CC @kevmoo

### DIFF
--- a/packages/web_benchmarks/lib/src/runner.dart
+++ b/packages/web_benchmarks/lib/src/runner.dart
@@ -55,6 +55,7 @@ class BenchmarkServer {
     required this.chromeDebugPort,
     required this.headless,
     required this.treeShakeIcons,
+    this.useWasm = false,
     this.initialPage = defaultInitialPage,
   });
 
@@ -74,6 +75,9 @@ class BenchmarkServer {
 
   /// Whether to build the app in CanvasKit mode.
   final bool useCanvasKit;
+
+  /// Whether to build the app with dart2wasm.
+  final bool useWasm;
 
   /// The port this benchmark server serves the app on.
   final int benchmarkServerPort;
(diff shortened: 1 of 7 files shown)


packages/web_benchmarks/lib/src/runner.dart
```diff
@@ -55,6 +55,7 @@ class BenchmarkServer {
     required this.chromeDebugPort,
     required this.headless,
     required this.treeShakeIcons,
+    this.useWasm = false,
```
yjbanov, line 58: I wonder if this will have the same effect as `useCanvasKit` above. `useCanvasKit` ended up being not extensible enough, so we'll need to make a breaking change to accommodate Skwasm. I wonder if we should instead have a `CompilationOptions` object that can be extended in the future. cc @eyebrowsoffire

kenzieschmoll: I added a CompilationOptions class. PTAL @yjbanov @eyebrowsoffire

packages/web_benchmarks/lib/src/runner.dart
```diff
@@ -343,3 +339,33 @@ class BenchmarkServer {
...
 }
+
+/// Compilation options for bulding a Flutter web app.
+///
+/// This object holds metadata that is used to determine how the benchmark app
+/// should be built.
+class CompilationOptions {
+  /// Creates a [CompilationOptions] object.
+  const CompilationOptions({
+    this.renderer = WebRenderer.html,
```
kevmoo, line 350: The default should ABSOLUTELY be `canvaskit`. I'd say ditch the `html` option entirely!

kevmoo: Well, this is published. Maybe just default to `canvaskit`

eyebrowsoffire (approved):
Looks good to me!

PR #5623, opened 2023-12-08
[tool] Use ^ for Dart SDK

stuartmorgan:
The update-min-sdk command has been using explicitly ranges for the Dart SDK, because that used to be required. Current versions of `pub` no longer require that, and using `^` for Dart ranges is okay, so this updates the tooling to use that format in the future.

Also removes the special casing that changed the upper bound from 3 to 4, since we have passed the point where we are generating upper bounds less than 4 anyway.

To minimize churn, this doesn't update the existing pubspecs. We can incrementally adopt this going forward as we roll dependencies forward.

Fixes https://github.com/flutter/flutter/issues/139806

### DIFF
--- a/script/tool/lib/src/update_min_sdk_command.dart
+++ b/script/tool/lib/src/update_min_sdk_command.dart
@@ -69,16 +69,8 @@ class UpdateMinSdkCommand extends PackageLoopingCommand {
         YamlEditor(package.pubspecFile.readAsStringSync());
     if (dartRange != null &&
         (dartRange.min ?? Version.none) < _dartMinVersion) {
-      Version upperBound = _dartMinVersion.nextMajor;
-      // pub special-cases 3.0.0 as an upper bound to be treated as 4.0.0, and
-      // using 3.0.0 is now an error at upload time, so special case it here.
-      if (upperBound.major == 3) {
-        upperBound = upperBound.nextMajor;
-      }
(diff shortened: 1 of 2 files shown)


ditman (approved):
The `flutter` constraint forbids an upper range, so we're going to end up with something like this:

```yaml
sdk: ^3.2.0
flutter: ">=3.16.0"
```

(which had me doing a double take to grok why they're different :P)

ditman:
> Yes, but they were also different before (Dart requires upper bound, Flutter forbids it), so it's just differently different :)

Agreed, the `flutter` dependency now sticks out a little bit. Maybe we should propose a `+` prefix so `+3.16.0` means "anything after 3.16.0, with no upper bound" :P

ditman:
The tree thinks that it needs to be red:

Screenshot 2023-12-08 at 8 36 17 PM

But in reality, [build 605](https://ci.chromium.org/ui/p/flutter/builders/prod/Linux_android%20android_platform_tests_shard_4%20master/605/overview) passed.

The dashboard must have lost the memo.

(Manually squashing and merging this PR to get the dashboard unstuck.)

ditman:
OK, it seems that the dashboard will retry 3x automatically without us having to use the github UI, and that if I retry *from github* the dashboard will not pick it up. Retries after "red" should be launched from the dashboard, rather than github.

PR #5625, opened 2023-12-08
[ci] Fix indentation in labeler.yml

stuartmorgan:
Fixes an indentation bug in the config file that causes the task to fail.


ditman (approved):
LGTM! (UGH yaml (x_x))

PR #5628, opened 2023-12-08
[path_provider] De-flake getExternalStorageDirectories test

gmackall:
Fixes the flaky path provider test. 

Directories.listSync will sometimes not return the files in the directory (as discussed in https://github.com/flutter/flutter/issues/139378), but the files are there and (after letting it loop for quite a while and comparing to current flake) they always appear in the output of `Process.runSync('ls', ...)`. 

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
> Directories.listSync will sometimes not return the files in the directory (as discussed in [flutter/flutter#139378](https://github.com/flutter/flutter/issues/139378)), but the files are there and (after letting it loop for quite a while and comparing to current flake) they always appear in the output of `Process.runSync('ls', ...)`.

That's super weird. Is there a Dart issue filed that we can reference in a comment?

packages/path_provider/path_provider/example/integration_test/path_provider_test.dart
```diff
@@ -117,6 +117,8 @@ void _verifySampleFile(Directory? directory, String name) {
 
   file.writeAsStringSync('Hello world!');
   expect(file.readAsStringSync(), 'Hello world!');
-  expect(directory.listSync(), isNotEmpty);
+  // This check intentionally avoids using Directory.listSync due to
```
stuartmorgan, line 120: For the app-facing package we should probably make this codepath Android-only, and have other platforms keep using `listSync`.

gmackall: Changed!

stuartmorgan (approved):
LGTM once autoformat is happy.

gmackall:
> LGTM once autoformat is happy.

I don't know why I assumed that was the changelog/version checker complaining 🤦. Thanks for taking another look, I'll fix that and then land

gmackall:
Tree is blocked on two failures, one of which is this test failing (https://github.com/flutter/packages/pull/5619 only disabled the test in `path_provider_android` but this same test is also run by the app-facing package on android).

Should this land on red to fix, or does the other blocker need to be resolved first?

PR #5630, opened 2023-12-08
[web_benchmarks] Add support for analyzing benchmark results

kenzieschmoll:
This PR adds a new top level library `analysis.dart` that exposes methods:
* `computeDelta`
* `computeAverage`

These were already added to DevTools in https://github.com/flutter/devtools/pull/6918 and https://github.com/flutter/devtools/pull/6920, respectively, but I think it makes more sense to upstream this logic into `package:web_benchmarks` so that other users of this package can take advantage of it.

As part of this PR, I have updated the README to give an example of how to use the analysis features.


elliette (approved):

packages/web_benchmarks/lib/analysis.dart
```diff
@@ -0,0 +1,142 @@
...
+
+export 'src/benchmark_result.dart';
+
+/// Returns the average of the benchmark results in [results].
+///
+/// Each element in [results] is expected to have identical benchmark names and
+/// metrics; otherwise, an [Exception] will be thrown.
+BenchmarkResults computeAverage(List<BenchmarkResults> results) {
+  if (results.isEmpty) {
+    throw Exception('Cannot take average of empty list.');
```
yjbanov, line 16: Would `ArgumentError` be more appropriate?

packages/web_benchmarks/lib/analysis.dart
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  BenchmarkResults totalSum = results.first;
+  for (int i = 1; i < results.length; i++) {
+    final BenchmarkResults current = results[i];
+    totalSum = totalSum._sumWith(current);
+  }
```
yjbanov, line 23: Not sure if it helps here, but since the element and result types are the same a shortcut is to use `Iterable.reduce`.

packages/web_benchmarks/lib/analysis.dart
```diff
@@ -0,0 +1,142 @@
...
+
+  final Map<String, List<Map<String, Object?>>> average = totalSum.toJson();
```
yjbanov, line 25: It's interesting that you had to pipe it through JSON. It tells me something is missing. Can we extend the API of `BenchmarkResults` so you can do the same computation without JSON?

packages/web_benchmarks/lib/analysis.dart
```diff
@@ -0,0 +1,142 @@
...
+      average[benchmark]![i][BenchmarkScore.valueKey] = averageValue;
+    }
+  }
+  return BenchmarkResults.parse(average);
+}
+
+/// Computes the delta between [test] and [baseline], and returns the results
+/// as a JSON object where each benchmark score entry contains a new field
+/// 'delta' with the metric value comparison.
+Map<String, List<Map<String, Object?>>> computeDelta(
```
yjbanov, line 40: Coupling the API to JSON seems to break separation of concerns. JSON is just a serialization format. Can we wrap the result into a regular strongly-typed object and then give it a `toJson` method for cases when you just want to save it to file or send over the network?

kenzieschmoll: Added an optional field `delta` to `BenchmarkScore` to use the strongly-typed objects. PTAL.

packages/web_benchmarks/lib/src/benchmark_result.dart
```diff
@@ -31,11 +35,18 @@ class BenchmarkScore {
   /// The result of measuring a particular metric in this benchmark run.
   final num value;
 
+  /// Optional delta value describing the difference between this metric's score
+  /// and the score of a matching metric from another [BenchmarkResults].
+  ///
+  /// This value may be assigned by the [computeDelta] analysis method.
+  num? delta;
```
yjbanov, line 42: This class is designed to be immutable. Is it possible to instead make this field `final`, add it to the constructor as an optional field, and then have `computeDelta` create new `BenchmarkScore` objects (and a new `BenchmarkResults` object) that contain the deltas, while keeping the baseline and test object intact?

kenzieschmoll: done

packages/web_benchmarks/lib/analysis.dart
```diff
@@ -48,22 +49,36 @@ BenchmarkResults computeDelta(
...
-      // Add the delta to the [testMetric].
-      score.delta = (score.value - baselineScore.value).toDouble();
+      final BenchmarkScore scoreWithDelta = BenchmarkScore(
+        metric: testScore.metric,
+        value: testScore.value,
+        delta: (testScore.value - baselineScore.value).toDouble(),
+      );
+      deltaScores
+        ..removeAt(i)
+        ..insert(i, scoreWithDelta);
```
yjbanov, line 70: This still treats `BenchmarkResults` as mutable, and the reason this works is due to the misfortune that in Dart lists and maps are mutable by default. Can we create a new `Map<String, List<BenchmarkScore>> scoreDeltas`, populate it with deltas and then `return BencharkResults(scoreDeltas)`?

kenzieschmoll: Okay switched to building a map and then parsing into a new BenchmarkResults object. I also add `growable: false` to the construction of each scores list so that the lists are actually immutable.

yjbanov (approved):
lgtm

PR #5701, opened 2023-12-18
[go_router] Add `titleBuilder` to `GoRoute` and `GoRouterState`

getBoolean:
This is something I hacked together, bringing the `titleBuilder` from `auto_route` to `go_router`. I suspect it will need some changes, as I am not familiar with the internals of GoRouter.

- Issue: https://github.com/flutter/flutter/issues/140297

## Pre-launch Checklist

- [x] I signed the [CLA].


chunhtai:
This seems to be a very limited use case, I think the root problem is that the shellroute doesn't have information about the current child. I think this is something we should figure out a better API to expose such thing. WDYT?

getBoolean:
@chunhtai yes that would work, though I wasn't able to come up with a more generic solution without casting. Perhaps a solution could involve generics.

getBoolean:
I experimented with having adding generic type that the custom data builder can return. it is possible, but i had to add generics to nearly every go_router class. (Probably not what we want, but what are your thoughts?)
https://github.com/getBoolean/flutter_packages/commit/80a70229bf2edd7770d71deb8f03d419ab7bdfba

getBoolean:
In 0ddddb6, I converted the title builder to extensions, similar to Flutter's theme extensions. I think this would work the best @chunhtai 

Usage:

```dart
GoRoute(
  path: 'b',
  extensions: <GoRouterStateExtensionBuilder>[
    (_, __) => const RouteTitleExtension('B'),
    (_, __) => const OtherCustomDataExtension('B.2'),
  ],
  builder: (BuildContext context, GoRouterState state) {
    return const Scaffold(
      body: Text('Screen 2'),
    );
  },
)

class RouteTitleExtension
    implements GoRouterStateExtension<RouteTitleExtension> {
  const RouteTitleExtension(this.title);

  final String title;
}

GoRouterState.of(context).extension<RouteTitleExtension>(context).title
```

getBoolean:
New PR opened based on 13.0.0 https://github.com/flutter/packages/pull/5736

PR #5717, opened 2023-12-18
[various] Sync lints with flutter/flutter

goderbauer:
Applying the latest analysis_options.yaml from flutter/flutter to this repo. Most fixes were auto-generated by `dart fix`.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -164,7 +164,7 @@ linter:
     - prefer_is_not_empty
     - prefer_is_not_operator
     - prefer_iterable_whereType
-    - prefer_mixin
+    # - prefer_mixin # DIFFERENT FROM FLUTTER/FLUTTER: enable when v2.1.7 of plugin_platform_interface is the oldest supported version (which makes MockPlatformInterfaceMixin a mixin class)
     # - prefer_null_aware_method_calls # "call()" is confusing to people new to the language since it's not documented anywhere
     - prefer_null_aware_operators
     - prefer_relative_imports
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,13 +1,14 @@
 # Specify analysis options.
 #
 # This file is a copy of analysis_options.yaml from flutter repo
-# as of 2022-07-27, but with some modifications marked with
+# as of 2023-12-18, but with some modifications marked with
 # "DIFFERENT FROM FLUTTER/FLUTTER" below. The file is expected to
 # be kept in sync with the master file from the flutter repo.
 
 analyzer:
   language:
     strict-casts: true
+    strict-inference: true
     strict-raw-types: true
   errors:
     # allow self-reference to deprecated members (we do this because otherwise we have
@@ -21,11 +22,10 @@ analyzer:
 linter:
   rules:
     # This list is derived from the list of all available lints located at
-    # https://github.com/dart-lang/linter/blob/master/example/all.yaml
+    # https://github.com/dart-lang/linter/blob/main/example/all.yaml
     - always_declare_return_types
     - always_put_control_body_on_new_line
     # - always_put_required_named_parameters_first # we prefer having parameters in the same order as fields https://github.com/flutter/flutter/issues/10219
-    - always_require_non_null_named_parameters
     - always_specify_types
     # - always_use_package_imports # we do this commonly
     - annotate_overrides
@@ -33,7 +33,7 @@ linter:
     - avoid_bool_literals_in_conditional_expressions
     # - avoid_catches_without_on_clauses # blocked on https://github.com/dart-lang/linter/issues/3023
     # - avoid_catching_errors # blocked on https://github.com/dart-lang/linter/issues/3023
-    - avoid_classes_with_only_static_members
+    # - avoid_classes_with_only_static_members # we do this commonly for `abstract final class`es
     - avoid_double_and_int_checks
     - avoid_dynamic_calls
     - avoid_empty_else
@@ -42,7 +42,7 @@ linter:
     - avoid_field_initializers_in_const_classes
     # - avoid_final_parameters # incompatible with prefer_final_parameters
     - avoid_function_literals_in_foreach_calls
-    - avoid_implementing_value_types
+    # - avoid_implementing_value_types # see https://github.com/dart-lang/linter/issues/4558
     - avoid_init_to_null
     - avoid_js_rounded_ints
     # - avoid_multiple_declarations_per_line # seems to be a stylistic choice we don't subscribe to
@@ -54,8 +54,6 @@ linter:
     - avoid_relative_lib_imports
     - avoid_renaming_method_parameters
     - avoid_return_types_on_setters
-    - avoid_returning_null
-    - avoid_returning_null_for_future
     - avoid_returning_null_for_void
     # - avoid_returning_this # there are enough valid reasons to return `this` that this lint ends up with too many false positives
     - avoid_setters_without_getters
@@ -77,17 +75,19 @@ linter:
     - cast_nullable_to_non_nullable
     # - close_sinks # not reliable enough
     - collection_methods_unrelated_type
-    # - combinators_ordering # DIFFERENT FROM FLUTTER/FLUTTER: This isn't available on stable yet.
+    - combinators_ordering
     # - comment_references # blocked on https://github.com/dart-lang/linter/issues/1142
     - conditional_uri_does_not_exist
     # - constant_identifier_names # needs an opt-out https://github.com/dart-lang/linter/issues/204
     - control_flow_in_finally
     - curly_braces_in_flow_control_structures
+    - dangling_library_doc_comments
     - depend_on_referenced_packages
     - deprecated_consistency
+    # - deprecated_member_use_from_same_package # we allow self-references to deprecated members
     # - diagnostic_describe_all_properties # enabled only at the framework level (packages/flutter/lib)
     - directives_ordering
-    # - discarded_futures # not yet tested
+    # - discarded_futures # too many false positives, similar to unawaited_futures
     # - do_not_use_environment # there are appropriate times to use the environment, especially in our tests and build logic
     - empty_catches
     - empty_constructor_bodies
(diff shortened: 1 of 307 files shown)


packages/rfw/lib/src/dart/text.dart
```diff
@@ -2127,7 +2003,7 @@ class _Parser {
   void _advance() {
     assert(_source.current is! _EofToken);
     final bool advanced = _source.moveNext();
-    assert(advanced == true); // see https://github.com/dart-lang/sdk/issues/47017
+    assert(advanced); // see https://github.com/dart-lang/sdk/issues/47017
```
yjbanov, line 2006: The `== true` was there on purpose. See the issue link.

packages/rfw/lib/src/dart/text.dart
```diff
@@ -2457,6 +2333,6 @@ class _Parser {
       throw ParserException._expected(expectation, _source.current);
     }
     final bool more = _source.moveNext();
-    assert(more == false); // see https://github.com/dart-lang/sdk/issues/47017
+    assert(!more); // see https://github.com/dart-lang/sdk/issues/47017
```
yjbanov, line 2336: Same here.

packages/go_router/example/lib/books/src/screens/books.dart
```diff
@@ -105,10 +102,8 @@ class _BooksScreenState extends State<BooksScreen>
     switch (index) {
       case 1:
         context.go('/books/new');
-        break;
       case 2:
         context.go('/books/all');
-        break;
       case 0:
       default:
```
chunhtai, line 108: the default should probably be removed

packages/go_router_builder/example/lib/shell_route_with_keys_example.dart
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// ignore_for_file: public_member_api_docs
+// ignore_for_file: public_member_api_docs, unreachable_from_main
```
chunhtai, line 5: should this be in the example folder analyze_options?

packages/rfw/lib/src/dart/text.dart
goderbauer: The issue is marked as fixed, though. Should we leave the `== true` out and remove the comment as well or do you prefer to still keep the `== true`?

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
```diff
@@ -28,35 +28,34 @@ const MethodCodec _codec = JSONMethodCodec();
 ///
 /// This is a class instead of an enum to allow future customizability e.g.
 /// opening a link in a specific iframe.
-class LinkTarget {
```
yjbanov, line 31: The dart doc above explains why this is a class instead of an enum. We should either change the doc, or if the old reasoning is still valid, leave it as is.

gspencergoog (approved):
LGTM for `flutter_adaptive_scaffold`.

packages/rfw/lib/src/dart/text.dart
yjbanov: It looks like it was fixed for real, so we should remove the link to the issue then.

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
yjbanov: cc @mdebbar

packages/rfw/lib/src/dart/text.dart
goderbauer: see previous response.

packages/go_router/example/lib/books/src/screens/books.dart
goderbauer: That doesn't seem related to applying the new lints and I'd like to leave that out of this already sizable PR.

packages/go_router_builder/example/lib/shell_route_with_keys_example.dart
goderbauer: The examples folder currently doesn't have its own analyze_options and introducing one is a bit of a process in the repro. I would like to defer on that for a future PR.

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
ditman: The way I understand this is that [`target` is an arbitrary string](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target), but there are values with a special meaning, hence why this isn't an enum.

Can users add their own enum values to LinkTarget? (I can't remember if LinkTarget had a custom constructor for users though)

jmagman (approved):
iOS packages LGTM

chunhtai (approved):
LGTM

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
goderbauer: LinkTarget only had a private constructor, so it wasn't user-extensible. I am happy to revert this back to a class and slap an ignore for the lint on it. Or I can remove the doc comment about the enum. Let me know what you prefer.

keyonghan (approved):
`metrics_center` LGTM

bparrishMines (approved):
LGTM for webview_flutter, in_app_purchase, and camera pacakges

tarrinneal (approved):
lgtm image_picker, pigeon, shared_prefs, video_player

domesticmouse (approved):
LGTM for `flutter_markdown`

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
yjbanov: I imagine this was motivated by a somewhat richer `window.open` API than what the plugin supports currently: https://developer.mozilla.org/en-US/docs/Web/API/Window/open

reidbaker (approved):

packages/url_launcher/url_launcher_platform_interface/lib/link.dart
mdebbar: The idea here was to mirror the [HTML target attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target), which has a set of predefined values (`_blank`, `_self`, etc), but can also take an arbitrary string that refers to an already open window or iframe.

Right now we don't expose this capability to users, but if someone asks for it, we will. My preference would be to keep this as a class so when the day comes, we don't have to switch it back from enum to a class.

goderbauer: Sounds good. I moved this back to be a class.

cbracken (approved):

gmackall (approved):

Piinks (approved):

analysis_options.yaml
```diff
@@ -157,7 +164,7 @@ linter:
     - prefer_is_not_empty
     - prefer_is_not_operator
     - prefer_iterable_whereType
-    # - prefer_mixin # Has false positives, see https://github.com/dart-lang/linter/issues/3018
+    # - prefer_mixin # DIFFERENT FROM FLUTTER/FLUTTER: enable when v2.1.7 of plugin_platform_interface is the oldest supported version (which makes MockPlatformInterfaceMixin a mixin class)
```
stuartmorgan, line 167: What does this comment mean exactly? The only version of any package we support is the latest version.

packages/flutter_migrate/lib/src/base/io.dart
```diff
@@ -25,6 +25,7 @@
 /// about any additional exports that you add to this file, as doing so will
 /// increase the API surface that we have to test in Flutter tools, and the APIs
 /// in `dart:io` can sometimes be hard to use in tests.
+library;
```
stuartmorgan, line 28: Can we remove this and just make the comments  `//` instead? I'm not sure why they are `///` in the first place.

packages/plugin_platform_interface/test/plugin_platform_interface_test.dart
```diff
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+// ignore_for_file: unreachable_from_main
```
stuartmorgan, line 5: Why do we need this in a test file?

stuartmorgan (approved):
LGTM with a couple of questions. Thanks for doing this!

analysis_options.yaml
goderbauer: The `analyze_downgraded` check was pulling in a version of the package where `MockPlatformInterfaceMixin` wasn't a mixin class yet and hence analysis would fail on packages that were using it as a mixin. As soon as  the `analyze_downgraded` check pulls in version 2.1.7 or newer of this package, the lint can be enabled. I guess, we could force this by updating the pubspec.yamls of all packages that depend on plugin_platform_interface to:

```yaml
plugin_platform_interface: ^2.1.7
```

If it's okay to do that, I can attempt this in a follow-up PR.

packages/flutter_migrate/lib/src/base/io.dart
goderbauer: Done

analysis_options.yaml
stuartmorgan: Yes, upping the dependencies to 2.1.7 would be fine.

(That's the only way `analyze_downgraded` would pull 2.1.7. It doesn't have its own logic, it just runs `pub downgrade` to get the oldest allowed package set.)

packages/plugin_platform_interface/test/plugin_platform_interface_test.dart
goderbauer: Added a comment to explain this file-level ignore. I didn't want to add it to individual instances here because then it would show up in the docs since the unreachable code is in a docregion.

As to why this is needed for a test file: The lint is mostly targeting tests to catch instances where a test case is deleted or refactored, but its supporting infrastructure (e.g. public mock classes and helper methods) isn't removed.

analysis_options.yaml
goderbauer: Awesome, thanks for confirming. I will prepare a follow-up PR to do just that after this one lands.

PR #5735, opened 2023-12-21
Disable `presubmit: false` targets for recipes CQ

keyonghan:
These targets are not enabled in presubmit, but enabled in CQ. This caused latest changes not surfaced from CQ side: CQ runs are based on out dated try runs. For example: `Mac_arm64 ios_platform_tests_shard_1 stable` is using the old build executed on 12/13: https://ci.chromium.org/ui/p/flutter/builders/try/Mac_arm64%20ios_platform_tests_shard_1%20stable

This PR disables CQ for targets that are with `presubmit: false`, to unblock recipes CL: https://flutter-review.googlesource.com/c/recipes/+/53301

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -1130,7 +1130,6 @@ targets:
     timeout: 60
     properties:
       channel: stable
-      add_recipes_cq: "true"
       version_file: flutter_stable.version
       target_file: macos_platform_tests.yaml
       env_variables: >-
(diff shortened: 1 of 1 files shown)


godofredoc (approved):

stuartmorgan:
So what would catch a recipe change that broke handling of non-`main` channels after this PR? We would have to hit it as OOB breakage in the packages repo and then roll back the recipe? That doesn't seem desireable.

keyonghan:
There are still `stable` targets enabled in recipes cq, e.g. [Linux_android android_build_all_packages stable](https://github.com/flutter/packages/blob/main/.ci.yaml#L403). Is this enough?

This PR disables only those that are not running in packages presubmit.

stuartmorgan:
Ah, yes, that's probably fine then.

PR #5736, opened 2023-12-22
[go_router] Add `topRoute` to `GoRouterState`

getBoolean:
This PR exposes the current top route to `GoRouterState`, this allows `ShellRoute` to know what is the current child and process the state accordingly.

- Issue: https://github.com/flutter/flutter/issues/140297

This could be used like this, given that each `GoRoute` had the `name` parameter given
```dart
StatefulShellRoute.indexedStack(
  parentNavigatorKey: rootNavigatorKey,
  builder: (
    BuildContext context,
    GoRouterState state,
    StatefulNavigationShell navigationShell,
  ) {
    final String? routeName =
        GoRouterState.of(context).topRoute.name;
    final String title = switch (routeName) {
      'a' => 'A',
      'b' => 'B',
      _ => 'Unknown',
    };
    return Column(
      children: <Widget>[
        Text(title),
        Expanded(child: navigationShell),
      ],
    );
  },
  ...
}
```

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/go_router/lib/src/match.dart
```diff
@@ -720,13 +722,30 @@ class RouteMatchList with Diagnosticable {
...
+  /// The last leaf route or null if [matches] is empty
+  ///
+  /// If the last RouteMatchBase from [matches] is a ShellRouteMatch, it
+  /// recursively goes into its [ShellRouteMatch.matches] until it reach the leaf
+  /// [RouteMatch].
+  RouteMatch? get lastOrNull {
+    if (matches.isEmpty) {
+      return null;
+    }
+    if (matches.last is RouteMatch) {
```
chunhtai, line 743: this can reuse the `last` getter, `return last`

packages/go_router/lib/src/state.dart
```diff
@@ -74,6 +76,9 @@ class GoRouterState {
   /// ```
   final ValueKey<String> pageKey;
 
+  /// The current matched top route
```
chunhtai, line 79: This needs more description. May be mention what this mean if you this is a this GoRouteState represent a shellroute.

chunhtai (commented):
I think the code LGTM, would be good to add an example in the example/ to showcase how to use this new property as well

packages/go_router/lib/src/state.dart
getBoolean: I've updated the documentation with the following, is this what you were thinking of?

```dart
  /// The current matched top route associated with this state.
  ///
  /// If this state represents a [ShellRoute], the top [GoRoute] will be the current
  /// matched location associated with the [ShellRoute]. This allows the [ShellRoute]'s
  /// associated [GoRouterState] to be uniquely identified using [GoRoute.name]
  final GoRoute? topRoute;
```

packages/go_router/example/lib/shell_route_top_route.dart
```diff
@@ -0,0 +1,326 @@
...
+    final RouteMatch lastMatch = currentConfiguration.last;
+    final Uri location = lastMatch is ImperativeRouteMatch
+        ? lastMatch.matches.uri
+        : currentConfiguration.uri;
+    final bool canPop = location.pathSegments.length > 1;
+    final Widget button = canPop
+        ? BackButton(onPressed: GoRouter.of(context).pop)
+        : const SizedBox.shrink(key: ValueKey<String>('NoLeadingButton'));
+    return button;
+  }
```
getBoolean, line 216: I had to manually create the leading button since this is in a scaffold above the children. Is there a better way of doing this? For context, auto_route has `AutoLeadingButton`

hangyujin (approved):
LGTM

hangyujin:
@johnpryan hi john,  chun-heng is on vacation, can you take a second look at this PR?

johnpryan:
Yep I can review.

packages/go_router/example/lib/shell_route_top_route.dart
```diff
@@ -0,0 +1,326 @@
...
+final GlobalKey<NavigatorState> _shellNavigatorKey =
+    GlobalKey<NavigatorState>(debugLabel: 'shell');
+
+// This scenario demonstrates how to set up nested navigation using ShellRoute,
+// which is a pattern where an additional Navigator is placed in the widget tree
+// to be used instead of the root navigator. This allows deep-links to display
+// pages along with other UI components such as a BottomNavigationBar.
+//
+// This example demonstrates how to display a route within a ShellRoute and also
+// using the current topRoute to determine the title of the AppBar.
```
johnpryan, line 19: We should update this comment to describe this sample.

packages/go_router/example/lib/shell_route_top_route.dart
```diff
@@ -0,0 +1,326 @@
...
+          // Could also use an enum/constant strings for route names.
+          final String title = switch (routeName) {
+            'a' => 'A Screen',
+            'a.details' => 'A Details',
+            'b' => 'B Screen',
+            'b.details' => 'B Details',
+            'c' => 'C Screen',
+            'c.details' => 'C Details',
+            _ => 'Unknown',
+          };
```
johnpryan, line 49: I like this sample! It demonstrates the idea perfectly.

packages/go_router/example/lib/shell_route_top_route.dart
```diff
@@ -0,0 +1,326 @@
...
+            label: 'C Screen',
+          ),
+        ],
+        currentIndex: _calculateSelectedIndex(context),
+        onTap: (int idx) => _onItemTapped(idx, context),
+      ),
+    );
+  }
+
+  /// Animates the AppBar title and leading button when it changes.
```
johnpryan, line 188: We might not need to have an animated app bar for this sample, if we can simplify the code to just `AppBar(title: _getScreenTitle(context),)` that might be easier from a readability point of view.

johnpryan (approved):
Looks great!

packages/go_router/example/lib/shell_route_top_route.dart
getBoolean: writing isn't my strong suit, is this better?

```
// This example demonstrates how use topRoute in a ShellRoute to create the
// title in the AppBar above the child, which is different for each GoRoute.
```

packages/go_router/example/lib/shell_route_top_route.dart
getBoolean: Thanks! It would be better if it included path parameters too, but I wanted to keep the sample simple.

packages/go_router/example/lib/shell_route_top_route.dart
getBoolean: That makes sense, I've removed the animated related code, though it still needs the custom leading back button

packages/go_router/example/lib/shell_route_top_route.dart
johnpryan: That works, thanks!

johnpryan (approved):
LGTM, we should also add this API to go_router_builder if we haven't already.

hangyujin:
yes,  go_router_builder changes are usually done in separated PRs. 

@johnpryan  i don't think we need to change go_router_builder to add this `topRoute` API.  `topRoute` here is just part of [GoRouterState] and it's not set when GoRouter is built.

getBoolean:
Are we waiting for @chunhtai to return before merging this?

getBoolean:
Thanks!

PR #5742, opened 2023-12-22
[go_router] Expose full `Uri` on `GoRouterState` in `GoRouterRedirect`

ChopinDavid:
A number of developers have voiced the desire to be able to know whether a `GoRouterState` is the result of a deep-link or in-app navigation from within their `GoRouter`'s `redirect` method. This can be accomplished by exposing the `Uri`'s `scheme` and `host` on the `GoRouterState`. This way, we can know whether the `GoRouterState` is the result of a deep-link by checking `state.uri.scheme != null` or `state.uri.host != null`.

This PR would close [#103659](https://github.com/flutter/flutter/issues/103659#issuecomment-1865112881).

No tests were broken as a result of this change, and we have added coverage for this change through new tests.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/go_router/lib/src/information_provider.dart
+++ b/packages/go_router/lib/src/information_provider.dart
@@ -105,16 +105,9 @@ class GoRouteInformationProvider extends RouteInformationProvider
     final bool replace;
     switch (type) {
       case RouteInformationReportingType.none:
-        const DeepCollectionEquality deepCollectionEquality =
-            DeepCollectionEquality();
-        if (deepCollectionEquality.equals(
-                _valueInEngine.uri.path, routeInformation.uri.path) &&
-            deepCollectionEquality.equals(_valueInEngine.uri.queryParameters,
-                routeInformation.uri.queryParameters) &&
-            deepCollectionEquality.equals(
-                _valueInEngine.uri.fragment, routeInformation.uri.fragment) &&
-            deepCollectionEquality.equals(
-                _valueInEngine.state, routeInformation.state)) {
+        if (!_valueHasChanged(
+            newLocationUri: routeInformation.uri,
+            newState: routeInformation.state)) {
           return;
         }
         replace = _valueInEngine == _kEmptyRouteInformation;
@@ -143,14 +136,9 @@ class GoRouteInformationProvider extends RouteInformationProvider
 
   void _setValue(String location, Object state) {
     final Uri uri = Uri.parse(location);
-    const DeepCollectionEquality deepCollectionEquality =
-        DeepCollectionEquality();
+
     final bool shouldNotify =
-        !deepCollectionEquality.equals(_value.uri.path, uri.path) ||
-            !deepCollectionEquality.equals(
-                _value.uri.queryParameters, uri.queryParameters) ||
-            !deepCollectionEquality.equals(_value.uri.fragment, uri.fragment) ||
-            !deepCollectionEquality.equals(_value.state, state);
+        _valueHasChanged(newLocationUri: uri, newState: state);
     _value = RouteInformation(uri: Uri.parse(location), state: state);
     if (shouldNotify) {
       notifyListeners();
@@ -245,6 +233,19 @@ class GoRouteInformationProvider extends RouteInformationProvider
       _valueInEngine = _kEmptyRouteInformation;
     }
     notifyListeners();
+  }
+
+  bool _valueHasChanged(
+      {required Uri newLocationUri, required Object? newState}) {
+    const DeepCollectionEquality deepCollectionEquality =
+        DeepCollectionEquality();
+    return !deepCollectionEquality.equals(
+            _value.uri.path, newLocationUri.path) ||
+        !deepCollectionEquality.equals(
+            _value.uri.queryParameters, newLocationUri.queryParameters) ||
+        !deepCollectionEquality.equals(
+            _value.uri.fragment, newLocationUri.fragment) ||
+        !deepCollectionEquality.equals(_value.state, newState);
   }
 
   @override
--- a/packages/go_router/lib/src/information_provider.dart
+++ b/packages/go_router/lib/src/information_provider.dart
@@ -11,10 +11,6 @@ import 'package:flutter/widgets.dart';
 
 import 'match.dart';
 
-// TODO(chunhtai): remove this ignore and migrate the code
-// https://github.com/flutter/flutter/issues/124045.
-// ignore_for_file: deprecated_member_use
-
 /// The type of the navigation.
 ///
 /// This enum is used by [RouteInformationState] to denote the navigation
@@ -85,7 +81,7 @@ class GoRouteInformationProvider extends RouteInformationProvider
     Listenable? refreshListenable,
   })  : _refreshListenable = refreshListenable,
         _value = RouteInformation(
-          location: initialLocation,
+          uri: Uri.parse(initialLocation),
           state: RouteInformationState<void>(
               extra: initialExtra, type: NavigatingType.go),
         ),
@@ -96,8 +92,8 @@ class GoRouteInformationProvider extends RouteInformationProvider
   final Listenable? _refreshListenable;
 
   static WidgetsBinding get _binding => WidgetsBinding.instance;
-  static const RouteInformation _kEmptyRouteInformation =
-      RouteInformation(location: '');
+  static final RouteInformation _kEmptyRouteInformation =
+      RouteInformation(uri: Uri.parse(''));
 
   @override
   void routerReportsNewRouteInformation(RouteInformation routeInformation,
@@ -109,7 +105,7 @@ class GoRouteInformationProvider extends RouteInformationProvider
     final bool replace;
     switch (type) {
       case RouteInformationReportingType.none:
-        if (_valueInEngine.location == routeInformation.location &&
+        if (_valueInEngine.uri.path == routeInformation.uri.path &&
             const DeepCollectionEquality()
                 .equals(_valueInEngine.state, routeInformation.state)) {
           return;
@@ -122,10 +118,7 @@ class GoRouteInformationProvider extends RouteInformationProvider
     }
     SystemNavigator.selectMultiEntryHistory();
     SystemNavigator.routeInformationUpdated(
-      // TODO(chunhtai): remove this ignore and migrate the code
-      // https://github.com/flutter/flutter/issues/124045.
-      // ignore: unnecessary_null_checks, unnecessary_non_null_assertion
-      location: routeInformation.location!,
+      uri: routeInformation.uri,
       state: routeInformation.state,
       replace: replace,
     );
@@ -146,8 +139,8 @@ class GoRouteInformationProvider extends RouteInformationProvider
 
   void _setValue(String location, Object state) {
     final bool shouldNotify =
-        _value.location != location || _value.state != state;
-    _value = RouteInformation(location: location, state: state);
+        _value.uri.path != location || _value.state != state;
+    _value = RouteInformation(uri: Uri.parse(location), state: state);
     if (shouldNotify) {
       notifyListeners();
     }
@@ -272,13 +265,6 @@ class GoRouteInformationProvider extends RouteInformationProvider
   Future<bool> didPushRouteInformation(RouteInformation routeInformation) {
     assert(hasListeners);
     _platformReportsNewRouteInformation(routeInformation);
-    return SynchronousFuture<bool>(true);
-  }
-
-  @override
-  Future<bool> didPushRoute(String route) {
-    assert(hasListeners);
-    _platformReportsNewRouteInformation(RouteInformation(location: route));
     return SynchronousFuture<bool>(true);
   }
 }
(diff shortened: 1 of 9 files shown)


ChopinDavid:
@superwes It seems Dan would need to sign the Google CLA to pass all checks as the first commit of this PR was authored by him. [See here](https://github.com/flutter/packages/pull/5742/checks?check_run_id=19904204363).

packages/go_router/lib/src/parser.dart
```diff
@@ -86,7 +86,8 @@ class GoRouteInformationParser extends RouteInformationParser<RouteMatchList> {
         // to uri, empty path check might be required here; see
         // https://github.com/flutter/packages/pull/5113#discussion_r1374861070
```
chunhtai, line 87: The entire todo and ignores above can be removed, and we need to do empty path check here as well per comment above

chunhtai (commented):
Since we are migrating to uri, can you double check every place with todo linked with https://github.com/flutter/flutter/issues/124045 ?

ChopinDavid:
@chunhtai I went ahead and migrated all occurrences of `RouteInformation.location` per the migration guide. I did have a few questions:

- It seems that [this TODO](https://github.com/wwt/flutter-packages/blob/a266a281374c99063c67e815a7b22b515876ab25/packages/go_router/lib/src/information_provider.dart#L136C1-L138) can go ahead and be removed?
- I went ahead and removed the `didPushRoute` override entirely from `InformationProvider`. Does this sound correct?

packages/go_router/lib/src/parser.dart
ChopinDavid: > The entire todo and ignores above can be removed, and we need to do empty path check here as well per comment above

I am not sure that we necessarily need to do the empty path check. It seems that you feel we need to do this check so that we do not re-introduce [this issue](https://github.com/flutter/flutter/issues/133928). But when launching our company's app through a deep link with no path (https://www.wwt.com) from a "dead" state, we are not seeing this `GoException: no routes for location` error message and the app launches as expected.

Maybe you can check to see if this issue has been re-introduced on your side?

packages/go_router/lib/src/information_provider.dart
```diff
@@ -109,7 +105,7 @@ class GoRouteInformationProvider extends RouteInformationProvider
     final bool replace;
     switch (type) {
       case RouteInformationReportingType.none:
-        if (_valueInEngine.location == routeInformation.location &&
+        if (_valueInEngine.uri.path == routeInformation.uri.path &&
```
chunhtai, line 108: This would also need to check the query parameter and hash fragment

packages/go_router/lib/src/information_provider.dart
```diff
@@ -149,8 +142,8 @@ class GoRouteInformationProvider extends RouteInformationProvider
 
   void _setValue(String location, Object state) {
     final bool shouldNotify =
-        _value.location != location || _value.state != state;
-    _value = RouteInformation(location: location, state: state);
+        _value.uri.path != location || _value.state != state;
```
chunhtai, line 145: same here this should also account for query parameter and hashfragment

packages/go_router/lib/src/parser.dart
chunhtai: Let me check and I will get back to you

chunhtai:
>

> @chunhtai I went ahead and migrated all occurrences of `RouteInformation.location` per the migration guide. I did have a few questions:
> 
> * It seems that [this TODO](https://github.com/wwt/flutter-packages/blob/a266a281374c99063c67e815a7b22b515876ab25/packages/go_router/lib/src/information_provider.dart#L136C1-L138) can go ahead and be removed?

It seems so, yes can you try to remove it and see if the analyzer complains or not

> * I went ahead and removed the `didPushRoute` override entirely from `InformationProvider`. Does this sound correct?

That sounds correct.

packages/go_router/lib/src/information_provider.dart
ChopinDavid: So something like
```dart
if (_valueInEngine.uri.path == routeInformation.uri.path &&
            _valueInEngine.uri.queryParameters ==
                routeInformation.uri.queryParameters &&
            _valueInEngine.uri.fragment == routeInformation.uri.fragment &&
            const DeepCollectionEquality()
                .equals(_valueInEngine.state, routeInformation.state))
```
?

packages/go_router/lib/src/parser.dart
chunhtai: I am testing in android, it looks like when cold start without the path the uri doesn't contain the host name and scheme. I am not sure what happen here.

If I warm start without the path, it would show no route for location. It seems to me we need to handle empty path here.

packages/go_router/lib/src/information_provider.dart
chunhtai: you probably need to do collection equality for query parameter

packages/go_router/lib/src/parser.dart
ChopinDavid: Yeah, I am actually seeing that we get the no route for location error when launching `www.domain.com` while the app is already running in the background.

It's almost as if we need to treat all `www.domain.com` URIs as `www.domain.com/` so that the location defaults to `/`. It makes sense that having a location of an empty string wouldn't match anything.

chunhtai:
looks like the CLA is not signed, this can happen if the commit history contain too many unrelated commit, you probably should do a sqush and force push, or just reset

ChopinDavid:
> looks like the CLA is not signed, this can happen if the commit history contain too many unrelated commit, you probably should do a sqush and force push, or just reset

The first commit on this branch was made by another member of my team. He probably just needs to sign the CLA. I will check with him in the coming days.

packages/go_router/lib/src/parser.dart
ChopinDavid: @chunhtai Does this sound like the right approach to you?

ChopinDavid: @chunhtai we went ahead and added the empty path check to ensure that `routeInformation` with an empty `.uri.path` has a `/` appended to the end of it. This behavior matches [what we had previously been doing with `routeInformation.location`](https://github.com/flutter/flutter/blob/b62bb518f4ce2a2df6ca7abec74e73629b61aab5/packages/flutter/lib/src/widgets/router.dart#L66-L77).

packages/go_router/lib/src/information_provider.dart
ChopinDavid: I believe I have done these correctly, I went ahead and added DeepCollectionEquality checks to the `queryParameters` and `fragment`s, as well as the `path`s and `state`s.

packages/go_router/lib/src/information_provider.dart
ChopinDavid: Same as above ⬆️, believe these have been fixed but let me know if you think otherwise.

packages/go_router/lib/src/information_provider.dart
```diff
@@ -137,17 +137,21 @@ class GoRouteInformationProvider extends RouteInformationProvider
...
   }
 
   void _setValue(String location, Object state) {
+    final Uri uri = Uri.parse(location);
+    const DeepCollectionEquality deepCollectionEquality =
+        DeepCollectionEquality();
     final bool shouldNotify =
-        _value.location != location || _value.state != state;
-    _value = RouteInformation(location: location, state: state);
+        !deepCollectionEquality.equals(_value.uri.path, uri.path) ||
```
hangyujin, line 149: code here is duplicated to above, extract it to a function

packages/go_router/lib/src/information_provider.dart
```diff
@@ -274,11 +278,4 @@ class GoRouteInformationProvider extends RouteInformationProvider
     _platformReportsNewRouteInformation(routeInformation);
     return SynchronousFuture<bool>(true);
   }
-
```
hangyujin, line 277: why remove this?

ChopinDavid: See [here](https://github.com/flutter/packages/pull/5742#issuecomment-1871364697).

[`didPushRoute` is now deprecated, and `didPushRouteInformation` should be used instead](https://github.com/flutter/flutter/blob/4721f96c3042281a80472047a4de06cd17755350/packages/flutter/lib/src/widgets/binding.dart#L84-L88).

hangyujin (approved):
LGTM, thank you!

ChopinDavid:
Anybody know why Linux repo_checks is failing? 😅

packages/go_router/pubspec.yaml
```diff
@@ -1,13 +1,13 @@
...
   deep linking, data-driven routes and more
-version: 13.0.1
+version: 13.0.2
 repository: https://github.com/flutter/packages/tree/main/packages/go_router
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+go_router%22
 
 environment:
   sdk: ">=3.0.0 <4.0.0"
-  flutter: ">=3.10.0"
+  flutter: ">=3.12.0"
```
hangyujin, line 10: revert the flutter version change

ChopinDavid: Should I revert the new changelog entry as well then?

ChopinDavid: I believe we need to be on flutter >= 3.12.0 as [`RouteInformation.uri` does not exist on flutter 3.10.0](https://github.com/flutter/flutter/blob/84a1e904f44f9b0e9c4510138010edcc653163f8/packages/flutter/lib/src/widgets/router.dart#L41-L74).

See also: [here](https://github.com/flutter/flutter/issues/103659#issuecomment-1861808022)

hangyujin: thanks for clarify this, looks like your test failed because the map here is not updated: https://github.com/flutter/packages/blob/f858453969324d13d4a98fc29a97f96d7c23df1b/script/tool/lib/src/common/core.dart#L61

ChopinDavid: Seems Linux repo_checks is still failing after upping min flutter version to 3.13.0. I am unable to see what is causing this step to fail.

hangyujin: the error messages are in checks-> linux repo_checks-> [View more details on flutter-dashboard] ->stdout 

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8758077883049129633/+/u/Run_package_tests/pubspec_validation/stdout

"  The minimum Dart version is 3.0.0, but the minimum Flutter version of 3.13.0 shipped with Dart 3.1.0. Please use consistent lower SDK bounds "

ChopinDavid: Ah, thanks, I'll try to get that fixed up.

ChopinDavid: Looks like everything's passing now, thanks @hangyujin

ChopinDavid:
It is worth mentioning, my team is finding an issue with iOS deeplinks not working from a terminated state. This PR should probably not be merged until we address this. More details to come. @hangyujin @johnpryan

ChopinDavid:
It seems that iOS deeplinks launched from a terminated app state sometimes do not work following this change. But it feels unclear to my team whether this PR introduced this issue. We are still trying to determine this, and feel free to see if you all can recreate this issue as well. I can provide more info about how to recreate this issue.

In the meantime, we are concerned about places in [`flutter/flutter`](https://github.com/flutter/flutter) where the Uri host/scheme are being stripped out:
 - [`WidgetsBindingObserver.didPushRouteInformation`](https://github.com/flutter/flutter/blob/1af09dde125c3fc1c4ec3012d463e1b2914307b2/packages/flutter/lib/src/widgets/binding.dart#L108-L119) 
 - [`_WidgetsAppState.didPushRouteInformation`](https://github.com/flutter/flutter/blob/1af09dde125c3fc1c4ec3012d463e1b2914307b2/packages/flutter/lib/src/widgets/app.dart#L1553-L1559)

cc: @chunhtai, you had [a TODO](https://github.com/flutter/flutter/blob/1af09dde125c3fc1c4ec3012d463e1b2914307b2/packages/flutter/lib/src/widgets/binding.dart#L106-L107) about updating this `didPushRouteInformation` method

ChopinDavid:
It seems that the issue we were seeing with iOS deep links from a terminated state did exist before any of this PR's changes were made. So this PR would not be introducing an issue there.

bawahakim:
FWIW, works well for me (at least on Android), and looking forward to having this in release, thanks @ChopinDavid ! 🚀

ZanderZhan:
Works well for me too, it's good having this in release as soon as possible, thanks @ChopinDavid ! 🚀

hangyujin:
@ChopinDavid  can you resolve the conflicts for this PR so it can be merged ?

hangyujin:
@johnpryan can you take a second look at this PR since chun-heng is on vacation?

ChopinDavid:
Hi guys, I will resolve this conflicts a bit later today. Thanks for the help everyone! 🙂🚀

ChopinDavid:
Looks like everything should be good to go now

johnpryan (approved):

PR #5312, opened 2023-11-03
[cupertino_icons] Add example to cupertino icons

foxtrotravi:
## Description of PR:

Adding an example app to showcase the usage of `cupertino_icons`. This addition will also increase the pub points of the package and will be helpful for other developers to quickly find relevant examples.

Fixes https://github.com/flutter/flutter/issues/137682

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #5775, opened 2024-01-02
[tool] Handle Flutter dev dependencies

stuartmorgan:
A non-Flutter package can have Flutter-based tests (e.g., cupertino_icons), in which case we need to use `flutter test` rather than `dart test` just like we would for a package with a non-dev Flutter dependency. This updates the `requiresFlutter` check to include dev dependencies as well as normal dependencies.


stuartmorgan:
See https://github.com/flutter/packages/pull/5312#discussion_r1426938508 for context/motivation.

tarrinneal (approved):

PR #5781, opened 2024-01-02
[video_player] Fix initial frame on macOS

stuartmorgan:
As with seeking while paused, initing a video and not playing it should show the first frame as soon as it is available, but it currently doesn't because the display link isn't running. This uses the same mechanism added for seek to ensure that a video reports a frame to the engine (thus populating the initially-blank textture) as soon as one is available after the player is created, even if it's not played.

Fixes https://github.com/flutter/flutter/issues/140782

## Pre-launch Checklist

- [x] I signed the [CLA].


cbracken (approved):
lgtm!

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
```diff
@@ -114,6 +114,10 @@ - (instancetype)initWithURL:(NSURL *)url
                 httpHeaders:(nonnull NSDictionary<NSString *, NSString *> *)headers
                   avFactory:(id<FVPAVFactory>)avFactory
                   registrar:(NSObject<FlutterPluginRegistrar> *)registrar;
+
+// Tells the player to run its frame updater until it receives a frame, regardless of the
+// play/pause state.
+- (void)expectFrame;
```
hellohuanlin, line 120: nit: maybe `startDisplayLinkIfNeeded`?

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
```diff
@@ -416,7 +420,7 @@ - (void)updatePlayingState {
   } else {
     [_player pause];
   }
-  _displayLink.running = _isPlaying;
+  _displayLink.running = _isPlaying || self.waitingForFrame;
```
hellohuanlin, line 423: can you also add a comment here about the paused case?

hellohuanlin (approved):

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
stuartmorgan: Done.

packages/video_player/video_player_avfoundation/darwin/Classes/FVPVideoPlayerPlugin.m
stuartmorgan: That would not be an accurate name; `startDisplayLinkIfNeeded` would just call `displayLink.running = true;`. If this were named by implementation, it would have to be called something like `startDisplayLinkIfNeededAndPreventStoppingUntilNextFrameAvailable`.

This is named by its conceptual purpose for callers, which is to instruct the player to expect that a frame will become available (and to handle state accordingly to make sure it's rendered), rather than its implementation.

PR #5785, opened 2024-01-02
[pointer_interceptor_web] Update package APIs and tests.

ditman:
## Changes:

* Updates the web implementation of the `pointer_interceptor` to `package:web`, and the latest `HtmlElementView` APIs available on Flutter.
* Resolves a timing issue in integration tests by waiting a few frames before doing any assertions from the DOM of a Flutter Web app, that should fix flutter rolls.

## Issues:

* Fixes https://github.com/flutter/flutter/issues/140834
* Fixes https://github.com/flutter/flutter/issues/139753
* Closes https://github.com/flutter/packages/pull/5673 (Credit to @balvinderz!)

## Tests

* Tested manually. All integration tests pass locally.

---

Co-authored-by: balvinderz <balvindersi2@gmail.com>

## Pre-launch Checklist

- [x] I signed the [CLA].


ditman:
_/cc @mdebbar migrated the package to the new `HtmlElementView.fromTagName` API!_

packages/pointer_interceptor/pointer_interceptor_web/example/integration_test/widget_test.dart
```diff
@@ -134,20 +112,29 @@ void main() {
...
 }
 
+Future<void> _fullyRenderApp(WidgetTester tester) async {
+  app.main();
+  // Because the app widget is not injected through the `tester`, we need to
+  // await a bunch of frames (~5) so the app is fully rendered, and the
+  // platform views injected in the DOM.
+  for (int i = 0; i < 5; i++) {
+    await tester.pump(const Duration(milliseconds: 16));
+  }
```
tarrinneal, line 132: Is there a less flaky way to wait for render? (I'm assuming no)

tarrinneal (approved):
Approved, might be worth waiting for @stuartmorgan to take a look though.

packages/pointer_interceptor/pointer_interceptor_web/example/integration_test/widget_test.dart
ditman: There's a `tester.pumpFrames(Widget, Duration time, [Duration interval])` but the internal implementation is just a for loop that does `tester.pump` `time/interval` times.

HOWEVER, now that I'm not using custom factories to render, I might be able to just `pumpWidget` with the app, and just pump a couple of frames directly! I'll give it a shot, thanks @tarrinneal!

ditman: (This seems to work. I've fixed the example app so we can pump its main Widget from the tester, instead of calling its "main", and now this just needs a couple of `pump` calls to get the platform view rendered into the DOM).

mdebbar (approved):
Looks good, thanks for the cleanup!

The only thing I would recommend is making `onElementCreated` a method so we don't keep creating a new closure on every build.

packages/pointer_interceptor/pointer_interceptor_web/example/integration_test/widget_test.dart
```diff
@@ -4,22 +4,21 @@
...
-import 'dart:html' as html;
-
 // Imports the Flutter Driver API.
 import 'package:flutter/src/widgets/framework.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 
 import 'package:pointer_interceptor_web_example/main.dart' as app;
 
+import 'package:web/web.dart' as web;
```
stuartmorgan, line 14: Nit: shouldn't this be up with the other packages? Usually we just have three blocks: `dart:`, `package:`, and path-based.

packages/pointer_interceptor/pointer_interceptor_web/example/lib/main.dart
```diff
@@ -2,50 +2,52 @@
...
-// ignore: avoid_web_libraries_in_flutter
-import 'dart:html' as html;
+import 'dart:js_interop';
 import 'dart:ui_web' as ui_web;
 
 import 'package:flutter/material.dart';
 import 'package:pointer_interceptor_platform_interface/pointer_interceptor_platform_interface.dart';
 import 'package:pointer_interceptor_web/pointer_interceptor_web.dart';
 
+import 'package:web/web.dart' as web;
```
stuartmorgan, line 12: Same here, I would not expect the newline above this.

packages/pointer_interceptor/pointer_interceptor_web/lib/pointer_interceptor_web.dart
```diff
@@ -2,78 +2,45 @@
...
-    final html.Element htmlElement = html.DivElement()
-      ..style.width = '100%'
-      ..style.height = '100%';
-    if (debug) {
-      htmlElement.style.backgroundColor = 'rgba(255, 0, 0, .5)';
-    }
-    return htmlElement;
-  }, isVisible: false);
-}
+import 'package:web/web.dart' as web;
```
stuartmorgan, line 10: Same.

stuartmorgan (approved):
LGTM

packages/pointer_interceptor/pointer_interceptor_web/example/integration_test/widget_test.dart
ditman: Yep, will tidy the imports and make the `onElementCreated` a tear-off as suggested by @mdebbar above.

stuartmorgan:
🎲

stuartmorgan:
> Rolling the 🎲 with `autosubmit`!

> 🎲

Maybe the problem is that the die keeps coming up "1" 🙂

ditman:
> Maybe the problem is that the die keeps coming up "1" 🙂

RFC 1149.5 specifies 4 as the standard IEEE-vetted random number[.](https://xkcd.com/221/)

tarrinneal:
My turn 🎲 🎲 🎲 🎲 🎲 🎲  (6)

ditman:
Removed bump to the master version, it seems other shards were breaking because of it.

PR #5786, opened 2024-01-03
[pigeon] java non null void

tarrinneal:
Fixes java requiring `Null` to be `NonNull`.

fixes: https://github.com/flutter/flutter/issues/139194

apparently also fixes an extra space that I removed a month ago and never merged.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/lib/java_generator.dart
```diff
@@ -449,7 +449,9 @@ class JavaGenerator extends StructuredGenerator<JavaOptions> {
 
       for (final Method func in api.methods) {
         final String resultType =
-            func.returnType.isNullable ? 'NullableResult' : 'Result';
+            func.returnType.isNullable || func.returnType.isVoid
+                ? 'NullableResult'
+                : 'Result';
```
stuartmorgan, line 454: We should probably extract a helper instead of repeating this non-trivial logic three times (and put a comment explaining why void is nullable).

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 15.0.3
```
stuartmorgan, line 1: Isn't this a breaking change for Java? I believe the failure is only when using the Java code with Kotlin, in which case many people will have currently working code with 15.x that will suddenly have the wrong type.

stuartmorgan (changes requested):
Why nullable `Void` rather than a `VoidResult` without an argument? If we need special-casing, the version that doesn't pass a useless argument seems more straightforward for clients.

stuartmorgan (approved):
LGTM

PR #5791, opened 2024-01-03
[web] Updates package:web dependency to ^0.5.0.

kevmoo:
This PR updates the packages that are using `package:web` to version `^0.5.0`.

Ancilliary changes:

* Bump `environment` to `flutter: ">=3.19.0"` and `sdk: ^3.3.0`.
* Bump version to next `Y`
* Clean-up code that was kept for compatibility with versions of `web: <0.5.0`.

The main exception to this is `package:google_sign_in_web`, which depends on a version of `google_identity_services_web` that has a dependency on package:web that is `<0.5.0`, so that package needs to have a range until `google_identity_services_web` gets published with the new ^0.5.0 dependency.

Co-Authored-By: David Iglesias<ditman@gmail.com>

### DIFF
--- a/packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
+++ b/packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
@@ -2,34 +2,102 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-/// Provides some useful tweaks to `package:web`.
-library package_web_tweaks;
+// ignore_for_file: public_member_api_docs
+
+@JS()
+library;
 
 import 'dart:js_interop';
+
 import 'package:web/web.dart' as web;
 
 /// This extension gives web.window a nullable getter to the `trustedTypes`
 /// property, which needs to be used to check for feature support.
 extension NullableTrustedTypesGetter on web.Window {
-  ///
   @JS('trustedTypes')
-  external web.TrustedTypePolicyFactory? get nullableTrustedTypes;
-}
+  external TrustedTypePolicyFactory? get nullableTrustedTypes;
 
-/// This extension allows a trusted type policy to create a script URL without
-/// the `args` parameter (which in Chrome currently fails).
-extension CreateScriptUrlWithoutArgs on web.TrustedTypePolicy {
-  ///
-  @JS('createScriptURL')
-  external web.TrustedScriptURL createScriptURLNoArgs(
-    String input,
-  );
+  external TrustedTypePolicyFactory get trustedTypes;
 }
 
 /// This extension allows setting a TrustedScriptURL as the src of a script element,
 /// which currently only accepts a string.
 extension TrustedTypeSrcAttribute on web.HTMLScriptElement {
-  ///
   @JS('src')
-  external set srcTT(web.TrustedScriptURL value);
+  external set srcTT(TrustedScriptURL value);
+}
+
+// TODO(kevmoo): drop all of this once https://github.com/dart-lang/web/pull/173
+// lands and pkg:web rolls to 0.5.1
+
+typedef CreateHTMLCallback = JSFunction;
+typedef CreateScriptCallback = JSFunction;
+typedef CreateScriptURLCallback = JSFunction;
+extension type TrustedHTML._(JSObject _) implements JSObject {
+  external static TrustedHTML fromLiteral(JSObject templateStringsArray);
+  external String toJSON();
+}
+extension type TrustedScript._(JSObject _) implements JSObject {
+  external static TrustedScript fromLiteral(JSObject templateStringsArray);
+  external String toJSON();
+}
+extension type TrustedScriptURL._(JSObject _) implements JSObject {
+  external static TrustedScriptURL fromLiteral(JSObject templateStringsArray);
+  external String toJSON();
+}
+extension type TrustedTypePolicyFactory._(JSObject _) implements JSObject {
+  external TrustedTypePolicy createPolicy(
+    String policyName, [
+    TrustedTypePolicyOptions policyOptions,
+  ]);
+  external bool isHTML(JSAny? value);
+  external bool isScript(JSAny? value);
+  external bool isScriptURL(JSAny? value);
+  external String? getAttributeType(
+    String tagName,
+    String attribute, [
+    String elementNs,
+    String attrNs,
+  ]);
+  external String? getPropertyType(
+    String tagName,
+    String property, [
+    String elementNs,
+  ]);
+  external TrustedHTML get emptyHTML;
+  external TrustedScript get emptyScript;
+  external TrustedTypePolicy? get defaultPolicy;
+}
+extension type TrustedTypePolicy._(JSObject _) implements JSObject {
+  external TrustedHTML createHTML(
+    String input,
+    JSAny? arguments,
+  );
+  external TrustedScript createScript(
+    String input,
+    JSAny? arguments,
+  );
+  external TrustedScriptURL createScriptURL(
+    String input,
+    JSAny? arguments,
+  );
+  @JS('createScriptURL')
+  external TrustedScriptURL createScriptURLNoArgs(
+    String input,
+  );
+  external String get name;
+}
+extension type TrustedTypePolicyOptions._(JSObject _) implements JSObject {
+  external factory TrustedTypePolicyOptions({
+    CreateHTMLCallback createHTML,
+    CreateScriptCallback createScript,
+    CreateScriptURLCallback createScriptURL,
+  });
+
+  external set createHTML(CreateHTMLCallback value);
+  external CreateHTMLCallback get createHTML;
+  external set createScript(CreateScriptCallback value);
+  external CreateScriptCallback get createScript;
+  external set createScriptURL(CreateScriptURLCallback value);
+  external CreateScriptURLCallback get createScriptURL;
 }
--- a/packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
+++ b/packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
@@ -11,13 +11,22 @@ import 'package:web/web.dart' as web;
 /// This extension gives web.window a nullable getter to the `trustedTypes`
 /// property, which needs to be used to check for feature support.
 extension NullableTrustedTypesGetter on web.Window {
-  /// Bindings to window.trustedTypes.
+  /// (Nullable) Bindings to window.trustedTypes.
   ///
   /// This may be null if the browser doesn't support the Trusted Types API.
   ///
   /// See: https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API
   @JS('trustedTypes')
-  external TrustedTypePolicyFactory? get trustedTypes;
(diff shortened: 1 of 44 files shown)


ditman:
I'll take over landing this one!

packages/web_benchmarks/lib/client.dart
```diff
@@ -320,6 +318,7 @@ class LocalBenchmarkServerClient {
   /// DevTools Protocol.
   Future<void> startPerformanceTracing(String? benchmarkName) async {
     _checkNotManualMode();
+    // ignore: deprecated_member_use
```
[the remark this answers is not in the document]
ditman: Removed, the rest of the occurrences are in `mock` or _test files, not in user code.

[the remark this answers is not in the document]
ditman:
```console
$ grep deprecated_member_use * -Irn | grep "_web" | grep "dart:" | cut -d: -f1 | sort | uniq -c | sort -r
      8 shared_preferences/shared_preferences_web/example/integration_test/shared_preferences_web_test.dart
      4 image_picker/image_picker_for_web/example/integration_test/image_picker_for_web_test.dart
      2 webview_flutter/webview_flutter_web/test/web_webview_controller_test.mocks.dart
      2 webview_flutter/webview_flutter_web/test/legacy/webview_flutter_web_test.mocks.dart
      2 google_sign_in/google_sign_in_web/example/integration_test/web_only_test.mocks.dart
      2 google_sign_in/google_sign_in_web/example/integration_test/google_sign_in_web_test.mocks.dart
      2 google_maps_flutter/google_maps_flutter_web/example/integration_test/overlays_test.mocks.dart
      2 google_maps_flutter/google_maps_flutter_web/example/integration_test/google_maps_plugin_test.mocks.dart
      2 google_maps_flutter/google_maps_flutter_web/example/integration_test/google_maps_controller_test.mocks.dart
```

Do we want to remove these from `_test.dart` files as well?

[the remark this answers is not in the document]
stuartmorgan: Autogenerated mocks are fine; the `ignore`s there are also auto-generated so not something we maintaine.

The two test files there are specific exceptions called out in the new gardener docs. They are annotating places where we have integration tests of our own deprecated methods (integration tests are technically a different bundle, so trip that check instead of `deprecated_member_use_from_same_package`), and we don't track those as tech debt since they are not actually cross-package, and will be removed in the same PR that removes the deprecated APIs themselves. Since we don't want to track them, we don't want them to show up in the manual audit, so we `ignore` them.

ditman:
```console
$ grep js_util * -Irn | grep "_web" | grep "dart:" | cut -d: -f1 | sort | uniq -c | sort -r
      5 camera/camera_web/example/integration_test/camera_service_test.dart
      4 camera/camera_web/lib/src/shims/dart_js_util.dart
      3 camera/camera_web/lib/src/camera_service.dart
      1 google_maps_flutter/google_maps_flutter_web/lib/google_maps_flutter_web.dart
      1 google_maps_flutter/google_maps_flutter_web/example/integration_test/google_maps_plugin_test.dart
      1 camera/camera_web/example/integration_test/helpers/mocks.dart
```

js_util seems to be almost gone from the repo (except from known non-migrated packages)

ditman (commented):
~~I think this is ready to land, PTAL @stuartmorgan!~~

packages/web_benchmarks/lib/client.dart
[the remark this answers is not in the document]
ditman: Ah, fantastic! I totally missed the gardener docs about those two files 😅

I'll mark this conversation as resolved then!

ditman:
I think this is OK now. PTAL @stuartmorgan (and @kevmoo)!

kevmoo:
I'd review it if I could! Thanks @ditman

packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
```diff
@@ -2,34 +2,75 @@
...
-
 import 'dart:js_interop';
+
 import 'package:web/web.dart' as web;
 
+// TODO(kevmoo): Make this file unnecessary, https://github.com/dart-lang/web/issues/175
+
 /// This extension gives web.window a nullable getter to the `trustedTypes`
 /// property, which needs to be used to check for feature support.
 extension NullableTrustedTypesGetter on web.Window {
```
stuartmorgan, line 13: Why are we using extensions? https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-using-extension

packages/google_sign_in/google_sign_in_web/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.12.4
+
+* Gets ready to update to web code to package `web: ^0.5.0`.
```
stuartmorgan, line 3: This seems like an odd description since it actually does the update.

(Also, nit: this package is all web code, so doesn't need to specify that here.)

packages/google_sign_in/google_sign_in_web/pubspec.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
   google_identity_services_web: ^0.3.0
   google_sign_in_platform_interface: ^2.4.0
   http: ">=0.13.0 <2.0.0"
-  web: ">=0.3.0 <0.5.0"
+  web: ">=0.3.0 <0.6.0" # because google_identity_services
```
stuartmorgan, line 28: Is this comment explaining the lower limit or the upper limit? More details would help.

packages/metrics_center/pubspec.yaml
```diff
@@ -12,7 +12,7 @@ dependencies:
   _discoveryapis_commons: ^1.0.0
   crypto: ^3.0.1
   gcloud: ^0.8.2
-  googleapis: ^3.0.0
+  googleapis: '>=3.0.0 <13.0.0'
```
stuartmorgan, line 15: Wow, this is Pigeon-level version revving 😁

packages/video_player/video_player_web/example/integration_test/utils.dart
```diff
@@ -22,35 +19,40 @@ String getUrlForAssetAsNetworkSource(String assetKey) {
       '?raw=true';
 }
 
-@JS()
-@anonymous
-class _Descriptor {
+extension type Descriptor._(JSObject _) implements JSObject {
```
stuartmorgan, line 22: Same question here.

packages/metrics_center/pubspec.yaml
kevmoo: We could honestly do `^12.0.0` I think.

stuartmorgan: Normally I'd say we should favor keeping the broad compatibility, but since no packages depend on `metrics_center` it probably doesn't matter either way.

packages/google_sign_in/google_sign_in_web/pubspec.yaml
kevmoo: I *think* this is because `google_identity_services` has not been upgraded yet...

packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
ditman: Extension types and extensions seem to be The New Way of doing JS-interop, rather than the `@staticInterop` annotation. See this: https://github.com/flutter/flutter/pull/143274 (I followed @kevmoo's lead here)

packages/google_sign_in/google_sign_in_web/pubspec.yaml
ditman: The comment explains the upper limit, this is the package I'll immediately update after landing this PR, so the upper limit will live there only for one small publish.

packages/video_player/video_player_web/example/integration_test/utils.dart
ditman: (Same answer as above)

packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
stuartmorgan: Okay, if Hixie is on board with this use case an an exception to the Flutter style guide that's fine.

kevmoo: Should we ask @Hixie directly?

stuartmorgan: Most of the file is the same extension type pattern that he just approved in the other PR, so I was taking that as confirmation that it should be fine.

stuartmorgan (approved):
LGTM modulo the changelog nits

packages/google_sign_in/google_sign_in_web/CHANGELOG.md
ditman: Agh shoot, I did a grammar pass for this description in other packages, but this one wasn't an exact match and I forgot. Thanks for flagging this!

packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart
ditman: > (This MAY be rewritable as extension type NullableTrustedTypesGetter._(web.Window) implements web.Window... (I'm going to try!))

OK, I tried:

```dart
extension type NullableTrustedTypesGetter._(web.Window _) implements web.Window {
```

instead of

```dart
extension NullableTrustedTypesGetter on web.Window {
```

And the compiler seems to take it, HOWEVER, when using the `web.window` instance of the `web.Window` class, my "wrapper" is not automagically applied, and I cannot access the nullable setter (because it should be of type `NullableTrustedTypesGetter`). This could be fixed by casting `as NullableTrustedTypesGetter` but... IMO that makes the code less legible.

So the rule of thumb seems to be:

* My custom JS-Interop types need to `extension type ... implements JSObject` (as described in the docs, this is [how `class Window` is defined in `package:web`](https://github.com/dart-lang/web/blob/975e55cbe4ac683c33d5256a91f4133fa97b4c34/lib/src/dom/html.dart#L2707-L2709))
* Existing JS-interop types that come from elsewhere and need some "tweaks" (like `web.Window` in this example) need to be a normal `extension ... on pkg.ForeignType`, so it gets applied automatically without us having to cast or doing anything to that foreign object.

(Of course, the rule of thumb will break down when we discover more complex use-cases, so take it with a grain of salt!)

packages/metrics_center/pubspec.yaml
ditman: Let's do it then!

packages/google_sign_in/google_sign_in_web/CHANGELOG.md
ditman: I also made this small dependency change a `+Z` change; I'll bump `+Y` when the migration of `package:web` actually happens.

packages/metrics_center/pubspec.yaml
ditman: Bumped to ^12.0.0, see what happens.

PR #5670, opened 2023-12-13
[webview_flutter] Implement platform interface for JavaScript dialog

jsharp83:
Adds the platform interface implementation for JavaScript dailog.

This PR is part of a series of PRs that aim to close https://github.com/flutter/flutter/issues/30358#issuecomment-1645347616
The PR that contains all changes can be found at https://github.com/flutter/packages/pull/4704

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #4704, opened 2023-08-15
[webview_flutter] Add interface for showing javascript dialog message

jsharp83:
* there are cases where Web calls System Popup with javascript on webview_flutter
* At this time, the message comes in the WKUIDelegate part in iOS.
   * https://developer.apple.com/documentation/webkit/wkuidelegate/1537406-webview
   * https://developer.apple.com/documentation/webkit/wkuidelegate/1536489-webview
* Android also has a interface on WebChromeClient
   * https://developer.android.com/reference/android/webkit/WebChromeClient#onJsAlert(android.webkit.WebView,%20java.lang.String,%20java.lang.String,%20android.webkit.JsResult)
* It was implemented according to the requirements of the code review of the https://github.com/flutter/packages/pull/4555
* Related issue: https://github.com/flutter/flutter/issues/30358#issuecomment-1645347616
* Related Interface PR: https://github.com/flutter/packages/pull/5670

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[...]

PR #5795, opened 2024-01-04
[webview_flutter_wkwebview] Add javascript panel interface for wkwebview

jsharp83:
* There are cases where Web calls System Popup with javascript on webview_flutter
* At this time, the message comes in the WKUIDelegate part in iOS.
   * https://developer.apple.com/documentation/webkit/wkuidelegate/1537406-webview
   * https://developer.apple.com/documentation/webkit/wkuidelegate/1536489-webview
* Related issue: https://github.com/flutter/flutter/issues/30358#issuecomment-1645347616
* Related Interface PR: https://github.com/flutter/packages/pull/5670
* The PR that contains all changes can be found at https://github.com/flutter/packages/pull/4704

## Pre-launch Checklist

- [x] I signed the [CLA].


hellohuanlin:
i briefly remembered that i have reviewed a similar PR before. Why are we opening a new PR?

jsharp83:
@hellohuanlin 

Previous PR is https://github.com/flutter/packages/pull/4704#issuecomment-1875633043
Since this PR includes implementations for multiple packages, I was requested to split it into separate PRs for each package. 
This PR only contains modifications related to wkwebview.

I invited other reviewers as you suggested. Thank you.

packages/webview_flutter/webview_flutter_wkwebview/example/lib/main.dart
```diff
@@ -605,6 +666,67 @@ class SampleMenu extends StatelessWidget {
...
+              TextButton(
+                  onPressed: () {
+                    Navigator.of(ctx).pop(true);
+                  },
+                  child: const Text('OK')),
+            ],
+          );
+        });
+
+    return result.runtimeType == bool && result as bool;
```
bparrishMines, line 708: I think this would be easier to read as `result is bool ? result : false`.

packages/webview_flutter/webview_flutter_wkwebview/example/lib/main.dart
```diff
@@ -605,6 +666,67 @@ class SampleMenu extends StatelessWidget {
...
+              TextButton(
+                  onPressed: () {
+                    Navigator.of(ctx).pop('Text test');
+                  },
+                  child: const Text('Enter')),
+            ],
+          );
+        });
+
+    return result.runtimeType == String ? result as String : '';
```
bparrishMines, line 728: This one could be `result is String ? result : ''`.

Although I just realized that the `prompt()` dialog can technically return a null String and should provide support for this according to Android docs: https://developer.android.com/reference/android/webkit/WebChromeClient#onJsPrompt(android.webkit.WebView,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20android.webkit.JsPromptResult). I'll actually go ahead and create a PR to update the interface method.

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFUIDelegateHostApi.m
```diff
@@ -131,6 +188,46 @@ - (void)webView:(WKWebView *)webView
...
+- (void)webView:(WKWebView *)webView
+    runJavaScriptAlertPanelWithMessage:(NSString *)message
+                      initiatedByFrame:(WKFrameInfo *)frame
+                     completionHandler:(void (^)(void))completionHandler {
+  [self.UIDelegateAPI runJavaScriptAlertPanelForDelegateWithIdentifier:self
+                                                               message:message
+                                                                 frame:frame
+                                                     completionHandler:^{
+                                                       completionHandler();
+                                                     }];
```
bparrishMines, line 201: I think this can be simplified to:

```suggestion
                                                     completionHandler:completionHandler];
```

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFUIDelegateHostApi.m
```diff
@@ -131,6 +188,46 @@ - (void)webView:(WKWebView *)webView
...
+- (void)webView:(WKWebView *)webView
+    runJavaScriptConfirmPanelWithMessage:(NSString *)message
+                        initiatedByFrame:(WKFrameInfo *)frame
+                       completionHandler:(void (^)(BOOL))completionHandler {
+  [self.UIDelegateAPI runJavaScriptConfirmPanelForDelegateWithIdentifier:self
+                                                                 message:message
+                                                                   frame:frame
+                                                       completionHandler:^(BOOL isConfirmed) {
+                                                         completionHandler(isConfirmed);
+                                                       }];
```
bparrishMines, line 213: Same

```suggestion
                                                       completionHandler:completionHandler];
```

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFUIDelegateHostApi.m
```diff
@@ -131,6 +188,46 @@ - (void)webView:(WKWebView *)webView
...
+                         initiatedByFrame:(WKFrameInfo *)frame
+                        completionHandler:(void (^)(NSString *_Nullable))completionHandler {
+  [self.UIDelegateAPI
+      runJavaScriptTextInputPanelForDelegateWithIdentifier:self
+                                                    prompt:prompt
+                                               defaultText:defaultText
+                                                     frame:frame
+                                         completionHandler:^(NSString *_Nullable inputMessage) {
+                                           completionHandler(inputMessage);
+                                         }];
```
bparrishMines, line 228: Same

```suggestion
                                         completionHandler:completionHandler];
```

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -780,11 +792,30 @@ class WKUIDelegate extends NSObject {
...
+      runJavaScriptAlertDialog;
+
+  /// Notifies the host application that the web page
+  /// wants to display a JavaScript confirm() dialog.
+  final Future<bool> Function(String message, WKFrameInfo frame)?
+      runJavaScriptConfirmDialog;
+
+  /// Notifies the host application that the web page
+  /// wants to display a JavaScript prompt() dialog.
+  final Future<String> Function(
```
bparrishMines, line 807: I think we should change this to be able to return a nullable value `Future<String?>`

packages/webview_flutter/webview_flutter_wkwebview/pigeons/web_kit.dart
```diff
@@ -805,6 +806,40 @@ abstract class WKUIDelegateFlutterApi {
...
+    String message,
+    WKFrameInfoData frame,
+  );
+
+  /// Callback to Dart function `WKUIDelegate.runJavaScriptTextInputPanel`.
+  @ObjCSelector(
+    'runJavaScriptTextInputPanelForDelegateWithIdentifier:prompt:defaultText:frame:',
+  )
+  @async
+  String runJavaScriptTextInputPanel(
```
bparrishMines, line 837: Can you change this to
```suggestion
  String? runJavaScriptTextInputPanel(
```
and regenerate the code.

bparrishMines (changes requested):

packages/webview_flutter/webview_flutter_wkwebview/example/lib/main.dart
jsharp83: @bparrishMines I just remove the result of dynamic type on https://github.com/flutter/packages/pull/5795/commits/aebf9483de36ece51334484119904f8bc92a73c2

packages/webview_flutter/webview_flutter_wkwebview/example/lib/main.dart
jsharp83: @bparrishMines 

I agree with your opinion. but one thing I worried that confirm() dialog also should be handle null case with this opinion.

I just thought that dismissing dialog without selection should be false in confirm and empty string in prompt.
But if we handle it as optional null, confirm dialog also return null when we just dismiss dialog.

bparrishMines: For `confirm()`, I considered the same thing.  However, I think canceling the operation is the same as returning false. 

`prompt()` is slightly different because there could be scenario where a user might want to return an empty string. And this could be considered different from a user that wants to cancel the operation.

packages/webview_flutter/webview_flutter_wkwebview/pigeons/web_kit.dart
bparrishMines: Actually since the platform interface change will require a breaking change: https://github.com/flutter/packages/pull/5901. I'll update this in another PR as well.

bparrishMines (approved):
LGTM

cc @stuartmorgan or @hellohuanlin for secondary review. 

I'll update the version on approval

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFDataConverters.m
```diff
@@ -167,7 +167,7 @@ WKAudiovisualMediaTypes FWFNativeWKAudiovisualMediaTypeFromEnumData(
 
 FWFNSUrlRequestData *FWFNSUrlRequestDataFromNativeNSURLRequest(NSURLRequest *request) {
   return [FWFNSUrlRequestData
-              makeWithUrl:request.URL.absoluteString
+              makeWithUrl:request.URL.absoluteString.length > 0 ? request.URL.absoluteString : @""
```
hellohuanlin, line 170: im a bit confused here - will absoluteString ever be `nil`? if so, may be less confusing to do nil check rather than length check?

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFUIDelegateHostApi.m
```diff
@@ -88,6 +88,63 @@ - (void)requestMediaCapturePermissionForDelegateWithIdentifier:(FWFUIDelegate *)
...
+- (void)runJavaScriptAlertPanelForDelegateWithIdentifier:(FWFUIDelegate *)instance
+                                                 message:(NSString *)message
+                                                   frame:(WKFrameInfo *)frame
+                                       completionHandler:(void (^)(void))completionHandler {
+  [self runJavaScriptAlertPanelForDelegateWithIdentifier:[self identifierForDelegate:instance]
+                                                 message:message
+                                                   frame:FWFWKFrameInfoDataFromNativeWKFrameInfo(
+                                                             frame)
+                                              completion:^(FlutterError *error) {
+                                                NSAssert(!error, @"%@", error);
```
hellohuanlin, line 101: when will error happen? should we pipe the error to dark side rather than doing assert here?

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFDataConverters.m
jsharp83: https://github.com/flutter/packages/blob/83c2c4da97c554996c488037c7ca7c1d5a076a9b/packages/webview_flutter/webview_flutter_wkwebview/example/integration_test/legacy/webview_flutter_test.dart#L1124-L1149

@hellohuanlin 
It can be nil with this test case.
I updated code to nil check.
Thanks

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFUIDelegateHostApi.m
jsharp83: https://github.com/flutter/packages/blob/83c2c4da97c554996c488037c7ca7c1d5a076a9b/packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFUIDelegateHostApi.m#L34-L91

@hellohuanlin In this case, I just followed previous implementation. I also agree with that NSAssert can be dangerous somethime, but the error only came from pigeon generation part like follow code and I understood that it is serious situation if channel is not established.

https://github.com/flutter/packages/blob/83c2c4da97c554996c488037c7ca7c1d5a076a9b/packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFGeneratedWebKitApis.m#L3094-L3098

hellohuanlin (approved):

packages/webview_flutter/webview_flutter_wkwebview/ios/Classes/FWFUIDelegateHostApi.m
bparrishMines: We are still deciding on a better alternative for using `NSAssert` in callback methods: https://github.com/flutter/flutter/issues/140410

@jsharp83 is correct that we probably won't be able to send the error back to Dart, so we need a way to handle it on the platform side.

Definitely feel free to add feedback to this issue!

PR #5801, opened 2024-01-04
Update platform label rules for shared iOS/macOS

stuartmorgan:
Update automatic platform label rules for more implementations that are shared between iOS and macOS.


jmagman (approved):

cbracken (approved):
lgtm!

PR #5805, opened 2024-01-04
[file_selector_android] Attempt to close system dialogs before integration tests run

bparrishMines:
Attempt to close any system dialogs before running the test. This idea came from https://stackoverflow.com/questions/39457305/android-testing-waited-for-the-root-of-the-view-hierarchy-to-have-window-focus. 

There may be a system dialog before the test runs that prevents the FlutterView from getting focus. Which also prevents any of the espresso actions to run

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Can we temporarily add the roll into this PR, to test that it actually fixes the issue we're having?

stuartmorgan (commented):
Changes LGTM assuming they bear out in tests.

If this is it, we should also file an infra issue about the fact that the Android emulators suddenly started failing tests due to the presence of a system dialog, since that seems like a fundamental issue that we shouldn't have to work around in each test.

stuartmorgan (approved):

bparrishMines:
> Changes LGTM assuming they bear out in tests.
> 
> If this is it, we should also file an infra issue about the fact that the Android emulators suddenly started failing tests due to the presence of a system dialog, since that seems like a fundamental issue that we shouldn't have to work around in each test.

I'm not sure if this is a problem that can be fixed by our infra. The stackoverflow link shows that it is also a problem with non-Flutter Android integration tests. It may just be related to Android 34 emulators since I don't remember it being flaky before.

gmackall (approved):

tarrinneal (approved):
another domino falls!

gmackall:
It looks like 2 of the 4 checks the packages tree are red on are due to this error - should we land this on red or does the fix for the other flakes need to land first?

tarrinneal:
merging on red to fix tree

PR #5811, opened 2024-01-05
[ci] Check repo-level package metadata

stuartmorgan:
Adds a new tool command (and runs it in CI) to check that each package:
- is listed correctly in the repo-level README.md table
- has a CODEOWNERS entry

In the future we could add other things (e.g., auto-label), but these were the main things we've had issues with recently.

Updates README.md and CODEOWNERS to fix failures it found:
- Adds a couple of missing CODEOWNERS
- Expands the web implementation CODEOWNERS to individual packages so that we don't have to special-case handling in the tool
- Fixes some minor mistakes in README.md
- URL-encodes all `:`s in the README.md links (which is why ever line shows as changed); it worked without that in practice, but it should really be encoded, and having it consistently encoded made things easier for the tooling.


script/tool/lib/src/repo_package_info_check_command.dart
```diff
@@ -0,0 +1,205 @@
...
+
+import 'package:file/file.dart';
+
+import 'common/core.dart';
+import 'common/output_utils.dart';
+import 'common/package_looping_command.dart';
+import 'common/repository_package.dart';
+
+const int _exitBadTableEntry = 3;
+const int _exitUknownPackageEntry = 4;
```
tarrinneal, line 13: typo: unknown

tarrinneal (approved):
looks good besides this typo

script/tool/lib/src/repo_package_info_check_command.dart
stuartmorgan: Fixed.

PR #5846, opened 2024-01-08
[various] Add iOS privacy manifests

stuartmorgan:
Adds privacy manifests to all iOS plugins.

While we only *need* to do the plugins listed [here](https://developer.apple.com/support/third-party-SDK-requirements/) for now, the wording of the page:
> The following are commonly used SDKs in apps on the App Store

suggests that the list of things for which this is required is just an arbitrary cutoff rather than a conceptual distinction, so it seems safest to just assume the list will grow over time and do all of them. To ensure that, this includes new repo tooling to check that a manifest is specified in the podspec.

The large caveat is that we do not currently know if this actually works. This is the method of inclusion that seems to be [the consensus among people using Cocoapods](https://github.com/CocoaPods/CocoaPods/issues/10325), as bundling it directly as a `resource` causes problems for clients who do not use `use_frameworks`. (In theory it seems like a manifest would not actually be *required* in that case since there is no framework, but it has the potential to actually stomp top-level resources.) Hopefully the automated analysis that Apple will eventually roll out will tolerate the file being bundled in a resource bundle in the framework rather than a top-level manifest file. If not, however, it's not clear how Cocoapods can be supported, so we can adopt this common approach for now under the assumption that eventually tooling will adapt to the reality of the ecosystem, and revisit the exact bundling later if necessary.

Only `shared_preferences` has a non-empty manifest, as it is our only plugin that uses a required reason API, and none of our plugins themselves collect private data. Ideally for that plugin we would instead use `C56D.1`, which is for wrappers, but as currently written we can't use it since it's exclusively a wrapper. If that changes in the future based on our pending request, we can revisit. For now, however, this reason should suffice since we don't currently allow reading from other app groups.

Fixes https://github.com/flutter/flutter/issues/131495
Fixes https://github.com/flutter/flutter/issues/139756
Fixes https://github.com/flutter/flutter/issues/139757
Fixes https://github.com/flutter/flutter/issues/139758
Fixes https://github.com/flutter/flutter/issues/139759
Fixes https://github.com/flutter/flutter/issues/139760
See also https://github.com/flutter/flutter/issues/139761

## Pre-launch Checklist

- [x] I signed the [CLA].


jmagman (approved):
LGTM

omarzl:
Consider setting a different name for the resource bundle name since Cocoapods will generate the same `bundle id` for the framework (if linked dynamically) and for the resource bundle.

For example:
A framework `camera_avfoundation.framework` would have the bundle id `org.cocoapods.camera_avfoundation` in its `Info.plist`
and the resource bundle `camera_avfoundation.bundle` will have the same one `org.cocoapods.camera_avfoundation`

We have had issues in the past with this id collision (like images not been displayed) since iOS loads the bundles and registers them with their id, not the filename nor file location.

Example of suggested changes:
```
 s.name             = 'camera_avfoundation'
 s.resource_bundles = {'camera_avfoundation_resources' => ['Resources/PrivacyInfo.xcprivacy']}
```

stuartmorgan:
> Consider setting a different name for the resource bundle name since Cocoapods will generate the same `bundle id` for the framework (if linked dynamically) and for the resource bundle.

Thanks for flagging that. It's unfortunate that the Cocoapod docs don't mention that when they talk about how to name the bundles to avoid collisions :|

stuartmorgan:
> If I should guess, in this scenario it won't cause any issue. But if in the future you need to include any other asset in those resource bundles and access them using `Bundle(identifier:)` it could cause a problem

I don't think we can rely on it not causing problems even as-is; even if we're not trying to load anything from the resource bundle, it's possible that the resource bundle could actually stomp the framework bundle for some lower-level lookup purpose.

tarrinneal (approved):

packages/image_picker/image_picker_ios/ios/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array/>
+	<key>NSPrivacyCollectedDataTypes</key>
```
vashworth, line 9: > Only shared_preferences has a non-empty manifest, as it is our only plugin that uses a required reason API, and none of our plugins themselves collect private data. 

Wouldn't plugin like image_picker_ios be using `NSPrivacyCollectedDataTypePhotosorVideos` for collecting "The user's photos or videos"?

stuartmorgan: The "Describing data use in privacy manifests" page never defines the term "collect", so currently we're assuming it's the same as the definition Apple gives [for existing privacy declarations](https://developer.apple.com/app-store/app-privacy-details/#data-collection), which is:

> “Collect” refers to transmitting data off the device in a way that allows you and/or your third-party partners to access it for a period longer than what is necessary to service the transmitted request in real time.

Our plugin does not collect any data under this definition. Plugin *clients* may or may not (under this definition) collect data using the plugin, but if they are doing such collection they would need to declare that at the app level since that's where the collection would be.

vashworth: I see, okay makes sense. Thanks for clarifying.

vashworth (approved):
LGTM

PR #5856, opened 2024-01-10
[multicast_dns] Allow decode of binary encoded TXT

ajhemphill91:
Follow up on https://github.com/flutter/flutter/issues/141218. Added a unit test with a binary-encoded payload.

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/multicast_dns/lib/src/resource_record.dart
+++ b/packages/multicast_dns/lib/src/resource_record.dart
@@ -338,11 +338,15 @@ class TxtResourceRecord extends ResourceRecord {
   const TxtResourceRecord(
     String name,
     int validUntil, {
-    required this.text,
+    required this.attrs,
   }) : super(ResourceRecordType.text, name, validUntil);
 
-  /// The raw text from this record.
-  final String text;
+  /// List of name/value pairs in the TXT record.
+  final List<String> attrs;
+
+  /// Text of this record combined with newlines. Keep in mind binary-encoded values may legally contain newlines,
+  /// and as such this property should not be used to parse the record.
+  String get text => attrs.join('\n');
 
   @override
   String get _additionalInfo => 'text: $text';
--- a/packages/multicast_dns/lib/src/packet.dart
+++ b/packages/multicast_dns/lib/src/packet.dart
@@ -329,9 +329,8 @@ List<ResourceRecord>? decodeMDnsResponse(List<int> packet) {
       case ResourceRecordType.text:
         checkLength(offset + readDataLength);
         // The first byte of the buffer is the length of the first string of
-        // the TXT record. Further length-prefixed strings may follow. We
-        // concatenate them with newlines.
-        final StringBuffer strings = StringBuffer();
+        // the TXT record. Further length-prefixed strings may follow.
+        final List<String> attrs = <String>[];
         int index = 0;
         while (index < readDataLength) {
           final int txtLength = data[offset + index];
@@ -343,11 +342,11 @@ List<ResourceRecord>? decodeMDnsResponse(List<int> packet) {
             Uint8List.view(data.buffer, offset + index, txtLength),
             allowMalformed: true,
           );
-          strings.writeln(text);
+          attrs.add(text);
           index += txtLength;
         }
         offset += readDataLength;
-        return TxtResourceRecord(fqdn, validUntil, text: strings.toString());
+        return TxtResourceRecord(fqdn, validUntil, attrs: attrs);
       default:
         checkLength(offset + readDataLength);
         offset += readDataLength;
--- a/packages/multicast_dns/test/decode_test.dart
+++ b/packages/multicast_dns/test/decode_test.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:multicast_dns/src/packet.dart';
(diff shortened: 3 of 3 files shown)


ajhemphill91:
Side note, the decoding algorithm for TXT data is still not exactly in spec (see [RFC6763 6.3-6.5](https://datatracker.ietf.org/doc/html/rfc6763#section-6.3)). 

In particular:

> DNS-SD TXT record strings
   beginning with an '=' character (i.e., the key is missing) MUST be
   silently ignored.

> The characters of a key MUST be printable US-ASCII values (0x20-0x7E)
   [[RFC20](https://datatracker.ietf.org/doc/html/rfc20)], excluding '=' (0x3D).

These don't matter to me too much at the present time. It does look like the openthread border router's [implementation](https://github.com/openthread/ot-br-posix/blob/897028fe9ef1b7604d8911f1c31b048d70c15c8a/src/mdns/mdns.cpp#L132) has these restrictions on the '=' character within the key.

stuartmorgan:
From triage: ping @dnfield on this review

packages/multicast_dns/lib/src/resource_record.dart
```diff
@@ -338,11 +338,15 @@ class TxtResourceRecord extends ResourceRecord {
   const TxtResourceRecord(
     String name,
     int validUntil, {
-    required this.text,
+    required this.attrs,
```
jmagman, line 341: Apologies for the delay in this review.

I believe this would be a breaking change to [TxtResourceRecord](https://pub.dev/documentation/multicast_dns/latest/multicast_dns/TxtResourceRecord/TxtResourceRecord.html)?

How about a new named constructor:

```dart
  TxtResourceRecord.attributes(
    String name,
    int validUntil, {
    required List<String> attributes,
  }) : this(name, validUntil, text: attributes.join('\n'));
```
Also total nit but `attributes` instead of `attrs`?

jmagman:
Hi @ajhemphill91, have you been able to circle back to the requested change for the named constructor?  Appreciate your contribution!

ajhemphill91:
Sorry for the delay, I have been on another project recently. If you do have a preference to keep the original constructor (and thus the original ```final String text``` property) for API stability, your suggested changes sound good. Internally, this constructor is only referenced once in packet.dart and then in the unit tests which are updated by this commit. 

The ```text``` property as a String is not valid since it represents a list of [encoded strings](http://www.zeroconf.org/Rendezvous/txtrecords.html) which [are treated as binary information](https://datatracker.ietf.org/doc/html/rfc1035#section-3.3). Basically it mutates the record by adding the ```\n``` characters.

Actually now that I am looking at it, ```encodeResponseRecord``` ought to fail to produce correct results too since the record is mutated by the ```text``` property. However it looks as though this property is never referenced anywhere.

stuartmorgan:
From triage: @ajhemphill91 Are you still planning on updating this PR?

ajhemphill91:
I can apply the nit, but changing back to a named constructor that joins the records with a newline nullifies this PR in the first place. What would you like me to do?

jmagman:
Sorry, didn't realize this PR was waiting for my response.

> but changing back to a named constructor that joins the records with a newline nullifies this PR in the first place


Hm, yeah I can't think of a way to do this without something unpleasant like a `text.split('\n')`
```dart
  TxtResourceRecord(
    String name,
    int validUntil, {
    required this.text,
  })  : attrs = text.split('\n'),
        super(ResourceRecordType.text, name, validUntil);

  TxtResourceRecord.attributes(
    String name,
    int validUntil, {
    required this.attrs,
  })  : text = attrs.join('\n'),
        super(ResourceRecordType.text, name, validUntil);
```

Adding @cbracken for his opinion here.  Ideally this could be done without a breaking change.

stuartmorgan:
@ajhemphill91 I've re-added the checklist from the template to the PR description; in the future please don't remove it. This PR is missing several required steps (notably for this discussion, a version change).

ajhemphill91:
Per the [versioning guide](https://dart.dev/tools/pub/versioning) should I then update the pubspec.yaml version to 0.4.0? I can take care of that and the CHANGELOG.md if everyone is okay with the breaking change.

stuartmorgan:
@jmagman Are you okay with the breaking change?

jmagman:
> @jmagman Are you okay with the breaking change?

LGTM, I can't remember if this will impact the tool but it would be easy to fix up

packages/multicast_dns/lib/src/resource_record.dart
cbracken: A way to do this without it being a breaking change would be to:
1. Make `text` nullable and mark it deprecated.
2. Add `attrs` as nullable.
3. Throw if (a) both or (b) neither are null.

After the deprecation period, mark `attrs` as non-nullable and remove `text`.

If very few packages are relying on this, then it might not be worth the effort.

cbracken (approved):
This seems fine to me. Suggestion on a non-breaking way to do this in my comment above.

cbracken:
@ajhemphill91 Looks like Dart formatting and CHANGELOG format checks are failing. See the log here for details:
https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20repo_checks/4671/overview

Once those are done, I leave it up to you as to whether you want to do the multi-step dance to go through the deprecation, then breaking change removal, or just go straight to the breaking change and land this as-is.

stuartmorgan:
@ajhemphill91 Is this still something you are planning on updating?

stuartmorgan:

PR #5872, opened 2024-01-11
[video_player] Skip flaky tests in certain circumstances on Android

stuartmorgan:
Temporary workaround for https://github.com/flutter/flutter/issues/141145

This doesn't fully skip the test because this problem doesn't happen in CI device tests or locally, so rather than totally lose coverage this detects the very specific markers of this flake and skips only when that happens. This should allow the tree to stop being so red with the minimal coverage loss while we investigate this further.

Part of https://github.com/flutter/flutter/issues/141145

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player/example/integration_test/video_player_test.dart
+++ b/packages/video_player/video_player/example/integration_test/video_player_test.dart
@@ -134,23 +134,17 @@ void main() {
         await controller.play();
         await tester.pumpAndSettle(_playDuration);
         // Android emulators in our CI have frequent flake where the video
-        // reports as still playing, but doesn't advance at all; if that
+        // reports as still playing (usually without having advanced at all
+        // past the seek position, but sometimes having advanced some); if that
         // happens, the thing being tested hasn't even had a chance to happen
         // due to CI issues, so just report it as skipped.
         // TODO(stuartmorgan): Remove once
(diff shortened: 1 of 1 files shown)


stuartmorgan:
Looks like this isn't working; need to experiment more to see why. Things that don't repro locally are the worst.

stuartmorgan:
> Looks like this isn't working; need to experiment more to see why

Captured what I found [here](https://github.com/flutter/flutter/issues/141145#issuecomment-1887909829) and relaxed the check accordingly. This should work now; it's slightly less targeted unfortunately.

stuartmorgan:
@reidbaker This should be working now, so it ready for review.

reidbaker (approved):
Another option would be to try to call the dismiss dialog call in this situation instead of skipping.

stuartmorgan:
> Another option would be to try to call the dismiss dialog call in this situation instead of skipping.

The reason I didn't do that is that unlike the other cases, this is a Dart integration test, which means it's running in the actual example app, with no special native code. So to dismiss the dialog (if we even can—it wasn't 100% clear to me from initial research if we could do from a normal app with a special permission set in the app, or if it was now fully locked down) we would need to build native code to do that, and a method channel to call it from Dart, into the example app, which is a lot more code to write for what I hope will be a temporary hack.

PR #5884, opened 2024-01-12
[video_player] Windows Support

azchohfi:
Creates a video_player_windows package that uses Windows' Media Foundation APIs to deliver a decent video player on windows.

Fixes: https://github.com/flutter/flutter/issues/37673

## Pre-launch Checklist

- [X] I signed the [CLA].


packages/video_player/video_player_windows/AUTHORS
stuartmorgan, line 1: New packages don't need to copy the existing AUTHORS, they only need to include the actual authors of the new package. (A bunch of packages have the mega-list because we didn't used to do per-package AUTHORS, and so when we started doing it we seeded them from the mega-list rather than doing a ton of work to figure out which ones should be in each package).

packages/video_player/video_player_windows/README.md
```diff
@@ -0,0 +1,15 @@
+# video\_player\_windows
+
+The Windows implementation of [`video_player`][1].
+
+## Usage
+
+This package is [endorsed][2], which means you can simply use `video_player`
```
stuartmorgan, line 7: Do you think this is ready enough for prime-time that we'll want to endorse it now, or still more alpha (in which case we could hold off on endorsing it, and ask people to testing it out by depending on it directly). We don't have a bright line on this, it's mostly about whether it's basically feature complete.

packages/video_player/video_player_windows/example/.gitignore
```diff
@@ -0,0 +1 @@
+lib/generated_plugin_registrant.dart
```
stuartmorgan, line 1: This should never exist in this package AFAIK; was this something you actually saw?

packages/video_player/video_player_windows/pigeons/messages.dart
```diff
@@ -0,0 +1,73 @@
...
+  dartOut: 'lib/src/messages.g.dart',
+  dartTestOut: 'test/test_api.g.dart',
+  cppHeaderOut: 'windows/messages.h',
+  cppSourceOut: 'windows/messages.cpp',
+  cppOptions: CppOptions(
+    namespace: 'Messages',
+  ),
+  copyrightHeader: 'pigeons/copyright.txt',
+))
+class TextureMessage {
```
stuartmorgan, line 17: It's not critical, but since this is a new implementation I would rather not carry this legacy style over. `video_player` was the first real testbed for Pigeon, and this approach of having a message object for everything predates Pigeon's ability to have multiple and non-class arguments. It should be trivial to just inline all of the fields as arguments now, and not need the classes.

packages/video_player/video_player_windows/pubspec.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+dependencies:
+  flutter:
+    sdk: flutter
+  video_player_platform_interface: ">=6.1.0 <7.0.0"
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  pigeon: ^9.2.5
```
stuartmorgan, line 27: Let's use 16. I would expect not too many changes for Windows code, but there have been some improvements.

packages/video_player/video_player_windows/windows/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+target_compile_definitions(${PLUGIN_NAME} PRIVATE FLUTTER_PLUGIN_IMPL)
+target_include_directories(${PLUGIN_NAME} INTERFACE
+  "${CMAKE_CURRENT_SOURCE_DIR}/include")
+target_link_libraries(${PLUGIN_NAME} PRIVATE flutter flutter_wrapper_plugin)
+
+# List of absolute paths to libraries that should be bundled with the plugin
+set(video_player_bundled_libraries
+  ""
+  PARENT_SCOPE
+)
```
stuartmorgan, line 67: We'll need native unit test structures set up for the plugin; see https://github.com/flutter/packages/blob/14b94ff834c06dbf9220e0066657806c6c7d552c/packages/url_launcher/url_launcher_windows/windows/CMakeLists.txt#L37-L75 and https://github.com/flutter/packages/blob/main/packages/url_launcher/url_launcher_windows/example/windows/CMakeLists.txt#L49-L52

stuartmorgan (commented):
Very exciting!

I started to review but ran out of time for today after a few high level notes. Publishing those now so they don't have to wait for the rest of the review, but I'll try to do the real review ASAP since I'd love to see Windows support for `video_player` landed.

packages/video_player/video_player_windows/README.md
azchohfi: Do we have a list of features that would define a minimal viable product for endorsement? I'm pretty sure this is going to support everything that ExoPlayer supports for Android and iOS. There are 3 things on my mental todo list that still need to be "fixed", and/or at desired features:
1) V-Sync (still seeing some artifacts when screen is large enough), needs to be fixed before merging this.
2) m3u8 files. I know that Media Foundation supports this, but it is failing on the existing code path, so I removed the test that was validating it.
3) DRM content (not sure if this is a must, but should be possible at least when Composition is enabled and available, AFAIK)

My C++ is not the best, so I know this will have some level of mistakes/bugs, but once the PR is reviewed properly merged, I don't see it being less feature complete than the Android or iOS implementations.

packages/video_player/video_player_windows/example/.gitignore
azchohfi: Nope, just a copy/paste mistake. I can remove it.

packages/video_player/video_player_windows/AUTHORS
azchohfi: Done.

packages/video_player/video_player_windows/pigeons/messages.dart
azchohfi: Oh yeah, absolutely! This simplified the pigeon file a LOT.

packages/video_player/video_player_windows/pubspec.yaml
azchohfi: Done.

packages/video_player/video_player_windows/README.md
stuartmorgan: > Do we have a list of features that would define a minimal viable product for endorsement? I'm pretty sure this is going to support everything that ExoPlayer supports for Android and iOS.

Sounds feature-complete for endorsement to me :)

> 2. m3u8 files. I know that Media Foundation supports this, but it is failing on the existing code path, so I removed the test that was validating it.

This is definitely something we should track in an issue (I'm sure people will hit it), but I would not block endorsement on it.

> 3. DRM content (not sure if this is a must, but should be possible at least when Composition is enabled and available, AFAIK)

This doesn't work on any platform currently, since we are using textures rather than platform views, and DRM systems generally don't want to write frames into arbitrary buffers. (Eventually we will likely implement a platform-view-based alternative for DRM playback, but that's not something that would block this; based on various discussions our current expectation is that long term every platform will need both texture-based and platform-view-based playback options due to the tradeoffs inherent in platform views.)

packages/video_player/video_player_windows/windows/CMakeLists.txt
azchohfi: Done.

azchohfi:
When I updated to 16.0.0 of Pigeon, found a bug: https://github.com/flutter/flutter/issues/141499

stuartmorgan:
> When I updated to 16.0.0 of Pigeon, found a bug: [flutter/flutter#141499](https://github.com/flutter/flutter/issues/141499)

Thanks, that should probably be an easy fix.

FYI we're generally moving away from `dartHostTestHandler` in favor of just mocking/faking the host API itself, as we've generally found over time that it makes the tests simpler.

azchohfi:
@stuartmorgan I have no idea why the tests are timing out. They are working fine locally.

tarrinneal (commented):
Just a little nit from me. 

I think this generally looks great! I also don't have a lot of experience with c++ so I can't speak to best practices there.

packages/video_player/video_player_windows/windows/test/video_player_windows_test.cpp
stuartmorgan, line 1: I would also expect that we would need a `video_player_test.cpp` with at least a few unit tests of `VideoPlayer` that interact with fake versions of the underlying objects, so that we have a structure for doing substantive tests of the wrapper functionality.

packages/video_player/video_player_windows/windows/test/mocks.h
```diff
@@ -0,0 +1,107 @@
...
+  MockTextureRegistrar() {
+    ON_CALL(*this, RegisterTexture)
+        .WillByDefault([this](flutter::TextureVariant* texture) -> int64_t {
+          EXPECT_TRUE(texture);
+          this->texture_ = texture;
+          this->texture_id_ = 1000;
+          return this->texture_id_;
+        });
+
+    // Deprecated pre-Flutter-3.4 version.
```
stuartmorgan, line 53: We aren't supporting any versions of Flutter before 3.4 in this plugin, so there's cruft here that can be removed.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,254 @@
...
+          },
+          [this](const flutter::EncodableValue* arguments)
+              -> std::unique_ptr<
+                  flutter::StreamHandlerError<flutter::EncodableValue>> {
+            this->_eventSink = nullptr;
+            return nullptr;
+          }));
+}
+
+VideoPlayer::~VideoPlayer() { m_valid = false; }
```
stuartmorgan, line 94: Why do we need to change an ivar in the destructor? `m_valid` won't be a valid memory location as soon as this call completes.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+bool VideoPlayer::IsValid() { return m_valid; }
```
stuartmorgan, line 96: It looks like this is only ever false after the destructor runs; calling a method on a destructed object has undefined behavior, so if that's the case this method can never do anything deterministic.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,254 @@
...
+  auto lock = m_compositionLock.lock();
+
+  RECT rect;
+  if (GetWindowRect(m_window, &rect)) {
+    int width = rect.right - rect.left;
+    int height = rect.bottom - rect.top;
+    m_windowSize = {static_cast<float>(width), static_cast<float>(height)};
+  }
+
+  // If the window has not been initialized yet, use a default size of 640x480
```
stuartmorgan, line 131: Nit: comments all need to end with periods, per the style guide.

stuartmorgan (changes requested):
I did a fairly high-level review, mostly about style guide notes, test structure, exceptions, Pigeon boundary. I haven't reviewed the specifics of the video implementation; I'll leave that to @cbracken since he'll most need to be familiar with it.

I haven't had a chance to play with it on my Windows machine yet, but I'm very much looking forward to seeing it in action :)

packages/video_player/video_player_windows/windows/MediaEngineExtension.cpp
```diff
@@ -0,0 +1,114 @@
...
+    m_uriType = ExtensionUriType::CustomSource;
+    // Invoke the callback synchronously since no outstanding work is required.
+    THROW_IF_FAILED(callback->Invoke(asyncResult.get()));
+  } else {
+    THROW_HR(MF_E_UNEXPECTED);
+  }
+
+  return S_OK;
+}
+CATCH_RETURN();
```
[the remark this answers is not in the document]
azchohfi: It is basically a catch-all:
https://github.com/microsoft/wil/blob/8501377980a52d98d1b51d04f60f571a7f0dc3db/include/wil/result_macros.h#L1364

packages/video_player/video_player_windows/windows/video_player_plugin.cpp
```diff
@@ -0,0 +1,236 @@
...
+    videoPlayers.at((int64_t)i)->Dispose();
+  }
+  videoPlayers.clear();
+
+  return {};
+}
+
+ErrorOr<int64_t> VideoPlayerPlugin::Create(
+    const std::string* asset, const std::string* uri,
+    const std::string* package_name, const std::string* format_hint,
```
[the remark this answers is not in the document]
azchohfi: Makes sense. I'll remove it. What about package_name? Should I open an issue on flutter itself? I don't think there is a Flutter API on Windows to load an asset of a package, right? https://docs.flutter.dev/ui/assets/assets-and-images#sharing-assets-with-the-underlying-platform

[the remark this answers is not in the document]
stuartmorgan: You can reference https://github.com/flutter/flutter/issues/103199 with a TODO.

I would do that [here](https://github.com/flutter/packages/blob/bd4cfbf6b85483bb3bd55e6ebf9e6938f42d54a1/packages/video_player/video_player_windows/lib/src/windows_video_player.dart#L42), and throw an UnimplementedError, and remove the Pigeon+native plumbing for it, and then we can decide how it should work when re-adding it later.

packages/video_player/video_player_windows/pigeons/messages.dart
```diff
@@ -0,0 +1,31 @@
...
+
+import 'package:pigeon/pigeon.dart';
+
+@ConfigurePigeon(PigeonOptions(
+  dartOut: 'lib/src/messages.g.dart',
+  dartTestOut: 'test/test_api.g.dart',
+  cppHeaderOut: 'windows/messages.h',
+  cppSourceOut: 'windows/messages.cpp',
+  cppOptions: CppOptions(
+    namespace: 'Messages',
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,130 @@
...
+#include <memory>
+#include <mutex>
+#include <sstream>
+#include <string>
+
+#include "MediaEngineWrapper.h"
+#include "MediaFoundationHelpers.h"
+#include "messages.h"
+
+using namespace Messages;
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,130 @@
...
+
+using namespace winrt;
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,130 @@
...
+
+class VideoPlayer {
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,130 @@
...
+ public:
+  VideoPlayer(IDXGIAdapter* adapter, HWND window, std::wstring uri,
+              flutter::EncodableMap httpHeaders);
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,130 @@
...
+  FlutterDesktopGpuSurfaceDescriptor* ObtainDescriptorCallback(size_t width,
+                                                               size_t height);
+
+  void Init(flutter::BinaryMessenger* messenger,
+            std::function<void()> textureFrameAvailableCallback,
+            int64_t textureId);
+
+  virtual ~VideoPlayer();
+
+  flutter::TextureVariant texture;
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,130 @@
...
+
+  FlutterDesktopGpuSurfaceDescriptor m_descriptor{};
+  std::mutex m_buffer_mutex;
+  winrt::com_ptr<IDXGIAdapter> m_adapter;
+  HWND m_window;
+  std::function<void()> m_textureFrameAvailableCallback;
+
+  bool isInitialized = false;
+
+  void SendInitialized();
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+#include "video_player.h"
+
+#include <flutter/event_channel.h>
+#include <flutter/event_stream_handler.h>
+#include <flutter/event_stream_handler_functions.h>
+#include <flutter/standard_method_codec.h>
+#include <shobjidl.h>
+#include <windows.h>
+
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,254 @@
...
+#undef GetCurrentTime
+
+using namespace winrt;
+
+VideoPlayer::VideoPlayer(IDXGIAdapter* adapter, HWND window, std::wstring uri,
+                         flutter::EncodableMap httpHeaders)
+    : VideoPlayer(adapter, window) {
+  // Create a source resolver to create an IMFMediaSource for the content URL.
+  // This will create an instance of an inbuilt OS media source for playback.
+  winrt::com_ptr<IMFSourceResolver> sourceResolver;
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  UpdateVideoSize();
+
+  return &m_descriptor;
+}
+
+void VideoPlayer::OnMediaInitialized() {
+  // Start playback
+  m_mediaEngineWrapper->SeekTo(0);
+  if (!this->isInitialized) {
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+void VideoPlayer::OnMediaError(MF_MEDIA_ENGINE_ERR error, HRESULT hr) {
+  LOG_HR_MSG(hr, "MediaEngine error (%d)", error);
+}
+
+void VideoPlayer::OnMediaStateChange(
+    media::MediaEngineWrapper::BufferingState bufferingState) {
+  if (bufferingState ==
+      media::MediaEngineWrapper::BufferingState::HAVE_NOTHING) {
+    this->SetBuffering(true);
```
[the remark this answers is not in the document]
azchohfi: Done.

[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+    uint32_t width;
+    uint32_t height;
+    m_mediaEngineWrapper->GetNativeVideoSize(width, height);
+    event.insert({flutter::EncodableValue("width"),
+                  flutter::EncodableValue((int32_t)width)});
+    event.insert({flutter::EncodableValue("height"),
+                  flutter::EncodableValue((int32_t)height)});
+
+    if (this->_eventSink) {
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,130 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#pragma once
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,130 @@
...
+#include <shobjidl.h>
+#include <unknwn.h>
+#include <wincodec.h>
+#include <winrt/Windows.Foundation.Collections.h>
+
+#include <future>
+#include <map>
+#include <memory>
+#include <sstream>
+#include <string>
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player_plugin.h
```diff
@@ -0,0 +1,69 @@
...
+
+class VideoPlayerPlugin : public flutter::Plugin, public WindowsVideoPlayerApi {
+ public:
+  static void RegisterWithRegistrar(flutter::PluginRegistrarWindows* registrar);
+
+  VideoPlayerPlugin(flutter::BinaryMessenger* messenger, HWND window,
+                    IDXGIAdapter* adapter,
+                    flutter::TextureRegistrar* textureRegistry);
+
+  std::optional<FlutterError> Initialize() override;
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player_plugin.h
```diff
@@ -0,0 +1,69 @@
...
+#include <wil/win32_helpers.h>
+#include <wincodec.h>
+#include <windows.h>
+
+#include <map>
+#include <memory>
+#include <sstream>
+#include <string>
+
+#include "include/video_player_windows/video_player_windows.h"
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player_plugin.h
```diff
@@ -0,0 +1,69 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <flutter/event_channel.h>
+#include <flutter/event_stream_handler.h>
+#include <flutter/event_stream_handler_functions.h>
+#include <flutter/plugin_registrar_windows.h>
+#include <flutter/standard_codec_serializer.h>
+#include <flutter/standard_method_codec.h>
```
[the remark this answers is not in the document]
azchohfi: Nop, not needed. Done.

packages/video_player/video_player_windows/windows/video_player_plugin.cpp
```diff
@@ -0,0 +1,236 @@
...
+  if (searchPlayer->second->IsValid()) {
+    searchPlayer->second->Pause();
+  }
+
+  return {};
+}
+
+std::optional<FlutterError> VideoPlayerPlugin::SetMixWithOthers(
+    bool mix_with_others) {
+  mixWithOthers = mix_with_others;
```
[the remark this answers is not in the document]
azchohfi: Forgot to remove. Done.

packages/video_player/video_player_windows/windows/video_player_plugin.cpp
```diff
@@ -0,0 +1,236 @@
...
+  auto searchPlayer = videoPlayers.find(texture_id);
+  if (searchPlayer == videoPlayers.end()) {
+    return FlutterError("player_not_found", "This player ID was not found");
+  }
+  if (searchPlayer->second->IsValid()) {
+    searchPlayer->second->Dispose();
+    videoPlayers.erase(texture_id);
+  }
+
+  return {};
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player_plugin.cpp
```diff
@@ -0,0 +1,236 @@
...
+}
+
+std::optional<FlutterError> VideoPlayerPlugin::SetMixWithOthers(
+    bool mix_with_others) {
+  mixWithOthers = mix_with_others;
+
+  return {};
+}
+
+VideoPlayerPlugin::~VideoPlayerPlugin() {}
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player_plugin.cpp
```diff
@@ -0,0 +1,236 @@
...
+    return FlutterError("player_not_found", "This player ID was not found");
+  }
+  if (searchPlayer->second->IsValid()) {
+    searchPlayer->second->Play();
+  }
+
+  return {};
+}
+
+ErrorOr<int64_t> VideoPlayerPlugin::Position(int64_t texture_id) {
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/test/mocks.h
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.cpp
azchohfi: Done.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.h
```diff
@@ -0,0 +1,113 @@
...
+
+// This class handles creation and management of the MediaFoundation
+// MediaEngine. It uses the provided IMFMediaSource to feed media
+// samples into the MediaEngine pipeline.
+class MediaEngineWrapper
+    : public winrt::implements<MediaEngineWrapper, IUnknown> {
+ public:
+  using ErrorCB = std::function<void(MF_MEDIA_ENGINE_ERR, HRESULT)>;
+
+  enum class BufferingState { HAVE_NOTHING = 0, HAVE_ENOUGH = 1 };
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.h
```diff
@@ -0,0 +1,113 @@
...
+        m_playbackEndedCB(playbackEndedCB),
+        m_timeUpdateCB(timeUpdateCB) {
+    // The initialize callback is required
+    THROW_HR_IF(E_INVALIDARG, !m_initializedCB);
+  }
+  ~MediaEngineWrapper() {
+    m_shouldExitLoop = true;
+    if (m_backgroundThread.joinable()) {
+      m_backgroundThread.join();
+    }
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.h
```diff
@@ -0,0 +1,113 @@
...
+  bool EnsureTextureCreated(DWORD width, DWORD height);
+  bool UpdateDXTexture();
+  bool UpdateDXTexture(DWORD width, DWORD height);
+  void InitializeVideo();
+  void CreateMediaEngine(IMFMediaSource* mediaSource);
+  void OnLoaded();
+  void OnError(MF_MEDIA_ENGINE_ERR error, HRESULT hr);
+  void OnBufferingStateChange(BufferingState state);
+  void OnPlaybackEnded();
+  void OnTimeUpdate();
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.h
```diff
@@ -0,0 +1,113 @@
...
+
+#include <flutter/plugin_registrar_windows.h>
+
+#include <thread>
+#include <tuple>
+
+#include "MediaEngineExtension.h"
+#include "MediaFoundationHelpers.h"
+
+namespace media {
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player_plugin.cpp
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/lib/src/windows_video_player.dart
```diff
@@ -0,0 +1,173 @@
...
+  @override
+  Future<void> setMixWithOthers(bool mixWithOthers) {
+    return _api.setMixWithOthers(mixWithOthers);
+  }
+
+  EventChannel _eventChannelFor(int textureId) {
+    return EventChannel('flutter.io/videoPlayer/videoEvents$textureId');
+  }
+
+  static const Map<VideoFormat, String> _videoFormatStringMap =
```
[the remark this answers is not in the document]
azchohfi: Removed.

azchohfi:
Went through most of the comments and fixed most of the issues.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,254 @@
...
+  if (_eventSink) {
+    _eventSink->Success(flutter::EncodableMap(
+        {{flutter::EncodableValue("event"),
+          flutter::EncodableValue(buffering ? "bufferingStart"
+                                            : "bufferingEnd")}}));
+  }
+}
+
+void VideoPlayer::SendInitialized() {
+  if (isInitialized) {
```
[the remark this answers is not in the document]
azchohfi: This is to avoid sending the initialized event more than once.

packages/video_player/video_player_windows/windows/video_player_plugin.h
```diff
@@ -0,0 +1,69 @@
...
+  std::optional<FlutterError> SetPlaybackSpeed(int64_t texture_id,
+                                               double speed) override;
+  std::optional<FlutterError> Play(int64_t texture_id) override;
+  ErrorOr<int64_t> Position(int64_t texture_id) override;
+  std::optional<FlutterError> SeekTo(int64_t texture_id,
+                                     int64_t position) override;
+  std::optional<FlutterError> Pause(int64_t texture_id) override;
+  std::optional<FlutterError> SetMixWithOthers(bool mix_with_others) override;
+
+  virtual ~VideoPlayerPlugin();
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.cpp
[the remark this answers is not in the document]
stuartmorgan: I don't follow; if `isInitialized` is true, and I call `SendInitialized` several times, then this check is going to pass every time and multiple events would be sent.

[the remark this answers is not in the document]
azchohfi: I see the confusion. The check is outside of the calls to this method, not inside. I could change this. This is mostly a port of the Android version, and that logic came with it. I just didn't question most of it.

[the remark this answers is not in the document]
azchohfi: Ahhh, I see. The problem is not even that. Every time that the method is called, that variable is set to true, so there is no reason to check for it. I can just remove the check. My bad.

packages/video_player/video_player_windows/example/.gitignore
```diff
@@ -0,0 +1 @@
+.vs/
```
loic-sharma, line 1: Is this right? Should we mirror other example `.gitignore`s? For example: https://github.com/flutter/packages/blob/main/packages/url_launcher/url_launcher_windows/example/.gitignore

stuartmorgan: Generally we use the `flutter create`d `.gitignore` without changes for the example apps.

azchohfi: I believe this should be added, maybe to the root .gitignore. Since Flutter generates Visual Studio projects for Windows, whenever I open the project files, it automatically creates the .vs folder, and it tries to check it in. At least in the samples it does.
It is added on the recommended .gitignore for Visual Studio: https://github.com/github/gitignore/blob/main/VisualStudio.gitignore#L36C4-L36C5

stuartmorgan: How are you opening them? The generated projects are in the `build` folder, which is already excluded.

azchohfi: I see. It wasn't doing that since I originally used the .gitignore from the Android project here:
https://github.com/flutter/packages/blob/main/packages/video_player/video_player_android/example/.gitignore
I can use the default one.

stuartmorgan:
I tried running this locally, and I'm getting an assertion failure in the `THROW_IF_FAILED(D3D11CreateDevice(...))` in `InitializeVideo` on startup as a modal dialog; this may well be what's happening in CI.

Interestingly if I press Ignore, the app then launches and the video plays.

stuartmorgan:
`MediaEngineWrapper::Initialize` is calling:
- `InitializeVideo`
- `CreateMediaEngine`
  - which itself has a call to `InitializeVideo` early on, and `InitializeVideo` doesn't appear to handle being called multiple times (since it passes `d3d11_device_.put()` each time, and re-using `put` without nulling it is a programming error.

I'm curious how you aren't seeing this locally.

azchohfi:
Oh, nice catch. No idea why it is not failing on my box. I've tested it in two Win11 machines with no issues. I'll update this to properly propagate initialization errors back to dart.

azchohfi:
This seems to be an issue on the media foundation samples. It has exactly the same pattern. It isn't a bigger issue there since the d3d11Device is a local variable there, not a member field like here, but it shouldn't be doing that anyway.

packages/video_player/video_player_windows/windows/video_player_plugin.h
```diff
@@ -0,0 +1,69 @@
...
+
+ private:
+  std::map<int64_t, std::unique_ptr<VideoPlayer>> videoPlayers;
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player_plugin.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  return textureId;
+}
+
+std::optional<FlutterError> VideoPlayerPlugin::Dispose(int64_t texture_id) {
+  auto searchPlayer = videoPlayers.find(texture_id);
+  if (searchPlayer == videoPlayers.end()) {
+    return FlutterError("player_not_found", "This player ID was not found");
+  }
+  if (searchPlayer->second->IsValid()) {
```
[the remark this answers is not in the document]
azchohfi: This check was unecessary. I did that code before properly joining the background thread on disposal.

packages/video_player/video_player_windows/windows/video_player.cpp
azchohfi: Not needed. Same as before. I removed this code.

packages/video_player/video_player_windows/windows/video_player.cpp
azchohfi: Same as the other one. Removed.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.h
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+#include "MediaEngineWrapper.h"
+#include "MediaFoundationHelpers.h"
+
+using namespace Microsoft::WRL;
+
+namespace media {
+
+namespace {
+class MediaEngineCallbackHelper
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/MediaFoundationHelpers.h
```diff
@@ -0,0 +1,146 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#pragma once
+
+namespace media {
+
+class MFPlatformRef {
```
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,106 @@
...
+#include "messages.h"
+
+namespace video_player_windows {
+
+class VideoPlayer {
+ public:
+  VideoPlayer(IDXGIAdapter* adapter, HWND window, std::wstring uri,
+              flutter::EncodableMap http_headers);
+
+  virtual ~VideoPlayer() = default;
```
loic-sharma, line 51: Should `~VideoPlayer` call `Dispose`? Alternatively, should the destructor verify that `Dispose` was called?

packages/video_player/video_player_windows/windows/video_player_plugin.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  search_player->second->Play();
+
+  return std::nullopt;
+}
+
+ErrorOr<int64_t> VideoPlayerPlugin::GetPosition(int64_t texture_id) {
+  auto search_player = video_players_.find(texture_id);
+  int64_t position = 0;
+  if (search_player != video_players_.end()) {
```
loic-sharma, line 178: Should this return a `player_not_found` error?

packages/video_player/video_player_windows/windows/video_player.h
```diff
@@ -0,0 +1,106 @@
...
+  void SendBufferingUpdate();
+  void SeekTo(int64_t seek);
+  int64_t GetTextureId();
+
+  FlutterDesktopGpuSurfaceDescriptor* ObtainDescriptorCallback(size_t width,
+                                                               size_t height);
+
+  void Init(flutter::BinaryMessenger* messenger,
+            std::function<void(int64_t)> texture_frame_available_callback,
+            flutter::TextureRegistrar* texture_registry);
```
loic-sharma, line 69: Should we remove the `Init` method and move these arguments to the `VideoPlayer` constructor?

From my understanding, the `Init` is the shared arguments regardless of whether the asset is remote or local.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.cpp
```diff
@@ -0,0 +1,506 @@
...
+    // Ensure that callbacks are valid
+    THROW_HR_IF(E_INVALIDARG, !on_loaded_cb_);
+    THROW_HR_IF(E_INVALIDARG, !error_cb_);
+    THROW_HR_IF(E_INVALIDARG, !buffering_state_change_cb_);
+    THROW_HR_IF(E_INVALIDARG, !playback_ended_cb_);
+    THROW_HR_IF(E_INVALIDARG, !time_update_cb_);
+  }
+  virtual ~MediaEngineCallbackHelper() = default;
+
+  void DetachParent() {
```
loic-sharma, line 55: Who calls this? Should the destructor call thsi?

This doesn't appear to be on the `IMFMediaEngineNotify` interface and I couldn't find any references in this code base. Consider adding a comment here.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.cpp
```diff
@@ -0,0 +1,506 @@
...
+        error_cb_(error_cb),
+        buffering_state_change_cb_(buffering_state_change_cb),
+        playback_ended_cb_(playback_ended_cb),
+        time_update_cb_(time_update_cb) {
+    // Ensure that callbacks are valid
+    THROW_HR_IF(E_INVALIDARG, !on_loaded_cb_);
+    THROW_HR_IF(E_INVALIDARG, !error_cb_);
+    THROW_HR_IF(E_INVALIDARG, !buffering_state_change_cb_);
+    THROW_HR_IF(E_INVALIDARG, !playback_ended_cb_);
+    THROW_HR_IF(E_INVALIDARG, !time_update_cb_);
```
loic-sharma, line 51: @stuartmorgan Does the packages repo have a way to do debug-only assertions, similar to the engine's `FML_DCHECK`? Ideally we would use that here instead of `THROW_HR_IF_`.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,246 @@
...
+FlutterDesktopGpuSurfaceDescriptor* VideoPlayer::ObtainDescriptorCallback(
+    size_t width, size_t height) {
+  // Lock buffer mutex to protect texture processing.
+  std::lock_guard<std::mutex> buffer_lock(buffer_mutex_);
+
+  media_engine_wrapper_->UpdateSurfaceDescriptor(
+      static_cast<uint32_t>(width), static_cast<uint32_t>(height),
+      [&]() { texture_frame_available_callback_(texture_id_); }, descriptor_);
+
+  UpdateVideoSize();
```
loic-sharma, line 98: Why don't we pass the `width` and `height` down to `UpdateVideoSize`? `UpdateVideoSize` appears to use the Flutter view's `HWND` to determine the video's size, which I suspect is incorrect if the video doesn't cover the entire Flutter view `HWND`.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.cpp
azchohfi: Nice catch, this was missing. Should be called from the dctor.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+  // If the window has not been initialized yet, use a default size of 640x480.
+  const bool window_initialized =
+      window_size_.Width != 0 && window_size_.Height != 0;
+  float width = window_initialized ? window_size_.Width : 640;
+  float height = window_initialized ? window_size_.Height : 480;
+
+  if (media_engine_wrapper_) {
+    // Call into media engine wrapper on MTA thread to resize the video surface.
+    RunSyncInMTA([&]() {
```
loic-sharma, line 128: Is this necessary? It looks like `MediaEngineWrapper::OnWindowUpdate` also calls `RunSyncInMTA`.

packages/video_player/video_player_windows/windows/video_player.cpp
```diff
@@ -0,0 +1,246 @@
...
+                  flutter::StreamHandlerError<flutter::EncodableValue>> {
+            event_sink_ = nullptr;
+            return nullptr;
+          }));
+}
+
+FlutterDesktopGpuSurfaceDescriptor* VideoPlayer::ObtainDescriptorCallback(
+    size_t width, size_t height) {
+  // Lock buffer mutex to protect texture processing.
+  std::lock_guard<std::mutex> buffer_lock(buffer_mutex_);
```
loic-sharma, line 92: Could you expand the comments on what `buffer_mutex_` and `lock_` guard?

My understanding: 
1. Only `VideoPlayer::ObtainDescriptorCallback` locks `buffer_mutex_`
2. Only `VideoPlayer::UpdateVideoSize` locks `lock_`
3. `VideoPlayer::UpdateVideoSize` is called only by `VideoPlayer::ObtainDescriptorCallback`

My questions:
1. What kind of problems do `lock_` and `buffer_mutex_` protect against?
1. Why do we need both `lock_` and `buffer_mutex_`? It looks like `buffer_mutex_` is always locked before `lock_` is locked.

packages/video_player/video_player_windows/windows/video_player_plugin.cpp
```diff
@@ -0,0 +1,219 @@
...
+  if (search_player == video_players_.end()) {
+    return FlutterError(
+        "player_not_found",
+        "Player not found for texture id: " + std::to_string(texture_id));
+  }
+
+  auto& player = search_player->second;
+
+  position = player->GetPosition();
+  player->SendBufferingUpdate();
```
loic-sharma, line 187: As someone not familiar with the video player, this is unexpected. Why is this necessary? Consider adding a comment explaining this.

loic-sharma:
Thank you for sending this pull request, this is amazing!!

I did a first pass from a Windows perspective. I'm still ramping up on Windows Media Foundation, so I'll need to do another pass later.

/cc @yaakovschectman who also works on Windows

packages/video_player/video_player_windows/windows/MediaEngineWrapper.cpp
stuartmorgan: Every package's code has to stand alone, or it wouldn't compile in actual use by clients, so there is nothing like `FML` for this repository.

stuartmorgan:
I played with the latest version locally a bit, and I can confirm that those assertions are gone for me! A few things I noticed:
- Playback seems intermittently choppy. It feels like maybe the buffer swap is being driven consistently? (This is just speculation, since it feels somewhat like the bad playback I had in early versions of the macOS version when I hadn't gotten the display link timer set up correctly.)
- I'm getting logging from the engine about an event being sent from a non-platform thread. I'm going to see if I can quickly track that down.
- Resizing the window pretty reliably crashes, sometimes with logging from the engine that sounds like it might be an assertion failure.

stuartmorgan:
> * I'm getting logging from the engine about an event being sent from a non-platform thread. I'm going to see if I can quickly track that down.

Looks like this is the `event_sink_` call in `VideoPlayer::SendInitialized`, which is coming from `MFWorkItem::Invoke` on a thread called (if I'm reading this VS UI correctly) `ntdll.dll thread`.

And then also `VideoPlayer::SetBuffering` on the same thread, from `MediaEngineCallbackHelper::EventNotify`

packages/video_player/video_player_windows/windows/video_player.cpp
azchohfi: This was a mess and not necessary. I've simplified it.

packages/video_player/video_player_windows/windows/video_player.cpp
azchohfi: Nice catch. Fixed.

stuartmorgan:
> * Resizing the window pretty reliably crashes

Running this under the debugger, in one case I saw an assertion failure in `MediaEngineWrapper::EnsureTextureCreated`; as with the earlier crash, it's `texture_.put()` being non-null.

But the actual crash seems to be an abort coming from deep in the Flutter engine on the raster thread. I don't have a debug engine on Windows so I don't have more detail. Given that it's resize, I'm guessing that the most likely situation is that a buffer is being destroyed while the engine is still using it.

packages/video_player/video_player_windows/windows/video_player.cpp
azchohfi: Yes, you are right. I've updated it. It also simplified things since I don't need to pass the window's HWND around anymore.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.cpp
azchohfi: Fixed.

packages/video_player/video_player_windows/windows/video_player.h
azchohfi: The plugin's Dispose is now calling dispose for everything.

packages/video_player/video_player_windows/windows/video_player.h
azchohfi: I don't think it makes much difference.

azchohfi: I can update this if you want, but I wanted to keep this implementation closer to the Android and iOS.

azchohfi:
I created a generic RunOnMainThread method, that adds a delegate to a queue, registers a top-level Window Proc Delegate, which dequeues the delegate and run it. This solved these issues, but I think other plugins might benefit from this, as there is an equivalent in Android and iOS.

stuartmorgan:
> This solved these issues, but I think other plugins might benefit from this, as there is an equivalent in Android and iOS.

https://github.com/flutter/flutter/issues/134346 has some related discussion; we've considered, rejected, and now are reconsidering building that into the plugin APIs.

packages/video_player/video_player_windows/windows/video_player.h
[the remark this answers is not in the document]
azchohfi: Done.

packages/video_player/video_player_windows/test/windows_video_player_test.dart
```diff
@@ -0,0 +1,416 @@
...
+
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:video_player_platform_interface/video_player_platform_interface.dart';
+import 'package:video_player_windows/video_player_windows.dart';
+
+import 'test_api.g.dart';
+
+class _ApiLogger implements TestHostVideoPlayerApi {
+  final List<String> log = <String>[];
```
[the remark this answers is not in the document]
azchohfi: Replaced this with mockito.

packages/video_player/video_player_windows/test/windows_video_player_test.dart
```diff
@@ -0,0 +1,416 @@
...
+    expect(VideoPlayerPlatform.instance, isA<WindowsVideoPlayer>());
+  });
+
+  group('$WindowsVideoPlayer', () {
+    final WindowsVideoPlayer player = WindowsVideoPlayer();
+    late _ApiLogger log;
+
+    setUp(() {
+      log = _ApiLogger();
+      TestHostVideoPlayerApi.setup(log);
```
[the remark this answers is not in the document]
azchohfi: Using mockito now.

packages/video_player/video_player_windows/windows/MediaEngineWrapper.cpp
loic-sharma: Gotcha, that makes sense. Closing

packages/video_player/video_player_windows/windows/video_player.h
loic-sharma: I don't feel strongly about this. Mirroring other platform's video player plugin is completely valid. Feel free to close this comment if you prefer as is.

lucasjinreal:
When would this feature merge? It's really needed!

FZY456:
Hello, I've found an issue with video_player_windows. It throws an error when playing a locally stored video file with a filename containing Chinese characters.

The error log is as follows: 

```
 #0      WindowsVideoPlayerApi.create (package:video_player_windows/src/messages.g.dart:89:7)
 <asynchronous suspension>
 #1      WindowsVideoPlayer.create (package:video_player_windows/src/windows_video_player.dart:56:27)
 <asynchronous suspension>
 #2      VideoPlayerController.initialize (package:video_player/video_player.dart:434:19)
```

stuartmorgan:
@FZY456 You've only included the stack there, not the actual error.

FZY456:
> 您只包含了堆栈，而不是实际错误。

`PlatformException(uri_load_failed, , null, null)`

FZY456:
> @FZY456 You've only included the stack there, not the actual error.

`PlatformException(uri_load_failed, , null, null)`

stuartmorgan:
From triage: What's the current status of this PR? Is it just waiting for an updated review?

azchohfi:
No, review is probably still updated, but there is work on **my** end. I just didn't have time to do yet. My bad.

Andrekarma:
Any update on this pr?

stuartmorgan:
From triage: @azchohfi Is this still something you're planning to come back to?

azchohfi:
Unfortunately, I do not have time right now to look at this. I would say that most of the code works but it needs some tweaking to make it production ready. Feel free to work on top of this to get it merged.

LeoSandbox:
Hey guys, I was wondering if you had any timeline in mind? I will be using this for my Flutter game cinematic that should run on MacOS and Windows. 
[https://store.steampowered.com/app/2489720/Drakontias/](https://store.steampowered.com/app/2489720/Drakontias/)

stuartmorgan:
As there isn't a planned timeline to finish this PR, I'm going to go ahead and close it to reflect the fact that it's not active. The issue has been marked as having a partial PR, and we encourage anyone in the community interested in this functionality to pick up the work from here.

OutdatedGuy:
For whoever willing to take the work from here, some observations on choppiness:

If you run the example app and switch to "Asset" tab, the butterfly video will start playing in a very choppy manner. Try making the window smaller by re-sizing it slowly and at some threshold the video starts playing smoothly.

On my machine, when the **VideoPlayer** size was _607.2 x 341.55_ the video was playing smoothly and when the size was _608 x 342_ the video became very choppy.

You can try it yourself by wrapping the `VideoPlayer` widget inside `_ButterFlyAssetVideo` with a `LayoutBuilder`:

```dart
LayoutBuilder(
  builder: (context, constraints) {
    print('VideoPlayer size: ${constraints.maxWidth} x ${constraints.maxHeight}');
    return VideoPlayer(_controller);
  },
),
```

PR #5886, opened 2024-01-13
Bump `release` to use 3.16.6

stuartmorgan:
google_maps_flutter_ios now requires 3.16.6, so we need to bump the release script so that it can be published by CI.

Fixes the tree.


stuartmorgan:
I'll just publish the failing release manually to avoid a revert/reland cycle, since it's just a single package.

bparrishMines (approved):
LGTM

PR #5899, opened 2024-01-16
Add "infinity", "negativeInfinity", and "nan" keywords to rfwtxt.

Hixie:
## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/rfw/lib/src/dart/text.dart
+++ b/packages/rfw/lib/src/dart/text.dart
@@ -40,6 +40,9 @@ import 'model.dart';
 ///    expressed as hex literals. Numbers are explicitly 64 bit precision;
 ///    specifically, signed 64 bit integers, or binary64 floating point numbers.
 ///
+///  * three keywords, "infinity", "negativeInfinity", and "nan", allow one
+///    to express the non-finite values of the binary64 format.
+///
 ///  * files are always rooted at a map. (Different versions of JSON are
 ///    ambiguous or contradictory about this.)
 ///
@@ -141,15 +144,20 @@ import 'model.dart';
 ///
 /// ```
 /// double ::= "-"? digit+ ("." digit+)? (("e" | "E") "-"? digit+)?
+///            | "infinity" | "negativeInfinity" | "nan"
 /// ```
 ///
-/// Floating point numbers are represented by an optional negative sign
+/// Finite floating point numbers are represented by an optional negative sign
 /// indicating the number is negative, a significand with optional fractional
 /// component in the form of digits in base ten giving the integer component
 /// followed optionally by a decimal point and further base ten digits giving
 /// the fractional component, and an exponent which itself is represented by an
 /// optional negative sign indicating a negative exponent and a sequence of
 /// digits giving the base ten exponent itself.
+///
+/// Infinity, negative infinity, and one arbitrary "not-a-number" value are
+/// represented by the keywords "infinity", "negativeInfinity", and "nan"
+/// respectively.
 ///
 /// The numbers represented must be values that can be expressed in the IEEE754
 /// binary64 format.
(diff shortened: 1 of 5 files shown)


packages/rfw/lib/src/dart/text.dart
```diff
@@ -2215,6 +2223,18 @@ class _Parser {
         _advance();
         return false;
       }
```
tugorez, line 2225: Should these identifiers be added to the reserved words list?

packages/rfw/test/text_test.dart
```diff
@@ -193,7 +193,9 @@ void main() {
     expect(parseDataFile('{ "a": -1e-16 }'), <String, Object?>{ 'a': -0.0000000000000001 });
     expect(parseDataFile('{ "a": -1e-17 }'), <String, Object?>{ 'a': -0.00000000000000001 });
     expect(parseDataFile('{ "a": -1e-18 }'), <String, Object?>{ 'a': -0.000000000000000001 });
-    expect(parseDataFile('{ "a": -1e-19 }'), <String, Object?>{ 'a': -0.0000000000000000001 });
+    expect(parseDataFile('{ "a": infinity }'), <String, Object?>{ 'a': double.infinity });
+    expect(parseDataFile('{ "a": negativeInfinity }'), <String, Object?>{ 'a': double.negativeInfinity });
+    expect(parseDataFile('{ "a": nan }')['a'], isNaN);
```
tugorez, line 198: nit: For the sake of consistency, should this particular test be

expect(..., <String, Object?> { 'a': double.nan });

peixinli (dismissed):
Thanks for adding this!

packages/rfw/test/text_test.dart
Hixie: that would fail, because `double.nan == double.nan` is false

packages/rfw/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: rfw
 description: "Remote Flutter widgets: a library for rendering declarative widget description files at runtime."
 repository: https://github.com/flutter/packages/tree/main/packages/rfw
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+rfw%22
-version: 1.0.16
+version: 1.0.19
```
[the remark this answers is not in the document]
Hixie: i have a bunch of other PRs for this package and gave them all different numbers. I'll pick the right one when I land it. :-)

packages/rfw/lib/src/dart/text.dart
Hixie: Hmm. If we add them to the reserved words list then this becomes a breaking change.

tugorez: Wondering if something like https://github.com/flutter/flutter/issues/141833#issuecomment-1901153459 would be useful here as well (instead of changing the language). The implementation of number could be as follows.

```
final result = _fetch(...);
if (result is int) return result as int;
if (result is double) return result as double;
if (result == 'infinity') return double.infinity;
if (result == 'nan') return double.nan;
```

Hixie: That would similarly be a breaking change though.

Hixie:
We could make infinity and negative infinity be written as `+infinity` and `-infinity` and not support NaN. That would avoid the breaking change aspect (because those two syntaxes are errors today).

Hixie:
We could also use `∞` and `-∞`.

stuartmorgan:
From triage: what's the status of this PR? Should it be a Draft?

Hixie:
There doesn't seem to be a huge amount of interest here so I'm tempted to just close it for now until we have a clearer need.

Hixie:
closing per last comment. If you would like this feature please file a bug and cc me. Thanks!

PR #5796, opened 2024-01-04
[webview_flutter_android] Add javascript panel interface for android

jsharp83:
* There are cases where Web calls System Popup with javascript on webview_flutter
* Android has a interface on WebChromeClient
   * https://developer.android.com/reference/android/webkit/WebChromeClient#onJsAlert(android.webkit.WebView,%20java.lang.String,%20java.lang.String,%20android.webkit.JsResult)

* Related issue: https://github.com/flutter/flutter/issues/30358#issuecomment-1645347616
* Related Interface PR: https://github.com/flutter/packages/pull/5670
* The PR that contains all changes can be found at https://github.com/flutter/packages/pull/4704

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[...]

PR #5901, opened 2024-01-16
[webview_flutter_platform_interface] Adds support to return `null` in `PlatformWebViewController.setOnJavaScriptTextInputDialog`

bparrishMines:
Based on the documentation for `prompt` and [WebChromClient.onJsPrompt](https://developer.android.com/reference/android/webkit/WebChromeClient#onJsPrompt(android.webkit.WebView,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20android.webkit.JsPromptResult)), this method should support returning a null value for the scenario that a user cancels the operation to enter text. 

Side note: It seems reasonable the `confirm` dialog can just return `false` if the user cancels the operation, so no changes to that one.

## Pre-launch Checklist

- [x] I signed the [CLA].


bparrishMines:
Whoops it looks like I forgot the android implementation was already submitted.

packages/webview_flutter/webview_flutter_android/test/android_webview_controller_test.mocks.dart
```diff
@@ -870,19 +870,6 @@ class MockAndroidWebViewController extends _i1.Mock
         returnValue: _i9.Future<void>.value(),
         returnValueForMissingStub: _i9.Future<void>.value(),
       ) as _i9.Future<void>);
-
-  @override
-  _i9.Future<void> setOnJavaScriptTextInputDialog(
```
bparrishMines, line 875: https://github.com/flutter/packages/pull/5796 has not landed yet, so I removed these generated overrides to have the `pathified` test pass.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/platform_webview_controller.dart
```diff
@@ -318,7 +318,7 @@ abstract class PlatformWebViewController extends PlatformInterface {
   /// Sets a callback that notifies the host application that the web page
   /// wants to display a JavaScript prompt() dialog.
   Future<void> setOnJavaScriptTextInputDialog(
-      Future<String> Function(JavaScriptTextInputDialogRequest request)
+      Future<String?> Function(JavaScriptTextInputDialogRequest request)
```
stuartmorgan, line 321: This is a breaking change; we'll need to make a replacement method instead (e.g., `setOnJavaScriptTextInputDialogCancelable`).

We should also document what `null` means in its doc comment.

stuartmorgan (changes requested):

packages/webview_flutter/webview_flutter_platform_interface/lib/src/platform_webview_controller.dart
bparrishMines: Is there not a way to submit this without a breaking change? The Android and iOS implementation PRs have not landed yet.

stuartmorgan: It's already breaking our own tests in other packages to try to change it, and there's no reason a third-party implementation of `webview_flutter` couldn't already be relying on it.

bparrishMines: Thats fair. I'll create a separate method for it. It's unfortunate that I didn't catch this earlier.

stuartmorgan: Yes, the reason I set the process up to have the individual PRs only split out once the combined PR has been fully reviewed is to try to avoid this kind of thing, but it's inevitable that sometimes we just won't notice things until that final read-through.

bparrishMines:
closing since this isn't on my current list of todos. Added https://github.com/flutter/flutter/issues/145799 to track interest.

PR #5907, opened 2024-01-17
[rfw] Add support for widget builders

tugorez:
This PR adds support for widget builders.

https://github.com/flutter/flutter/issues/141658

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/rfw/lib/src/dart/model.dart
```diff
@@ -439,6 +439,22 @@ class ConstructorCall extends BlobNode {
...
+/// blobs.
+class FunctionDeclaration extends BlobNode {
+  /// Creates a [FunctionDeclaration] for a function that returns a widget.
+  const FunctionDeclaration(this.argumentName, this.constructorCall);
+
+  /// RFW functions can only take a single argument binded to this identifier.
+  final String argumentName;
+
+  /// RFW functions can only return a [ConstructorCall].
+  final ConstructorCall constructorCall;
```
Hixie, line 452: Should we support Switch here too?

tugorez: I don't think it is stricly needed. There could be two potential use cases of switch here:

#### A user wants to changed the passed value based on a given `args` or `state` entry value. 

For this particular case ,they can use a switch wrapping the builders

```
Builder(
  builder: switch args.down {
     true: (foo) => Text(...),
     false: (foo) => Container(...),
  }
);
```

#### A user wants to changed the passed value based on the scoped builder var
If I'm not wrong, you're suggesting something like this

```
Builder(
  builder:  (foo) => switch foo.down {
    case true:  (foo) => Text(...)
    case false:  (foo) => Text(...)
  }
);
```

but the same can be achieved by doing

Builder(
  builder:  (foo) => Foo(down: foo.down),
);

widget Foo = switch args.down {
 ...
}


Do you see any value in supporting switch here as well? (We can always add it later).

packages/rfw/lib/src/flutter/runtime.dart
```diff
@@ -423,6 +445,25 @@ class Runtime extends ChangeNotifier {
...
+        // TODO(tugorez): bind the builder argument.
+        final DynamicMap subArguments = _bindArguments(
+          context,
+          node.constructorCall.arguments,
+          arguments,
+          stateDepth,
+          usedWidgets,
+        ) as DynamicMap;
+        return _applyConstructorAndBindArguments(
+          FullyQualifiedWidgetName(context.library, node.constructorCall.name),
```
tugorez, line 459: Wonder if this is fine, technically speaking this is an anonymous function.

packages/rfw/test/runtime_test.dart
```diff
@@ -1088,4 +1088,156 @@ void main() {
...
+        );
+      ''');
+      await tester.pumpWidget(widget);
+
+      final Finder textFinder = find.byType(Text);
+      expect(textFinder, findsOneWidget);
+      expect(tester.widget<Text>(textFinder).data, '(1 + 2) * 3 = 9');
+    });
+
+    testWidgets('Widget builders - swich works with builder', (WidgetTester tester) async {
```
Hixie, line 1200: typo ("swich" -> "switch")

packages/rfw/test/runtime_test.dart
```diff
@@ -1088,4 +1088,156 @@ void main() {
...
+        );
+      ''');
+      await tester.pumpWidget(widget);
+
+      final Finder textFinder = find.byType(Text);
+      expect(textFinder, findsOneWidget);
+      expect(tester.widget<Text>(textFinder).data, 'The result is not zero');
+    });
+  });
 }
```
Hixie, line 1243: these tests are great. we should also add some tests that verify that things update correctly when the data is updated. for example, have a version of the calculator test where the operands come from a data reference, and the data is updated several times and you see if that causes the render tree to change appropriately.

packages/rfw/lib/src/dart/model.dart
tugorez: Ended up implementing this.

packages/rfw/lib/src/dart/binary.dart
```diff
@@ -468,6 +474,14 @@ class _BlobDecoder {
     return ConstructorCall(name, _readMap(_readArgument)!);
   }
 
+  WidgetBuilderDeclaration _readWidgetBuilder() {
+    final String argumentName = _readString();
+    final int type = _readByte();
+    assert(type == _msWidget || type == _msSwitch);
```
Hixie, line 480: what guarantees that this type is only one of those two values?

i would have expected this to be a FormatException rather than a debug-only assert.

packages/rfw/lib/src/dart/model.dart
```diff
@@ -439,6 +439,26 @@ class ConstructorCall extends BlobNode {
   String toString() => '$name($arguments)';
 }
 
+/// Representation of functions that return widgets in Remote Flutter library blobs.
```
Hixie, line 442: "Remote Flutter Widgets library blobs."

packages/rfw/lib/src/dart/model.dart
```diff
@@ -439,6 +439,26 @@ class ConstructorCall extends BlobNode {
...
+class WidgetBuilderDeclaration extends BlobNode {
+  /// Creates a [WidgetBuilderDeclaration].
```
Hixie, line 444: this documentation is insufficient per our style guide which says that documentation needs to have more information than can be guessed from just the name of the member being documented.
https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#documentation-dartdocs-javadocs-etc

packages/rfw/lib/src/dart/model.dart
```diff
@@ -534,6 +554,31 @@ class DataReference extends Reference {
   String toString() => 'data.${parts.join(".")}';
 }
 
+/// Reference to the [DynamicMap] passed into the widget builder.
+///
+/// This class is used to represent references to a function argument.
+/// In "(scope) => Container(width: scope.width)" this represents "scope.width".
```
Hixie, line 560: Use \` rather than " for code snippets.

packages/rfw/lib/src/dart/model.dart
```diff
@@ -534,6 +554,31 @@ class DataReference extends Reference {
...
+///
+/// See also:
+/// 
+///   * [WidgetBuilderDeclaration] which represents a widget builder definition.
```
Hixie, line 564: we usually put a comma after the `]` in see also lists

packages/rfw/lib/src/dart/model.dart
```diff
@@ -439,6 +439,26 @@ class ConstructorCall extends BlobNode {
   String toString() => '$name($arguments)';
 }
 
+/// Representation of functions that return widgets in Remote Flutter library blobs.
+class WidgetBuilderDeclaration extends BlobNode {
+  /// Creates a [WidgetBuilderDeclaration].
+  const WidgetBuilderDeclaration(this.argumentName, this.widget);
+
+  /// The name associated with the passed [DynamicMap].
+  final String argumentName;
```
Hixie, line 448: also needs more detail

packages/rfw/lib/src/dart/model.dart
```diff
@@ -534,6 +554,31 @@ class DataReference extends Reference {
...
+/// 
+///   * [WidgetBuilderDeclaration] which represents a widget builder definition.
+class WidgetBuilderArgReference extends Reference {
+  /// Wraps the given [parts] associated to the [argumentName] as an [WidgetBuilderArgReference].
+  ///
+  /// The parts must not be mutated after the object is created.
+  const WidgetBuilderArgReference(this.argumentName, super.parts);
+
+  /// References the function argument name.
+  final String argumentName;
```
Hixie, line 572: needs more detail

packages/rfw/lib/src/dart/text.dart
```diff
@@ -2306,10 +2377,15 @@ class _Parser {
         _advance();
         return _withSourceRange(StateReference(_readParts()), start);
       }
+      if (widgetBuilderScope.contains(identifier)) {
+        final SourceLocation? start = _getSourceLocation();
+        _advance();
+        return _withSourceRange(WidgetBuilderArgReference(identifier, _readParts()), start);
+      }
```
Hixie, line 2384: let's put this just before the loop identifiers logic (after `set`), to keep the function more consistent.

packages/rfw/lib/src/dart/text.dart
```diff
@@ -2327,16 +2403,39 @@ class _Parser {
...
+    final String argumentName = _readIdentifier();
+    _checkIsNotReservedWord(argumentName, argumentNameToken);
+    _expectSymbol(_SymbolToken.closeParen);
+    _expectSymbol(_SymbolToken.equals);
+    _expectSymbol(_SymbolToken.greatherThan);
+    final Object widget = _readValue(
+      extended: true,
+      widgetBuilderScope: <String>[...widgetBuilderScope, argumentName],
+    );
+    assert(widget is ConstructorCall || widget is Switch);
```
Hixie, line 2425: as in the binary section, this seems like it should be an exception, not an assert. Asserts are for things that, if they occur, indicate an error in the code, but here you're using it to catch an error in the data being parsed, unless i misunderstand something.

packages/rfw/lib/src/flutter/runtime.dart
```diff
@@ -126,6 +129,25 @@ abstract class DataSource {
   /// non-widget nodes replaced by [ErrorWidget].
   List<Widget> childList(List<Object> argsKey);
 
+  /// Build the builder at the given key.
+  ///
+  /// If the node specified is not a widget, returns an [ErrorWidget].
```
Hixie, line 134: surely the "node specified" should be a builder?

packages/rfw/lib/src/flutter/runtime.dart
```diff
@@ -126,6 +129,25 @@ abstract class DataSource {
...
+  ///
+  /// See also:
+  ///
+  ///  * [optionalBuilder], which returns null if the widget is missing.
+  Widget builder(List<Object> argsKey, DynamicMap builderArg);
+
+  /// Build the builder at the given key.
+  ///
+  /// If the node specified is not a widget, returns null.
```
Hixie, line 143: same here

packages/rfw/lib/src/flutter/runtime.dart
```diff
@@ -1095,6 +1437,17 @@ class _WidgetState extends State<_Widget> implements DataSource {
     return subscription.value;
   }
 
+  Object _widgetBuilderArgResolver(List<Object> rawDataKey) {
+    final _Key widgetBuilderArgKey = _Key(_kWidgetBuilderArgSection, rawDataKey);
+    final _Subscription subscription = _subscriptions[widgetBuilderArgKey] ??= _Subscription(
```
Hixie, line 1442: putIfAbsent may be more idiomatic

packages/rfw/lib/src/flutter/runtime.dart
```diff
@@ -1095,6 +1437,17 @@ class _WidgetState extends State<_Widget> implements DataSource {
...
+      widget.widgetBuilderScope,
```
Hixie, line 1443: i don't understand how this works. how do you get the right widget builder argument, e.g. in a nested case?

packages/rfw/test/runtime_test.dart
```diff
@@ -1088,4 +1088,156 @@ void main() {
...
+
+        widget test = Calculator(
+          operand1: 1,
+          operand2: 2,
+          operation: 'sum',
+          builder: (result1) => Calculator(
+            operand1: result1.result,
+            operand2: 3,
+            operation: 'multiply',
+            builder: (result2) => CoolText(text: ['(1 + 2) * 3 = ', result2.result]),
```
Hixie, line 1189: there should be a version of this test that tries to read `result1`

packages/rfw/lib/src/flutter/runtime.dart
tugorez: This might help https://github.com/flutter/packages/blob/e7a48123e5e14b316a73acbd599bc9056cf4799a/packages/rfw/lib/src/flutter/runtime.dart#L548

tugorez: Not for this PR but I think we can generalize (and simplify) state, data and this newly introduced builder scope concept in a single scope entity (pretty much like interpreters).

packages/rfw/test/runtime_test.dart
tugorez: Great call @Hixie, done!

packages/rfw/test/runtime_test.dart
```diff
@@ -1173,6 +1178,84 @@ void main() {
...
+      await tester.pumpWidget(widget);
+
+      final Finder textFinder = find.byType(Text);
+
+      data.update('increment', 1);
+      await tester.pump();
+      expect(textFinder, findsOneWidget);
+      expect(tester.widget<Text>(textFinder).data, 'Counter: 1');
+
+      await tester.tap(textFinder);
```
Hixie, line 1235: can we check that it also updated the rendering without the tap?

tugorez: Sure thing, done!

packages/rfw/lib/src/dart/text.dart
tugorez: Great call. Fixed!

packages/rfw/lib/src/flutter/runtime.dart
tugorez: ACK. FWIW considered it, tried it but actually prefer this version. Let me know if you feel strongly about it.

packages/rfw/lib/src/dart/model.dart
tugorez: PTAL

packages/rfw/test/runtime_test.dart
```diff
@@ -1088,4 +1088,286 @@ void main() {
     data.update('c', 'test');
     expect(log, <String>['leaf: 2', 'root: {a: [2, 3], b: [q, r]}', 'root: {a: [2, 3], b: [q, r], c: test}']);
   });
+
+  group('Widget Builders', () {
+    late DynamicContent data;
+    late List<RfwEvent> dispatchedEvents;
+
+    Widget rwfWidget(String library, {Map<String, Object?>? initialData}) {
```
Hixie, line 1096: was this meant to be "rwfWidget"?

packages/rfw/test/runtime_test.dart
```diff
@@ -1088,4 +1088,286 @@ void main() {
     data.update('c', 'test');
     expect(log, <String>['leaf: 2', 'root: {a: [2, 3], b: [q, r]}', 'root: {a: [2, 3], b: [q, r], c: test}']);
   });
+
+  group('Widget Builders', () {
+    late DynamicContent data;
+    late List<RfwEvent> dispatchedEvents;
```
Hixie, line 1094: please avoid groups and instead put all the shared state into each test, that way we can much more easily change individual tests, refactor them, move them to other files, etc.

Hixie:
Here's a test I wrote while I was playing with this, might be worth adding (I didn't see anything testing quite everything this test checks):

```dart
    testWidgets('Widget builders - works nested dynamically', (WidgetTester tester) async {
      const LibraryName coreLibraryName = LibraryName(<String>['core']);
      const LibraryName localLibraryName = LibraryName(<String>['local']);
      const LibraryName remoteLibraryName = LibraryName(<String>['remote']);
      final Map<String, VoidCallback> handlers = <String, VoidCallback>{};
      final Runtime runtime = Runtime();
      runtime.update(coreLibraryName, createCoreWidgets());
      runtime.update(localLibraryName, LocalWidgetLibrary(<String, LocalWidgetBuilder> {
        'Builder': (BuildContext context, DataSource source) {
          final String? id = source.v<String>(<String>['id']);
          if (id != null) {
            handlers[id] = source.voidHandler(<String>['handler'])!;
          }
          return source.builder(<String>['builder'], <String, Object?>{
            'param1': source.v<String>(<String>['arg1']),
            'param2': source.v<String>(<String>['arg2']),
          });
        },
      }));
      runtime.update(remoteLibraryName, parseLibraryFile('''
        import core;
        import local;

        widget test { state1: 'strawberry' } = Builder(
          arg1: data.a1,
          arg2: 'apple',
          id: 'A',
          handler: set state.state1 = 'STRAWBERRY',
          builder: (builder1) => Builder(
            arg1: data.b1,
            arg2: 'banana',
            builder: (builder2) => Text(
              textDirection: 'ltr',
              text: [
                state.state1, ' ', builder1.param1, ' ', builder1.param2, ' ', builder2.param1, ' ', builder2.param2,
              ],
            ),
          ),
        );
      '''));
      final DynamicContent data = DynamicContent(<String, Object?>{
        'a1': 'apricot',
        'b1': 'blueberry',
      });
      await tester.pumpWidget(RemoteWidget(
        runtime: runtime,
        data: data,
        widget: const FullyQualifiedWidgetName(remoteLibraryName, 'test'),
      ));
      final Finder textFinder = find.byType(Text);
      expect(tester.widget<Text>(textFinder).data, 'strawberry apricot apple blueberry banana');
      data.update('a1', 'APRICOT');
      await tester.pump();
      expect(tester.widget<Text>(textFinder).data, 'strawberry APRICOT apple blueberry banana');
      data.update('b1', 'BLUEBERRY');
      await tester.pump();
      expect(tester.widget<Text>(textFinder).data, 'strawberry APRICOT apple BLUEBERRY banana');
      handlers['A']!();
      await tester.pump();
      expect(tester.widget<Text>(textFinder).data, 'STRAWBERRY APRICOT apple BLUEBERRY banana');
    });
```

packages/rfw/lib/src/dart/model.dart
```diff
@@ -439,6 +439,28 @@ class ConstructorCall extends BlobNode {
...
 }
 
+/// Representation of functions that return widgets in Remote Flutter Widgets library blobs.
+class WidgetBuilderDeclaration extends BlobNode {
+  /// Represents a callback that takes a single argument [argumentName] and returns the [widget].
+  const WidgetBuilderDeclaration(this.argumentName, this.widget);
+
+  /// The callback single argument name.
+  ///
+  /// In `Builder(builder: (scope) => Container());` [argumentName] is "scope".
```
Hixie, line 449: minor grammar nit, after "in foo" there should be a comma

Hixie: as in:
```suggestion
  /// In `Builder(builder: (scope) => Container());`, [argumentName] is "scope".
```

packages/rfw/lib/src/dart/model.dart
```diff
@@ -534,6 +556,43 @@ class DataReference extends Reference {
   String toString() => 'data.${parts.join(".")}';
 }
 
+/// Reference to the single argument of type [DynamicMap] passed into the widget builder.
+///
+/// This class is used to represent references to a function argument.
+/// In `(scope) => Container(width: scope.width)` this represents "scope.width".
```
Hixie, line 562: similarly,
```suggestion
/// In `(scope) => Container(width: scope.width)`, this represents "scope.width".
```

packages/rfw/lib/src/dart/model.dart
```diff
@@ -534,6 +556,43 @@ class DataReference extends Reference {
...
+///   * [WidgetBuilderDeclaration], which represents a widget builder definition.
+class WidgetBuilderArgReference extends Reference {
+  /// Wraps the given [argumentName] and [parts] as a [WidgetBuilderArgReference].
+  ///
+  /// The parts must not be mutated after the object is created.
+  const WidgetBuilderArgReference(this.argumentName, super.parts);
+
+  /// A reference to a [WidgetBuilderDeclaration.argumentName].
+  ///
+  /// In `Builder(builder: (scope) => Text(text: scope.result.text));`
```
Hixie, line 575:
```suggestion
  /// In `Builder(builder: (scope) => Text(text: scope.result.text));`,
```

packages/rfw/lib/src/flutter/runtime.dart
Hixie: yeah, i often prefer this style too, to be honest.

Hixie:
Coverage is no longer 100% with this PR, please see test_coverage/bin/test_coverage.dart (you can run it using `run_tests.sh`).

Hixie:
This looks great. This is a particularly tricky thing to implement and this seems to be an excellent implementation of it. Just some minor nits, but then I think it should be good to go!

packages/rfw/test/runtime_test.dart
tugorez: Done

packages/rfw/test/runtime_test.dart
tugorez: Great call. Done!

Hixie (approved):
LGTM

PR #5873, opened 2024-01-11
[google_maps_flutter] Drop iOS 11 support

stuartmorgan:
Now that the latest Flutter stable requires iOS 12, and will auto-adjust projects, we can drop iOS 11 support in this plugin without breaking users. Having a minimum of iOS 12 means that Cocoapods should not resolve to anything older than Google Maps 6.x, which is the version that add arm64 simulator support, so the workaround to explicitly not build that can be removed. This means `google_maps_flutter` will no longer force projects into the buggier Rosetta iOS simulator path.

This folds the iOS 11 and iOS 12 examples together, since (per `examples/README.md`) the iOS 11 version is the one containing most of the tests. To minimize the potential of losing tests due to mistakes, this was done by deleting the iOS *12* example, renaming the iOS 11 example, and then updating it to a minimum of 12.

Fixes https://github.com/flutter/flutter/issues/94491

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[...]

PR #5913, opened 2024-01-17
[ci] Work around Google Maps podspec lint issue

stuartmorgan:
https://github.com/flutter/packages/pull/5873 removed a plugin-level workaround for a `podspec-lint` issue, under the incorrect belief that it was no longer necessary. Rather than re-add that, which degrades the experience for plugin clients who are targeting iOS 13+ with their apps, this adds a hack in the CI tooling itself to modify the podspec to require iOS 13 before linting.

This could in theory cause false-positive lints about deprecated APIs, but currently doesn't, and that could be addressed in the unlikely event that it comes up.

See https://github.com/flutter/flutter/issues/94491

### DIFF
--- a/script/tool/lib/src/podspec_check_command.dart
+++ b/script/tool/lib/src/podspec_check_command.dart
@@ -128,6 +128,19 @@ class PodspecCheckCommand extends PackageLoopingCommand {
     final String podspecBasename = podspec.basename;
     print('Linting $podspecBasename');
 
+    // Hack the google_maps_flutter plugin's podspec to require iOS 13+, so that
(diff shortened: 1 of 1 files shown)


stuartmorgan:
test-exempt: is a test

(I did consider writing unit tests for this, but since the *only* purpose of this code is to make https://github.com/flutter/flutter/issues/141493 work, I don't see any value to doing so: either it works and the CI we need is green, or it doesn't and the CI we need will be red.)

stuartmorgan:
Never mind; we have to put the plugin workaround back. I'll update the issue.

PR #5800, opened 2024-01-04
[video_player_web] migrates to package:web

balvinderz:
Updates the web implementation of `video_player_web` to `package:web`.

Also: prevents an infinite event loop when seeking to the end of a video after it's done.

### Issues

* Fixes: https://github.com/flutter/flutter/issues/139752

Co-authored-by: ToddZeil <120418414+ToddZeil@users.noreply.github.com>

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [X] I signed the [CLA].
- [X] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [X] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making, or this PR is [test-exempt].
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #5920, opened 2024-01-18
[video_player] Fix pause infinite loop on completed

ToddZeil:
When a video is complete, and not on a loop setting.
Pause() function is called even if video is already paused. 

(_applyPlayPause() function is used for play and pause which can lead to playing a video when attempting to restart) 

Fixes https://github.com/flutter/flutter/issues/143141

Edited: Also this may only be an issue on web

## Pre-launch Checklist

- [ y] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ y] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ y] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [y ] I signed the [CLA].
- [y ] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ y] I listed at least one issue that this PR fixes in the description above.
- [y ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [y ] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [not needed ] I updated/added relevant documentation (doc comments with `///`).
- [fixed curren, added one more test ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [Y, one fixed test ] All existing and new tests are passing.


packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 ## NEXT
 
+* Description of the change.
+
```
tarrinneal, line 4: This should be removed

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
...
+## 2.8.3
+
 * Updates support matrix in README to indicate that iOS 11 is no longer supported.
 * Clients on versions of Flutter that still support iOS 11 can continue to use this
   package with iOS 11, but will not receive any further updates to the iOS implementation.
+* Fix infinite pause loop caused by on completed to call pause even when video is pause.
```
tarrinneal, line 10: Fixes infinite pause loop caused by call to pause even when video is already paused.

tarrinneal (changes requested):
A couple nits here.

I'm not certain if this will break the work-around for the video completing but not truly reaching the last frame or not. Maybe @stuartmorgan has some insight.

ToddZeil:
> A couple nits here.
> 
> I'm not certain if this will break the work-around for the video completing but not truly reaching the last frame or not. Maybe @stuartmorgan has some insight.

There are a couple of other ways we could do this if your not sure and think they will be better. 
Either, split up the play pause into two different functions, or only call the pause part of the function if the timer is not null

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -460,7 +460,9 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
           // position=value.duration. Instead of setting the values directly,
           // we use pause() and seekTo() to ensure the platform stops playing
           // and seeks to the last frame of the video.
-          pause().then((void pauseResult) => seekTo(value.duration));
+          if (value.isPlaying) {
+            pause().then((void pauseResult) => seekTo(value.duration));
+          }
```
stuartmorgan, line 465: This changed logic would definitely not be guaranteed to actually do what this code is supposed to be doing; if the native side reported reaching the end as:
- playback stopped
- video completed

then the seek code would never run.

ToddZeil: @stuartmorgan have updated the logic to account for this

tarrinneal (approved):
This change tracks for me, approved pending @stuartmorgan review.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -1246,7 +1246,7 @@ void main() {
 
     bool currentIsCompleted = controller.value.isCompleted;
 
-    final void Function() isCompletedTest = expectAsync0(() {}, count: 2);
+    final void Function() isCompletedTest = expectAsync0(() {});
```
stuartmorgan, line 1249: It is unclear to me why this is changing, or how this test is related to the behavior your PR is intended to fix. This PR should be adding one or more new tests that clearly cover the specific behavior you are fixing.

stuartmorgan (changes requested):

packages/video_player/video_player/test/video_player_test.dart
ToddZeil: @stuartmorgan updated that test and added one new test

ToddZeil: @stuartmorgan turns out that test was failing as was not in a play state as test implied

ToddZeil: @stuartmorgan are these changes okay ?

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -1267,6 +1267,50 @@ void main() {
       }
     });
 
+    controller.value = controller.value.copyWith(isPlaying: true);
+    fakeVideoEventStream.add(VideoEvent(eventType: VideoEventType.completed));
+  });
+
+  test(
+      'isCompleted seeks position to max duration and pauses the video first if is playing',
```
stuartmorgan, line 1275: Isn't the thing you want to fix that you want to *not* call pause if it's *not* playing? That path would also need a test.

stuartmorgan (changes requested):
Looking at the code more deeply here, it's not clear to me what the reason for this change is. Why is calling `pause` while paused a problem? The PR description just says that it's a problem because it calls `_applyPlayPause()`, but with the paused value state, that will call `_videoPlayerPlatform.pause`. If calling `pause` on the platform object will start *playing* a video, that seems like a platform bug.

Could you file an issue with clear repro steps and an explanation of what exactly is happening?

stuartmorgan:
Thanks for filing the issue. Could you elaborate, either here or (preferably) in the issue about what exactly is going wrong? It's not clear to me how the change in this PR relates to the problem shown in the issue. What exactly is happening in the web implementation that is causing the incorrect behavior?

ToddZeil:
> Thanks for filing the issue. Could you elaborate, either here or (preferably) in the issue about what exactly is going wrong? It's not clear to me how the change in this PR relates to the problem shown in the issue. What exactly is happening in the web implementation that is causing the incorrect behavior?

The listener when completed  goes into an infinite loop which continuously calls any functions in the onCompleted switch case statement.
Specifically this is repeatedly called
`  case VideoEventType.completed:
          // In this case we need to stop _timer, set isPlaying=false, and
          // position=value.duration. Instead of setting the values directly,
          // we use pause() and seekTo() to ensure the platform stops playing
          // and seeks to the last frame of the video.
          pause().then((void pauseResult) => seekTo(value.duration));
          value = value.copyWith(isCompleted: true);`

stuartmorgan:
> The listener when completed goes into an infinite loop which continuously calls any functions in the onCompleted switch case statement.

It's not clear to me why this would be the case. Could you provide the specific call sequence in the infinite loop? In other words, when you say:

> Specifically this is repeatedly called

what code is repeatedly calling that?

ToddZeil:
> > The listener when completed goes into an infinite loop which continuously calls any functions in the onCompleted switch case statement.
> 
> It's not clear to me why this would be the case. Could you provide the specific call sequence in the infinite loop? In other words, when you say:
> 
> > Specifically this is repeatedly called
> 
> what code is repeatedly calling that?

This is the code which seems to be repeatedly calling it

/.pub-cache/hosted/pub.dev/video_player_web-2.1.3/lib/src/video_player.dart

`    _videoElement.onEnded.listen((dynamic _) {
      setBuffering(false);
      _eventController.add(VideoEvent(eventType: VideoEventType.completed));
    });`

stuartmorgan:
It seems like more root-causing is still needed here. As far as I can see the only change the current version of the PR makes is not calling `pause()` when paused, but `pause()` on web just calls `pause()` on the `<video>` element, and according to MDN:
> The HTMLMediaElement.pause() method will pause playback of the media, if the media is already in a paused state this method will have no effect.

Are those docs incorrect? Is calling `pause()` when already paused at the end of the video triggering an `onEnded` event? That seems very surprising if so, but if not I don't see how this PR would fix the behavior.

ToddZeil:
The infinite loop must be coming from the html code

But I believe this issue is coming from the _applyPlayPause function in the
video_player

Which while in the infinite loop, will actually instantly cancel the timer
which is used for the video playing

stuartmorgan:
So this PR doesn't fix the infinite loop, just removes one side effect of the infinite loop?

If there's an infinite loop, we should root cause and fix the loop itself, not the side effect.

ToddZeil:
Yes that's correct,
Olay I will need to look into that then as from a quick look I couldn't
work out where the 'ended' event was getting added to the video event
stream in the web package.

ToddZeil:
The loop is coming from 
'flutter/flutter/bin/cache/pkg/sky_engine/lib/html/html_dart2js.dart'
Specifically: `ElementStream<Event> get onEnded => endedEvent.forElement(this);`


This is used in video_player_web

`    _videoElement.onEnded.listen((dynamic _) {
      setBuffering(false);
       _eventController.add(VideoEvent(eventType: VideoEventType.completed));
    });`
    
    The interesting thing is sometimes the video replays and the stream continues to be called as completed even after the video has replayed
    
   I can't find the code where the event is added to the stream, I'm not sure this is even in our control ?

stuartmorgan:
The creation of the event is done by the browser, but it is likely the result of something the plugin is doing; it's extremely unlikely that browser behavior is to create an infinite stream of `onEnded` events without any input.

Is seeking to the end when already at the end creating a new `onEnded`?

ToddZeil:
You must be right, as I comment out the code that seeks to end `onCompleted`

And `onEnded` event only happens the once on ended.

ToddZeil:
So we could keep PR how it is 

Or update the seek to function like so:
` Future<void> seekTo(Duration position) async {
    if (_isDisposedOrNotInitialized || position == value.duration) {
      return;
    }`

stuartmorgan:
> So we could keep PR how it is

The PR as it is now does't fix the infinite loop.

> Or update the seek to function like so: ` Future<void> seekTo(Duration position) async { if (_isDisposedOrNotInitialized || position == value.duration) { return; }`

Have you verified that this doesn't regress the cases that the code was added to handle? `value` does not alway reflect the actual underlying position, so if the issue is that `seekTo` causes problems on web in this scenario, a much safer fix would be to make seeking a no-op at the platform implementation level if the target value is the same as the actual, live, native position.

ToddZeil:
The current PR does stop the infinite loop as it doesn't call seek to if the video is paused.
We just didn't understand why the infinite loop was occurring, before we dived into it.

But there is potential to fix this via `video_player_web` `seekTo`
By checking if the ` _videoElement.currentTime ` is the same time as `position` already.

stuartmorgan:
> The current PR does stop the infinite loop as it doesn't call `seekTo` if the video is paused.

This is likely to regress other edge cases, per my previous comment. (Sorry, I thought you mean keep the earlier version of the PR.)

> Or update the seek to function like so: ` Future<void> seekTo(Duration position) async { if (_isDisposedOrNotInitialized || position == value.duration) { return; }`

This change has the same problem, it just moves it to a different part of the code.

The place to fix a web-specific issue in the seek behavior here is in the web implementation, where we have access to the live duration information and we can be entirely confident that it won't regress the issue this code was added to fix in the first place (on mobile).

ToddZeil:
I did mean the PR in the current state, forgot you mentioned the possibility of regression though.
What were your concerns here? as completed should still get completed once while in the played state


There is an issue with doing through `video_player_web`
Is that the `position` times and `videoElement` times don't actually match up
So we would have to convert the `video_player_web`  `videoElement` into milliseconds as type `int` 
Would this cause any other concerns?

`  void seekTo(Duration position) {
    assert(!position.isNegative);
    if (position.inMilliseconds.toDouble() / 1000 ==
        _videoElement.currentTime) {
      return;
    }
    _videoElement.currentTime = position.inMilliseconds.toDouble() / 1000;
    debugPrint('position: ${position.inMilliseconds}' );
    debugPrint('currentTime: ${_videoElement.currentTime}' );
    debugPrint('duration: ${_videoElement.duration}'  );
  }
`

position: 4037
currentTime: 4.036666
duration: 4.036667

stuartmorgan:
> What were your concerns here? as completed should still get completed once while in the played state

See [this comment](https://github.com/flutter/packages/pull/5920#discussion_r1457776375)

> There is an issue with doing through `video_player_web`
> Is that the `position` times and `videoElement` times don't actually match up
> So we would have to convert the `video_player_web` `videoElement` into milliseconds as type `int`
> Would this cause any other concerns?

I'm not sure I understand the question. How would that be different than [the current code](https://github.com/flutter/packages/blob/abb04bb89cca97e653764a60bde0ee4e816db322/packages/video_player/video_player_web/lib/src/video_player.dart#L214-L218)?

ToddZeil:
I did add the fix for checking positions as so
` else if (value.position != value.duration) {
            seekTo(value.duration);
          }`
But I think you mentioned maybe the value might not be exactly what we need
          
Updated the PR now if you want to check this solution

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -1267,6 +1267,7 @@ void main() {
       }
     });
 
+    controller.value = controller.value.copyWith(isPlaying: true);
```
ditman, line 1270: Is this needed?

ditman (commented):
I think I've minimally reproduced the issue in JS, here:

* https://jsfiddle.net/ditman/nmwvLah8/

Are the changes to `video_player` needed? This is only a fix to `video_player_web`, right?

packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -205,8 +205,14 @@ class VideoPlayer {
   /// Moves the playback head to a new `position`.
   ///
   /// `position` cannot be negative.
+  ///
+  /// Does nothing if the new position is the same as the current position.
   void seekTo(Duration position) {
     assert(!position.isNegative);
+    // Don't seek if video is already at position.
+    if (position.inMilliseconds == (_videoElement.currentTime * 1000).round()) {
```
stuartmorgan, line 213: The common code with `getPosition()` should be extracted to a shared private helper so that they are guaranteed to stay in sync, since what's important here is that it matches what the higher-level code got from `getPosition`.

packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -205,8 +205,14 @@ class VideoPlayer {
   /// Moves the playback head to a new `position`.
   ///
   /// `position` cannot be negative.
+  ///
+  /// Does nothing if the new position is the same as the current position.
   void seekTo(Duration position) {
     assert(!position.isNegative);
+    // Don't seek if video is already at position.
```
stuartmorgan, line 212: This should explain why.

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
-## NEXT
+## 2.8.3
```
stuartmorgan, line 1: The changes to this package should be reverted; we don't artificially constrain allowed versions in the app-facing packages.

stuartmorgan (changes requested):

ToddZeil:
PR has been updated if you can check again, when you can

packages/video_player/video_player/CHANGELOG.md
stuartmorgan: This has not been addressed.

packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -205,16 +205,27 @@ class VideoPlayer {
...
   }
 
   /// Returns the current playback head position as a [Duration].
   Duration getPosition() {
     _sendBufferingRangesUpdate();
-    return Duration(milliseconds: (_videoElement.currentTime * 1000).round());
+    return Duration(milliseconds: _getVideoCurrentTimeInMilliseconds());
+  }
+
+  int _getVideoCurrentTimeInMilliseconds() {
```
stuartmorgan, line 227: This can use `get` rather than be a method, as it's only doing trivial calculation and has no side effects.

ToddZeil:
Thanks, have updated now

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Updates support matrix in README to indicate that iOS 11 is no longer supported.
 * Clients on versions of Flutter that still support iOS 11 can continue to use this
   package with iOS 11, but will not receive any further updates to the iOS implementation.
+* Web, Fix infinite pause loop caused by seekTo marking the video as completed when already completed.
```
ditman, line 6: Please, revert all changes to the `video_player` package, they're not needed.

`flutter pub get` (or `flutter pub upgrade`) will take care of fetching the fixed version of `video_player_web` once the fix is published.

(Sorry I wasn't clearer before!)

ToddZeil:
Thanks for all the help guys, should be all done now!

packages/video_player/video_player_web/CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
-## NEXT
+## 2.1.4
 
+* Fixes infinite pause loop caused by seekTo marking the video as completed when already completed.
```
stuartmorgan, line 3: Nit: Please use backticks on `seekTo` to mark it as code.

packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -205,18 +205,29 @@ class VideoPlayer {
   /// Moves the playback head to a new `position`.
   ///
   /// `position` cannot be negative.
+  ///
+  /// Does nothing if the new position is the same as the current position.
   void seekTo(Duration position) {
     assert(!position.isNegative);
+    // Don't seek if video is already at position.
+    // As seeking when completed will trigger another completed event ('onEnded')
```
stuartmorgan, line 213: This is a sentence fragment; please combine it with the line above: "Don't seek if the video is already at the target position, as seeking when completed [...]"

packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -205,18 +205,29 @@ class VideoPlayer {
...
+    // to avoid potentially firing extra 'onEnded' events when the video is already over
```
stuartmorgan, line 214: "when the video is already over" is redundant with the earlier "when completed", so should be removed.

stuartmorgan, line 214: This needs a period at the end of the sentence, since comments should be properly punctuated sentences.

stuartmorgan (changes requested):
(Fixing review state; I thought I could clear my request without approving, but apparently not.)

ditman:
Adding a small commit here to address @stuartmorgan's concerns, and get the branch updated with the latest `main`.

packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -205,18 +205,29 @@ class VideoPlayer {
   /// Moves the playback head to a new `position`.
   ///
   /// `position` cannot be negative.
+  ///
+  /// Does nothing if the new position is the same as the current position.
   void seekTo(Duration position) {
     assert(!position.isNegative);
+    // Don't seek if video is already at target position
+    // as seeking when completed will trigger another completed event ('onEnded').
+    // This avoids potentially firing extra 'onEnded' events.
```
ditman, line 214:
```suggestion
    // Don't seek if video is already at target position.
    //
    // This is needed because the core plugin will pause and seek to the end of
    // the video when it finishes, and that causes an infinite loop of `ended`
    // events on the web.
    //
    // See: https://github.com/flutter/flutter/issues/77674
```

ditman (commented):
We need a test, I think we can write one easily by listening to the events coming from the videoplayer (two seeks to the same position should only yield a single seek event to the core plugin)

ToddZeil:
Okay where do we put tests for this package ? 
Will it go in here `packages/packages/video_player/video_player/test/video_player_test.dart` ?

ditman:
@ToddZeil there's already several tests here:

* https://github.com/flutter/packages/tree/main/packages/video_player/video_player_web/example/integration_test

TL;DR: Assuming that `chromedriver` and `dart` are in your `$PATH`, and Chrome is in a standard location, inside the `video_player_web` directory, you can do something like this:

```bash
PLUGIN_TOOLS=/work/flutter/packages/script/tool/bin/flutter_plugin_tools.dart # Adjust this
dart $PLUGIN_TOOLS drive-examples --current-package --web --run-chromedriver
```

(I normally use Linux)

There's more instructions on how to run tests in the [README.md of the example](https://github.com/flutter/packages/tree/main/packages/video_player/video_player_web/example#readme)

ToddZeil:
I have set up the integration tests, not sure exactly how these work but the current seekTo doesn't seem to check for anything other than completion ?

`testWidgets('can seek to position', (WidgetTester tester) async {
      expect(
        VideoPlayerPlatform.instance.seekTo(
          await textureId,
          const Duration(seconds: 1),
        ),
        completes,
      );
    });`
    
    Adding seekTo seems to add the following events to the tests:
    ` VideoEventType:VideoEventType.bufferingStart,
            VideoEventType:VideoEventType.bufferingUpdate,
            VideoEventType:VideoEventType.isPlayingStateUpdate,
            VideoEventType:VideoEventType.bufferingEnd,
            VideoEventType:VideoEventType.completed
`

But when I comment out the new code it doesn't seem to effect anything with in the test?
It doesn't seem to call completed multiple times?
Could I be doing something wrong?

ditman:
> Could I be doing something wrong?

@ToddZeil can you share the code of the test that you've written? Or I can try and write one tomorrow :)

My plan is to:

* start a video paused
* seek to a position
* assert the events (probably `seeked` or some combination of `buffering`?)
* seek to the same position
* assert that now new events fire in a reasonable amount of time (500ms or similar).

I that test should fail if we comment out the fix (?)

ToddZeil:
Your test sounds like it should work! 
The main issue was happening when looping was off and `seekTo` was seeking to complete the video

I was just trying to find out what events the `seekTo` function emitted, but I couldn't get multiple events firing through

This is the test I was trying:

`testWidgets('multiple calls to seekTo will emit a seek event',
        (WidgetTester tester) async {
      final int videoPlayerId = await textureId;
      final Stream<VideoEvent> eventStream =
          VideoPlayerPlatform.instance.videoEventsFor(videoPlayerId);

      final Future<List<VideoEvent>> stream = eventStream.timeout(
        const Duration(seconds: 2),
        onTimeout: (EventSink<VideoEvent> sink) {
          sink.close();
        },
      ).toList();

      await VideoPlayerPlatform.instance.setLooping(videoPlayerId, false);
      await VideoPlayerPlatform.instance.setVolume(videoPlayerId, 0);

      await VideoPlayerPlatform.instance.play(videoPlayerId);


      await VideoPlayerPlatform.instance
          .seekTo(await textureId, const Duration(milliseconds: 300));

      await VideoPlayerPlatform.instance
          .seekTo(await textureId, const Duration(milliseconds: 350));

      await VideoPlayerPlatform.instance
          .seekTo(await textureId, const Duration(milliseconds: 350));

      await VideoPlayerPlatform.instance
          .seekTo(await textureId, const Duration(seconds: 99));

      await VideoPlayerPlatform.instance
          .seekTo(await textureId, const Duration(seconds: 99));

      await VideoPlayerPlatform.instance
          .seekTo(await textureId, const Duration(milliseconds: 100));

      // Let the video play, until we stop seeing events for two seconds
      final List<VideoEvent> events = await stream;

      await VideoPlayerPlatform.instance.pause(videoPlayerId);

      // The expected list of event types should look like this:
      // 1. isPlayingStateUpdate (videoElement.onPlaying)
      // 2. bufferingStart,
      // 3. bufferingUpdate (videoElement.onWaiting),
      // 4. initialized (videoElement.onCanPlay),
      // 5. bufferingEnd (videoElement.onCanPlayThrough),
      expect(
          events.map((VideoEvent e) => e.eventType),
          equals(<VideoEventType>[
            VideoEventType.isPlayingStateUpdate,
            VideoEventType.bufferingStart,
            VideoEventType.bufferingUpdate,
            VideoEventType.initialized,
            VideoEventType.bufferingEnd,
          ]));
    });`

ToddZeil:
Did get a chance to look into this test ? Or have further hints or advice ? Thanks

ditman:
@ToddZeil I didn't have time to look at the test, but I'm going to work next in migrating the `video_player_web` to the latest `package:web`, so I can sneak your fix there :)

The problem is that we don't have any "seek" event that we can check that wasn't propagated:

* https://github.com/flutter/packages/blob/main/packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart#L279

(We could maybe inject a fake video player and ensure we don't call "seek" multiple times, but I'm not sure... I'll try to add you as a Co-Author in the migration PR!)

ToddZeil:
Okay interesting!

Cool hopefully we can get it in, thanks for that :)

ditman:
This fix has landed alongside #5800! Closing.

ToddZeil:
Great thank you!

PR #5927, opened 2024-01-18
[camera]fix a sample buffer memory leak on pause resume recording

hellohuanlin:
This memory leak happens when pause/resume the recording. Depending on camera resolution, about 1-2MB leaks for every resume, so it can add up pretty fast if we frequently pause/resume. 

We only reference this sample buffer inside the scope of this method (e.g. directly piping into file IO), so no need to retain it again, as it's already retained and will be released afterwards by apple. (We did keep around the pixel buffer, but we already manually [retain](https://github.com/flutter/packages/blob/d7ee75ad59ad7bc45e659d0599e935e9e7981ea1/packages/camera/camera_avfoundation/ios/Classes/FLTCam.m#L403) and [release](https://github.com/flutter/packages/blob/d7ee75ad59ad7bc45e659d0599e935e9e7981ea1/packages/camera/camera_avfoundation/ios/Classes/FLTCam.m#L415) that)

Bug was introduced in https://github.com/flutter/plugins/pull/1370. 

https://github.com/flutter/flutter/issues/132013

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
+++ b/packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
@@ -514,7 +514,6 @@ NSString *const errorMethod = @"error";
       return;
     }
 
-    CFRetain(sampleBuffer);
     CMTime currentSampleTime = CMSampleBufferGetPresentationTimeStamp(sampleBuffer);
 
     if (_videoWriter.status != AVAssetWriterStatusWriting) {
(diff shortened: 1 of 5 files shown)


packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -564,18 +563,19 @@ - (void)captureOutput:(AVCaptureOutput *)output
...
-      [self newAudioSample:sampleBuffer];
     }
-
-    CFRelease(sampleBuffer);
   }
 }
 
-- (CMSampleBufferRef)adjustTime:(CMSampleBufferRef)sample by:(CMTime)offset CF_RETURNS_RETAINED {
+- (CMSampleBufferRef)copySampleBufferWithAdjustedTime:(CMSampleBufferRef)sample
+                                                   by:(CMTime)offset CF_RETURNS_RETAINED {
```
hellohuanlin, line 578: just wanted to put the word "copy" in method name, so the caller remembers to release it

jmagman, line 578: Is `CF_RETURNS_RETAINED` needed with the `copy` rename?  I thought that was only needed if it was a `copy` but the name didn't include it.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -514,7 +514,6 @@ - (void)captureOutput:(AVCaptureOutput *)output
       return;
     }
 
-    CFRetain(sampleBuffer);
```
jmagman, line 517: Is the bug that this is retained but not released before the `if(_audioIsDisconnected)` early returns?

jmagman: (I think you're right it doesn't need to be retained at all since `sampleBuffer` isn't used outside the scope of this method)
>     Clients that need to reference the CMSampleBuffer object outside of the scope of this method must CFRetain it and then CFRelease it when they are finished with it.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
hellohuanlin: you are right. let me remove it. just wanted to put `copy` in the name so +1 is clear from caller site

jmagman (approved):

PR #5928, opened 2024-01-18
[ci] Run Swift formatter and linter during CI formatting

jmagman:
Make `swift-format` CIPD package available on the `PATH` https://flutter-review.googlesource.com/c/recipes/+/54020

Run `format --no-clang-format --no-java --no-kotlin --no-dart`  on the macOS builder so it doesn't duplicate same `format` call run on Linux.

Filter out generated files until https://github.com/flutter/flutter/issues/141799 is done.

Also add `swift-format lint` call during `format` command.  Fix the one casing issue it found in a test file.

Failing run:
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8758535881172004177/+/u/Run_package_tests/Swift_format/stdout

Successful run:
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8758492350529121249/+/u/Run_package_tests/Swift_format/stdout

Fixes https://github.com/flutter/flutter/issues/41129

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -1071,7 +1071,11 @@ targets:
     properties:
       add_recipes_cq: "true"
       version_file: flutter_master.version
-      target_file: macos_check_podspecs.yaml
+      target_file: macos_repo_checks.yaml
+      dependencies: >
+        [
+          {"dependency": "swift_format", "version": "build_id:8797338980206841409"}
+        ]
 
   ### macOS desktop tasks ###
   # macos-platform_tests builds all the packages on ARM, so this build is run
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -1147,6 +1147,10 @@ targets:
         {
           "CHANNEL": "master"
         }
+      dependencies: >
+        [
+          {"dependency": "swift_format", "version": "build_id:8797338979890974865"}
+        ]
 
   - name: Mac_arm64 custom_package_tests stable
     recipe: packages/packages
@@ -1160,6 +1164,10 @@ targets:
         {
           "CHANNEL": "stable"
         }
+      dependencies: >
+        [
+          {"dependency": "swift_format", "version": "build_id:8797338979890974865"}
+        ]
 
   ### iOS tasks ###
   # ios_platform_tests builds all the packages on ARM, so this build is run
--- a/.ci/targets/macos_repo_checks.yaml
+++ b/.ci/targets/macos_repo_checks.yaml
@@ -7,7 +7,7 @@ tasks:
     infra_step: true # Note infra steps failing prevents "always" from running.
   - name: Swift format
     script: script/tool_runner.sh
-    args: ["format", "--fail-on-change", "--swift", "--no-dart", "--no-clang-format", "--no-kotlin", "--no-java" ]
+    args: ["format", "--fail-on-change", "--no-dart", "--no-clang-format", "--no-kotlin", "--no-java" ]
     always: true
   - name: validate iOS and macOS podspecs
     script: script/tool_runner.sh
--- a/.ci/targets/macos_repo_checks.yaml
+++ b/.ci/targets/macos_repo_checks.yaml
@@ -7,8 +7,11 @@ tasks:
     infra_step: true # Note infra steps failing prevents "always" from running.
   - name: Swift format
     script: script/tool_runner.sh
+    # Non-Swift languages are formatted on Linux builders.
+    # Skip them on Mac builder to avoid duplication.
     args: ["format", "--fail-on-change", "--no-dart", "--no-clang-format", "--no-kotlin", "--no-java" ]
     always: true
   - name: validate iOS and macOS podspecs
     script: script/tool_runner.sh
     args: ["podspec-check"]
+    always: true
--- a/.ci/targets/macos_repo_checks.yaml
+++ b/.ci/targets/macos_repo_checks.yaml
@@ -5,6 +5,10 @@ tasks:
   - name: update pods repo
     script: .ci/scripts/update_pods.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
+  - name: Swift format
+    script: script/tool_runner.sh
+    args: ["format", "--fail-on-change", "--swift", "--no-dart", "--no-clang-format", "--no-kotlin", "--no-java" ]
+    always: true
   - name: validate iOS and macOS podspecs
     script: script/tool_runner.sh
     args: ["podspec-check"]
--- a/packages/file_selector/file_selector_macos/example/macos/RunnerTests/RunnerTests.swift
+++ b/packages/file_selector/file_selector_macos/example/macos/RunnerTests/RunnerTests.swift
@@ -40,7 +40,7 @@ class TestViewProvider: NSObject, ViewProvider {
   var window: NSWindow? = NSWindow()
 }
 
-class exampleTests: XCTestCase {
+class ExampleTests: XCTestCase {
 
   func testOpenSimple() throws {
     let panelController = TestPanelController()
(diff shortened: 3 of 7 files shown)


.ci.yaml
```diff
@@ -1071,7 +1071,11 @@ targets:
     properties:
       add_recipes_cq: "true"
       version_file: flutter_master.version
-      target_file: macos_check_podspecs.yaml
+      target_file: macos_repo_checks.yaml
+      dependencies: >
+        [
+          {"dependency": "swift_format", "version": "build_id:8797338979890974865"}
```
jmagman, line 1077: The `version` tag is `version:2@505.chromium.2` which fails because the `@` isn't valid in the regex `^[a-z0-9_]+$` 
```
 ERROR: projects/flutter: cr-buildbucket.cfg: in "cr-buildbucket.cfg" (buckets #4 - try / swarming / builders #454 - Mac_x64 check_podspecs / caches #4): name: "packages_main_swift_format_version_2@505_chromium_2" does not match "^[a-z0-9_]+$"
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8758535881024973153/+/u/luci_validate/stdout

I don't think it's important which tag we use, `build_id` seems fine too.

jmagman: Actually it does matter because the tags are different on arm and x64.  I'll see if I can get the 3pp folks to push a shared version tag that matches the regex.

packages/file_selector/file_selector_macos/example/macos/RunnerTests/RunnerTests.swift
```diff
@@ -40,7 +40,7 @@ class TestViewProvider: NSObject, ViewProvider {
   var window: NSWindow? = NSWindow()
 }
 
-class exampleTests: XCTestCase {
+class ExampleTests: XCTestCase {
```
jmagman, line 43: linter (not format) error:
```
RunnerTests.swift:43:7: warning: [TypeNamesShouldBeCapitalized] rename the class 'exampleTests' using UpperCamelCase; for example, 'ExampleTests'
```

script/tool/lib/src/format_command.dart
```diff
@@ -342,6 +359,13 @@ class FormatCommand extends PackageCommand {
         (String filePath) => extensions.contains(path.extension(filePath)));
   }
 
+  Iterable<String> _filterGeneratedFiles(Iterable<String> files) {
```
stuartmorgan, line 362: Rather than filtering out generated files, we just need Pigeon to do Swift formatting in CI; [this code](https://github.com/flutter/packages/blob/129e08c1dff159b1b49484e1f838f0a471273cd7/packages/pigeon/tool/shared/generation.dart#L239-L246) should include Swift formatting. If we changed the default then it'll just work, otherwise we need to add `--swift` there.

stuartmorgan (changes requested):

script/tool/lib/src/format_command.dart
stuartmorgan: Oh, but that's not going to work on the Linux machine where we run this.

Hm, do we have a `clang-format` CIPD package for macOS already? If so, we could change [this](https://github.com/flutter/packages/blob/129e08c1dff159b1b49484e1f838f0a471273cd7/packages/pigeon/tool/run_tests.dart#L211) to macOS instead of Linux, and adjust the dependencies in ci.yaml. If not I'll need to think more about how we can handles this sanely, and we can land with the exclusion in the meantime.

stuartmorgan: Maybe it's easier to change the Pigeon validator to check specific languages for changes instead of checking everything for changes, and then run it for different languages on different CI hosts the way we do the `format` check itself. I'll play with that today.

stuartmorgan: > Maybe it's easier to change the Pigeon validator to check specific languages for changes instead of checking everything for changes, and then run it for different languages on different CI hosts the way we do the `format` check itself.

I think https://github.com/flutter/packages/pull/5944 should make this work.

.ci.yaml
```diff
@@ -1071,7 +1071,11 @@ targets:
     properties:
       add_recipes_cq: "true"
       version_file: flutter_master.version
-      target_file: macos_check_podspecs.yaml
+      target_file: macos_repo_checks.yaml
+      dependencies: >
+        [
+          {"dependency": "swift_format", "version": "build_id:8797338980206841409"}
```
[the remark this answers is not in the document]
jmagman: Done.

.ci.yaml
```diff
@@ -1143,6 +1147,10 @@ targets:
         {
           "CHANNEL": "master"
         }
+      dependencies: >
+        [
+          {"dependency": "swift_format", "version": "build_id:8797338979890974865"}
```
jmagman, line 1152: The arm64 and x64 `build_id` numbers are different, but it's pointing to the same version.
https://github.com/flutter/packages/pull/5928#discussion_r1457919953

.ci/targets/macos_repo_checks.yaml
```diff
@@ -5,6 +5,10 @@ tasks:
   - name: update pods repo
     script: .ci/scripts/update_pods.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
+  - name: Swift format
+    script: script/tool_runner.sh
+    args: ["format", "--fail-on-change", "--swift", "--no-dart", "--no-clang-format", "--no-kotlin", "--no-java" ]
```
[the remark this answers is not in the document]
jmagman: Done.

script/tool/lib/src/format_command.dart
jmagman: I changed my comment about removing it into a TODO.  How about I land this, then you rebase https://github.com/flutter/packages/pull/5944 onto ToT and remove the `_filterGeneratedFiles` in that PR?

.ci/targets/macos_repo_checks.yaml
```diff
@@ -5,6 +5,10 @@ tasks:
   - name: update pods repo
     script: .ci/scripts/update_pods.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
+  - name: Swift format
+    script: script/tool_runner.sh
+    args: ["format", "--fail-on-change", "--no-dart", "--no-clang-format", "--no-kotlin", "--no-java" ]
+    always: true
   - name: validate iOS and macOS podspecs
```
stuartmorgan, line 12: This should get an `always` too.

.ci/targets/macos_repo_checks.yaml
```diff
@@ -5,6 +5,10 @@ tasks:
   - name: update pods repo
     script: .ci/scripts/update_pods.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
+  - name: Swift format
+    script: script/tool_runner.sh
+    args: ["format", "--fail-on-change", "--no-dart", "--no-clang-format", "--no-kotlin", "--no-java" ]
```
stuartmorgan, line 10: It's probably worth a comment saying that these languages are done in `repo_checks.yaml`, so are excluded here to avoid duplication.

script/tool/lib/src/format_command.dart
stuartmorgan: 👍🏻

stuartmorgan (approved):
LGTM with nit

.ci/targets/macos_repo_checks.yaml
jmagman: Done.

.ci/targets/macos_repo_checks.yaml
jmagman: Done.

script/tool/lib/src/format_command.dart
stuartmorgan: > Rather than filtering out generated files, we just need Pigeon to do Swift formatting in CI; [this code](https://github.com/flutter/packages/blob/129e08c1dff159b1b49484e1f838f0a471273cd7/packages/pigeon/tool/shared/generation.dart#L239-L246) should include Swift formatting.

Oops, now that `swift` is enabled by default, that same code will need (until my PR lands) `--no-swift`.

PR #5936, opened 2024-01-19
Expose registered widget libraries and local widget library widgets.

johnmccutchan:
This exposes the necessary information to allow rfwpg to display registered libraries and widgets in a list or tree.

### DIFF
--- a/packages/rfw/CHANGELOG.md
+++ b/packages/rfw/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.0.18
+
+* Exposes `WidgetLibrary`s registered in `Runtime`.
+* Exposes widgets map in `LocalWidgetLibrary`.
+
 ## 1.0.17
 
 * Adds support for tracking source locations of `BlobNode`s and
--- a/packages/rfw/lib/src/flutter/runtime.dart
+++ b/packages/rfw/lib/src/flutter/runtime.dart
@@ -4,10 +4,13 @@
 
 // This file is hand-formatted.
 
+import 'dart:collection';
+
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 
-import '../dart/model.dart';
+import '../../formats.dart';
+
 import 'content.dart';
 
 /// Signature of builders for local widgets.
@@ -169,6 +172,22 @@ class LocalWidgetLibrary extends WidgetLibrary {
   LocalWidgetBuilder? findConstructor(String name) {
     return _widgets[name];
   }
+
+  /// The widgets defined by this [LocalWidgetLibrary].
+  ///
+  /// The returned map is an immutable view of the map provided to the constructor.
+  /// They keys are the unqualified widget names, and the values are the corresponding
+  /// [LocalWidgetBuilder]s.
+  ///
+  /// The map never changes during the lifetime of the [LocalWidgetLibrary], but a new
+  /// instance of an [UnmodifiableMapView] is returned each time this getter is used.
+  ///
+  /// See also:
+  ///
+  ///  * [createCoreWidgets], a function that creates a [Map] of local widgets.
+  UnmodifiableMapView<String, LocalWidgetBuilder> get widgets {
+    return UnmodifiableMapView<String, LocalWidgetBuilder>(_widgets);
+  }
 }
 
 class _ResolvedConstructor {
@@ -224,6 +243,24 @@ class Runtime extends ChangeNotifier {
   void clearLibraries() {
     _libraries.clear();
     _clearCache();
+  }
+
+  /// The widget libraries imported in this [Runtime].
+  ///
+  /// The returned map is an immutable view of the map updated by calls to
+  /// [update] and [clearLibraries].
+  /// 
+  /// The keys are instances [LibraryName] which encode fully qualified library
+  /// names, and the values are the corresponding [WidgetLibrary]s.
+  /// 
+  /// The returned map is an immutable copy of the registered libraries
+  /// at the time of this call. 
+  ///
+  /// See also:
+  ///
+  ///  * [update] and [clearLibraries], functions that populate this map.
+  UnmodifiableMapView<LibraryName, WidgetLibrary> get libraries {
+    return UnmodifiableMapView<LibraryName, WidgetLibrary>(Map<LibraryName, WidgetLibrary>.from(_libraries));
   }
 
   final Map<FullyQualifiedWidgetName, _ResolvedConstructor?> _cachedConstructors = <FullyQualifiedWidgetName, _ResolvedConstructor?>{};
--- a/packages/rfw/test/runtime_test.dart
+++ b/packages/rfw/test/runtime_test.dart
@@ -349,6 +349,18 @@ void main() {
 
   testWidgets('Runtime', (WidgetTester tester) async {
     final Runtime runtime = Runtime()
+          ..update(const LibraryName(<String>['core']), createCoreWidgets());
+    expect(runtime.libraries.length, 1);
+    final LibraryName libraryName = runtime.libraries.entries.first.key;
+    expect('$libraryName', 'core');
+    final WidgetLibrary widgetLibrary = runtime.libraries.entries.first.value;
+    expect(widgetLibrary, isA<LocalWidgetLibrary>());
+    widgetLibrary as LocalWidgetLibrary;
+    expect(widgetLibrary.widgets.length, greaterThan(1));
+  });
+
+  testWidgets('Runtime', (WidgetTester tester) async {
+    final Runtime runtime = Runtime()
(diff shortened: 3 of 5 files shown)


packages/rfw/test/runtime_test.dart
```diff
@@ -347,6 +347,17 @@ void main() {
     expect(tester.widget<ColoredBox>(find.byType(ColoredBox)).color, const Color(0x00000002));
   });
 
+  testWidgets('Runtime', (WidgetTester tester) async {
+    final Runtime runtime = Runtime()
+          ..update(const LibraryName(<String>['core']), createCoreWidgets());
+    for (final MapEntry(key: LibraryName libName, value: WidgetLibrary lib) in runtime.libraries.entries) {
+      expect(libName.toString(), equals('core'));
+      if (lib case final LocalWidgetLibrary localWidgetLibrary) {
```
Hixie, line 355: I have no idea what this line of code does.

packages/rfw/lib/src/flutter/runtime.dart
```diff
@@ -169,6 +171,11 @@ class LocalWidgetLibrary extends WidgetLibrary {
   LocalWidgetBuilder? findConstructor(String name) {
     return _widgets[name];
   }
+
+  /// Returns an unmodifiable map from widget name to LocalWidgetBuilder.
```
Hixie, line 175: LocalWidgetBuilder should be a hyperlink.

This should probably include more detail, also. If someone is reading these docs, they have a question. What's the question that might lead them here?

For example, it could be something like

```
/// The widgets defined by this [LocalWidgetLibrary].
///
/// The returned map is an immutable view of the map provided to the constructor.
/// They keys are the unqualified widget names, and the values are the corresponding
/// [LocalWidgetBuilder]s.
///
/// The map never changes during the lifetime of the [LocalWidgetLibrary], but a new
/// instance of an [UnmodifiableMapView] is returned each time this getter is used.
///
/// See also:
///
///  * [createCoreWidgets], a function that creates a [Map] of local widgets.
```

packages/rfw/lib/src/flutter/runtime.dart
```diff
@@ -226,6 +233,11 @@ class Runtime extends ChangeNotifier {
     _clearCache();
   }
 
+  /// Returns an unmodifiable map from LibraryName to WidgetLibrary.
```
Hixie, line 236: same as above, docs should be more detailed

packages/rfw/test/runtime_test.dart
```diff
@@ -347,6 +347,17 @@ void main() {
     expect(tester.widget<ColoredBox>(find.byType(ColoredBox)).color, const Color(0x00000002));
   });
 
+  testWidgets('Runtime', (WidgetTester tester) async {
+    final Runtime runtime = Runtime()
+          ..update(const LibraryName(<String>['core']), createCoreWidgets());
+    for (final MapEntry(key: LibraryName libName, value: WidgetLibrary lib) in runtime.libraries.entries) {
```
Hixie, line 353: avoid abbreviations (https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-abbreviations)

packages/rfw/test/runtime_test.dart
```diff
@@ -347,6 +347,17 @@ void main() {
...
+      expect(libName.toString(), equals('core'));
```
Hixie, line 354: `'$libName'`

Hixie: nit: (also you don't need to explicitly specify `equals`)

packages/rfw/test/runtime_test.dart
```diff
@@ -347,6 +347,17 @@ void main() {
...
+      if (lib case final LocalWidgetLibrary localWidgetLibrary) {
+        expect(localWidgetLibrary.widgets.length, greaterThan(1));
```
Hixie, line 356: why is this expectation conditional?

Hixie (changes requested):

packages/rfw/test/runtime_test.dart
johnmccutchan: this is a "pattern" it checks if "lib" is an instance of LocalWidgetLibrary and then creates a local variable named localWidgetLibrary that is available in the if block.

packages/rfw/lib/src/flutter/runtime.dart
johnmccutchan: Done.

packages/rfw/test/runtime_test.dart
johnmccutchan: Done.

packages/rfw/test/runtime_test.dart
johnmccutchan: done.

packages/rfw/test/runtime_test.dart
johnmccutchan: In this test it would have always executed but I've rewritten this section to not use any conditional logic.

packages/rfw/lib/src/flutter/runtime.dart
johnmccutchan: Done.

packages/rfw/test/runtime_test.dart
```diff
@@ -347,6 +347,17 @@ void main() {
     expect(tester.widget<ColoredBox>(find.byType(ColoredBox)).color, const Color(0x00000002));
   });
 
+  testWidgets('Runtime', (WidgetTester tester) async {
+    final Runtime runtime = Runtime()
+          ..update(const LibraryName(<String>['core']), createCoreWidgets());
+    for (final MapEntry<LibraryName, WidgetLibrary>(key: LibraryName libraryName, value: WidgetLibrary widgetLibrary) in runtime.libraries.entries) {
```
tugorez, line 353: Given that we know it has a single entry should we use something like map.entries.single and then make the assertions?

packages/rfw/test/runtime_test.dart
Hixie: it seems like `if (lib is LocalWidgetLibrary)` would be easier to understand?

packages/rfw/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.0.18
+
+* Exposes `WidgetLibrary`s registered in `Runtime.
```
Hixie, line 3: missing \`

johnmccutchan: Done.

packages/rfw/lib/src/flutter/runtime.dart
```diff
@@ -226,6 +244,25 @@ class Runtime extends ChangeNotifier {
...
 
+  /// The widget libraries imported in this [Runtime].
+  ///
+  /// The returned map is an immutable view of the map updated by calls to
+  /// [update] and [clearLibraries].
+  /// 
+  /// The keys are instances [LibraryName] which encode fully qualified library
+  /// names, and the values are the corresponding [WidgetLibrary]s.
+  /// 
+  /// The map may change during the lifetime of the [Runtime], and a new
```
Hixie, line 255: That's rather terrifying as an API contract. You get a map that can arbitrarily change out from under you without notification? Should we return a copy instead?

packages/rfw/test/runtime_test.dart
```diff
@@ -347,6 +347,17 @@ void main() {
...
+      expect('$libraryName', 'core');
+      expect(widgetLibrary, isA<LocalWidgetLibrary>());
+      final LocalWidgetLibrary localWidgetLibrary = widgetLibrary as LocalWidgetLibrary;
```
Hixie, line 356: (fwiw I believe you don't need the variable here, you can just say `widgetLibrary as LocalWidgetLibrary;` and then use `widgetLibrary` as the new type, due to promotion. Either is fine though.)

Hixie (approved):
LGTM modulo minor comments

packages/rfw/test/runtime_test.dart
johnmccutchan: Interesting, Dart has changed a lot in the last 7 years.

packages/rfw/test/runtime_test.dart
johnmccutchan: Done.

packages/rfw/test/runtime_test.dart
johnmccutchan: Done.

packages/rfw/lib/src/flutter/runtime.dart
johnmccutchan: Done.

johnmccutchan:
Thanks all for the reviews!

PR #5958, opened 2024-01-23
[camera_platform_interface] Add appleCaptureDeviceType to CameraDescription

rickcasson:
To more easily differentiate between the available cameras on iOS devices, the optional appleCaptureDeviceType parameter can become populated when discovering devices.

*List which issues are fixed by this PR. You must list at least one issue.*
Closes  flutter/flutter#142025

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #5957, opened 2024-01-22
[camera_avfoundation] Add all missing iOS capture device types

rickcasson:
Several of the physical and logical cameras were missing when discovering available capture devices on iOS,  so I completed the list based on Apple's developer documentation: https://developer.apple.com/documentation/avfoundation/avcapturedevice/devicetype

@bibaswan-bhawal had previously made a PR (#5892) that added one of these types, but I think it would be helpful to have the complete list available.

*List which issues are fixed by this PR. You must list at least one issue.*
closes flutter/flutter#134151
closes flutter/flutter#130073

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #5961, opened 2024-01-23
[camera] Export apple capture device type

rickcasson:
This exports the enum `AppleCaptureDeviceType` so it can used by the user to identify and select the camera they would like to use. This PR is dependant on #5957, #5958, and #5960 

Closes https://github.com/flutter/flutter/issues/142025

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/lib/camera.dart
+++ b/packages/camera/camera/lib/camera.dart
@@ -4,7 +4,6 @@
 
 export 'package:camera_platform_interface/camera_platform_interface.dart'
     show
-        AppleCaptureDeviceType,
         CameraDescription,
         CameraException,
         CameraLensDirection,
--- a/packages/camera/camera/lib/camera.dart
+++ b/packages/camera/camera/lib/camera.dart
@@ -4,6 +4,7 @@
 
 export 'package:camera_platform_interface/camera_platform_interface.dart'
     show
+        AppleCaptureDeviceType,
         CameraDescription,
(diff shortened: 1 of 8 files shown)


stuartmorgan:
Thanks for the contribution!

Please see https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins. It looks like you skipped to steps 3 and 5; this needs a holistic review before individual PRs should be opened.

packages/camera/camera/lib/camera.dart
```diff
@@ -4,6 +4,7 @@
 
 export 'package:camera_platform_interface/camera_platform_interface.dart'
     show
+        AppleCaptureDeviceType,
```
stuartmorgan, line 7: This is unlikely to be something we want to add at this layer. The goal of the app-facing package is to provide abstractions, not directly mirror every low-level detail of every platform, as the latter will result in an extremely confusing API.

We have patterns around having high-level abstractions, and then providing platform-specific subclasses with platform details if necessary, which would seem like a better option here. The first thing to evaluate is whether we actually need platform-specific details at all here; having different types of cameras isn't unique to iOS, so a design proposal that surveys the relevant APIs/distinctions across platforms would be the best place to start.

stuartmorgan (changes requested):

PR #6011, opened 2024-01-29
[ci] Reduce tasks slightly

stuartmorgan:
Help buy a bit of time until we eliminate the max number of tasks for `release` to wait for, and be more efficient with CI resources:
- Folds the very fast repo-tooling-unit-test step into the existing repo checks task, cutting down on machine overhead.
- Removes the `master` version of downgraded analysis, since it's hard to imagine a plausible scenario where it fails but the `stable` version does not, so we don't need to spend resources on both.

See https://github.com/flutter/flutter/issues/141196

### DIFF
--- a/.ci/targets/repo_checks.yaml
+++ b/.ci/targets/repo_checks.yaml
@@ -2,6 +2,8 @@ tasks:
   - name: prepare tool
     script: .ci/scripts/prepare_tool.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
+  - name: tool unit tests
+    script: .ci/scripts/plugin_tools_tests.sh
   - name: format
     script: script/tool_runner.sh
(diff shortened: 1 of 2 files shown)


tarrinneal (approved):

.ci/targets/repo_checks.yaml
```diff
@@ -2,6 +2,8 @@ tasks:
   - name: prepare tool
     script: .ci/scripts/prepare_tool.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
+  - name: tool unit tests
+    script: .ci/scripts/plugin_tools_tests.sh
```
ditman, line 6: As per the note above, should this one be an `infra_step: true` as well?

ditman (approved):
ACK!

PR #6022, opened 2024-01-31
[camerax] Clarify how CameraX uses preset resolution

camsim99:
Improves documentation to clarify that CameraX uses the preset resolution to target for [UseCase]s and are not guaranteed.

Fixes https://github.com/flutter/flutter/issues/141766.

## Pre-launch Checklist

- [x] I signed the [CLA].


camsim99:
Not sure about the version changes here and I assume if I do keep the version changes, then I will need to split this to land the changes in the interface and in the plugin separately.

packages/camera/camera_platform_interface/lib/src/types/resolution_preset.dart
```diff
@@ -4,7 +4,14 @@
 
 /// Affect the quality of video recording and image capture:
 ///
-/// If a preset is not available on the camera being used a preset of lower quality will be selected automatically.
+/// If a preset is not available on the camera being used, a preset of lower
+/// quality will be selected automatically.
+///
+/// For the `camera_android_camerax` platform implementation of the plugin,
```
stuartmorgan, line 10: We generally try not to document specific implementations in the platform interface, and in fact try to remove existing comments like that (which usually predate federation) incrementally. I would prefer we combine this with the part just above in a more generic way. E.g.,

> This is treated as a target resolution, and exact values are not guaranteed. Platform implementations may fall back to a higher or lower resolution if the specific preset is not available.

We could also remove the very specific details in the entries below, and just make them say `~240p`, `~480p`, `~720p`, `~1080p`, `~2160p`, so it's clearer that we aren't promising exact values (and to make it more future-proof; it's already not saying what happens on Windows for instance).

stuartmorgan (changes requested):

packages/camera/camera_platform_interface/lib/src/types/resolution_preset.dart
camsim99: Modified this! I liked your phrasing for making it future-proof.

PR #6043, opened 2024-02-02
[pigeon] Adds Dart implementation of ProxyApi

bparrishMines:
Part of https://github.com/flutter/flutter/issues/134777

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/lib/dart_generator.dart
```diff
@@ -94,7 +102,10 @@ class DartGenerator extends StructuredGenerator<DartOptions> {
     indent.writeln('// ${getGeneratedCodeWarning()}');
     indent.writeln('// $seeAlsoWarning');
     indent.writeln(
-      '// ignore_for_file: public_member_api_docs, non_constant_identifier_names, avoid_as, unused_import, unnecessary_parenthesis, prefer_null_aware_operators, omit_local_variable_types, unused_shown_name, unnecessary_import, no_leading_underscores_for_local_identifiers',
+      '// ignore_for_file: public_member_api_docs, non_constant_identifier_names, '
+      'avoid_as, unused_import, unnecessary_parenthesis, prefer_null_aware_operators, '
+      'omit_local_variable_types, unused_shown_name, unnecessary_import, '
+      'no_leading_underscores_for_local_identifiers, camel_case_types',
```
bparrishMines, line 108: The only difference is the `camel_case_types` added to the end.

packages/pigeon/example/app/lib/src/messages.g.dart
```diff
@@ -3,7 +3,7 @@
 // found in the LICENSE file.
 // Autogenerated from Pigeon, do not edit directly.
 // See also: https://pub.dev/packages/pigeon
-// ignore_for_file: public_member_api_docs, non_constant_identifier_names, avoid_as, unused_import, unnecessary_parenthesis, prefer_null_aware_operators, omit_local_variable_types, unused_shown_name, unnecessary_import, no_leading_underscores_for_local_identifiers
+// ignore_for_file: public_member_api_docs, non_constant_identifier_names, avoid_as, unused_import, unnecessary_parenthesis, prefer_null_aware_operators, omit_local_variable_types, unused_shown_name, unnecessary_import, no_leading_underscores_for_local_identifiers, camel_case_types
```
stuartmorgan, line 6: Why do we need the new ignore? We generally want to follow language conventions like capitalization in generators.

packages/pigeon/lib/dart_generator.dart
```diff
@@ -434,6 +452,194 @@ final BinaryMessenger? ${_varNamePrefix}binaryMessenger;
...
+  @override
+  void writeInstanceManager(
+    DartOptions generatorOptions,
+    Root root,
+    Indent indent, {
+    required String dartPackageName,
+  }) {
+    indent.writeln(proxyApiBaseClass);
+
+    indent.writeln(
```
stuartmorgan, line 464: Should these calls to templates be `writeln`, or just `write`? It seems odd to use `writeln` for arbitrary multi-line chunks of text.

packages/pigeon/lib/dart_generator.dart
```diff
@@ -434,6 +452,194 @@ final BinaryMessenger? ${_varNamePrefix}binaryMessenger;
...
+  void writeProxyApi(
+    DartOptions generatorOptions,
+    Root root,
+    Indent indent,
+    AstProxyApi api, {
+    required String dartPackageName,
+  }) {
+    const String codecName = '_${classNamePrefix}ProxyApiBaseCodec';
+
+    // Each api has a private codec instance used by every host method,
```
stuartmorgan, line 507: Nit: API

packages/pigeon/lib/dart_generator.dart
```diff
@@ -434,6 +452,194 @@ final BinaryMessenger? ${_varNamePrefix}binaryMessenger;
...
+      final Set<AstProxyApi> interfaceApisFromSuperClasses =
+          recursiveFindAllInterfaceApis(
+        api.superClass!.associatedProxyApi!,
+      );
+      for (final AstProxyApi proxyApi in interfaceApisFromSuperClasses) {
+        flutterMethodsFromSuperClasses.addAll(proxyApi.methods);
+      }
+    }
+
+    // Ast class used by code_builder to generate the code.
```
stuartmorgan, line 548: Nit: AST

packages/pigeon/lib/dart_generator.dart
```diff
@@ -434,6 +452,194 @@ final BinaryMessenger? ${_varNamePrefix}binaryMessenger;
...
+          codecName: codecName,
+        ))
+        ..methods.add(_proxyApiCopyMethod(
+          apiName: api.name,
+          unattachedFields: api.unattachedFields,
+          declaredAndInheritedFlutterMethods: flutterMethodsFromSuperClasses
+              .followedBy(flutterMethodsFromInterfaces)
+              .followedBy(api.flutterMethods),
+        )),
+    );
```
stuartmorgan, line 637: This seems pretty nice. Definitely something to consider for the other Dart generation.

packages/pigeon/lib/dart_generator.dart
```diff
@@ -787,10 +1001,818 @@ if (${_varNamePrefix}replyList == null) {
...
+            );
+        },
+      );
+    }
+  }
+
+  /// The detached constructor present for every ProxyApi.
+  ///
+  /// This constructor doesn't include a host method call to create a new native
+  /// class instance. It is mainly used when the native side once to create a
```
stuartmorgan, line 1146: s/once/wants/?

packages/pigeon/pigeons/proxy_api_tests.dart
```diff
@@ -0,0 +1,469 @@
...
+
+import 'package:pigeon/pigeon.dart';
+
+enum ProxyApiTestEnum {
+  one,
+  two,
+  three,
+}
+
+/// The core interface that each host language plugin must implement in
```
stuartmorgan, line 13: This needs to be adjusted for this class so it's not a direct copy/paste of the other comment.

packages/pigeon/pigeons/proxy_api_tests.dart
```diff
@@ -0,0 +1,469 @@
...
+    List<Object?>? nullableListParam,
+    // ignore: avoid_unused_constructor_parameters
+    Map<String?, Object?>? nullableMapParam,
+    // ignore: avoid_unused_constructor_parameters
+    ProxyApiTestEnum? nullableEnumParam,
+    // ignore: avoid_unused_constructor_parameters
+    ProxyApiSuperClass? nullableProxyApiParam,
+  );
+
+  late bool aBool;
```
stuartmorgan, line 57: Why does everything need to be `late` in the interface definition?

stuartmorgan:
(Mostly small comments; There are some big walls of text so I was skimming in the generator, but it looks like the overall structure here is pretty well factored.)

packages/pigeon/lib/dart/templates.dart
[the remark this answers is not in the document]
bparrishMines: I wouldn't expect these to have many changes in the future. And if they do, it shouldn't be changes that require adding logic when generating them. This should at least be true for `InstanceManager`, `InstanceManagerApi`, and `ProxyApiBaseClass`. I expect all of these to always be basic templates.

However, `ProxyApiBaseCodec` could require logic later to solve some complex issues, but I don't have any plans for it yet.

packages/pigeon/lib/dart_generator.dart
```diff
@@ -2,9 +2,13 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:code_builder/code_builder.dart' as cb;
+import 'package:collection/collection.dart' as collection;
```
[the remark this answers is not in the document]
bparrishMines: tl/dr; I removed this import and changed my usages of `mapIndexed` to `indexMap`.

But yes, I was using `mapIndexed` from `collection` instead of `indexMap` from the `functional.dart`. And these imports collide because they have a method called `mergeMaps`. I was originally using a different method from `collection`, but it doesn't look like I'm using it anymore. So i switched to just use the method from `functional.dart`. I do think we should eventually remove the methods from `functional.dart` that are already provided by the `collection` package.

packages/pigeon/lib/dart_generator.dart
bparrishMines: this change was undone

packages/pigeon/lib/dart_generator.dart
bparrishMines: I changed it to `indent.format`

packages/pigeon/platform_tests/shared_test_plugin_code/lib/src/generated/proxy_api_tests.gen.dart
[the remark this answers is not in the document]
bparrishMines: Sorry, this file is all generated so that I can write the integration tests and verify the generated code doesn't cause any lint warnings. I assumed the review for this file would only require someone to skim it.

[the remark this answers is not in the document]
tarrinneal: In a sense this is the code that needs the most review, since it's the actual functioning code from what you've written. The rest of it is just preference and future work looking.

packages/pigeon/lib/ast.dart
```diff
@@ -177,6 +178,46 @@ class AstProxyApi extends Api {
...
+    if (superClass != null) {
+      final Set<AstProxyApi> interfaceApisFromSuperClasses =
+          recursiveFindAllInterfaceApis(
+        superClass!.associatedProxyApi!,
+      );
+      for (final AstProxyApi proxyApi in interfaceApisFromSuperClasses) {
+        yield* proxyApi.methods;
+      }
+    }
+  }
```
tarrinneal, line 219: I don't know if performance really matters here, but ideally these methods would only ever be ran once, and the data stored on the class.

packages/pigeon/lib/ast.dart
```diff
@@ -177,6 +178,46 @@ class AstProxyApi extends Api {
...
 
+  /// A list of AstProxyApis where each `extends` the API that follows it.
+  Iterable<AstProxyApi> get allSuperClasses => recursiveGetSuperClassApisChain(
+        this,
+      );
+
+  /// All ProxyApis this API `implements` and all the interfaces those APIs
+  /// `implements`.
+  Iterable<AstProxyApi> get apisOfInterfaces =>
+      recursiveFindAllInterfaceApis(this);
```
tarrinneal, line 189: These getters call methods that are defined in the tools file but - I think - are only called here. Should they be moved into this class instead? Would prevent the `this` argument. Probably doesn't matter.

packages/pigeon/lib/ast.dart
bparrishMines: Since some of the fields of `AstProxyApi` that this method uses are mutable, I don't think we can run it only once. And they are mutable because of the step in `pigeon_lib.dart` where the `associatedProxyApi` is added to all the type declarations.

packages/pigeon/lib/ast.dart
bparrishMines: Good catch! I didn't realize that this was the only place that I was using them. I moved the full method implementation into this class.

packages/pigeon/lib/dart_generator.dart
```diff
@@ -521,15 +727,22 @@ final BinaryMessenger? ${_varNamePrefix}binaryMessenger;
...
+        root.apis.any((Api api) => api is AstProxyApi);
+
+    if (hasHostMethod) {
       _writeCreateConnectionError(indent);
     }
-    if (hasFlutterApi || generatorOptions.testOutPath != null) {
+    if (hasFlutterMethod || generatorOptions.testOutPath != null) {
       _writeWrapResponse(generatorOptions, root, indent);
     }
   }
```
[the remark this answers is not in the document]
bparrishMines: I created https://github.com/flutter/flutter/issues/144642 to tracking moving the methods to `Generator`.

packages/pigeon/lib/ast.dart
tarrinneal: Doesn't the code that associates the api's make copies? Maybe I'm looking in the wrong place.

bparrishMines: Unless I'm missing something, a new api is not created: https://github.com/flutter/packages/blob/main/packages/pigeon/lib/pigeon_lib.dart#L1297 and https://github.com/flutter/packages/blob/main/packages/pigeon/lib/pigeon_lib.dart#L1337.

Only `NamedType`s (e.g. fields, methods parameters, method returnType, constructor parameters, etc...) are copied. The original apis, methods, and constructors are retained.

But even so, as long as the fields are mutable, it could cause a bug in the future if a contributor uses the method. I do think we should probably restructure the AST with immutable classes in the future. Mutable data classes can end up being a foot-gun.

tarrinneal (approved):
I think we're good to land this now, if @stuartmorgan wants to get another pass in now's the time.

Please add an issue as mentioned above, and I think it's good to land.

stuartmorgan:
I've reviewed manual versions of the output in `webview_flutter` in the past, so if you're happy with the generator aspects of this I don't feel like I need to review again.

packages/pigeon/lib/ast.dart
tarrinneal: Can you add a todo for this and tag me in it?

PR #6044, opened 2024-02-02
Skip the wasm test for now, it doesn't even actually test wasm currently

eyebrowsoffire:
The semantics for the wasm stuff is changing, and this is blocking the roll. See https://github.com/flutter/flutter/issues/142809

Also, this test wasn't actually testing the wasm build previously, since it is still serving the `build/web` directory even though `--wasm` used to actually output to `build/web_wasm`.


eyebrowsoffire:
No versioning needed, we're just skipping a test and not changing any non-test code in the package itself.

yjbanov (approved):

PR #6051, opened 2024-02-05
[video_player] Call `super.dispose` last in README example

tp:
As per https://api.flutter.dev/flutter/widgets/State/dispose.html

> Implementations of this method should end with a call to the inherited method, as in super.dispose().

### DIFF
--- a/packages/video_player/video_player/CHANGELOG.md
+++ b/packages/video_player/video_player/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.8.5
+
+* Updates example to call `super.dispose()` last.
+
 ## 2.8.4
 
 * Removes `_ambiguate` methods from code.
--- a/packages/video_player/video_player/README.md
+++ b/packages/video_player/video_player/README.md
@@ -118,8 +118,8 @@ class _VideoAppState extends State<VideoApp> {
 
   @override
   void dispose() {
-    super.dispose();
     _controller.dispose();
+    super.dispose();
   }
 }
 ```
--- a/packages/video_player/video_player/example/lib/basic.dart
+++ b/packages/video_player/video_player/example/lib/basic.dart
@@ -66,8 +66,8 @@ class _VideoAppState extends State<VideoApp> {
 
   @override
   void dispose() {
-    super.dispose();
     _controller.dispose();
+    super.dispose();
   }
 }
 // #enddocregion basic-example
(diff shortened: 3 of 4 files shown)


tarrinneal:
Good catch on this! Our README code samples are pulled directly from code snippets in the example app and must match what is in the app to pass the repo_checks tests.

https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#readme-code

you can just manually change the snippet to match, or change the snippet and run the code I linked to.

You'll also need to update the changlelog and versioning for this package or the changes you added wont be synced to the published package.

https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates

tarrinneal (changes requested):
Good catch on this! Our README code samples are pulled directly from code snippets in the example app and must match what is in the app to pass the repo_checks tests.

https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#readme-code

you can just manually change the snippet to match, or change the snippet and run the code I linked to.

You'll also need to update the changlelog and versioning for this package or the changes you added wont be synced to the published package.

https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates

stuartmorgan:
From triage: @tp, are you still planning on updating this PR to address the review comments above?

tarrinneal:
> @tarrinneal Thanks for the detailed explanation. I have updated the PR accordingly, and now only one job that fails across multiple open PRs (and seemingly unrelated) fails.

Yep, tree-status isn't one you need to worry about. I'll add this back into my queue! Thanks for fixing it up!

tarrinneal (approved):
thanks for putting the time into this!

tarrinneal:
@stuartmorgan for second review

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.8.4
+
+* Update example to call `super.dispose()` last
```
stuartmorgan, line 3: Updates

https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style

stuartmorgan, line 3: This is missing a period.

https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style

stuartmorgan (changes requested):

stuartmorgan (approved):

stuartmorgan:
Your rebase removed the version bump; PRs can't add changes to versions that have already been published.

PR #6059, opened 2024-02-05
[camerax] Implements `setFocusPoint`, `setExposurePoint`, `setExposureOffset`

camsim99:
This PR implements `setFocusPoint`, `setExposurePoint`, `setExposureOffset` and makes some small fixes here and there, each of which I have left a comment about for context.

Part of https://github.com/flutter/flutter/issues/120468 & https://github.com/flutter/flutter/issues/120467.

~NOTE: Should land after https://github.com/flutter/packages/pull/6068.~ done :)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -446,13 +465,67 @@ class AndroidCameraCameraX extends CameraPlatform {
 
   /// Gets the supported step size for exposure offset for the selected camera in EV units.
   ///
-  /// Returns 0 when exposure compensation is not supported.
-  ///
```
camsim99, line 450: This explicitly went against the platform interface; 0 should be returned when step sizes are not supported. So, this is an unrelated fix for that fact.

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/CameraControlHostApiImpl.java
```diff
@@ -83,6 +83,12 @@ public void onSuccess(Void voidResult) {
             }
 
             public void onFailure(Throwable t) {
+              if (t instanceof CameraControl.OperationCanceledException) {
+                // Operation was canceled due to camera being closed or a new request was submitted, which
+                // is not actionable and should not block a new value from potentially being submitted.
+                result.success(null);
+                return;
+              }
```
camsim99, line 91: A semi-related fix: some `CameraControl` operations will throw an exception if interrupted or could not have completed due to the camera being closed. Changed this to result in a success since this is not an exception could/should be handled by someone using the plugin.

For an example of why, the example camera app has a slider to change the exposure offset, and this results in many of these exceptions being thrown because it makes many rapid calls to `CameraControl.setExposureCompensationIndex`.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -452,13 +481,66 @@ class AndroidCameraCameraX extends CameraPlatform {
...
     final ExposureState exposureState = await cameraInfo!.getExposureState();
-    return exposureState.exposureCompensationStep;
+    final double exposureOffsetStepSize =
+        exposureState.exposureCompensationStep;
+    if (exposureOffsetStepSize == 0) {
+      // CameraX returns a step size of 0 if exposure compensation is not
+      // supported for the device.
+      throw CameraException(exposureCompensationNotSupported,
+          'Exposure compensation not supported for the device.');
+    }
```
camsim99, line 495: Looking for feedback on throwing an exception here/how to handle exposure compensation not being supported on the device.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -452,13 +481,66 @@ class AndroidCameraCameraX extends CameraPlatform {
...
+
+    try {
+      await cameraControl!
+          .setExposureCompensationIndex(roundedExposureCompensationIndex);
+    } on PlatformException catch (e) {
+      throw CameraException(
+          'setExposureOffsetFailed',
+          e.message ??
+              'Setting the camera exposure compensation index failed.');
+    }
```
camsim99, line 530: Looking for feedback on throwing an exception here and forwarding the exception from `CameraControl` in order to conform with the platform interface, in the sense that the method

> Throws a `CameraException` when an illegal offset is supplied.

as per its documentation. Another option is explicitly checking in this method, but was worried that such an asynchronous call could impact performance.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -832,14 +910,15 @@ class AndroidCameraCameraX extends CameraPlatform {
...
-  /// Updates [cameraInfo] to the information corresponding to [camera] and
-  /// adds observers to the [LiveData] of the [CameraState] of the current
-  /// [camera], saved as [liveCameraState].
+  /// Updates [cameraInfo] and [cameraControl] to the information corresponding
+  /// to [camera] and adds observers to the [LiveData] of the [CameraState] of
+  /// the current [camera], saved as [liveCameraState].
   ///
   /// If a previous [liveCameraState] was stored, existing observers are
   /// removed, as well.
   Future<void> _updateCameraInfoAndLiveCameraState(int cameraId) async {
```
camsim99, line 919: This is mistakenly not called after updating the `camera` instance` when video capture is started. This will be fixed in https://github.com/flutter/packages/pull/6068.

packages/camera/camera_android_camerax/lib/src/camera_control.dart
```diff
@@ -158,11 +171,20 @@ class _CameraControlHostApiImpl extends CameraControlHostApi {
       CameraControl instance, int index) async {
     final int identifier = instanceManager.getIdentifier(instance)!;
     try {
-      return setExposureCompensationIndex(identifier, index);
+      final int? exposureCompensationIndex =
+          await setExposureCompensationIndex(identifier, index);
+      if (exposureCompensationIndex == null) {
```
camsim99, line 176: Part of [previously mentioned fix](https://github.com/flutter/packages/pull/6059/files#r1481836864).

packages/camera/camera_android_camerax/lib/src/camera_control.dart
```diff
@@ -158,11 +171,20 @@ class _CameraControlHostApiImpl extends CameraControlHostApi {
...
+        return Future<int?>.value();
+      }
+      return exposureCompensationIndex;
     } on PlatformException catch (e) {
       SystemServices.cameraErrorStreamController.add(e.message ??
           'Setting the camera exposure compensation index failed.');
-      return Future<int?>.value();
+      // Surfacing error to plugin layer to maintain consistency of
+      // setExposureOffset implementation across platform implementations.
+      rethrow;
```
camsim99, line 187: Looking for feedback on this as per https://github.com/flutter/packages/pull/6059/files#r1481841622.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -69,6 +73,9 @@ class AndroidCameraCameraX extends CameraPlatform {
   @visibleForTesting
   CameraInfo? cameraInfo;
 
+  /// The [CameraControl] instance that corresponds to the [camera] instance.
+  CameraControl? cameraControl;
+
```
camsim99, line 78: Decided to save this whenever the related `camera` is updated versus making async calls to retrieve it every time that it's needed.

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/MeteringPointHostApiImpl.java
```diff
@@ -25,17 +32,34 @@ public class MeteringPointHostApiImpl implements MeteringPointHostApi {
...
+
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+        display = activity.getDisplay();
+      } else {
+        display =
+            ((WindowManager) activity.getSystemService(Context.WINDOW_SERVICE)).getDefaultDisplay();
+      }
+
+      DisplayOrientedMeteringPointFactory factory =
+          getDisplayOrientedMeteringPointFactory(display, 1f, 1f, cameraInfo);
```
camsim99, line 61: A slightly larger fix: I originally implemented this host API with a different `MeteringPointFactory`, which was not technically wrong but would have required that the plugin manually handle changes in camera and device orientation and can be tricky. This factory automatically does that based on the current camera's `CameraInfo` and the current display. So, I made changes in this PR to the `MeteringPoint` wrapper to accommodate that fact.

camsim99:
To test if the coordinates are correct for focus/exposure points, I have been manually testing and following documentation. Would love if anyone would be willing to run the example app and verify, though! This is really a note to self to see if anyone else can manually run this to double check before landing this.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -981,4 +1062,89 @@ class AndroidCameraCameraX extends CameraPlatform {
...
+                  // mode specified, but this logic is safe because this plugin
+                  // only uses points that explicitly have mode
+                  // FocusMeteringAction.flagAe or FocusMeteringAction.flagAf.
+                  meteringPointInfo.$2 != meteringMode)
+              .toList();
+
+      if (newMeteringPointInfos.isEmpty) {
+        // If no other metering points were specified, cancel any previously
+        // started focus and metering actions.
+        await cameraControl!.cancelFocusAndMetering();
```
gmackall, line 1114: Should we set `currentFocusMeteringAction` to null before returning? Or do we need to keep it around even though it is canceled?

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -981,4 +1062,89 @@ class AndroidCameraCameraX extends CameraPlatform {
...
+        return;
+      }
+      currentFocusMeteringAction =
+          proxy.createFocusMeteringAction(newMeteringPointInfos);
+    } else if (meteringPoint.x < 0 ||
+        meteringPoint.x > 1 ||
+        meteringPoint.y < 0 && meteringPoint.y > 1) {
```
gmackall, line 1121: This should be `... meteringPoint.y < 0 || meteringPoint.y > 1`, right? Not &&?

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -981,4 +1062,89 @@ class AndroidCameraCameraX extends CameraPlatform {
...
+      throw CameraException('meteringPointInvalid',
+          'The coordinates of a metering point for an auto-focus or auto-exposure action must be within (0,0) and (1,1).');
```
gmackall, line 1123: Nit: include the provided (actual) coordinates in the log message, something like 
> ...must be within (0,0) and (1,1), but point $meteringPoint was provided for metering mode $meteringMode

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -981,4 +1062,89 @@ class AndroidCameraCameraX extends CameraPlatform {
...
+      // meteringMode.
+      if (currentFocusMeteringAction == null) {
+        // Attempting to clear a metering point from a previous action, but no
+        // such action exists.
+        return;
+      }
+
+      // Remove metering point with specified meteringMode from current focus
+      // and metering action, as only one focus or exposure point may be set
+      // at once in this plugin.
```
gmackall, line 1100: Why do we have this restriction? Also, do you happen to know what the use case is for having multiple focus or exposure points set? I just want to better understand why we are taking a different approach than CameraX and what the impact is

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -981,4 +1062,89 @@ class AndroidCameraCameraX extends CameraPlatform {
...
+  ///     [currentFocusMeteringAction] that do not share a metering mode with
+  ///     [meteringPoint] and
+  ///   * [meteringPoint] with the specified [meteringMode].
+  /// If [meteringPoint] is null, this action includes only metering points and
+  /// their modes previously added to [currentFocusMeteringAction] that do not
+  /// share a metering mode with [meteringPoint]. If there are no such metering
+  /// points, then the previously enabled focus and metering actions will be
+  /// canceled.
+  Future<void> _startFocusAndMeteringFor(
+      {required Point<double>? meteringPoint,
```
gmackall, line 1087: nit: Maybe rename to `point`, given that this isn't a `MeteringPoint`, and that we later use (what is currently named) `meteringPoint` to make a `MeteringPoint`

gmackall (commented):
Added some comments!

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/CameraControlHostApiImpl.java
camsim99: I believe the camera closes if there is an error or if all `UseCase`s are bound. Regardless, I agree that it should not silently fail; that's why I made it return `null` and added a `CameraErrorEvent` to the error event stream on the Dart side. Open to discussion on that choice though :) Ideally, we could know when the camera closes versus the operation being canceled because a new one was set, because I don't think we should be surfacing the latter as an exception as it does without this `catch` statement.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -430,6 +447,18 @@ class AndroidCameraCameraX extends CameraPlatform {
     captureOrientationLocked = false;
   }
 
+  /// Sets the exposure point for automatically determining the exposure values.
+  ///
+  /// Supplying `null` for the [point] argument will result in resetting to the
+  /// original exposure point value.
+  ///
+  /// [cameraId] is not used.
```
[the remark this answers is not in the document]
camsim99: Because it's part of the platform interface 😅 I think, though, it is included in all/most methods in the plugin for the purpose of creating various camera event streams. I plan to clarify this when fixing https://github.com/flutter/flutter/issues/126270.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: Yeah something like that! Returning a negative number or throwing an exception are really the only two options I can think of honestly, since we cannot return null.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -452,13 +481,68 @@ class AndroidCameraCameraX extends CameraPlatform {
...
+    }
+    return exposureOffsetStepSize;
+  }
+
+  /// Sets the exposure offset for the selected camera.
+  ///
+  /// The supplied [offset] value should be in EV units. 1 EV unit represents a
+  /// doubling in brightness. It should be between the minimum and maximum offsets
+  /// obtained through `getMinExposureOffset` and `getMaxExposureOffset` respectively.
+  /// Throws a `CameraException` when an illegal offset is supplied.
```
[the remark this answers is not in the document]
camsim99: Both are technically true because setting the offset will fail when an illegal offset is supplied (I made sure this is true since these docs are from the platform interface and I wanted to match that behavior). I modified the doc to include that case :)

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -981,4 +1062,89 @@ class AndroidCameraCameraX extends CameraPlatform {
...
         videoQuality: videoQuality, fallbackStrategy: fallbackStrategy);
   }
+
+  // Methods for configuring auto-focus and auto-exposure:
+
+  /// Starts a foucs and metering action.
+  ///
+  /// This method will merge the current action's metering points with that
+  /// specified such that only one auto-exposure metering point and one
+  /// auto-focus metering point will be part of the action. This method would
```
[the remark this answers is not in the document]
camsim99: Modified, let me know if it reads better

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: This was an older design decision made for the plugin, but I'm not entirely why. I suppose you may want to balance the exposure based on two focal points or have multiple points of focus. CameraX does warn [in the docs](https://developer.android.com/reference/androidx/camera/core/CameraControl#startFocusAndMetering(androidx.camera.core.FocusMeteringAction)) that this may not be available for all devices, though. Perhaps it's a newer feature and that's why it wasn't supported in the past.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: Yes you're right, thank you!

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: Yes, thank you!!

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/CameraControlHostApiImpl.java
gmackall: Makes sense! I made the comment after seeing `result.success` here but didn't connect that it led to an error in `camera_control.dart`, thanks for clarifying

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
gmackall: I suppose I lean towards a special negative return value, because it isn't an actual error (or at least CameraX doesn't think it is). But this is a weak preference, if you have any reason for preferring throwing an error instead.

packages/camera/camera_android_camerax/test/android_camera_camerax_test.dart
```diff
@@ -389,6 +407,9 @@ void main() {
     // Verify the camera's CameraInfo instance got updated.
     expect(camera.cameraInfo, equals(mockCameraInfo));
 
+    // Verify camera's CameraControl instance gotu pdated.
```
gmackall, line 410:
```suggestion
    // Verify camera's CameraControl instance got updated.
```

gmackall (commented):
LGTM outside of the decision on how to handle the "Exposure compensation not supported for the device" situation

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
bparrishMines: Do you know how the other platforms handle this? I'm not familiar with how `exposureOffsetStepSize` works, but it seems the app-facing package indicates that a zero should be returned when it is not supported: https://pub.dev/documentation/camera/latest/camera/CameraController/getExposureOffsetStepSize.html

I think we should just return 0 if the other platforms are doing this as well.

camsim99: Ah okay. I interpreted that as still suggesting that the camera supports exposure compensation, though, which it doesn't. I am okay returning 0 though it is technically different in that way, since one way or another the developer will find this out when attempting to set exposure index. @gmackall thoughts?

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
bparrishMines: I'm fine with throwing an exception here since a user is explicitly warned about it in the app-facing package.

Although this does highlight the need for a way to check if a feature is supported without using a try/catch for every method. Something we will consider when the camera platform interface is updated.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
gmackall: I'm not sure if my terminology here is correct, but wouldn't returning a 0 here still indicate that exposure compensation IS supported on this device? Just that it is done without stepping? Or am I misunderstanding

bparrishMines: Ah, I see. If we want to return a `-1` to indicate that the feature is not supported, then we would need to add that to the app-facing documentation. Do you know how `camera_android` handles this? Does it throw an exception if exposure compensation is not supported or does it also return a default value?

gmackall: I don't think that documentation is indicating that the exposure step size can be negative, just that the exposure setting itself can be negative (and then explaining how that would interact with a given step size). I don't think it would make much sense for the step size itself to be negative, though I'm not 100% sure its impossible (the documentation isn't clear). @camsim99 is it possible to get confirmation that this value can't be negative, so we can safely use the negative value here to indicate it being unsupported?

camsim99: > I don't think that documentation is indicating that the exposure step size can be negative, just that the exposure setting itself can be negative (and then explaining how that would interact with a given step size).

This is correct. The formula is (Exposure compensation index) * (exposure offset step size) = (exposure offset), so that's just explaining that the offset can be negative if you set the compensation index to a negative value. The offset step size, at least for CameraX, will always be some positive value, unless un-supported.

camsim99: So, returning a negative value would be clear if we can modify the platform interface.

For `camera_android`, the [docs](https://developer.android.com/reference/android/hardware/camera2/CameraCharacteristics#CONTROL_AE_COMPENSATION_STEP) for the value it requests says that the value is available on all devices, so that's an interesting contrast. I assume CameraX just chooses not to return this value if it knows that trying to set an exposure offset will fail on the device. Would have to double check that though.

bparrishMines: Rereading the docs, I realized i misread the value that can be negative. Thanks for verifying that!

camsim99: Thanks for the research!

I changed it to return 0 since that is consistent with iOS and CameraX and added a note in the docs.

gmackall (approved):

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: After talking again with @gmackall, decided to change to -1 and modify the platform interface. I think it will be helpful to have a defined return value for unsupported exposure compensation going forward and help differentiate it from the other platforms, since iOS seems to not support step sizes by always returning 0 and it doesn't seem that exposure control is implemented on web or Windows.

@stuartmorgan let us know if you have any concerns about modifying the interface, though I believe this addition will help future-proof it. Windows could also, then, return -1.

camsim99: Going to land this for now since it's blocking me, but I can revisit this if needed and otherwise, will proceed with a PR for the platform interface.

stuartmorgan: > @stuartmorgan let us know if you have any concerns about modifying the interface, though I believe this addition will help future-proof it.

What do you mean by "modify the platform interface"? If it's just a docs change that sounds reasonable given the discussion above, if it's a code change that's complicated.

camsim99: Just a docs change :) https://github.com/flutter/packages/pull/6182

PR #6076, opened 2024-02-07
[image_picker] Suppress analyzer warning about deprecation of UnmodifiableUint8ListView

jason-simmons:
See https://github.com/flutter/flutter/issues/143110

### DIFF
--- a/packages/image_picker/image_picker_platform_interface/lib/src/types/picked_file/html.dart
+++ b/packages/image_picker/image_picker_platform_interface/lib/src/types/picked_file/html.dart
@@ -27,6 +27,8 @@ class PickedFile extends PickedFileBase {
 
   Future<Uint8List> get _bytes async {
     if (_initBytes != null) {
+      // TODO(stuartmorgan): Remove this directive, https://github.com/flutter/flutter/issues/143113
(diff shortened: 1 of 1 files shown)


packages/image_picker/image_picker_platform_interface/lib/src/types/picked_file/html.dart
```diff
@@ -27,6 +27,8 @@ class PickedFile extends PickedFileBase {
 
   Future<Uint8List> get _bytes async {
     if (_initBytes != null) {
+      // See https://github.com/flutter/flutter/issues/143110
```
stuartmorgan, line 30: This should be a TODO to do the migration, with a link to an issue that stays open and describes the migration needed and information about what release the replacement is available in. The goal isn't to leave `ignore`s indefinitely, but to clearly track the need to remove them and the timeline for doing so.

jason-simmons: Filed https://github.com/flutter/flutter/issues/143113 and updated the PR

PTAL

tarrinneal (approved):
Assuming this doesn't need to be published to pub, should be good to land as is.

PR #6077, opened 2024-02-07
[WIP][legacy_material] Supporting legacy Material Design

Piinks:
⚠️  This is a WIP. ⚠️ 

Would fix https://github.com/flutter/flutter/issues/143499, in the long run

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/legacy_material/lib/src/theme_data.dart
```diff
@@ -0,0 +1,1841 @@
...
+        ? const IconThemeData(color: Colors.white)
+        : const IconThemeData(color: Colors.black);
+
+    // COMPONENT THEMES
+    // TODO(Piinks): We may need to create custom component themes as we go
+    //  forward here to get the right styling. For example, we may need to
+    //  create LegacyAppBarTheme, that includes more properties than the default
+    //  theme provides. The default themes could be expanded upon as well, but
+    //  should be done separately so that development of this package is not
+    //  stalled by having to wait for stable releases.
```
QuncCccccc, line 331: Just a question here. If we have new properties in `LegacyXTheme` instead of `XTheme`, how would these properties be applied to the widgets? In the widget implementation, should we also get something like `legacyTheme = LegacyXTheme.of(context)` to get the widget style?

QuncCccccc (commented):
Thanks so much for this migration🤩 Overall looks good to me!

So far one problem that I can think of is, some widget "styles" are not controlled by widget theme data. For example, if we global search "material3 ? ", navigation bar and input decorator have some constants to customize the according widgets. Also for Switch, there is a config class which have 14 properties to provide some default constants based on M2/M3. Might need to consider if we should add all of them in `SwitchThemeData`🤣

packages/legacy_material/lib/src/theme_data.dart
Piinks: Thank you for catching this! This is an older comment I meant to remove, because you are very right - adding new properties to `LegacyXThemeData` wouldn't mean the widget `X` would adopt them.

> If we have new properties in LegacyXTheme instead of XTheme, how would these properties be applied to the widgets?

In order to do this right, we'll need to expand all component themes to include all of the properties we need, and then have the widgets adopt all of the properties from the theme if it is provided. 

 > In the widget implementation, should we also get something like legacyTheme = LegacyXTheme.of(context) to get the widget style?

No, because the component theme should come from Theme.of(context). Right now component theme look ups fallback on the primary theme:

https://github.com/flutter/flutter/blob/1e6a9024738787f862db8991ed325efead5b09d0/packages/flutter/lib/src/material/switch_theme.dart#L223-L226

We wouldn't be able to override a static `.of(context)` of a component theme in a way that would work for the widget, and we don't want the framework to depend on this package. What I envision right now is, in `LegacyThemeData` it will create `LegacyXThemeData` for widget `X`, which instead of having null property values, it will supply the default values from the M2Defaults we have scattered all over the framework. So when we call on `Theme.of(context).xTheme` we get the legacy theme data.

There would definitely be limitations; we are trying our best to support an outdated spec here. But I think it will be the best experience we can provide.

Piinks: Before moving forward with this, I am working on a doc that will lay it all out. :)

PR #6111, opened 2024-02-13
Allow deprecated members from the Dart SDK/Flutter Framework to roll in

matanlurey:
Namely, without breaking the tree. This is a deliberate policy decision change.

See https://github.com/flutter/flutter/issues/143312.


stuartmorgan:
The only answer I see to [my question about a replacement process covering this repo](https://github.com/flutter/flutter/issues/143312#issuecomment-1939344947) seems to have been a link to https://github.com/flutter/flutter/issues/143344 which is described in the comment linking to it as "a catch-all issue for "I wish we had X tooling or process" and in the issue description says "It's also not clear who or when this work will get done".

I would like some basic understand of the replacement process as it relates to this repository (even if it's not the idealized final process) before approving removing the process we have. Absent some specific replacement, I don't think this PR will be a net improvement for the reasons I described [here](https://github.com/flutter/flutter/issues/143312#issuecomment-1939316383). And for packages, which have to be published and then proactively picked up by users to avoid build breakage, delaying fixes from the deprecation point to the removal point isn't just a wash, it's a substantial net negative for all of our package clients in a way that is not true for the engine or framework repos, or for google3.

johnmccutchan:
> Does that process cover all affected repos? E.g., will flutter/packages be reviewed by that process, or does this involve new work that the ecosystem team needs to plan for?

This is a policy change where we are no longer letting a repo force upstream teams to account for deprecation errors, if you want to add a tax on yourself maintaining that repo that's you and your team's decision but it can't impact the framework engine or SDK teams.

> I would like some basic understand of the replacement process as it relates to this repository (even if it's not the idealized final process) before approving removing the process we have. 

We should have a tool that works on *any* Dart repository that can enumerate call sites of deprecated methods. As I wrote the previous sentence, it occurred to me that the analyzer cli must support identifying all calls to deprecated methods because we have to annotate our code to tell the analyzer to ignore this problem. So we have a production ready tool that can identify all of these issues on any given repository. 

Given this tool, I think there are two practical issues to tackle:

1) How do we stop new calls to deprecated APIs from being added? We add a pre-submit that counts the number of deprecated method usages *before* and *after* the PR is applied. If the count goes up, we block the PR. This is trivial to implement.

2) Where do we surface this information for everyone? Using deprecated methods is a code health signal so I would like to see our infrastructure surface this number along side things like code coverage, warnings, and other lints.  This dashboard should also show *all* callers of a specific deprecated API so that maintainers can quickly identify all call sites that need to be updated.

Would you feel more comfortable if I filed a more narrow bug around these changes?

stuartmorgan:
> This is a policy change where we are no longer letting a repo force upstream teams to account for deprecation errors, if you want to add a tax on yourself maintaining that repo that's you and your team's decision

So does that mean the answer to my question is that no, there is no replacement process covering this repository, nor is there any concrete plan to create one?

> but it can't impact the framework engine or SDK teams.

I haven't seen anyone address my point that absent a robust replacement process, all this does is kick the can down the road and then impact the same people. If we land this and then don't do anything else, the process will change from:
- roller breaks on deprecation of an SDK method, forcing someone to deal with it to unblock the roller.

to
- roller breaks on removal of an SDK method that has been deprecated for a while without anyone noticing because there was no process for surfacing and fixing it, forcing someone to deal with it to unblock the roller.

And as noted above, in *this* repository, that timing change would be quite bad **for all clients of these packages**. I'm not going to LGTM a change that I don't believe will actually make things appreciably better for any of the upstream parts of our team but will absolutely make things worse for our customers.

I'm happy to approve this once we actually have a concrete plan for a replacement process that will make this a net positive rather than a net negative (or alternately, if someone can explain to me what substantial benefit to the Flutter and Dart teams I am missing here that justifies making things worse for clients).

johnmccutchan:
> I haven't seen anyone address my point that absent a robust replacement process, all this does is kick the can down the road and then impact the same people. If we land this and then don't do anything else, the process will change from:
> 
> * roller breaks on deprecation of an SDK method, forcing someone to deal with it to unblock the roller.
> 
> to
> 
> * roller breaks on removal of an SDK method that has been deprecated for a while without anyone noticing because there was no process for surfacing and fixing it, forcing someone to deal with it to unblock the roller.
> 

I don't see how forcing a human to annotate all call sites with "// ignore: deprecated_method_use" is anything more than busy work that forces our static analysis tool to *hide* the problem from developers. 

The existing process matches my definition of papering over a problem and kicking the can down the road, and unfortunately, it also comes with a bunch of pointless busy work that interrupts people from doing impactful work.

A serious way of dealing with the problem is to not hide the analysis errors from maintainers, this will encourage maintainers to update their down-stream code faster and _actually_ deal with the problem instead of papering over it.

> And as noted above, in _this_ repository, that timing change would be quite bad **for all clients of these packages**. I'm not going to LGTM a change that I don't believe will actually make things appreciably better for any of the upstream parts of our team but will absolutely make things worse for our customers.
> 

If someone lands a change that removes deprecated methods *before* all callers of that method are updated, that is a breaking change and should immediately reverted. 

I'll just state this explicitly: The developers that could land such a change are your colleagues and are working on the same project. This means that everyone's incentives are aligned such that your scenario above won't ever happen because removing the API before the callers have been fixed isn't acceptable and will immediately be reverted. 

> I'm happy to approve this once we actually have a concrete plan for a replacement process that will make this a net positive rather than a net negative (or alternately, if someone can explain to me what substantial benefit to the Flutter and Dart teams I am missing here that justifies making things worse for clients).

I've described a concrete plan on how to avoid adding new uses of deprecated methods, how we should surface these code health signals to developers on the team, and how we will deal with accidental breakages caused by removing a method before all of its callers are fixed.

stuartmorgan:
> I don't see how forcing a human to annotate all call sites with "// ignore: deprecated_method_use" is anything more than busy work that forces our static analysis tool to _hide_ the problem from developers.

When they do that, they also have to (in this repo; I don't know what other repos enforce in code review) file an issue explaining what the required transition is, and leave TODOs on the `ignore`s that reference that issue, and then we track those issues, updating the code once it's possible to do so on `stable`.

So the current process requires someone to affirmatively engage with this repo and inject a work item into our stream that we have a process for tracking.

If you want to argue there are better ways to do that, great, but we need an actual better replacement, not just to remove a system that, while not perfect, *does work*, without any concrete replacement.

> A serious way of dealing with the problem is to not hide the analysis errors from maintainers, this will encourage maintainers to update their down-stream code faster and _actually_ deal with the problem instead of papering over it.

I don't consider adding TODOs and a tracking issue triaged in our normal process to be hiding things from maintainers.

> > And as noted above, in _this_ repository, that timing change would be quite bad **for all clients of these packages**. I'm not going to LGTM a change that I don't believe will actually make things appreciably better for any of the upstream parts of our team but will absolutely make things worse for our customers.
> 
> If someone lands a change that removes deprecated methods _before_ all callers of that method are updated, that is a breaking change and should immediately reverted.

I don't think you're understanding the scope of the problem with respect to flutter/packages. The call sites are in packages published to pub, which means that in practice updating all call sites before removing the method does not just require someone to land changes here, it requires that, **and also for every client to update their packages to get that change**. It's a completely different dynamic than the Dart/engine/framework relationships.

If we replace deprecated code shortly after it's **deprecated** (once it's possible to do so on stable), then the new version of the package without that code will, in practice, probably be published 6+ months before the API is released. Clients then have 6+ months wherein they will, if they update their dependencies, be future-proof. The vast majority of clients will never even notice.

If we replace the deprecated code when it's **removed**, then clients have at most 1-3 months (depending on when exactly it's removed in the `stable` cycle) to update. If they don't then their project will have compilation errors the next time they update Flutter. Compilation errors in code they don't use and may not even understand why they have. We occasionally get issues filed for cases like this with the ecosystem in general, and a lot of people find it very confusing and don't understand what they need to do.

> I'll just state this explicitly: The developers that could land such a change are your colleagues and are working on the same project. This means that everyone's incentives are aligned such that your scenario above won't ever happen because removing the API before the callers have been fixed isn't acceptable and will immediately be reverted.

The developers that could land such a change do not work on the projects of the millions of Flutter clients, and thus cannot guarantee that all callers are fixed. No revert will happen when a bunch of people's projects break because the fix is for all of them to update their dependencies. But even if it's technically up to them to fix their own projects, making a case we know is a pain point and making it substantially more common hurts developer experience.

What you are describing makes sense for usages in flutter/flutter, in flutter/engine, and in google3. It **does not work** in this repo because the call sites aren't here, they are massively distributed to machines we don't control. We can avoid breaking this repo's tree, sure, but not our clients. The goal of the CI in this repo is fundamentally to make sure the packages work as well as possible for our clients.

stuartmorgan:
Also, as more evidence that the flutter/packages use case does not appear to have been thought through in this policy change:

> 1. How do we stop new calls to deprecated APIs from being added? We add a pre-submit that counts the number of deprecated method usages _before_ and _after_ the PR is applied. If the count goes up, we block the PR. This is trivial to implement.

Such a feature would not be usable in this repository, because there is an up-to-three-month-window where adding uses of any given deprecated API may be *required*, in the very common case of an API being deprecated at the same time as the replacement is introduced (meaning the replacement isn't compatible with `stable` at the time of deprecation).

stuartmorgan (approved):
I've instituted a manual process in the ecosystem gardener rotation as a replacement.

I'll do a follow-up PR to clean up the now-pointless annotations that are already in the repo, and ensure that those that didn't get TODOs and issues get tracked by issues.

PR #6119, opened 2024-02-13
[tools] Ignore analysis options files in .symlinks

stuartmorgan:
When running `analyze` in a local tree that has done builds, there can be false positives of the "unexpected analysis options" check due to the .symlinks directory making other packages' analysis options show up in the check. This avoids following links to prevent those false positives.


tarrinneal (approved):
Thank you for fixing this! Can all tooling ignore sylink files?

PR #6060, opened 2024-02-05
[interactive_media_ads] Creates and adds the `interactive_media_ads` plugin

bparrishMines:
Part of https://github.com/flutter/flutter/issues/134228

Initial implementation of the plugin. 

* Adds minimum changes to pass CI (licenses, gradle, podpec changes, etc..).
* Adds platform interface and tests for it.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[...]

PR #6132, opened 2024-02-14
Raise gradle wrapper version of legacy project

bparrishMines:
When running `flutter create` for a new plugin on `master`, this error is printed from the legacy build_all_packages test.

Screenshot 2024-02-09 at 3 08 35 PM

It looks like only 6.7.1 is required to work with new plugins, so this is only a minor bump.


stuartmorgan (approved):
LGTM

(FYI @reidbaker)

reidbaker:
I dont see why updating to 6.7.1 would fix either compileSdkVersion not being set or calling substring on a null. 


https://docs.gradle.org/6.7.1/release-notes.html

bparrishMines:
@reidbaker I found the full error again: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8756540924836675809/+/u/Run_package_tests/build_all_packages_for_Android_-_legacy_version/stdout?format=raw

It looks like it is the minimum for `kotlin-android`

```
> Failed to apply plugin 'kotlin-android'.
   > The current Gradle version 6.7 is not compatible with the Kotlin Gradle plugin. Please use Gradle 6.7.1 or newer, or the previous version of the Kotlin plugin.
```

The [IMA plugin](https://github.com/flutter/packages/pull/6060) uses Kotlin.

reidbaker (approved):

PR #6140, opened 2024-02-15
Explicitly pass the web renderer into the tests.

eyebrowsoffire:
The web tests currently all assume to be using the html renderer. `flutter test` previously erroneously was serving files for the HTML renderer by default (which is not the intended behavior). After this was fixed, the tests started running on the CanvasKit renderer and some of them broke. If we are relying on the html renderer, we should explicitly pass it when running `flutter test`


stuartmorgan:
This tooling has unit test, so this will need a test.

@ditman It seems like we have a coverage problem here; is there an issue tracking the fact that we are only testing with one renderer?

eyebrowsoffire:
@stuartmorgan I went ahead and filed https://github.com/flutter/flutter/issues/143543

stuartmorgan (approved):
LGTM.

Do we need to do the same thing for [`flutter drive`](https://github.com/flutter/packages/blob/main/script/tool/lib/src/drive_examples_command.dart)?

eyebrowsoffire:
> Do we need to do the same thing for [flutter drive](https://github.com/flutter/packages/blob/main/script/tool/lib/src/drive_examples_command.dart)?

Probably a good idea. Let me look

PR #6147, opened 2024-02-16
[ci/various] Update for 3.19 stable

stuartmorgan:
Does the steps from https://github.com/flutter/flutter/wiki/Updating-Packages-repo-for-a-stable-release to account for stable now being 3.19, and N-2 (and thus our minimum supported min version) being 3.13.

These changes are version-bump-exempt [by repo policy](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version).

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -295,10 +295,10 @@ targets:
     timeout: 30
     properties:
       target_file: analyze_legacy.yaml
-      channel: "3.13.9"
+      channel: "3.16.9"
       env_variables: >-
         {
-          "CHANNEL": "3.13.9"
+          "CHANNEL": "3.16.9"
         }
 
   - name: Linux analyze_legacy N-2
@@ -306,10 +306,10 @@ targets:
     timeout: 30
     properties:
       target_file: analyze_legacy.yaml
-      channel: "3.10.6"
+      channel: "3.13.9"
       env_variables: >-
         {
-          "CHANNEL": "3.10.6"
+          "CHANNEL": "3.13.9"
         }
 
   - name: Linux_android custom_package_tests master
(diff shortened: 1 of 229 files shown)


.ci.yaml
```diff
@@ -295,21 +295,21 @@ targets:
     timeout: 30
     properties:
       target_file: analyze_legacy.yaml
-      channel: "3.13.9"
+      channel: "3.16.9"
```
reidbaker, line 298: Why does most of this pr update to 3.13 but the legacy test updates to 3.16? Is the legacy test by definition only one flutter release old?

.ci/targets/repo_checks.yaml
```diff
@@ -19,7 +19,7 @@ tasks:
     script: .ci/scripts/tool_runner.sh
     args:
       - "pubspec-check"
-      - "--min-min-flutter-version=3.10.0"
+      - "--min-min-flutter-version=3.13.0"
```
reidbaker, line 22: nonblocking? Why not make this the highest patch version?

reidbaker (approved):

.ci.yaml
stuartmorgan: > Why does most of this pr update to 3.13 but the legacy test updates to 3.16? Is the legacy test by definition only one flutter release old?

There are two legacy tests: N-1, and N-2. This is the N-1 test, which is by definition one release old, and the one just below is N-2, which is by definition two releases old.

(We have both main because if a package decides to drop N-2 but still support N-1, we want to make sure it doesn't accidentally uses APIs from N, but it won't be able to run the N-2 tests.)

ditman:
Thanks for bumping this!

PR #6162, opened 2024-02-19
[in_app_purchase_android] Add UserChoiceBilling mode. 

reidbaker:
Add UserChoiceBilling billing mode option. 

Fixes flutter/flutter/issues/143004 

Left in draft until: 
- [x] connect, disconnect and set user mode test is added to java
- [x] dart tests for method channel 
- [x] Callback api added to dart 
- [x] Test for call back api
- [x] Example using callback api in example app

This does not have an End to end working example with play integration. I am currently stuck at the server side play integration part. 

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/in_app_purchase/in_app_purchase_android/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 0.3.2
 
+* Adds UserChoiceBillingapis to platform addition.
```
stuartmorgan, line 3: "`UserChoiceBilling` APIs"?

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
```diff
@@ -15,11 +18,31 @@ final class BillingClientFactoryImpl implements BillingClientFactory {
...
     BillingClient.Builder builder = BillingClient.newBuilder(context).enablePendingPurchases();
-    if (billingChoiceMode == BillingChoiceMode.ALTERNATIVE_BILLING_ONLY) {
-      // https://developer.android.com/google/play/billing/alternative/alternative-billing-without-user-choice-in-app
-      builder.enableAlternativeBillingOnly();
+    switch (billingChoiceMode) {
+      case BillingChoiceMode.ALTERNATIVE_BILLING_ONLY:
+        // https://developer.android.com/google/play/billing/alternative/alternative-billing-without-user-choice-in-app
+        builder.enableAlternativeBillingOnly();
+        break;
+      case BillingChoiceMode.USER_CHOICE_BILLING:
```
stuartmorgan, line 31: Should this have a docs link too, like the above?

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
```diff
@@ -15,11 +18,31 @@ final class BillingClientFactoryImpl implements BillingClientFactory {
...
+          Log.e(
+              "BillingClientFactoryImpl",
+              "userChoiceBillingListener null when USER_CHOICE_BILLING set. Defaulting to PLAY_BILLING_ONLY");
+        }
+        break;
+      case BillingChoiceMode.PLAY_BILLING_ONLY:
+        // Do nothing
+        break;
+      default:
+        Log.w("BillingClientFactoryImpl", "Unknown BillingChoiceMode " + billingChoiceMode);
```
stuartmorgan, line 44: Shouldn't this be `.e`, and also mention that it's doing fallback to Play-only?

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
```diff
@@ -233,6 +235,34 @@ static HashMap<String, Object> fromBillingResult(BillingResult billingResult) {
     return info;
   }
 
+  static HashMap<String, Object> fromUserChoiceDetails(UserChoiceDetails userChoiceDetails) {
+    HashMap<String, Object> info = new HashMap<>();
+    info.put("externalTransactionToken", userChoiceDetails.getExternalTransactionToken());
+    info.put("originalExternalTransactionId", userChoiceDetails.getOriginalExternalTransactionId());
+    info.put("productsList", fromProductsList(userChoiceDetails.getProducts()));
```
stuartmorgan, line 242: The Dart object's field is just called `products`; is `List` added automagically by JSON serialization, or is this a mismatch?

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
```diff
@@ -233,6 +235,34 @@ static HashMap<String, Object> fromBillingResult(BillingResult billingResult) {
...
+    for (Product product : productsList) {
+      output.add(fromProduct(product));
+    }
+    return output;
+  }
+
+  static HashMap<String, Object> fromProduct(Product product) {
+    HashMap<String, Object> info = new HashMap<>();
+    info.put("productId", product.getId());
+    info.put("productOfferToken", product.getOfferToken());
```
stuartmorgan, line 260: Why do these two have an extra "product" prefix relative to Dart (but the type below matches directly)?

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -507,6 +532,10 @@ enum BillingChoiceMode {
   /// Billing through app provided flow.
   @JsonValue(1)
   alternativeBillingOnly,
+
+  /// Users can choose play billing or alternative billing.
```
stuartmorgan, line 536: Nit: Play

packages/in_app_purchase/in_app_purchase_android/CHANGELOG.md
reidbaker: done

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
reidbaker: Done

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
reidbaker: done

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/billing_client_wrapper.dart
reidbaker: Done and fixed some other locations in this file.

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
reidbaker: Confirmed mismatch and added test.

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
reidbaker: Mismatch, test added.

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
```diff
@@ -15,11 +18,34 @@ final class BillingClientFactoryImpl implements BillingClientFactory {
...
+          // https://developer.android.com/google/play/billing/alternative/alternative-billing-with-user-choice-in-app
+          builder.enableUserChoiceBilling(userChoiceBillingListener);
+        } else {
+          Log.e(
+              "BillingClientFactoryImpl",
+              "userChoiceBillingListener null when USER_CHOICE_BILLING set. Defaulting to PLAY_BILLING_ONLY");
+        }
+        break;
+      case BillingChoiceMode.PLAY_BILLING_ONLY:
+        // Do nothing
```
gmackall, line 42: nit: end comment with period

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
```diff
@@ -15,11 +18,34 @@ final class BillingClientFactoryImpl implements BillingClientFactory {
...
+        // https://developer.android.com/google/play/billing/alternative/alternative-billing-without-user-choice-in-app
+        builder.enableAlternativeBillingOnly();
+        break;
+      case BillingChoiceMode.USER_CHOICE_BILLING:
+        if (userChoiceBillingListener != null) {
+          // https://developer.android.com/google/play/billing/alternative/alternative-billing-with-user-choice-in-app
+          builder.enableUserChoiceBilling(userChoiceBillingListener);
+        } else {
+          Log.e(
+              "BillingClientFactoryImpl",
```
gmackall, line 37: What do you think about enforcing this being an error case, instead of defaulting? I think I might be confusing for developers to default to `PLAY_BILLING_ONLY`, even if they have made a mistake in how they configured for `USER_CHOICE_BILLING`. 

Not a strong preference, though, as long as the messaging is clear

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java
```diff
@@ -233,6 +235,34 @@ static HashMap<String, Object> fromBillingResult(BillingResult billingResult) {
     return info;
   }
 
+  static HashMap<String, Object> fromUserChoiceDetails(UserChoiceDetails userChoiceDetails) {
```
gmackall, line 238: Nit: add docs links to the places we are essentially copying the structure of android classes here, so 
https://developer.android.com/reference/com/android/billingclient/api/UserChoiceDetails#getOriginalExternalTransactionId() here 
and 
https://developer.android.com/reference/com/android/billingclient/api/UserChoiceDetails.Product
below

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
gmackall: After reading further, we should never reach this case right? We define the listener, and we basically do it as a pass through where we pass the `UserChoiceDetails` in a copied hashmap to the dart side?

And the only time that listener is null is when the `BillingChoiceMode` is not `USER_CHOICE_BILLING`?

Can't we instead move that logic where we construct the listener (lines 547-557 in `MethodCallHandlerImpl`) inside this class and only invoke it when we actually need to make the listener, so it is never null?

gmackall (commented):
partial (java side)

packages/in_app_purchase/in_app_purchase_android/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 0.3.2
 
+* Adds UserChoiceBilling API's to platform addition.
```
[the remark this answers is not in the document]
reidbaker: done

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
reidbaker: done

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
reidbaker: I agree that we should not every hit this case as the code is currently written but I added the additional fallaback logic to protect against future bugs and give us a way to know that we were in a bad state. 

I didnt want to create the listener in this class because it needed to invoke methodChannel and I thought the methodchannelimpl class would be where you would look for that logic.

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -18,6 +18,12 @@ part 'billing_client_wrapper.g.dart';
 @visibleForTesting
 const String kOnPurchasesUpdated =
     'PurchasesUpdatedListener#onPurchasesUpdated(BillingResult, List<Purchase>)';
+
+/// Method idendifier for the userSelectedAlternativeBilling method channel method.
```
gmackall, line 22:
```suggestion
/// Method identifier for the userSelectedAlternativeBilling method channel method.
```

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImpl.java
gmackall: Part of why I suggested this is that we have existing code where we do something similar to construct a listener in this plugin, see https://github.com/flutter/packages/blob/a10b360a21531d03fb3795b7c083dda5c4796f96/packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/PluginPurchaseListener.java#L39.

I still lean towards moving the listener construction, so we can eliminate the bad case here, but if you have a strong preference to keep the structure as is I'm fine that way too

reidbaker: How about a compromise. I am worried about the march 13th deadline for apps to 
 migrate to this api. I have filed https://github.com/flutter/flutter/issues/144851 and assigned it to myself along with some other cleanup bugs that I can do without impacting the public facing api.

gmackall (approved):
LGTM with https://github.com/flutter/flutter/issues/144851 as a follow up, considering deadline for developers to adopt these new apis is soon

PR #6166, opened 2024-02-20
[ci] Promote Windows Arm64 build master tests

loic-sharma:
Part of https://github.com/flutter/flutter/issues/129813

## Pre-launch Checklist

- [x] I signed the [CLA].


.ci.yaml
```diff
@@ -1490,6 +1491,7 @@ targets:
 
   - name: Windows_x64 windows-build_all_packages stable
     recipe: packages/packages
+    presubmit: false
```
stuartmorgan, line 1494: I assume this was supposed to go on line 1510 instead (Win arm64 stable)?

If capacity is low, I would suggest just deleting that one instead. The chances of breaking only the Windows arm64 build of plugins only on stable seems extremely low, so we could halve our Windows arm64 usage for this repo with probably no appreciable impact on coverage.

loic-sharma: Good catch, fixed!

The majority of Windows Arm64 capacity is used by Dart's tests, not Flutter's tests. I think keeping the Windows Arm64 stable test is fine unless you'd prefer to remove it.

My understanding is that the infra team would need to significantly increase their number of Windows Arm64 machines to also support the try pool, so for now they're skipping that.

stuartmorgan: I would probably remove it absent some reason we believe it will catch actual issues that the other bot won't, but I don't feel strongly about it.

stuartmorgan (approved):
LGTM

PR #6168, opened 2024-02-21
[flutter_markdown] Support wasm

kevmoo:
Drop usage of dart:html

### DIFF
--- a/packages/flutter_markdown/lib/src/_functions_web.dart
+++ b/packages/flutter_markdown/lib/src/_functions_web.dart
@@ -3,7 +3,6 @@
 // found in the LICENSE file.
 
 import 'dart:js_interop';
-import 'dart:js_interop_unsafe';
 
 import 'package:flutter/cupertino.dart' show CupertinoTheme;
 import 'package:flutter/material.dart' show Theme;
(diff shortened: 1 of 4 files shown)


matanlurey:
test-exempt: non-semantic refactor

packages/flutter_markdown/lib/src/_functions_web.dart
```diff
@@ -50,20 +51,15 @@ final MarkdownStyleSheet Function(BuildContext, MarkdownStyleSheetBaseTheme?)
...
-    case MarkdownStyleSheetBaseTheme.material:
-    default: // ignore: no_default_cases
-      result = MarkdownStyleSheet.fromTheme(Theme.of(context));
-  }
+  final MarkdownStyleSheet result = switch (baseTheme) {
+    MarkdownStyleSheetBaseTheme.platform when _userAgent.contains('Mac OS X') =>
+      MarkdownStyleSheet.fromCupertinoTheme(CupertinoTheme.of(context)),
+    MarkdownStyleSheetBaseTheme.cupertino =>
+      MarkdownStyleSheet.fromCupertinoTheme(CupertinoTheme.of(context)),
+    MarkdownStyleSheetBaseTheme.material ||
```
srujzs, line 59: nit: unnecessary, right?

packages/flutter_markdown/lib/src/_functions_web.dart
```diff
@@ -84,3 +80,9 @@ Widget _handleDataSchemeUri(
   }
   return const SizedBox();
 }
+
+final String _userAgent = globalContext
```
srujzs, line 84: I'm guessing you can't use `package:web`?

srujzs (approved):

packages/flutter_markdown/lib/src/_functions_web.dart
kevmoo: was trying to avoid it since it's for ONE SINGLE property access.

packages/flutter_markdown/lib/src/_functions_web.dart
kevmoo: yeah, trying to keep the existing...whatever...

packages/flutter_markdown/lib/src/_functions_web.dart
srujzs: Another alternative is using a dotted top-level:

```
@JS('window.navigator.userAgent')
external JSObject get userAgent;
```

kevmoo: Could be `JSString` return value?

srujzs: Sure. I should have double-checked the type before posting the example.

PR #6183, opened 2024-02-22
[pigeon] Allows kotlin generator to skip error class generation

tarrinneal:
Allows kotlin generator to skip error class generation by creating `includeErrorClass` argument.

This may be considered only a partial fix, as it requires another pigeon file with the class generated rather than creating a new file with the class.

This solution seems to me to solve the problem well enough as creating another file is a non-trivial task and there is already a working solution - renaming the error class per file.

fixes: https://github.com/flutter/flutter/issues/142099

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/pigeon/CHANGELOG.md
+++ b/packages/pigeon/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 17.0.1
+## 17.1.0
 
 * [kotlin] Adds `includeErrorClass` to `KotlinOptions`.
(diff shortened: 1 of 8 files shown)


stuartmorgan (commented):
How does sharing work without an import of the file that defines it?

tarrinneal:
> How does sharing work without an import of the file that defines it?

As long as it's in the same directory, an import isn't necessary. That's why this is a problem in the first place.

stuartmorgan (approved):
Weird; LGTM then.

PR #6158, opened 2024-02-19
[google_maps_flutter] Add support for marker clustering - platform interface

jokerttu:
This PR introduces support for marker clustering

This is prequel PR for: https://github.com/flutter/packages/pull/4319
Containing only changes to `google_maps_flutter_platform_interface` package.

Follow up PR:s will hold the platform and app-facing plugin implementations.

Linked issue: https://github.com/flutter/flutter/issues/26863

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[...]

PR #6185, opened 2024-02-23
[google_maps_flutter] Add marker clustering support - android implementation

jokerttu:
This PR introduces support for marker clustering for Android platform

An example usage is available in the example application at `./packages/google_maps_flutter/google_maps_flutter_android/example` on the page `Manage clustering`

Containing only changes to `google_maps_flutter_android` package.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
@@ -92,8 +92,9 @@ class ClusterManagersController
       throw new IllegalArgumentException("clusterManagerId was null");
     }
     ClusterManager<MarkerBuilder> clusterManager =
-        new ClusterManager<>(context, googleMap, markerManager);
-    ClusterRenderer clusterRenderer = new ClusterRenderer(context, googleMap, clusterManager, this);
+        new ClusterManager<MarkerBuilder>(context, googleMap, markerManager);
+    ClusterRenderer<MarkerBuilder> clusterRenderer =
+        new ClusterRenderer<MarkerBuilder>(context, googleMap, clusterManager, this);
     clusterManager.setRenderer(clusterRenderer);
     initListenersForClusterManager(clusterManager, this, clusterItemClickListener);
     clusterManagerIdToManager.put(clusterManagerId, clusterManager);
@@ -209,13 +210,13 @@ class ClusterManagersController
    * ClusterRenderer builds marker options for new markers to be rendered to the map. After cluster
    * item (marker) is rendered, it is sent to the listeners for control.
    */
-  private static class ClusterRenderer extends DefaultClusterRenderer<MarkerBuilder> {
+  private static class ClusterRenderer<T extends MarkerBuilder> extends DefaultClusterRenderer<T> {
     private final ClusterManagersController clusterManagersController;
 
     public ClusterRenderer(
         Context context,
         GoogleMap map,
-        ClusterManager<MarkerBuilder> clusterManager,
+        ClusterManager<T> clusterManager,
         ClusterManagersController clusterManagersController) {
       super(context, map, clusterManager);
       this.clusterManagersController = clusterManagersController;
@@ -223,14 +224,14 @@ class ClusterManagersController
 
     @Override
     protected void onBeforeClusterItemRendered(
-        @NonNull MarkerBuilder item, @NonNull MarkerOptions markerOptions) {
+        @NonNull T item, @NonNull MarkerOptions markerOptions) {
       // Builds new markerOptions for new marker created by the ClusterRenderer under
       // ClusterManager.
       item.update(markerOptions);
     }
 
     @Override
-    protected void onClusterItemRendered(@NonNull MarkerBuilder item, @NonNull Marker marker) {
+    protected void onClusterItemRendered(@NonNull T item, @NonNull Marker marker) {
       super.onClusterItemRendered(item, marker);
       clusterManagersController.onClusterItemRendered(item, marker);
     }
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
@@ -160,6 +160,8 @@ class ClusterManagersController
   @SuppressWarnings("unchecked")
   private static String getClusterManagerId(Object clusterManagerData) {
     Map<String, Object> clusterMap = (Map<String, Object>) clusterManagerData;
+    // The "clusterManagerId" key is set at:
+    // https://github.com/flutter/packages/blob/dce6f0c91fcf01ee3a67e3475155ec9f16ce2a66/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cluster_manager.dart#L62
     return (String) clusterMap.get("clusterManagerId");
   }
 
@@ -172,7 +174,9 @@ class ClusterManagersController
     ClusterManager<MarkerBuilder> clusterManager = clusterManagerIdToManager.get(clusterManagerId);
     if (clusterManager == null) {
       result.error(
-          "Invalid clusterManagerId", "getClusters called with invalid clusterManagerId", null);
+          "Invalid clusterManagerId",
+          "getClusters called with invalid clusterManagerId:" + clusterManagerId,
+          null);
       return;
     }
 
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
@@ -1,0 +1,239 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package io.flutter.plugins.googlemaps;
+
+import android.content.Context;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.android.gms.maps.GoogleMap;
+import com.google.android.gms.maps.model.Marker;
+import com.google.android.gms.maps.model.MarkerOptions;
+import com.google.maps.android.clustering.Cluster;
+import com.google.maps.android.clustering.ClusterItem;
+import com.google.maps.android.clustering.ClusterManager;
+import com.google.maps.android.clustering.view.DefaultClusterRenderer;
+import com.google.maps.android.collections.MarkerManager;
+import io.flutter.plugin.common.MethodChannel;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Controls cluster managers and exposes interfaces for adding and removing cluster items for
+ * specific cluster managers.
+ */
+class ClusterManagersController
+    implements GoogleMap.OnCameraIdleListener,
+        ClusterManager.OnClusterClickListener<MarkerBuilder> {
+  @NonNull private final Context context;
+  @NonNull private final HashMap<String, ClusterManager<MarkerBuilder>> clusterManagerIdToManager;
+  @NonNull private final MethodChannel methodChannel;
+  @Nullable private MarkerManager markerManager;
+  @Nullable private GoogleMap googleMap;
+
+  @Nullable
+  private ClusterManager.OnClusterItemClickListener<MarkerBuilder> clusterItemClickListener;
+
+  @Nullable
+  private ClusterManagersController.OnClusterItemRendered<MarkerBuilder>
+      clusterItemRenderedListener;
+
+  ClusterManagersController(MethodChannel methodChannel, Context context) {
+    this.clusterManagerIdToManager = new HashMap<>();
+    this.context = context;
+    this.methodChannel = methodChannel;
+  }
+
+  void init(GoogleMap googleMap, MarkerManager markerManager) {
+    this.markerManager = markerManager;
+    this.googleMap = googleMap;
+  }
+
+  void setClusterItemClickListener(
+      @Nullable ClusterManager.OnClusterItemClickListener<MarkerBuilder> listener) {
+    clusterItemClickListener = listener;
+    initListenersForClusterManagers();
+  }
+
+  void setClusterItemRenderedListener(
+      @Nullable ClusterManagersController.OnClusterItemRendered<MarkerBuilder> listener) {
+    clusterItemRenderedListener = listener;
+  }
+
+  private void initListenersForClusterManagers() {
+    for (Map.Entry<String, ClusterManager<MarkerBuilder>> entry :
+        clusterManagerIdToManager.entrySet()) {
+      initListenersForClusterManager(entry.getValue(), this, clusterItemClickListener);
+    }
+  }
+
+  private void initListenersForClusterManager(
+      ClusterManager<MarkerBuilder> clusterManager,
+      @Nullable ClusterManager.OnClusterClickListener<MarkerBuilder> clusterClickListener,
+      @Nullable ClusterManager.OnClusterItemClickListener<MarkerBuilder> clusterItemClickListener) {
+    clusterManager.setOnClusterClickListener(clusterClickListener);
+    clusterManager.setOnClusterItemClickListener(clusterItemClickListener);
+  }
+
+  /** Adds new ClusterManagers to the controller. */
+  void addClusterManagers(@NonNull List<Object> clusterManagersToAdd) {
+    for (Object clusterToAdd : clusterManagersToAdd) {
+      addClusterManager(clusterToAdd);
+    }
+  }
+
+  /** Adds new ClusterManager to the controller. */
+  void addClusterManager(Object clusterManagerData) {
+    String clusterManagerId = getClusterManagerId(clusterManagerData);
+    if (clusterManagerId == null) {
+      throw new IllegalArgumentException("clusterManagerId was null");
+    }
+    ClusterManager<MarkerBuilder> clusterManager =
+        new ClusterManager<>(context, googleMap, markerManager);
+    ClusterRenderer clusterRenderer = new ClusterRenderer(context, googleMap, clusterManager, this);
+    clusterManager.setRenderer(clusterRenderer);
+    initListenersForClusterManager(clusterManager, this, clusterItemClickListener);
+    clusterManagerIdToManager.put(clusterManagerId, clusterManager);
+  }
+
+  /** Removes ClusterManagers by given cluster manager IDs from the controller. */
+  public void removeClusterManagers(@NonNull List<Object> clusterManagerIdsToRemove) {
+    for (Object rawClusterManagerId : clusterManagerIdsToRemove) {
+      if (rawClusterManagerId == null) {
+        continue;
+      }
+      String clusterManagerId = (String) rawClusterManagerId;
+      removeClusterManager(clusterManagerId);
+    }
+  }
+
+  /**
+   * Removes the ClusterManagers by the given cluster manager ID from the controller. The reference
+   * to this cluster manager is removed from the clusterManagerIdToManager and it will be garbage
+   * collected later.
+   */
+  private void removeClusterManager(Object clusterManagerId) {
+    // Remove the cluster manager from the hash map to allow it to be garbage collected.
+    final ClusterManager<MarkerBuilder> clusterManager =
+        clusterManagerIdToManager.remove(clusterManagerId);
+    if (clusterManager == null) {
+      return;
+    }
+    initListenersForClusterManager(clusterManager, null, null);
+    clusterManager.clearItems();
+    clusterManager.cluster();
+  }
+
+  /** Adds item to the ClusterManager it belongs to. */
+  public void addItem(MarkerBuilder item) {
+    ClusterManager<MarkerBuilder> clusterManager =
+        clusterManagerIdToManager.get(item.clusterManagerId());
+    if (clusterManager != null) {
+      clusterManager.addItem(item);
+      clusterManager.cluster();
+    }
+  }
+
+  /** Removes item from the ClusterManager it belongs to. */
+  public void removeItem(MarkerBuilder item) {
+    ClusterManager<MarkerBuilder> clusterManager =
+        clusterManagerIdToManager.get(item.clusterManagerId());
+    if (clusterManager != null) {
+      clusterManager.removeItem(item);
+      clusterManager.cluster();
+    }
+  }
+
+  /** Called when ClusterRenderer has rendered new visible marker to the map. */
+  void onClusterItemRendered(@NonNull MarkerBuilder item, @NonNull Marker marker) {
+    // If map is being disposed, clusterItemRenderedListener might have been cleared and
+    // set to null.
+    if (clusterItemRenderedListener != null) {
+      clusterItemRenderedListener.onClusterItemRendered(item, marker);
+    }
+  }
+
+  /** Reads clusterManagerId from object data. */
+  @SuppressWarnings("unchecked")
+  private static String getClusterManagerId(Object clusterManagerData) {
+    Map<String, Object> clusterMap = (Map<String, Object>) clusterManagerData;
+    return (String) clusterMap.get("clusterManagerId");
+  }
+
+  /**
+   * Requests all current clusters from the algorithm of the requested ClusterManager and converts
+   * them to result response.
+   */
+  public void getClustersWithClusterManagerId(
+      String clusterManagerId, MethodChannel.Result result) {
+    ClusterManager<MarkerBuilder> clusterManager = clusterManagerIdToManager.get(clusterManagerId);
+    if (clusterManager == null) {
+      result.error(
+          "Invalid clusterManagerId", "getClusters called with invalid clusterManagerId", null);
+      return;
+    }
+
+    final Set<? extends Cluster<MarkerBuilder>> clusters =
+        clusterManager.getAlgorithm().getClusters(googleMap.getCameraPosition().zoom);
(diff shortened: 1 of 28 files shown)


reidbaker:
Did not mean to close.

reidbaker (commented):
Partial review. Sorry you waited so long this slipped though my radar. Will continue the review tomorrow.

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/MarkersControllerTest.java
camsim99, line 1: Can you test the functionality you added to `onMapsMarkerTap`, `onClusterItemMarker`, `changeMarker` here or wherever is appropriate? Seems like they are called by `GoogleMapsController`, but testing there would require faking the method calls I assume.

camsim99 (commented):
Looks great! Just left some comments about nits, tests, and naming :) Would not merge until @reidbaker gives a second review though

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
```diff
@@ -0,0 +1,222 @@
...
+        addClusterManager(clusterToAdd);
+      }
+    }
+  }
+
+  /** Adds new ClusterManager. */
+  void addClusterManager(Object clusterManagerData) {
+    String clusterManagerId = getClusterManagerId(clusterManagerData);
+    if (clusterManagerId == null) {
+      throw new IllegalArgumentException("clusterManagerId was null");
```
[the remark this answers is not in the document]
jokerttu: How would you like me to change this?
Exceptions thrown in the native code are converted to PlatformExceptions by the method channel and this is handled as an PlatformException at dart code.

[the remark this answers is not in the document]
jokerttu: IllegalArgumentExceptions are used in the Android platform on similar places.
Exceptions thrown in the method channel method handlers are converted to PlatformExceptions by the method channel code:

E/flutter (12022): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: **PlatformException**(error, clusterManagerId was null, null, java.lang.IllegalArgumentException: clusterManagerId was null

As clusterManagerUpdates should be created by the [ClusterManagerUpdates](https://github.com/flutter/packages/blob/dce6f0c91fcf01ee3a67e3475155ec9f16ce2a66/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cluster_manager_updates.dart#L11) class this error should not be never seen by the developers.

How would you like me to change this handling?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
```diff
@@ -0,0 +1,222 @@
...
+    if (clusterItemListener != null) {
+      clusterItemListener.onClusterItemMarker(item, marker);
+    }
+  }
+
+  /** Reads clusterManagerId from object data. */
+  @SuppressWarnings("unchecked")
+  private static String getClusterManagerId(Object clusterManagerData) {
+    Map<String, Object> clusterMap = (Map<String, Object>) clusterManagerData;
+    return (String) clusterMap.get("clusterManagerId");
```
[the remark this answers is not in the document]
jokerttu: Current implementation of the google maps plugin uses custom serialisation for method channel communication.
These strings are hardcoded through the native implementation matching the data structure serialised at the dart layer.

Corresponding mapping for this is done here:
https://github.com/flutter/packages/blob/dce6f0c91fcf01ee3a67e3475155ec9f16ce2a66/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cluster_manager.dart#L62

Should I add this link to before the line, as this might get outdated (even with the ref) at some point.
The current communication layer relies a lot of these hardcoded mapping keys.
Moving the plugin method channel over pigeon in the future would improve the type-safety a lot.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -149,18 +156,23 @@ public void onMapReady(GoogleMap googleMap) {
...
     if (mapReadyResult != null) {
       mapReadyResult.success(null);
       mapReadyResult = null;
     }
-    setGoogleMapListener(this);
+    markerManager = new MarkerManager(googleMap);
+    markerCollection = markerManager.newCollection();
     updateMyLocationSettings();
-    markersController.setGoogleMap(googleMap);
+    markersController.setCollection(markerCollection);
```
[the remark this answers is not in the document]
jokerttu: There is no reason to move the `setGoogleMapListener(this)` listener initialization to different position and was only done to re-organize listeners to same place and improve readability; I reverted this change as it is not needed by this PR.

Some of the listeners are removed (eg. `googleMap.setOnInfoWindowClickListener(this)`) as those are handled by the markerCollection listeners set in `setMarkerCollectionListener`

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
```diff
@@ -0,0 +1,222 @@
...
+    @Override
+    protected void onClusterItemRendered(@NonNull MarkerBuilder item, @NonNull Marker marker) {
+      super.onClusterItemRendered(item, marker);
+      clusterManagersController.onClusterItemMarker(item, marker);
+    }
+  }
+
+  /** Interface for handling situations where clusterManager adds new visible marker to the map. */
+  public interface OnClusterItemMarker<T extends ClusterItem> {
+    void onClusterItemMarker(@NonNull T item, @NonNull Marker marker);
```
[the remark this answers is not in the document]
jokerttu: renamed the method and interface to match the method call where it is called: `onClusterItemRendered`

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterItemListener.java
```diff
@@ -0,0 +1,11 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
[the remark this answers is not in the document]
jokerttu: This method combines both:
ClusterManagersController.OnClusterItemMarker<MarkerBuilder>,
and 
ClusterManager.OnClusterItemClickListener<MarkerBuilder>
under same listener.

Listener with type ClusterItemListener is registered for clusterManagerController.
It is possible to have two listenerer objects instead of this combinations, and two setter functions to set the listener.

Instead of method   
```
void setClusterItemListener(@Nullable ClusterItemListener listener) {
    clusterItemListener = listener;
    initListenersForClusterManagers(this, listener);
}
```

there need to be two methods:

```
  void setOnClusterItemClickListener(@Nullable ClusterManager.OnClusterItemClickListener<MarkerBuilder> listener) {
    onClusterItemClickListener= listener;
    initListenersForClusterManagers(this, listener);
  }
```
  
  and
  
```
  void setOnClusterItemRenderedListener(@Nullable ClusterManagersController.OnClusterItemMarker<MarkerBuilder>listener) {
    onClusterItemRenderedListener = listener;
  }
```

Is this approach better?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterListener.java
```diff
@@ -0,0 +1,9 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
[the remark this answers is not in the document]
jokerttu: Removed this file

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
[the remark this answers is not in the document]
jokerttu: I don't know good way to do this, as this method is called by the the google maps sdk when the marker on map is clicked.
It could be possible to position marker in middle of the map and try to tap that location programmically, but that sounds a bit flaky?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
```diff
@@ -55,11 +59,27 @@ void removeMarkers(List<Object> markerIdsToRemove) {
...
+      return;
+    }
+    final MarkerController markerController = markerIdToController.remove(markerId);
+    final String clusterManagerId = markerBuilder.clusterManagerId();
+    if (clusterManagerId != null) {
+      // Remove marker from clusterManager.
+      clusterManagersController.removeItem(markerBuilder);
+    } else if (markerController != null && this.markerCollection != null) {
+      // Remove marker from map and markerCollection
+      markerController.removeFromCollection(markerCollection);
```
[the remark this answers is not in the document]
jokerttu: When marker does not have clusterManagerId it is added to the map by MarkersController and need to be removed from the map. If marker is set to be controller by the clusterManager, the marker is removed from clusterManager istead.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
```diff
@@ -0,0 +1,222 @@
...
+        clusterManagerIdToManager.get(item.clusterManagerId());
+    if (clusterManager != null) {
+      clusterManager.removeItem(item);
+      clusterManager.cluster();
+    }
+  }
+
+  /** Called when ClusterRenderer has rendered new visible marker to the map. */
+  void onClusterItemMarker(@NonNull MarkerBuilder item, @NonNull Marker marker) {
+    if (clusterItemListener != null) {
```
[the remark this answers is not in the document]
jokerttu: Added comment

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkerBuilder.java
```diff
@@ -7,23 +7,54 @@
...
+    markerOptionsToUpdate.alpha(markerOptions.getAlpha());
+    markerOptionsToUpdate.anchor(markerOptions.getAnchorU(), markerOptions.getAnchorV());
+    markerOptionsToUpdate.draggable(markerOptions.isDraggable());
+    markerOptionsToUpdate.flat(markerOptions.isFlat());
+    markerOptionsToUpdate.icon(markerOptions.getIcon());
+    markerOptionsToUpdate.infoWindowAnchor(
+        markerOptions.getInfoWindowAnchorU(), markerOptions.getInfoWindowAnchorV());
+    markerOptionsToUpdate.title(markerOptions.getTitle());
+    markerOptionsToUpdate.snippet(markerOptions.getSnippet());
+    markerOptionsToUpdate.position(markerOptions.getPosition());
```
reidbaker, line 40: Why is positions updated on line 30 and here?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkerBuilder.java
```diff
@@ -7,23 +7,54 @@
...
+    this.markerId = markerId;
+    this.clusterManagerId = clusterManagerId;
   }
 
   MarkerOptions build() {
     return markerOptions;
   }
 
+  /** Update existing markerOptions with builder values */
+  void update(MarkerOptions markerOptionsToUpdate) {
```
reidbaker, line 29: So that I can understand can you explain how markerOptionsToUpdate updates information?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkerController.java
```diff
@@ -7,82 +7,139 @@
...
+  // Holds a weak reference to a Marker instance. The clustering library
+  // dynamically manages markers, adding and removing them from the map
+  // as needed based on user interaction or data changes.
+  private final WeakReference<Marker> weakMarker;
   private final String googleMapsMarkerId;
   private boolean consumeTapEvents;
 
   MarkerController(Marker marker, boolean consumeTapEvents) {
-    this.marker = marker;
+    this.weakMarker = new WeakReference<>(marker);
```
reidbaker, line 24: Can you add a comment here explaining why a weakreference is needed? I believe that you would not have added the complexity if it was not needed but it helps when someone else comes to modify the code.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
```diff
@@ -79,7 +99,10 @@ void hideMarkerInfoWindow(String markerId, MethodChannel.Result result) {
       markerController.hideInfoWindow();
       result.success(null);
     } else {
-      result.error("Invalid markerId", "hideInfoWindow called with invalid markerId", null);
+      result.error(
+          "Invalid markerId",
+          "hideInfoWindow called with invalid markerId or for hidden cluster marker",
```
reidbaker, line 104: Consider including the markerId in this error message.

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ConvertTest.java
```diff
@@ -26,4 +31,54 @@ public void ConvertToPointsConvertsThePointsWithFullPrecision() {
     Assert.assertEquals(latitude, latLng.latitude, 1e-15);
     Assert.assertEquals(longitude, latLng.longitude, 1e-15);
   }
+
+  @Test
+  public void ConvertClustersToJsonReturnsCorrectData() {
+    String clusterManagerId = "cm_1";
+    StaticCluster<MarkerBuilder> cluster = new StaticCluster<>(new LatLng(43.00, -87.90));
```
reidbaker, line 38: This test could be easier to read if you had position1/2/3 defined a the top then passed those positions into the constructor and compared to those positions in the asserts.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
```diff
@@ -146,18 +176,58 @@ void onInfoWindowTap(String googleMarkerId) {
...
+
   private void addMarker(Object marker) {
     if (marker == null) {
       return;
     }
-    MarkerBuilder markerBuilder = new MarkerBuilder();
-    String markerId = Convert.interpretMarkerOptions(marker, markerBuilder);
+    String markerId = getMarkerId(marker);
+    if (markerId == null) {
+      throw new IllegalArgumentException("markerId was null");
```
reidbaker, line 196: Specifically here but a comment more broadly about the java side code and tests. Please add some more coverage of the non happy path cases in tests.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -181,6 +182,11 @@ class GoogleMapsFlutterAndroid extends GoogleMapsFlutterPlatform {
     return _events(mapId).whereType<MapLongPressEvent>();
   }
 
+  @override
+  Stream<ClusterTapEvent> onClusterTap({required int mapId}) {
+    return _events(mapId).whereType<ClusterTapEvent>();
```
reidbaker, line 187: Document/test the behavior here if onClusterTap is called with an invalid mapId

packages/google_maps_flutter/google_maps_flutter_android/lib/src/utils/cluster_manager_utils.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:google_maps_flutter_platform_interface/google_maps_flutter_platform_interface.dart';
+
+/// Converts a Set of Cluster Managers into object serializable in JSON.
+Object serializeClusterManagerSet(Set<ClusterManager> clusterManagers) {
```
reidbaker, line 8: Please add cluster_manager_utils_test that covers these methods.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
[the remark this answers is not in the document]
stuartmorgan: (Tarrin just recently landed support for channel suffixes in Pigeon, which unblocks converting maps, so I'm hoping to do the conversion this quarter.)

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkerBuilder.java
jokerttu: Google maps markers are created with MarkerOptions object.
Items added to clusterManager are instances of MarkerBuilder that extends the ClusterItem class. The clustering process uses data from this MarkerBuilder instance to update the MarkerOptions, that it then uses to create or re-create the marker to the map.

It is important to store this data to MarkerBuilder object instance per each marker to allow to update marker data even when marker instance is not drawn on the map.
Here is clustering callback that uses this update method tuo pass the data from the MarkerBuilder to the markerOptions instance, the clustermanager uses to create the visible marker on the map.

```java
    @Override
    protected void onBeforeClusterItemRendered(
        @NonNull MarkerBuilder item, @NonNull MarkerOptions markerOptions) {
      // Builds new markerOptions for new marker created by the ClusterRenderer under
      // ClusterManager.
      item.update(markerOptions);
    }
```

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkerController.java
jokerttu: Isn't the comment on few lines above is enough explaining the need for the weakReference?
As the clustering process might remove the marker from the map (when clustered), the marker should not be accessed anymore in MarkerController. Instead markerbuilder is updated, so that if the marker is declustered, it is used to update the marker settings before  drawing.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
jokerttu: Reverted the change, as just noticed that this changes the existing behaviour and is breaking change.
Eventhough it would be nice to give extra information for the developers that the error could be caused because marker is removed by the clusterManager, changing the existing API might not be an option here.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
```diff
@@ -88,15 +111,22 @@ void isInfoWindowShown(String markerId, MethodChannel.Result result) {
     if (markerController != null) {
       result.success(markerController.isInfoWindowShown());
     } else {
-      result.error("Invalid markerId", "isInfoWindowShown called with invalid markerId", null);
+      result.error(
```
jokerttu: Reverted the change, as just noticed that this changes the existing behaviour and is breaking change.
Eventhough it would be nice to give extra information for the developers that the error could be caused because marker is removed by the clusterManager, changing the existing API might not be an option here.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
jokerttu: This is path that should never hit as the dart side always serializes markerId to the marker object. I'll add test to this in separate commit if possible.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -824,6 +872,19 @@ private void updateInitialMarkers() {
     markersController.addMarkers(initialMarkers);
   }
 
+  @Override
+  public void setInitialClusterManagers(Object initialClusterManagers) {
```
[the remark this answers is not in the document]
jokerttu: Added test cases to cover these methods

packages/google_maps_flutter/google_maps_flutter_android/lib/src/utils/cluster_manager_utils.dart
jokerttu: Added tests for cluster_manager_utils

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
jokerttu: Please check how this _events method is used elsewhere in the file.
_events function just creates event matcher that matches the mapId, and creating stream with invalid mapId will just create stream that does not match any events ever.
The stream is created for the current mapId and I don't see case where the mapId could be invalid.
Testing this method would test mostly the dart Stream functionality.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+  }
+
+  /** Reads clusterManagerId from object data. */
+  @SuppressWarnings("unchecked")
+  private static String getClusterManagerId(Object clusterManagerData) {
+    Map<String, Object> clusterMap = (Map<String, Object>) clusterManagerData;
+    // The "clusterManagerId" key is set at:
+    // https://github.com/flutter/packages/blob/dce6f0c91fcf01ee3a67e3475155ec9f16ce2a66/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cluster_manager.dart#L62
+    return (String) clusterMap.get("clusterManagerId");
```
camsim99, line 166: Nit: Maybe just point to the class/method this is in versus the link?

packages/google_maps_flutter/google_maps_flutter_android/lib/src/utils/cluster_manager_utils.dart
```diff
@@ -0,0 +1,34 @@
...
+  return clusterManagers
+      .map<Object>((ClusterManager cm) => serializeClusterManager(cm))
+      .toList();
+}
+
+/// Converts a Cluster Manager into object serializable in JSON.
+Object serializeClusterManager(ClusterManager clusterManager) {
```
camsim99, line 15: Nit: can this be private? Don't see it used anywhere else

camsim99 (approved):
Left some nits but all of my comments have been addressed! Looks good :)

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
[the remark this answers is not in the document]
jokerttu: @reidbaker @stuartmorgan do you still need changes to this?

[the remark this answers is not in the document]
stuartmorgan: I'm fine with this staying as is for now since a) it should never actually happen, and b) it will go away when we convert to Pigeon, which I plan to do this quarter.

camsim99 (approved):
Thanks for fixing my nits! @reidbaker for second review

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkerController.java
reidbaker: yes that comment works for me.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
reidbaker: Ok so what you are saying is that an invalid map id is a user error and is safe to call?

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
jokerttu: Yes it is. The app facing plugin implementation is also hiding the map id from the developer interface.
If anyone uses these plarform interface implementations directly, it is user error to listen wron mapId.
Note that this whole multiplexing behaviour will be removed when this plugin is transformed to use pigeon messaging.

PR #6186, opened 2024-02-23
[google_maps_flutter] Add marker clustering support - iOS implementation

jokerttu:
This PR introduces support for marker clustering for iOS platform

An example usabe is available in the example application at ./packages/google_maps_flutter/google_maps_flutter_ios/example/ios12 on the page `Manage clustering`

Containing only changes to `google_maps_flutter_ios` package.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.h
```diff
@@ -0,0 +1,55 @@
...
+/// @param identifier identifier of the ClusterManager.
+/// @return GMUClusterManager if found otherwise nil.
+- (GMUClusterManager *)clusterManagerWithIdentifier:(NSString *)identifier;
+
+/// Converts all clusters from the specific ClusterManager to result object response
+/// and updates that to the provided result object.
+///
+/// @param identifier identifier of the ClusterManager.
+/// @param result FlutterResult object to be updated with cluster data.
+- (void)clustersWithIdentifier:(NSString *)identifier result:(FlutterResult)result;
```
hellohuanlin, line 44: `cluster` or `clusters`?

hellohuanlin: the method name feels like it's returning a cluster. 

If i understand it right, how about something like `sendResultWithClusterIdentifier`?

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.h
```diff
@@ -0,0 +1,55 @@
...
+
+/// Called when cluster marker is tapped on the map.
+///
+/// @param cluster GMUStaticCluster object.
+- (void)didTapOnCluster:(GMUStaticCluster *)cluster;
+
+/// Calls cluster method of all ClusterManagers.
+- (void)clusterAll;
```
hellohuanlin, line 52: how about `invokeAllClusters`?

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.m
```diff
@@ -0,0 +1,144 @@
...
+// found in the LICENSE file.
+
+#import "FLTClusterManagersController.h"
+#import "FLTGoogleMapJSONConversions.h"
+#import "GoogleMarkerUtilities.h"
+
+@interface FLTClusterManagersController ()
+
+@property(strong, nonatomic)
+    NSMutableDictionary<NSString *, GMUClusterManager *> *clusterManagerIdToManagers;
```
hellohuanlin, line 12: `ID` rather than `Id`

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.m
```diff
@@ -0,0 +1,144 @@
...
+      continue;
+    }
+    [response addObject:clusterDict];
+  }
+  result(response);
+}
+
+- (void)didTapOnCluster:(GMUStaticCluster *)cluster {
+  NSDictionary *clusterDict = [self serializableDictionaryForCluster:cluster];
+  if (clusterDict != nil) {
```
hellohuanlin, line 94: nit: `if (clusterDict)`

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.m
```diff
@@ -0,0 +1,144 @@
...
+/// Returns the cluster manager id for given cluster.
+///
+/// @param cluster identifier of the ClusterManager.
+/// @return id NSString if found otherwise nil.
+- (NSString *)clusterManagerIdentifierForCluster:(GMUStaticCluster *)cluster {
+  if ([cluster.items count] == 0) {
+    return nil;
+  }
+
+  if (cluster.items.firstObject && [cluster.items.firstObject isKindOfClass:[GMSMarker class]]) {
```
hellohuanlin, line 110: `nil` will be just NO when checking `isKindOfClass`

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
```diff
@@ -547,28 +575,38 @@ - (void)mapView:(GMSMapView *)mapView idleAtCameraPosition:(GMSCameraPosition *)
 }
 
 - (BOOL)mapView:(GMSMapView *)mapView didTapMarker:(GMSMarker *)marker {
-  NSString *markerId = marker.userData[0];
-  return [self.markersController didTapMarkerWithIdentifier:markerId];
+  if ([marker.userData conformsToProtocol:@protocol(GMUCluster)]) {
+    GMUStaticCluster *cluster = marker.userData;
```
hellohuanlin, line 579: should it be just `GMUCluster` since that's what you checked to conform to?

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -232,52 +258,106 @@ @interface FLTMarkersController ()
...
+  FLTGoogleMapMarkerController *controller =
+      [[FLTGoogleMapMarkerController alloc] initWithMarker:marker
+                                          markerIdentifier:markerIdentifier
+                                  clusterManagerIdentifier:clusterManagerIdentifier
+                                                   mapView:self.mapView];
+  [controller interpretMarkerOptions:markerToAdd registrar:self.registrar];
+  if (clusterManagerIdentifier && [clusterManagerIdentifier isKindOfClass:[NSString class]]) {
+    id clusterManager =
+        [_clusterManagersController clusterManagerWithIdentifier:clusterManagerIdentifier];
+    if (marker && [clusterManager isKindOfClass:[GMUClusterManager class]]) {
```
hellohuanlin, line 306: do we need type check here? can the `_clusterManagersController` here directly return `GMUClusterManager` type?

hellohuanlin (commented):
The code looks quite clean. But I feel I don't know enough about clusters to stamp it just yet. Could you also invite the reviewers from your previous PR?

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.h
jokerttu: This method converts all clusters from the specific ClusterManager to serializable array. 
This method itself does not send anything, it just serializes the data to the given result object. 
Renamed the method to `serializeClustersWithIdentifier` to match better the implementation.

Also the method documentation updated to match better the implementation.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.h
jokerttu: This is used to trigger clustering algorithm on each cluster manager. When markers are added, modified or removed from clustermanager the clustering is not triggered for each operation, and clustering is called after all changed are done for better performance.
Changed the name to `invokeClusteringForAllClusterManagers` to be more descriptive.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.m
jokerttu: Renamed this to `clusterManagerIdentifierToManagers`

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
jokerttu: Fixed the check

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.m
jokerttu: removed the unnecessary check `cluster.items.firstObject && `

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.m
jokerttu: changed to null check

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.h
hellohuanlin: > This method itself does not send anything, it just serializes the data to the given result object.

Not sure if i follow. iirc `FlutterResult` is a closure that sends data to the dart side, is it not the case?

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec
```diff
@@ -22,8 +22,15 @@ Downloaded by pub (not CocoaPods).
   # has been confirmed to be compatible via an example in examples/. See discussion
   # in https://github.com/flutter/flutter/issues/86820 for why this is so broad.
   s.dependency 'GoogleMaps', '< 9.0'
+  # Once the plugin requires iOS version 13+, this should be updated.
+  s.dependency 'Google-Maps-iOS-Utils', '4.1.0'
   s.static_framework = true
   s.platform = :ios, '12.0'
+  # Enabling Swift support for Google-Maps-iOS-Utils
+  s.swift_version = '5.0'
```
hellohuanlin, line 30: why is swift required to use this library? this doesn't look like a swift only library.

hellohuanlin (commented):
can you add some unit tests on native side?

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec
jokerttu: Swift is required by the utils library (v4.1.0): https://github.com/googlemaps/google-maps-ios-utils/tree/v4.1.0:

Without following lines:
```
  s.swift_version = '5.0'
  s.xcconfig = {
    'LD_RUNPATH_SEARCH_PATHS' => '$(inherited) /usr/lib/swift',
  }
```

Following exception is thrown:
```
Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Termination Reason: DYLD 1 Library missing
Library not loaded: @rpath/libswiftCore.dylib
```

Seems like this works without the swift_version line `s.swift_version = '5.0'`
But i think it should be kept in as well to match the swift version on utils lib:
(https://github.com/googlemaps/google-maps-ios-utils/blob/v4.1.0/Google-Maps-iOS-Utils.podspec#L22)
@stuartmorgan any comments on this?

jokerttu:
@hellohuanlin 
> can you add some unit tests on native side?
iOS does not have a lot of tests currently to use as a base for map control tests.
PartiallyMockedMapView is really stub, and need to have at least mocked versions for adding and handling the marker items. clustering will use other map methods as well, atleast camera location and visible area need to mocked as well.
Also cluster focus click might call animated camera methods internally.
I'll look into these asap, but not sure when I am able to finalize these as these might take some time to develop.

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec
hellohuanlin: You may wanna put `5.3` rather than `5.0` from the link you posted: 

> Note: This feature is only available with Swift 5.3 (Xcode 12) or later.

jokerttu: It seems to be related to the Swift Package Manager

jokerttu:
> can you add some unit tests on native side?

Added some native unit tests

hellohuanlin (approved):
Thanks for adding those tests. The overall code looks clean. 

@cbracken feel free to give it another pass (if you see fit) since you are the new owner. Otherwise I think it's good to land.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.h
```diff
@@ -0,0 +1,60 @@
...
+///
+/// @param identifiers List of clusterManagerIds to remove.
+
+- (void)removeClusterManagersWithIdentifiers:(NSArray<NSString *> *)identifiers;
+
+/// Returns the ClusterManager for the given identifier.
+///
+/// @param identifier identifier of the ClusterManager.
+/// @return GMUClusterManager if found otherwise nil.
+- (GMUClusterManager *)clusterManagerWithIdentifier:(NSString *)identifier;
```
stuartmorgan, line 37: The API declaration asserts that the return value will never be null, but the comment just above says the opposite.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.m
```diff
@@ -0,0 +1,146 @@
...
+#import "FLTClusterManagersController.h"
+#import "FLTGoogleMapJSONConversions.h"
+#import "GoogleMarkerUtilities.h"
+
+@interface FLTClusterManagersController ()
+
+@property(strong, nonatomic)
+    NSMutableDictionary<NSString *, GMUClusterManager *> *clusterManagerIdentifierToManagers;
+@property(strong, nonatomic) FlutterMethodChannel *methodChannel;
+@property(strong, nonatomic) GMSMapView *mapView;
```
stuartmorgan, line 14: These all need declaration comments too; the style guide does not exempt private declarations from the commenting requirements.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.m
```diff
@@ -0,0 +1,146 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FLTClusterManagersController.h"
```
stuartmorgan, line 5: Nit: Blank line below this one per style guide convention.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMarkerUserData.h
```diff
@@ -0,0 +1,13 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Foundation/Foundation.h>
+
+// Defines user data object for markers.
+@interface GoogleMarkerUserData : NSObject
+
+@property(nonatomic, strong) NSString *markerIdentifier;
+@property(nonatomic, strong) NSString *clusterManagerIdentifier;
```
stuartmorgan, line 11: These need declaration comments.

stuartmorgan, line 11: These should both be `copy`.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMarkerUtilities.h
```diff
@@ -0,0 +1,38 @@
...
+/// @param markerIdentifier Identifier of marker.
+/// @param clusterManagerIdentifier Identifier of cluster manager.
+/// @param marker GMSMarker object.
++ (void)setMarkerIdentifier:(NSString *)markerIdentifier andClusterManagerIdentifier:(NSString *)clusterManagerIdentifier for:(GMSMarker*)marker;
+
+/// Get MarkerIdentifier from GMSMarker UserData.
+///
+/// @param marker GMSMarker object.
+/// @return NSString if found otherwise nil.
++ (NSString *)getMarkerIdentifierFrom:(GMSMarker *)marker;
```
stuartmorgan, line 28: Same comment as in the other file; this contradicts the comment directly above it.

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec
```diff
@@ -22,8 +22,15 @@ Downloaded by pub (not CocoaPods).
   # has been confirmed to be compatible via an example in examples/. See discussion
   # in https://github.com/flutter/flutter/issues/86820 for why this is so broad.
   s.dependency 'GoogleMaps', '< 9.0'
+  # Once the plugin requires iOS version 13+, this should be updated.
```
stuartmorgan, line 25: This day has come :) We now require iOS 14+.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
```diff
@@ -580,6 +614,37 @@ class GoogleMapsFlutterIOS extends GoogleMapsFlutterPlatform {
     GoogleMapsInspectorPlatform.instance =
         GoogleMapsInspectorIOS((int mapId) => _channel(mapId));
   }
+
+  /// Parses cluster data from dynamic json objects and returns [Cluster] object.
+  /// Used by `cluster#onTap` method call handler and inspectors [getClusters] response parser.
+  static Cluster parseCluster(
+      String clusterManagerIdString,
+      Object positionObject,
+      Map<dynamic, dynamic> boundsMap,
```
stuartmorgan, line 623: `dynamic` is very strongly discouraged in all new code; please use `Object` or `Object?` as appropriate.

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMarkerUtilities.h
```diff
@@ -0,0 +1,38 @@
...
+
+/// Get ClusterManagerIdentifier from GMSMarker UserData.
+///
+/// @param marker GMSMarker object.
+/// @return NSString if found otherwise nil.
++ (NSString *)getClusterManagerIdentifierFrom:(GMSMarker *)marker;
```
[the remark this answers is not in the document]
jokerttu: added `nullable` to the signature

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMarkerUtilities.h
jokerttu: added `nullable` to the signature

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMarkerUserData.h
jokerttu: fixed

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTClusterManagersController.h
jokerttu: added `nullable` to the signature

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec
jokerttu: Note that the clustering still works with the version 4.1.0 as well, so updating it might not be mandatory for now.

stuartmorgan: Sounds like this is blocked on https://github.com/googlemaps/google-maps-ios-utils/issues/456 then.

I would image that will be fixed soon, because Google Maps 7.3 isn't compliant with the upcoming App Store privacy requirements, so the current version of Google-Maps-iOS-Utils will become essentially unshippable once enforcement of those requirements starts.

jmagman:
@jokerttu is this ready for re-review?

jokerttu:
> @jokerttu is this ready for re-review?

@jmagman not yet, this is waiting new release of Google-Maps-iOS-Utils library to have support latest Google Maps SDK versions. See discussion here: https://github.com/flutter/packages/pull/6186#discussion_r1581115795

stuartmorgan:
FYI I have raised the blocking issue internally with the Google Maps team.

stuartmorgan:
Update from triage: I haven't heard anything yet on when the utils SDK will be usable for us.

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec
cedvdb: @stuartmorgan FYI that repository you linked does not seem to be actively maintained

stuartmorgan: @cedvdb See https://github.com/googlemaps/google-maps-ios-utils/issues/473

Rexios80:
It looks like the iOS utils pod might be updated enough to land this. Worth checking out: https://github.com/googlemaps/google-maps-ios-utils/issues/473#issuecomment-2261690953

jokerttu:
As version 5.0.0 and version 6.0.0 for Google-Maps-iOS-Utils has been published, I am able to continue with this.
Meanwhile the package has been migrated to pigeon, so this needs now a bit more changes than version number...

jokerttu:
@stuartmorgan this is again ready for review.

Updated version dependency for Google-Maps-iOS-Utils:
`s.dependency 'Google-Maps-iOS-Utils', '>= 5.0.0', '< 7.0.0'`
 ios14 uses version 5.0.0 for GoogleMaps Sdk version 8.x
 ios15 uses version 6.0.0 for GoogleMaps Sdk version 9.x

Migrated to pigeon messaging

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMarkerUtilities.h
```diff
@@ -0,0 +1,38 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <GoogleMaps/GoogleMaps.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface GoogleMarkerUtilities : NSObject
```
stuartmorgan, line 9: Please don't create a class just to hold static methods; if there's no state, and it's never initialized, then there's no meaningful class. C functions are a fully supported part of Objective-C, so unlike Java this kind of container is not necessary.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMarkerUserData.h
```diff
@@ -0,0 +1,17 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Foundation/Foundation.h>
+
+/// Defines user data object for markers.
+@interface GoogleMarkerUserData : NSObject
```
stuartmorgan, line 8: `FGMGoogleMarkerUserData`. All classes must be prefixed to avoid runtime collisions.

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec
```diff
@@ -25,8 +25,14 @@ Downloaded by pub (not CocoaPods).
   # Versions earlier than 8.4 can't be supported because that's the first version
   # that supports privacy manifests.
   s.dependency 'GoogleMaps', '>= 8.4', '< 10.0'
+  s.dependency 'Google-Maps-iOS-Utils', '>= 5.0.0', '< 7.0.0'
```
stuartmorgan, line 28: Please put a comment about the mapping from `GoogleMaps` versions to these versions, to explain why there's a range.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/utils/cluster_manager.dart
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/// Converts a Cluster Manager into object serializable in JSON.
+Object serializeClusterManager(ClusterManager clusterManager) {
+  final Map<String, Object> json = <String, Object>{};
+  json['clusterManagerId'] = clusterManager.clusterManagerId.value;
+  return json;
+}
+
+/// Converts a Cluster Manager updates into object serializable in JSON.
```
stuartmorgan, line 21: This function should no longer be necessary.

packages/google_maps_flutter/google_maps_flutter_ios/pigeons/messages.dart
```diff
@@ -334,6 +367,9 @@ abstract class MapsCallbackApi {
   @ObjCSelector('didTapCircleWithIdentifier:')
   void onCircleTap(String circleId);
 
+  /// Called when a marker cluster is tapped.
```
stuartmorgan, line 370: `@ObjCSelector('didTapCluster:')`

packages/google_maps_flutter/google_maps_flutter_ios/pigeons/messages.dart
```diff
@@ -364,4 +400,9 @@ abstract class MapsInspectorApi {
   PlatformTileLayer? getTileOverlayInfo(String tileOverlayId);
   @ObjCSelector('zoomRange')
   PlatformZoomRange getZoomRange();
+  // TODO(stuartmorgan): Make the generic type non-nullable once supported.
+  // https://github.com/flutter/flutter/issues/97848
+  // The consuming code treats the entries as non-nullable.
+  @ObjCSelector('getClustersWithIdentifier:')
```
stuartmorgan, line 406: `clustersWithIdentifier:`

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
```diff
@@ -171,9 +175,14 @@ - (instancetype)initWithMapView:(GMSMapView *_Nonnull)mapView
...
-    _markersController = [[FLTMarkersController alloc] initWithMapView:_mapView
-                                                       callbackHandler:_dartCallbackHandler
-                                                             registrar:registrar];
+    _clusterManagersController =
+        [[FLTClusterManagersController alloc] initWithCallbackHandler:_dartCallbackHandler
+                                                              mapView:_mapView];
+    _markersController =
+        [[FLTMarkersController alloc] initWithClusterManagersController:_clusterManagersController
+                                                        callbackHandler:_dartCallbackHandler
+                                                                mapView:_mapView
```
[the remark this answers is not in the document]
jokerttu: I standardized this as well to take the mapView first

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -454,53 +482,86 @@ - (instancetype)initWithMapView:(GMSMapView *)mapView
...
+                                  clusterManagerIdentifier:clusterManagerIdentifier
+                                                   mapView:self.mapView];
+  [controller interpretMarkerOptions:markerToAdd
+                           registrar:self.registrar
+                         screenScale:[self getScreenScale]];
+  if (clusterManagerIdentifier && [clusterManagerIdentifier isKindOfClass:[NSString class]]) {
+    GMUClusterManager *clusterManager =
+        [_clusterManagersController clusterManagerWithIdentifier:clusterManagerIdentifier];
+    if (marker && clusterManager) {
+      [clusterManager addItem:(id<GMUClusterItem>)marker];
```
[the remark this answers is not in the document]
jokerttu: Added conformsToProtocol check to make sure to work with possible future changes on GMUClusterItem protocol.

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec
jokerttu: I don't think this is related to GoogleMaps versions, more to supported iOS versions, 14 and 15.
Should we remove this range and let cocoapods to automatically resolve proper supported version or should we keep is as the dependency is tested only with versions 5 and 6.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMarkerUserData.h
jokerttu: Other files have FLT prefix, I used it, if this really need to use FGM I can rename them again

stuartmorgan: See https://github.com/flutter/flutter/issues/102601 for context on why many files currently use `FLT`.

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec
stuartmorgan: > Should we remove this range

We should not remove the range; see lines 21 and 22 of this file.

> and let cocoapods to automatically resolve proper supported version

Cocoapods has no way of knowing whether the code in this plugin will work with `Google-Maps-iOS-Utils` version 7, so cannot resolve the proper supported version if there is no upper bound.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMClusterManagersController.h
```diff
@@ -0,0 +1,70 @@
...
+
+#import <Flutter/Flutter.h>
+#import <GoogleMaps/GoogleMaps.h>
+@import GoogleMapsUtils;
+
+#import "messages.g.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+// Defines cluster managers controller interface which
```
stuartmorgan, line 13: Comments should be written for an audience that understand the basic mechanics of the language; a class comment does not need to say that `@interface` defines an interface.

Also, this should be a `///` comment.

```objc
/// A controller that manages all of the cluster managers on a map.
```

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMClusterManagersController.h
```diff
@@ -0,0 +1,70 @@
...
+
+/// Creates ClusterManagers and initializes them.
+///
+/// @param clusterManagersToAdd List of clustermanager object data.
+- (void)addClusterManagers:(NSArray<FGMPlatformClusterManager *> *)clusterManagersToAdd;
+
+/// Removes requested ClusterManagers from the controller.
+///
+/// @param identifiers List of clusterManagerIds to remove.
+
```
stuartmorgan, line 37: No blank lines between a declaration comment and the thing it is commenting.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMClusterManagersController.h
```diff
@@ -0,0 +1,70 @@
...
+/// @return GMUClusterManager if found; otherwise, nil.
+- (nullable GMUClusterManager *)clusterManagerWithIdentifier:(NSString *)identifier;
+
+/// Converts clusters managed by the specified ClusterManager to
+/// a serializable array of clusters.
+///
+/// This method fetches and serializes clusters at the current zoom
+/// level from the ClusterManager identified by the given identifier.
+/// If the specified ClusterManager identifier does not exist, an empty
+/// array is returned.
```
stuartmorgan, line 52: This does not appear to be true; this is the error condition unless I'm reading the code wrong.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMClusterManagersController.h
```diff
@@ -0,0 +1,70 @@
...
+///
+/// @param identifier The identifier of the ClusterManager to serialize.
+/// @return An array of FGMPlatformCluster objects representing the clusters. `nil` is returned only
+/// when `error != nil`.
```
stuartmorgan, line 56: Presumably you mean `*error != nil`; it's not valid to pass `nil` as the value of `error`. I would just say "[...] only if `error` is populated."

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMClusterManagersController.h
```diff
@@ -0,0 +1,70 @@
...
+- (nullable NSArray<FGMPlatformCluster *> *)
+    clustersWithIdentifier:(NSString *)identifier
+                     error:(FlutterError *_Nullable __autoreleasing *_Nonnull)error;
+
+/// Called when cluster marker is tapped on the map.
```
stuartmorgan, line 61: "when a cluster marker"

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMClusterManagersController.h
```diff
@@ -0,0 +1,70 @@
...
+///
+/// @param cluster GMUStaticCluster object.
```
stuartmorgan, line 63: This is a restatement of the parameter declaration in English, not an explanation. The conceptual description, which is what a declaration comment should contain, is "The cluster that was tapped on."

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMClusterManagersController.h
```diff
@@ -0,0 +1,70 @@
...
+/// Removes requested ClusterManagers from the controller.
+///
+/// @param identifiers List of clusterManagerIds to remove.
+
+- (void)removeClusterManagersWithIdentifiers:(NSArray<NSString *> *)identifiers;
+
+/// Returns the ClusterManager for the given identifier.
+///
+/// @param identifier identifier of the ClusterManager.
+/// @return GMUClusterManager if found; otherwise, nil.
```
stuartmorgan, line 43: The comment does not need to say what the return type is; the code already does that.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMClusterManagersController.h
```diff
@@ -0,0 +1,70 @@
...
+@interface FGMClusterManagersController : NSObject
+
+/// Initializes FLTClusterManagersController.
+///
+/// @param callbackHandler A callback handler.
+/// @param mapView A map view that will be used to display clustered markers.
+- (instancetype)initWithMapView:(GMSMapView *)mapView
+                callbackHandler:(FGMMapsCallbackApi *)callbackHandler;
+
+/// Creates ClusterManagers and initializes them form JSON data.
```
stuartmorgan, line 24: You've used `ClusterManager` throughout these comments, but that's not an actual type. In general, you should not use type names in declaration comments unless they resolve some ambiguity (e.g., in the conversion function comments, where the point is that it's two different concrete class representations of the same conceptual thing); you should use English. If you do need to use a type name to resolve ambiguity, it needs to be an exact type name.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMClusterManagersController.m
```diff
@@ -0,0 +1,137 @@
...
+  }
+}
+
+- (void)addClusterManager:(NSString *)identifier {
+  id<GMUClusterAlgorithm> algorithm = [[GMUNonHierarchicalDistanceBasedAlgorithm alloc] init];
+  id<GMUClusterIconGenerator> iconGenerator = [[GMUDefaultClusterIconGenerator alloc] init];
+  id<GMUClusterRenderer> renderer =
+      [[GMUDefaultClusterRenderer alloc] initWithMapView:self.mapView
+                                    clusterIconGenerator:iconGenerator];
+  GMUClusterManager *clusterManager = [[GMUClusterManager alloc] initWithMap:self.mapView
```
stuartmorgan, line 56: What's the purpose of this local variable, rather than assigning directly into the map?

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMMarkerUserData.h
```diff
@@ -0,0 +1,42 @@
...
+/// The identifier of the marker.
+@property(nonatomic, copy) NSString *markerIdentifier;
+
+/// The identifier of the cluster manager.
+/// This property is set only if the marker is managed by a cluster manager.
+@property(nonatomic, copy, nullable) NSString *clusterManagerIdentifier;
+
+@end
+
+/// Sets MarkerId and optionally ClusterManagerId to GMSMarker UserData.
```
stuartmorgan, line 21: This comment appears to reference two types that don't exist.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMMarkerUserData.h
```diff
@@ -0,0 +1,42 @@
...
+///
+/// @param markerIdentifier Identifier of marker.
+/// @param clusterManagerIdentifier Optional identifier of cluster manager.
+/// @param marker GMSMarker object.
```
stuartmorgan, line 25: This doesn't tell me anything about its use, only information that is obvious from the function signature to anyone who can read Objecitve-C.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMMarkerUserData.h
```diff
@@ -0,0 +1,42 @@
...
+extern void FGMSetIdentifiersToMarkerUserData(NSString *markerIdentifier,
+                                              NSString *_Nullable clusterManagerIdentifier,
+                                              GMSMarker *marker);
+
+/// Get MarkerIdentifier from GMSMarker UserData.
```
stuartmorgan, line 30: MarkerIdentifier is not a type, nor is UserData.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMMarkerUserData.h
```diff
@@ -0,0 +1,42 @@
...
+///
+/// @param marker GMSMarker object.
+/// @return NSString if found; otherwise, nil.
+extern NSString *_Nullable FGMGetMarkerIdentifierFromMarker(GMSMarker *marker);
+
+/// Get ClusterManagerIdentifier from GMSMarker UserData.
```
stuartmorgan, line 36: ClusterManagerIdentifier is not a type, nor is UserData.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.h
```diff
@@ -4,7 +4,7 @@
 
 #import <Flutter/Flutter.h>
 #import <GoogleMaps/GoogleMaps.h>
-
```
stuartmorgan, line 7: This blank line was [correct](https://google.github.io/styleguide/objcguide.html#order-of-includes).

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -3,26 +3,34 @@
 // found in the LICENSE file.
 
 #import "GoogleMapMarkerController.h"
+#import "FGMMarkerUserData.h"
```
stuartmorgan, line 6: There should be a blank line above this; it was missing before.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.h
```diff
@@ -4,7 +4,7 @@
...
+#import "FGMClusterManagersController.h"
 #import "GoogleMapController.h"
```
stuartmorgan, line 8: Although this should be in its own group above everything else.

stuartmorgan (changes requested):

jokerttu:
@stuartmorgan 
Fixed the issues mentioned in the review.
Rebased and fixed conflicts with heatmap support.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.h
stuartmorgan: Sorry, this comment was wrong; this is a header file not an implementation file. So these should be down with the `message.g.h` import.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
```diff
@@ -477,6 +500,7 @@ class GoogleMapsFlutterIOS extends GoogleMapsFlutterPlatform {
     Set<Polyline> polylines = const <Polyline>{},
     Set<Circle> circles = const <Circle>{},
     Set<TileOverlay> tileOverlays = const <TileOverlay>{},
+    Set<ClusterManager> clusterManagers = const <ClusterManager>{},
```
stuartmorgan, line 503: Oops, I missed this before; this change shouldn't be made. It compiles because it's an optional param, but nothing should ever pass this, so we don't need to add it.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
```diff
@@ -491,6 +515,7 @@ class GoogleMapsFlutterIOS extends GoogleMapsFlutterPlatform {
           polygons: polygons,
           polylines: polylines,
           circles: circles,
+          clusterManagers: clusterManagers,
```
stuartmorgan, line 518: And then this is removed as well.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
```diff
@@ -506,6 +531,7 @@ class GoogleMapsFlutterIOS extends GoogleMapsFlutterPlatform {
     Set<Polyline> polylines = const <Polyline>{},
     Set<Circle> circles = const <Circle>{},
     Set<TileOverlay> tileOverlays = const <TileOverlay>{},
+    Set<ClusterManager> clusterManagers = const <ClusterManager>{},
```
stuartmorgan, line 534: Same here and below.

stuartmorgan (changes requested):
Sorry, I missed a couple of things; once those are done this will be good to land.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
jokerttu: I'll remove these...
Noticed that this is for slipped through for android, and should be reverted.
https://github.com/flutter/packages/blob/4354f65e1314755f13d4d9db28b9fc08096b8f81/packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart#L601

stuartmorgan: I figured it probably did when I realized I'd missed it here; it's not a big deal but it wouldn't hurt to clean it up later.

stuartmorgan: Actually, scratch that; it would be a breaking change to the Android implementation package to remove it. That's not worth the churn.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
jokerttu: Fixed

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
jokerttu: removed

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
jokerttu: These lines are now removed

jokerttu:
@stuartmorgan seems like I need separate approval from you before I can add autosubmit label.

stuartmorgan (approved):
LGTM. Sorry, I should have pressed the other button last time to avoid that.

PR #6187, opened 2024-02-23
[google_maps_flutter] Add marker clustering support - web implementation

jokerttu:
This PR introduces support for marker clustering for Web platform

Containing only changes to `google_maps_flutter_web` package.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_web/README.md
+++ b/packages/google_maps_flutter/google_maps_flutter_web/README.md
@@ -52,14 +52,14 @@ Now you should be able to use the Google Maps plugin normally.
 
 ## Marker clustering
 
-If you need marker clustering support, modify the <head> tag to load the [js-markerclusterer](https://github.com/googlemaps/js-markerclusterer#install) library, like so:
+If you need marker clustering support, modify the <head> tag to load the [js-markerclusterer](https://github.com/googlemaps/js-markerclusterer#install) library. Ensure you are using the currently supported version `2.5.3`, like so:
 
 ```html
 <head>
 
   <!-- // Other stuff -->
 
-  <script src="https://unpkg.com/@googlemaps/markerclusterer@2.5.0/dist/index.min.js"></script>
+  <script src="https://unpkg.com/@googlemaps/markerclusterer@2.5.3/dist/index.min.js"></script>
 </head>
 ```
 
--- a/packages/google_maps_flutter/google_maps_flutter_web/README.md
+++ b/packages/google_maps_flutter/google_maps_flutter_web/README.md
@@ -50,6 +50,19 @@ To request multiple libraries, separate them with commas:
 
 Now you should be able to use the Google Maps plugin normally.
 
+## Marker clustering
+
+If you need marker clustering support, modify the <head> tag to load the [js-markerclusterer](https://github.com/googlemaps/js-markerclusterer#install) library, like so:
+
+```html
+<head>
+
+  <!-- // Other stuff -->
+
+  <script src="https://unpkg.com/@googlemaps/markerclusterer@2.5.0/dist/index.min.js"></script>
+</head>
+```
+
 ## Limitations of the web version
 
 The following map options are not available in web, because the map doesn't rotate there:
--- a/packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering.dart
@@ -74,7 +74,7 @@ class ClusterManagersController extends GeometryController {
         _clusterManagerIdToMarkerClusterer[clusterManagerId];
     if (markerClusterer != null) {
       markerClusterer.addMarker(marker, true);
-      markerClusterer.dirty = true;
+      markerClusterer.render();
     }
   }
 
@@ -86,23 +86,9 @@ class ClusterManagersController extends GeometryController {
           _clusterManagerIdToMarkerClusterer[clusterManagerId];
       if (markerClusterer != null) {
         markerClusterer.removeMarker(marker, true);
-        markerClusterer.dirty = true;
+        markerClusterer.render();
       }
     }
-  }
-
-  /// Renders all cluster managers that are marked as dirty.
-  ///
-  /// This mechanism is used to batch changes to the cluster managers to avoid
-  /// unnecessary re-renders.
-  /// [MarkerClusterer] is marked as dirty when a marker is added or removed.
-  void renderDirty() {
-    _clusterManagerIdToMarkerClusterer.values
-        .where((MarkerClusterer markerClusterer) => markerClusterer.dirty)
-        .forEach((MarkerClusterer markerClusterer) {
-      markerClusterer.dirty = false;
-      markerClusterer.render();
-    });
   }
 
   /// Returns list of clusters in [MarkerClusterer] with given
--- a/packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering.dart
@@ -1,0 +1,153 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:google_maps/google_maps.dart' as gmaps;
+import 'package:google_maps_flutter_platform_interface/google_maps_flutter_platform_interface.dart';
+
+import '../google_maps_flutter_web.dart';
+import 'marker_clustering_js_interop.dart';
+import 'types.dart';
+
+/// A controller class for managing marker clustering.
+///
+/// This class maps [ClusterManager] objects to javascript [MarkerClusterer]
+/// objects and provides an interface for adding and removing markers from
+/// clusters.
+class ClusterManagersController extends GeometryController {
+  /// Creates a new [ClusterManagersController] instance.
+  ///
+  /// The [stream] parameter is a required [StreamController] used for
+  /// emitting map events.
+  ClusterManagersController(
+      {required StreamController<MapEvent<Object?>> stream})
+      : _streamController = stream,
+        _clusterManagerIdToMarkerClusterer =
+            <ClusterManagerId, MarkerClusterer>{};
+
+  // The stream over which cluster managers broadcast their events
+  final StreamController<MapEvent<Object?>> _streamController;
+
+  // A cache of [MarkerClusterer]s indexed by their [ClusterManagerId].
+  final Map<ClusterManagerId, MarkerClusterer>
+      _clusterManagerIdToMarkerClusterer;
+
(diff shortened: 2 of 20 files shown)


jokerttu:
Web does not have example app to verify the clustering functionality.

Instead use app-facing package from this PR to test the clustering support on web platform: https://github.com/flutter/packages/pull/4319
It temporary uses dependency overriding to use this PR ref head as web package source.

To clone and start run the app-facing example app with web clustering from this PR,  follow these steps:
1. Clone and open git@github.com:flutter/packages.git
2. Fetch pull request ref: `git fetch origin pull/4319/head`
3. Create local branch from fetch head: `git checkout -b clustering-pr-4319 FETCH_HEAD`
4. Start example on using web platform at  `/packages/google_maps_flutter/google_maps_flutter/example`

This version of the app-facing plugin fetches the packages using PR refs for platform implementations:
```
dependency_overrides:
  google_maps_flutter_android:
    git:
      url: git@github.com:flutter/packages.git
      path: packages/google_maps_flutter/google_maps_flutter_android
      ref: refs/pull/6185/head
  google_maps_flutter_ios:
    git:
      url: git@github.com:flutter/packages.git
      path: packages/google_maps_flutter/google_maps_flutter_ios
      ref: refs/pull/6186/head
  google_maps_flutter_web:
    git:
      url: git@github.com:flutter/packages.git
      path: packages/google_maps_flutter/google_maps_flutter_web
      ref: refs/pull/6187/head
```

jokerttu:
@ditman is it possible to get review for this PR as iOS is already greenlighted and Android has only minor test related tasks to do.

ditman:
Seems to work on my machine!

Screenshot 2024-04-15 at 5 13 25 PM

packages/google_maps_flutter/google_maps_flutter_web/README.md
```diff
@@ -50,6 +50,19 @@ To request multiple libraries, separate them with commas:
 
 Now you should be able to use the Google Maps plugin normally.
 
+## Marker clustering
+
+If you need marker clustering support, modify the <head> tag to load the [js-markerclusterer](https://github.com/googlemaps/js-markerclusterer#install) library, like so:
+
```
ditman, line 56: We should specify the exact version(s) of the library that the js-interop code of the plugin supports.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering.dart
```diff
@@ -0,0 +1,153 @@
...
+      }
+    }
+  }
+
+  /// Renders all cluster managers that are marked as dirty.
+  ///
+  /// This mechanism is used to batch changes to the cluster managers to avoid
+  /// unnecessary re-renders.
+  /// [MarkerClusterer] is marked as dirty when a marker is added or removed.
+  void renderDirty() {
```
ditman, line 99: When manually testing, it seemed that some clusters disappeared, or didn't render at all after modifying them (with the example app), until I changed the zoom level. Is this function too strict? Should it re-render more?

packages/google_maps_flutter/google_maps_flutter_web/pubspec.yaml
```diff
@@ -23,7 +23,8 @@ dependencies:
   flutter_web_plugins:
     sdk: flutter
   google_maps: ^7.1.0
-  google_maps_flutter_platform_interface: ^2.5.0
+  google_maps_flutter_platform_interface: ^2.6.0
+  js: ^0.6.7
```
ditman, line 27: On version 0.5.5 we removed the `package:js` dependency so this package could compile to Wasm, please don't add it back! There's a new (better) way of doing js-interop.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering_js_interop.dart
```diff
@@ -0,0 +1,116 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:google_maps/google_maps.dart' as gmaps;
+import 'package:js/js.dart';
```
ditman, line 6: This needs to be done with `package:web` and `dart:js_interop` instead. See docs:

* https://dart.dev/interop/js-interop

Some examples (take a look at the whole files, there's examples on how to extend an existing package:web type, and how to create a custom object with an external factory):

* https://github.com/flutter/packages/blob/9a2490419268365911191f9d8388f986111bcc24/packages/video_player/video_player_web/example/integration_test/pkg_web_tweaks.dart#L41
* https://github.com/flutter/packages/blob/9a2490419268365911191f9d8388f986111bcc24/packages/google_maps_flutter/google_maps_flutter_web/lib/src/map_styler.dart#L17

ditman (changes requested):
The marker clustering is super intrusive with the markers themselves, but I guess that's the way it was done in the SDK.

Most of my comments are nitpicking, but there's one blocking observation: **we can't add `package:js` to the dependencies**. I hope the links I left help migrating the js-interop layer of markerclusterer to something that is wasm-friendly.

Thanks for the PR and the patience!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering_js_interop.dart
jokerttu: @ditman thanks for the review.
I once tried using dart:js_interop instead of package:js but got trouble with [gmaps.Marker references on bindings](https://github.com/a14n/dart-google-maps/blob/master/lib/src/generated/google_maps_core.js.g.dart#L1611)

[google_maps](https://pub.dev/packages/google_maps) package, that still uses js-package (not js_interop). 
See: https://github.com/a14n/dart-google-maps/blob/master/pubspec.yaml#L10
And for example https://github.com/a14n/dart-google-maps/blob/master/lib/src/generated/google_maps_core.js.g.dart#L1611 for marker bindings.

I am not sure is it possible to use js bindings from the google_maps package inside the js_interop extensions.
I can try again, but I don't think it is possible. So I think the  [google_maps](https://pub.dev/packages/google_maps) package need to be updated to use js_interop before this clustering support can land for the web platform.

I'll try again in separate branch and update the state.

ditman: @jokerttu I think the current definition should allow extending the gmaps.Marker. If you prepare a branch but you encounter errors, do cc me so I can try and help!

The goal is to **not use `package:js` in this package anymore.** Note that the JS annotation is now also available from `dart:js_interop` ([doc](https://api.flutter.dev/flutter/dart-js_interop/JS-class.html)).

jokerttu: @ditman here is the commit to separate branch:
https://github.com/CodemateLtd/packages/commit/3770ee2fcd3574b3320c16193e1bf835c0685b84

I was not yet able to solve the issue with the google_maps library:
example:
```
../../google_maps_flutter_web/lib/src/marker_clustering_js_interop.dart:22:28: Error: Type 'GMap?' is not a valid type in the signature of `dart:js_interop` external APIs or APIs converted via `toJS`. The only valid types are: JS types from `dart:js_interop`, @staticInterop types, void, bool, num, double, int, String, extension types that erases to one of these types, or a type parameter that is bound to a static interop type.
 - 'GMap' is from 'package:google_maps/src/generated/google_maps_core.js.g.dart' ('../../../../../../../.pub-cache/hosted/pub.dev/google_maps-7.1.0/lib/src/generated/google_maps_core.js.g.dart').
Use one of the valid types instead.
  external gmaps.GMap? get map;
```

ditman: (I'm looking at this right now, will report back!)

ditman:
@jokerttu I've created a fix for the JS-interop layer and left it on your repo:

* https://github.com/CodemateLtd/packages/pull/6

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering_js_interop.dart
ditman: Fixed, link to PR below!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering.dart
ditman: This doesn't seem to happen in "--profile" mode :/

jokerttu: I removed the whole dirty handling, as the clustering seems to work really fast even without it and it seems like an unnecessary optimization at the moment.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering_js_interop.dart
jokerttu: @ditman thanks, looks great. Merged and this PR is updated.
After reading your PR it is really clear how the type conversion should be done with external JS types;
Even when this is actually explained in https://dart.dev/interop/js-interop/usage and https://dart.dev/interop/js-interop/js-types it would be good to have better example, at the type conversion chapter, of the case where external javascript types are used as a function parameters, as in this case.
Thanks again!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering_js_interop.dart
```diff
@@ -0,0 +1,164 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// TODO(srujzs): Needed for https://github.com/dart-lang/sdk/issues/54801. Once
+// we publish a version with a min SDK constraint that contains this fix,
+// remove.
+@JS()
+library;
```
jokerttu, line 9: @ditman @srujzs 
Should I remove these lines and update the SDK constraints to 3.3.1 within this PR, or should this be updated separately?
See: https://github.com/CodemateLtd/packages/pull/6#discussion_r1581648828

ditman: I think we can clean them up all together later.

ditman:
Giving this a final test, want to see if the plugin still works without adding the clustering JS dependency (obviously NOT on the clustering screens, but elsewhere :))

ditman (approved):
I still believe marker_clustering_js_interop.dart should be a separate package, and that it should inject the JS as needed, but for now this is good.

I tested without the JS in the page, and everything works until you attempt to "Add cluster manager", and then it fails with a fairly explicit `Uncaught TypeError: Cannot read properties of undefined (reading 'MarkerClusterer')`, which is fine.

Let's ship this one!

PR #4064, opened 2023-05-23
[google_maps] Prepares packages to endorse web.

ditman:
Prepares the google_maps packages to endorse the Web platform:

* `[google_maps_flutter]` Changes `example/integration_test` to run in the web.
  * Splits the only file there in 3 for them to be slightly more manageable.
  * (Does not require publishing)
* `[google_maps_flutter_platform_interface]` Adds a test coming from the core plugin to the unit tests of this package.
  * (Does not require publishing)
* `[google_maps_flutter_web]` Changes to add an "inspector" object, and to conform with the tests in the core package.
  * Implements a `GoogleMapsInspectorPlatform` to allow integration tests to inspect parts of the internal state of a gmap.
  * Fires a `MapStyleException` when an invalid JSON is used in `setMapStyle` (was `FormatException` previously), to conform with the expected behavior in the core plugin tests.
  * (Requires publishing)

## Issues

* Part of: https://github.com/flutter/flutter/issues/80688

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[...]

PR #6192, opened 2024-02-23
[google_maps_flutter] Add `style` to widget

stuartmorgan:
Adds `style` to the `GoogleMap` widget construction, and internally to the `MapConfiguration` structure, wiring updates to style through the existing configuration update system instead of needing to be set via a separate channel. This makes things more consistent (style is conceptually similar to many of the other configuration options, and is not something that's likely to require frequent dynamic updates), and more importantly allows setting the style as part of creation instead of asynchronously after the map has likely already been displayed.

Because of https://github.com/flutter/flutter/issues/24327, this adds a way to query for style errors after the fact. (It's possible that it's why `setMapStyle` was done that way in the first place; the initial PR doesn't explain.) While this is slightly clunky, it's unlikely that this method would actually be used in production code anyway, since presumably people are defining their styles at build time, not runtime, and thus probably don't need runtime error handling. Long term, the solution to this problem is likely to invert the widget/controller relationship, as was done in webview_flutter, but that's a major breaking change and not something we're ready to take on at this point.

Since we don't really want two recommended ways of doing the same thing, this deprecates the existing `setMapStyle` method in favor of the new approach.

Fixes https://github.com/flutter/flutter/issues/66207

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
```diff
@@ -506,7 +513,7 @@ - (void)setMyLocationButtonEnabled:(BOOL)enabled {
 }
 
 - (NSString *)setMapStyle:(NSString *)mapStyle {
-  if (mapStyle == (id)[NSNull null] || mapStyle.length == 0) {
+  if (mapStyle.length == 0) {
     self.mapView.mapStyle = nil;
```
hellohuanlin, line 517: should we clear the error here?

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
```diff
@@ -658,6 +665,10 @@ - (void)interpretMapOptions:(NSDictionary *)data {
   if (myLocationButtonEnabled && myLocationButtonEnabled != (id)[NSNull null]) {
     [self setMyLocationButtonEnabled:[myLocationButtonEnabled boolValue]];
   }
+  NSString *style = data[@"style"];
+  if (style) {
```
hellohuanlin, line 669: will this be NSNull?

hellohuanlin (commented):
only reviewed the ios part

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
stuartmorgan: No, the corresponding JSON serialization code on the Dart side is:

```dart
if (config.style != null) 'style': config.style!,
```

Now that we are using in-package method channels everywhere, meaning that method channels in our plugins now map directly from Dart to a single implementation, I don't think there's much value in doing defensive coding against potential changes in Dart, because the only reason someone would change that Dart code would be to change this one other endpoint at the same time (unlike the old days, when someone might change the Dart code to do something for Android, not realizing that it would break iOS because they didn't even look).

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
stuartmorgan: The caller is responsible for using the return value to update the error state.

stuartmorgan:
@ditman Any idea what those web failures mean? It looks like the driver itself failed rather than a specific test, but it happened to both `master` and `stable` which suggest that it's my PR.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
hellohuanlin: if the caller successfully set the style (to be nil), shouldn't we mark our internal state as no error? (The error ivar is used to track the error happened in previous setStyle call right?)

stuartmorgan: I don't understand what code specifically you are referring to as "we" and "our" in that statement, so I don't know how to answer.

> (The error ivar is used to track the error happened in previous setStyle call right?)

The same code that set it previously would clear it. I'm not sure what the call sequence you are concerned about here is exactly.

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -136,6 +137,19 @@ class GoogleMap extends StatefulWidget {
...
+  /// Set to null to clear any previous custom styling.
+  ///
+  /// If problems were detected with the [mapStyle], including un-parsable
+  /// styling JSON, unrecognized feature type, unrecognized element type, or
+  /// invalid styler keys, the style is left unchanged, and the error can be
+  /// retrieved with [GoogleMapController.getStyleError].
+  ///
+  /// The style string can be generated using the
+  /// [map style tool](https://mapstyle.withgoogle.com/).
+  final String? style;
```
reidbaker, line 151: Is there any way at all to make this a type safe object?
 
 If not can we at least point users to https://developers.google.com/maps/documentation/cloud-customization which does not require app updates and a stringly typed api.

packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
```diff
@@ -190,11 +190,17 @@ class GoogleMapController {
   /// Also, refer [iOS](https://developers.google.com/maps/documentation/ios-sdk/style-reference)
   /// and [Android](https://developers.google.com/maps/documentation/android-sdk/style-reference)
   /// style reference for more information regarding the supported styles.
+  @Deprecated('Use GoogleMap.style instead.')
   Future<void> setMapStyle(String? mapStyle) {
     return GoogleMapsFlutterPlatform.instance
         .setMapStyle(mapStyle, mapId: mapId);
   }
 
+  /// Returns the last style error, if any.
```
reidbaker, line 199: Should this say that it will get errors for the current mapid only?

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
stuartmorgan: MapController is a per-map object; it's created with this ID.

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
stuartmorgan: > Is there any way at all to make this a type safe object?

Not in any useful way. The native objects are opaque objects created from a string that is supposed to be a JSON string. So the options are:
- Make a do-nothing `MapStyle` object that just wraps a String, but in practice that would almost certainly just convert every call from `style: someString` to `style: MapStyle(someString)`
- Pre-parse the JSON on the Dart side, only to convert it back to a string and then have the Maps SDK re-parse it anyway.

> If not can we at least point users to https://developers.google.com/maps/documentation/cloud-customization which does not require app updates and a stringly typed api.

If you follow the link there in this comment for generating the JSON style, you'll see that we don't really need to do that at the plugin level, as the SDK's docs and tool are very aggressive about this. (Also, there are significant caveats, like cloud style requires a per-load charge, and doesn't work on all Android devices, which is more than I want to get into in the comments here. I'd rather err on the side of being a neutral wrapper of the SDKs, and let the SDK docs steer people.)

ditman (approved):
I like this change! Maybe consider calling the new property `legacyStyle` to follow the naming used in the maps sdk documentation? https://developers.google.com/maps/documentation/javascript/cloud-customization/legacy-overview

(I don't know if this is referred to as "legacy" in other platforms, though)

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
stuartmorgan: > It can definitely be made a type-safe object

Sorry, technically there is a third option: we could—in our wrapper that is already under-resourced—reverse-engineer a complex object graph from docs, thus building and maintaining a much more complex API than what the SDKs provide, creating a ton of work for us, only to throw away all that structure and re-encode it as a string anyway at the SDK boundary. And thereby make the existing tools for these styles harder for our clients to use. And also being inconsistent with our `setMapStyle`, making it significantly harder to migrate to this new API.

It's theoretically possible, but I don't see any actual upsides.

> HOWEVER, one thing that Google Maps SDK seems to be doing is to call this feature "legacy styling", so maybe this property could be called: `legacyStyle`

Their high-level docs call it legacy style. The actual SDKs, and the code-level API docs for those SDKs, do not use that language anywhere that I have seen.

I do not want to be in the business of being significantly more opinionated in the code about what APIs people should or should not use than the people actually maintaining the SDK.

I also don't want `setMapStyle` (the existing method) and `legacyStyle` (the new, preferred approach), since that will make the deprecated one seem preferable and almost certainly confuse everyone.

stuartmorgan:
I'll split out the first sub-PR. Meanwhile, @ditman any thoughts on my question above about the web integration test failure?

ditman:
> @ditman any thoughts on my question above about the web integration test failure?

Nope, let me clone your branch and attempt to step through the test, (maybe an unexpected `null` somewhere?)

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
ditman: SGTM, `legacyStyle` seems to be only mentioned in the Cloud Styling docs.

hellohuanlin (approved):

stuartmorgan:
It looks like it doesn't like my new test. Maybe I've created a situation where it's throwing out of the constructor and `onMapCreated` never gets called. If so, I must be missing a test case in the web implementation package.

I missed that the failure was google_maps_flutter rather than google_maps_flutter_web, so I thought I couldn't repro locally. I should be able to figure it out if it's locally reproducible.

ditman:
I've pushed a small fix to the test, it just needed to use the `pumpMap` helper.

I introduced that helper when we endorsed the package (and integration_tests started to attempt to run on the web). The web platform needs a little bit more "scaffolding" (literally) on the app than mobile to properly render a functional Map widget.

Here's the PR where `pumpMap` was originally added:

* https://github.com/flutter/packages/pull/4064

stuartmorgan:
Awesome, thanks! I think I wrote the test in Android or iOS and then duplicated it to there, not realizing it used a different structure.

stuartmorgan:
Oh, it looks like this is related to https://github.com/flutter/packages/pull/5254 rather than my PR, given discussion like https://github.com/dart-lang/sdk/issues/55039

stuartmorgan:
Hopefully this will go away once https://github.com/flutter/packages/pull/6216 lands and publishes.

stuartmorgan:
That did indeed fix it. This is now just the app-facing bits, with the update constraints, so landing.

PR #6195, opened 2024-02-23
[tool] Ignore GeneratedPluginRegistrant.swift for `format`

stuartmorgan:
This file fails `swift-format lint`, so including it in `format` causes lots of warnings when run locally.


tarrinneal (approved):

tarrinneal:
This code clearly breaks webview, I can't believe you even uploaded it smh my head

PR #6201, opened 2024-02-25
[image_picker] Add limit parameter to pickMultiImage and pickMultipleMedia to ios and Android

pdenert:
Adds limit parameter to `MediaOptions` and `MultiImagePickerOptions` and supports its use on iOS and Android. The `limit` argument defines how many images or media files can be select.

Fixes: [flutter/flutter#85772](https://github.com/flutter/flutter/issues/85772)

## Pre-launch Checklist

- [X] I signed the [CLA].


packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
```diff
@@ -203,12 +203,15 @@ abstract class ImagePickerPlatform extends PlatformInterface {
   /// image types such as JPEG. If compression is not supported for the image that is picked,
   /// a warning message will be logged.
   ///
+  /// The `limit` argument defines how many images can be select.
+  ///
   /// If no images were picked, the return value is null.
   @Deprecated('Use getMultiImageWithOptions instead.')
```
tarrinneal, line 209: This method is deprecated and shouldn't be updated for this change.

The actual implementation of this deprecation is backwards. This method should be calling the newer method rather than the reverse.

Updating that would make this change to the arguments unneeded.

tarrinneal: Changing this would also make the changes to non ios/android sub-packages unneeded and revertable.

tarrinneal: as well as clean up some unneeded test changes.

tarrinneal (changes requested):
Seems like a solid change once the method deprecation is changed. Thanks for putting this together, and let me know if you need any help/guidance with anything!

packages/image_picker/image_picker/lib/image_picker.dart
```diff
@@ -128,6 +128,7 @@ class ImagePicker {
     double? maxWidth,
     double? maxHeight,
     int? imageQuality,
+    int? limit,
```
ditman, line 131: The new `limit` parameter should be documented in the doc-block above, since this function is directly used by end users of the plugin, and they might want to [know what it does](https://pub.dev/documentation/image_picker/1.0.7/image_picker/ImagePicker/pickMultiImage.html). 

(This is also a good place to remark that on the web this parameter is ignored!)

packages/image_picker/image_picker/lib/image_picker.dart
```diff
@@ -128,6 +128,7 @@ class ImagePicker {
...
     bool requestFullMetadata = true,
   }) {
     final ImageOptions imageOptions = ImageOptions.createAndValidate(
```
ditman, line 134: Should this assert that `limit == null || limit > 0`?

packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
ditman: I think this is a breaking change, that's why there's the "WithOptions" version, so we can add a new parameter to the constructor of the object (but not to a method that is going to be overridden)

packages/image_picker/image_picker_platform_interface/lib/src/types/media_options.dart
```diff
@@ -13,11 +13,15 @@ class MediaOptions {
   const MediaOptions({
     this.imageOptions = const ImageOptions(),
     required this.allowMultiple,
+    this.limit,
```
ditman, line 16: What happens when this object is created with `allowMultiple` as `false`, but `limit` is not `null`?

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
```diff
@@ -40,11 +40,13 @@ class MethodChannelImagePicker extends ImagePickerPlatform {
     double? maxWidth,
     double? maxHeight,
     int? imageQuality,
+    int? limit,
```
ditman, line 43: Please, do not add functionality to this method, if you can avoid it. Everything that returns `PickedFile` has been removed from public access of the plugin since v1.0.0 (see [changelog](https://pub.dev/packages/image_picker/changelog#100))

packages/image_picker/image_picker_for_web/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 3.1.0
 
+* Add limit parameter to `pickMultiImage` and `pickMultipleMedia` to ios and Android.
```
ditman, line 3: The web package shouldn't have any changes, since the `limit` parameter should only be added to the `MultiImagePickerOptions`, right?

ditman (commented):
I think that this PR could be much simplified if it were: "add a new parameter to `MultiImagePickerOptions`, and then use it in iOS and Android". For example, the web package wouldn't need any changes, because it doesn't care about the new parameter of that object at all.

There's a lot of extra code because this is updating a bunch of deprecated methods (as @tarrinneal mentioned in his review above!), which... we might be even able to delete...

packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
ditman: @tarrinneal I've created this issue:

* https://github.com/flutter/flutter/issues/144286

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java
```diff
@@ -297,8 +299,19 @@ private void launchPickMediaFromGalleryIntent(Messages.GeneralOptions generalOpt
...
+            effectiveLimit = Math.toIntExact(limit);
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R && SdkExtensions.getExtensionVersion(Build.VERSION_CODES.R) >= 2) {
+            effectiveLimit = MediaStore.getPickImagesMaxLimit();
+        } else {
+            effectiveLimit = Integer.MAX_VALUE;
+        }
+
         pickMediaIntent =
-            new ActivityResultContracts.PickMultipleVisualMedia()
+            new ActivityResultContracts.PickMultipleVisualMedia(effectiveLimit)
```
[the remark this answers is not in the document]
pdenert: No, than it's exactly the same. Calling it with no argument makes it assign to `maxItems` result of `getMaxItems`:

```java
PickMultipleVisualMedia(
    private val maxItems: Int = getMaxItems()
)
```

```java
internal fun getMaxItems() = if (PickVisualMedia.isSystemPickerAvailable()) {
    MediaStore.getPickImagesMaxLimit()
} else {
    Integer.MAX_VALUE
}
```

packages/image_picker/image_picker_for_web/CHANGELOG.md
pdenert: Right, I removed changes from web, linux, macos and windows packages. I only left dependency_overrides, to verify if changes to other packages would not mess things up.

packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart
pdenert: I have removed the limit from deprecated methods

packages/image_picker/image_picker_platform_interface/lib/src/types/media_options.dart
pdenert: It's firstly verify if the allowMultiple is true and only then query for photos with specified limit

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
pdenert: Removed

pdenert:
I've cleaned up packages which remain unchanged. 
I've removed limit from deprecated methods.
I've updated iOS and Android implementations

ditman:
Thanks for reverting the `web` package @pdenert!

I checked the web APIs again and there's nothing new to `limit` the number of files that the user selects (as in: disable file selection once the user selects `limit` files).

We could *count* how many they've selected and throw an exception, or discard extras if they've picked too many, but that's probably confusing, and can be implemented on the application layer too.

pdenert:
> We could count how many they've selected and throw an exception, or discard extras if they've picked too many, but that's probably confusing, and can be implemented on the application layer too.

Yeah, I feel the same way. If we can't limit choice on the UI side, then it's better to do nothing.

packages/image_picker/image_picker_platform_interface/lib/src/types/media_options.dart
pdenert: So if the `allowMultiple` is `false`, then it allow to select only one file

packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerPlugin.java
```diff
@@ -317,7 +320,18 @@ public void pickImages(
...
+
+      if (limit != null) {
+          effectiveLimit = Math.toIntExact(limit);
+      } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R && SdkExtensions.getExtensionVersion(Build.VERSION_CODES.R) >= 2) {
+          effectiveLimit = MediaStore.getPickImagesMaxLimit();
+      } else {
+          effectiveLimit = Integer.MAX_VALUE;
+      }
+
+      delegate.chooseMultiImageFromGallery(options, generalOptions.getUsePhotoPicker(), effectiveLimit, result);
```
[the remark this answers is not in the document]
pdenert: Works same as: https://github.com/flutter/packages/pull/6201#discussion_r1506770328

pdenert:
@tarrinneal do you mind look at this again?

vashworth (changes requested):
The iOS side looks good, however, the usage of `pickMultiImageWithMaxSize` and `makeWithMaxSize` within `image_picker_ios/example/ios/RunnerTests/ImagePickerPluginTests.m` will need to be updated to use new parameter. 

Also, additional tests should be added to test the new `limit` parameter correctly sets the `maxImageCount`. I've included an example you can use below. You should also include ones where the limit is nil.

<details><summary>Test example</summary>

```objc
- (void)testPickMultiImageWithLimit {
  FLTImagePickerPlugin *plugin = [[FLTImagePickerPlugin alloc] init];
  [plugin pickMultiImageWithMaxSize:[[FLTMaxSize alloc] init]
                            quality:nil
                       fullMetadata:NO
                              limit:@(2)
                         completion:^(NSArray<NSString *> *_Nullable result,
                                      FlutterError *_Nullable error){
                         }];
  XCTAssertEqual(plugin.callContext.maxImageCount, 2);
}

- (void)testPickMediaWithLimitAllowsMultiple {
  FLTImagePickerPlugin *plugin = [[FLTImagePickerPlugin alloc] init];
  FLTMediaSelectionOptions *mediaSelectionOptions =
      [FLTMediaSelectionOptions makeWithMaxSize:[FLTMaxSize makeWithWidth:@(100) height:@(200)]
                                   imageQuality:nil
                            requestFullMetadata:NO
                                  allowMultiple:YES
                                          limit:@(2)];

  [plugin pickMediaWithMediaSelectionOptions:mediaSelectionOptions
                                  completion:^(NSArray<NSString *> *_Nullable result,
                                               FlutterError *_Nullable error){
                                  }];

  XCTAssertEqual(plugin.callContext.maxImageCount, 2);
}

- (void)testPickMediaWithLimitMultipleNotAllowed {
  FLTImagePickerPlugin *plugin = [[FLTImagePickerPlugin alloc] init];
  FLTMediaSelectionOptions *mediaSelectionOptions =
      [FLTMediaSelectionOptions makeWithMaxSize:[FLTMaxSize makeWithWidth:@(100) height:@(200)]
                                   imageQuality:nil
                            requestFullMetadata:NO
                                  allowMultiple:NO
                                          limit:@(2)];

  [plugin pickMediaWithMediaSelectionOptions:mediaSelectionOptions
                                  completion:^(NSArray<NSString *> *_Nullable result,
                                               FlutterError *_Nullable error){
                                  }];

  XCTAssertEqual(plugin.callContext.maxImageCount, 1);
}

```

</details>

pdenert:
@vashworth thanks for the test examples. I've also added tests for variants where the limit is nil

vashworth (commented):
The test [Linux repo_checks](https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20repo_checks/6301/overview) is failing on formatting issues:
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8754354673776286193/+/u/Run_package_tests/format/stdout

Looks like a number of files need to be formatted: https://github.com/flutter/packages/blob/main/script/tool/README.md#format-code

packages/image_picker/image_picker_platform_interface/lib/src/types/multi_image_picker_options.dart
```diff
@@ -6,11 +6,31 @@ import 'image_options.dart';
...
 class MultiImagePickerOptions {
-  /// Creates an instance with the given [imageOptions].
+  /// Creates an instance with the given [imageOptions] and [limit].
   const MultiImagePickerOptions({
     this.imageOptions = const ImageOptions(),
+    this.limit,
   });
 
+  /// Creates an instance with the given [imageOptions] and [limit].
+  /// Throws if limit is not valid.
```
tarrinneal, line 16: same multi line comment issue

tarrinneal: Might as well be more specific about what a valid limit would be.

packages/image_picker/image_picker_platform_interface/lib/src/types/multi_image_picker_options.dart
```diff
@@ -6,11 +6,31 @@ import 'image_options.dart';
...
+  }
+
   /// The image-specific options for picking.
   final ImageOptions imageOptions;
+
+  /// The maximum number of images to select. Default null value means no limit.
+  final int? limit;
+
+  /// Validates that all values are within required ranges. Throws if not.
+  static void _validateOptions({int? limit}) {
```
tarrinneal, line 31: Same method name comment as before.

packages/image_picker/image_picker_platform_interface/lib/src/types/multi_image_picker_options.dart
```diff
@@ -6,11 +6,31 @@ import 'image_options.dart';
...
+    _validateOptions(limit: limit);
+  }
+
   /// The image-specific options for picking.
   final ImageOptions imageOptions;
+
+  /// The maximum number of images to select. Default null value means no limit.
+  final int? limit;
+
+  /// Validates that all values are within required ranges. Throws if not.
```
tarrinneal, line 30: Both of these comments need to be broken up into multi line comments by sentence. (check the style guide I linked earlier)

packages/image_picker/image_picker_platform_interface/lib/src/types/media_options.dart
tarrinneal: Probably worth throwing a test together to validate this so it doesn't change in the future.

tarrinneal (changes requested):
Quick review of the platform interface. I'll try to get a full review in asap for you.

packages/image_picker/image_picker_platform_interface/lib/src/method_channel/method_channel_image_picker.dart
```diff
@@ -72,13 +73,18 @@ class MethodChannelImagePicker extends ImagePickerPlatform {
       throw ArgumentError.value(maxHeight, 'maxHeight', 'cannot be negative');
     }
 
+    if (limit != null && limit < 2) {
+      throw ArgumentError.value(limit, 'limit', 'cannot be lower than 2');
+    }
```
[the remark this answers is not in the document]
pdenert: This is an Android requirement, iOS doesn't have this kind of limitation. Although there's a little difference between calling it with 1 and 2, because if you call iOS picker with limit equal or more than 2, then you have to confirm your selection with the 'add' button. And if you call iOS picker with limit = 1, then picker takes the first selected photo without confirmation. 
I thought it would be better to set such a requirement everywhere, rather than doing a different behaviour for Android.

```
@RequiresApi(19)
open class PickMultipleVisualMedia(
    private val maxItems: Int = getMaxItems()
) : ActivityResultContract<PickVisualMediaRequest, List<@JvmSuppressWildcards Uri>>() {

    init {
        require(maxItems > 1) {
            "Max items must be higher than 1"
        }
    }
```

packages/image_picker/image_picker_platform_interface/lib/src/types/media_options.dart
```diff
@@ -13,11 +13,32 @@ class MediaOptions {
...
 
   /// Whether to allow for selecting multiple media.
   final bool allowMultiple;
+
+  /// The maximum number of images to select. Default null value means no limit.
+  final int? limit;
+
+  /// Validates that all values are within required ranges. Throws if not.
+  static void _validateOptions({int? limit}) {
+    if (limit != null && limit < 2) {
```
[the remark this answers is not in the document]
pdenert: Answer in: https://github.com/flutter/packages/pull/6201#discussion_r1512744350

vashworth (approved):
Thanks for the contribution @pdenert! Great to see this added.

iOS side LGTM. I'll leave the interface and android side to @tarrinneal and @gmackall

packages/image_picker/image_picker_platform_interface/lib/src/types/media_options.dart
pdenert: Yeah, but I'm also wondering if it's not to much open approach. Maybe be more strict would be less confusing to users? So to require that `allowMultiple` should be `true` when limit is not `null` and also remove `MediaOptions.createAndValidate` and call `_validate` in default constructor?

It will look like that:

```dart
MediaOptions({
  this.imageOptions = const ImageOptions(),
  required this.allowMultiple,
  this.limit,
}) {
  _validate(allowMultiple: allowMultiple, limit: limit);
}

static void _validate({required bool allowMultiple, int? limit}) {
  if (!allowMultiple && limit != null) {
    throw ArgumentError.value(
      allowMultiple,
      'allowMultiple',
      'cannot be false, when limit is not null',
    );
  }

  if (limit != null && limit < 2) {
    throw ArgumentError.value(limit, 'limit', 'cannot be lower then 2');
  }
}
```

What do you think?

tarrinneal: I'm torn about it. I think throwing when we don't need to (there is a logical option) is not great. Being more strict is also good though.

@stuartmorgan what do you think?

pdenert: Another option is to create a new MediaOptions.withLimit constructor and remove the limit passing from the default constructor:

```dart
MediaOptions({
  this.imageOptions = const ImageOptions(),
  required this.allowMultiple,
}) : limit = null;

MediaOptions.withLimit({
  this.imageOptions = const ImageOptions(),
  required this.limit,
}) : allowMultiple = true; 
```

stuartmorgan: I would throw for invalid combinations of parameters. Specialized constructors aren't a great option for a class that we expect to grow over time, because the combinatorics can become quite bad over time.

gmackall (commented):
Android side mostly LGTM (one comment about code duplication), but similar to iOS it will need native unit tests for the changes to `ImagePickerDelegate` and `ImagePickerPlugin` (in [`ImagePickerDelegateTest`](https://github.com/flutter/packages/blob/38711aca66fa91cff1815839f99e9819f3c27d60/packages/image_picker/image_picker_android/android/src/test/java/io/flutter/plugins/imagepicker/ImagePickerDelegateTest.java) and [`ImagePickerPluginTest`](https://github.com/flutter/packages/blob/38711aca66fa91cff1815839f99e9819f3c27d60/packages/image_picker/image_picker_android/android/src/test/java/io/flutter/plugins/imagepicker/ImagePickerPluginTest.java) respectively).

Also, the existing tests will need to have the limit parameter added where needed... I think some of them should be failing as is? See [this test](https://github.com/flutter/packages/blob/38711aca66fa91cff1815839f99e9819f3c27d60/packages/image_picker/image_picker_android/android/src/test/java/io/flutter/plugins/imagepicker/ImagePickerDelegateTest.java#L163), for example, which should certainly fail without the new limit parameter. cc @stuartmorgan do you happen to know if these tests aren't running as part of pre-submit?

Finally it looks like there is a bug in the lint?
```
/packages/packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/ImagePickerDelegate.java:309: Error: Call requires API level 33 (current min is 16): android.provider.MediaStore#getPickImagesMaxLimit [NewApi]
            effectiveLimit = MediaStore.getPickImagesMaxLimit();
```
You've done the guarding correctly ([according to the docs](https://developer.android.com/reference/android/provider/MediaStore#getPickImagesMaxLimit()) this method is available in both 33+, and 30+ w/ extension 2), but the lint is still warning. So we might need to suppress that warning.

stuartmorgan:
> Also, the existing tests will need to have the limit parameter added where needed... I think some of them should be failing as is? See [this test](https://github.com/flutter/packages/blob/38711aca66fa91cff1815839f99e9819f3c27d60/packages/image_picker/image_picker_android/android/src/test/java/io/flutter/plugins/imagepicker/ImagePickerDelegateTest.java#L163), for example, which should certainly fail without the new limit parameter. cc @stuartmorgan do you happen to know if these tests aren't running as part of pre-submit?

Native unit tests are the step after lint in `android_platform_tests`, and CI stops after failures unless we explicitly tell it not to, so currently any failures in native unit tests would be masked by the lint failure.

packages/image_picker/image_picker_platform_interface/lib/src/types/media_options.dart
pdenert: I've updated PR with a version that throws for invalid parameter combinations.

pdenert: Adding body with validate to the default MediaOptions constructor makes it non const, so it'll break the contract for other packages (web, windows, mac, linux). Finally, I've added to the MediaOptions createAndValidate constructor, the same as it's done for ImageOptions, and mimic it's behaviour in further implementation to be consistent.

pdenert:
@gmackall  I've extracted getting final limit to `getLimitFromOptions` in `ImagePickerUtils` and suppressed the lint warning. I've also updated the native tests and added new ones for limit

pdenert:
> Quick review of the platform interface. I'll try to get a full review in asap for you.

@tarrinneal hey, may you add your full review soon?

pdenert:
> > Quick review of the platform interface. I'll try to get a full review in asap for you.
> 
> @tarrinneal hey, may you add your full review soon?

@tarrinneal just friendly reminder that I'm still waiting for review

tarrinneal:
Sorry for the delays.

This seems ready to break up into it's respective pr's as per https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changing-federated-plugins

Let me know if you need help with any of this.

pdenert:
PR with platform interface changes only is ready here: https://github.com/flutter/packages/pull/6434

pdenert:
@tarrinneal I have updated the PR to the published platform interface changes

pdenert:
@tarrinneal may you look at this soon?

tarrinneal (approved):
Thanks for putting in the time and effort on this pr. 

(and sorry for the delays in approval)

PR #6205, opened 2024-02-26
[google_map_flutter] Add style to widget - platform impls

stuartmorgan:
Platform implementations portion of https://github.com/flutter/packages/pull/6192

Adds handling of `MapConfiguration.style` and implements new `getStyleError` method.

Part of https://github.com/flutter/flutter/issues/66207

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
```diff
@@ -279,15 +282,36 @@ class GoogleMapController {
...
+  // source of truth for style info. Currently it's tracked and handled
+  // separately since style didn't used to be part of the configuration.
+  List<gmaps.MapTypeStyle> _updateStylesFromConfiguration(
+      MapConfiguration update) {
+    if (update.style != null) {
+      // Provide async access to the error rather than throwing, to match the
+      // behavior of other platforms where there's no mechanism to return errors
+      // from configuration updates.
+      try {
+        _lastStyles = _mapStyles(update.style);
```
ditman, line 295: Do we want a test that asserts that `_lastStyles` remain unchanged when `update.style` is wrong? (not sure if we want that to be a feature or is a side-effect of this particular implementation)

ditman (approved):
Web bits lgtm!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
stuartmorgan: That does seem worth testing. I couldn't see a way to add it without adding another kind of ugly test hook, so I did that; let me know if you have a better option. (Long term we may want to wrap that opaque maps object instead of having to add these hooks at this layer.)

ditman: The new hook is fine. There seems to exist a way to fetch styles from the gmaps.GMap object through its MVCObject interface:

```dart
map.get('styles')
```

([`get` DartDocs](https://pub.dev/documentation/google_maps/7.1.0/google_maps/MVCObject/get.html))

But I'm not sure if whatever that returns is directly comparable with what we pass in!

stuartmorgan: We shouldn't need to compare with what we pass in, just the before/after, but I'm getting null from that getter. Maybe this is related to needing to pump things differently in the maps integration tests?

For now I'm just going to land this since it's a pattern we already have, and we can revisit the structure of these tests later if necessary.

stuartmorgan:
(Landing without reviews from the others since it's identical to the version already approved in the main PR.)

PR #6216, opened 2024-02-28
Add library annotations for js interop

srujzs:
These are needed for 3.3 in order to address issue https://github.com/dart-lang/sdk/issues/54801. This commit also cleans up some unnecessary annotations.

Also publishes 0.5.7.

https://github.com/dart-lang/sdk/issues/55039


stuartmorgan (approved):
I changed this from a "minor" version bump to a "bugfix" version bump (using Dart conventions for <1.0); LGTM with that change.

PR #6218, opened 2024-02-28
[tool] Add features to support GCB auto-publish flow

stuartmorgan:
Adds the flowing to the tool:
- A new `--exact-match-only` flag to be used with `--packages` to prevent group matching (i.e., a selection like `--packages=path_provider --exact-match-only` would only run on `packages/path_provider/path_provider`, not `packages/path_provider/*`).
- Two new `publish` command flags:
  - `--tag-for-auto-publish`, to do all the steps that `publish` currently does except for the real `pub publish`, so it would dry-run the publish and then create and push the tag if successful.
  - `--already-tagged`, to skip the step of adding and pushing a tag, and replace it with a check that `HEAD` already has the expected tag.

This set of additions supports a workflow where the current `release` step is changed to use `--tag-for-auto-publish`, and then the separate auto-publish system would publish each package with `... publish --already-tagged --packages=<some package> --exact-match-only`.

See https://github.com/flutter/packages/pull/5005#discussion_r1344542422 for previous discussion/context.

Part of https://github.com/flutter/flutter/issues/126827

### DIFF
--- a/script/tool/lib/src/publish_command.dart
+++ b/script/tool/lib/src/publish_command.dart
@@ -60,7 +60,7 @@ class PublishCommand extends PackageLoopingCommand {
         _stdin = stdinput ?? io.stdin {
     argParser.addFlag(_alreadyTaggedFlag,
         help:
-            'Instead of tagging, validates that the current checkout is aleardy tagged with the expected version.\n'
+            'Instead of tagging, validates that the current checkout is already tagged with the expected version.\n'
             'This is primarily intended for use in CI publish steps triggered by tagging.',
         negatable: false);
     argParser.addMultiOption(_pubFlagsOption,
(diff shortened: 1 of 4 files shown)


script/tool/lib/src/publish_command.dart
```diff
@@ -58,6 +58,11 @@ class PublishCommand extends PackageLoopingCommand {
   })  : _pubVersionFinder =
             PubVersionFinder(httpClient: httpClient ?? http.Client()),
         _stdin = stdinput ?? io.stdin {
+    argParser.addFlag(_alreadyTaggedFlag,
+        help:
+            'Instead of tagging, validates that the current checkout is aleardy tagged with the expected version.\n'
```
tarrinneal, line 63: typo: already

script/tool/lib/src/publish_command.dart
```diff
@@ -83,13 +88,20 @@ class PublishCommand extends PackageLoopingCommand {
...
+        negatable: false);
   }
 
+  static const String _alreadyTaggedFlag = 'already-tagged';
   static const String _pubFlagsOption = 'pub-publish-flags';
   static const String _remoteOption = 'remote';
   static const String _allChangedFlag = 'all-changed';
   static const String _dryRunFlag = 'dry-run';
   static const String _skipConfirmationFlag = 'skip-confirmation';
+  static const String _tagForAutopublishFlag = 'tag-for-autopublish';
```
tarrinneal, line 104: is auto publish one word?

tarrinneal (approved):
little nits

script/tool/lib/src/publish_command.dart
stuartmorgan: Little known fact, "aleardy" is actually the adverbal form of "all ears"; this is validating that it was tagged in a way that makes it all ears.

(Fixed.)

script/tool/lib/src/publish_command.dart
stuartmorgan: Nope! Fixed.

PR #6221, opened 2024-02-28
[webview] Add macOS support in implementation package

stuartmorgan:
Adds initial support for macOS.

Known limitations:
- Scroll APIs are not supported, and currently throw unimplemented. Longer term we could consider JS polyfills, but we'd probably want to make them opt-in since injecting JS into every page has some risks.
- Setting the background color doesn't work since the view structure is different on macOS. We may be able to use a layer-backed view to support this in the future.

Since many use cases don't require these features, my preference is to land this without that support, rather than blocking on having full parity, and then the specific features can be added later based on community interest (demand and/or contributions).

This requires 3.24, as 3.22 does not yet have gesture support for macOS platform views, and pushing this to users without gesture support would be very confusing.

Part of https://github.com/flutter/flutter/issues/41725


packages/webview_flutter/webview_flutter_wkwebview/lib/src/ui_kit/ui_kit.dart
stuartmorgan, line 1: @bparrishMines I'm curious what you think of this approach to handle the fact that `WKWebView` has a different base class on iOS vs macOS.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_proxy.dart
```diff
@@ -23,26 +36,19 @@ WKWebsiteDataStore _defaultWebsiteDataStore() =>
 /// it intends to return.
 class WebKitProxy {
   /// Constructs a [WebKitProxy].
-  const WebKitProxy({
-    this.createWebView = WKWebView.new,
+  WebKitProxy({
+    WebViewConstructor? createWebView,
```
stuartmorgan, line 40: @bparrishMines Unfortunately I had to make this non-const because of the `Platform` switch. Better ideas welcome!

packages/webview_flutter/webview_flutter_wkwebview/lib/src/web_kit/web_kit.dart
```diff
@@ -1168,17 +1173,89 @@ class WKWebView extends UIView {
...
   }
+}
 
-  /// The custom user agent string.
-  ///
-  /// Represents [WKWebView.customUserAgent](https://developer.apple.com/documentation/webkit/wkwebview/1414950-customuseragent?language=objc).
-  Future<String?> getCustomUserAgent() {
-    return _webViewApi.getCustomUserAgentForInstances(this);
+/// The iOS version of a WKWebView.
+class WKWebViewIOS extends WKWebView implements UIView {
```
stuartmorgan, line 1179: @bparrishMines This is the other half of how I'm handling the different parenting: the non-abstract classes are these subclasses which are different for the two platforms.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_proxy.dart
bparrishMines: You should be able to set this with a constant Function. Either by putting `(Platform.isIOS ? WKWebViewIOS.new : WKWebViewMacOS.new);` in a static method or a  top-level function.

I've also have been waiting for this feature! : https://github.com/dart-lang/language/issues/1048

packages/webview_flutter/webview_flutter_wkwebview/lib/src/ui_kit/ui_kit.dart
bparrishMines: I like this approach for sharing code between the platforms. Especially the part of creating a platform agnostic `WKWebView` class. I can also see this working if we need to add the `NSView` class for macOS. Although I'm not sure how this translates for the Pigeon wrapper generator since the Dart classes would become completely generated.

packages/webview_flutter/webview_flutter_wkwebview/darwin/Classes/FLTWebViewFlutterPlugin.m
```diff
@@ -106,10 +125,12 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
   SetUpFWFNSUrlHostApi(registrar.messenger,
                        [[FWFURLHostApiImpl alloc] initWithBinaryMessenger:registrar.messenger
                                                           instanceManager:instanceManager]);
+#if TARGET_OS_IOS
   SetUpFWFUIScrollViewDelegateHostApi(
```
hellohuanlin, line 129: is there a macos version of scroll view delegate?

packages/webview_flutter/webview_flutter_wkwebview/darwin/Classes/FWFScrollViewHostApi.m
```diff
@@ -24,47 +24,65 @@ - (instancetype)initWithInstanceManager:(FWFInstanceManager *)instanceManager {
...
 }
 
 - (NSArray<NSNumber *> *)
     contentOffsetForScrollViewWithIdentifier:(NSInteger)identifier
                                        error:(FlutterError *_Nullable *_Nonnull)error {
+#if TARGET_OS_IOS
   CGPoint point = [[self scrollViewForIdentifier:identifier] contentOffset];
   return @[ @(point.x), @(point.y) ];
+#else
+  return @[ @(0), @(0) ];
```
hellohuanlin, line 53: will this function be called by macOS? if not, maybe use `API_AVAILABLE` annotation? same below

packages/webview_flutter/webview_flutter_wkwebview/darwin/Classes/FWFWebViewConfigurationHostApi.m
```diff
@@ -95,21 +95,25 @@ - (void)setAllowsInlineMediaPlaybackForConfigurationWithIdentifier:(NSInteger)id
                                                              error:
                                                                  (FlutterError *_Nullable *_Nonnull)
                                                                      error {
+#if TARGET_OS_IOS
   [[self webViewConfigurationForIdentifier:identifier] setAllowsInlineMediaPlayback:allow];
+#endif
+  // No-op, rather than error out, on macOS, since it's not a meaningful option on macOS and it's
+  // easier for clients if it's just ignored.
```
hellohuanlin, line 102: do we want to just `NSLog` here in case of confusion?

stuartmorgan:
@alterhuman Please see my comments in the issue; there is no plan to land this PR until platform view gesture support for macOS reaches `stable`, and it hasn't even landed on `master` yet.

wanjm:
> @alterhuman Please see my comments in the issue; there is no plan to land this PR until platform view gesture support for macOS reaches `stable`, and it hasn't even landed on `master` yet.

gesture support has landed on mater on Mar 9th, does it mean that I can use mater and the code to do a prototype?

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_proxy.dart
stuartmorgan: Thanks for the tip! I've re-added the consts I removed.

Unfortunately I had to duplicate all the arguments, because I couldn't make a static function that returned `Platform.isIOS ? WKWebViewIOS.new : WKWebViewMacOS.new` and then call that because it still didn't want me calling the function in the const constructor, so I had to instead make it a wrapper that calls the right constructor each time, passing all the args through. (If I missed a way to make that work please let me know!)

packages/webview_flutter/webview_flutter_wkwebview/darwin/Classes/FLTWebViewFlutterPlugin.m
stuartmorgan: AFAICT the only way to implement scroll functionality on the macOS version of WKWebView without mucking around with internal view details is JS.

packages/webview_flutter/webview_flutter_wkwebview/darwin/Classes/FWFWebViewConfigurationHostApi.m
stuartmorgan: In practice, people are going to be calling this with `YES`, which is already how macOS works, so it would be a spammy warning that wouldn't really mean anything.

gvsakhil:
Any update on when this will be available?

stuartmorgan:
@bparrishMines @hellohuanlin This is ready for review. As noted in the PR description it won't land until the next stable, but since the PR is in a good state now we can work through reviews now so it's ready to go when the next stable arrives.

packages/webview_flutter/webview_flutter_wkwebview/example/integration_test/legacy/webview_flutter_test.dart
```diff
@@ -757,7 +757,10 @@ Future<void> main() async {
       isPaused = await controller.runJavascriptReturningResult('isPaused();');
       expect(isPaused, _webviewBool(false));
     });
-  });
+  },
+      // OGG playback is not supported on macOS, so the test data would need
+      // to be changed to support macOS.
+      skip: Platform.isMacOS);
```
cbracken, line 763: nit: Is this indentation correct?

packages/webview_flutter/webview_flutter_wkwebview/darwin/Classes/FWFInstanceManager.m
```diff
@@ -140,17 +140,17 @@ - (long)identifierWithStrongReferenceForInstance:(nonnull NSObject *)instance {
...
   });
-  return identifierNumber ? identifierNumber.longValue : NSNotFound;
+  return identifierNumber != nil ? identifierNumber.longValue : NSNotFound;
 }
 
 - (BOOL)containsInstance:(nonnull NSObject *)instance {
   BOOL __block containsInstance;
   dispatch_sync(_lockQueue, ^{
-    containsInstance = [self.identifiers objectForKey:instance];
+    containsInstance = [self.identifiers objectForKey:instance] != nil;
```
hellohuanlin, line 153: nit: (probably unrelated to this pr so no need to change): can just use pthread or NSLock if `__block` feels tedious to write. 

Using queues as mutex locks feels a bit overkill although many people are doing it so 🤷. iirc i ran a benchmark super long time ago and locks (even the slow NSLock) are A LOT faster than queues.

hellohuanlin (approved):
Only reviewed the objc part and looks very clean

packages/webview_flutter/webview_flutter_wkwebview/example/integration_test/legacy/webview_flutter_test.dart
stuartmorgan: I don't know why the formatter does this; it's something to do with having a comment inline. We hit this with `skip` frequently, unfortunately.

The only alternative I can get the formatter to do is by adding a comma after `isMacOS`, but then it indents the entire test more, which is annoying for consistency with other tests.

cbenhagen:
Can you add some information why setting a background is not supported on macOS? I guess most will want to make the background transparent. Is there a possibility to allow this special case?

stuartmorgan:
> Can you add some information why setting a background is not supported on macOS?

Sorry, I'm not sure what you are asking exactly. The PR description and my issue comment about the limitations of the PR discuss this; if you are looking for details that aren't there you'll need to be more specific.

> Is there a possibility to allow this special case?

As always, anyone is welcome to submit PRs adding functionality they need.

bparrishMines (approved):
LGTM

This looks fine to land as is. I'll probably need to change the structure when the Swift ProxyApi generator is finished since I don't think it's compatible with this structure anymore. Mainly because the generated native file would reference the native class directly and the Dart class would be generated.

LouiseHsu:
@stuartmorgan is this okay to merge or are there still outstanding issues?

stuartmorgan:
> This will not be landed until the next full stable release after 3.22, as 3.22 does not yet have gesture support for macOS platform views, and pushing this to users without gesture support would be very confusing.

Per the PR description:

>> This will not be landed until the next full stable release after 3.22, as 3.22 does not yet have gesture support for macOS platform views, and pushing this to users without gesture support would be very confusing.

cbracken (approved):
Clearing out my pending reviews post-holiday. lgtm!

stuartmorgan:
This is blocked on Flutter `main` reporting a version number higher than 3.24 (and that change rolling into this repo).

igor-lukyanov:
@stuartmorgan Does this mean that the branch will be merged into a future version of Flutter, not 3.24? Trying to understand the plans to adjust own expectations or start looking for a workaround. Thank you.

stuartmorgan:
> @stuartmorgan they mean when will the package `webview_flutter` be updated with macOS support.

I'm not going to assume that I know that someone means something very different from what they said, so I answered the question they actually asked.

> I have the same question to prioritise feature development order in my flutter app.

It will land when it's no longer blocked; see [my recent comment](https://github.com/flutter/packages/pull/6221#issuecomment-2273848038).

wanjm:
> This is blocked on Flutter `main` reporting a version number higher than 3.24 (and that change rolling into this repo).

what do you expected the version number to be on the branch main?

PR #6224, opened 2024-02-29
[pigeon] Allow multi instance support with message channel name suffix

tarrinneal:
Adds the ability to instantiate any api using a suffix string to allow for unique instances. 

partial https://github.com/flutter/flutter/issues/66710

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/pigeon/example/app/android/app/src/main/java/io/flutter/plugins/Messages.java
+++ b/packages/pigeon/example/app/android/app/src/main/java/io/flutter/plugins/Messages.java
@@ -268,11 +268,20 @@ public class Messages {
     }
     /** Sets up an instance of `ExampleHostApi` to handle messages through the `binaryMessenger`. */
     static void setUp(@NonNull BinaryMessenger binaryMessenger, @Nullable ExampleHostApi api) {
+      setUp(binaryMessenger, "", api);
+    }
+
+    static void setUp(
+        @NonNull BinaryMessenger binaryMessenger,
+        @NonNull String messageChannelSuffix,
+        @Nullable ExampleHostApi api) {
+      messageChannelSuffix = messageChannelSuffix.isEmpty() ? "" : "." + messageChannelSuffix;
       {
         BasicMessageChannel<Object> channel =
             new BasicMessageChannel<>(
                 binaryMessenger,
-                "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.getHostLanguage",
+                "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.getHostLanguage"
+                    + messageChannelSuffix,
                 getCodec());
         if (api != null) {
           channel.setMessageHandler(
@@ -295,7 +304,8 @@ public class Messages {
         BasicMessageChannel<Object> channel =
             new BasicMessageChannel<>(
                 binaryMessenger,
-                "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.add",
+                "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.add"
+                    + messageChannelSuffix,
                 getCodec());
         if (api != null) {
           channel.setMessageHandler(
@@ -324,7 +334,8 @@ public class Messages {
         BasicMessageChannel<Object> channel =
             new BasicMessageChannel<>(
                 binaryMessenger,
-                "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.sendMessage",
+                "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.sendMessage"
+                    + messageChannelSuffix,
                 getCodec());
         if (api != null) {
           channel.setMessageHandler(
@@ -356,9 +367,16 @@ public class Messages {
   /** Generated class from Pigeon that represents Flutter messages that can be called from Java. */
   public static class MessageFlutterApi {
     private final @NonNull BinaryMessenger binaryMessenger;
+    private final String messageChannelSuffix;
 
     public MessageFlutterApi(@NonNull BinaryMessenger argBinaryMessenger) {
+      this(argBinaryMessenger, "");
+    }
+
+    public MessageFlutterApi(
+        @NonNull BinaryMessenger argBinaryMessenger, @NonNull String messageChannelSuffix) {
       this.binaryMessenger = argBinaryMessenger;
+      this.messageChannelSuffix = messageChannelSuffix.isEmpty() ? "" : "." + messageChannelSuffix;
     }
 
     /** Public interface for sending reply. */
@@ -369,7 +387,8 @@ public class Messages {
 
     public void flutterMethod(@Nullable String aStringArg, @NonNull Result<String> result) {
       final String channelName =
-          "dev.flutter.pigeon.pigeon_example_package.MessageFlutterApi.flutterMethod";
+          "dev.flutter.pigeon.pigeon_example_package.MessageFlutterApi.flutterMethod"
+              + messageChannelSuffix;
       BasicMessageChannel<Object> channel =
           new BasicMessageChannel<>(binaryMessenger, channelName, getCodec());
       channel.send(
--- a/packages/pigeon/example/app/ios/Runner/Messages.g.swift
+++ b/packages/pigeon/example/app/ios/Runner/Messages.g.swift
@@ -131,9 +131,13 @@ class ExampleHostApiSetup {
   /// The codec used by ExampleHostApi.
   static var codec: FlutterStandardMessageCodec { ExampleHostApiCodec.shared }
   /// Sets up an instance of `ExampleHostApi` to handle messages through the `binaryMessenger`.
-  static func setUp(binaryMessenger: FlutterBinaryMessenger, api: ExampleHostApi?) {
+  static func setUp(
+    binaryMessenger: FlutterBinaryMessenger, api: ExampleHostApi?, messageChannelSuffix: String = ""
+  ) {
+    let channelSuffix = messageChannelSuffix.count > 0 ? ".\(messageChannelSuffix)" : ""
     let getHostLanguageChannel = FlutterBasicMessageChannel(
-      name: "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.getHostLanguage",
+      name:
+        "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.getHostLanguage\(channelSuffix)",
       binaryMessenger: binaryMessenger, codec: codec)
     if let api = api {
       getHostLanguageChannel.setMessageHandler { _, reply in
@@ -148,7 +152,7 @@ class ExampleHostApiSetup {
       getHostLanguageChannel.setMessageHandler(nil)
     }
     let addChannel = FlutterBasicMessageChannel(
-      name: "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.add",
+      name: "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.add\(channelSuffix)",
       binaryMessenger: binaryMessenger, codec: codec)
     if let api = api {
       addChannel.setMessageHandler { message, reply in
@@ -166,7 +170,7 @@ class ExampleHostApiSetup {
       addChannel.setMessageHandler(nil)
     }
     let sendMessageChannel = FlutterBasicMessageChannel(
-      name: "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.sendMessage",
+      name: "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.sendMessage\(channelSuffix)",
       binaryMessenger: binaryMessenger, codec: codec)
     if let api = api {
       sendMessageChannel.setMessageHandler { message, reply in
@@ -193,14 +197,16 @@ protocol MessageFlutterApiProtocol {
 }
 class MessageFlutterApi: MessageFlutterApiProtocol {
   private let binaryMessenger: FlutterBinaryMessenger
-  init(binaryMessenger: FlutterBinaryMessenger) {
+  private let messageChannelSuffix: String
+  init(binaryMessenger: FlutterBinaryMessenger, messageChannelSuffix: String = "") {
     self.binaryMessenger = binaryMessenger
+    self.messageChannelSuffix = messageChannelSuffix.count > 0 ? ".\(messageChannelSuffix)" : ""
   }
   func flutterMethod(
     aString aStringArg: String?, completion: @escaping (Result<String, FlutterError>) -> Void
   ) {
     let channelName: String =
-      "dev.flutter.pigeon.pigeon_example_package.MessageFlutterApi.flutterMethod"
+      "dev.flutter.pigeon.pigeon_example_package.MessageFlutterApi.flutterMethod\(messageChannelSuffix)"
     let channel = FlutterBasicMessageChannel(name: channelName, binaryMessenger: binaryMessenger)
     channel.sendMessage([aStringArg] as [Any?]) { response in
       guard let listResponse = response as? [Any?] else {
--- a/packages/pigeon/example/app/macos/Runner/messages.g.h
+++ b/packages/pigeon/example/app/macos/Runner/messages.g.h
@@ -56,11 +56,17 @@ NSObject<FlutterMessageCodec> *PGNExampleHostApiGetCodec(void);
 extern void SetUpPGNExampleHostApi(id<FlutterBinaryMessenger> binaryMessenger,
                                    NSObject<PGNExampleHostApi> *_Nullable api);
 
+extern void SetUpPGNExampleHostApiWithSuffix(id<FlutterBinaryMessenger> binaryMessenger,
+                                             NSObject<PGNExampleHostApi> *_Nullable api,
+                                             NSString *messageChannelSuffix);
+
 /// The codec used by PGNMessageFlutterApi.
 NSObject<FlutterMessageCodec> *PGNMessageFlutterApiGetCodec(void);
 
 @interface PGNMessageFlutterApi : NSObject
 - (instancetype)initWithBinaryMessenger:(id<FlutterBinaryMessenger>)binaryMessenger;
+- (instancetype)initWithBinaryMessenger:(id<FlutterBinaryMessenger>)binaryMessenger
+                   messageChannelSuffix:(nullable NSString *)messageChannelSuffix;
 - (void)flutterMethodAString:(nullable NSString *)aString
                   completion:(void (^)(NSString *_Nullable, FlutterError *_Nullable))completion;
 @end
--- a/packages/pigeon/example/app/macos/Runner/messages.g.m
+++ b/packages/pigeon/example/app/macos/Runner/messages.g.m
@@ -138,9 +138,21 @@ NSObject<FlutterMessageCodec> *PGNExampleHostApiGetCodec(void) {
 
 void SetUpPGNExampleHostApi(id<FlutterBinaryMessenger> binaryMessenger,
                             NSObject<PGNExampleHostApi> *api) {
+  SetUpPGNExampleHostApiWithSuffix(binaryMessenger, api, @"");
+}
+
+void SetUpPGNExampleHostApiWithSuffix(id<FlutterBinaryMessenger> binaryMessenger,
+                                      NSObject<PGNExampleHostApi> *api,
+                                      NSString *messageChannelSuffix) {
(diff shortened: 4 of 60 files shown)


packages/pigeon/platform_tests/test_plugin/windows/test_plugin.cpp
```diff
@@ -23,24 +23,55 @@ using core_tests_pigeontest::AnEnum;
...
+  result(std::nullopt);
+}
+
 // static
 void TestPlugin::RegisterWithRegistrar(
     flutter::PluginRegistrarWindows* registrar) {
   auto plugin = std::make_unique<TestPlugin>(registrar->messenger());
 
-  HostIntegrationCoreApi::SetUp(registrar->messenger(), plugin.get());
+  HostIntegrationCoreApi::SetUp(registrar->messenger(), plugin.get(), nullptr);
```
stuartmorgan: Remove the last argument here.

packages/pigeon/platform_tests/test_plugin/windows/test_plugin.cpp
```diff
@@ -23,24 +23,55 @@ using core_tests_pigeontest::AnEnum;
...
 using flutter::EncodableMap;
 using flutter::EncodableValue;
 
+// static
+void TestSmallApi::RegisterWithRegistrar(
+    flutter::PluginRegistrarWindows* registrar,
+    const std::string& message_channel_suffix) {
+  auto api = std::make_unique<TestSmallApi>(registrar->messenger());
+
+  HostSmallApi::SetUp(registrar->messenger(), api.get(),
```
stuartmorgan, line 39: You're passing a raw pointer to `api` here, but `api` is a smart pointer whose only owner is the local scope of this function, so as soon as `RegisterWithRegistrar` returns, the handler for Pigeon calls is a dangling pointer.

You either need to make `TestSmallApi` a `Plugin` and use `AddPlugin` to keep it around, or you need to remove this whole quasi-plugin method and just have `TestPlugin` call the constructor for this object and keep it around as an ivar. I think the latter would be less confusing, since this isn't actually a plugin with a registration method that Flutter will call.

stuartmorgan (commented):
With the changes below the tests run, but there's a bad allocation exception being thrown that I'm still looking into.

packages/pigeon/platform_tests/test_plugin/windows/test_plugin.cpp
tarrinneal: Makes sense, I tried to keep the structure as similar to the other class as possible, but it seems it was a mistake here. Thanks

packages/pigeon/example/app/android/app/src/main/java/io/flutter/plugins/Messages.java
```diff
@@ -268,11 +268,19 @@ public interface ExampleHostApi {
...
+        @NonNull BinaryMessenger binaryMessenger,
+        @NonNull String messageChannelSuffix,
+        @Nullable ExampleHostApi api) {
       {
         BasicMessageChannel<Object> channel =
             new BasicMessageChannel<>(
                 binaryMessenger,
-                "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.getHostLanguage",
+                "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.getHostLanguage"
+                    + messageChannelSuffix,
```
stuartmorgan, line 283: I was thinking about this again after offline discussion, and I think we actually need to add the separator ourselves even though it'll make the code more complicated. Otherwise the following could happen:
- API declares a method called `foo`, and another called `foobar`.
- A non-suffixed version of the API is instantiated.
- A suffixed version is also instantiated, with the suffix `bar`.

Now Pigeon has tried to register two handlers for the same name: `dev.flutter.pigeon.package_name.ApiName.foobar`

This is an edge case, but with dynamic suffixes based on something other than a counter it's not impossible to imagine happening, and it would be *really* confusing for clients if it ever did. And we can prevent that edge case by just adding the separator any time there's a non-empty suffix.

packages/pigeon/example/app/android/app/src/main/java/io/flutter/plugins/Messages.java
```diff
@@ -356,9 +366,16 @@ public void error(Throwable error) {
...
+    private final String messageChannelSuffix;
 
     public MessageFlutterApi(@NonNull BinaryMessenger argBinaryMessenger) {
+      this(argBinaryMessenger, "");
+    }
+
+    public MessageFlutterApi(
+        @NonNull BinaryMessenger argBinaryMessenger, @NonNull String messageChannelSuffix) {
       this.binaryMessenger = argBinaryMessenger;
+      this.messageChannelSuffix = messageChannelSuffix;
```
stuartmorgan, line 378: We can add it at this stage so we only have to have the logic in one spot.

packages/pigeon/example/app/macos/Runner/messages.g.m
```diff
@@ -218,13 +235,24 @@ - (instancetype)initWithBinaryMessenger:(NSObject<FlutterBinaryMessenger> *)bina
   self = [super init];
```
stuartmorgan, line 235: While it probably doesn't matter much here, the expectation in Obj-C (which can matter in subclassing, for instance) is that this method would delegate to the more-argument-rich version of init. So `return [self initWithBinaryMessenger:binaryMessenger messageChannelSuffix:@""];`

packages/pigeon/example/app/macos/Runner/messages.g.h
```diff
@@ -56,11 +56,17 @@ NSObject<FlutterMessageCodec> *PGNExampleHostApiGetCodec(void);
...
+                                             NSObject<PGNExampleHostApi> *_Nullable api,
+                                             NSString *messageChannelSuffix);
+
 /// The codec used by PGNMessageFlutterApi.
 NSObject<FlutterMessageCodec> *PGNMessageFlutterApiGetCodec(void);
 
 @interface PGNMessageFlutterApi : NSObject
 - (instancetype)initWithBinaryMessenger:(id<FlutterBinaryMessenger>)binaryMessenger;
+- (instancetype)initWithBinaryMessenger:(id<FlutterBinaryMessenger>)binaryMessenger
+                   messageChannelSuffix:(NSString *)messageChannelSuffix;
```
stuartmorgan, line 69: Optional: it would be a little more idiomatic to make this `nullable NSString *` and then convert `nil` to `@""` internally. (On the other hand, it's very unlikely anyone would ever use this method with no suffix, given the other method.)

packages/pigeon/lib/dart_generator.dart
```diff
@@ -349,7 +352,7 @@ $resultAt != null
         indent.newln();
       }
       indent.write(
-          'static void setup(${api.name}? api, {BinaryMessenger? binaryMessenger}) ');
+          "static void setup(${api.name}? api, {BinaryMessenger? binaryMessenger, String messageChannelSuffix = '',}) ");
```
stuartmorgan, line 355: I don't suppose I could convince you to make this a new method called `setUp`, and the old one deprecated and delegate to the new version? 😬 

(I know it's not really related, but you're touching it so I had to try...)

packages/pigeon/platform_tests/alternate_language_test_plugin/ios/Classes/AlternateLanguageTestPlugin.m
```diff
@@ -500,4 +508,53 @@ - (void)callFlutterEchoNullableEnum:(nullable AnEnumBox *)AnEnumBoxed
...
+
+@end
+
+@interface AlternateLanguageTestAPIWithSuffix ()
+@end
+
+@implementation AlternateLanguageTestAPIWithSuffix
++ (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar
+                       suffix:(NSString *)suffix {
+  AlternateLanguageTestAPIWithSuffix *api = [[AlternateLanguageTestAPIWithSuffix alloc] init];
```
stuartmorgan, line 551: For all the languages, I think this secondary plugin should take the suffix as an argument, and it should use that in its echo implementation, so that you can address the problem you mentioned of not currently actually testing that they are registering their respective handlers.

stuartmorgan (changes requested):

packages/pigeon/example/app/android/app/src/main/java/io/flutter/plugins/Messages.java
tarrinneal: Makes perfect sense.

packages/pigeon/platform_tests/alternate_language_test_plugin/ios/Classes/AlternateLanguageTestPlugin.m
tarrinneal: I did actually write a test that confirms there are multiple (dart) implementations with unique suffixes. Since the integration tests with multiple channels pass, and we know that the dart api's are using the correct channels, we also know that the platforms are using them.

stuartmorgan: Ah, clever! I didn't look closely at the unit test. SGTM.

packages/pigeon/example/app/android/app/src/main/java/io/flutter/plugins/Messages.java
tarrinneal: if only it were so simple

tarrinneal:
This is fully ready for review @stuartmorgan @hellohuanlin and @bparrishMines for anyone who wants to give it a look

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -131,9 +131,12 @@ class ExampleHostApiSetup {
...
   static var codec: FlutterStandardMessageCodec { ExampleHostApiCodec.shared }
   /// Sets up an instance of `ExampleHostApi` to handle messages through the `binaryMessenger`.
-  static func setUp(binaryMessenger: FlutterBinaryMessenger, api: ExampleHostApi?) {
+  static func setUp(
+    binaryMessenger: FlutterBinaryMessenger, api: ExampleHostApi?, messageChannelSuffix: String = ""
+  ) {
     let getHostLanguageChannel = FlutterBasicMessageChannel(
-      name: "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.getHostLanguage",
+      name:
+        "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.getHostLanguage\(messageChannelSuffix.count > 0 ? ".\(messageChannelSuffix)" : "")",
```
hellohuanlin, line 139: this can be just `"dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.getHostLanguage\(messageChannelSuffix)"`

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -193,14 +198,16 @@ protocol MessageFlutterApiProtocol {
 }
 class MessageFlutterApi: MessageFlutterApiProtocol {
   private let binaryMessenger: FlutterBinaryMessenger
-  init(binaryMessenger: FlutterBinaryMessenger) {
+  private let messageChannelSuffix: String
+  init(binaryMessenger: FlutterBinaryMessenger, messageChannelSuffix: String = "") {
     self.binaryMessenger = binaryMessenger
+    self.messageChannelSuffix = messageChannelSuffix.count > 0 ? ".\(messageChannelSuffix)" : ""
```
hellohuanlin, line 204: can be `self.messageChannelSuffix = messageChannelSuffix`

packages/pigeon/example/app/macos/Runner/messages.g.h
```diff
@@ -56,11 +56,17 @@ NSObject<FlutterMessageCodec> *PGNExampleHostApiGetCodec(void);
...
+                                             NSObject<PGNExampleHostApi> *_Nullable api,
+                                             NSString *messageChannelSuffix);
+
 /// The codec used by PGNMessageFlutterApi.
 NSObject<FlutterMessageCodec> *PGNMessageFlutterApiGetCodec(void);
 
 @interface PGNMessageFlutterApi : NSObject
 - (instancetype)initWithBinaryMessenger:(id<FlutterBinaryMessenger>)binaryMessenger;
+- (instancetype)initWithBinaryMessenger:(id<FlutterBinaryMessenger>)binaryMessenger
+                   messageChannelSuffix:(nullable NSString *)messageChannelSuffix;
```
hellohuanlin, line 69: is it intended for this to be nullable (while swift version is nonnull)?

packages/pigeon/example/app/macos/Runner/messages.g.m
```diff
@@ -138,9 +138,21 @@ - (FlutterStandardReader *)readerWithData:(NSData *)data {
...
                             NSObject<PGNExampleHostApi> *api) {
+  SetUpPGNExampleHostApiWithSuffix(binaryMessenger, api, @"");
+}
+
+void SetUpPGNExampleHostApiWithSuffix(id<FlutterBinaryMessenger> binaryMessenger,
+                                      NSObject<PGNExampleHostApi> *api,
+                                      NSString *messageChannelSuffix) {
   {
+    messageChannelSuffix = messageChannelSuffix.length > 0
+                               ? [NSString stringWithFormat:@"%@%@", @".", messageChannelSuffix]
```
hellohuanlin, line 149: `[NSString stringWithFormat:@".%@", messageChannelSuffix]`

packages/pigeon/example/app/macos/Runner/messages.g.m
```diff
@@ -158,8 +170,15 @@ void SetUpPGNExampleHostApi(id<FlutterBinaryMessenger> binaryMessenger,
     }
   }
   {
+    messageChannelSuffix = messageChannelSuffix.length > 0
+                               ? [NSString stringWithFormat:@"%@%@", @".", messageChannelSuffix]
```
hellohuanlin, line 174: same here

hellohuanlin (commented):
reviewed the ios part

packages/pigeon/example/app/ios/Runner/Messages.g.swift
tarrinneal: Needs the `.` when it exists

packages/pigeon/example/app/macos/Runner/messages.g.h
tarrinneal: becuase @stuartmorgan told me to. I can make the swift version nullable too, if you think that's more "swift style" than the current implementation.

tarrinneal: In general I don't know that it needs to be nullable, seems weird to have a nullable argument when there's already a version of the method that doesn't have that parameter at all.

tarrinneal: it is nullable in swift anyway isn't it? it has a default value at least

hellohuanlin: it's better to have non-null in swift, so that you have 1 less case to deal with. 

Just wanted point out the fact that the divergence of the 2 APIs.

stuartmorgan: > In general I don't know that it needs to be nullable, seems weird to have a nullable argument when there's already a version of the method that doesn't have that parameter at all.

It's not that it needs to be, it's that the standard practice for Obj-C APIs (e.g., what you'll see in Foundation APIs) is that if you have multiple variants of methods, the extras are generally nullable, indicating that the shorter methods are equivalent to calling the longer ones with nulls.

If it makes life harder for it to be nullable, you could make it non-nullable, I just think it'll just feel somewhat odd to Obj-C developers.

tarrinneal: It's not any harder for it to be nullable, the logic is basically the same. I think it's good as is.

hellohuanlin (approved):

packages/pigeon/lib/dart_generator.dart
```diff
@@ -970,7 +983,8 @@ if (${_varNamePrefix}replyList == null) {
   /// Converts Constructors from the pigeon AST to a `code_builder` Constructor
   /// for a ProxyApi.
   Iterable<cb.Constructor> _proxyApiConstructors(
-    Iterable<Constructor> constructors, {
+    Iterable<Constructor> constructors,
+    AstProxyApi api, {
```
bparrishMines, line 987: I don't think this variable is used in this method.

tarrinneal: oops meant to remove that

packages/pigeon/platform_tests/alternate_language_test_plugin/android/src/main/java/com/example/alternate_language_test_plugin/CoreTests.java
```diff
@@ -1269,11 +1271,20 @@ void callFlutterEchoNullableEnum(
...
         @NonNull BinaryMessenger binaryMessenger, @Nullable HostIntegrationCoreApi api) {
+      setUp(binaryMessenger, "", api);
+    }
+
+    static void setUp(
+        @NonNull BinaryMessenger binaryMessenger,
+        @NonNull String messageChannelSuffix,
+        @Nullable HostIntegrationCoreApi api) {
       {
+        messageChannelSuffix = messageChannelSuffix.isEmpty() ? "" : "." + messageChannelSuffix;
```
bparrishMines, line 1282: Instead of having this generated at the start of every method handler, could this be generated as the first line of `setUp`. This also applies to the other languages.

tarrinneal: Yeah, I did that on the others, I just missed this one.

unknown-undefined:
It appears there could be a bug within the Objective-C generator. Specifically, the issue involves the messageChannelSuffix being repeatedly modified in the generated code. Initially, it's changed to .messageChannelSuffix, but with subsequent generations, an additional dot is prepended each time, leading to ..messageChannelSuffix, then ...messageChannelSuffix, and so on, eventually resulting in .......messageChannelSuffix. 
```void SetUpSDKWithSuffix(id<FlutterBinaryMessenger> binaryMessenger, NSObject<OBPlayerSDK> *api, NSString *messageChannelSuffix) {
  {
      messageChannelSuffix = messageChannelSuffix.length > 0 ? [NSString stringWithFormat: @".%@", messageChannelSuffix] : @"";
    FlutterBasicMessageChannel *channel =
      [[FlutterBasicMessageChannel alloc]
        initWithName:[NSString stringWithFormat:@"%@%@", @"SDK.joinChannel", messageChannelSuffix]
        binaryMessenger:binaryMessenger
        codec:OBPlayerSDKGetCodec()];
   ...
  }
  {
      messageChannelSuffix = messageChannelSuffix.length > 0 ? [NSString stringWithFormat: @".%@", messageChannelSuffix] : @"";
...
```

tarrinneal:
> It appears there could be a bug within the Objective-C generator in `c55a91673d7b`. Specifically, the issue involves the messageChannelSuffix being repeatedly modified in the generated code. Initially, it's changed to `.messageChannelSuffix`, but with subsequent generations, an additional dot is prepended each time, leading to `..messageChannelSuffix`, then `...messageChannelSuffix`, and so on, eventually resulting in `.......messageChannelSuffix`.
> 
> ```
>   {
>       messageChannelSuffix = messageChannelSuffix.length > 0 ? [NSString stringWithFormat: @".%@", messageChannelSuffix] : @"";
>     FlutterBasicMessageChannel *channel =
>       [[FlutterBasicMessageChannel alloc]
>         initWithName:[NSString stringWithFormat:@"%@%@", @"SDK.joinChannel", messageChannelSuffix]
>         binaryMessenger:binaryMessenger
>         codec:SDKGetCodec()];
>    ...
>   }
>   {
>       messageChannelSuffix = messageChannelSuffix.length > 0 ? [NSString stringWithFormat: @".%@", messageChannelSuffix] : @"";
> ...
> ```

I don't think this was an issue, but I was already planning on changing it anyway (and have).

bparrishMines (approved):
LGTM

stuartmorgan:
Could be a difference in local vs bot `clang-format` version. The real fix is to remove the stray semicolon from the ends of those two lines though, then it'll be both correct and consistent :)

stuartmorgan (approved):
LGTM

PR #6228, opened 2024-02-29
[webview_flutter] Make WebKitWebViewWidget unit tests stateless

stuartmorgan:
This eliminates the state in the Dart unit tests for this class, bringing it into alignment with https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#make-each-test-entirely-self-contained.
- `testController` is now returned by `buildWidget`, which clarifies and enforces that the variable can only be used after `buildWidget` is called
- All the mocks are now in a data object that is created via a `configureMocks` call in each test.

The latter will allow easily adjusting specific mocks in individual tests (vs having to create a new group that has an almost-identical duplicate of `setUp`, but with slight changes) which will be useful when adding macOS support, where the mock `WKWebView` will need to be a different class when testing macOS/iOS codepath divergence.


bparrishMines (approved):
LGTM

PR #6262, opened 2024-03-04
[in_app_purchase] Convert to Pigeon

stuartmorgan:
WIP

Replaces manual method channels with Pigeon.

Unfortunately a significant amount of JSON code can't be removed yet even though it's now unused by the plugin, because it's part of the API of the public wrappers, so clients may be using it. We could do a follow-up breaking change later to remove all the JSON logic.

Fixes https://github.com/flutter/flutter/issues/117910

## Pre-launch Checklist

- [ ] I signed the [CLA].


stuartmorgan:
This should be ready to review. There's a lot of code, but I went ahead and did all the methods in one PR since having a hybrid of manual method channel code and Pigeon code was kind of weird, particularly in tests. Each method is its own commit though, so it's possible to review incrementally, and I could cut it up if preferred. (It might be useful to look at one incremental commit to see the pattern, and then review in bulk after that?)

@reidbaker If you could give this a manual sanity test, that would be great, since I don't have a setup to be able to do actual end-to-end tests.

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/billing_client_wrapper.dart
```diff
@@ -256,14 +218,24 @@ class BillingClient {
...
   Future<PurchasesResultWrapper> queryPurchases(ProductType productType) async {
-    return PurchasesResultWrapper.fromJson(
-        (await channel.invokeMapMethod<String, dynamic>(
-              'BillingClient#queryPurchasesAsync(QueryPurchaseParams, PurchaseResponseListener)',
-              <String, dynamic>{
-                'productType': const ProductTypeConverter().toJson(productType)
-              },
-            )) ??
-            <String, dynamic>{});
+    // TODO(stuartmorgan): Investigate whether forceOkResponseCode is actually
```
stuartmorgan, line 221: @reidbaker @gmackall FYI, someone who can dig into the expected behavior should probably take a look at this ASAP. This is a Dart conversion of [this code](https://github.com/flutter/packages/pull/6262/files#diff-3c947ed9281a5b9d78b07931af2e897b70781f8c13769b7c638270e8fb263af0L479-L481), but for the reasons explained in this comment, I'm pretty skeptical that the Java code was correct. I deliberately didn't change it here since this PR is supposed to be a functional no-op, but I did want to flag it.

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/MethodCallHandlerImpl.java
```diff
@@ -144,155 +97,107 @@ void setActivity(@Nullable Activity activity) {
...
-      result.error(ACTIVITY_UNAVAILABLE, "Not attempting to show dialog", null);
+      result.error(new FlutterError(ACTIVITY_UNAVAILABLE, "Not attempting to show dialog", null));
       return;
     }
-    billingClient.showAlternativeBillingOnlyInformationDialog(
-        activity,
-        billingResult -> {
-          result.success(fromBillingResult(billingResult));
-        });
+    try {
```
stuartmorgan, line 140: I had to add all of these try/catch wrappers to all the `*Async` methods because it turns out the plugin is relying on them as part of its API 🙁. We have an integration test that checks that when we pass an empty product query, for instance, that we get back a platform exception (with a specific string, even though we don't control that string, so that could be fun someday).

Ideally we should be preventing invalid cases explicitly instead, but that would change behavior. These `FlutterError`s match the generic error made by the method channel code in the engine, so it's doing the same thing, it's just more explicit now (which means more boilerplate than pre-Pigeon, but also that we can see the anti-pattern—and the need to fix it—more clearly).

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/MethodCallHandlerImpl.java
```diff
@@ -144,155 +97,107 @@ void setActivity(@Nullable Activity activity) {
...
-      default:
-        result.notImplemented();
-    }
-  }
-
-  private void showAlternativeBillingOnlyInformationDialog(final MethodChannel.Result result) {
-    if (billingClientError(result)) {
+  public void showAlternativeBillingOnlyInformationDialog(
+      @NonNull Result<PlatformBillingResult> result) {
+    if (billingClient == null) {
```
reidbaker, line 132: I think this is a behavior change for "endConnection" It used to be safe to call multiple times aka on a null billing client and with this check is is no longer ok.

reidbaker: nvm there is no longer a shared onMethodCall.

packages/in_app_purchase/in_app_purchase_android/android/src/test/java/io/flutter/plugins/inapppurchase/TranslatorTest.java
```diff
@@ -164,20 +163,20 @@ public void fromBillingResult() {
...
   @Test
   public void fromBillingResult_debugMessageNull() {
     BillingResult newBillingResult =
         BillingResult.newBuilder().setResponseCode(BillingClient.BillingResponseCode.OK).build();
-    Map<String, Object> billingResultMap = Translator.fromBillingResult(newBillingResult);
+    Messages.PlatformBillingResult platformResult = Translator.fromBillingResult(newBillingResult);
 
-    assertEquals(billingResultMap.get("responseCode"), newBillingResult.getResponseCode());
-    assertEquals(billingResultMap.get("debugMessage"), newBillingResult.getDebugMessage());
+    assertEquals(platformResult.getResponseCode().longValue(), newBillingResult.getResponseCode());
```
reidbaker, line 178: It seems odd that we would need to get a long value here are you sure  it is needed?

reidbaker (commented):
Partial review. Stopped at https://github.com/flutter/packages/pull/6262/files#r1521913336

packages/in_app_purchase/in_app_purchase_android/android/src/test/java/io/flutter/plugins/inapppurchase/TranslatorTest.java
stuartmorgan: Where I have these in tests it's because the Java compiler was angry at me for trying to pass a `Long` and a `long` (or sometimes `int`) to `assertEquals`, which wants the same kind of object.

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/billing_client_wrapper.dart
reidbaker: Keep the code you have as is for this pr.

reidbaker (approved):

PR #6263, opened 2024-03-05
[pigeon] Fixes double prefixes added to enum names for Objc HostApis and FlutterApis

bparrishMines:
In some areas of the Objc generator, the the name created from [_objcTypeForDartType](https://github.com/flutter/packages/blob/main/packages/pigeon/lib/objc_generator.dart#L1353) is being passed to [_enumName](https://github.com/flutter/packages/blob/main/packages/pigeon/lib/objc_generator.dart#L1196) and both of these methods add a prefix to enums. The locations are when they are used in `HostApi` or `FlutterApi` methods, but the name of the generated enum would be correct. e.g. `FLTEnumName` vs `FLTFLTEnumName`.

This fixes the locations where this is happening by passing the AST name to `_enumName` instead. 

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/pigeon/CHANGELOG.md
+++ b/packages/pigeon/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 17.1.3
+
+* [objc] Fixes double prefixes added to enum names.
+
 ## 17.1.2
 
 * [swift] Separates message call code generation into separate methods.
--- a/packages/pigeon/lib/generator_tools.dart
+++ b/packages/pigeon/lib/generator_tools.dart
@@ -13,7 +13,7 @@ import 'ast.dart';
 /// The current version of pigeon.
 ///
 /// This must match the version in pubspec.yaml.
-const String pigeonVersion = '17.1.2';
+const String pigeonVersion = '17.1.3';
 
 /// Read all the content from [stdin] to a String.
 String readStdin() {
--- a/packages/pigeon/lib/objc_generator.dart
+++ b/packages/pigeon/lib/objc_generator.dart
@@ -359,7 +359,7 @@ class ObjcHeaderGenerator extends StructuredGenerator<ObjcOptions> {
       String? lastArgType;
       String? returnType;
       final String enumReturnType = _enumName(
-        returnTypeName.baseName,
+        func.returnType.baseName,
         suffix: ' *_Nullable',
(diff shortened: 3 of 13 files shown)


packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 17.1.3
```
stuartmorgan, line 1: Isn't this a breaking change for Obj-C clients?

bparrishMines: I assumed it was a bug fix since this creates code that doesn't compile.

bparrishMines: It generates correctly in other places. At least this was the case for me.

tarrinneal:
What scenario is missing from our integration tests that would have triggered this?

bparrishMines:
@tarrinneal It looks like the scenario is a combination of:
1. Using a primitive enum as a method parameter or return value
2. Generates Objective-c code with a prefix

It looks like [CoreTests.gen.h](https://github.com/flutter/packages/blob/main/packages/pigeon/platform_tests/alternate_language_test_plugin/ios/Classes/CoreTests.gen.h) doesn't use a prefix. Although scenario 1 does occur for it.

Would it be a good idea to update core tests with a prefix?

tarrinneal:
> Would it be a good idea to update core tests with a prefix?

yup

bparrishMines:
@tarrinneal Adding the prefix to core tests also found a location where the prefix wasn't added to the enum. I think a better solution might be to actually get rid of the `_enumName` method altogether and just use `_ObjcType` instead. It seems redundant to `_ObjcType` and it seems to be error prone to have both. I will try and explore this option. What are your thoughts?

tarrinneal:
I can take this over if you'd like.

packages/pigeon/CHANGELOG.md
stuartmorgan: Yeah, if it's generating bad references to correctly named declarations, that's just a bugfix.

tarrinneal:
@stuartmorgan this is ready for final review

stuartmorgan (approved):
LGTM

PR #5766, opened 2023-12-28
[local_auth] Implemented plugin to support local auth in macOS

OutdatedGuy:
Created plugin `local_auth_macos` to support macOS platform for biometric and password authentication for the [local_auth](https://pub.dev/packages/local_auth) plugin.

The Objective-C code for macOS is built upon the implementation of iOS with necessary changes to work on macOS.

**Working Demo:**

https://github.com/flutter/packages/assets/74326345/71f31ec1-8d67-4208-9d06-ef2dd8f7ad89

<hr />

resolves https://github.com/flutter/flutter/issues/140685

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[...]

PR #6267, opened 2024-03-05
[local_auth_darwin] macOS Support 

alexrabin-sentracam:
Adds macOS support for local_auth_darwin

Screenshot 2024-03-05 at 8 30 35 AM

Screenshot 2024-03-05 at 8 30 56 AM

## Cancelled Example:

Screenshot 2024-03-05 at 8 31 12 AM

## Success Example

Screenshot 2024-03-05 at 8 31 32 AM

## Error Example
Screenshot 2024-03-05 at 4 01 58 PM

https://github.com/flutter/flutter/issues/140685

## Pre-launch Checklist

- [X] I signed the [CLA].


alexrabin-sentracam:
@stuartmorgan That pr doesn't seem to have any momentum and is outdated so I figured I would take this on.

packages/local_auth/local_auth_darwin/CHANGELOG.md
```diff
@@ -1,12 +1,16 @@
+## 2.0.0
+
+- Adds MacOS Support. Requires MacOS 10.15 or higher
+
 ## 1.2.2
 
-* Adds compatibility with `intl` 0.19.0.
+- Adds compatibility with `intl` 0.19.0.
```
stuartmorgan, line 7: Please do not reformat the file.

packages/local_auth/local_auth_darwin/CHANGELOG.md
```diff
@@ -1,12 +1,16 @@
+## 2.0.0
+
+- Adds MacOS Support. Requires MacOS 10.15 or higher
```
stuartmorgan, line 3: macOS

stuartmorgan, line 3: Where does the 10.15 requirement come from?

stuartmorgan, line 3: "Support" is not a title or proper noun, so should not be capitalized.

packages/local_auth/local_auth_darwin/README.md
```diff
@@ -1,6 +1,6 @@
 # local_auth_darwin
 
-The iOS implementation of [`local_auth`][1].
+The iOS and MacOS implementation of [`local_auth`][1].
```
stuartmorgan, line 3: macOS

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -153,6 +153,26 @@ - (void)showAlertWithMessage:(NSString *)message
...
+  NSAlert *alert = [[NSAlert alloc] init];
+  [alert setMessageText:message];
+  [alert addButtonWithTitle:dismissButtonTitle];
+  if (openSettingsButtonTitle != nil) {
+    [alert addButtonWithTitle:openSettingsButtonTitle];
+  }
+  [alert beginSheetModalForWindow:NSApp.keyWindow
+                completionHandler:^(NSModalResponse returnCode) {
+                  if (returnCode == NSAlertSecondButtonReturn) {
+                    NSURL *url = [NSURL URLWithString:@"x-apple.systempreferences:com.apple."
```
stuartmorgan, line 166: https://github.com/flutter/packages/pull/5766/files#r1486380427

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -153,6 +153,26 @@ - (void)showAlertWithMessage:(NSString *)message
...
      openSettingsButtonTitle:(NSString *)openSettingsButtonTitle
                   completion:(FLADAuthCompletion)completion {
+#if TARGET_OS_OSX
+  NSAlert *alert = [[NSAlert alloc] init];
+  [alert setMessageText:message];
+  [alert addButtonWithTitle:dismissButtonTitle];
+  if (openSettingsButtonTitle != nil) {
+    [alert addButtonWithTitle:openSettingsButtonTitle];
+  }
+  [alert beginSheetModalForWindow:NSApp.keyWindow
```
stuartmorgan, line 163: See https://github.com/flutter/packages/pull/5766/files#r1486389711 for discussion of the window that should be used; there is no guarantee that the Flutter content is in the key window.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -257,12 +279,14 @@ - (void)handleError:(NSError *)authError
 
 #pragma mark - AppDelegate
 
+#if TARGET_OS_IOS
```
stuartmorgan, line 282: This should have a comment explaining why it's not relevant for macOS.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin.podspec
```diff
@@ -16,8 +16,10 @@ Downloaded by pub (not CocoaPods).
   s.documentation_url = 'https://pub.dev/packages/local_auth_darwin'
   s.source_files = 'Classes/**/*'
   s.public_header_files = 'Classes/**/*.h'
-  s.dependency 'Flutter'
+  s.ios.dependency 'Flutter'
+  s.osx.dependency 'FlutterMacOS'
   s.platform = :ios, '12.0'
```
stuartmorgan, line 21: This should be updated to use the same format as the new line below it.

packages/local_auth/local_auth_darwin/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -227,7 +227,7 @@
 		97C146E61CF9000F007C117D /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastUpgradeCheck = 1510;
+				LastUpgradeCheck = 1430;
```
stuartmorgan, line 230: Please don't downgrade the iOS project metadata.

packages/local_auth/local_auth_darwin/example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "1510"
+   LastUpgradeVersion = "1430"
```
stuartmorgan, line 3: Same.

packages/local_auth/local_auth_darwin/lib/types/auth_messages_darwin.dart
```diff
@@ -6,12 +6,12 @@ import 'package:flutter/foundation.dart';
 import 'package:intl/intl.dart';
 import 'package:local_auth_platform_interface/types/auth_messages.dart';
 
-/// Class wrapping all authentication messages needed on iOS.
+/// Class wrapping all authentication messages needed on iOS & MacOS.
 /// Provides default values for all messages.
 @immutable
-class IOSAuthMessages extends AuthMessages {
+class DarwinAuthMessages extends AuthMessages {
```
stuartmorgan, line 12: See https://github.com/flutter/packages/pull/5766#issuecomment-1878641342 and https://github.com/flutter/packages/pull/5766#discussion_r1486398912

packages/local_auth/local_auth_darwin/example/macos/RunnerTests/RunnerTests.swift
stuartmorgan, line 1: This file should not exist; macOS should be running the same native unit tests as iOS is, sharing the file in `darwin/Tests/`.

packages/local_auth/local_auth_darwin/lib/types/auth_messages_darwin.dart
stuartmorgan: (Also, renaming public classes is a breaking change, which this doesn't otherwise need to be.)

packages/local_auth/local_auth_darwin/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: local_auth_darwin
 description: iOS implementation of the local_auth plugin.
 repository: https://github.com/flutter/packages/tree/main/packages/local_auth/local_auth_darwin
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+local_auth%22
-version: 1.2.2
+version: 2.0.0
```
stuartmorgan, line 5: Once my comments about strings are resolved, this should be able to be 1.3.0.

stuartmorgan (changes requested):

packages/local_auth/local_auth_darwin/CHANGELOG.md
alexrabin-sentracam: Oops, my prettier on my vs code must have done that automatically. Will fix

packages/local_auth/local_auth_darwin/CHANGELOG.md
alexrabin-sentracam: Since we are using LABiometryTypeFaceID that requires macOS 10.15 or higher https://developer.apple.com/documentation/localauthentication/labiometrytype/labiometrytypefaceid?language=objc

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin.podspec
alexrabin-sentracam: What do you mean?

stuartmorgan: `s.ios.deployment_target = ...`

packages/local_auth/local_auth_darwin/CHANGELOG.md
stuartmorgan: Why not just not support face ID when running on macOS 10.14?

alexrabin-sentracam: The LABiometryTypeFaceID enum value is only available on macOS 10.15 or higher. We can't use this value on macOS 10.14

packages/local_auth/local_auth_darwin/lib/types/auth_messages_darwin.dart
alexrabin-sentracam: I put back iOSAuthMessages and created MacOSAuthMessages

packages/local_auth/local_auth_darwin/CHANGELOG.md
alexrabin-sentracam: Fixed

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
alexrabin-sentracam: I am not seeing how to get the current window with the registrar. Can you link me to some documentation please?

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
alexrabin-sentracam: https://gist.github.com/rmcdongit/f66ff91e0dad78d4d6346a75ded4b751

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
alexrabin-sentracam: Nvm I found it

alexrabin-sentracam: NSWindow* keyWindow = self.registrar.view.window;

alexrabin-sentracam:
@stuartmorgan  I have addressed all your comments

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -130,12 +132,31 @@ - (nullable NSNumber *)deviceCanSupportBiometricsWithError:
...
+          return biometrics;
+        }
+      }
+      if (context.biometryType == LABiometryTypeTouchID) {
+        [biometrics
+            addObject:[FLADAuthBiometricWrapper makeWithValue:FLADAuthBiometricFingerprint]];
+      }
+#endif
+
+#if TARGET_OS_IOS
```
stuartmorgan, line 152: Why is all of this code duplicated and ifdef'd rather than the `@available` check including iOS?

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -130,12 +132,31 @@ - (nullable NSNumber *)deviceCanSupportBiometricsWithError:
   if ([context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics
                            error:&authError]) {
     if (authError == nil) {
+#if TARGET_OS_OSX
+      // @available(macOS 10.15, *) is not a BOOL expression.
```
stuartmorgan, line 136: This comment doesn't seem necessary; this is standard language behavior.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
stuartmorgan: You've marked this as resolved, but I don't see a code comment in the current version, as requested by the linked review.

More importantly though, what the link you provided appears to be describing is a reverse-engineered system that uses private keys and has a proven track record of fragility to OS updates. That means that:
- there's no guarantee that Apple would consider this app-store compliant, and
- it is very likely to break without warning in the future, leaving a button that doesn't actually work.

Am I misunderstanding the discussion there? If not, then this needs a different solution for macOS that is more robust even if it's not as featureful.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -49,13 +49,15 @@ - (instancetype)initWithOptions:(nonnull FLADAuthOptions *)options
...
 @property(nonatomic, strong) NSObject<FLADAuthContextFactory> *authContextFactory;
+@property(nonatomic, strong) NSObject<FlutterPluginRegistrar> *registrar;
 @end
 
 @implementation FLALocalAuthPlugin
 
 + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
   FLALocalAuthPlugin *instance = [[FLALocalAuthPlugin alloc] init];
   [registrar addApplicationDelegate:instance];
+  instance.registrar = registrar;
```
stuartmorgan, line 60: This should be an init parameter, as it's both required and unchanging.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -153,6 +174,27 @@ - (void)showAlertWithMessage:(NSString *)message
...
      openSettingsButtonTitle:(NSString *)openSettingsButtonTitle
                   completion:(FLADAuthCompletion)completion {
+#if TARGET_OS_OSX
+  NSAlert *alert = [[NSAlert alloc] init];
+  [alert setMessageText:message];
+  [alert addButtonWithTitle:dismissButtonTitle];
+  if (openSettingsButtonTitle != nil) {
+    [alert addButtonWithTitle:openSettingsButtonTitle];
+  }
+  NSWindow *keyWindow = self.registrar.view.window;
```
stuartmorgan, line 184: "Key window" is a specific term with a defined meaning in macOS development; this is not the key window. It's the window containing the Flutter content.

There's no ambiguity here though, so the variable can just be called `window`.

packages/local_auth/local_auth_darwin/darwin/Classes/messages.g.m
```diff
@@ -41,7 +41,7 @@ - (instancetype)initWithValue:(FLADAuthResult)value {
 }
 @end
 
-/// Pigeon equivalent of the subset of BiometricType used by iOS.
+/// Pigeon equivalent of the subset of BiometricType used by iOS & macOS.
```
stuartmorgan, line 44: Please see line 4 of this file; you are hand-editing a generated file rather than the source of generation.

stuartmorgan (changes requested):
Based on the CI failure message I suspect you didn't add any source files to the RunnerTest target's build.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
alexrabin-sentracam: When I try to combine both availability statements I get this error:
```
 @available does not guard availability here; use if (@available) instead [-Wunsupported-availability-guard]
       if (@available(macOS 10.15, *) || @available(iOS 11.0, *)) {
```

That was why I put the comment saying @available is not a BOOl expression and is why I had to separate the if statements. What would you suggest we do then?

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
alexrabin-sentracam: Fixed

packages/local_auth/local_auth_darwin/darwin/Classes/messages.g.m
alexrabin-sentracam: Reverted

stuartmorgan:
> > Based on the CI failure message I suspect you didn't add any source files to the RunnerTest target's build.
> 
> I had to remove them since the CI said there were duplicate tests from the iOS project

This won't pass CI without native unit tests. You'll need to provide an exact error message you were getting from having them, since that description doesn't make sense to me. We have this pattern in a number of other plugins.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
alexrabin-sentracam: What do you suggest then in order to open the settings app? Here is another source with all the url schemes for settings: https://www.mbsplugins.de/archive/2020-04-05/MacOS_System_Preference_Links

alexrabin-sentracam:
> This won't pass CI without native unit tests. You'll need to provide an exact error message you were getting from having them, since that description doesn't make sense to me. We have this pattern in a number of other plugins.

Okay I'll put back the tests then

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
alexrabin-sentracam: Nvm fixed: https://forums.developer.apple.com/forums/thread/5237

alexrabin-sentracam:
@stuartmorgan I found this reference for all the url schemes that Apple has within the settings app: https://www.mbsplugins.de/archive/2020-04-05/MacOS_System_Preference_Links. If theses are up to par with you then what do you suggest we do for the settings button?

stuartmorgan:
> @stuartmorgan I found this reference for all the url schemes that Apple has within the settings app: https://www.mbsplugins.de/archive/2020-04-05/MacOS_System_Preference_Links.

That's not official documentation from Apple, so it doesn't address my earlier comments.

> If theses aren't up to par with you then what do you suggest we do for the settings button?

If you are not able to find a supported way to open settings programmatically, then I wouldn't attempt to provide a button that claims that it will do that, and would instead provide a high-level explanation of what the user needs to do.

stuartmorgan:
It still looks from the error message like the RunnerTests target is empty. Are you sure the file is added to that target? Are you able to run those tests in Xcode locally?

alexrabin-sentracam:
@stuartmorgan I found an article talking about how to find all the available app link URL's that Apple Supports and the plist keys that are needed to support them: https://www.linkedin.com/pulse/creating-web-links-launch-mac-apps-preferences-script-nick-tong

`sudo grep -A1 -r "NSPrefPaneAllowsXAppleSystemPreferencesURLScheme" /System/Library/PreferencePanes`

```
<key>NSPrefPaneAllowsXAppleSystemPreferencesURLScheme</key>
<true/>
```

I updated the Readme accordingly.

I also added the Tests for macOS.

stuartmorgan:
> @stuartmorgan I found an article talking about how to find all the available app link URL's that Apple Supports and the plist keys that are needed to support them: https://www.linkedin.com/pulse/creating-web-links-launch-mac-apps-preferences-script-nick-tong

Please read my comments above; a random person posting an article on LinkedIn is not the same as official documentation from Apple.

> I updated the Readme accordingly.

Adding an undocumented key that is used *for preference panes* to an application is unlikely to do anything, so is not something we would instruct people to do.

> I also added the Tests for macOS.

Based on the CI error, your change to the project appears to have broken the ability to build the example app.

alexrabin-sentracam:
> > @stuartmorgan I found an article talking about how to find all the available app link URL's that Apple Supports and the plist keys that are needed to support them: https://www.linkedin.com/pulse/creating-web-links-launch-mac-apps-preferences-script-nick-tong
> 
> Please read my comments above; a random person posting an article on LinkedIn is not the same as official documentation from Apple.
> 
> > I updated the Readme accordingly.
> 
> Adding an undocumented key that is used _for preference panes_ to an application is unlikely to do anything, so is not something we would instruct people to do.

Okay I am going to remove the settings button in the pop up and just update the messaging to tell the user to go to the settings app manually.

alexrabin-sentracam:
@stuartmorgan @LouiseHsu is there anything else needed before this pr can be merged?

stuartmorgan:
Please see https://github.com/flutter/flutter/wiki/Tree-hygiene#tldr for expected review times. It's been less than a full business day since you posted the last version.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -71,6 +72,14 @@ - (instancetype)initWithContextFactory:(NSObject<FLADAuthContextFactory> *)facto
   return self;
 }
 
+- (instancetype)initWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
```
stuartmorgan, line 75: There should not be any codepaths that don't set the registrar. This class should follow standard Obj-C convention of having an init method that has every possible parameter, and any necessary convenience inits should delegate to that one.

I would not expect the `initWithContextFactory:` to be necessary any more, since tests should always be specifying all arguments, so you should be able to just add a parameter to that method.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -153,6 +166,19 @@ - (void)showAlertWithMessage:(NSString *)message
...
+  [alert addButtonWithTitle:dismissButtonTitle];
+  NSWindow *window = self.registrar.view.window;
+  [alert beginSheetModalForWindow:window
+                completionHandler:^(NSModalResponse returnCode) {
+                  [self handleSucceeded:NO withCompletion:completion];
+                }];
+  return;
+#endif
+
+#if TARGET_OS_IOS
```
stuartmorgan, line 181: These are mutually exclusive blocks, so should use an `#if/#else` construction, not two separate `#if`s.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.m
stuartmorgan, line 1: There's no testing of the alert construction, or that it's using the registry. There should be a test that validates that the alert uses the window of the view provided by the registry, and that the expected strings are used in the alert. That will mean adding an injectable alert factory, similar to the existing auth factory.

packages/local_auth/local_auth_darwin/lib/local_auth_darwin.dart
```diff
@@ -2,19 +2,21 @@
...
 import 'types/auth_messages_ios.dart';
+import 'types/auth_messages_macos.dart';
 
 export 'package:local_auth_darwin/types/auth_messages_ios.dart';
 export 'package:local_auth_platform_interface/types/auth_messages.dart';
 export 'package:local_auth_platform_interface/types/auth_options.dart';
 export 'package:local_auth_platform_interface/types/biometric_type.dart';
 
-/// The implementation of [LocalAuthPlatform] for iOS.
+/// The implementation of [LocalAuthPlatform] for iOS & macOS.
```
stuartmorgan, line 19: Please use "and" instead of "&", to match our standard style.

packages/local_auth/local_auth_darwin/lib/local_auth_darwin.dart
```diff
@@ -93,11 +97,11 @@ class LocalAuthDarwin extends LocalAuthPlatform {
   @override
   Future<bool> isDeviceSupported() async => _api.isDeviceSupported();
 
-  /// Always returns false as this method is not supported on iOS.
+  /// Always returns false as this method is not supported on iOS & macOS.
```
stuartmorgan, line 100: "iOS or macOS"

packages/local_auth/local_auth_darwin/lib/local_auth_darwin.dart
```diff
@@ -118,4 +122,26 @@ class LocalAuthDarwin extends LocalAuthPlatform {
...
+        break;
+      }
+    }
+    return AuthStrings(
+      reason: localizedReason,
+      lockOut: messages?.lockOut ?? macOSLockOut,
+      goToSettingsButton: messages?.goToSettingsButton ?? goToSettings,
+      goToSettingsDescription:
+          messages?.goToSettingsDescription ?? macOSGoToSettingsDescription,
+      // TODO(stuartmorgan): The default's name is confusing here for legacy
```
stuartmorgan, line 141: What legacy reasons require a new constant to have a confusing name?

packages/local_auth/local_auth_darwin/lib/src/messages.g.dart
```diff
@@ -36,15 +36,16 @@ enum AuthResult {
   errorPasscodeNotSet,
 }
 
-/// Pigeon equivalent of the subset of BiometricType used by iOS.
+/// Pigeon equivalent of the subset of BiometricType used by iOS & MacOS.
```
stuartmorgan, line 39: "and"

Same throughout this file (and the rest of the PR if there are other instances).

packages/local_auth/local_auth_darwin/lib/src/messages.g.dart
```diff
@@ -36,15 +36,16 @@ enum AuthResult {
...
 enum AuthBiometric {
+  /// Face is only used on iOS
```
stuartmorgan, line 41: This is missing a period.

packages/local_auth/local_auth_darwin/lib/src/messages.g.dart
stuartmorgan, line 39: macOS

packages/local_auth/local_auth_darwin/lib/types/auth_messages_macos.dart
```diff
@@ -0,0 +1,107 @@
...
+import 'package:local_auth_platform_interface/types/auth_messages.dart';
+
+/// Class wrapping all authentication messages needed on macOS.
+/// Provides default values for all messages.
+@immutable
+class MacOSAuthMessages extends AuthMessages {
+  /// Constructs a new instance.
+  const MacOSAuthMessages({
+    this.lockOut,
+    this.goToSettingsButton,
```
stuartmorgan, line 16: This is no longer used.

packages/local_auth/local_auth_darwin/lib/types/auth_messages_macos.dart
```diff
@@ -0,0 +1,107 @@
...
+/// from the current dialog.
+String get macOSGoToSettings => Intl.message('Go to settings',
+    desc: 'Message shown on a button that the user can click to go to '
+        'settings pages from the current dialog. Maximum 30 characters.');
+
+/// Message advising the user to re-enable biometrics on their device.
+/// It shows in a dialog on macOS.
+String get macOSLockOut => Intl.message(
+    'Biometric authentication is disabled. Please lock and unlock your screen to '
+    'enable it.',
```
stuartmorgan, line 91: Are these instructions correct for macOS?

packages/local_auth/local_auth_darwin/lib/local_auth_darwin.dart
```diff
@@ -40,7 +42,9 @@ class LocalAuthDarwin extends LocalAuthPlatform {
             biometricOnly: options.biometricOnly,
             sticky: options.stickyAuth,
             useErrorDialogs: options.useErrorDialogs),
-        _pigeonStringsFromAuthMessages(localizedReason, authMessages));
+        defaultTargetPlatform == TargetPlatform.macOS
```
stuartmorgan, line 45: This should be using `Platform.is*` (or a local-test-overridable wrapper around that), since it's interfacing with native code; if non-test code is executing this code with the default target overridden, it still needs to use the strings that correspond to the actual native code that will execute.

stuartmorgan (changes requested):

packages/local_auth/local_auth_darwin/lib/types/auth_messages_macos.dart
stuartmorgan: Sending a string from Dart to the host isn't using the string in a meaningful sense, it's just passing it around. The reason these strings exist is to provide them to native code in order to display them to the user in a native dialog. There is no longer any code path that results in this string being displayed to the user, so it's dead code.

alexrabin-sentracam: If I remove MacOSAuthMessages, should I rename iOSAuthMessages to DarwinAuthMessages?

stuartmorgan: > If I remove MacOSAuthMessages

You should not remove `MacOSAuthMessages`. You should remove the parts of it that are never used.

> should I rename iOSAuthMessages to DarwinAuthMessages?

No; as discussed [here](https://github.com/flutter/packages/pull/6267#discussion_r1546793569) they are not the same. That would also re-create a different version of the same problem: a public API surface that allows setting some values for macOS that will never be used on macOS.

packages/local_auth/local_auth_darwin/lib/local_auth_darwin.dart
alexrabin-sentracam: That was a comment that was copied from auth_messages_ios

stuartmorgan: I'm aware of that. I wrote that comment when writing new code that used an existing constant that we couldn't change without breaking clients.

That context doesn't apply to a completely new constant, so I don't understand why the TODO and the confusing name were copied into this file.

jmagman:
@LouiseHsu could you take a look at this one?

stuartmorgan:
Is this ready for re-review? There are still unaddressed comments and failing tests, so I assumed it was a work-in-progress.

alexrabin-sentracam:
I could use some assistance on the comments @stuartmorgan left me. I am not seeing a clean way to use the Platform.isMacos and still have the dart test cases pass. 

I am also not well versed in testing the alert construction as you requested.

stuartmorgan:
> I am not seeing a clean way to use the Platform.isMacos and still have the dart test cases pass.

Per my earlier comment, you can wrap those calls in something that can have an alternate implementation injected for unit tests, using standard dependency injection patterns.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.m
alexrabin-sentracam: Added alert construction test

alexrabin-sentracam:
> > I am not seeing a clean way to use the Platform.isMacos and still have the dart test cases pass.
> 
> Per my earlier comment, you can wrap those calls in something that can have an alternate implementation injected for unit tests, using standard dependency injection patterns.

All fixed. I addressed all your comments @stuartmorgan . This pr is ready for re-review.

stuartmorgan:
@alexrabin-sentracam I'll re-review when I have time (I currently have some very high priority work that will delay it); it's not necessary to constantly merge the branch in the meantime.

alexrabin-sentracam:
@stuartmorgan Okay sounds good. Thank you.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin.m
```diff
@@ -257,12 +302,15 @@ - (void)handleError:(NSError *)authError
 
 #pragma mark - AppDelegate
 
+// This method is called when the app is resumed from the background only on iOS
```
stuartmorgan, line 305: This is missing a period.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin_Test.h
```diff
@@ -2,16 +2,33 @@
...
+#endif
+
 #import <LocalAuthentication/LocalAuthentication.h>
 
 /// Protocol for a source of LAContext instances. Used to allow context injection in unit tests.
 @protocol FLADAuthContextFactory <NSObject>
 - (LAContext *)createAuthContext;
 @end
 
+@protocol FLADAlertFactory <NSObject>
```
stuartmorgan, line 18: This needs a declaration comment, per the style guide.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin_Test.h
```diff
@@ -2,16 +2,33 @@
...
+
+#if TARGET_OS_OSX
+- (NSAlert *)createNSAlert;
```
stuartmorgan, line 21: As do the methods.

packages/local_auth/local_auth_darwin/darwin/Classes/FLALocalAuthPlugin_Test.h
```diff
@@ -2,16 +2,33 @@
...
+#elif TARGET_OS_IOS
+- (UIAlertController *)createUIAlert:(NSString *)message;
+#endif
+
+@end
+
 @interface FLALocalAuthPlugin ()
 /// Returns an instance that uses the given factory to create LAContexts.
```
stuartmorgan, line 29: This comment needs to be updated to reflect the changes.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.m
```diff
@@ -49,11 +92,30 @@ - (void)setUp {
   self.continueAfterFailure = NO;
 }
 
+- (StubAlertFactory *)createStubAlertFactory {
```
stuartmorgan, line 95: This needs a declaration comment, per style guide.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.m
```diff
@@ -38,6 +38,49 @@ - (LAContext *)createAuthContext {
 
 @end
 
+@interface StubAlertFactory : NSObject <FLADAlertFactory>
+#if TARGET_OS_OSX
+@property(copy, nonatomic) NSAlert *alert;
```
stuartmorgan, line 43: Why are these properties `copy`?

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.m
```diff
@@ -38,6 +38,49 @@ - (LAContext *)createAuthContext {
...
+@end
+
+@implementation StubAlertFactory
+
+#if TARGET_OS_OSX
+
+- (instancetype)initWithNSAlert:(NSAlert *)alert {
+  self = [super init];
+  if (self) {
+    _alert = alert;
```
stuartmorgan, line 59: These aren't doing copies, which violates the current API contract. (This will be resolved if you don't declare them as copy though.)

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.m
```diff
@@ -166,12 +237,70 @@ - (void)testFailedAuthWithBiometrics {
...
+                                                                  sticky:NO
+                                                         useErrorDialogs:YES]
+                          strings:strings
+                       completion:^(FLADAuthResultDetails *_Nullable resultDetails,
+                                    FlutterError *_Nullable error) {
+                         XCTAssertTrue([NSThread isMainThread]);
+                         // Tests that the alert window is the same as the registrar window
+                         XCTAssertEqual(alertFactory.alert.window, reg.view.window);
+                         // TODO(stuartmorgan): Fix this; this was the pre-Pigeon-migration
+                         // behavior, so is preserved as part of the migration, but a failed
```
stuartmorgan, line 282: How can an entirely new platform implementation have previous behavior that needs to be preserved?

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.m
```diff
@@ -49,11 +92,30 @@ - (void)setUp {
   self.continueAfterFailure = NO;
 }
 
+- (StubAlertFactory *)createStubAlertFactory {
+  StubAlertFactory *alertFactory;
+#if TARGET_OS_OSX
+
+  id mockNSAlert = OCMClassMock([NSAlert class]);
```
stuartmorgan, line 99: Why is this a mock? The only thing that this appears to be used for in the test is to read a public property back out of it, which can be done with a real object, avoiding the need for a mock.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.m
```diff
@@ -166,12 +237,70 @@ - (void)testFailedAuthWithBiometrics {
...
+  XCTestExpectation *expectation = [self expectationWithDescription:@"Result is called"];
+  [plugin authenticateWithOptions:[FLADAuthOptions makeWithBiometricOnly:YES
+                                                                  sticky:NO
+                                                         useErrorDialogs:YES]
+                          strings:strings
+                       completion:^(FLADAuthResultDetails *_Nullable resultDetails,
+                                    FlutterError *_Nullable error) {
+                         XCTAssertTrue([NSThread isMainThread]);
+                         // Tests that the alert window is the same as the registrar window
+                         XCTAssertEqual(alertFactory.alert.window, reg.view.window);
```
stuartmorgan, line 280: I would expect there to also be a test that if the user provides a custom message, the alert has that message.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.m
```diff
@@ -166,12 +237,70 @@ - (void)testFailedAuthWithBiometrics {
...
+                         // authentication should return failure, not an error that results in a
+                         // PlatformException.
+                         XCTAssertEqual(resultDetails.result, FLADAuthResultErrorNotAvailable);
+                         XCTAssertNil(error);
+                         [expectation fulfill];
+                       }];
+  [self waitForExpectationsWithTimeout:kTimeout handler:nil];
+}
+
+#endif
```
stuartmorgan, line 292: You've changed the iOS codepaths to make the alert testable; there needs to be at least one new test covering those changes.

stuartmorgan (changes requested):

jmagman:
@alexrabin-sentracam looks like there's some merge conflicts, let us know when this is ready to re-review!

jmagman:
@LouiseHsu would you mind taking a look at this?

stuartmorgan:
Is this actually ready for re-review? There are still open comments from the last review, so I was assuming it was not yet ready.

alexrabin-sentracam:
@stuartmorgan I am seeing this error in the LUCI after the tests were converted to swift. 

> xcodebuild: error: Failed to build workspace Runner with scheme Runner.: Cannot test target “RunnerTests” on “My Mac”: My Mac’s macOS 13.6.1 doesn’t match RunnerTests’s macOS 14.0 deployment target.
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8746482769422894929/+/u/Run_package_tests/native_test/stdout?format=raw

Do you have any recommendations for what I should do?

stuartmorgan:
> @stuartmorgan I am seeing this error in the LUCI after the tests were converted to swift.

Why do you believe the error is related to the tests being converted to Swift? You changed the `MACOSX_DEPLOYMENT_TARGET` in [this commit](https://github.com/flutter/packages/pull/6267/commits/a91e4e912d9d7e33d3dc33a38600130481acdb76#diff-1464ddfcf50b83c234b3529d7d64d867e68def0e1e0b95c9a5ff3944f7aa1a37).

packages/local_auth/local_auth_darwin/darwin/Tests/OCSwiftMock.h
stuartmorgan, line 1: Please do not re-introduce OCMock; I removed it intentionally. New mocks/stubs should use the same pattern the other tests now use, which is to introduce a protocol that has the desired methods, and then have tests implement that protocol. See [here](https://github.com/flutter/packages/blob/decb46042bc5b000c0de27f2ff1ab4beeaa7b921/packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/include/local_auth_darwin/FLALocalAuthPlugin_Test.h#L10-L20), [here](https://github.com/flutter/packages/blob/main/packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/FLALocalAuthPlugin.m#L11-L56), and [here](https://github.com/flutter/packages/blob/decb46042bc5b000c0de27f2ff1ab4beeaa7b921/packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift#L26-L70) for an example of the pattern. (I would expect the new wrappers you would need to be simpler than that one.)

alexrabin-sentracam:
> @alexrabin-sentracam do you need some help with the OCMock bit?

@LouiseHsu Yes please, I'd appreciate that

stuartmorgan:
> > @alexrabin-sentracam do you need some help with the OCMock bit?
> 
> @LouiseHsu Yes please, I'd appreciate that

I've pushed changes that restructure to avoid the need for OCMock. @alexrabin-sentracam some follow-up work is still needed:
- The changes to actually add OCMock to the project need to be undone; I didn't do any project changes.
- I did not have to change any tests as part of the work, which suggests that none of the tests are actually checking anything about the interactions with `NSAlert`/`UIAlertController` to make sure they are as expected. That's an important part of the test coverage. The structure I set up will allow for querying that information, but I didn't actually add the tests.

stuartmorgan:
From triage: @alexrabin-sentracam Are you still planning on updating this now that the OCMock removal has been handled?

alexrabin-sentracam:
@stuartmorgan I've been assigned higher priorities at my job. I don't know when I'll be able to take a look at this 😕

stuartmorgan (approved):
I've done the remaining cleanup, and added basic testing of the dialog display for both macOS and iOS.

@LouiseHsu this is ready for final review.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -310,14 +324,53 @@ class FLALocalAuthPluginTests: XCTestCase {
...
+    plugin.authenticate(
+      with: FLADAuthOptions.make(
+        withBiometricOnly: false,
+        sticky: false,
+        useErrorDialogs: true),
+      strings: strings
+    ) { resultDetails, error in
+      #if os(macOS)
+        expectation.fulfill()
+      #else
```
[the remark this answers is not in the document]
stuartmorgan: Good point; reworked to be a TODO comment above the `#if`, and removed the `#else`.

stuartmorgan:
(That last failure was just the current macOS host flake we're having, unlike the previous ones.)

PR #6274, opened 2024-03-06
[webview_flutter_wkwebview] Fixes JSON.stringify() cannot serialize cyclic structures

LinXunFeng:
Using the `replacer` parameter of `JSON.stringify()` to remove cyclic object to resolve the following error.

```
TypeError: JSON.stringify cannot serialize cyclic structures.
```

~~Related solution: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value~~

Fixes https://github.com/flutter/flutter/issues/144535

## Pre-launch Checklist

- [x] I signed the [CLA].


jmagman:
This is in wkwebview but maybe @ditman could review the js code?

ditman:
Can do, thanks for the mention @jmagman, I normally disregard webview stuff because I end up flagged as an owner (but I'm not :P)

bparrishMines:
I don't think we want to incorporate this into the plugin since this is an intentional error from the platform and I wouldn't consider serializing cyclical objects as a core feature of using a WebView. It also seems like something that could be handled by the application. Although we do have other scripts for achieving platform parity (e.g. disabling zoom on iOS is not possible without injecting a script). 

@ditman @stuartmorgan What are your thoughts on this since you have more experience with web development?

ditman:
The biggest problem I see with this is that this is a "lossy" fix; what you deserialize is *not* what you were serializing; `"[Cyclic]"` might be good for the json encoding to not crash, but it doesn't tell you how to recover that cycle when you parse JSON back to an object?

(Also if the cycle is not important: remove it before attempting to serialize to JSON?)

bparrishMines (approved):
LGTM

I didn't originally realize this was preventing an error from a method that was already a part of the script we were already injecting. I think this solution looks good.

cc @ditman or @stuartmorgan for secondary review.

bparrishMines:
> The biggest problem I see with this is that this is a "lossy" fix; what you deserialize is _not_ what you were serializing; `"[Cyclic]"` might be good for the json encoding to not crash, but it doesn't tell you how to recover that cycle when you parse JSON back to an object?
> 
> (Also if the cycle is not important: maybe remove it before attempting to serialize to JSON?)

Yea, I was considering this as well. But do think it is reasonable for a user to expect that `console.log` should be able to handle any object? Does the problem lie with using `JSON.stringify()` instead of just using `String()`? I would consider calling `console.log` as the equivalent of calling `print(...)` in a Flutter app and would be surprised if this caused an error. It might be a breaking change to move away from `JSON.stringify()` though, so that was my reasoning for avoiding the crash as the best solution right now.

bparrishMines:
Looking at the issue and the Android output, it looks like Android basically just calls [String()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/String) on an object passed to the console. Android just returns `[object Object]` when passed a recursive object.

ditman:
Ah! I didn't realize this was to implement `console.log`. In that case, you're probably fine. This works in the browser because the browser doesn't attempt to "toString" the whole thing, it just gives you a tree-like structure for the object:

Screenshot 2024-03-08 at 12 11 24 PM

(You get tired of expanding `self` before the browser runs out of memory, I think :P)

The PR is a C+P from MDN however; @stuartmorgan licensing concerns?

stuartmorgan:
> The PR is a C+P from MDN however; @stuartmorgan licensing concerns?

What is done in this PR [is not permitted in Flutter](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#code-that-is-not-copyrighted-the-flutter-authors).

This code is not complicated enough to warrant us adding a `third_party` directory to this plugin; we should get a clean implementation of the webkit_webview_controller.dart (by giving someone who has not seen this code just the unit test and a high-level description of the goal).

stuartmorgan (changes requested):
(Marking as changes needed to ensure this doesn't accidentally land as-is.)

LinXunFeng:
@stuartmorgan Thanks for your reminder.

I have adjusted the implementation to solve this issue by removing the cyclic object, please review the code again.

stuartmorgan:
Unless I'm missing something, the new version is destroying any duplicate objects, which is not the same thing as just removing cycles.

LinXunFeng:
@stuartmorgan I've re-adjusted the removal logic.

jmagman:
cc @hellohuanlin to review

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
```diff
@@ -630,11 +630,34 @@ class WebKitWebViewController extends PlatformWebViewController {
...
+    // Using the replacer parameter of JSON.stringify() to solve the error
+    // TypeError: JSON.stringify cannot serialize cyclic structures.
+    // See https://github.com/flutter/flutter/issues/144535.
+    //
+    // Considering this is just looking at the logs printed via console.log,
+    // the cyclic object is not important, so remove it.
     const WKUserScript overrideScript = WKUserScript(
       '''
+function removeCyclicObject() {
+  const levelObjects = [];
```
stuartmorgan, line 642: I'm not sure what this naming means; what is a "level object"?

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
```diff
@@ -630,11 +630,34 @@ class WebKitWebViewController extends PlatformWebViewController {
...
   Future<void> _injectConsoleOverride() {
+    // Using the replacer parameter of JSON.stringify() to solve the error
+    // TypeError: JSON.stringify cannot serialize cyclic structures.
+    // See https://github.com/flutter/flutter/issues/144535.
+    //
+    // Considering this is just looking at the logs printed via console.log,
+    // the cyclic object is not important, so remove it.
     const WKUserScript overrideScript = WKUserScript(
       '''
+function removeCyclicObject() {
```
stuartmorgan, line 641: This should have a name that cannot plausibly cause collisions on actual web pages, since it is injected into the page's context. E.g., maybe a Flutter prefix and a dynamically generated UUID? @ditman Any suggestions on best practice for avoiding naming collisions in JS?

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
```diff
@@ -630,11 +630,34 @@ class WebKitWebViewController extends PlatformWebViewController {
   }
 
   Future<void> _injectConsoleOverride() {
+    // Using the replacer parameter of JSON.stringify() to solve the error
+    // TypeError: JSON.stringify cannot serialize cyclic structures.
+    // See https://github.com/flutter/flutter/issues/144535.
+    //
+    // Considering this is just looking at the logs printed via console.log,
+    // the cyclic object is not important, so remove it.
     const WKUserScript overrideScript = WKUserScript(
```
stuartmorgan, line 639: Please add a comment briefly explaining the approach at a high level so readers don't need to reverse-engineer what is being done here.

stuartmorgan (changes requested):

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
ditman: @stuartmorgan At the expense of some performance this function could be defined inside of the `log` function that calls it (line 659)... But if we're not worried about a function `log` colliding with the functions of a webpage, I wouldn't worry too much about `removeCyclicObject` either :)

(In the engine we normally prefix with `_` the name of the function so there's a fewer change of a global collision, but there's no guarantee that it won't collide. For a UUID you'd have to strip the `-` characters for it to be a valid function name, but it could definitely work)

ditman: > We wouldn't need to do it in JS; we'd generate it in Dart, and then string interpolate that into the string we inject.

Ah you're right, it's "free" then. `_flutter_log_override` sounds like a good name. You can even have a small `_flutter_webview_plugin_overrides` "namespace" object that contains all the methods that you're injecting, so it's easy to hide/move them around.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
LinXunFeng: `levelObjects` is used to record objects whose properties are currently being traversed. (Maybe I should rename it to `stack` or something? )

As shown in the figure below, the properties of the outermost object (blue box) are currently being traversed, and obj1 (green box) is also an object, so the properties of obj1 will be traversed next and pushed in levelObjects.

image

When traversing all the properties of obj1 (green box) and starting to traverse obj2, currentParentObj will point to the outermost object (blue box) and the last element obj1 (green box) of the current `levelObjects` is not equal, then obj1 will be popped up, and obj2 will be pushed to levelObjects.

image

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
LinXunFeng: Done.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
LinXunFeng: Done.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
```diff
@@ -630,33 +630,65 @@ class WebKitWebViewController extends PlatformWebViewController {
...
-  var log = {
-    level: type,
-    message: message
-  };
-
-  window.webkit.messageHandlers.fltConsoleMessage.postMessage(JSON.stringify(log));
-}
+var _flutter_webview_plugin_overrides = _flutter_webview_plugin_overrides || {
+  removeCyclicObject: function() {
+    const levelObjects = [];
```
stuartmorgan, line 651: Based on the description in the PR comment, `traversalStack` would probably be a clearer name.

stuartmorgan (approved):
LGTM with one naming suggestion.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
LinXunFeng: Thx.

PR #6285, opened 2024-03-08
Replace deprecated UIGraphicsBeginImageContextWithOptions with UIGraphicsImageRenderer

jmagman:
Fixes https://github.com/flutter/flutter/issues/144602


packages/image_picker/image_picker_ios/ios/Classes/FLTImagePickerImageUtil.m
```diff
@@ -28,16 +28,36 @@ - (instancetype)initWithImages:(NSArray<UIImage *> *)images interval:(NSTimeInte
...
-  bool hasMaxWidth = maxWidth != nil;
-  bool hasMaxHeight = maxHeight != nil;
+  BOOL hasMaxWidth = maxWidth != nil;
+  BOOL hasMaxHeight = maxHeight != nil;
+
+  if ((originalWidth == maxWidth.doubleValue && originalHeight == maxHeight.doubleValue) ||
+      (!hasMaxWidth && !hasMaxHeight)) {
+    // Nothing to scale.
+    return image;
+  }
```
jmagman, line 58: This code is new, and unrelated to the fix.  If there is nothing to scale, return the original image instead of doing extra work.

packages/image_picker/image_picker_ios/example/ios/RunnerTests/ImageUtilTests.m
```diff
@@ -13,26 +13,82 @@ @interface ImageUtilTests : XCTestCase
...
+  CGFloat scaledHeight = 2;
+  UIImage *scaledImage = [FLTImagePickerImageUtil scaledImage:image
+                                                     maxWidth:@(scaledWidth)
+                                                    maxHeight:@(scaledHeight)
+                                          isMetadataAvailable:YES];
+  XCTAssertEqual(scaledImage.size.width, scaledWidth);
+  XCTAssertEqual(scaledImage.size.height, scaledHeight);
+
+  // Check the corners. Format is "R G B A"
+  XCTAssertEqualObjects(ColorStringAtPixel(scaledImage, 0, 0), @"1 0.776471 0 1");  // yellow
```
vashworth, line 63: Nit: How about making these color string representations consts with descriptive names so they can be reused and seem less like a magic value?

vashworth (approved):
LGTM with a small suggestion

packages/image_picker/image_picker_ios/example/ios/RunnerTests/ImageUtilTests.m
jmagman: I can do that for the two tests that are the same.  These exact values would changes based on how small the image is scaled.

PR #6286, opened 2024-03-08
[video_player_web] Add an alternate mode that renders images directly into the canvas

jezell:
This allows the web video player to avoid issues relating to having too many active overlaygroups. When this mode is enabled as follows:

```dart
VideoPlayerPlugin.renderVideoAsTexture = true;`
```

The video player will use createImageFromImageBitmap to grab frames as they are available and render them using RawImage instead.

Fixes:

https://github.com/flutter/flutter/issues/144591

## Pre-launch Checklist

- [ X ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ X ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ X] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [ X] I signed the [CLA].
- [ X ] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ X ] I [linked to at least one issue that this PR fixes] in the description above.
- [ X ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [ X ] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [ X ] I updated/added relevant documentation (doc comments with `///`).
- [ X ] All existing and new tests are passing.

### DIFF
--- a/packages/video_player/video_player_web/lib/src/browser_detection.dart
+++ b/packages/video_player/video_player_web/lib/src/browser_detection.dart
@@ -1,0 +1,83 @@
+import 'package:web/web.dart' as web;
+
+/// Browser detection copied from engine/browser_detection.dart
+
+/// The HTML engine used by the current browser.
+enum BrowserEngine {
+  /// The engine that powers Chrome, Samsung Internet Browser, UC Browser,
+  /// Microsoft Edge, Opera, and others.
+  ///
+  /// Blink is assumed in case when a more precise browser engine wasn't
+  /// detected.
+  blink,
+
+  /// The engine that powers Safari.
+  webkit,
+
+  /// The engine that powers Firefox.
+  firefox,
+}
+
+/// html webgl version qualifier constants.
+abstract class WebGLVersion {
+  /// WebGL 1.0 is based on OpenGL ES 2.0 / GLSL 1.00
+  static const int webgl1 = 1;
+
+  /// WebGL 2.0 is based on OpenGL ES 3.0 / GLSL 3.00
+  static const int webgl2 = 2;
+}
+
+/// Lazily initialized current browser engine.
+final BrowserEngine _browserEngine = _detectBrowserEngine();
+
+/// Override the value of [browserEngine].
+///
+/// Setting this to `null` lets [browserEngine] detect the browser that the
+/// app is running on.
+///
+/// This is intended to be used for testing and debugging only.
+BrowserEngine? debugBrowserEngineOverride;
+
+/// Returns the [BrowserEngine] used by the current browser.
+///
+/// This is used to implement browser-specific behavior.
+BrowserEngine get browserEngine {
+  return debugBrowserEngineOverride ?? _browserEngine;
+}
+
+BrowserEngine _detectBrowserEngine() {
+  final String vendor = web.window.navigator.vendor;
+  final String agent = web.window.navigator.userAgent.toLowerCase();
+  return detectBrowserEngineByVendorAgent(vendor, agent);
+}
+
+/// Detects browser engine for a given vendor and agent string.
+///
+/// Used for testing this library.
+BrowserEngine detectBrowserEngineByVendorAgent(String vendor, String agent) {
+  if (vendor == 'Google Inc.') {
+    return BrowserEngine.blink;
+  } else if (vendor == 'Apple Computer, Inc.') {
+    return BrowserEngine.webkit;
+  } else if (agent.contains('Edg/')) {
+    // Chromium based Microsoft Edge has `Edg` in the user-agent.
+    // https://docs.microsoft.com/en-us/microsoft-edge/web-platform/user-agent-string
+    return BrowserEngine.blink;
+  } else if (vendor == '' && agent.contains('firefox')) {
+    // An empty string means firefox:
+    // https://developer.mozilla.org/en-US/docs/Web/API/Navigator/vendor
+    return BrowserEngine.firefox;
+  }
+
+  // Assume Blink otherwise, but issue a warning.
+  // ignore: avoid_print
+  print(
+      'WARNING: failed to detect current browser engine. Assuming this is a Chromium-compatible browser.');
+  return BrowserEngine.blink;
+}
+
+/// Whether the current browser is Safari.
+bool get isSafari => browserEngine == BrowserEngine.webkit;
+
+/// Whether the current browser is Firefox.
+bool get isFirefox => browserEngine == BrowserEngine.firefox;
--- a/packages/video_player/video_player_web/lib/video_player_web.dart
+++ b/packages/video_player/video_player_web/lib/video_player_web.dart
@@ -14,8 +14,8 @@ import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:video_player_platform_interface/video_player_platform_interface.dart';
 import 'package:web/web.dart' as web;
 
-import 'src/video_player.dart';
 import 'src/browser_detection.dart';
+import 'src/video_player.dart';
 
 /// The web implementation of [VideoPlayerPlatform].
 ///
--- a/packages/video_player/video_player_web/lib/video_player_web.dart
+++ b/packages/video_player/video_player_web/lib/video_player_web.dart
@@ -4,7 +4,6 @@
 
 import 'dart:async';
 import 'dart:js_interop';
-import 'dart:js_interop_unsafe';
 import 'dart:js_util';
 import 'dart:math';
 import 'dart:ui' as ui;
@@ -212,8 +211,7 @@ class _AdapterVideoPlayerRendererState
   }
 
   bool get rendererCanvasKit {
-    final JSAny? r = web.window.getProperty('flutterCanvasKit'.toJS);
-    return r != null;
+    return hasProperty(web.window, 'flutterCanvasKit');
   }
 
   @override
--- a/packages/video_player/video_player_web/lib/video_player_web.dart
+++ b/packages/video_player/video_player_web/lib/video_player_web.dart
@@ -15,6 +15,7 @@ import 'package:video_player_platform_interface/video_player_platform_interface.
 import 'package:web/web.dart' as web;
 
 import 'src/video_player.dart';
+import 'src/browser_detection.dart';
 
 /// The web implementation of [VideoPlayerPlatform].
 ///
@@ -239,7 +240,7 @@ class _AdapterVideoPlayerRendererState
     final web.HTMLVideoElement element = widget.player.videoElement;
     isPlaying = !!(!element.paused && !element.ended && element.readyState > 2);
 
-    if (isPlaying) {
+    if (isPlaying && (isSafari || isFirefox)) {
       desiredMode = VideoRenderMode.canvas;
     } else {
       desiredMode = VideoRenderMode.texture;
--- a/packages/video_player/video_player_web/lib/video_player_web.dart
+++ b/packages/video_player/video_player_web/lib/video_player_web.dart
@@ -95,8 +95,11 @@ class VideoPlayerPlugin extends VideoPlayerPlatform {
       ..style.border = 'none'
       ..style.height = '100%'
       ..style.width = '100%'
-      ..style.pointerEvents = 'none'
-      ..crossOrigin = 'anonymous';
+      ..style.pointerEvents = 'none';
+
+    if (mode != VideoRenderMode.html) {
+      videoElement.crossOrigin = 'anonymous';
+    }
 
     return videoElement;
   }
--- a/packages/video_player/video_player_web/lib/video_player_web.dart
+++ b/packages/video_player/video_player_web/lib/video_player_web.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
 import 'dart:js_util';
 import 'dart:math';
 import 'dart:ui' as ui;
@@ -19,7 +20,12 @@ import 'src/video_player.dart';
 /// The web implementation of [VideoPlayerPlatform].
 ///
 /// This class implements the `package:video_player` functionality for the web.
-abstract class VideoElementPlayerPlugin extends VideoPlayerPlatform {
+class VideoPlayerPlugin extends VideoPlayerPlatform {
+  /// Registers this class as the default instance of [VideoPlayerPlatform].
+  static void registerWith(Registrar registrar) {
+    VideoPlayerPlatform.instance = VideoPlayerPlugin();
+  }
+
   // Map of textureId -> VideoPlayer instances
   final Map<int, VideoPlayer> _videoPlayers = <int, VideoPlayer>{};
 
@@ -71,7 +77,7 @@ abstract class VideoElementPlayerPlugin extends VideoPlayerPlatform {
             'web implementation of video_player cannot play content uri'));
     }
 
-    final web.HTMLVideoElement videoElement = createElement(textureId);
+    final web.HTMLVideoElement videoElement = _createElement(textureId);
 
     final VideoPlayer player = VideoPlayer(videoElement: videoElement)
       ..initialize(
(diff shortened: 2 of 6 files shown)


ditman:
Should this be a different implementation of `video_player_web`? Do you need both rendering modes at the same time, or do you always use the image-to-canvas mode? (Also, does this play sound?)

jezell:
@ditman I believe these rendering issues are unique to the web player, as the other platforms are already doing something like this to render the video into textures on their respective platforms, so making this the default way that videos work would bring the platforms a bit more inline with each other in terms of behavior. I don't know of a good reason to keep the old behavior around forever or as the default. Might actually be better if the default of the setting was true, or the old behavior was put into a legacy player package. That would be quite a bit of code duplication just for a toggle on how the rendering part happens though. At the end of the day all paths go through a video element. Maybe the renderer widget itself could be abstracted so packages could choose a rendering strategy instead of having a boolean toggle? Might be a little cleaner.

ditman:
@jezell what browser are you targeting? I'm guessing Safari?

This can't be the default method because `requestVideoFrameCallback` [doesn't seem to be supported at all on Firefox](https://caniuse.com/mdn-api_htmlvideoelement_requestvideoframecallback)?

I guess it can be made to fallback to the html element in older browsers, but how can this be more performant than the native tag?

jezell:
@ditman performance is secondary to the fact that overlay groups cause rendering to entirely stop working. 

Regardless, the performance issue isn't the native video element itself, it's that the native video element is inserted into an overlay group and the overlay groups cause issues. That's why the engine has a hard stop a a really low number. Nothing should ever depend on an overlay groups or an HtmlElementView in a framework intended to be used generically.  If overlay groups didn't exist, the performance would be that of the native element, but overlay groups do exist. 

To avoid the problems with overlay groups and use a native element is possible, but only if you avoid using HtmlElementView entirely and just render the element on top of the canvas manually. IMO, that's probably how overlay groups should work in the first place (rendering always 100% below or 100% above the canvas), but that's not how they work. At the moment, there is a large cost to overlay groups so anything that forces you to use them will make you pay a performance price.

jezell:
@ditman It wouldn't be a bad idea to make the rendering strategy that uses the native player avoid overlay groups entirely. Then you could pick between rendering into the canvas at a reasonable cost or rendering on top of the canvas at no cost. That would still be better than not being able to ever put a video on a page that happens to have 8 SVGs or 8 webrtc cameras on it.

jezell:
> This can't be the default method because `requestVideoFrameCallback` [doesn't seem to be supported at all on Firefox](https://caniuse.com/mdn-api_htmlvideoelement_requestvideoframecallback)?
> 
> I guess it can be made to fallback to the html element in older browsers, but how can this be more performant than the native tag?

Most frameworks use requestAnimationFrame when requestVideoFrameCallback isn't available. It's a reasonable strategy, since RAF is used all over the place, including in flutter itself as a fallback for various missing APIs.

ditman:
@jezell I'm not opposed to merging this, because I like that it brings new features that were impossible before (filtering, probably access to the stream of frames) but it's a lot of experimental APIs (and more that could be used, like the Web Codecs API, I guess). Let me bring it up with the team, I know there were concerns about transferring images like this in firefox for the core framework (I linked the performance issues in the related issue of this PR)

jezell:
@ditman added firefox support via RAF, works fine for playing full screen videos in Firefox for me in places where they crash due to overlay group limitations on the current player.

jezell:
@ditman Thanks. I understand that this is a bit cutting edge and scary, but it does work better than the current method, and I think providing the fallback in the plugin if you prefer the old behavior should help address the concern -- or at the very least including the new logic, but making you turn on this as "experimental" to start.

ditman:
There's some dart analyze issues, but those should be straightforward to fix.

A couple of tests are complaining, though:

```
The following PlatformException was thrown running a test:
PlatformException(MEDIA_ERR_SRC_NOT_SUPPORTED,
MEDIA_ELEMENT_ERROR: Format error, The video has been found to be
unsuitable (missing or in a format not supported by your
browser)., null)
```

ditman:
Another issue with this new technique is that it exposes the video_player_web to the same CORS problems that the NetworkImage has in the canvaskit renderer: [videos need to satisfy CORS restrictions](https://stackoverflow.com/a/29571550), which is not necessarily true with the `<video>` tag.

See: https://github.com/flutter/flutter/issues/107187#issuecomment-1200079139

I guess we must keep the platform-view based implementation for these cross-domain cases.

jezell:
@ditman yeah, the crossOrigin is set to anonymous right now to at least allow some things through, but seeing that NetworkImage has same problem, at least it's consistent with other assets. 

I don't think there's a good reason not to forbid using the platform layer upon request, but if you opt into it, at least then you'd be choosing  the risky behavior and it could be properly documented to make sure you know what you are signing up for. I'd guess the average consumer doesn't know they have to be very careful about overlay groups until it's too late and their renderer is misbehaving. Could probably also be detected and auto fallback, or maybe defaulted based on the renderer you are using? If using the HTML renderer maybe you'd prefer the video element because you don't have the webgl limits to content with? Since that's the default in places like mobile safari that are constrained for texture space it could actually be a good thing to default to the old way there.

stuartmorgan:
> the other platforms are already doing something like this to render the video into textures on their respective platforms

We expect to introduce platform view versions of playback for all of them over time, because the texture approach doesn't work with DRM-protected videos.

> I don't know of a good reason to keep the old behavior around forever

I would expect the same issues with DRM on the web. [This](https://w3c.github.io/encrypted-media/#media-element-restrictions) send to be explicitly saying that accessing frames of DRM-protected videos can't be relied on.

jezell:
> We expect to introduce platform view versions of playback for all of them over time, because the texture approach doesn't work with DRM-protected videos.

Good point. Hopefully all platforms will support both rendering into a texture or using a platform view depending on the needs of the app then. At least for us it would not be optimal to get reduced functionality for all videos because some videos can be DRM'd. Likely DRM'd content can get away with the limitations of platform views on webs as typically apps like Netflix don't let you watch more than one video at a time. Other cases like WebRTC, collaboration apps, etc. often do end up with a lot of videos on the screen at once and none of them are DRM'd, which is where we are currently running into trouble with platform views on web.

Another possibility I was thinking could help here on web (at least for videos) would be to only use the platform view when the video is playing, and to swap to a captured RawImage when it is not playing. Then, as long as you don't need multiple videos playing at once, you are only burning a single layer. Wouldn't be a great solution for WebRTC scenarios, but those aren't covered by this plugin anyway.

stuartmorgan:
> Hopefully all platforms will support both rendering into a texture or using a platform view depending on the needs of the app then.

That is my current expectation for the long term state, yes.

jezell:
Updated PR based on feedback in this thread.

In auto render mode (default) which will choose a rendering strategy dynamically at runtime. A specific strategy can also be selected to so that people can choose platform views if they want things like DRM to work or texture mode if they want shaders to work and no platform layers to be used.

In auto mode:

- When flutter is configured with the html renderer, auto mode will use platform views as there are no shaders to take advantage of and there is not such bad behavior when they are exhausted.

- When flutter is using canvaskit renderer, auto mode will only use a platform view for playing videos. When videos are not playing, it will capture an image to a texture. This could offer a better default than the current player with minimal performance impact, as it doesn't require the per frame overhead of createImageBitmap and drastically minimizes the number of platform layers that are being requested. This eliminates, for example, a current issue where simply putting multiple videos in a feed or trying to preview videos in a grid can exhaust the available overlay groups and cause rendering issues even if they aren't currently playing.

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -144,10 +157,453 @@ class VideoPlayerPlugin extends VideoPlayerPlatform {
...
+        if (widget.player.videoElement.parentElement == null &&
+            mode != VideoRenderMode.html) {
+          web.document.body!.appendChild(widget.player.videoElement);
+        }
+      }
+    }
+  }
+
+  bool get rendererCanvasKit {
+    return hasProperty(web.window, 'flutterCanvasKit');
```
ditman, line 218: This

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -144,10 +157,453 @@ class VideoPlayerPlugin extends VideoPlayerPlatform {
...
+    extends _VideoPlayerRendererState<_TextureVideoPlayerRenderer> {
+  @override
+  Future<void> capture() async {
+    if (!widget.paused || lastFrameTime != widget.element.currentTime) {
+      lastFrameTime = widget.element.currentTime;
+      try {
+        final web.ImageBitmap newSource =
+            await promiseToFuture<web.ImageBitmap>(
+                web.window.createImageBitmap(widget.element));
+        final ui.Image img = await ui_web.createImageFromImageBitmap(newSource);
```
ditman, line 470: This

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -144,10 +157,453 @@ class VideoPlayerPlugin extends VideoPlayerPlatform {
...
+class _CanvasVideoPlayerRendererState
+    extends _VideoPlayerRendererState<_CanvasVideoPlayerRenderer> {
+  @override
+  Future<void> capture() async {
+    lastFrameTime = widget.element.currentTime;
+    if (canvas != null) {
+      final web.CanvasRenderingContext2D context =
+          canvas!.getContext('2d')! as web.CanvasRenderingContext2D;
+
+      context.drawImage(widget.element, 0, 0, canvas!.width, canvas!.height);
```
ditman, line 540: This

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -144,10 +157,453 @@ class VideoPlayerPlugin extends VideoPlayerPlatform {
...
 
   /// Sets the audio mode to mix with other sources (ignored)
   @override
   Future<void> setMixWithOthers(bool mixWithOthers) => Future<void>.value();
+
+  /// The mode to render the video with.
+  VideoRenderMode mode = VideoRenderMode.auto;
+}
+
+class _AdapterVideoPlayerRenderer extends StatefulWidget {
```
ditman, line 175: (Extract to separate files)

packages/video_player/video_player_web/lib/src/browser_detection.dart
ditman, line 1: This whole file

ditman (commented):
@jezell we're discussing internally if there's anything else that could be exposed through `ui_web` to make this more efficient/better integrated with the renderer mechanisms; (for example, there's a new `skwasm` renderer coming that attempts to do off-thread rendering, and I'm not sure how this would play with it), but also the way the renderer is determined, or the browser detection code, could be vended from the engine so everything is guaranteed to stay in sync with whatever it says.

One quick thing, could you split the `_AdapterVideoPlayerRenderer` to separate file(s), so we keep the `video_player_web.dart` as "the place that has the platform channel overrides" and the new stuff that knows how to render the video Widget in different files?

Also, there's no need for classes to be "private" (prefixed by _) in this package, because people shouldn't be using it directly on their code.

packages/video_player/video_player_web/lib/video_player_web.dart
jezell: @ditman It's definitely not the fastest way the browser supports, but it is the only way that's currently exposed to get a frame into the renderer. I have some suggestions here in this thread:

https://github.com/flutter/flutter/issues/144815#issuecomment-1987309476

Ideally, in browsers where a VideoFrame is available, it could be transferred to the worker for rendering. There are some constraints with VideoFrames though that you can't hold onto them without blocking the decoding, so might require some additional plumbing to make work properly. Maybe WebCodecs would be the best option at some point: https://github.com/w3c/webcodecs/issues/43. But also not doable without some work in the flutter engine.

packages/video_player/video_player_web/lib/video_player_web.dart
jezell: @ditman worth some benchmarking, but other than the native element, this is probably the fastest path to get a video frame rendered in FF and Safari. It avoids the transfer to the worker, which the slow path in FF / Safari and avoids the GC churn caused by getting bitmaps every frame. 

This mode only exists to allow the player to only use a platform layer while playing and use a texture while not playing. Ideally, there wouldn't be a need for the canvas element because ideally we could swap from a texture straight to the video element when it starts playing. However, the challenge is that you can't move a video element in the DOM without resetting it, so if you try to move the video element into the platform layer when it is created, it will immediately stop playing. 

I haven't experimented with invisible platform layers yet, but it's possible that an alternative could be using an invisible platform layer for the video when it's not playing and then swapping it to be a visible layer later.  If the video element was not removed from the DOM during a swap from invisible to visible, there's a possibility it could work, but depends on whether Chrome would allow us to get the frames while the video was in the invisible layer and whether the element would get recreated by that swap.

jezell: @ditman worth some benchmarking, but other than the native element, this is probably the fastest path to get a video frame rendered in FF and Safari. It avoids the transfer to the worker, which the slow path in FF / Safari and avoids the GC churn caused by getting bitmaps every frame. 

This mode only exists to allow the player to only use a platform layer while playing and use a texture while not playing. Ideally, there wouldn't be a need for the canvas element because ideally we could swap from a texture straight to the video element when it starts playing. However, the challenge is that you can't move a video element in the DOM without resetting it, so if you try to move the video element into the platform layer when it is created, it will immediately stop playing. 

I haven't experimented with invisible platform layers yet, but it's possible that an alternative could be using an invisible platform layer for the video when it's not playing and then swapping it to be a visible layer later.  If the video element was not removed from the DOM during a swap from invisible to visible, there's a possibility it could work, but depends on whether Chrome would allow us to get the frames while the video was in the invisible layer and whether the element would get recreated by that swap.

packages/video_player/video_player_web/lib/src/browser_detection.dart
jezell: @ditman yeah, that file is in the browser engine, but can it be imported from here? It would definitely be ideal to not have to double implement it.

ditman: Nope, it can't be imported now, it needs to be reexported from the `ui_web` package to work.

stuartmorgan:
From triage: @ditman What's the status of this PR? Is it blocked by potential SDK changes?

stuartmorgan:
From triage: @ditman Ping on the question above?

ditman:
Sorry I missed the earlier message!

There's at least two relevant changes to the SDK:

* https://main-api.flutter.dev/flutter/dart-ui_web/browser.html (exposed the browser detection from the engine, currently in `master`)
* we're looking at deprecating the HTML renderer

The priority right now is in doing the work to move to canvaskit (js)/skwasm (wasm). Platform views are staying, so the current implementation of the engine stays. I'm still unsure on how the image copying would play with the rest of the framework (but currently there's a pattern where people do custom rendering to bytes in Flutter, and I guess that'd need to continue working)

I think this should land as an opt-in feature for the video player web, I'll give this some testing after I wrap up some unrelated stuff.

jezell:
@ditman React Native Skia is using MakeLazyImageFromTextureSource for a faster path here. I really think that API should land in web_ui to allow this to be improved by binding the WebGL texture straight to the video element.

kevmoo:
Are we blocked here? Or still in discussion...

jezell:
@kevmoo Fix requires some changes that are in main branch, how does the plugin release cycle work? 

I had pointed out the potential problem here:

https://github.com/flutter/flutter/issues/150592

I think video_player_web is kind of a bad place to put the lower level video rendering logic because there are a lot of places that need the same logic, and video_player_web is way higher level, so it can't be used by camera_web or livekit_client. Not sure what approach the team wants to go with here. It's easy enough to hack it in to all these places manually, but feels wrong having to duplicate the code and then keep updating it as the web engine gets better.

stuartmorgan:
> Fix requires some changes that are in main branch, how does the plugin release cycle work?

Please see https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#supported-flutter-versions for this repository's support model.

chunhtai:
(triage) hi @jezell , are we still blocked? or do you know what is the status of this pr?

stuartmorgan:

PR #6292, opened 2024-03-09
[google_sign_in] Update README with instructions to use this package without Firebase

lukey78:
There is some confusion about using this package without Firebase, especially for correctly integrating it with Android. The README update is based on the the following web resources:

- https://github.com/flutter/flutter/issues/20903
- https://github.com/flutter/flutter/issues/36673
- https://github.com/flutter/flutter/issues/33393
- https://medium.com/codebrew/flutter-google-sign-in-without-firebase-3680713966fb

[36673](https://github.com/flutter/flutter/issues/36673)
[33393](https://github.com/flutter/flutter/issues/33393)
[20903](https://github.com/flutter/flutter/issues/20903)


stuartmorgan:
Thanks for the contribution, but this goes into a lot of details we wouldn't want to add to the README. E.g., explaining concepts like bundle IDs or the mechanics of Android app signing are well out of scope for the plugin, and would apply equally to non-Flutter applications. There are also a number of things here that would tend to become outdated very quickly, such as references to specific UI strings in external products.

This level of detail is probably best for something like a StackOverflow post, or a codelab. In the context of this README, what we would want is something much shorter and more evergreen, ideally consisting of links to the relevant official documentation for the wrapped SDKs. The only steps that we would want to cover in detail here would be anything specific to the Flutter plugin.

lukey78:
Apparently - based on the vast amount of issues and articles regarding this issue - it seems that many people struggle with using the plugin without Firebase. I think the docs should tell something about the correct usage. And of course all of that is related to the Flutter plugin - this plugin seems to be the only way to make Google Sign In to work with Flutter. The standard OAuth2/OIDC plugins like flutter_web_auth_2 will not work anymore due to the deprecation of custom URI schemes.

stuartmorgan:
> I think the docs should tell something about the correct usage.

We agree, which is why https://github.com/flutter/flutter/issues/36673 is open. I'd be happy to review a PR along the lines of what I described at the end of my comment above.

PR #3879, opened 2023-05-02
[webview_flutter_android] Added the functionality to fullscreen html5 video

paulppn:
At this moment on android the fullscreen html5 video does not work. This PR solves that issues, adding the functionality that the video is played fullscreen when you click on the fullscreen-button in the video player.

Fixes https://github.com/flutter/flutter/issues/27101

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ x I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #6335, opened 2024-03-15
[webview_flutter_android] Fixes iframe navigation with `onNavigationRequest`

petermnt:
This PR reintroduces the check on `WebResourceRequest.isForMainFrame` in the `shouldOverrideUrlLoading` callback.
This check already existed in the implementation before the change to Pigeon, but was not taken over.  See https://github.com/flutter/packages/blob/418bef0d6f3100e7a4bafb86537285e6d2095159/packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/FlutterWebViewClient.java#L99

The reasoning was already already in comment there:
```
      // Since we cannot call loadUrl for a subframe, we currently only allow the delegate to stop
      // navigations that target the main frame, if the request is not for the main frame
      // we just return false to allow the navigation.
      //
      // For more details see: https://github.com/flutter/flutter/issues/25329#issuecomment-464863209
```
 
Currently when a `NavigationDeletage.onNavigationRequest` is set and a navigation request for an iframe happens it will automatically be blocked and handed to the Flutter `onNavigationRequest`.  
If the request gets a `NavigationDecision.navigate` it will be loaded into the  mainframe through a `loadUrl`.  
If the request gets a `NavigationDecision.prevent` it will never be executed.

This causes an issue when for example an iframe loads local data.
Example
```data:text/html,<script>onresize=function(){parent.postMessage(0,'*')}<\/script>```

An iframe is allowed to do this, but with the current implementation this data will be loaded in the main frame, leading to a white page. 

Issue: https://github.com/flutter/flutter/issues/145208

## Pre-launch Checklist

- [x] I signed the [CLA].

### CODE REVIEW

File: packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/WebViewClientCompatImplTest.java [hunk 1]
```diff
@@ -0,0 +1,21 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package io.flutter.plugins.webviewflutter;
+
+import android.webkit.WebViewClient;
+
+public class WebViewClientCompatImplTest extends WebViewClientTest {
```
bparrishMines, line 9: Extending a test seems to violate: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#make-each-test-entirely-self-contained. I would prefer just having all the tests added to `WebViewClientTest` with the different classes.

bparrishMines (commented):
This is a good fix! It looks like I missed this code in the pre `webview v4` code: https://github.com/flutter/plugins/pull/1236/files#diff-8ba24b285f0228d5dd613ad64b1e63f6ce4be9819ec1a7e102a2dd269de569c6R53

I think we should also change `AndroidWebViewController` to only call `onNavigationRequest` for the main frame. Otherwise, someone may return `prevent` to the callback and it would be ignored for an unknown reason.

packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/WebViewClientCompatImplTest.java
petermnt: Fair enough, I will make the change tomorrow

petermnt:
> I think we should also change `AndroidWebViewController` to only call `onNavigationRequest` for the main frame. Otherwise, someone may return `prevent` to the callback and it would be ignored for an unknown reason.

I was thinking the same, but for now restored the 'old' behaviour. I guess it was like that to still inform the flutter side.

My current fix indeed needs this in our implementation.
```dart
  if (!navigation.isMainFrame) {
      return _platformRepo.isIOS ? NavigationDecision.navigate : NavigationDecision.prevent;
    }
```

I will apply your suggestion tomorrow.

packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/WebViewClientCompatImplTest.java
petermnt: I leaned into the 'Embrace code duplication in tests.' and 'Prefer more test files, avoid long test files'

The test that's independent of the implementation is in `WebViewClientTest` and the tests that should be tested on each implementation are duplicated in `WebViewClientCompatImplTest` and `WebViewClientImplTest`

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewClientHostApiImpl.java
```diff
@@ -87,6 +87,12 @@ public void onReceivedError(
     @Override
     public boolean shouldOverrideUrlLoading(
         @NonNull WebView view, @NonNull WebResourceRequest request) {
+      if (!request.isForMainFrame()) {
```
bparrishMines, line 90: I think your original implementation here was correct. We should still have the callback made to the Dart `WebViewClient`. However, `AndroidWebviewController` should check that the callback is for the main frame before calling `onNavigationRequest`: https://github.com/flutter/packages/blob/main/packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart#L1467

petermnt: I see, I will make that adaptation tomorrow

packages/webview_flutter/webview_flutter_android/example/lib/main.dart
```diff
@@ -23,10 +23,10 @@ const String kNavigationExamplePage = '''
 <head><title>Navigation Delegate Example</title></head>
 <body>
 <p>
-The navigation delegate is set to block navigation to the youtube website.
+The navigation delegate is set to block navigation to the pub.dev website.
```
bparrishMines, line 26: This should stay youtube since the test navigation delegate is blocking youtube.

```suggestion
The navigation delegate is set to block navigation to the youtube website.
```

packages/webview_flutter/webview_flutter_android/example/lib/main.dart
```diff
@@ -23,10 +23,10 @@ const String kNavigationExamplePage = '''
...
 </p>
 <ul>
-<ul><a href="https://www.youtube.com/">https://www.youtube.com/</a></ul>
+<ul><a href="https://pub.dev/">https://pub.dev/</a></ul>
```
bparrishMines, line 29: Same here

```suggestion
<ul><a href="https://www.youtube.com/">https://www.youtube.com/</a></ul>
```

packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/WebViewClientTest.java
```diff
@@ -66,51 +53,9 @@ public void tearDown() {
...
-    when(mockRequest.isForMainFrame()).thenReturn(true);
-    when(mockRequest.getRequestHeaders()).thenReturn(null);
-
-    final GeneratedAndroidWebView.WebResourceRequestData data =
-        WebViewClientFlutterApiImpl.createWebResourceRequestData(mockRequest);
-    assertEquals(data.getRequestHeaders(), new HashMap<String, String>());
-  }
-
   @Test
   public void setReturnValueForShouldOverrideUrlLoading() {
```
bparrishMines, line 57: Since it looks like this PR is creating separate tests files for `WebViewClient` and `WebViewClientCompat`, you can go ahead and copy this file to the new tests and then delete this file.

bparrishMines (changes requested):
Most of this looks good. Just a few more comments.

@stuartmorgan This is ready for a secondary review. And what are your thoughts on this being a bug fix on needing to be a breaking change?

packages/webview_flutter/webview_flutter_android/example/lib/main.dart
petermnt: It isn't as far as I could see. 
https://github.com/flutter/packages/blob/d1d3bd6163977bd15417918b942fc18c221aa229/packages/webview_flutter/webview_flutter_android/example/lib/main.dart#L192

That's actually the reason I changed it, because the delegate didn't match the example.

packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/WebViewClientTest.java
petermnt: I actually left this one out on purpose, as this test is not dependent on an implementation.
This way there's 3 files with clear distinction
* `WebViewClientTest` which tests things that are not dependent on the implementation
* `WebViewClientImplTest` and `WebViewClientCompatImplTest` which test the 2 implementations.

packages/webview_flutter/webview_flutter_android/example/lib/main.dart
bparrishMines: Ah, you're right. It looks like the url was changed in https://github.com/flutter/packages/pull/3879, but the example html wasn't. Thanks for catching this!

packages/webview_flutter/webview_flutter_android/android/src/test/java/io/flutter/plugins/webviewflutter/WebViewClientTest.java
bparrishMines: Yes, but both implementations have `setReturnValueForShouldOverrideUrlLoading`, right. So I think both files and implementations should have there own test for it.

petermnt: I've had a look at the implementation (I only looked at the test before) and now see that the internal implementation is indeed dependant on the `instanceof` check.  So there would indeed be a point of testing both.

I also noticed that `WebViewClientCompatImplTest` still had the test, so `WebViewClientTest` was redundant anyway.

Adding the test to `WebViewClientImplTest` is not possible with the current setup as I can't mock the static call to `Build.VERSION.SDK_INT`.

I have removed the redundant `WebViewClientTest`. The 'missing' test for `WebViewClientImplTest` was already missing and not directly related to my PR.

bparrishMines (approved):
LGTM

@stuartmorgan for secondary review

stuartmorgan (approved):
LGTM

PR #6355, opened 2024-03-19
Update multiple packages to depend on versions with iOS privacy manifest included

vbuberen:
Updated multiple packages `pubspec.yaml` to depend on iOS implementation versions that include privacy manifest required by Apple. I touched packages that were referenced in https://github.com/flutter/flutter/issues/131940 issue. 
The reason to do this change is that in a few projects I noticed that updating `url_launcher` dependency to the currently latest 6.2.5 doesn't also update `url_launcher_ios` to version with privacy manifest (appeared in 6.2.4) included in the plugin's `pubspec.yaml` it is declared as `^6.2.0`, thus, version resolution picks 6.2.0 and that is it.

Considering that there is already a deadline of May 1, 2024 from Apple on when apps would have to include privacy manifests, it would be good to let devs already prepare their projects. This change should be especially useful for `shared_preferences`, which has `NSPrivacyAccessedAPITypes` array declared in its privacy manifest and which is already asked by Apple in emails after uploading new builds:

Screenshot 2024-03-19 at 12 36 47

Didn't update changelogs due to the fact that plugins stayed in almost the same versions range, where those minor changes to iOS part had no serious changed.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/image_picker/image_picker/CHANGELOG.md
+++ b/packages/image_picker/image_picker/CHANGELOG.md
@@ -1,9 +1,10 @@
-## NEXT
+## 1.0.8
 
 * Updates minimum supported SDK version to Flutter 3.13/Dart 3.1.
 * Updates support matrix in README to indicate that iOS 11 is no longer supported.
 * Clients on versions of Flutter that still support iOS 11 can continue to use this
   package with iOS 11, but will not receive any further updates to the iOS implementation.
+* Updates minimum iOS implementation version to include a privacy manifest.
 
 ## 1.0.7
 
--- a/packages/video_player/video_player/pubspec.yaml
+++ b/packages/video_player/video_player/pubspec.yaml
@@ -3,11 +3,11 @@ description: Flutter plugin for displaying inline video with other Flutter
   widgets on Android, iOS, and web.
 repository: https://github.com/flutter/packages/tree/main/packages/video_player/video_player
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+video_player%22
-version: 2.8.5
+version: 2.8.6
 
 environment:
-  sdk: ">=3.1.0 <4.0.0"
-  flutter: ">=3.13.0"
+  sdk: ">=3.2.3 <4.0.0"
+  flutter: ">=3.16.6"
 
 flutter:
   plugin:
@@ -26,7 +26,7 @@ dependencies:
     sdk: flutter
   html: ^0.15.0
   video_player_android: ^2.3.5
-  video_player_avfoundation: ^2.5.0
+  video_player_avfoundation: ^2.5.6
   video_player_platform_interface: ">=6.1.0 <7.0.0"
   video_player_web: ^2.0.0
 
--- a/packages/image_picker/image_picker/pubspec.yaml
+++ b/packages/image_picker/image_picker/pubspec.yaml
@@ -3,7 +3,7 @@ description: Flutter plugin for selecting images from the Android and iOS image
   library, and taking new pictures with the camera.
 repository: https://github.com/flutter/packages/tree/main/packages/image_picker/image_picker
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+image_picker%22
-version: 1.0.7
+version: 1.0.8
 
 environment:
   sdk: ^3.1.0
@@ -30,7 +30,7 @@ dependencies:
     sdk: flutter
   image_picker_android: ^0.8.7
   image_picker_for_web: ">=2.2.0 <4.0.0"
-  image_picker_ios: ^0.8.8
+  image_picker_ios: ^0.8.9+1
   image_picker_linux: ^0.2.1
   image_picker_macos: ^0.2.1
   image_picker_platform_interface: ^2.8.0
--- a/packages/path_provider/path_provider/CHANGELOG.md
+++ b/packages/path_provider/path_provider/CHANGELOG.md
@@ -1,9 +1,10 @@
-## NEXT
+## 2.1.3
 
 * Updates minimum supported SDK version to Flutter 3.13/Dart 3.1.
 * Updates support matrix in README to indicate that iOS 11 is no longer supported.
 * Clients on versions of Flutter that still support iOS 11 can continue to use this
   package with iOS 11, but will not receive any further updates to the iOS implementation.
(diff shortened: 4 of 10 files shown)


stuartmorgan:
Normally we don't impose version constraints that aren't required by the code, but we have [precendent](https://github.com/flutter/plugins/pull/6595#issuecomment-1287065534) for doing this when there's a store-level requirement that is imposed, so this seems reasonable.

> in the plugin's `pubspec.yaml` it is declared as `^6.2.0`, thus, version resolution picks 6.2.0 and that is it.

Version resolution will pick the newest version it can. The reason existing projects don't get updated versions of `url_launcher_ios` when just changing `url_launcher`'s version manually is pubspec.lock, not that the resolver won't give newer versions.

> it would be good to let devs already prepare their projects

Devs can already prepare their projects simply by running `flutter pub upgrade`. We can in practice speed the update process due to many people choosing for some reason to only update their direct dependencies rather than also updating their transitive dependencies, but developers could (and should) update transitive dependencies without any change by us.

> Didn't update changelogs due to the fact that plugins stayed in almost the same versions range, where those minor changes to iOS part had no serious changed.

What is the purpose of changing the constraints without publishing that change?

vbuberen:
> Version resolution will pick the newest version it can. The reason existing projects don't get updated versions of url_launcher_ios when just changing url_launcher's version manually is pubspec.lock, not that the resolver won't give newer versions.

Alright, got it. In such case I have a few projects with some limitations. And yes, it is manual change of version in pubspec.yaml + click on `Get` button in the IDE. I know quite a lot of people who do updates the same way, not just my direct colleagues, so this PR is something that would help ensure that such people also get updated versions resolved when next releases of touched plugins will be released.

> Devs can already prepare their projects simply by running `flutter pub upgrade`.

I think you meant `flutter pub upgrade [dependency name]` as there are enough projects with a huge set of dependencies where it is not the best option to run an upgrade for everything instead of doing part by part for smaller set of changes.

> What is the purpose of changing the constraints without publishing that change?

You misunderstood my message. The change needs publishing. The thing is that those versions of iOS part were compatible already before, so this bump doesn't change much and, thus, from my point of view is not something worth mentioning separately.

But in general it seems like changes from this PR are unwanted . You could just reject the PR with request for changes if you expect some or close it instead.

stuartmorgan:
> > Devs can already prepare their projects simply by running `flutter pub upgrade`.
> 
> I think you meant `flutter pub upgrade [dependency name]` as there are enough projects with a huge set of dependencies where it is not the best option to run an upgrade for everything instead of doing part by part for smaller set of changes.

In this case I do mean `flutter pub upgrade`. I expect most developers have multiple transitive dependencies that they will need to update to comply with the new App Store polices, and tracking down all of the relevant transitive dependencies one at a time by hand is likely to be pretty painful. Obviously it's their choice either way though.

> But in general it seems like changes from this PR are unwanted

No, as I said we have precedent for doing this kind of change specifically in the case of app store requirements. I was just clarifying the context/reason we would accept it, since the PR description suggested that this change was required in order for clients to update, which isn't actually the case. I clarify this whenever anyone says it in any PR or issue, because it's a relatively common misconception, and my hope is that the more places it's corrected, the less widespread the misconception will be.

stuartmorgan (changes requested):
> - [ ] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning), or this PR is [exempt from version changes](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates).
> - [ ] I updated CHANGELOG.md to add a description of the change, [following repository CHANGELOG style](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style).

These steps are required for the PR to land.

packages/video_player/video_player/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Flutter plugin for displaying inline video with other Flutter
   widgets on Android, iOS, and web.
 repository: https://github.com/flutter/packages/tree/main/packages/video_player/video_player
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+video_player%22
-version: 2.8.3
+version: 2.8.4
 
 environment:
   sdk: ">=3.1.0 <4.0.0"
```
stuartmorgan, line 9: These two values need to be rev'd to fix [this failure](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8752915517182459137/+/u/Run_package_tests/analyze_-_legacy/stdout); the package currently fails resolution on 3.13.

packages/image_picker/image_picker/CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
-## NEXT
+## 1.0.8
 
 * Updates minimum supported SDK version to Flutter 3.13/Dart 3.1.
 * Updates support matrix in README to indicate that iOS 11 is no longer supported.
 * Clients on versions of Flutter that still support iOS 11 can continue to use this
   package with iOS 11, but will not receive any further updates to the iOS implementation.
+* Bumps min version of iOS part to start with version with privacy manifest included.
```
stuartmorgan, line 7: I'm having trouble parsing this (I'm not sure what it means for the minimum version of something to start with something means). How about:

> Updates minimum iOS implementation version to include a privacy manifest.

(This applies throughout the PR.)

stuartmorgan (changes requested):

vbuberen:
@stuartmorgan Would you mind checking this PR again?

The failing integration test is something I can't tackle as locally this test passes Ok for me in my setup with Android emulators. I see that the reason for failing on CI is in some small time diff in position, which seems a bit odd: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8751747162297807793/+/u/Run_package_tests/drive_examples/stdout#L23859_0 
I saw this issue https://github.com/flutter/flutter/issues/141145 and the comment so assume that this test is expected to be flaky, even though previously there was a different issue with it?

stuartmorgan (approved):
LGTM

stuartmorgan:
@tarrinneal Could you do the secondary review here?

(This will need conflicts resolved, but that can wait until it has all the reviews so that it doesn't get re-conflicted.)

tarrinneal (approved):

vbuberen:
@stuartmorgan Looks like this one can be merged. Second approve is done and I have already updated the branch to resolve conflicts.

PR #6371, opened 2024-03-21
[pigeon] Kotlin implementation for ProxyApis 

bparrishMines:
Kotlin portion of https://github.com/flutter/flutter/issues/134777

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
```diff
@@ -0,0 +1,475 @@
...
+
+    when {
+      instanceManager.containsInstance(value) -> {
+        stream.write(128)
+        writeValue(stream, instanceManager.getIdentifierForStrongReference(value))
+      }
+      else -> super.writeValue(stream, value)
+    }
+  }
+}
```
bparrishMines, line 426: @camsim99 FYI

@stuartmorgan @tarrinneal This is one of the places where the design changes from the implementation in the doc.  I wanted to improve the scenario of when a method returns a `ProxyApi` or a list/map of a `ProxyApi`:

```java
class Basket {
  Apple giveMeAnApple();
  List<Apple> giveMeApples();
}
```

Typically we solve this by needing to handle this manually when we implement the api:

```java
class  PigeonApiBasketImpl {
  Apple giveMeAnApple(Basket instance) {
    final Apple apple = instance.giveMeAnApple();
    if (!instanceManager.containsInstance(apple)) {
      instanceManager.addHostCreatedInstance(apple);
    }

    return apple;
  }

  List<Apple> giveMeApples(Basket instance) {
    final List<Apple> apples = instance.giveMeApples();
    for (Apple apple in apples) {
      if (!instanceManager.containsInstance(apple)) {
        instanceManager.addHostCreatedInstance(apple);
      }
    }

    return apples;
  }
}
```

This change makes the codec the central class that contains all the proxy api implementations, so the `writeValue` can handle creating new dart instances by calling `pigeon_newInstance`.

This also:
* Creates a registrar like class that ensures a user implements each proxy api.
* A central class with every api implementation prevents the need for multiple instances of the same api implementation.

tarrinneal: This may end up overlapping with some of the codec work I'm about to do as well. I think it's generally a good idea to create a central codec though. We may end up with multiple codecs implementing each other.

tarrinneal: My main concern at this point is that the number of custom inputs is limited to around 120, so if there are a lot of proxy apis, enums, and or data classes, we may not be able to accommodate them all.

bparrishMines: As of right now, the generation of the `ProxyApi` codec is separate from Host/Flutter APIs codecs. Mainly because:
1. ProxyApis don't support data classes
2. ProxyApis support other ProxyApis in the codec and Host/Flutter APIs don't.

So they currently don't really share anything. If you create a central custom codec for all the APIs, then `ProxyApis` should only need one dedicated spot that uses the `InstanceManager`: https://github.com/flutter/packages/blob/main/packages/pigeon/platform_tests/shared_test_plugin_code/lib/src/generated/proxy_api_tests.gen.dart#L374. 

Also, as a side note, since all data classes have an `encode/decode` could you create a base data class for all the data classes and only use one spot.

tarrinneal:
Are you planning to add the full implementation and test suite to this pr?

bparrishMines:
> Are you planning to add the full implementation and test suite to this pr?

I wasn't planning on doing that for this PR, but I can if you prefer the full implementation. This PR just adds the Kotlin ProxyApi utility classes, including the `InstanceManager` and the tests for it. And also the bare minimum `ProxyApi` class declaration.

bparrishMines:
@tarrinneal I updated the PR to have the full implementation and tests. Its ready for a review.

packages/pigeon/lib/kotlin/templates.dart
```diff
@@ -0,0 +1,344 @@
...
+              wrapError(exception)
+            }
+            reply.reply(wrapped)
+          }
+        } else {
+          channel.setMessageHandler(null)
+        }
+      }
+    }
+  }
```
tarrinneal, line 320: a lot of this code is essentially duplicated with the host api. I'd prefer if that code could be gened from the same code. to avoid future work duplication

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -839,6 +1099,615 @@ class KotlinGenerator extends StructuredGenerator<KotlinOptions> {
...
+      ' * needed by any implementation.\n'
+      ' */',
+    );
+    indent.writeScoped(
+      'abstract class $registrarName(val binaryMessenger: BinaryMessenger) {',
+      '}',
+      () {
+        indent.writeln('val instanceManager: PigeonInstanceManager');
+        indent.format(
+          'private var _codec: StandardMessageCodec? = null\n'
```
tarrinneal, line 1121: This will need to change to use the new codec - once the codec pr lands of course

tarrinneal (commented):
quick little glance through, I think most of this pr is good as is. Any suggestions I've made here are generally open for discussion if you want to keep them.

packages/pigeon/lib/kotlin/templates.dart
bparrishMines: This will also require changes to the Dart code, so I created https://github.com/flutter/packages/pull/6831

packages/pigeon/lib/kotlin/templates.dart
```diff
@@ -0,0 +1,344 @@
...
+  /** Interface for listening when a weak reference of an instance is removed from the manager.  */
+  interface $_finalizationListenerClassName {
+    fun onFinalize(identifier: Long)
+  }
+
+  private val identifiers = java.util.WeakHashMap<Any, Long>()
+  private val weakInstances = HashMap<Long, java.lang.ref.WeakReference<Any>>()
+  private val strongInstances = HashMap<Long, Any>()
+  private val referenceQueue = java.lang.ref.ReferenceQueue<Any>()
+  private val weakReferencesToIdentifiers = HashMap<java.lang.ref.WeakReference<Any>, Long>()
```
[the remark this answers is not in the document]
bparrishMines: I used the Kotlin->Java converter from Android Studio and this is what it used. I tried using the Kotlin version of `WeakReference`, but it seems to not be supported by this version of Kotlin. Same goes for the other `java.util` and `java.lang` classes.

packages/pigeon/lib/ast.dart
```diff
@@ -4,6 +4,7 @@
 
 import 'package:collection/collection.dart' show ListEquality;
 import 'package:meta/meta.dart';
+import 'kotlin_generator.dart';
```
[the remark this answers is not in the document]
bparrishMines: I can't add this to the normal kotlin options class because these options are used for each `@ProxyApi` class. The normal kotlin options is used for the whole file. I changed it to 

```dart
import 'kotlin_generator.dart' show KotlinProxyApiOptions;
```

though.

packages/pigeon/lib/kotlin/templates.dart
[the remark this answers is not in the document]
tarrinneal: What version do we need to use the kotlin version? Maybe we need a todo?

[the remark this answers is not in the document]
bparrishMines: It looks like it was created in `1.3`. This plugin uses `1.7.10`. But it looks like the api is still experimental: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native.ref/-weak-reference/. 

No version, including the latest `2.0`, has `WeakReference` as not experimental. And it looks like experimental apis require an opt-in, which I think is something we wouldn't want to do: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.experimental/-experimental-native-api/.

packages/pigeon/lib/kotlin/templates.dart
```diff
@@ -0,0 +1,256 @@
...
+ * <p>When an instance is added with an identifier, either can be used to retrieve the other.
+ *
+ * <p>Added instances are added as a weak reference and a strong reference. When the strong
+ * reference is removed with [remove] and the weak reference is deallocated, the
+ * `finalizationListener` is made with the instance's identifier. However, if the strong reference
+ * is removed and then the identifier is retrieved with the intention to pass the identifier to Dart
+ * (e.g. calling [getIdentifierForStrongReference]), the strong reference to the
+ * instance is recreated. The strong reference will then need to be removed manually again.
+ */
+@Suppress("UNCHECKED_CAST", "MemberVisibilityCanBePrivate", "unused")
```
stuartmorgan, line 24: What's `unused` for? Could we annotate specific things (with comments) for that one instead?

packages/pigeon/lib/kotlin/templates.dart
```diff
@@ -0,0 +1,256 @@
...
+    private const val tag = "$instanceManagerClassName"
+
+    /**
+     * Instantiate a new manager.
+     *
+     *
+     * When the manager is no longer needed, [stopFinalizationListener] must be called.
+     *
+     * @param finalizationListener the listener for garbage collected weak references.
+     * @return a new `$instanceManagerClassName`.
```
stuartmorgan, line 73: Optional (for all these comments): the Kotlin style guide [recommends against Javadoc style](https://kotlinlang.org/docs/coding-conventions.html#documentation-comments).

packages/pigeon/lib/kotlin/templates.dart
```diff
@@ -0,0 +1,256 @@
...
+   * @param instance the instance whose presence in this manager is to be tested.
+   * @return whether this manager contains the given `instance`.
+   */
+  fun containsInstance(instance: Any?): Boolean {
+    logWarningIfFinalizationListenerHasStopped()
+    return identifiers.containsKey(instance)
+  }
+
+  /**
+   * Stop the periodic run of the [$_finalizationListenerClassName] for instances that have been garbage
```
stuartmorgan, line 177: Nit: Stops

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -452,6 +490,317 @@ class KotlinGenerator extends StructuredGenerator<KotlinOptions> {
...
+              () {
+            indent.writeln('StandardMessageCodec()');
+          });
+          indent.newln();
+
+          indent.writeln('/**');
+          indent.writeln(
+              '* Sets up an instance of `$instanceManagerApiName` to handle messages from the');
+          indent.writeln('* `binaryMessenger`.');
+          indent.writeln('*/');
```
stuartmorgan, line 543: Wouldn't multi-line blocks like this be more readable with `format` instead of a bunch of `writeln`s?

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -452,6 +490,317 @@ class KotlinGenerator extends StructuredGenerator<KotlinOptions> {
...
+    final Iterable<AstProxyApi> allProxyApis =
+        root.apis.whereType<AstProxyApi>();
+
+    _writeProxyApiRegistrar(indent, allProxyApis: allProxyApis);
+
+    const String codecName = '${classNamePrefix}ProxyApiBaseCodec';
+
+    // Sort APIs where edges are an API's super class and interfaces.
+    //
+    // This sorts the apis to have child classes be listed before their parent
```
stuartmorgan, line 625: APIs

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -839,6 +1201,624 @@ class KotlinGenerator extends StructuredGenerator<KotlinOptions> {
...
+          'val codec: StandardMessageCodec\n'
+          '  get() {\n'
+          '    if (_codec == null) {\n '
+          '      _codec = PigeonProxyApiBaseCodec(this)\n'
+          '    }\n'
+          '    return _codec!!\n'
+          '  }\n',
+        );
+        indent.format(
+          'init {\n'
```
stuartmorgan, line 1234: `'''` to improve legibility?

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTestApiImpls.kt
```diff
@@ -0,0 +1,680 @@
...
+
+  override fun attachedField(pigeon_instance: ProxyApiTestClass): ProxyApiSuperClass {
+    return ProxyApiSuperClass()
+  }
+
+  override fun staticAttachedField(): ProxyApiSuperClass {
+    return ProxyApiSuperClass()
+  }
+
+  override fun aBool(pigeon_instance: ProxyApiTestClass): Boolean {
```
stuartmorgan, line 78: Is there a reason these are using hard-code values instead of being echo methods like the existing tests? Echo allows for more test coverage. E.g., being able to test that a nullable-return method can successfully return both a null and a non-null value.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
```diff
@@ -0,0 +1,3947 @@
...
+ */
+class ProxyApiTestsError(
+    val code: String,
+    override val message: String? = null,
+    val details: Any? = null
+) : Throwable()
+/**
+ * Maintains instances used to communicate with the corresponding objects in Dart.
+ *
+ * <p>Objects stored in this container are represented by an object in Dart that is also stored in
```
stuartmorgan, line 54: Why does this comment block have `<p>`s when none of the others do?

packages/pigeon/lib/kotlin/templates.dart
tarrinneal: Ideally we would just use the addDocumentationComments method for this, but idk if it's too complicated to plumb into these templates.

tarrinneal (commented):
comment comments

packages/pigeon/lib/generator_tools.dart
```diff
@@ -494,6 +500,48 @@ Map<TypeDeclaration, List<int>> getReferencedTypes(
...
+
+  if (allReferencedTypes.isEmpty) {
+    return null;
+  }
+
+  final TypeDeclaration typeWithHighestRequirement = allReferencedTypes.reduce(
+    (TypeDeclaration one, TypeDeclaration two) {
+      return onCompare(onGetApiRequirement(one)!, onGetApiRequirement(two)!) > 0
+          ? one
+          : two;
```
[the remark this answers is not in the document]
bparrishMines: I tried this, but it changed the formatting to: 

```dart
final TypeDeclaration typeWithHighestRequirement = allReferencedTypes.reduce(
    (TypeDeclaration typeOne, TypeDeclaration typeTwo) {
      return onCompare(onGetApiRequirement(typeOne)!,
                  onGetApiRequirement(typeTwo)!) >
              0
          ? typeOne
          : typeTwo;
    },
  );
```

packages/pigeon/lib/kotlin/templates.dart
bparrishMines: I removed it and it looks like it actually isn't needed.

packages/pigeon/lib/kotlin/templates.dart
bparrishMines: It would be pretty complicated since `addDocumentationComments` requires an `Indent` and the helper method returns `<String>[]`.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
bparrishMines: Forgot to remvoe

packages/pigeon/lib/kotlin_generator.dart
bparrishMines: I've tried to use `'''` with format, but the code that comes out (before running the formatter) includes the empty spaces unless you do something like this: https://github.com/flutter/packages/blob/main/packages/pigeon/lib/dart_generator.dart#L414

I updated `Indent.format` with a flag that can trim the indentation: https://github.com/flutter/packages/pull/6371/commits/78128b8dee3001f49b8f4bedaa5ae5731e9e71c6. Let me know if you believe there is a cleaner way.

cc @tarrinneal

packages/pigeon/lib/ast.dart
```diff
@@ -139,6 +140,7 @@ class AstProxyApi extends Api {
     required this.fields,
     this.superClass,
     this.interfaces = const <TypeDeclaration>{},
+    this.kotlinOptions,
```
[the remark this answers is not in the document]
bparrishMines: Since the generated code references the native class, each ProxyApi needs to provide the full class name with the package. I also think `Method.swiftFunction` is similar since it provides additional instructions for the Swift generator.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
```diff
@@ -0,0 +1,4085 @@
...
+ *
+ * Added instances are added as a weak reference and a strong reference. When the strong reference
+ * is removed with [remove] and the weak reference is deallocated, the
+ * `finalizationListener.onFinalize` is called with the instance's identifier. However, if the
+ * strong reference is removed and then the identifier is retrieved with the intention to pass the
+ * identifier to Dart (e.g. calling [getIdentifierForStrongReference]), the strong reference to the
+ * instance is recreated. The strong reference will then need to be removed manually again.
+ */
+@Suppress("UNCHECKED_CAST", "MemberVisibilityCanBePrivate")
+class PigeonInstanceManager(private val finalizationListener: PigeonFinalizationListener) {
```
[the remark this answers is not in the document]
bparrishMines: `fileSpecificClassNameComponent` was just added in https://github.com/flutter/packages/pull/6600. I updated this PR to use it.

packages/pigeon/lib/generator_tools.dart
```diff
@@ -76,11 +77,37 @@ class Indent {
...
+    int? shortestIndentation;
+    if (trimIndentation) {
+      for (final String line in lines) {
+        if (line.trim().isNotEmpty) {
+          final int indentationLength = line.length - line.trimLeft().length;
+          shortestIndentation = shortestIndentation == null
+              ? indentationLength
+              : min(shortestIndentation, indentationLength);
+        }
+      }
```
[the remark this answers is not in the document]
bparrishMines: This is a really nice improvement. It looks like I need to brush up on my functional programming!

gmackall:
From android triage: Reid is out for the rest of this week, does this need another Android reviewer in the meantime, or is work still ongoing?

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTestApiImpls.kt
```diff
@@ -0,0 +1,699 @@
...
+import androidx.annotation.RequiresApi
+import io.flutter.plugin.common.BinaryMessenger
+
+class ProxyApiTestClass : ProxyApiSuperClass(), ProxyApiInterface
+
+open class ProxyApiSuperClass
+
+interface ProxyApiInterface
+
+@RequiresApi(25) class ClassWithApiRequirement
```
reidbaker, line 16: Does pigeon have a different minimum api level from Flutter? I know this is a test but was surprised to see the annotation and I dont see where we protect against this classes reference on api 24 and below.

packages/pigeon/test/kotlin/proxy_api_test.dart
```diff
@@ -0,0 +1,944 @@
...
+        final StringBuffer sink = StringBuffer();
+        const KotlinGenerator generator = KotlinGenerator();
+        generator.generate(
+          const KotlinOptions(),
+          root,
+          sink,
+          dartPackageName: DEFAULT_PACKAGE_NAME,
+        );
+        final String code = sink.toString();
+        final String collapsedCode = _collapseNewlineAndIndentation(code);
```
reidbaker, line 413: If this test is not generated consider adding how someone should update these strings.

reidbaker (approved):

tarrinneal:
@bparrishMines I think there needs to be some plumbing added to have the kotlin files generate using the tool/generate script

bparrishMines:
@tarrinneal What do you mean? I have been using `dart run tool/generate.dart` to generate the Kotlin code successfully. Is this the script you are referring to?

packages/pigeon/example/app/android/app/src/main/kotlin/dev/flutter/pigeon_example_app/Messages.g.kt
```diff
@@ -84,7 +84,7 @@ data class MessageData(
   }
 }
 
-private object MessagesPigeonCodec : StandardMessageCodec() {
+private open class MessagesPigeonCodec : StandardMessageCodec() {
```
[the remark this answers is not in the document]
reidbaker: Thanks that is not a level of kotlin nuance I would have known.

tarrinneal:
> @tarrinneal What do you mean? I have been using `dart run tool/generate.dart` to generate the Kotlin code successfully. Is this the script you are referring to?

yep, If it's working then don't worry about it :)

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTestApiImpls.kt
bparrishMines: This was added only to ensure that wrapping a class with an API requirement doesn't break the generated code or cause lint warnings. There is an example implementation that protects against it in `ProxyApiTestApiImpls.kt`: https://github.com/flutter/packages/pull/6371/files#diff-87909fbd5b2dd9a1652c6dfd584c2ff3f387a4d4a8f4a00e32e553b6aacb6272R691.

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -921,6 +1324,667 @@ if (wrapped == null) {
...
+          _writeMethodDeclaration(
+            indent,
+            name: 'get$hostProxyApiPrefix${api.name}',
+            isAbstract:
+                api.hasAnyHostMessageCalls() || api.unattachedFields.isNotEmpty,
+            isOpen:
+                !api.hasAnyHostMessageCalls() && api.unattachedFields.isEmpty,
+            documentationComments: <String>[
+              'An implementation of [$hostProxyApiPrefix${api.name}] used to add a new Dart instance of',
+              '`${api.name}` to the Dart `InstanceManager`.'
```
tarrinneal, line 1398: I'm not going to mark all of them, but here's another

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -921,6 +1324,667 @@ if (wrapped == null) {
...
+    required String dartPackageName,
+    required String fullKotlinClassName,
+  }) {
+    indent.writeln('@Suppress("LocalVariableName")');
+    indent.writeScoped(
+      'fun setUpMessageHandlers(binaryMessenger: BinaryMessenger, api: $kotlinApiName?) {',
+      '}',
+      () {
+        indent.writeln(
+          'val codec = api?.pigeonRegistrar?.codec ?: StandardMessageCodec()',
```
tarrinneal, line 1590: Everything should use the PigeonCodec instead of the StandardCodec.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTestApiImpls.kt
tarrinneal: Why was this change necessary?

tarrinneal (changes requested):
A few more things here, should be easy.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTestApiImpls.kt
bparrishMines: It made the feature mentioned in another comment possible: https://github.com/flutter/packages/pull/6371#discussion_r1536177357

Essentially, there wasn't a way for me to retrieve the constructor values in the `ProxyApiBaseCodec` for a class if the original `dartDefaultConstructor` was used.

packages/pigeon/platform_tests/shared_test_plugin_code/lib/integration_tests.dart
```diff
@@ -2230,3 +2879,142 @@ class _SmallFlutterApi implements FlutterSmallApi {
     return msg;
   }
 }
+
+ProxyApiTestClass _createGenericProxyApiTestClass({
```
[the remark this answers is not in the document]
bparrishMines: What do you mean by using a builder? As in the builder pattern that is typically used in Java?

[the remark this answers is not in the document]
reidbaker: Yes but I wouldnt call it a java pattern. On my first pass I thought these were positional args but these are named parameters and that is much more dart-y. Ignore this feedback. 

 https://en.wikipedia.org/wiki/Builder_pattern

tarrinneal (approved):
I think it's at a place where we can land it and start making incremental changes as needed. Good job with it all, and thanks for your patience!

PR #6372, opened 2024-03-21
[pigeon] Adds @SwiftClass annotation

tarrinneal:
Adds annotation for @SwiftClass which will cause the swift generator to create a `class` instead of a `struct` in the generated Swift code. 

This will allow for recursive classes as well as allow for Objc interop, without forcing users to lose the potential benefits of structs if they prefer them instead.

Also creates recursive data class integration test to check for coverage on all languages.

fixes https://github.com/flutter/flutter/issues/145175

## Pre-launch Checklist

- [x] I signed the [CLA].


tarrinneal:
@LouiseHsu Do you also need the ability to add the `@objc` attribute to generated classes?

packages/pigeon/pigeons/core_tests.dart
```diff
@@ -48,6 +48,7 @@ class AllTypes {
 }
 
 /// A class containing all supported nullable types.
+@SwiftClass()
```
hellohuanlin, line 51: should we test both class and struct cases?

tarrinneal: yes, but that will delay this pr from landing as the testing is extensive. I'm already working on it though.

LouiseHsu:
> @LouiseHsu Do you also need the ability to add the `@objc` attribute to generated classes?

Yes, that would be helpful!

tarrinneal:
> > @LouiseHsu Do you also need the ability to add the `@objc` attribute to generated classes?
> 
> Yes, that would be helpful! <3

For api classes and data classes? or just data classes?

LouiseHsu:
> > > @LouiseHsu Do you also need the ability to add the `@objc` attribute to generated classes?
> > 
> > 
> > Yes, that would be helpful! <3
> 
> For api classes and data classes? or just data classes?

I think just data classes will be good for now :)

tarrinneal:
Might be best to combine the two annotations (the inheritance being optional within the class one)

tarrinneal:
> Might be best to combine the two annotations (the inheritance being optional within the class one)

Done.

@stuartmorgan I'm hoping we can push this through in the morning, if you have time to review it.

packages/pigeon/platform_tests/test_plugin/windows/pigeon/core_tests.gen.h
```diff
@@ -279,6 +284,129 @@ class AllNullableTypes {
   std::optional<AnEnum> a_nullable_enum_;
   std::optional<std::string> a_nullable_string_;
   std::optional<flutter::EncodableValue> a_nullable_object_;
+  std::optional<AllNullableTypes> all_nullable_types_;
```
stuartmorgan, line 287: Oof. This won't work; it's the same problem as what you're fixing in Swift, basically. `std::optional` stores the actual class, not a pointer to the class, so the `AllNullableTypes` class now has infinite size.

I'll work on adjusting the C++ generators to use `std::unique_ptr` internally for any custom class storage.

stuartmorgan (changes requested):

packages/pigeon/lib/ast.dart
```diff
@@ -639,6 +639,8 @@ class Class extends Node {
   Class({
     required this.name,
     required this.fields,
+    this.isSwiftObjcInteropClass = false,
```
hellohuanlin, line 642: nit: `isNSObjectSubclass`

packages/pigeon/lib/cpp_generator.dart
```diff
@@ -1169,15 +1205,79 @@ return EncodableValue(EncodableList{
...
+  }
+
+  /*
+  
+      final String className = classDefinition.name;
+      _writeFunctionDeclaration(indent, className,
+          isConstructor: true, parameters: <String>['const $className& other']);
+      _writeFunctionDeclaration(indent, 'operator=',
+          returnType: '$className&',
+          parameters: <String>['const $className& other']);
```
hellohuanlin, line 1259: delete these?

packages/pigeon/lib/swift_generator.dart
```diff
@@ -125,11 +125,31 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
         indent, classDefinition.documentationComments, _docCommentSpec,
         generatorComments: generatedComments);
 
-    indent.write('struct ${classDefinition.name} ');
+    if (classDefinition.isSwiftClass) {
+      String inheritNSObject = '';
+      if (classDefinition.isSwiftObjcInteropClass) {
+        inheritNSObject = ': NSObject';
+        indent.writeln('@objc');
```
hellohuanlin, line 132: iirc `@objc` isn't necessary since it's already a subclass of `NSObject`, which is already exposed to objc runtime.

packages/pigeon/README.md
```diff
@@ -27,6 +27,10 @@ Custom classes, nested datatypes, and enums are also supported.
 
 Nullable enums in Objective-C generated code will be wrapped in a class to allow for nullability.
 
+By default, custom classes in Swift are defined as structs. 
+If you need classes instead - for recursive data, or Objective-C interop -
+use the @SwiftClass annotation when defining the class.
```
hellohuanlin, line 32: Can we add a comment explaining why we need this? (e.g. struct doesn't support recursive definition)

hellohuanlin (approved):
I like the idea of using annotation, because it can get complicated to detect recursive definition and decide `struct` vs `class` for developers.

packages/pigeon/lib/swift_generator.dart
hellohuanlin: Oh in addition to above, we probably need to add `@objc` for all custom members that are not inherited from `NSObject` (basically all the properties and the `init` function). 

Can you also write some ObjC tests to make sure it works?

hellohuanlin (changes requested):

packages/pigeon/README.md
tarrinneal: it is noted there already, is it not clear enough?

packages/pigeon/lib/swift_generator.dart
tarrinneal: all properties? even primitives? or only custom types?

packages/pigeon/README.md
hellohuanlin: > If you need classes instead - for recursive data, 
Can you add a note that Swift `struct` doesn't support recursively defined data type. 

> or Objective-C interop -
> use the @SwiftClass annotation when defining the class.

using `SwiftClass` isn't enough, we also need it to be `NSObject` subclass.

packages/pigeon/lib/swift_generator.dart
hellohuanlin: All properties and methods that are not inherited from `NSObject` need to be annotated, in order for them to be accessibly by objc runtime. The class definition itself and its members inherited from `NSObject` are already implicitly annotated.

hellohuanlin: So yes, even the properties of primitive types.

stuartmorgan (approved):
LGTM once @hellohuanlin's concerns are addressed. (Marking now since I'll be OOO.)

tarrinneal:
@hellohuanlin I removed the @objc feature for now to unblock this pr. I'll throw up a new pr with just that in a short while.

hellohuanlin (approved):
As discussed offline, feel free to land this first and create a separate PR for objc interop.

tarrinneal:
Landing to fix the tree

PR #6384, opened 2024-03-25
[flutter_markdown] Ensure customize nested bullet list style.

Kurogoma4D:
See https://github.com/flutter/flutter/issues/145670 .

## Pre-launch Checklist

- [x] I signed the [CLA].


domesticmouse (approved):
PTAL @stuartmorgan

stuartmorgan:
> * [ ]  I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning), or this PR is [exempt from version changes](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates).
> * [ ]  I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style).

CI will not pass without completing these steps; that's why `repo_checks` is red.

Kurogoma4D:
I've updated them (I'm new to contribute this package, so I'm not sure this versioning is correct).

tarrinneal:
@domesticmouse This is a breaking change isn't it?

Kurogoma4D:
So should I increase the minor version?

tarrinneal:
> So should I increase the minor version?

can you just make the parameter named and optional?

tarrinneal:
> > @domesticmouse This is a breaking change isn't it?
> 
> Do we follow SemVer in these packages? If we do, then yeah it is.

We're generally stricter than that. No breaking changes, we would deprecate this method and add a new one. I'm not sure if that's the policy for this package. @stuartmorgan will know.

stuartmorgan:
> We're generally stricter than that. No breaking changes, we would deprecate this method and add a new one. I'm not sure if that's the policy for this package.

That's for platform interface packages. For regular packages we can do breaking changes (with a corresponding major version bump), but we generally evaluate the importance of the use case and whether we need to do it with a breaking change or there is a less disruptive option.

tarrinneal:
> That's for platform interface packages. 

Right. 

> So should I increase the minor version?

@Kurogoma4D That should be fine. I think change is worth the breaking change.

You should also make a note in the changelog about what this pr breaks https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog

packages/flutter_markdown/lib/src/widget.dart
```diff
@@ -46,7 +46,11 @@ typedef MarkdownCheckboxBuilder = Widget Function(bool value);
 /// Signature for custom bullet widget.
 ///
 /// Used by [MarkdownWidget.bulletBuilder]
-typedef MarkdownBulletBuilder = Widget Function(int index, BulletStyle style);
+typedef MarkdownBulletBuilder = Widget Function(
+  int index,
+  BulletStyle style,
+  int nestLevel,
```
stuartmorgan, line 52: I haven't done a full review here, but I assume the breaking change is that we're adding a parameter to a method that package clients need to implement?

If so, it would be better to make the breaking change replace it with a parameter object, so that next time it doesn't need a breaking change.

Kurogoma4D: I agree.
Which is better to use class or record, as a parameter object here?

stuartmorgan: Adding a field to a record later would be breaking (since it changes the type), so it needs to be a class to get the benefit.

Kurogoma4D:
@domesticmouse @stuartmorgan
Could you review again?

domesticmouse (approved):
PTAL @stuartmorgan

toan88291:
> @domesticmouse @stuartmorgan Could you review again?

"This branch has conflicts that must be resolved"
I am also having this problem, can you solve it sooner.

domesticmouse:
Kurogoma4D, I think I've resolved the conflicts appropriately, please check. 

@tarrinneal @stuartmorgan PTAL

tarrinneal (approved):

Kurogoma4D:
@domesticmouse Thanks, LGTM :)

PR #6403, opened 2024-03-26
[in_app_purchase] Replaced deprecated ProrationMode with ReplacementMode

vongrad:
Replace deprecated `ProrationMode` in Android's billing client with `ReplacementMode`. 

https://github.com/flutter/flutter/issues/128957

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/.ci/legacy_project/all_packages/android/app/build.gradle
+++ b/.ci/legacy_project/all_packages/android/app/build.gradle
@@ -29,7 +29,7 @@ android {
 
     defaultConfig {
         applicationId "com.example.all_packages"
-        minSdkVersion 16
+        minSdkVersion flutter.minSdkVersion
         targetSdkVersion 30
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
--- a/packages/espresso/CHANGELOG.md
+++ b/packages/espresso/CHANGELOG.md
@@ -1,6 +1,7 @@
-## NEXT
+## 0.3.0+8
 
-* Updates minimum supported SDK version to Flutter 3.13/Dart 3.1.
+* Updates minSdkVersion to 19.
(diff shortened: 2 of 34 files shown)


vongrad:
I had to update the versions in both `in_app_purchase` and `in_app_purchase_android` pubspec.yaml files, but now the in_app_purchase cannot resolve the in_app_purchase_android ^0.3.3 dependency when running your pipeline checks. What would be the right approach?

bparrishMines (commented):
@vongrad It looks like you are changing the name of the `ProrationMode` class. This is a breaking change for the Android implementation, so it would need a major version bump (e.g. 0.4.0) if that is the goal of the PR. 

Also, you are bumping the android implementation in `packages/in_app_purchase/in_app_purchase/pubspec.yaml`, but that version doesn't exist on `pub.dev` until this PR would land.

But overall I think we should actually deprecate `ProrationMode` on the Dart side and add `ReplacementMode` as a separate enum. And add it as a separate parameter in `launchBillingFlow`. This would give users time to transition without breaking them immediately. 

Since this looks like it is only changing an Android specific method, you shouldn't need to change `in_app_purchase` in this PR.

vongrad:
@bparrishMines I agree that its probably better to introduce a new enum for `ReplacementMode` instead of placing it in as a replacement. 
In this case, I suppose I should leave the version of `in_app_purchase` untouched and only increase the version of `in_app_purchase_android`. What would you prefer to have as a version number?

bparrishMines:
Since this is adding a new feature, this should bump `in_app_purchase_android` to `0.3.3`.

vongrad:
Sure thing, I will submit a new PR during next week once I get some free time.

vongrad:
I made a new PR with the mentioned changes and in general re-worked it a bit. Please see here: https://github.com/flutter/packages/pull/6515

PR #6404, opened 2024-03-26
[google_sign_in_ios] Upgrade GoogleSignIn iOS SDK to 7.1

jmagman:
1. Update GoogleSignIn iOS SDK dependency to [7.1](https://github.com/google/GoogleSignIn-iOS/releases/tag/7.1.0), which supports privacy manifests.
1. Update "fetch deps step" to run `pod repo update` every time to avoid missing spec failure. Also added a fetch deps step to the all_packages builds, since those could also use a `pod repo update` but I didn't want to add more logic to those bash scripts.
```
     [!] CocoaPods could not find compatible versions for pod "GoogleSignIn":
       In Podfile:
         google_sign_in_ios (from `Flutter/ephemeral/.symlinks/plugins/google_sign_in_ios/darwin`) was resolved to 0.0.1, which depends on
           GoogleSignIn (~> 7.1)
     None of your spec sources contain a spec satisfying the dependency: `GoogleSignIn (~> 7.1)`.
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8752192509749702705/+/u/Run_package_tests/download_Dart_and_iOS_deps/stdout

3. Looks like the new versions of [`GTMSessionFetcher`](https://github.com/CocoaPods/Specs/blob/master/Specs/c/e/3/GTMSessionFetcher/3.3.2/GTMSessionFetcher.podspec.json) and [`AppAuth`](https://github.com/CocoaPods/Specs/blob/master/Specs/b/b/9/AppAuth/1.7.3/AppAuth.podspec.json) don't define modules, so there's an error building it statically:
```
     [!] The following Swift pods cannot yet be integrated as static libraries:
     The Swift pod `GTMAppAuth` depends upon `GTMSessionFetcher` and `AppAuth`, which do not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
```
I filed https://github.com/google/gtm-session-fetcher/issues/384 and https://github.com/openid/AppAuth-iOS/issues/844.

In the meantime, I updated the example apps to `use_frameworks!`, which would be on for Flutter Swift apps, but not Objective-C ones.  We could add something to the tool to detect this case, and suggest `use_frameworks!` be added?  

4.  Even though google_sign_in_ios does not contain Swift files, for some reason, there is a "pod lib lint" warning complaining `swift_version` isn't set. This seems related to `GTMAppAuth` dependency constraint that went from an Objective-C-only to Swift pod.  So I set `swift_version` since it's harmless.

```
- WARN  | swift: The validator used Swift `4.0` by default because no Swift version was specified. To specify a Swift version during validation, add the `swift_versions` attribute in your podspec. Note that usage of a `.swift-version` file is now deprecated.
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8752100979634945505/+/u/Run_package_tests/validate_iOS_and_macOS_podspecs/stdout

## Issues

* Fixes https://github.com/flutter/flutter/issues/145777
* Fixes https://github.com/flutter/flutter/issues/145866
* See also https://github.com/flutter/flutter/issues/137140

## Pre-launch Checklist

- [x] I signed the [CLA].


ditman:
Thanks for updating the `extension_google_sign...` package too. I think these changes need a pubspec/CHANGELOG update, so we publish them to pub.dev, right?

jmagman:
> Nope, the extension is only updating the iOS code for the example app.

The repo_checks look happy.  I can add a NEXT if you want though.

ditman:
> > Nope, the extension is only updating the iOS code for the example app.
> 
> The repo_checks look happy. I can add a NEXT if you want though.

Nope, no need, I thought that something more significant had changed. If users don't need to know about it, no need to put it in a changelog!

ditman (approved):
This LGTM, but @vashworth or @stuartmorgan's reviews are more valuable than mine for sure :)

script/tool/lib/src/fetch_deps_command.dart
```diff
@@ -79,22 +80,22 @@ class FetchDepsCommand extends PackageLoopingCommand {
...
     // https://github.com/flutter/flutter/blob/fb7a763c640d247d090cbb373e4b3a0459ac171b/packages/flutter_tools/bin/podhelper.rb#L130
-    if (getBoolArg(platformIOS)) {
-      final int exitCode = await processRunner.runAndStream(
+    final bool precacheIos = getBoolArg(platformIOS);
+    final bool precacheMacOS = getBoolArg(platformMacOS);
+    if (precacheIos || precacheMacOS) {
+      final int precacheExitCode = await processRunner.runAndStream(
         flutterCommand,
-        <String>['precache', '--ios'],
+        <String>['precache', '--${precacheIos ? 'ios' : 'macos'}'],
```
vashworth, line 88: Although it doesn't seem that any uses of "fetch-deps" does this currently, it is possible for both `getBoolArg(platformIOS)` and `getBoolArg(platformMacOS)` to be true. Perhaps would be better to make it a function and pass in the platform for each?

script/tool/lib/src/fetch_deps_command.dart
```diff
@@ -79,22 +80,22 @@ class FetchDepsCommand extends PackageLoopingCommand {
     // `pod install` requires having the platform artifacts precached. See
     // https://github.com/flutter/flutter/blob/fb7a763c640d247d090cbb373e4b3a0459ac171b/packages/flutter_tools/bin/podhelper.rb#L47
     // https://github.com/flutter/flutter/blob/fb7a763c640d247d090cbb373e4b3a0459ac171b/packages/flutter_tools/bin/podhelper.rb#L130
-    if (getBoolArg(platformIOS)) {
-      final int exitCode = await processRunner.runAndStream(
+    final bool precacheIos = getBoolArg(platformIOS);
```
stuartmorgan, line 83: precacheIOS

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#capitalize-identifiers-consistent-with-their-spelling

script/tool/lib/src/fetch_deps_command.dart
stuartmorgan: IIRC the tool will accept multiple platforms, in which case this can be:
```dart
<String>[
  'precache',
  if (precacheIOS)
    '--ios',
  if (precacheMacOS)
    '--macos',
],
```

stuartmorgan (approved):
LGTM

And it looks like the module issue was fixed upstream (and published?), so hopefully not many people will hit that.

.ci/targets/ios_build_all_packages.yaml
```diff
@@ -2,6 +2,10 @@ tasks:
   - name: prepare tool
     script: .ci/scripts/prepare_tool.sh
     infra_step: true # Note infra steps failing prevents "always" from running.
+  - name: download Dart and iOS deps
+    script: .ci/scripts/tool_runner.sh
+    args: ["fetch-deps", "--ios", "--supporting-target-platforms-only"]
```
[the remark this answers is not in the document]
jmagman: Hm, looks like `macos_platform_tests master - packages` was flaky at the edge of taking an hour and being killed, and this change is pushing it more reliably to time out:

Timeout on this PR: https://ci.chromium.org/ui/p/flutter/builders/try/Mac_arm64%20macos_platform_tests%20master%20-%20packages/12077/overview

An example of a timeout in prod https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_arm64%20macos_platform_tests%20master%20-%20packages/1659/overview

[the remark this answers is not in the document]
jmagman: Filed https://github.com/flutter/flutter/issues/146254

.ci.yaml
```diff
@@ -1091,9 +1091,10 @@ targets:
   # TODO(stuartmorgan): Remove "- packages" from all task names once
   # flutter/plugins is merged into this repo and turned down; it's here only
   # because names must be unique across all flutter repositories.
-  - name: Mac_arm64 macos_platform_tests master - packages
+  - name: Mac_x64 macos_platform_tests master - packages
```
jmagman, line 1094: For science, I wonder if these are hanging because they are on arm bots...

jmagman:
> And it looks like the module issue was fixed upstream (and published?), so hopefully not many people will hit that.

Neither published yet...

.ci.yaml
stuartmorgan: We shouldn't change this without changing the build-all platform the other direction. And we should do it with new bots so we don't lose all macOS coverage for 1-2 days while this is in bringup.

stuartmorgan:
> > And it looks like the module issue was fixed upstream (and published?), so hopefully not many people will hit that.
> 
> Neither published yet...

Hm, so are we okay with everyone with Obj-C apps being broken and having an error message that doesn't give them the right solution? It seems like maybe we should wait a little longer to see if they get published soon.

jmagman:
> Hm, so are we okay with everyone with Obj-C apps being broken and having an error message that doesn't give them the right solution? It seems like maybe we should wait a little longer to see if they get published soon.

AppAuth published its fix https://github.com/openid/AppAuth-iOS/releases/tag/1.7.4
Still waiting on `GTMSessionFetcher`.

.ci/targets/ios_build_all_packages.yaml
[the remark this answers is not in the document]
jmagman: This is resolved.

.ci.yaml
jmagman: Reverted this change since the hang is fixed upstream and wasn't related to the arch anyway.

stuartmorgan:
https://github.com/CocoaPods/Specs/blob/df2cd82c7ff09bc5fa92d015a5f0a86c49931ac6/Specs/c/e/3/GTMSessionFetcher/3.4.0/GTMSessionFetcher.podspec.json#L27

🎉

packages/google_sign_in/google_sign_in_ios/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: google_sign_in_ios
 description: iOS implementation of the google_sign_in plugin.
 repository: https://github.com/flutter/packages/tree/main/packages/google_sign_in/google_sign_in_ios
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+google_sign_in%22
-version: 5.7.5
+version: 5.7.6
```
jmagman, line 5: Hm, do you think this should be a minor bump instead @stuartmorgan?

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios.podspec
```diff
@@ -15,12 +15,22 @@ Enables Google Sign-In in Flutter apps.
   s.source_files = 'Classes/**/*.{h,m}'
   s.public_header_files = 'Classes/**/*.h'
   s.module_map = 'Classes/FLTGoogleSignInPlugin.modulemap'
-  s.dependency 'GoogleSignIn', '~> 7.0.0'
+
+  # AppAuth is a transitive dependency of GoogleSignIn.
+  # Depend on 1.7.4 or higher, which defines modules.
+  s.dependency 'AppAuth', '~> 1.7.4'
```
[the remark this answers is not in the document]
jmagman: Done.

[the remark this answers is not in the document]
jmagman: Thanks for catching.  Done.

packages/google_sign_in/google_sign_in_ios/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -386,11 +386,19 @@
 			);
 			inputPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-Runner/Pods-Runner-resources.sh",
+				"${PODS_CONFIGURATION_BUILD_DIR}/AppAuth/AppAuthCore_Privacy.bundle",
+				"${PODS_CONFIGURATION_BUILD_DIR}/GTMAppAuth/GTMAppAuth_Privacy.bundle",
+				"${PODS_CONFIGURATION_BUILD_DIR}/GTMSessionFetcher/GTMSessionFetcher_Core_Privacy.bundle",
+				"${PODS_CONFIGURATION_BUILD_DIR}/GTMSessionFetcher/GTMSessionFetcher_Full_Privacy.bundle",
```
jmagman, line 392: Woo!

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios.podspec
```diff
@@ -16,10 +16,11 @@ Enables Google Sign-In in Flutter apps.
...
 
-  # AppAuth is a transitive dependency of GoogleSignIn.
-  # Depend on 1.7.4 or higher, which defines modules.
-  s.dependency 'AppAuth', '~> 1.7.4'
-  s.dependency 'GoogleSignIn', '~> 7.1'
+  # AppAuth and GTMSessionFetcher are GoogleSignIn transitive dependencies.
+  # Depend on versions which defines modules.
+  s.dependency 'AppAuth', '>= 1.7.4'
+  s.dependency 'GTMSessionFetcher', '>= 3.4.0'
+  s.dependency 'GoogleSignIn', '>= 7.1'
```
stuartmorgan, line 23: Isn't `~> 7.1` right for this one? We don't want to get 8.x without opting in because we do call its APIs directly.

packages/google_sign_in/google_sign_in_ios/pubspec.yaml
stuartmorgan: Since this one *should* be a no-op for clients, bugfix seems fine. I don't feel strongly between the two here.

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios.podspec
jmagman: My brain refuses to retain what `~>` does.  Done.

stuartmorgan: For the future: we could consider just never using it, since we all have to look it up every time, and manually do the ranges ourselves. But hopefully in the medium term we won't need podspecs at all.

stuartmorgan:
Merged to pull in the fix for my OOB maps breakage.

stuartmorgan:
🟢 🎉

ynmain:
@jmagman 
Hi.
I'm in trouble.
`google_sign_in_ios 5.7.6`
I just updated to the latest version
cocoapods cannot be updated due to MLKitVision dependencies.

Can we solve this problem?
https://github.com/firebase/flutterfire/issues/12742#issuecomment-2093003407

stuartmorgan:
@ynmain Please use [the issue tracker](https://github.com/flutter/flutter/issues) to submit issues, not comments on closed PRs.

The issue here is not with `google_sign_in`; see https://github.com/flutter-ml/google_ml_kit_flutter/issues/612

PR #6426, opened 2024-03-28
Fix unnecessary toList/fromList calls during encode/decode process

tarrinneal:
Fix unnecessary toList/fromList calls during encode/decode process.

also makes some kotlin code more idiomatic.

removes the ability to have collisions with the name `list`.

fixes https://github.com/flutter/flutter/issues/119351

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
> removes double encode/decode on custom classes.

Per offline discussion I think we should adjust the PR description here, as well as the changelog. It's not that things were double-encoding custom classes, it's that we were making the encoding logic needlessly complex by calling ToList/FromList directly instead of letting the encoder do the encoding internally.

(This PR will actually make the byte encoding slightly *less* efficient; before a custom class was encoded as `<list type marker><list size><list elements>` and now it will be `<custom class type marker><list type marker><list size><list elements>`. Maybe that's why it was originally written the other way, but I don't think that's a worthwhile optimization for the conceptual complexity it adds.)

packages/pigeon/lib/objc_generator.dart
```diff
@@ -1488,21 +1488,11 @@ String _makeObjcSignature({
 void generateObjcHeader(ObjcOptions options, Root root, Indent indent) {}
 
 String _listGetter(String list, NamedType field, int index, String? prefix) {
```
stuartmorgan, line 1490: This is only used in one place; now that it's one line we should just inline it at the call site. The string itself is pretty self-explanatory.

packages/pigeon/lib/swift_generator.dart
```diff
@@ -214,7 +214,7 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
           final String fieldName = field.name;
           final String nullsafe = field.type.isNullable ? '?' : '';
           if (field.type.isClass) {
```
stuartmorgan, line 216: This is now identical to the final `else`, just written slightly differently; they should be consolidated and normalized (either with everything using `fieldName`, or with `fieldName` eliminated in favor of `field.name`).

packages/pigeon/lib/swift_generator.dart
```diff
@@ -576,11 +578,10 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
...
       if (type.isClass) {
         indent.writeln('var $variableName: $fieldType? = nil');
-        indent
-            .write('if let ${variableName}List: [Any?] = nilOrValue($value) ');
+        indent.write(
+            'if let ${variableName}List: $fieldType = nilOrValue($value) ');
         indent.addScoped('{', '}', () {
-          indent.writeln(
-              '$variableName = $fieldType.fromList(${variableName}List)');
+          indent.writeln('$variableName = ${variableName}List as $fieldType');
```
stuartmorgan, line 584: Shouldn't these no longer have `List` in the name?

(Also, is this still actually different than the `else` branch?)

packages/pigeon/lib/swift_generator.dart
```diff
@@ -603,8 +604,7 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
       }
     } else {
       if (type.isClass) {
-        indent.writeln(
-            'let $variableName = $fieldType.fromList($value as! [Any?])!');
+        indent.writeln('let $variableName = $value as! $fieldType');
```
stuartmorgan, line 607: Is this different than the `else`?

packages/pigeon/pigeons/core_tests.dart
```diff
@@ -39,7 +39,7 @@ class AllTypes {
   Int64List a8ByteArray;
   Float64List aFloatArray;
   // ignore: always_specify_types, strict_raw_type
-  List aList;
+  List list;
```
stuartmorgan, line 42: I think it's worth commenting on why this name doesn't match all the other ones, so someone doesn't think it's a mistake a change it back later.

stuartmorgan (changes requested):

packages/pigeon/example/app/android/app/src/main/kotlin/dev/flutter/pigeon_example_app/Messages.g.kt
```diff
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 // Autogenerated from Pigeon, do not edit directly.
 // See also: https://pub.dev/packages/pigeon
+@file:Suppress("UNCHECKED_CAST", "LocalVariableName", "ArrayInDataClass")
```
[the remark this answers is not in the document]
tarrinneal: most of them are full file regardless, it gives "redundant suppression" errors.

stuartmorgan:
Now I remember why this error looked familiar: it's a bug in VS 2019, which was fixed in VS 2022.

We actually shouldn't still be using VS 2019 in CI, but it's going to be tricky to update that right now so I'll dig up the workaround. IIRC it will just require turning on exceptions for the test app.

tarrinneal:
> Now I remember why this error looked familiar: it's [a bug in VS 2019, which was fixed in VS 2022](https://developercommunity.visualstudio.com/t/stdany-doesnt-link-when-exceptions-are-disabled/376072).
> 
> We actually shouldn't still be using VS 2019 in CI, but it's going to be tricky to update that right now so I'll dig up the workaround. IIRC it will just require turning on exceptions for the test app.

Seems like it may be something else still...

stuartmorgan:
> > Now I remember why this error looked familiar: it's [a bug in VS 2019, which was fixed in VS 2022](https://developercommunity.visualstudio.com/t/stdany-doesnt-link-when-exceptions-are-disabled/376072).
> > We actually shouldn't still be using VS 2019 in CI, but it's going to be tricky to update that right now so I'll dig up the workaround. IIRC it will just require turning on exceptions for the test app.
> 
> Seems like it may be something else still...

Or it is that same problem, but I put the workaround on the example app binary when the failure is coming from a unit test file that's not compiled in that binary, and I need to change the build settings of the thing that was actually failing to compile 🤦🏻

packages/pigeon/lib/swift_generator.dart
```diff
@@ -969,6 +945,9 @@ String _swiftTypeForDartType(TypeDeclaration type) {
 
 String _nullsafeSwiftTypeForDartType(TypeDeclaration type) {
   final String nullSafe = type.isNullable ? '?' : '';
+  if (type.baseName == 'Map') {
```
stuartmorgan, line 948: Maybe a comment in the code here for why Map needs special handling? It's not obvious to me what this change is for.

stuartmorgan (approved):
LGTM! Just a few minor comments about possible simplifications.

PR #6438, opened 2024-03-29
[file_selector_android] Modifies `getDirectoryPath`, `openFile`, `openFiles` to return file/directory paths instead of URIs

camsim99:
## Overview

Changes `getDirectoryPath`, `openFile`, and `openFiles` to return the directory path instead of the URI representing the directory or file.

Fixes https://github.com/flutter/flutter/issues/141250.

## Technical Details

Technically, `getDirectoryPath` returns the path of the selected directory directly, whereas `openFile` and `openFiles` actually copy the content that the URI points to into another file and returns the path of the file. This is a direct result of the fact that Android **highly discourages** folks from accessing files through direct paths, so it's extremely difficult to do reliably without this approach (can confirm by how much effort it took to get this PR out...).

This approach [is also used in `image_picker_android`](https://github.com/flutter/packages/blob/main/packages/image_picker/image_picker_android/android/src/main/java/io/flutter/plugins/imagepicker/FileUtils.java) and this PR uses essentially the same logic.

I think we should move forward with this short term solution to unblock folks that need access to directories and files in Dart, but I strongly advise that we re-evaluate for the long term. I elaborate on this in [this doc](https://docs.google.com/document/d/12pJDtl0yubyc68UqKo2hQ7XJwv86_ixCjNemQ7ENCBQ/edit?usp=sharing) and welcome feedback!

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileSelectorApiImpl.java
```diff
@@ -204,7 +206,16 @@ public void getDirectoryPath(
             public void onResult(int resultCode, @Nullable Intent data) {
               if (resultCode == Activity.RESULT_OK && data != null) {
                 final Uri uri = data.getData();
-                result.success(uri.toString());
+                final Uri docUri =
```
gmackall, line 209: The `uri` we get in the line above [can be null](https://developer.android.com/reference/android/content/Intent#getData()), so I think we should handle that error case as well (I believe this [has caused issues](https://github.com/flutter/flutter/issues/104168) in `image_picker`)

Also, it wasn't added in this PR but it looks like the intent itself is nullable? So the line above where we get the `uri` could also produce a NullPointerException, which would be good to handle as well.

gmackall (commented):
Mostly lgtm, one comment on potential null pointer exceptions. I mostly skipped over the `FileUtils.java` and corresponding test because I'm assuming its copied over from image picker, let me know if thats not the case.

camsim99:
@gmackall to respond to your comment and clarify for any reviewers, the `FileUtils.java` that I added is different from that used in image_picker in the following ways:

1. I wrote the `getPathFromUri` method, so please review!
2. `getPathFromCopyOfFileFromUri` is the same logic used in image_picker except I removed image-specific assumptions (the only notable one being an assumption about the extension for files being .jpeg if it could not otherwise be found).

In short, a review on the whole file wouldn't hurt!

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -0,0 +1,198 @@
...
+
+public class FileUtils {
+
+  /** URI authority that represents access to external storage providers. */
+  public static final String EXTERNAL_DOCUMENT_AUTHORITY = "com.android.externalstorage.documents";
+
+  /**
+   * Retrieves path of directory represented by the specified {@code Uri}.
+   *
+   * <p>Intended to handle any cases needed to return paths from URIS retrieved from open
```
gmackall, line 51: nit 
```suggestion
   * <p>Intended to handle any cases needed to return paths from URIs retrieved from open
```

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -0,0 +1,198 @@
...
+   * Intent.ACTION_OPEN_FILES}, or {@code Intent.ACTION_OPEN_DOCUMENT_TREE}.
+   *
+   * <p>Will return the path for on-device directories, but does not handle external storage
+   * volumes.
+   */
+  @NonNull
+  public static String getPathFromUri(@NonNull Context context, @NonNull Uri uri) {
+    String uriAuthority = uri.getAuthority();
+
+    if (uriAuthority.equals(EXTERNAL_DOCUMENT_AUTHORITY)) {
```
gmackall, line 62: `uriAuthority` can be null, I'd recommend swapping the order here to handle that case automatically
so
```suggestion
    if (EXTERNAL_DOCUMENT_AUTHORITY.equals(uriAuthority)) {
```

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -0,0 +1,198 @@
...
+      String uriDocumentId = DocumentsContract.getDocumentId(uri);
+      String documentStorageVolume = uriDocumentId.split(":")[0];
```
gmackall, line 64: Preference for splitting this once, storing it, checking the length, and handling that case (instead of splitting multiple times and assuming the length is what we need)

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -0,0 +1,198 @@
...
+   * <p>Each file is placed in its own directory to avoid conflicts according to the following
+   * scheme: {cacheDir}/{randomUuid}/{fileName}
+   *
+   * <p>File extension is changed to match MIME type of the file, if known. Otherwise, the extension
+   * is left unchanged.
+   *
+   * <p>If the original file name is unknown, a predefined "file_selector" filename is used and the
+   * file extension is deduced from the mime type.
+   *
+   * <p>Will return null if closing the output stream when done copying file contents does not for
```
gmackall, line 99: I'm not sure I follow the first case here "when done copying file contents does not for sure complete", can you clarify?

gmackall (commented):
Added some more comments on `FileUtils.java`!

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -0,0 +1,198 @@
...
+   */
+  @NonNull
+  public static String getPathFromUri(@NonNull Context context, @NonNull Uri uri) {
+    String uriAuthority = uri.getAuthority();
+
+    if (EXTERNAL_DOCUMENT_AUTHORITY.equals(uriAuthority)) {
+      String uriDocumentId = DocumentsContract.getDocumentId(uri);
+      String documentStorageVolume = uriDocumentId.split(":")[0];
+
+      // Non-primary storage volumes come from SD cards, USB drives, etc. and are
```
reidbaker, line 66: Where are they handled? Also can you include a link to your source that "primary" is the correct string.

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -0,0 +1,198 @@
...
+   * <p>If the original file name is unknown, a predefined "file_selector" filename is used and the
+   * file extension is deduced from the mime type.
+   *
+   * <p>Will return null if closing the output stream when done copying file contents does not for
+   * sure complete or if a security exception is encountered when opening the input stream.
+   */
+  @Nullable
+  public static String getPathFromCopyOfFileFromUri(@NonNull Context context, @NonNull Uri uri) {
+    try (InputStream inputStream = context.getContentResolver().openInputStream(uri)) {
+      String uuid = UUID.randomUUID().toString();
```
reidbaker, line 105: Testing random things can be hard consider having a way to set a seed that can be used in tests.

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -0,0 +1,198 @@
...
+    try {
+      if (uriFile.getScheme().equals(ContentResolver.SCHEME_CONTENT)) {
+        final MimeTypeMap mime = MimeTypeMap.getSingleton();
+        extension = mime.getExtensionFromMimeType(context.getContentResolver().getType(uriFile));
+      } else {
+        extension =
+            MimeTypeMap.getFileExtensionFromUrl(
+                Uri.fromFile(new File(uriFile.getPath())).toString());
+      }
+    } catch (Exception e) {
```
reidbaker, line 156: Is there a more scoped exception you can catch here? If not eating an exception then not logging can make for some difficult to debug code.

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
camsim99: Changed this to a UUID based on the URI and updated FileSelectorAndroidTest to check for the expected path since we can now reproduce it in tests!

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
camsim99: "primary" is the correct String because [this constant](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/provider/DocumentsContract.java;l=255?q=Documentscont&ss=android%2Fplatform%2Fsuperproject%2Fmain) is used as the root of the document ID. I verified in my own testing and looking through examples of getting the getting paths from URIs. There is no documentation about it probably because Android does not want folks accessing files this way.

I did not handle those cases because I honestly didn't think we needed to handle these cases in this plugin, but now that you mention it, there was never that restriction. I would prefer to leave this as a TODO and file an issue, though, because I would have to figure out how to test this myself and do some similar deducing since there's no documentation and wouldn't like to block this PR on that but let me know what you think.

I think that this entire comment is demonstrative of  why we should prioritize finding a longer term solution that isn't reliant on file paths BTW :) it will be hard for us to find future-proof solutions and handle all cases with this approach.

gmackall (approved):
lgtm!

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
reidbaker: You should link that constant in this code because that will be super hard to find later.

reidbaker: Also my understanding of this api is that a user is picking something and the host app has no control over what a user picks. Throwing an exception doesn't feel wrong but it seems odd to not give a caller a way to control or filter what users can see to the files that the caller can actually access.

reidbaker (approved):
One philosophy question about how we should signal to callers what to do. But the code looks good.

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
camsim99: You make a good point. In theory, the host app should not be tasked to control what kind of files the user picks and the plugin should just take care of it. So, the exception is meant to signal an issue with the plugin for users to report to us to handle. I don't love that solution but I do not feel like silently failing is much better since we won't find out as fast about the issues. Maybe we can catch the exception and send back a dummy value to the Flutter app that signifies an error was encountered?

Again, this issue is specific to this approach of using file paths because Android should handle all relevant file types in its storage framework.

reidbaker: Let me back up to a more simple statement. If a developer uses this plugin on android can their user select files that would cause us to throw this exception? My understanding of the codepaths is yes, that the file selector utility lets people pick things from anywhere on their device. 

I get that a silent failure is not ok. What I am hearing in your answer is that the caller cannot avoid these errors but also they are outside of what you think the plugin can handle in this pr? Is that correct?

camsim99: > If a developer uses this plugin on android can their user select files that would cause us to throw this exception? 

You're right -- theoretically yes, but we do not expect it. This is only the answer because I am not on the Android team with this sort of insight and no one else is besides them.

> What I am hearing in your answer is that the caller cannot avoid these errors but also they are outside of what you think the plugin can handle in this pr? Is that correct?

Yes -- I argue that we spend more time on a long term solution. There's no further progress I can make on this PR to avoid those errors until they are encountered, unfortunately.

reidbaker: Ok those seem to be reasonable arguments. Merge away

PR #6446, opened 2024-04-01
[Tool] Add ability to check dependencies independently of dev-dependencies, exclude integration_test from dependencies

reidbaker:
Create a linter that ensures that `integration_test` is not used in dependencies. 

Will be paired with a change to documentation 
```
If you are considering adding an external dependency:

Consider other options, and discuss with #hackers-ecosystem in Discord.
* If you add a dev_dependency on an external package, pin it to a specific version if at all possible.
* If you add a dependency on an external package in an example/, pin it to a specific version if at all possible.
* Some dependencies should only be linked as dev dependencies like integration_test 
```

Related to flutter/flutter/issues/145992
## Pre-launch Checklist

- [x] I signed the [CLA].


script/tool/lib/src/pubspec_check_command.dart
```diff
@@ -542,17 +542,25 @@ class PubspecCheckCommand extends PackageLoopingCommand {
...
+    // Dev dependencies
+    for (final Map<String, Dependency> dependencies
+        in <Map<String, Dependency>>[pubspec.devDependencies]) {
+      dependencies.forEach((String name, Dependency dependency) {
+        if (!_shouldAllowDevDependency(name, dependency)) {
+          badDependencies.add(name);
+        }
+      });
+    }
+
```
stuartmorgan, line 563: I think it would be less code/complexity if you left the old structure in place and then just added a second section here:
```dart
// Ensure that dev-only dependencies aren't in `dependencies`.
const List<String> devOnlyDependencies = <String>['integration_test'];
pubspec.dependencies.forEach((String name, Dependency dependency) {
  if (devOnlyDependencies.contains(name)) {
    badDependencies.add(name);
  }
});
```

(In fact, that last loop probably condenses to something like `badDependencies.addAll(pubspec.dependencies.keys.where(devOnlyDependencies.contains));`

reidbaker: Done, Left in the loop form though because I find it easier to read.

stuartmorgan (approved):
LGTM

script/tool/lib/src/pubspec_check_command.dart
```diff
@@ -553,6 +554,23 @@ class PubspecCheckCommand extends PackageLoopingCommand {
         }
       });
     }
+
+    // Ensure that dev-only dependencies aren't in `dependencies`.
+    const List<String> devOnlyDependencies = <String>['integration_test'];
+    // pigeon/platform_tests/shared_test_plugin_code is allowed to violate
+    // the dev only dependencies rule beause pidgeon has generated tests that
+    // are intended to ship to customers.
+    if (pubspec.name == 'shared_test_plugin_code') {
```
reidbaker, line 563: @stuartmorgan requesting re-review because I am not a fan of how I exempted pidgeon from the new check.

stuartmorgan: You can just conditionalize this whole check on being publishable, and remove the warning:
https://github.com/flutter/packages/blob/dcb4f853b2476d5106aabe88186b1fad8ad623b4/script/tool/lib/src/pubspec_check_command.dart#L209

script/tool/lib/src/pubspec_check_command.dart
```diff
@@ -553,6 +554,19 @@ class PubspecCheckCommand extends PackageLoopingCommand {
         }
       });
     }
+
+    // Ensure that dev-only dependencies aren't in `dependencies`.
+    const List<String> devOnlyDependencies = <String>['integration_test'];
+    // non published packages like pidgeon subpackages are allowed to violate
```
stuartmorgan, line 560: Non-published

stuartmorgan (approved):
LGTM with one minor comment-comment.

script/tool/lib/src/pubspec_check_command.dart
reidbaker: fixed

ekuleshov: @stuartmorgan @reidbaker should this same lint also be applied to dependencies like `test` or `flutter_test` ?

stuartmorgan: Probably; we'd need an audit of the repo (e.g., by adding them and running the script locally) to see if we have any legitimate cases for it. I'd be happy to review a PR.

ekuleshov: @stuartmorgan what is the command to run updated `pubspec_check_command.dart` on the `packages` repository? Also, what launch config to use to be able to debug this script?

Also, wouldn't this lint be useful as part of the `https://pub.dev/packages/flutter_lints` or `https://pub.dev/packages/lints` ?

stuartmorgan: > @stuartmorgan what is the command to run updated `pubspec_check_command.dart` on the `packages` repository? Also, what launch config to use to be able to debug this script?

The tool is documented [here](https://github.com/flutter/packages/blob/main/script/tool/README.md), but I'm not sure what you mean by launch config.

> Also, wouldn't this lint be useful as part of the `https://pub.dev/packages/flutter_lints` or `https://pub.dev/packages/lints` ?

It's not an actual lint, it's custom repository tooling, so it can't be added to packages that are just lists of lints to enable.

ekuleshov: > > @stuartmorgan what is the command to run updated `pubspec_check_command.dart` on the `packages` repository? Also, what launch config to use to be able to debug this script?
> 
> The tool is documented [here](https://github.com/flutter/packages/blob/main/script/tool/README.md), but I'm not sure what you mean by launch config.

Okay, it looks like the CI runs it with these params:

`dart run script/tool/bin/flutter_plugin_tools.dart pubspec-check --allow-dependencies=script/configs/allowed_unpinned_deps.yaml --allow-pinned-dependencies=script/configs/allowed_pinned_deps.yaml --min-min-flutter-version=3.13.0`

Adding `test` and `flutter_test` to the check brings up these packages:
```
  packages/css_colors
  packages/web_benchmarks
```

> > Also, wouldn't this lint be useful as part of the `https://pub.dev/packages/flutter_lints` or `https://pub.dev/packages/lints` ?
> 
> It's not an actual lint, it's custom repository tooling, so it can't be added to packages that are just lists of lints to enable.

Understand. Though the `dart analyze` is also being run by the CI. E.g. the `sort_pub_dependencies` lint is enforced on pubspecs.

I'm just saying maybe such test dependency check could/should be moved into lints. I've seen a few published packages messing test dependencies and Dart pub tools don't allow to specify dependency excludes down stream.

stuartmorgan: > Adding `test` and `flutter_test` to the check brings up these packages:
> 
> ```
>   packages/css_colors
>   packages/web_benchmarks
> ```

Those both look like mistakes that could be fixed in a PR adding the check.

> I'm just saying maybe such test dependency check could/should be moved into lints.

You could certainly propose new lints in the Dart linter repo. Many things in flutter/packages repo tooling are things that in theory I would like to explore moving to various upstream places (Dart lints, pub, etc.), but don't have the time currently.

ekuleshov: > > Adding `test` and `flutter_test` to the check brings up these packages:
> > ```
> >   packages/css_colors
> >   packages/web_benchmarks
> > ```

Okay. Let's try this https://github.com/flutter/packages/pull/6472

PR #6447, opened 2024-04-01
[go_router] Fixes deep links with no path

kforjan:
This PR addresses an issue where root deep links did not correctly navigate to '/'. The problem originated from the assumption that only paths (like '/', '/details', etc.) would be passed to the `canonicalUri` method in `path_utils.dart`. However, in the case of deep links, the full URL (including the scheme and domain) would be passed, causing the trailing slash to be incorrectly removed for root URLs.

The first part of the fix modifies the `canonicalUri` method to check the actual path using `uri.path`. This ensures that the trailing slash is preserved for root URLs, even when the full URL is passed to the method. Importantly, even if '/' or '/details' is passed as a URI to `canonicalUri`, `uri.path` will still return '/' or '/details', ensuring consistent behavior.

The second part of the fix modifies the `parseRouteInformationWithDependencies` function in `parser.dart` to correctly handle URLs with query parameters or fragments. Previously, the trailing slash was added after the query parameters or fragments, which is incorrect. The fix ensures that the trailing slash is added immediately after the base URL, before any query parameters or fragments.

Preserving the trailing slash for root URLs is crucial because the `_matchByNavigatorKey` method in `match.dart` uses `uri.path` as the `remainingLocation` parameter. (why is that method relevant? because `parseRouteInformationWithDependencies` uses `findMatch`, which calls `_getLocRouteMatches`, which calls `match` and it calls `_matchByNavigatorKey`). If the trailing slash is removed from the root deep link URL, `uri.path` will not return '/', and the URL will not match any routes in the Go router.

To validate these changes, new tests have been added. In `path_utils_test.dart`, tests have been added to verify that the trailing slash is not removed from a URL that is not just the path. In `parser_test.dart`, a new test has been added to verify that a URI with an empty path is correctly parsed. This test covers the case where `routeInformation.uri.hasEmptyPath` is true, which was not previously tested.

Issues fixed:
 - flutter/flutter#133928

## Pre-launch Checklist

- [x] I signed the [CLA].


kforjan:
@hangyujin @johnpryan I have noticed that the code owner who is added as a reviewer hasn't been active for some time. Could you take a look at this, or do you know who could provide some feedback? Thank you!

hangyujin:
@kforjan I will take a look

packages/go_router/lib/src/path_utils.dart
```diff
@@ -125,20 +125,20 @@ String canonicalUri(String loc) {
     throw GoException('Location cannot be empty.');
   }
   String canon = Uri.parse(loc).toString();
+  final Uri uri = Uri.parse(canon);
```
hangyujin, line 128: line128 should be after line 129

I can think of a corner test case  ```/profile/?```

packages/go_router/lib/src/path_utils.dart
```diff
@@ -125,20 +125,20 @@ String canonicalUri(String loc) {
...
   canon = canon.endsWith('?') ? canon.substring(0, canon.length - 1) : canon;
 
   // remove trailing slash except for when you shouldn't, e.g.
   // /profile/ => /profile
   // / => /
   // /login?from=/ => login?from=/
-  canon = canon.endsWith('/') && canon != '/' && !canon.contains('?')
+  canon = uri.path.endsWith('/') && uri.path != '/' && !uri.hasQuery
```
hangyujin, line 135: add `&& !uri.hasFragment`

packages/go_router/lib/src/parser.dart
```diff
@@ -79,11 +79,19 @@ class GoRouteInformationParser extends RouteInformationParser<RouteMatchList> {
...
+      if (routeInformation.uri.hasQuery) {
+        newUri += '?${routeInformation.uri.query}';
+      }
+      if (routeInformation.uri.hasFragment) {
+        newUri += '#${routeInformation.uri.fragment}';
+      }
+      initialMatches = configuration.findMatch(newUri, extra: state.extra);
+    } else {
+      initialMatches = configuration.findMatch(routeInformation.uri.toString(),
+          extra: state.extra);
```
hangyujin, line 93: nit: add comma

hangyujin (approved):
Thanks for your contribution. This looks great to me, the fix looks clean and also thank you for your detailed description and tests.

chunhtai (approved):
LGTM

DennisKragekjaer:
image

DennisKragekjaer:
So I still see an issue, also after update to new version 13.2.3

chunhtai:
Since this is merged, please file a new issue with repro.

kforjan:
Hello @DennisKragekjaer! I have just tried using the exact same url with a same custom scheme setup like the one in your example. It worked for me and I couldn't reproduce your problem. Please note that `/link` is the path which you're trying to open in your example, so you have to include a `GoRoute` with the path `/link` in the `routes` parameter which you provide to your `GoRouter` instance. If this doesn't help - please open an issue with reproduction steps.

Note for @chunhtai: while I investigated this, I noticed an edge case that I missed in my original implementation. When using custom schemes `uri.origin` cannot be used. After parsing a custom scheme with `Uri.parse() `calling `uri.origin` will produce the following Exception:
```
Bad state: Origin is only applicable to schemes http and https: customScheme://example
```
It doesn't break the flow or affect the app launch in any way, rather it just produces this error message in the debug console. It happens only with custom schemes with no path, for example:
`customScheme://example` will have an exception in the console while `customScheme://example/` won't have it. 
It isn't breaking for those cases, but it should be fixed anyway. One way to do it is to use `uri.schema` and `uri.host` to construct the uri, since both are available for custom schemas and http/https, but I am not sure if there are any other cases which I am missing where schema and host wont be in the same format as schema://host

DennisKragekjaer:
@kforjan Thanks for looking into this. The thing is I have no idea where this "dk.kragekjaer.mille...." comes from.

PR #6453, opened 2024-04-03
[in_app_purchase] Convert Android data objects to Pigeon

stuartmorgan:
Follow-up to https://github.com/flutter/packages/pull/6262, converting the nested data objects from the legacy JSON conversion to typed Pigeon objects.

Deprecates all the JSON code in the wrapper objects, but does not remove it because, unfortunately, the wrapper objects were both platform data transfer objects *and* the public API surface, so the JSON serialization/deserialization code that was part of those objects for platform channel usage is part of the public API as well. At some point we can do a breaking change to clean up that code (https://github.com/flutter/flutter/issues/146202).

Fixes https://github.com/flutter/flutter/issues/117910

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Messages.java
+++ b/packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Messages.java
@@ -116,6 +116,19 @@ public class Messages {
     }
   }
 
+  /** Pigeon version of Java ProductDetails.RecurrenceMode. */
+  public enum PlatformRecurrenceMode {
+    FINITE_RECURRING(0),
+    INFINITE_RECURRING(1),
+    NON_RECURRING(2);
+
+    final int index;
+
+    private PlatformRecurrenceMode(final int index) {
+      this.index = index;
+    }
+  }
+
   /**
    * Pigeon version of Java QueryProductDetailsParams.Product.
    *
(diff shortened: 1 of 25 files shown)


packages/in_app_purchase/in_app_purchase_android/lib/src/pigeon_converters.dart
```diff
@@ -110,16 +156,104 @@ PlatformProductType platformProductTypeFromWrapper(ProductType type) {
   };
 }
 
+/// Creates a [PricingPhaseWrapper] from its Pigeon equivalent.
```
stuartmorgan, line 159: In most Pigeon conversions this new code would be offset by removing a bunch of manual JSON conversion on the Dart side (like it is on the Java side in this PR); in this case it's net new code for now because we can't remove the generated JSON conversion yet (as noted in the PR description).

packages/in_app_purchase/in_app_purchase_android/pigeons/messages.dart
```diff
@@ -12,14 +12,25 @@ import 'package:pigeon/pigeon.dart';
   copyrightHeader: 'pigeons/copyright.txt',
 ))
 
-/// Pigeon version of Java Product.
-class PlatformProduct {
-  PlatformProduct({required this.productId, required this.productType});
+/// Pigeon version of Java QueryProductDetailsParams.Product.
+class PlatformQueryProduct {
```
stuartmorgan, line 16: This rename is because I discovered that there are at least two different things called `Product` in the Billing Client API surface, so the old name was too generic.

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Messages.java
```diff
@@ -264,6 +360,324 @@ ArrayList<Object> toList() {
...
+  public static final class PlatformOneTimePurchaseOfferDetails {
+    private @NonNull Long priceAmountMicros;
+
+    public @NonNull Long getPriceAmountMicros() {
+      return priceAmountMicros;
+    }
+
+    public void setPriceAmountMicros(@NonNull Long setterArg) {
+      if (setterArg == null) {
+        throw new IllegalStateException("Nonnull field \"priceAmountMicros\" is null.");
```
reidbaker, line 377: setPrinceAmountMicros I think is only called by in plugin java code so this would be a plugin author error. Is that why this is not a PlatfromException?

reidbaker (approved):

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Messages.java
stuartmorgan: Pigeon generates these errors for the native classes in Java so that if someone violates their own API constraints in native plugin code it'll create an error on the Java side, where it happened, rather than the Dart side (when it's read) further from the mistake.

PR #4251, opened 2023-06-19
[go_router] Add support for preloading branches of StatefulShellRoute

tolo:
Adds support for preloading branches in a `StatefulShellRoute`. This functionality was initially part of an early implementation of flutter/packages#2650, however it was decided to implement this in a separate PR. The current implementation is a rewrite of the original implementation to better fit the final version of `StatefulShellRoute` (and go_router in general).

This fixes issue flutter/flutter#127804.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[...]

PR #2650, opened 2022-09-27
[go_router] Nested stateful navigation with ShellRoute

tolo:
Added functionality for building route configuration with support for preserving state in nested navigators. This change introduces a new shell route class called `StatefulShellRoute`, that uses separate navigators for its child routes as well as preserving state in each navigation branch. This is convenient when for instance implementing a UI with a `BottomNavigationBar`, with a persistent navigation state for each tab (i.e. building a `Navigator` for each tab). 
An example showcasing a UI with BottomNavigationBar and StatefulShellRoute has also been added ([`stateful_shell_route.dart`](https://github.com/tolo/flutter_packages/blob/nested-persistent-navigation/packages/go_router/example/lib/stateful_shell_route.dart)). 

Other examples of using `StatefulShellRoute` are also available in these repositories: 
* [stateful_books](https://github.com/tolo/stateful_books) - A fork of the Books example of go_router.
* [stateful_navbar](https://github.com/tolo/stateful_navbar) - A clone of the Flutter Material 3 Navigation Bar example.

<br/>
Below is a short example of how a `StatefulShellRoute` can be setup:

```dart
StatefulShellRoute(
  /// Each separate stateful navigation tree (i.e. Navigator) is represented by
  /// a StatefulShellBranch, which defines the routes that will be placed on that
  /// Navigator. StatefulShellBranch also makes it possible to configure
  /// things like an (optional) Navigator key, the default location (i.e. the
  /// location the branch will be navigated to when loading it for the first time) etc.
  branches: <StatefulShellBranch>[
    StatefulShellBranch(navigatorKey: optionalNavigatorKey, routes: <RouteBase>[
      GoRoute(
        path: '/a',
        builder: (BuildContext context, GoRouterState state) =>
        const RootScreen(label: 'A'),
        routes: <RouteBase>[
          GoRoute(
            path: 'details',
            builder: (BuildContext context, GoRouterState state) =>
            const DetailsScreen(label: 'A'),
          ),
        ],
      ),
    ]),
    /// The default location of a branch will by default be the first of the
    /// configured routes. To configure a different route, provide the
    /// defaultLocation parameter.
    StatefulShellBranch(defaultLocation: '/b/detail', routes: <RouteBase>[
      GoRoute(
        path: '/b',
        builder: (BuildContext context, GoRouterState state) =>
        const RootScreen(label: 'B'),
        routes: <RouteBase>[
          GoRoute(
            path: 'details',
            builder: (BuildContext context, GoRouterState state) =>
[...]

PR #6467, opened 2024-04-05
[go_router] Add support for preloading branches of StatefulShellRoute (revised solution)

tolo:
Adds support for preloading branches in a `StatefulShellRoute`. This functionality was initially part of an early implementation of flutter/packages#2650, however it was decided to implement this in a separate PR. The current implementation is a rewrite of the original implementation to better fit the final version of `StatefulShellRoute` (and go_router in general). 

**NOTE**: this is a revised version of the initial solution (see flutter/packages#4251), containing a substantially simpler implementation made possible thanks to recent refactoring in go_router.

This fixes issue flutter/flutter#127804. 

## Pre-launch Checklist

- [x] I signed the [CLA].


chunhtai:
I know I have been slacking off at this pr. I am sorry about this. My schedule is quite tight after I came back from vacation. This is still under my radar. I will try find time review this pr within 2 weeks

tolo:
> I know I have been slacking off at this pr. I am sorry about this. My schedule is quite tight after I came back from vacation. This is still under my radar. I will try find time review this pr within 2 weeks

No worries @chunhtai, I completely understand, have had far too much on my plate myself the last year 😅

packages/go_router/lib/src/route.dart
```diff
@@ -1242,8 +1281,40 @@ class StatefulNavigationShellState extends State<StatefulNavigationShell>
...
+        ShellRouteMatch? match;
+        matchList.visitRouteMatches((RouteMatchBase e) {
+          if (e is ShellRouteMatch && e.route == widget.route) {
+            match = e;
+            return false;
+          }
+          return true;
+        });
+
+        final Widget navigator = widget.shellRouteContext.navigatorBuilder(
```
cedvdb, line 1308: might want to null check match here

packages/go_router/lib/src/route.dart
```diff
@@ -57,7 +57,13 @@ typedef StatefulShellRoutePageBuilder = Page<dynamic> Function(
 
 /// Signature for functions used to build Navigators
 typedef NavigatorBuilder = Widget Function(
-    List<NavigatorObserver>? observers, String? restorationScopeId);
+    GlobalKey<NavigatorState> navigatorKey,
+    ShellRouteMatch match,
+    RouteMatchList matchList,
+    List<NavigatorObserver>? observers,
+    String? restorationScopeId);
+// typedef NavigatorBuilder = Widget Function(
```
cedvdb, line 65: seems like this line is unnecessary

packages/go_router/example/lib/stateful_shell_route.dart
```diff
@@ -60,6 +60,9 @@ class NestedTabNavigationExampleApp extends StatelessWidget {
                 ],
               ),
             ],
+            // To enable preloading of the initial locations of branches, pass
+            // 'true' for the parameter preload.
+            // preload: true,
```
cedvdb, line 65: I'd find it clearer if the last line was uncommented as `preload: false`

packages/go_router/lib/src/route.dart
```diff
@@ -1024,6 +1042,22 @@ class StatefulShellBranch {
...
+  ///
+  /// If this property is true, the branch will be loaded immediately when the
+  /// associated [StatefulShellRoute] is visited for the first time. In that
+  /// case, the branch will be preloaded by navigating to the initial location
+  /// (see [initialLocation]).
+  ///
+  /// *Note:* The primary purpose of branch preloading is to enhance the user
+  /// experience when switching branches, which might for instance involve
+  /// preparing the UI for animated transitions etc. Care must be taken to
+  /// **keep the preloading to an absolute minimum** to avoid any unnecessary
```
cedvdb, line 1057: The position here is overly dramatic imo. I'd change 

```
 Care must be taken to **keep the preloading to an absolute minimum** to avoid any unnecessary resource use.
```

into 

```
 Preloading involves resource usage for elements that are not rendered, as such this should be used sparingly.
```

or something like that.

packages/go_router/lib/src/route.dart
tolo: Good catch, will update this.

packages/go_router/example/lib/stateful_shell_route.dart
tolo: You may be right, I'll consider this.

packages/go_router/lib/src/route.dart
tolo: Great feedback. This wording was actually due to some rather lengthy discussion in the original PR around the potential negative aspects of preloading, as I recall it. There are always dangers of premature/unnecessary optimisation when exposing these kinds of APIs, so some level of cautionary language is certainly justified. But maybe it can be dialed back a bit.

unacorbatanegra:
Any updates on this?

packages/go_router/lib/src/route.dart
```diff
@@ -996,6 +1011,7 @@ class StatefulShellBranch {
     this.initialLocation,
     this.restorationScopeId,
     this.observers,
+    this.preload = false,
```
chunhtai, line 1014: should we check for initiallocation is set if preload is true?

packages/go_router/lib/src/route.dart
```diff
@@ -1032,6 +1048,21 @@ class StatefulShellBranch {
   /// The observers parameter is used by the [Navigator] built for this branch.
   final List<NavigatorObserver>? observers;
 
+  /// Whether this route branch should be loaded only when navigating to it for
```
chunhtai, line 1051: Can you inverse the statement? from reading it, it seem to not preload when this property is true

packages/go_router/lib/src/route.dart
```diff
@@ -1255,8 +1291,39 @@ class StatefulNavigationShellState extends State<StatefulNavigationShell>
     final bool locationChanged =
         previousBranchLocation != currentBranchLocation;
     if (locationChanged || !hasExistingNavigator) {
-      _branchNavigators[branch.navigatorKey] = shellRouteContext
-          .navigatorBuilder(branch.observers, branch.restorationScopeId);
+      _branchNavigators[branch.navigatorKey] =
```
chunhtai, line 1294: just realized somewhere in this method should probably remove entries in _branchNavigators the are no longer in the StatefulShellRoute?

chunhtai (commented):
Overall LGTM, just some comments

packages/go_router/lib/src/route.dart
tolo: Don't think that should be necessary, as it will fall back on the `defaultRoute` if `initialLocation` isn't provided.

packages/go_router/lib/src/route.dart
tolo: Right, forgot to change that. How about this: "Whether this route branch should be eagerly loaded when navigating to it for the first time..."?

packages/go_router/lib/src/route.dart
tolo: You're right, that is definitely necessary to do now. Will have a look at it.

packages/go_router/lib/src/route.dart
cedvdb: No, that's also confusing, it's not preloading when accessing the branch for the first time but when accessing the shell.

"Whether this "StatefulShellBranch" should be preloaded once the "StatefulShellRoute" has been accessed."

tolo: Ah, but of course, how about this: 
"Whether this route branch should be eagerly loaded when navigating to the associated StatefulShellRoute for the first time."

cedvdb: Yes that's much better

packages/go_router/lib/src/route.dart
tolo: @chuntai, I added a fix for removing old entries, but when attempting to write a test case, I ran into issues with duplicate GlobalKeys (_shellStateKey) when updating the dynamic RoutingConfig. Or - I should be clear - at first the test case worked, but not because of the code I added, but instead because the entire StatefulShellRoute was replaced/reloaded. But when I added the possibility to set the `GlobalKey<StatefulNavigationShellState> _shellStateKey` to maintain state, I started to get this error.  

When trying to troubleshoot it, I started to wonder whether dynamic RoutingConfig full supports ShellRoutes, at least with state. Added some test cases to test this, and it seems simple StatefulWidgets in a simple routing scenario works, but when adding a (regular) ShellRoute with a stateful shell I also get and issue with duplicate GlobalKeys. 

Not sure if I just missed something, but I anyway added a gist with the test cases here: https://gist.github.com/tolo/922fd838cdea30b17121533b0012eda4

cedvdb:
What's the status here ? It seems like you said that you have a test locally that's failing but it's unrelated to the changes here ?

tolo:
> What's the status here ? It seems like you said that you have a test locally that's failing but it's unrelated to the changes here ?

Yes, @chunhtai, have you had a chance to too look at my comment above (https://github.com/flutter/packages/pull/6467#discussion_r1644576099)?

The issue is however only with testing if obsolete branch navigators are removed (which is done dynamic RoutingConfig), so possibly this could be moved to a separate issue/PR. And another issue should probably be opened for fixing the support for dynamic RoutingConfig for (stateful) shell routes. What do you think @chunhtai?

chunhtai:
will take a look this week

packages/go_router/example/lib/others/custom_stateful_shell_route.dart
```diff
@@ -80,71 +86,69 @@ class NestedTabNavigationExampleApp extends StatelessWidget {
...
+            preload: true,
             // StatefulShellBranch will automatically use the first descendant
             // GoRoute as the initial location of the branch. If another route
             // is desired, specify the location of it using the defaultLocation
             // parameter.
-            // defaultLocation: '/c2',
+            // defaultLocation: '/b1',
             routes: <RouteBase>[
-              StatefulShellRoute(
+              StatefulShellRoute.indexedStack(
```
chunhtai, line 99: why changing this example? In stead of adding more complexity to this example, we should probably create a new example that target specific feature

packages/go_router/lib/src/route.dart
chunhtai: yes sounds good to me, can you reflect that to the doc string?

packages/go_router/lib/src/route.dart
chunhtai: can you file an issue for it? looks like a bug

chunhtai: is the fix in this pr already?

Shunt22:
Any status update?
Would love to have this feature in my project.

packages/go_router/lib/src/route.dart
tolo: It is now. 😊 Note though that I disabled two tests that fail due to the issue I mentioned above.

tolo:
@chunhtai, as mentioned above, I've included the fix for cleaning obsolete branches. But I'm thinking it might be better to postpone that fix until the issues with duplicate GlobalKeys is resolved.

packages/go_router/lib/src/route.dart
```diff
@@ -1241,25 +1285,71 @@ class StatefulNavigationShellState extends State<StatefulNavigationShell>
...
+        ShellRouteMatch? match;
+        matchList.visitRouteMatches((RouteMatchBase e) {
+          match = e is ShellRouteMatch && e.route == route ? e : match;
+          return match == null;
+        });
+        assert(match != null);
+
+        final Widget navigator = widget.shellRouteContext.navigatorBuilder(
+          branch.navigatorKey,
+          match!,
```
cedvdb, line 1330: `!` can be removed on this line

tolo: Oh? Maybe I'm missing something, but that gives me an error (since match is ShellRouteMatch?). Would have been nice if the assert would have that effect though.

cedvdb:
@chunhtai can this PR get some love ? It seems highly anticipated

tolo:
FYI, for those interested in `StatefulShellRoute` evolution, there is a related (draft) PR in the works: flutter/packages#7622. There are some ideas around quality-of-life improvements around StatefulShellRoute in there, so have a look if you're interested.

packages/go_router/lib/src/route.dart
```diff
@@ -1317,13 +1406,26 @@ class StatefulNavigationShellState extends State<StatefulNavigationShell>
             key: ObjectKey(branch),
             branch: branch,
             navigatorForBranch: (StatefulShellBranch b) =>
```
chunhtai, line 1408: while you on it, can you also rename the parameter `b` to be something more descriptive. Our style guide suggest no abbreviation

chunhtai (approved):
LGTM

hannah-hyj (approved):
LGTM!

Shunt22:
Hi!
Sorry, when will this be live?

cedvdb:
@chunhtai Can autosubmit be added ? as conflicts were already resolved 2 weeks ago

chunhtai:
Looks like there is some outdated excerpt
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8731887329080081345/+/u/Run_package_tests/README_snippet_validation/stdout?format=raw

I can't fix it through gihub ui.  @tolo can you try to fix it?

PR #6471, opened 2024-04-05
[packages] Set parallelizable to NO to reduce test flakiness in packages project tests

LouiseHsu:
fixes https://github.com/flutter/flutter/issues/146358

## Pre-launch Checklist

- [x] I signed the [CLA].
- [] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].

### DIFF
--- a/packages/google_sign_in/google_sign_in_ios/example/macos/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
+++ b/packages/google_sign_in/google_sign_in_ios/example/macos/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
@@ -39,7 +39,7 @@
       <Testables>
          <TestableReference
             skipped = "NO"
-            parallelizable = "YES">
+            parallelizable = "NO">
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "331C80D4294CF70F00263BE5"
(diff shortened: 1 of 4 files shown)


stuartmorgan:
Do we always want this to be true in every project? If so, we should probably make a repo tool change that enforces it, so we don't invariably forget for new projects in the future.

hellohuanlin:
interesting. if setting to NO works, i suspect the flakiness may be due to our OCMock usage where we somehow created some sort of side effect by mistake.

LouiseHsu:
> interesting. if setting to NO works, i suspect the flakiness may be due to our OCMock usage where we somehow created some sort of side effect by mistake.

this is probably true, but when it hangs its very difficult to debug, bc no stack trace or errors are generated

LouiseHsu:
> Do we always want this to be true in every project? If so, we should probably make a repo tool change that enforces it, so we don't invariably forget for new projects in the future.

I feel like we probably should, unless you think there would be a significant performance trade off if the tests aren't running parallel. But most of the packages that I looked at dont have this attribute set at all, these are the only ones.

jmagman:
> But most of the packages that I looked at dont have this attribute set at all

What's the Xcode default when you create a new project?

LouiseHsu:
> > But most of the packages that I looked at dont have this attribute set at all
> 
> What's the Xcode default when you create a new project?

parallelizable is set to YES by default when you create a new xcode project, but I think this is relatively new.

jmagman (approved):

stuartmorgan:
> interesting. if setting to NO works, i suspect the flakiness may be due to our OCMock usage where we somehow created some sort of side effect by mistake.

Global mocking in OCMock is a strong contender for the source of problems that only show up when running in parallel. That's one of several reasons I'm extremely against adding any new usage of it in the reviews I do.

stuartmorgan:
I'll file an issue for tooling enforcement of this; the PR can definitely land without the tooling part.

PR #6473, opened 2024-04-05
Add missing_code_block_language_in_doc_comment lint to flutter/packages.

kallentu:
Adds this Dartdoc-related lint to the flutter repository, in replacement of the Dartdoc warning (missingCodeBlockLanguage) because it will be deprecated and removed soon.

flutter/flutter already has this lint as well. Adding to flutter/engine with https://github.com/flutter/engine/pull/51944.

Lint Proposal: https://github.com/dart-lang/linter/issues/4904

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/rfw/lib/src/dart/binary.dart
+++ b/packages/rfw/lib/src/dart/binary.dart
@@ -220,7 +220,7 @@ Uint8List encodeLibraryBlob(RemoteWidgetLibrary value) {
 ///
 ///   For example, this switch:
 ///
-///   ```
+///   ```none
 ///   switch (args.a) {
 ///    0: 'z',
 ///    1: 'o',
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -112,6 +112,7 @@ linter:
     # - lines_longer_than_80_chars # not required by flutter style
     - literal_only_boolean_expressions
     # - matching_super_parameters # blocked on https://github.com/dart-lang/language/issues/2509
+    - missing_code_block_language_in_doc_comment
     - missing_whitespace_between_adjacent_strings
     - no_adjacent_strings_in_list
     - no_default_cases
--- a/packages/flutter_migrate/lib/src/base/common.dart
+++ b/packages/flutter_migrate/lib/src/base/common.dart
@@ -63,7 +63,7 @@ String getEnumName(dynamic enumItem) {
 /// Rationale:
 ///
 /// Consider the following snippet:
-/// ```
+/// ```dart
 /// try {
 ///   await foo();
 ///   ...
@@ -87,7 +87,7 @@ String getEnumName(dynamic enumItem) {
 /// [asyncGuard] is intended to wrap awaited expressions occurring in a `try`
 /// block. The behavior described above gives the behavior that users
 /// intuitively expect from `await`. Consider the snippet:
-/// ```
+/// ```dart
 /// try {
 ///   await asyncGuard(() async {
 ///     var c = Completer();
--- a/packages/flutter_migrate/lib/src/base/logger.dart
+++ b/packages/flutter_migrate/lib/src/base/logger.dart
@@ -1125,7 +1125,7 @@ class SpinnerStatus extends AnonymousSpinnerStatus {
 /// ```
 ///
 /// yields:
-/// ```
+/// ```none
 ///   Usage: app main_command <subcommand>
 ///          [arguments]
 /// ```
--- a/packages/go_router/lib/src/route.dart
+++ b/packages/go_router/lib/src/route.dart
@@ -76,7 +76,7 @@ typedef ExitCallback = FutureOr<bool> Function(BuildContext context);
 /// with the sub routes.
 ///
 /// For example these routes:
-/// ```
+/// ```none
 /// /         => HomePage()
 ///   family/f1 => FamilyPage('f1')
 ///     person/p2 => PersonPage('f1', 'p2') ← showing this page, Back pops ↑
@@ -84,7 +84,7 @@ typedef ExitCallback = FutureOr<bool> Function(BuildContext context);
 ///
 /// Can be represented as:
 ///
-/// ```
+/// ```dart
 /// final GoRouter _router = GoRouter(
 ///   routes: <GoRoute>[
 ///     GoRoute(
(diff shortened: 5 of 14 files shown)


stuartmorgan:
Is this lint in 3.4, or do we need to wait another stable release cycle?

Piinks:
(PR triage): What is the status of this PR?

kallentu:
> (PR triage): What is the status of this PR?

@Piinks 
Linux analyze legacy N-1 is failing because it's linting on the wrong things. There's a commit that just so happens to fall in 3.5 (https://github.com/dart-lang/sdk/commit/b6ac281270ac751175c5298ce78dd9f2cb3bd594) which isn't picked up in that tryjob. The rest of the lint work is done in 3.4 and the version of legacy N-1 is at Dart 3.4.x.

What's the best way forward? I don't want to be doing the fix for what's failing because it's wrong behaviour. I'm happy to wait longer, but I suspect N-1 will become N-2 and that will end up failing.

Piinks:
Hmm, I am not sure. @stuartmorgan would likely be best to advise.
I think you would have to bump the dependency on each affected package to make it no longer compatible with N-1 and N-2.

stuartmorgan:
The options are:
- Drop support for N-1 in those packages
- Wait for the next stable to do this, and drop N-2 in those packages
- Wait for two more stables (I would not recommend this one; the value of N-2 support is low)
- Add `// ignore` annotations for the N-1 false positives, with a comment explaining why and linking to an issue that tracks removing them after the next stable update. There's no warning for unnecessary `// ignore`s, so that will work on all versions.

Any of those is fine. Our policy is that we only commit to supporting N, so the bar for dropping N-1 is low; N-1/N-2 tests are to prevent *accidentally* doing something that doesn't compile in N-1/N-2 and shipping it in a version that claims to support that version, thus breaking clients.

kallentu:
> The options are:
> 
> * Drop support for N-1 in those packages
> * Wait for the next stable to do this, and drop N-2 in those packages
> * Wait for two more stables (I would not recommend this one; the value of N-2 support is low)
> * Add `// ignore` annotations for the N-1 false positives, with a comment explaining why and linking to an issue that tracks removing them after the next stable update. There's no warning for unnecessary `// ignore`s, so that will work on all versions.
> 
> Any of those is fine. Our policy is that we only commit to supporting N, so the bar for dropping N-1 is low; N-1/N-2 tests are to prevent _accidentally_ doing something that doesn't compile in N-1/N-2 and shipping it in a version that claims to support that version, thus breaking clients.

Added ignores as you suggested with the issue tracking it: https://github.com/flutter/flutter/issues/157620.

Fixed up pubspec and changelogs. Should be ready for review and to move forward.

yjbanov (approved):

packages/flutter_migrate/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 0.0.1+4
 
+* Added `missing_code_block_language_in_doc_comment` lint.
```
stuartmorgan, line 3: Nit: Adds

(Per the CHANGELOG style guide in the contributing docs.)

stuartmorgan: Same for the other CHANGELOGs as well.

packages/process/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 5.0.3
```
stuartmorgan, line 1: The changelog and pubspec changes for this package can be reverted, since now you're only adding a non-doc comment to an implementation file, which doesn't affect clients.

stuartmorgan (approved):
LGTM with minor changes.

packages/flutter_migrate/CHANGELOG.md
kallentu: Done, thanks for the reminder!

cbracken (approved):

chunhtai (approved):
LGTM

ditman (approved):
LGTM, but there's some TODOs that could be done now I think?

packages/rfw/lib/src/flutter/material_widgets.dart
```diff
@@ -64,10 +64,16 @@ import 'runtime.dart';
 ///    be supported. As a result, the following `ButtonBar` parameters are no longer
 ///    supported:
 ///
+// TODO(kallentu): Remove ignore and fix when stable is bumped.
+// https://github.com/flutter/flutter/issues/157620
+// ignore: missing_code_block_language_in_doc_comment
 ///    * `buttonMinWidth`
 ///    * `buttonHeight`
 ///    * `buttonAlignedDropdown`
```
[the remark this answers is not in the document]
kallentu: Yeah, this is known. It'll get fixed next release. https://github.com/flutter/flutter/issues/157620 explains a bit, but the commit that fixes it isn't in the right version so the n-1 tryjob fails. Solving this by putting a lint ignore for now. Will follow up later to clean it up :)

packages/webview_flutter/webview_flutter_web/lib/src/http_request_factory.dart
```diff
@@ -39,6 +39,9 @@ class HttpRequestFactory {
   ///
   /// The following is equivalent to the [getString] sample above:
   ///
+  // TODO(kallentu): Remove ignore and fix when stable is bumped.
+  // https://github.com/flutter/flutter/issues/157620
+  // ignore: missing_code_block_language_in_doc_comment
```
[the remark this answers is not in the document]
kallentu: Yeah fair enough. I can do that. I think I was hoping to just preserve the comments as are for now. But no harm, I'll fix these indented comments --> we prefer those anyways,

packages/rfw/lib/src/dart/binary.dart
```diff
@@ -159,6 +171,9 @@ Uint8List encodeLibraryBlob(RemoteWidgetLibrary value) {
 ///   one of the subparts of the imported library name. For example, `import
 ///   a.b` is encoded as:
 ///
+// TODO(kallentu): Remove ignore and fix when stable is bumped.
+// https://github.com/flutter/flutter/issues/157620
+// ignore: missing_code_block_language_in_doc_comment
 ///       02 00 00 00 00 00 00 00  01 00 00 00 00 00 00 00
```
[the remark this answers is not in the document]
kallentu: Sure. I can make that change.

ditman:
Thanks @kallentu!!

PR #6493, opened 2024-04-09
[camerax] Shorten interval for releasing weak references to Dart-wrapped native objects

camsim99:
This PR:

1. Shortens the `InstanceManager`'s default time interval that it waits to remove references to Dart-wrapped Android native objects. The shortened interval matches `webview_flutter_android`s `InstanceManager`; because it has been tested by that plugin, we expect it not to impact performance and this could help reduce overall memory usage by the plugin.
2. Dynamically shortens that same time interval whenever image streaming is started/stopped to account for the increased memory usage that this camera use case requires.

Fixes https://github.com/flutter/flutter/issues/145893.

## Pre-launch Checklist

- [x] I signed the [CLA].


gmackall (commented):
Is there a reason we can't set to `1000` all the time? The fact that we are switching seems to imply some sort of trade off to me (otherwise we could just stay at `1000` right?) but it isn't clear to me what that trade off is. 

Is it just the potential of performance impact? But we are fine with that potential performance impact when we need the shorter interval because image streaming demands it?

camsim99:
> Is there a reason we can't set to `1000` all the time? The fact that we are switching seems to imply some sort of trade off to me (otherwise we could just stay at `1000` right?) but it isn't clear to me what that trade off is.
> 
> Is it just the potential of performance impact? But we are fine with that potential performance impact when we need the shorter interval because image streaming demands it?

Yes, you hit the nail on the head--we just aren't sure that it wouldn't impact performance, but we know that to avoid a crash, we have to lower it for image streaming. We could lower the default below 3000, but it doesn't seem necessary for any other camera use cases so far.

gmackall (approved):

PR #6494, opened 2024-04-09
[many] Remove references to v1 embedding

gmackall:
I have a WIP effort to remove the v1 embedding entirely (https://github.com/flutter/flutter/pull/146523 is the latest pr).

`registerWith` references the v1 embedding, which has been deprecated for many years, so this PR removes it from all plugins.

Also removes some additional references, see in particular these three commits:

1. Modifies private `ActivityState` class to remove `PluginRegistry.Registrar` member in `image_picker`: https://github.com/flutter/packages/pull/6494/commits/c2e4c872917a32c61d77a72b3f2e8fe25c3d6599
2. Replaces `FlutterMain.getLookupKeyForAsset()` with `FlutterLoader.getLookupKeyForAsset()` in `google_maps` plugin: https://github.com/flutter/packages/pull/6494/commits/73c3de3807dc20333536affb86c87b82ac267bf9
3. Removes deprecated `RegistrarFlutterAssetManager` class in `webview_flutter`: https://github.com/flutter/packages/pull/6494/commits/cc842c69b2ed2a5e5c203d7c9514605823225b07

Fixes https://github.com/flutter/flutter/issues/70923

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
It's been a while since I was in conversations around the v1 wind-down, so I don't have all the context any more. What's the current state of v1 support on `stable`? IIRC the last I had been involved it was being moved behind a flag, but it was still possible to use that flag to force build an app in v1 mode. Is that still the case?

The way I would want this to go is:
- Make v1 fully impossible to use (may already be done).
- Once that's on stable, we can delete the v1 support from all of our plugins (updating the min supported Flutter version if necessary).
- We can also do outreach to top plugins at that point to let them know they can remove it as well.

I think I'd like to keep `registerWith` in the interface, with a prominent comment saying it will never be called and is only there to avoid breaking compilation of plugin, for a while; it should be pretty low-impact for us to keep just that declaration. Eventually we could remove it and break plugins on master, but it would be good to have that be a while after all the big plugins (including ours) have had a chance to remove it.

gmackall:
> What's the current state of v1 support on stable?

Currently still possible to build a v1 app by passing a flag (will error without the flag), though that will change in the next release (once https://github.com/flutter/flutter/pull/144726 gets picked up).

> I think I'd like to keep registerWith in the interface

To be a little nitpicky, `registerWith` itself isn't [in the interface](https://github.com/flutter/engine/blob/main/shell/platform/android/io/flutter/embedding/engine/plugins/FlutterPlugin.java), it is just an additional static method we have recommended people include. But (if people follow our recommendation) that static method references the inner type [io.flutter.plugin.common.PluginRegistry.Registrar](https://github.com/flutter/engine/blob/main/shell/platform/android/io/flutter/plugin/common/PluginRegistry.java#L80), which is part of the v1 embedding. 

But I'm fine leaving that interface around (with a comment warning it is slated for deletion) for a release to give plugin authors time to migrate, while deleting the rest of the v1 embedding now. Does that sound reasonable to you? Also I'm happy to help reach out to the top plugin authors, if we have a list somewhere.

gmackall:
Also in light of 
> Once that's on stable, we can delete the v1 support from all of our plugins (updating the min supported Flutter version if necessary).

I'll get this pr to tests passing, and then leave it around until the next release before requesting review.

stuartmorgan:
> For full transparency we have been attempting to approach it faster, as a flat out deletion rather than wind down

Right, I was thinking of the entire process of "fully supported" to "completely gone" when I say wind-down. I've been involved in discussions on and off since very early in the process.

> To be a little nitpicky, `registerWith` itself isn't [in the interface](https://github.com/flutter/engine/blob/main/shell/platform/android/io/flutter/embedding/engine/plugins/FlutterPlugin.java), it is just an additional static method we have recommended people include. But (if people follow our recommendation) that static method references the inner type [io.flutter.plugin.common.PluginRegistry.Registrar](https://github.com/flutter/engine/blob/main/shell/platform/android/io/flutter/plugin/common/PluginRegistry.java#L80), which is part of the v1 embedding.

Ah, that's more unfortunate in terms of code volume. But since it's just an interface we could strip out all the comments and at least get it down to a pretty small amount.

> But I'm fine leaving that interface around (with a comment warning it is slated for deletion) for a release to give plugin authors time to migrate, while deleting the rest of the v1 embedding now. Does that sound reasonable to you?

Yes, I'm on board with that. The main thing I had pushed back on in the past was deleting the v1 support from our plugins while the tool still supported v1 builds, because it seemed backward. It's much smoother if we delete things from the tool up (just as we do with dropping old OS versions).

> Also I'm happy to help reach out to the top plugin authors, if we have a list somewhere.

I'll get an updated list and do the outreach once the change hits `stable`; it's not that long a list so I can do it pretty quickly.

johnmccutchan:
I'm not sure I understand why we aren't landing this PR now.

0. The code has been removed from flutter tool. 
1. There are _no_ tests in this repository (or any repository) that are testing the functionality of v1 embedder.
2. The Android v1 embedder has been broken since July 2023 (I intentionally didn't implement necessary functionality)

Why are we keeping dead and untested code around in our repository?

stuartmorgan:
> I'm not sure I understand why we aren't landing this PR now.
> 
> 0. The code has been removed from flutter tool.

According to [this comment](https://github.com/flutter/packages/pull/6494#issuecomment-2046010323) it's been removed on `master`, but not `stable`. We have an explicit policy in this repository of supporting `stable` for our packages.

> 1. There are _no_ tests in this repository (or any repository) that are testing the functionality of v1 embedder.

Not having tests for v1 here doesn't mean we can intentionally break v1 support. It's in the same category as ["best effort" OS support](https://docs.flutter.dev/reference/supported-platforms).

> 2. The Android v1 embedder has been broken since July 2023 (I intentionally didn't implement necessary functionality)

Can you elaborate on what "broken" means, concretely? What happens, right now, if someone attempts to build an app on `stable` with the v1 flag?

johnmccutchan:
> > I'm not sure I understand why we aren't landing this PR now.
> > 
> > 0. The code has been removed from flutter tool.
> 
> According to [this comment](https://github.com/flutter/packages/pull/6494#issuecomment-2046010323) it's been removed on `master`, but not `stable`. We have an explicit policy in this repository of supporting `stable` for our packages.
> 
> > 1. There are _no_ tests in this repository (or any repository) that are testing the functionality of v1 embedder.
> 
> Not having tests for v1 here doesn't mean we can intentionally break v1 support. It's in the same category as ["best effort" OS support](https://docs.flutter.dev/reference/supported-platforms).
> 

This is an API that was deprecated 5+ years ago and has no users.

> > 2. The Android v1 embedder has been broken since July 2023 (I intentionally didn't implement necessary functionality)
> 
> Can you elaborate on what "broken" means, concretely? What happens, right now, if someone attempts to build an app on `stable` with the v1 flag?

The app will build but fail at runtime.

stuartmorgan:
> This is an API that [...] and has no users.
> [...]
> The app will build but fail at runtime.

Both of these things are completely new information to me. Is there more information on this somewhere?

johnmccutchan:
> > This is an API that [...] and has no users.
> > [...]
> > The app will build but fail at runtime.
> 
> Both of these things are completely new information to me. Is there more information on this somewhere?

New functionality added to the engine to better support platform views, impeller support (GL + Vulkan) were left unimplemented in the v1 embedder. They will throw or silently fail.

I would like to see this PR landed though as there is no benefit to keeping it out and by landing it we are able to delete a lot of long dead code that is causing confusion among new developers to the Android embedder.

stuartmorgan:
> New functionality added to the engine to better support platform views, impeller support (GL + Vulkan) were left unimplemented in the v1 embedder. They will throw or silently fail.

I'm still not clear on the details here based on that description. Some specific functionality (e.g., around platform views) throws, or it's literally impossible to run the app because it will immediately throw and crash?

> I would like to see this PR landed though as there is no benefit to keeping it out

I would like to fully understand this statement. Until less than an hour ago, every conversation around this was based on the premise that the v1 flag in `flutter` actually worked on `stable` at the time it was removed. This topic is something that has been under discussion for over a week (and in some sense, for several years), so I don't think taking a little bit longer to clearly explain and document (ideally not just here, but in https://github.com/flutter/flutter/issues/143531 which has no discussion of this) the state is unreasonable.

> by landing it we are able to delete a lot of long dead code that is causing confusion among new developers to the Android embedder.

I don't agree that landing this PR will allow us to immediately delete a bunch of engine code, for reasons discussed above. We need to allow a window of time where we do outreach to top plugin developers and get them to land similar PRs and then release updates, otherwise that PR will be *extremely* disruptive for anyone trying to build real-world apps against `master`.

johnmccutchan:
Stuart,

1) All of these methods and types were deprecated six years ago. According to our deprecation [policy](https://github.com/flutter/flutter/wiki/Tree-hygiene#deprecations), we have long exceeded the four-stable releases / one year criteria of deprecated code.

```
Deprecations are removed in a consistent "first-in-first-out" fashion. The lifetime for a Flutter deprecation is 1 year after reaching the stable channel, or after 4 stable releases, whichever is longer. Deprecations are still subject to the policy described on the [breaking changes page](https://flutter.dev/docs/release/breaking-changes) of the website. Where possible, prepare the dart fix tools and write appropriate migrations guides.
```

Additionally,

2) According to our breaking change [policy](https://github.com/flutter/flutter/wiki/Tree-hygiene#1-determine-if-your-change-is-a-breaking-change) indicates that this is not a breaking change as no tests fail when this PR lands.

stuartmorgan:
> All of these methods and types were deprecated six years ago.

And we have continued to encourage plugin authors to include this code in their plugins for all six of those years. Unfortunately the steps that were supposed to continue the process of turn-down—including changing that guidance and doing ecosystem outreach earlier—were disrupted by staffing changes.

> According to our deprecation [policy](https://github.com/flutter/flutter/wiki/Tree-hygiene#deprecations), we have long exceeded the four-stable releases / one year criteria of deprecated code.

I'm aware of what our general policy is. This is a special case that does not look at all like a normal API deprecation, for reasons I have already explained.

> Additionally,
> 
> 2. According to our breaking change [policy](https://github.com/flutter/flutter/wiki/Tree-hygiene#1-determine-if-your-change-is-a-breaking-change) indicates that this is not a breaking change as no tests fail when this PR lands.

This PR runs in a repository that doesn't have a `customer_tests` shard, and even if it did it wouldn't do anything because the consumer model for packages is dependency via pub. If you would like me to add text to that section of the wiki explaining that it doesn't apply to flutter/packages and why, I'm happy to do so.

Nothing in your last comment is responsive to the concerns and questions I raised in my previous comment. Is the goal here to reach consensus, or to explain why you are going to ignore my objections to changing the previously established plan for turning down v1?

johnmccutchan:
My goal here is to reduce technical complexity of our Android embedder and get things under control so that we can onboard new engineers without their minds exploding. 

My second goal is to avoid engaging in debates about technical decisions that don't materially impact any of our users.

stuartmorgan:
> My second goal is to avoid engaging in debates about technical decisions that don't materially impact any of our users.

If you don't agree with the premise that it will impact users to suddenly remove all of the engine code—and thus suddenly render most real world applications unbuildable for Android on `master` by breaking many extremely widely used plugins—with no warning, leaving the entire plugin ecosystem to scramble to release new versions that will compile on `master`, then there's nothing further to discuss.

If the Android team's plan is to land this in a way that I believe will be very disruptive, over my express objections, @reidbaker should sign off on it.

reidbaker:
Let get in a room at talk about it @stuartmorgan @johnmccutchan

johnmccutchan:
> > My second goal is to avoid engaging in debates about technical decisions that don't materially impact any of our users.
> 
> If you don't agree with the premise that it will impact users to suddenly remove all of the engine code—and thus suddenly render most real world applications unbuildable for Android on `master` by breaking many extremely widely used plugins—with no warning, leaving the entire plugin ecosystem to scramble to release new versions that will compile on `master`, then there's nothing further to discuss.
> 

That does not represent my position at all. I don't appreciate this straw man argument. 

I've asked repeatedly for examples of applications that would be negatively impacted and I've yet to hear one.

I've audited the thirty most popular plugins and aside from the ones fixed in this PR, I've found two instances of this ancient code. PRs have been submitted for both of them.

Given that the plugin ecosystem has had six years to remove long deprecated code, classifying this as a "scramble" is ridiculous. Our deprecation policy is that we keep code around for 1 year, it has been six.

reidbaker:
> > > My second goal is to avoid engaging in debates about technical decisions that don't materially impact any of our users.
> > 
> > 
> > If you don't agree with the premise that it will impact users to suddenly remove all of the engine code—and thus suddenly render most real world applications unbuildable for Android on `master` by breaking many extremely widely used plugins—with no warning, leaving the entire plugin ecosystem to scramble to release new versions that will compile on `master`, then there's nothing further to discuss.
> 
> That does not represent my position at all. I don't appreciate this straw man argument.
> 
> I've asked repeatedly for examples of applications that would be negatively impacted and I've yet to hear one.
> 
> I've audited the thirty most popular plugins and aside from the ones fixed in this PR, I've found two instances of this ancient code. PRs have been submitted for both of them.
> 
> Given that the plugin ecosystem has had six years to remove long deprecated code, classifying this as a "scramble" is ridiculous. Our deprecation policy is that we keep code around for 1 year, it has been six.

The issue I got hung up on was that our deprecation told people to use code we are now removing. So that wasn't really a depreciation.

stuartmorgan:
> I've asked repeatedly for examples of applications that would be negatively impacted and I've yet to hear one.

It's unfortunate that you never asked that here, or in another discussion that I was part of; I've certainly answered this question for people in other discussions. Offhand:
- Any application using any of our plugins.
- Any application using any of the plugins you sent PRs for yesterday.
- Any application using any long-tail plugin that is still following our previous explicit advice that plugin developers ignore the deprecation annotation in order to continue to support v1.

I can't give you a specific list of application names. We could pull specific metrics if you want numbers, but I would guess we're probably talking about well over half of all applications targeting Android that would become unbuildable on `master` with the current versions of plugins. I wouldn't be surprised if it's more than 75% of non-toy applications.

> Given that the plugin ecosystem has had six years to remove long deprecated code

We explicitly told plugin developers to ignore the annotation in plugins to continue supporting v1.

> Our deprecation policy is that we keep code around for 1 year, it has been six.

If someone can point me to an announcement to plugin developers that went out more than a year ago changing our official guidance, I would have a different position on this.

The spirit of the policy you are quoting would be that we wait a full year after an announcement that AFAIK we still have not made. I don't think we should wait that long, but I think we should wait longer than ~immediately after telling people the guidance has changed.

gmackall:
+1 to discussing in a meeting

reidbaker:
Recap of offline discussion. 

This pr will land after the 3.22 release is in stable and when it has been updated to require the modified plugins use flutter 3.22 this will ensure that pub upgrade does not break users of our plugins. 

The tool pr to remove the flag that allows building with v1 embeddings has landed in 3.22 and there is breaking change documentation going out with the release. 
The engine pr to remove the v1 embedding will be held until we can do some outreach to migrate popular plugins or a couple of months whichever comes sooner.

gmackall:
Hmm, tests fail with 
```
Running command: "flutter test --color" in /b/s/w/ir/x/w/packages/packages/camera/camera_android
Resolving dependencies...
The current Flutter SDK version is 3.22.0-35.0.pre.11.

Because camera_android requires Flutter SDK version >=3.22.0, version solving failed.


You can try the following suggestion to make the pubspec resolve:
* Try using the Flutter SDK version: 3.22.0. 
```

Looks like the commit used by ci is defined here
https://github.com/flutter/packages/blob/main/.ci/flutter_master.version
need to figure out why that is mapping to a pre 3.22.0 version (cc @stuartmorgan if you know more 🙂 )

stuartmorgan:
I raised this in the releases channel; we can file a tracking issue if it's not resolved relatively soon.

The set of steps to get this PR unblocked is:
- flutter/flutter lands a change incrementing `master`'s version past `3.22-*` so that it is newer than 3.22 stable as defined by Dart version resolution.
- That change rolls into flutter/packages (generally happens once a day, but we can do a manual roll whenever).

gmackall:
> I raised this in the releases channel; we can file a tracking issue if it's not resolved relatively soon.
> 
> The set of steps to get this PR unblocked is:
> 
> * flutter/flutter lands a change incrementing `master`'s version past `3.22-*` so that it is newer than 3.22 stable as defined by Dart version resolution.
> * That change rolls into flutter/packages (generally happens once a day, but we can do a manual roll whenever).

Sounds good, thanks for the escalation help. Also FYI I've added some more v1 reference removals, based on things that broke in a dry run attempt of rolling the embedding deletion and this pr into g3

stuartmorgan (changes requested):
These should all get version bumps; we want to get the versions without this code out into client's hands ASAP so that it has time to filter organically through the ecosystem in advance of the engine code removal, in order to minimize disruption.

packages/webview_flutter/webview_flutter_android/android/src/main/java/io/flutter/plugins/webviewflutter/WebViewFlutterPlugin.java
```diff
@@ -50,33 +48,13 @@ public class WebViewFlutterPlugin implements FlutterPlugin, ActivityAware {
    *
    * <p>THIS PLUGIN CODE PATH DEPENDS ON A NEWER VERSION OF FLUTTER THAN THE ONE DEFINED IN THE
    * PUBSPEC.YAML. Text input will fail on some Android devices unless this is used with at least
-   * flutter/flutter@1d4d63ace1f801a022ea9ec737bf8c15395588b9. Use the V1 embedding with {@link
-   * #registerWith} to use this plugin with older Flutter versions.
+   * flutter/flutter@1d4d63ace1f801a022ea9ec737bf8c15395588b9.
```
stuartmorgan, line 51: This whole paragraph can go; the hash landed in 2019 and was in Flutter 1.16, so this whole comment has been wrong for quite a long time.

stuartmorgan (approved):
LGTM with one nit.

PR #6500, opened 2024-04-11
Go router builder v2.3.4 non fatal

ho-doan:

## Pre-launch Checklist

- [ ] I signed the [CLA].


chunhtai (commented):
Looks like you need to sign CLA, can you also talk about the reason for this change?

The original code forces required parameter to be present in path parameter, which makes sense to me.

ho-doan:
@chunhtai 
When publishing stg I get a warning like this `Non-fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: Null check operator used on a null value
at GoRouter.of(go_router.dart:147)
at GoRouterHelper.go(go_router.dart:39)
at $PageContainerRouteExtension.go(router.g.dart:88)`

`HomeRoute(
      url: state.pathParameters['url']!, <- in this line
);`

chunhtai:
Can you provide a repro?

ho-doan:
[sample code gen](https://github.com/ho-doan/base_s/blob/master/template/lib/core/router/router.g.dart)
Maybe the next repo will continue to have this warning.
This sample repo is similar to the prod repo.
-> Line 77

packages/go_router_builder/example/lib/stateful_shell_route_initial_location_example.g.dart
```diff
@@ -69,7 +69,7 @@ extension $NotificationsRouteDataExtension on NotificationsRouteData {
   static NotificationsRouteData _fromState(GoRouterState state) =>
       NotificationsRouteData(
         section: _$NotificationsPageSectionEnumMap
-            ._$fromName(state.pathParameters['section']!),
+            ._$fromName(state.pathParameters['section'] ?? ''),
```
chunhtai, line 72: path parameters probably shouldn't be nullable?

chunhtai (commented):
In your case, the ProductRouter has a non null string id, what would be stored in that id if it is not set?

packages/go_router_builder/example/lib/stateful_shell_route_initial_location_example.g.dart
ho-doan: it's definitely not null, but firebase got a `!` error so instead of using it, I replaced it with `??` to solve problems

chunhtai: what I meant is whether the code gen should only generate `??` when this is a query parameter.

However, I still not sure what would be store in the ProductRouter.id if query parmeter is null. I think we should force the ProductRouter.id to have a default value if it is not null and is used as query parmeter.

chunhtai: @ho-doan WDYT?

chunhtai:
Closing this pr for now, feel free to reopen when you have time

PR #6505, opened 2024-04-11
[url_launcher][web] Link should work when triggered by keyboard

mdebbar:
### Background

You can think of the `Link` widget (on the web) as two components working together:

1. The `<a>` element created by the `Link` widget. This is essential to make all browser interactions feel natural (e.g. context menu, cmd+click, etc).
2. The children of `Link` widget. These are the widgets visible to the user (e.g. a button or a hyperlink text) and the user can interact with them the same way they would interact with any Flutter widgets (focus, pointer click, etc).

In order for the Link widget to navigate to a URI, the two components from above have to indicate their intent of navigation:

1. Some widget has to call `followLink` to indicate that the click successfully landed (i.e. hit tested) on it. E.g. if it's a button, then the `onPressed` callback should lead to a call to the Link's `followLink`.
2. The `<a>` element also has to receive an event to initiate the navigation.

### The PR

We used to only handle click events on the `<a>` element, and no handling for keyboard events was present. So when a user tabs their way to the Link, then hits "Enter", the following happens:

1. The focused widget (e.g. button) that received the "Enter" will correctly indicate its intent to navigate by calling `followLink`.
2. The intent from the `<a>` element is lost because we were only handling clicks and not keyboard events.

This PR adds handling of keyboard events so that it works similar to clicks.

Fixes https://github.com/flutter/flutter/issues/97863


packages/url_launcher/url_launcher_web/example/integration_test/link_widget_test.dart
```diff
@@ -199,6 +391,27 @@ html.Element _findSingleAnchor() {
   return foundAnchors.single;
 }
 
+void _simulateClick(html.Element target) {
+  final html.MouseEvent clickEvent = html.MouseEvent('click');
+  clickEvent.initMouseEvent('click', /*bubbles*/ true);
```
ditman, line 396: `initMouseEvent` seems to be [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/initMouseEvent), can we set `bubbles: true` into the [options](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/MouseEvent#options) of the constructor of `MouseEvent`?

packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -192,6 +208,27 @@ class LinkViewController extends PlatformViewController {
     await SystemChannels.platform_views.invokeMethod<void>('create', args);
   }
 
+  void _onDomKeydown() {
```
ditman, line 211: Does this method need to check if the keydown is "enter", "space" or something like that? This'd trigger the link pressing any key, right?

packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -142,13 +146,25 @@ class LinkViewController extends PlatformViewController {
...
   static int? _hitTestedViewId;
 
-  static late StreamSubscription<html.MouseEvent> _clickSubscription;
+  static final JSFunction _jsGlobalKeydownListener = _onGlobalKeydown.toJS;
+  static final JSFunction _jsGlobalClickListener = _onGlobalClick.toJS;
+
+  static void _onGlobalKeydown(html.KeyboardEvent event) {
+    // The keydown event is not directly associated with the target Link, so
+    // we need to look for the recently hit tested Link to handle the event.
+    if (_hitTestedViewId != null) {
```
ditman, line 155: How does this `_hitTestedViewId` get set when using the keyboard?

packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -106,10 +109,11 @@ class LinkViewController extends PlatformViewController {
...
       // listener.
 
-      _clickSubscription =
-          const html.EventStreamProvider<html.MouseEvent>('click')
-              .forTarget(html.window)
-              .listen(_onGlobalClick);
+      // We listen to `keydown` events in the capture phase to ensure that we
+      // receive the event even if the engine calls `stopPropagation`.
+      html.window
+          .addEventListener('keydown', _jsGlobalKeydownListener, _useCapture);
```
ditman, line 115: Can this be attached to the `<a>` element that is backing the widget, instead of the global window?

ditman (approved):
LGTM! Thanks for adding this! Small question about attaching the keydown listener to the window, instead of the link element (or the view of the Widget)

packages/url_launcher/url_launcher_web/lib/src/link.dart
mdebbar: I don't want to assume what keys the framework and/or app are using as a trigger for the link.

As I explained in the PR description, there are 2 conditions that have to be met in order for the link to navigate. The code in `_onDomKeydown` is only one part. The other part is the app/framework calling `followLink` when they decide that the key should trigger the link. Here's roughly the sequence of events:

1. User tabs through the app until the focus reaches the button wrapped by the Link widget.
    * (Keep in mind that this tabbing is happening purely in the framework; the browser has no idea what's focused)
2. User hits "Enter".
3. The framework received the keydown event.
    1. It decides to trigger the `onPressed` on the button (because the button has focus).
    2. The button's `onPressed` callback leads to a call to the Link's `followLink`.
    3. `followLink` calls `LinkViewController.registerHitTest` which sets `_hitTestedViewId` to the Link's platform view ID.
4. The `LinkViewController` also receives keydown event.
    1. It checks if `_hitTestedViewId` is already set (indicating that the framework already decided to trigger the link).
    2. If `_hitTestedViewId` is set, it navigates to that Link's URI.

packages/url_launcher/url_launcher_web/lib/src/link.dart
mdebbar: I tried to explain the flow in [this comment](https://github.com/flutter/packages/pull/6505#discussion_r1562659976). Please let me know if it's still unclear.

packages/url_launcher/url_launcher_web/lib/src/link.dart
mdebbar: See [this comment](https://github.com/flutter/packages/pull/6505#discussion_r1562659976) for context.

Long story short, when the user is tabbing through the Flutter UI elements, only Flutter is tracking where focus is; the browser has no idea which element is focused (in many cases, it can't because the focused button may be drawn on a `<canvas>`). From a browser's perspective, most of the time, the focus is on `<flutter-view>` (with some exception like text fields). So key events are sent to `<flutter-view>`, not to the actual button or `<a>` element.

packages/url_launcher/url_launcher_web/lib/src/link.dart
ditman: (This should be part of the documentation of the widget :P)

packages/url_launcher/url_launcher_web/lib/src/link.dart
mdebbar: I added this explanation to the code comments.

packages/url_launcher/url_launcher_web/lib/src/link.dart
mdebbar: Added as comments to the `_onGlobalKeydown` method since that's where this information is relevant.

PR #6506, opened 2024-04-11
[Many] Bump to min minSdkVersion of 19 across all plugins

gmackall:
* Bumps any `minSdkVersion`'s below 19 to 19, as the change to only support 19+ has been in stable for a bit now.
* Bumps the minimum required flutter version to 3.16 for the changed packages, as that is the first version in which we started automigrating `minSdkVersion`'s less than 19.
* Also bumps the `minSdkVersion` in the legacy project to `flutter.minSdkVersion`, as this would happen when running an app with `minSdkVersion` < 19.
* Removes some code branches for cases of sdk < 19.
* Fixes some failing tests in `video_player` - for full transparency, I have now idea how these could have been passing before?

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/espresso/pubspec.yaml
+++ b/packages/espresso/pubspec.yaml
@@ -7,7 +7,7 @@ version: 0.3.0+8
 
 environment:
   sdk: ^3.1.0
-  flutter: ">=3.13.0"
+  flutter: ">=3.16.0"
 
 flutter:
   plugin:
--- a/packages/espresso/pubspec.yaml
+++ b/packages/espresso/pubspec.yaml
@@ -6,7 +6,7 @@ issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+
 version: 0.3.0+8
 
 environment:
-  sdk: ^3.1.0
+  sdk: ^3.2.0
   flutter: ">=3.16.0"
 
 flutter:
--- a/packages/espresso/pubspec.yaml
+++ b/packages/espresso/pubspec.yaml
@@ -3,7 +3,7 @@ description: Java classes for testing Flutter apps using Espresso.
   Allows driving Flutter widgets from a native Espresso test.
 repository: https://github.com/flutter/packages/tree/main/packages/espresso
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+espresso%22
-version: 0.3.0+7
+version: 0.3.0+8
 
 environment:
   sdk: ^3.1.0
--- a/.ci/legacy_project/all_packages/android/app/build.gradle
+++ b/.ci/legacy_project/all_packages/android/app/build.gradle
@@ -29,7 +29,7 @@ android {
 
     defaultConfig {
         applicationId "com.example.all_packages"
-        minSdkVersion 16
+        minSdkVersion flutter.minSdkVersion
         targetSdkVersion 30
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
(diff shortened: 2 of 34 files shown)


reidbaker (approved):

stuartmorgan:
> as the change to only support 19+ has been in stable for a bit now

What's the Flutter version that has that change? We need to make sure none of the packages being changed here have a min Flutter SDK version lower than that.

gmackall:
> What's the Flutter version that has that change? We need to make sure none of the packages being changed here have a min Flutter SDK version lower than that.

If I had to pick a specific PR, it would be https://github.com/flutter/flutter/pull/129729, though the change that really matters is the change to https://docs.flutter.dev/reference/supported-platforms, which isn't versioned along with the Flutter SDK really.

But https://github.com/flutter/flutter/pull/129729 is in 3.16 at the earliest, so it sounds like this will need to wait until the next stable, at which the minimum allowed will be N-2=3.16 (according to https://github.com/flutter/flutter/wiki/Updating-Packages-repo-for-a-stable-release)?

stuartmorgan:
> If I had to pick a specific PR, it would be [flutter/flutter#129729](https://github.com/flutter/flutter/pull/129729), though the change that really matters is the change to https://docs.flutter.dev/reference/supported-platforms, which isn't versioned along with the Flutter SDK really.

For the purposes of the packages repo it is the PR that matters, not the policy on a wiki. The goal here is to make package changes in a way that does not suddenly break people's ability to compile their app when getting minor or bugfix version updates to the packages they use. Actively breaking people with non-breaking-version plugin changes and then pointing them to a wiki page they had no reason to look at if they file a bug would be a bad user experience. Automatically getting the last working version is a much better experience.

So when changing minimum OS versions the goal is to make is such that the `pub` resolver *will not resolve* a version that has dropped support for an OS version that the developer's current version of Flutter still supports building for. The PR that auto-migrates their project is the point at which their version of Flutter no longer supports building it.

> But [flutter/flutter#129729](https://github.com/flutter/flutter/pull/129729) is in 3.16 at the earliest, so it sounds like this will need to wait until the next stable, at which the minimum allowed will be N-2=3.16

Packages aren't required to support N-2, they just aren't allowed to support anything older; only supporting back to current `stable` is required. But packages can't *break* people on N-1 or N-2, so if you are making a change that would break compilation for people on N-2, then you need to set the minimum Flutter SDK constraint for the package(s) in question constraint to N-1 as part of the change.

https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#supported-flutter-versions

packages/espresso/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Java classes for testing Flutter apps using Espresso.
   Allows driving Flutter widgets from a native Espresso test.
 repository: https://github.com/flutter/packages/tree/main/packages/espresso
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+espresso%22
-version: 0.3.0+7
+version: 0.3.0+8
 
 environment:
   sdk: ^3.1.0
```
stuartmorgan, line 9: Based on the discussion in comments, every package being touched here that still lists Flutter 3.13 support needs to be changed to 3.16 instead.

stuartmorgan (changes requested):

stuartmorgan:
The Dart min SDK version needs to be consistent; you may want to use the `update-min-sdk` command mentioned on https://github.com/flutter/flutter/wiki/Updating-Packages-repo-for-a-stable-release to do this. You can run it with `--base-branch main --run-on-changed-packages` (assuming you branched from `main`)

gmackall:
I ended up making the change by hand - the script made changes to every package, I believe because I had changed the legacy test.

stuartmorgan:
Ah, yes. I get bitten by that occasionally with the version bump script; I need to make a new flag that is only for these local tool usages.

stuartmorgan (approved):
LGTM

PR #6511, opened 2024-04-13
[google_maps_flutter] update ios privacy manifest

rrpadilla:
The installation instructions for Google Maps Platform SDKs for iOS now have download links and instructions for **manually** adding Privacy Manifest files for use with the latest versions of the SDKs. You can read more about this [here](https://github.com/googlemaps/google-maps-ios-utils/issues/463#issuecomment-2035734010).

This PR synchronizes the iOS privacy manifest with the file provided by Google for the Maps SDK for iOS. 
The privacy manifest was downloaded from: [https://developers.google.com/maps/documentation/ios-sdk/config#add-apple-privacy-manifest-file](https://developers.google.com/maps/documentation/ios-sdk/config#add-apple-privacy-manifest-file).

As you can see in the new **[PrivacyInfo.xcprivacy](packages/google_maps_flutter/google_maps_flutter_ios/ios/Resources/PrivacyInfo.xcprivacy)** file Google Maps SDK is using the following **NSPrivacyAccessedAPITypes**:

- **NSPrivacyAccessedAPICategoryDiskSpace**
- **NSPrivacyAccessedAPICategorySystemBootTime**
- **NSPrivacyAccessedAPICategoryFileTimestamp**
- **NSPrivacyAccessedAPICategoryUserDefaults**

*Related issues:*
- [flutter/flutter/issues/145269](https://github.com/flutter/flutter/issues/145269)
- [flutter/flutter/issues/143232](https://github.com/flutter/flutter/issues/143232)
- [flutter/flutter/issues/131940](https://github.com/flutter/flutter/issues/131940#issuecomment-2006965339)

Fixes https://github.com/flutter/flutter/issues/94491

## Pre-launch Checklist

- [x] I signed the [CLA].


cbracken (commented):
We're going to do a little digging to figure out why the Maps SDK team is requesting these steps: https://developers.google.com/maps/documentation/ios-sdk/config#add-apple-privacy-manifest-file

jmagman:
@rrpadilla Did you get an App Store warning about the GoogleMaps frameworks specifically?  And if you make the google_maps_flutter manifest changes in this PR and resubmit your app, the warning goes away? I wouldn't expect the App Store validator would accept the `google_maps_flutter` manifest as a proxy for an embedded GoogleMaps framework, but I may be misunderstanding the mechanism...

stuartmorgan:
> @rrpadilla Did you get an App Store warning about the GoogleMaps frameworks specifically?

Capturing from discussion elsewhere: it looks like the vendored frameworks in the Pod are static frameworks, and then we're also forcing static build in `google_maps_flutter` (I'm not sure why offhand; maybe because of the Info.plist key lookup?), so by the time it gets to the App Store validator it's all in `Runner`.

stuartmorgan (commented):
If we do move forward with this (it's error-prone in the long term, and the static frameworks really should be including this information directly), we will need to make updating the SDK version part of this PR. According to [the docs](https://developers.google.com/maps/documentation/ios-sdk/config#add-apple-privacy-manifest-file):

> Important: These instructions for the Apple Privacy Manifest File apply only to the latest SDK version.

That will also mean changing the min supported iOS version in our podspec.

We will need to do that regardless though, because older versions won't be compliant with the new rules anwyay.

rrpadilla:
> @rrpadilla Did you get an App Store warning about the GoogleMaps frameworks specifically? And if you make the google_maps_flutter manifest changes in this PR and resubmit your app, the warning goes away? I wouldn't expect the App Store validator would accept the `google_maps_flutter` manifest as a proxy for an embedded GoogleMaps framework, but I may be misunderstanding the mechanism...

The first time I submitted the app to apple store connect I get an email about ITMS-91053: Missing API declaration, specifically the 4 listed below. 

```
NSPrivacyAccessedAPICategoryFileTimestamp
NSPrivacyAccessedAPICategorySystemBootTime
NSPrivacyAccessedAPICategoryUserDefaults
NSPrivacyAccessedAPICategoryDiskSpace
```

I created a privacy file as recommended [here](https://github.com/flutter/flutter/issues/131940#issuecomment-2006965339) and [here](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files#4284009). This is the privacy file I uploaded and the warning goes away. Basically I just copied the **NSPrivacyAccessedAPITypes** entries from the privacy manifest provided by google [here](https://developers.google.com/maps/documentation/ios-sdk/config#add-apple-privacy-manifest-file).

```
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>NSPrivacyTracking</key>
    <false/>
    <key>NSPrivacyTrackingDomains</key>
    <array/>
    <key>NSPrivacyCollectedDataTypes</key>
    <array/>
    <key>NSPrivacyAccessedAPITypes</key>
	<array>
		<dict>
			<key>NSPrivacyAccessedAPIType</key>
			<string>NSPrivacyAccessedAPICategoryDiskSpace</string>
			<key>NSPrivacyAccessedAPITypeReasons</key>
			<array>
				<string>85F4.1</string>
			</array>
		</dict>
		<dict>
			<key>NSPrivacyAccessedAPIType</key>
			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
			<key>NSPrivacyAccessedAPITypeReasons</key>
			<array>
				<string>C617.1</string>
			</array>
		</dict>
		<dict>
			<key>NSPrivacyAccessedAPIType</key>
			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
			<key>NSPrivacyAccessedAPITypeReasons</key>
			<array>
				<string>35F9.1</string>
			</array>
		</dict>
		<dict>
			<key>NSPrivacyAccessedAPIType</key>
			<string>NSPrivacyAccessedAPICategoryUserDefaults</string>
			<key>NSPrivacyAccessedAPITypeReasons</key>
			<array>
				<string>1C8F.1</string>
			</array>
		</dict>
	</array>
</dict>
</plist>
```

### How do I know Google Maps is causing this warning?

I build a flutter app with the **google_maps_flutter** plugin only and when I run the commands listed below (recommended by @stuartmorgan [here](https://github.com/flutter/flutter/issues/145269#issuecomment-2030025509)). Then I confirmed the [GoogleMapsPrivacy.bundle](https://developers.google.com/maps/documentation/ios-sdk/config#add-apple-privacy-manifest-file) was using those entries.

```
nm build/ios/Release-iphoneos/Runner.app/Runner | grep mach_absolute_time
nm build/ios/Release-iphoneos/Runner.app/Runner | grep systemUptime
nm build/ios/Release-iphoneos/Runner.app/Runner | grep NSFileCreationDate
nm build/ios/Release-iphoneos/Runner.app/Runner | grep NSFileModificationDate
nm build/ios/Release-iphoneos/Runner.app/Runner | grep fileModificationDate
nm build/ios/Release-iphoneos/Runner.app/Runner | grep NSURLContentModificationDateKey
nm build/ios/Release-iphoneos/Runner.app/Runner | grep NSURLCreationDateKey
nm build/ios/Release-iphoneos/Runner.app/Runner | grep getattrlist
nm build/ios/Release-iphoneos/Runner.app/Runner | grep fstatat
nm build/ios/Release-iphoneos/Runner.app/Runner | grep "stat$"
```

As you can see the [Google Maps Flutter iOS privacy file](https://github.com/flutter/packages/blob/main/packages/google_maps_flutter/google_maps_flutter_ios/ios/Resources/PrivacyInfo.xcprivacy) is empty.

stuartmorgan:
> As you can see the [Google Maps Flutter iOS privacy file](https://github.com/flutter/packages/blob/main/packages/google_maps_flutter/google_maps_flutter_ios/ios/Resources/PrivacyInfo.xcprivacy) is empty.

Which should be correct, because the manifest for the plugin should only describe what the code *in the plugin* does, and its dependencies should have their own manifests that describe what they do. These Google Maps SDK instructions are very unusual.

rrpadilla:
> Which should be correct, because the manifest for the plugin should only describe what the code _in the plugin_ does, and its dependencies should have their own manifests that describe what they do. These Google Maps SDK instructions are very unusual.

@stuartmorgan What is your recommendation?
- Update the [Google Maps Flutter iOS privacy file](https://github.com/flutter/packages/blob/main/packages/google_maps_flutter/google_maps_flutter_ios/ios/Resources/PrivacyInfo.xcprivacy).
- Update the Google Maps Flutter documentation and add a reference to the [Google Maps SDK privacy file](https://developers.google.com/maps/documentation/ios-sdk/config#add-apple-privacy-manifest-file).

stuartmorgan:
My recommendation is that we wait for [this](https://github.com/flutter/packages/pull/6511#pullrequestreview-2001946172) to land anything.

stuartmorgan:
> That will also involve changing the directories in `example` to remove iOS versions that are no longer applicable, and moving their native tests forward.

Actually, I can go ahead and do these changes and push them to the PR. They are bit tricky to get right.

rrpadilla:
> > That will also involve changing the directories in `example` to remove iOS versions that are no longer applicable, and moving their native tests forward.
> 
> Actually, I can go ahead and do these changes and push them to the PR. They are bit tricky to get right.

@stuartmorgan Thanks.

Maybe we have to do this in the [podspec file](https://github.com/flutter/packages/blob/main/packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios.podspec#L26).

```
  # Version 8.4 and the versions up to 9.0, not including 9.0 and higher.
  s.dependency 'GoogleMaps', '~> 8.4'
  s.static_framework = true
  s.platform = :ios, '15.0'
```

stuartmorgan:
Based on offline conversation, we've decided to go ahead with incorporating the stand-alone privacy bundle's manifest into the plugin's manifest for now. The expectation is that future versions of the Google Maps SDK will have it built in as expected, at which point we can unwind the workaround†.

I've pushed the following changes:
- Updated the podspec to require 8.4, and correspondingly iOS 14.
    - I've updated the CHANGELOG accordingly, and also made it a minor version bump since this is now much more than a bugfix change. Arguably it should be a major version bump since it can break people's builds, but Flutter 3.16 will give a clear error message explaining why and what the next steps are, and since earlier versions of the plugin won't be compatible with the App Store starting on May 1st anyway, the incremental impact of the breakage is, IMO, small. So I'm not doing a major version bump since that would just make the plumbing more complicated for us, and actually make it harder for people to get the fix they will absolutely need.
- I've removed the no-longer-useful `ios12` and `ios13` example directories, and folded the tests from `ios12` into `ios14`. (Technically I actually deleted `ios14`, renamed `ios12`, and then back-ported the minor changes into `ios12` to minimize potential errors.)
    - I left the multi-OS-version example structure in place because we'll need it again soon; the SDK release notes have already indicated that sometime this year iOS 14 will be dropped, so we'll want to go back to supporting a range. (@cbracken feel free to set up a quick VC with me to explain the unusual test structure and the context around it.)
- Since we are no longer supporting older versions of the SDK, I updated the code that was doing dynamic symbol lookup to get new features to just use them normally, and resolved a TODO about a deprecated API.
- Since we no longer need it, I removed the arm64 simulator workaround, fixing https://github.com/flutter/flutter/issues/94491

(†If the version that includes the manifest is also the version that drops iOS 14, which may well be the case based on the timings listed in the SDK release notes, we'll need to think about whether we want to unwind the workaround at the cost of losing iOS 14 support.)

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/Flutter/AppFrameworkInfo.plist
```diff
@@ -25,6 +25,6 @@
     <string>arm64</string>
   </array>
   <key>MinimumOSVersion</key>
-  <string>13.0</string>
+  <string>12.0</string>
```
cbracken, line 28: Shouldn't this be 14.0?

cbracken (commented):
lgtm modulo one comment.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/Flutter/AppFrameworkInfo.plist
stuartmorgan: I don't think so; AFAICT this is the plist for the `App.framework` that holds the compiled Dart code, and there's nothing in *that* framework that requires a higher deployment target than Flutter itself. The 13 here was the result of me doing an overzealous find/replace when I make `ios13` from `ios12` (and `ios13` was copied to make the original `ios14`).

stuartmorgan: (The one that matters is the Runner's `Info.plist`, and that is correctly set to 14.0.)

cbracken: sgtm! Thanks!

cbracken (approved):

stuartmorgan:
@jmagman Do you want to sign off on [this](https://github.com/flutter/packages/pull/6511#issuecomment-2059342641) as well? (In particular, dropping iOS 12 and 13 support without a breaking change, just the tooling error.)

jmagman (approved):
> - Since we no longer need it, I removed the arm64 simulator workaround, fixing https://github.com/flutter/flutter/issues/94491

I knew this day would eventually come.  🙂 

LGTM!

@rrpadilla thanks for bringing the unusual GoogleMaps case to our attention.

PR #6535, opened 2024-04-15
[video_player_android] Migrate ExoPlayer to ExoPlayer-Media3 1.3.1

emakar:
Resolves [#130272](https://github.com/flutter/flutter/issues/130272)

Migrated with https://developer.android.com/media/media3/exoplayer/migration-guide#usingscript

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/video_player/video_player_android/android/build.gradle
```diff
@@ -48,11 +48,12 @@ android {
     }
 
     dependencies {
-        def exoplayer_version = "2.18.7"
-        implementation "com.google.android.exoplayer:exoplayer-core:${exoplayer_version}"
-        implementation "com.google.android.exoplayer:exoplayer-hls:${exoplayer_version}"
-        implementation "com.google.android.exoplayer:exoplayer-dash:${exoplayer_version}"
-        implementation "com.google.android.exoplayer:exoplayer-smoothstreaming:${exoplayer_version}"
+        def exoplayer_version = "1.1.1"
```
emakar, line 51: version 1.1.1 corresponds to exoplayer 2.19.1
version 1.3.1 is the latest one https://github.com/androidx/media/blob/release/RELEASENOTES.md

packages/video_player/video_player_android/android/build.gradle
```diff
@@ -48,11 +48,12 @@ android {
...
+        implementation "androidx.media3:media3-exoplayer:${exoplayer_version}"
+        implementation "androidx.media3:media3-exoplayer-hls:${exoplayer_version}"
+        implementation "androidx.media3:media3-exoplayer-dash:${exoplayer_version}"
+        implementation "androidx.media3:media3-exoplayer-smoothstreaming:${exoplayer_version}"
+        debugImplementation "androidx.annotation:annotation-experimental:1.3.0"
```
emakar, line 56: `androidx.media3:media3-common:1.1.1` depends on `androidx.annotation:annotation-experimental:1.2.0`

however annotation-experimental 1.2.0 leads to lint failure:
```
jetified-annotation-experimental-1.2.0/jars/lint.jar: Error: Library lint checks out of date;
```

so, since lintDebug only is used - bump annotation-experimental to 1.3.0 in debug builds

media3 uses annotation-experimental 1.3.1 since https://github.com/androidx/media/releases/tag/1.2.0, and this could be removed later

packages/video_player/video_player_android/android/build.gradle
camsim99: Can you bump to the latest version?

packages/video_player/video_player_android/android/build.gradle
camsim99: What do you mean by removing it later? Is there a reason we shouldn't use it? Also, if we bump to 1.3.1 anyway, that will align.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -40,6 +42,7 @@
 import java.util.List;
 import java.util.Map;
 
+@OptIn(markerClass = UnstableApi.class)
```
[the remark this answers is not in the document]
emakar: updated, is it okay?

packages/video_player/video_player_android/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.4.14
```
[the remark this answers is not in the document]
emakar: bumped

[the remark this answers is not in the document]
stuartmorgan: I would also err on the side of making this a minor version bump even though it's not technically necessary according to semver.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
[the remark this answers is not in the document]
emakar: so, not migration to media3 then (at least until there is a stable alternative for such apis?)
or am I missing something?

packages/video_player/video_player_android/CHANGELOG.md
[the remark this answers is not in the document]
emakar: changed minor

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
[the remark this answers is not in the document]
emakar: another solution I see is adding non-endorsed plugin - video_player_android_media3 until media3 stabilizes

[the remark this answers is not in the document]
emakar: though, wasn't it the same all the time?
https://github.com/androidx/media/issues/331#issuecomment-1512711590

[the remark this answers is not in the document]
stuartmorgan: The fact that the old code was silently doing something dangerous doesn't mean we should explicitly opt the entire plugin into continuing to do that, or putting a comment that describes it as safe when it isn't.

What is the exact list of APIs we are using that are unstable? We should be using highly targeted annotations, and we should investigate each of them to determine if they are truly necessary, and if so comment why, and about the danger they pose.

packages/video_player/video_player_android/android/build.gradle
stuartmorgan: [These are the steps](https://github.com/flutter/packages/blob/9a2490419268365911191f9d8388f986111bcc24/.ci/targets/android_build_all_packages.yaml#L14-L25) to run that locally.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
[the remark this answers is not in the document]
emakar: >What is the exact list of APIs we are using that are unstable?

updated pr with fine grained opt-in
didn't find stable alternative for e.g. setMediaSource (whereas MediaSource itself is stable)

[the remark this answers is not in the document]
emakar: do I understand it correctly - even if we are using stable only API - any client can still use unstable api on its own or with some dependency, which potentially breaks it

so
1. all good with whatever api if we are the only dependency
2. high chance things would break if both video_player_android and some other dependency use unstable api with different minor/patch versions
3. even we use stable only api - any other dependency with unstable api gives problems - if it depends on older version

however, I don't understand how even stable api helps if media3 bumps major version (legally changing stable api) - gradle would still prefer it by default afaik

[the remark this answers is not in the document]
stuartmorgan: > any client can still use unstable api on its own or with some dependency, which potentially breaks it

I don't follow. How would plugin code using only stable APIs be broken by something else using unstable APIs?

> even we use stable only api - any other dependency with unstable api gives problems - if it depends on older version

If some other plugin chooses to use unstable APIs, and doesn't do any strict version control, then *it* could break, yes. That's the nature of using unstable APIs. We can't control what other developers choose to do in their own plugins. We can only attempt to make our own code robust and stable.

> however, I don't understand how even stable api helps if media3 bumps major version (legally changing stable api) - gradle would still prefer it by default afaik

Gradle does optimistic upgrades across *major* version changes? If so, that's bad, and we should revisit our versioning constraints in general. /cc @reidbaker Are you familiar with the details of how optimistic upgrades work?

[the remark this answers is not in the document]
emakar: > I don't follow. How would plugin code using only stable APIs be broken by something else using unstable APIs?

Plugin won't break its code - it would break client code if one uses older version
e.g. client uses setMediaSource in version 1.1.1, which is removed in version 1.4.0 which plugin uses (despite its all stable api in plugin)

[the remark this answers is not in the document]
emakar: > Gradle will consider all requested versions, wherever they appear in the dependency graph. Out of these versions, it will select the highest one. More information on version ordering [here](https://docs.gradle.org/current/userguide/single_versions.html#version_ordering).

https://docs.gradle.org/current/userguide/dependency_resolution.html

> In case of conflict, Gradle by default uses the newest of conflicting versions

https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ResolutionStrategy.html#org.gradle.api.artifacts.ResolutionStrategy:dependencySubstitution

packages/video_player/video_player_android/android/build.gradle
emakar: reproduced, thanks (one also needs `running prepare_tool.sh` and `create_all_packages_app.sh legacy`)

the issue could be simplified to:
- use gradle 6.x
- bump guava version to 32.1.x (media3-1.3.1 uses guava 32.1.3-android) in any plugin (e.g. video_player_android like here or camera_android_camerax)

then we'd face following - https://github.com/google/guava/issues/6801
gradle 7+ handles this fine with 32.1.2+
gradle 6 seem to need workarounds described at https://github.com/google/guava/releases/tag/v32.1.0 (not sure, investigating)

so, this is a generic issue with legacy gradle 6 project and guava 32, should I file it separately and use compatible media3 version in this pr?

emakar: added workaround for gradle 6.x

packages/video_player/video_player_android/android/build.gradle
```diff
@@ -47,12 +47,23 @@ android {
...
+    // Gradle 6.x workaround for
+    // see "Selecting the appropriate flavor"
+    // https://github.com/google/guava/releases/tag/v32.1.0
+    dependencies.constraints {
+        implementation("com.google.guava:guava") {
+            attributes {
+                attribute(Attribute.of("org.gradle.jvm.environment", String), "android")
+            }
+        }
+    }
```
emakar, line 59: gradle 7+ handles this fine with 32.1.2+
gradle 6 needs workaround described at
[google/guava@v32.1.0 (release)](https://github.com/google/guava/releases/tag/v32.1.0)
https://github.com/google/guava/issues/6612#issuecomment-1614992368

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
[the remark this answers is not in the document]
reidbaker: @stuartmorgan I think this comment is saying that gradle considers all dependencies in the scope of the app which means that versions specified in the plugin could be overridden by dependencies the app (or another plugin) has. Which both feels like a gotcha that might exist and also not how build systems should work and is not something I was aware of before.

packages/video_player/video_player_android/android/build.gradle
emakar: welp, it's fixed now for video_player_android, but fails somehow for project :espresso with 30.1.1-android (all_packages)
still investigating

packages/espresso/android/build.gradle
```diff
@@ -79,8 +79,18 @@ dependencies {
...
+    // dependencies.constraints {
+    //     api("com.google.guava:guava") {
+    //         attributes {
+    //             attribute(Attribute.of("org.gradle.jvm.environment", String), "android")
+    //         }
+    //     }
+    // }
+    // 2. make api dependency on guava instead implementation
+    implementation 'androidx.test.ext:truth:1.5.0'
+    implementation 'com.google.truth:truth:1.1.3'
```
emakar, line 93: issue with espresso was because workaround https://github.com/flutter/packages/pull/6535/commits/a3800a16550faf8ce7a1c56a893883fafe635901 works for `implementation` configuration, not `api`
and for it to work with `api` one needs `api` dependency on guava as well (which I think is worse solution)

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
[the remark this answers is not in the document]
stuartmorgan: I think it's much better to clearly scope the use of unstable APIs as the current commit does, if (as sounds like is the case) we are stuck with using them for now. I would also comment the annotations with a TODO pointing to an issue discussing the issue and the need to track a long term solution.

Based on the discussion above it sounds like:
- Short term, we are no worse off than before by moving forward with the migration (but @reidbaker and @camsim99 should weigh in as well) in terms of potential for the plugin to stop compiling for clients due to upstream changes.
- Long term, we need to revisit how we express dependencies in our Android plugins, to see if we can prevent optimistic updates across major versions. I'll file an issue for that.
- Long term, we need to revisit the use of unstable APIs in this plugin (the TODO).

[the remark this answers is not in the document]
emakar: added TODOs and created https://github.com/flutter/flutter/issues/147039

packages/espresso/android/build.gradle
```diff
@@ -79,8 +79,18 @@ dependencies {
 
     // Assertions
     api 'androidx.test.ext:junit:1.1.5'
-    api 'androidx.test.ext:truth:1.5.0'
-    api 'com.google.truth:truth:1.1.3'
+    // Don't make these 'api', otherwise another workaround for guava 32.1.+ on gradle 6.x is needed:
```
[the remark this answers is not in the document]
emakar: They might (I don't know whether it was intended)

It just lesser evil between workarounds I've found:
- either demote api here to implemenation (and since all_packages are built with it - it seemed ok to me)
- or add api to guava dependency (see https://github.com/flutter/packages/pull/6535#discussion_r1567278787 and https://github.com/flutter/packages/pull/6535#discussion_r1567520344)
- bumg minimal supported gradle to 7.x, since gradle 6 only is affected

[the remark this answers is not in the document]
stuartmorgan: > and since all_packages are built with it - it seemed ok to me

The deciding factor isn't whether one particular configuration happened to build, it's whether types from these dependencies are part of the public API surface.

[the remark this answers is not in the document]
emakar: Yes, and they are part of public API.
I just don't know what workaround to pick from, or any other solution
What do you prefer from mentioned above?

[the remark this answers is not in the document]
emakar: @stuartmorgan if we can't decide now on concrete workaround, I'd propose
- creating separate issue on guava 32.1.+ with gradle 6. It happened that media3 1.3 depends on newer guava, but there are other plugins - camera, espresso, google_sign_in, etc - which face the same issue once they bump guava dependency
- leave media3 at 1.1.1 or 1.2.1, remove commits a72c25811232b5264ad3a90e9f1f54d695b330c3 20d6823fa19fc90fdb5c20747d6c37d349cd6034

what do you think?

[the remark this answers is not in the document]
emakar: I'll remove workarounds for gradle 6 since all_projects uses 7 now - https://github.com/flutter/packages/pull/6591

packages/espresso/android/build.gradle
emakar: dropped commit

packages/video_player/video_player_android/android/build.gradle
emakar: dropped commit

emakar:
I've dropped controversial workarounds since they aren't needed anymore (all_packages uses gradle 7 now)
This is ready for review, please let me know if anything else is needed

stuartmorgan:
> I've dropped controversial workarounds since they aren't needed anymore (all_packages uses gradle 7 now)

Per [these comments](https://github.com/flutter/packages/pull/6591#issuecomment-2074789315) I think we should probably hold this until the next `stable` release if it relies on those changes.

gmackall:
From android triage: The next stable has been released, is this ready for re-review?

reidbaker:
@emakar Is this still something you plan to get to if not we understand and appreciate the effort you have put in so far.

emakar:
@reidbaker yeah, It seems https://github.com/flutter/packages/pull/6742 was a reland to main, but it's not in 3.22 and a next stable material?

or is this pr now unblocked for merge (once conflicts are resolved)?

stuartmorgan:
> but it's not in 3.22

Nothing in this repository is released as part of a Flutter release; everything here is published through `pub.dev`.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -85,7 +88,7 @@ final class VideoPlayer {
 
     MediaSource mediaSource = buildMediaSource(uri, dataSourceFactory, formatHint);
 
-    exoPlayer.setMediaSource(mediaSource);
+    unstableSetMediaSource(exoPlayer, mediaSource);
```
stuartmorgan, line 91: Do we actually need to use `MediaSource` at all? It looks like we're doing two things with the media source:
1. Setting the source type.
2. Injecting the data source factory.

From the APIs, it looks like 1. can be done with [`MediaItem` MIME types](https://developer.android.com/media/media3/exoplayer/media-items#handling-non-standard) and 2 can be done [as part of `ExoPlayer` creation](https://developer.android.com/media/media3/exoplayer/media-sources#customizing-media-source-creation). Are there reasons those wouldn't work?

emakar: Well, concrete MediaSource.Factory implementations (which are to be set to builder) are UnstableApi as well

so instead of unstable player.setMediaSource we'd have unstable createSomeFactory

emakar: This is however one unstable call less
I'll push a commit so you can see whether it's ok

stuartmorgan: Why do we need to create the concrete factories. Don't we just pass in `dataSourceFactory`, and then the internal implementation of ExoPlayer creates the concrete factories from types as needed when it internally creates `MediaSource`s from the `MediaItem`s we pass it? My impression from the docs is that `unstableBuildMediaSource` is largely duplicating logic that ExoPlayer already has.

emakar: Oh, wasn't sure DefaultMediaSourceFactory is enough, need to check it

icbaker: ExoPlayer dev here chiming in :)

I think the `MediaItem` way of writing this code (without `MediaSource`) would be more like:

```
ExoPlayer exoPlayer = new ExoPlayer.Builder(context)
    .setMediaSourceFactory(new DefaultMediaSourceFactory(dataSourceFactory))
    .build();
    
mediaItem = new MediaItem.Builder
    .setUri(uri)
    .setMimeType(mimeFromFormatHint(formatHint))
    .build();

exoPlayer.setMediaItem(mediaItem);
```

`mimeTypeFromFormatHint` would just map from your own constants to those in [`MimeTypes`](https://developer.android.com/reference/androidx/media3/common/MimeTypes).

See also https://developer.android.com/media/media3/exoplayer/media-items and some examples at https://developer.android.com/media/media3/exoplayer/customization too.

stuartmorgan: @icbaker Thanks for the confirmation!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -341,15 +322,15 @@ void sendInitialized() {
       event.put("event", "initialized");
       event.put("duration", exoPlayer.getDuration());
 
-      if (exoPlayer.getVideoFormat() != null) {
-        Format videoFormat = exoPlayer.getVideoFormat();
+      Format videoFormat = unstableGetVideoFormat(exoPlayer);
```
stuartmorgan, line 325: Can we use `getVideoSize` here instead of `getVideoFormat`? It appears that we just need width, height, and rotation, all of which [`VideoSize`](https://developer.android.com/reference/androidx/media3/common/VideoSize) has.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
emakar: thanks! updated the branch

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
emakar: thanks, updated

not sure I've applied rotation changes correctly though

emakar:
Hm, this test is now failing accidentally
```
The following TestFailure was thrown running a test:
Expected: a value greater than Duration:<0:00:00.000000>
  Actual: Duration:<0:00:00.000000>
   Which: is not a value greater than Duration:<0:00:00.000000>

#4      main.<anonymous closure>.<anonymous closure> (file:///b/s/w/ir/x/w/packages/packages/video_player/video_player_android/example/integration_test/video_player_test.dart:71:7)
<asynchronous suspension>
#5      testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:183:15)
<asynchronous suspension>
#6      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:1025:5)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)

This was caught by the test expectation on the following line:
  file:///b/s/w/ir/x/w/packages/packages/video_player/video_player_android/example/integration_test/video_player_test.dart line 71
The test description was:
  can be played
```

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -120,63 +123,14 @@ public void buildHttpDataSourceFactory(@NonNull Map<String, String> httpHeaders)
...
-        {
-          throw new IllegalStateException("Unsupported type: " + type);
-        }
-    }
+    unstableUpdateDataSourceFactory(
+        httpDataSourceFactory, httpHeaders, userAgent, httpHeadersNotEmpty);
   }
 
   public void recreateSurface(Context context) {
     ExoPlayer exoPlayer = new ExoPlayer.Builder(context).build();
```
icbaker, line 131: This line looks dodgy - I think you need to have the same new 'init ExoPlayer' code you have above here.

emakar: whoops, you are right!

stuartmorgan: > I suspect this may be due to the test calling into the `@VisibleForTesting` constructor above which does 'less' than the 'real' constructor.

Integration tests don't use any test-only Java code, they are driving the production native code from Dart. Only Java unit tests would call `@VisibleForTesting` codepaths.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -377,4 +334,36 @@ void dispose() {
       exoPlayer.release();
     }
   }
+
+  @Nullable
+  private static String mimeFromFormatHint(String formatHint) {
```
stuartmorgan, line 339: Nit: it would be good to annotate the parameter as nullable as well.

stuartmorgan (approved):
Thanks for all the iteration on this to put the plugin on a more stable API foundation!

A couple of minor comments, and @camsim99 will need to re-review as well, but this has my LGTM.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -80,16 +76,17 @@ final class VideoPlayer {
...
+            .setUri(dataSource)
+            .setMimeType(mimeFromFormatHint(formatHint))
+            .build();
 
     buildHttpDataSourceFactory(httpHeaders);
-    DataSource.Factory dataSourceFactory =
-        new DefaultDataSource.Factory(context, httpDataSourceFactory);
 
-    mediaSource = buildMediaSource(uri, dataSourceFactory, formatHint);
+    ExoPlayer exoPlayer = buildExoPlayer(context);
```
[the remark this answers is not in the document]
emakar: sounds good, will do

emakar:
thanks for the review!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -110,7 +111,7 @@ final class VideoPlayer {
   }
 
   @VisibleForTesting
-  public void buildHttpDataSourceFactory(@NonNull Map<String, String> httpHeaders) {
+  public void setupHttpDataSourceFactory(@NonNull Map<String, String> httpHeaders) {
```
stuartmorgan, line 114: Nit: "setup" is a noun, the verb is "set up", so this should be "setUpHttpDataSourceFactory".

stuartmorgan (approved):
Thanks!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
emakar: oh, well, this now looks like a settter?
chosed configure if you don't mind

emakar:
@camsim99 could you please review this?

camsim99 (approved):
This looks good to me!

PR #5523, opened 2023-11-30
[image_picker] Adopt code excerpts in README

mike-v2:
Improves image_picker_for_web README example and updates it to use code excerpts.

Part of [flutter/flutter#102679](https://github.com/flutter/flutter/issues/102679)

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #6537, opened 2024-04-15
Update local_auth_android minSdkVersion to 19

stuartmorgan:
https://github.com/flutter/packages/pull/6506 was supposed to update everything to 19+, but missed local_auth_android. This updates it to require 19 as well.

### DIFF
--- a/packages/local_auth/local_auth_android/android/build.gradle
+++ b/packages/local_auth/local_auth_android/android/build.gradle
@@ -29,7 +29,7 @@ android {
     compileSdk 34
 
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion flutter.minSdkVersion
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
 
--- a/packages/local_auth/local_auth_android/android/build.gradle
+++ b/packages/local_auth/local_auth_android/android/build.gradle
@@ -29,7 +29,7 @@ android {
     compileSdk 34
 
     defaultConfig {
-        minSdkVersion flutter.minSdkVersion
+        minSdkVersion 19
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
 
--- a/packages/local_auth/local_auth_android/CHANGELOG.md
+++ b/packages/local_auth/local_auth_android/CHANGELOG.md
@@ -1,6 +1,7 @@
-## NEXT
+## 1.0.38
 
-* Updates minimum supported SDK version to Flutter 3.13/Dart 3.1.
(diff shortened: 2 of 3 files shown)


stuartmorgan:

(Also, does fix a broken test)

stuartmorgan:
Example OOB presubmit failure this fixes: https://github.com/flutter/packages/pull/5523#issuecomment-2057130606

reidbaker (approved):

packages/local_auth/local_auth_android/android/build.gradle
```diff
@@ -29,7 +29,7 @@ android {
     compileSdk 34
 
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion flutter.minSdkVersion
```
reidbaker, line 32: I think this needs to be a 19

reidbaker (approved):

packages/local_auth/local_auth_android/android/build.gradle
stuartmorgan: A lot of our existing examples use `flutter.minSdkVersion`; I thought we only wanted to use a number when a specific example needed something higher than the min (e.g., the example for a plugin that requires 21).

stuartmorgan: Wait, this isn't the `/app/` file. Why is this line even here? I need to look at this again, I missed the path on this.

stuartmorgan: Oh, it's not the example, it's the actual plugin. I really can't read this morning.

Now I'm confused as to what's causing the error in the example app though. This should probably be 19 and get a version bump, but the example app is already using `flutter.minSdkVersion`.

stuartmorgan:
Okay, I think I see now. The error message had `example` in the build path because that's where we run the command, but the failing target was in `local_auth_android` itself, not the example, so it looks like the issue was that `local_auth_android` was trying to depend on another plugin (`flutter_plugin_android_lifecycle`) that required 19.

gmackall (approved):

gmackall:
So for my understanding, the reason why presubmits wouldn't have caught this on my PR is that the lint step here would have been running against the latest published version of `flutter_plugin_android_lifecycle` at the time (not the version that was modified in the PR)?

stuartmorgan:
Correct. We don't do a pathified version of native testing because it would be really expensive, and it's rarely a problem, but it's something we could reconsider.

(What I really wish was that we had an easy way to do optional tests, like being able to manually trigger pathified native Android tests on a PR like yours. Maybe we could do some trickery with a dummy path and path-based rules...)

PR #6540, opened 2024-04-15
[in_app_purchase] Add api to expose country code

reidbaker:
- **Add changelog and version bump**
- **dependency override from package tool**

Fixes flutter/flutter/issues/141627

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #6548, opened 2024-04-16
[in_app_purchase_platform_interface] Adds countryCode API

reidbaker:
Original pr https://github.com/flutter/packages/pull/6540

Related to https://github.com/flutter/flutter/issues/141627
## Pre-launch Checklist

- [x] I signed the [CLA].


LouiseHsu (approved):
lgtm!

stuartmorgan (approved):
LGTM. Usually we try to avoid documenting platform-specific behavior in the app-facing package for scalability and separation of concerns, but it seems like we probably can't provide much useful documentation here without it.

gmackall (approved):

reidbaker:
> LGTM. Usually we try to avoid documenting platform-specific behavior in the app-facing package for scalability and separation of concerns, but it seems like we probably can't provide much useful documentation here without it.

FWIW I agree I debated just linking to the underlying documentation but thought it would be easier for customers to just see the content where they were looking. If we decided we wanted to convert the underlying country codes into some standard then enforce it ourselves we could avoid the platform specific references but I think that most of the time the answer will either be the same or values will get passed to a system that assumes the platform implementation.

PR #6550, opened 2024-04-16
[two_dimensional_scrollables] Refactor Spans for common use

Piinks:
This relocates most of `table_span.dart` to `../common/span.dart`, renaming the classes to basic Spans instead of TableSpans.
This is because since a Span can represent a row or a column, the new base Span class will be used to build TreeRows in the upcoming TreeView.
In `table_span.dart`, the old classes are now typedefs. This is because (IMO) the Table and Tree APIs are better to work with when the naming conventions reflect what you are building. The Table and Tree uses of Span could also deviate in the future and it felt prudent to leave room for future changes that may be specific to only one or the other.

Otherwise a bit of tidying to get ready for TreeView!

Part of https://github.com/flutter/flutter/issues/42332

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/two_dimensional_scrollables/lib/src/common/implementation_span.dart
goderbauer, line 1: Should this file be named implemented_span since the (only) class in it is called ImplementedSpan?

packages/two_dimensional_scrollables/lib/src/common/implementation_span.dart
```diff
@@ -0,0 +1,141 @@
...
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+
+import 'span.dart';
+
+// This file is NOT exported as part of the package.
```
goderbauer, line 12: I wonder if we could enforce this somehow with a lint...

packages/two_dimensional_scrollables/lib/src/common/implementation_span.dart
```diff
@@ -0,0 +1,141 @@
...
+
+/// A contextual used by the [RenderTableViewport].
```
goderbauer, line 14: Is this missing a word? A contextual what?

packages/two_dimensional_scrollables/lib/src/common/implementation_span.dart
goderbauer: This is also violating the Flutter style, which requires everything to be exported (unless it is private with an underscore and hence not exportable): https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#structure

Why is it so important that this isn't exported? Could I use this as a building block for my own 2D scrollable?

packages/two_dimensional_scrollables/lib/src/common/implementation_span.dart
```diff
@@ -0,0 +1,141 @@
...
+///
+/// Used to track the layout information, and additional configurations of the
+/// span during multiple layout passes.
+abstract class ImplementedSpan
```
goderbauer, line 18: The name is kinda strange since it doesn't `implements Span`. If there's not a more descriptive name, would BaseSpan be appropriate (as in it is the base class for Spans)?

packages/two_dimensional_scrollables/lib/src/common/implementation_span.dart
```diff
@@ -0,0 +1,141 @@
...
+import 'package:flutter/widgets.dart';
+
+import 'span.dart';
+
+// This file is NOT exported as part of the package.
+
+/// A contextual used by the [RenderTableViewport].
+///
+/// Used to track the layout information, and additional configurations of the
+/// span during multiple layout passes.
```
goderbauer, line 17: The docs could say a word or two about its relationship to the Span class.

packages/two_dimensional_scrollables/lib/src/common/implementation_span.dart
```diff
@@ -0,0 +1,141 @@
...
+  /// The leading edge of the span in the viewport, orientated to the column or
+  /// row it represents.
+  double get leadingOffset => _leadingOffset;
+  late double _leadingOffset;
+
+  /// The extent of the span.
+  double get extent => _extent;
+  late double _extent;
+
+  /// The Span that was used to populate this implemented span.
```
goderbauer, line 30:
```suggestion
  /// The [Span] that was used to populate this implemented span.
```

packages/two_dimensional_scrollables/lib/src/common/implementation_span.dart
Piinks: Oh I thought maybe that did not apply in flutter/packages? 😅  Since not everything in the style guide applies to all repos..

Regardless, I am more concerned about folks using this than the alternative of a little code duplication between the tree and the table, so I will put this back where it was as a private class.

 I don't want folks to be confused by the Span (+subclasses) they are meant to be using, and this one that is really just meant for internal use, especially as we would not want folks to use this and manipulate the information that should only be managed by the render object.

goderbauer (approved):
LGTM

PR #6551, opened 2024-04-17
[ci] Add Linux desktop support to Linux custom_package_tests

stuartmorgan:
We're in the process of adding Linux desktop support to Pigeon (https://github.com/flutter/packages/pull/5100), so we need the Linux desktop dependencies in the relevant tasks.


tarrinneal (approved):

.ci.yaml
```diff
@@ -322,12 +322,14 @@ targets:
       version_file: flutter_master.version
       target_file: linux_custom_package_tests.yaml
       # Pigeon tests need Andoid deps (thus the Linux_android base), emulator,
-      # and clang-format.
+      # and Linux desktop deps.
```
robert-ancell, line 325: Was the comment about clang-format intentionally removed? I think it's still used in the tests right?

robert-ancell (approved):
Other than the question about the comment this looks good.

.ci.yaml
stuartmorgan: Clang is part of the Linux desktop deps, so I was simplifying the comment rather than have two different explanations for why clang was in the list.

PR #6553, opened 2024-04-17
[camera] Initial iOS Pigeon conversion

stuartmorgan:
Converts one platform channel method to Pigeon, setting up all the Pigeon plumbing and test scaffolding.

The Camera API surface is relatively large, so this lays a foundation for incremental conversion, minimizing the mixing of Pigeon setup with the individual method conversions.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera_avfoundation/example/ios/RunnerTests/AvailableCamerasTest.m
```diff
@@ -15,8 +15,7 @@ @implementation AvailableCamerasTest
 
 - (void)testAvailableCamerasShouldReturnAllCamerasOnMultiCameraIPhone {
   CameraPlugin *camera = [[CameraPlugin alloc] initWithRegistry:nil messenger:nil];
-  XCTestExpectation *expectation =
-      [[XCTestExpectation alloc] initWithDescription:@"Result finished"];
+  XCTestExpectation *expectation = [self expectationWithDescription:@"Result finished"];
```
stuartmorgan, line 18: The change to the utility method allows using `waitForExpectationsWithTimeout:` instead of having to list every expectation.

packages/camera/camera_avfoundation/example/ios/RunnerTests/AvailableCamerasTest.m
```diff
@@ -55,29 +54,26 @@ - (void)testAvailableCamerasShouldReturnAllCamerasOnMultiCameraIPhone {
...
-                         }];
+  __block NSArray<FCPPlatformCameraDescription *> *resultValue;
+  [camera
+      availableCamerasWithCompletion:^(NSArray<FCPPlatformCameraDescription *> *_Nullable result,
+                                       FlutterError *_Nullable error) {
+        XCTAssertNil(error);
+        resultValue = result;
+        [expectation fulfill];
+      }];
+  [self waitForExpectationsWithTimeout:30 handler:nil];
```
stuartmorgan, line 65: I'm not sure why we weren't actually using the expectation before, but now it's necessary because the queue dispatch isn't being bypassed in the test like it was before, so the block above is no longer run synchronously in the test.

packages/camera/camera_avfoundation/example/ios/RunnerTests/AvailableCamerasTest.m
```diff
@@ -55,29 +54,26 @@ - (void)testAvailableCamerasShouldReturnAllCamerasOnMultiCameraIPhone {
...
 
   // Verify the result
-  NSDictionary *dictionaryResult = (NSDictionary *)resultValue;
   if (@available(iOS 13.0, *)) {
-    XCTAssertTrue([dictionaryResult count] == 4);
+    XCTAssertEqual(resultValue.count, 4);
```
stuartmorgan, line 69: This is just a quality of life improvement for when the test fails. (I had a failure initially because I didn't add the `wait`, so `resultValue` was nil, and I got a failure with no actual info.)

packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
```diff
@@ -104,8 +106,12 @@ - (void)handleMethodCall:(FlutterMethodCall *)call result:(FlutterResult)result
...
 }
 
-- (void)handleMethodCallAsync:(FlutterMethodCall *)call result:(FlutterResult)result {
-  if ([@"availableCameras" isEqualToString:call.method]) {
+- (void)availableCamerasWithCompletion:
+    (nonnull void (^)(NSArray<FCPPlatformCameraDescription *> *_Nullable,
+                      FlutterError *_Nullable))completion {
+  // This doesn't interact with FLTCam, so can use an arbitrary thread rather than
+  // captureSessionQueue. It should still not be done on the main thread, however.
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
stuartmorgan, line 114: We will unfortunately need to have the thread dispatch in every individual method in the Pigeon version (instead of handleMethodCall doing it for everything at once). Almost everything needs to use `captureSessionQueue`, since `FLTCam` sometimes needs to redispatch to that queue in internal callbacks to interact with instance state, which means we can't just use task queues.

(Also IIRC task queues aren't implemented for macOS, and we want to enable this plugin for macOS at some point.)

It'll just be a couple of lines in each method though, so it shouldn't be too bad. I think we can simplify when we migrate to Swift.

packages/camera/camera_avfoundation/test/utils_test.dart
```diff
@@ -2,39 +2,29 @@
...
       );
       expect(
-        parseCameraLensDirection('external'),
+        cameraLensDirectionFromPlatform(PlatformCameraLensDirection.external),
         CameraLensDirection.external,
       );
     });
 
-    test(
-        'Should throw ArgumentException when invalid value is supplied when parsing camera lens direction',
```
stuartmorgan, line 30: This is no longer possible, so we don't need a test for it.

packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
hellohuanlin: I think it's still better to use `captureSessionQueue` for all session configurations, in case Apple made any assumptions in its internal implementation (e.g. I wouldn't be surprised if multiple capture session related API access some common internal states).

(`AVCaptureDeviceDiscoverySession`'s API doc doesn't say anything, but Apple's [sample project](https://developer.apple.com/documentation/avfoundation/capture_setup/avcam_building_a_camera_app) uses this convention)

stuartmorgan: Done. It hadn't occurred to me that in theory the unique device IDs could be thread-specific; that would be weird, but not implausible.

hellohuanlin: Ya I bet we should be fine either way, but just wanna be extra safe here.

hellohuanlin (approved):

packages/camera/camera_avfoundation/pubspec.yaml
```diff
@@ -20,13 +20,15 @@ dependencies:
   camera_platform_interface: ^2.7.0
   flutter:
     sdk: flutter
+  pigeon: ^18.0.0
```
emakar, line 23: shouldn't it come as dev dependency?

stuartmorgan: Yes, see https://github.com/flutter/packages/pull/6563

PR #6556, opened 2024-04-17
[in_app_purchase] Add countryCode implementation to android and storekit

reidbaker:
Part of flutter/flutter/issues/141627
reviewed pr https://github.com/flutter/packages/pull/6540

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/in_app_purchase/in_app_purchase_android/pubspec.yaml
```diff
@@ -2,11 +2,11 @@ name: in_app_purchase_android
...
 repository: https://github.com/flutter/packages/tree/main/packages/in_app_purchase/in_app_purchase_android
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+in_app_purchase%22
-version: 0.3.3
+version: 0.3.4
 
 environment:
-  sdk: ^3.1.0
-  flutter: ">=3.13.0"
+  sdk: ^3.2.3
+  flutter: ">=3.16.6"
```
stuartmorgan, line 9: Are these necessary? My comment in the main PR was specific to the app-facing package, since it depends on the iOS implementation that requires 3.16.6. This can probably continue to support 3.13.

reidbaker: I don't think so but I thought there might be some value in them all being the same range. I will revert this change.

stuartmorgan: It's actually the opposite, there's (a small amount of) value in not updating ranges when we don't actually need to. E.g., if the next thing that lands in iap_android is a bugfix, by not making this change people who are still using 3.13 can still get that bugfix if they `flutter pub upgrade`.

We don't go out of our way to make fixes available to people not on the latest version, but if we can do so without impacting our development it's a nice bonus.

stuartmorgan (approved):
LGTM

PR #6561, opened 2024-04-17
[in_app_purchase_storekit] migrate main plugin class to swift in preperation to storekit 2

LouiseHsu:
Part of https://github.com/flutter/flutter/issues/119106

This PR migrates the main InAppPurchasePlugin.m class to swift.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/in_app_purchase_storekit.h
```diff
@@ -0,0 +1,15 @@
...
+//  in_app_purchase_storekit.h
+//  Pods
+//
+//  Created by Louise Hsu on 4/11/24.
+//
+
+#ifndef in_app_purchase_storekit_h
+#define in_app_purchase_storekit_h
+
+#if __has_include("in_app_purchase_storekit-umbrella.h")
```
LouiseHsu, line 11: This has to be here b/c of this cocoapods issue -
https://github.com/CocoaPods/CocoaPods/issues/3767

Without this file, the generated "in_app_purchase_storekit-Swift.h" will keep trying to import an "in_app_purchase_storekit.h" which doesn't exist.

packages/in_app_purchase/in_app_purchase_storekit/pigeons/messages.dart
```diff
@@ -238,7 +238,7 @@ abstract class InAppPurchaseAPI {
   SKProductsResponseMessage startProductRequest(
       List<String> productIdentifiers);
 
-  void finishTransaction(Map<String, String?> finishMap);
+  void finishTransaction(Map<String, Object?> finishMap);
```
LouiseHsu, line 241: Even though the intended type here should be a nullable String, when pigeon generates obj C code, it is generated as type NSString. NSString is nullable in objective c, but when Xcode generates the corresponding swift method signature, it converts into a String, not a String?. NSString also cannot be annoted with _Nullable as it is a generic.

packages/in_app_purchase/in_app_purchase_storekit/example/shared/RunnerTests/Stubs.m
```diff
@@ -144,28 +144,6 @@ - (instancetype)initWithMap:(NSDictionary *)map {
 
 @end
 
-@interface InAppPurchasePluginStub ()
```
LouiseHsu, line 147: This plugin stub needed to be converted to swift as you cannot subclass a swift class in objective C.

hellohuanlin:
Could you run formatter first so it's easier for review?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+  {
+    return SKPaymentQueue.canMakePayments() as NSNumber
+  }
+
+  public func transactionsWithError(_ error: AutoreleasingUnsafeMutablePointer<FlutterError?>)
+    -> [SKPaymentTransactionMessage]?
+  {
+    let transactions = self.paymentQueueHandler?.getUnfinishedTransactions() ?? []
+    var transactionMaps: [SKPaymentTransactionMessage] = []
+    for transaction in transactions {
```
hellohuanlin, line 98: use `.map` or `.compactMap` to avoid the variable.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+    return transactionMaps
+  }
+
+  public func storefrontWithError(_ error: AutoreleasingUnsafeMutablePointer<FlutterError?>)
+    -> SKStorefrontMessage?
+  {
+    if #available(iOS 13.0, *) {
+      let storefront = self.paymentQueueHandler?.storefront
+      if storefront == nil {
+        return nil
```
hellohuanlin, line 112: Nit: `guard let storeFront = ... else { return nil }`

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+    }
+    return nil
+  }
+
+  public func startProductRequestProductIdentifiers(
+    _ productIdentifiers: [String],
+    completion: @escaping (SKProductsResponseMessage?, FlutterError?) -> Void
+  ) {
+    let request = getProductRequest(withIdentifiers: Set(productIdentifiers))
+    let handler = handlerFactory!(request)
```
hellohuanlin, line 124: will `handlerFactory` ever be nil?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+    requestHandlers.insert(handler)
+
+    handler.startProductRequest { response, startProductRequestError in
+      var error: FlutterError
+      if let startProductRequestError = startProductRequestError {
+        error = FlutterError(
```
hellohuanlin, line 130: if you move the `error` declaration inside this `if let`, you can use `let` and avoid `var`.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+          code: "storekit_platform_no_response",
+          message:
+            "Failed to get SKProductResponse in startRequest call. Error occurred on iOS platform",
+          details: productIdentifiers)
+        completion(nil, error)
+        return
+      }
+
+      response.products.forEach { product in
+        self.productsCache[product.productIdentifier] = product
```
hellohuanlin, line 149: `startProductRequest` should capture a `weak self`

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+            code: "storekit_invalid_payment_discount_object",
+            message:
+              "You have requested a payment and specified a payment discount with invalid properties. \(invalidError)",
+            details: paymentMap)
+          return
+        }
+      }
+    }
+
+    guard self.paymentQueueHandler?.add(payment) == true else {
```
hellohuanlin, line 206: no need `== true`

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+      error.pointee = FlutterError(
+        code: "storekit_duplicate_product_object",
+        message:
+          "There is a pending transaction for the same product identifier. Please either wait for it to be finished or finish it manually using `completePurchase` to avoid edge cases.",
+        details: paymentMap)
+      return
+    }
```
hellohuanlin, line 213: does it have any more logic after the guard? A guard statement is typically used as an early exit of a function

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+
+    let pendingTransactions = paymentQueueHandler!.getUnfinishedTransactions()
+
+    for transaction in pendingTransactions {
+      // Check if the current transaction's identifier matches the provided one,
+      // or both identifiers are nil and the product identifiers match.
+      if transaction.transactionIdentifier == transactionIdentifier
+        || (transactionIdentifier == NSNull().description
```
hellohuanlin, line 231: why is it comparing the `description`? `description` is typically used for logging or displaying, and isn't good for identification

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+    #if os(iOS)
+      paymentQueueHandler!.presentCodeRedemptionSheet()
+    #endif
+  }
+
+  public func retrieveReceiptDataWithError(
+    _ error: AutoreleasingUnsafeMutablePointer<FlutterError?>
+  ) -> String? {
+    var flutterError: FlutterError? = nil
+    let receiptData: String? = receiptManager!.retrieveReceiptWithError(&flutterError)
```
hellohuanlin, line 258: will `receiptManager` be nil?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+  public func refreshReceiptReceiptProperties(
+    _ receiptProperties: [String: Any]?, completion: @escaping (FlutterError?) -> Void
+  ) {
+    var request: SKReceiptRefreshRequest
+    if let receiptProperties = receiptProperties {
+      // If receiptProperties is not nil, this call is for testing.
+      var properties: [String: Any] = [:]
+      properties[SKReceiptPropertyIsExpired] = receiptProperties["isExpired"]!
+      properties[SKReceiptPropertyIsRevoked] = receiptProperties["isRevoked"]!
+      properties[SKReceiptPropertyIsVolumePurchase] = receiptProperties["isVolumePurchase"]!
```
hellohuanlin, line 275: nit: can just create dictionary literals 
```
let properties = {
  key1: value1,
  key2: value2,
}
```

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+    #if os(iOS)
+      if #available(iOS 13.4, *) {
+        paymentQueueHandler!.showPriceConsentIfNeeded()
+      }
+    #endif
+  }
+
+  public func handleTransactionsUpdated(_ transactions: [SKPaymentTransaction]) {
+    var maps: [[AnyHashable: Any]] = []
+    for transaction in transactions {
```
hellohuanlin, line 350: can you a `map` function

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+      let map = FIAObjectTranslator.getMapFrom(transaction)
+      maps.append(map)
+    }
+    transactionObserverCallbackChannel!.invokeMethod("updatedTransactions", arguments: maps)
+  }
+
+  public func handleTransactionsRemoved(_ transactions: [SKPaymentTransaction]) {
+    var maps: [[AnyHashable: Any]] = []
+    for transaction in transactions {
```
hellohuanlin, line 359: same here.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+  func getProductRequest(withIdentifiers productIdentifiers: Set<String>) -> SKProductsRequest {
+    return SKProductsRequest(productIdentifiers: productIdentifiers)
+  }
+
+  func getNonNullValue(from dictionary: [String: Any], forKey key: String) -> Any? {
+    let value = dictionary[key]
+    return value is NSNull ? nil : value
+  }
+
+  func getRefreshReceiptRequest(properties: [String: Any]?) -> SKReceiptRefreshRequest {
```
hellohuanlin, line 409: are these helpers only used inside this file? should use `private` if that's the case.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/in_app_purchase_storekit.h
hellohuanlin: can you add a comment there?

packages/in_app_purchase/in_app_purchase_storekit/example/ios/RunnerTests/RunnerTests-Bridging-Header.h
```diff
@@ -0,0 +1,7 @@
+//
+//  Use this file to import your target's public headers that you would like to
+//  expose to Swift.
+//
+
+#import "Stubs.h"
+#import "Stubs.m"
```
hellohuanlin, line 7: why is it importing the .m file?

packages/in_app_purchase/in_app_purchase_storekit/pigeons/messages.dart
hellohuanlin: im confused. why is NSString generic?

hellohuanlin (commented):
I briefly looked at the code and did not look into the details. I have to stop early as it's getting long, and there are enough feedback already.

packages/in_app_purchase/in_app_purchase_storekit/pigeons/messages.dart
LouiseHsu: https://github.com/flutter/packages/pull/6561#discussion_r1571294947

jmagman:
```
SwiftGeneratePch normal arm64 Compiling\ bridging\ header (in target 'RunnerTests' from project 'Runner')
    cd /Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos
    builtin-swiftTaskExecution -- /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-frontend -frontend -target arm64-apple-macos10.15 -Xllvm -aarch64-use-tbi -enable-objc-interop -stack-check -sdk /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk -I /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug -I /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/lib -F /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug -F /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/in_app_purchase_storekit -F /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/shared_preferences_foundation -F /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/Frameworks -F /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/OCMock -F /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/in_app_purchase_storekit -F /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/shared_preferences_foundation -no-color-diagnostics -enable-testing -g -module-cache-path /Users/chrome-bot/Library/Developer/Xcode/DerivedData/ModuleCache.noindex -swift-version 5 -enforce-exclusivity\=checked -Onone -D DEBUG -D COCOAPODS -D COCOAPODS -serialize-debugging-options -const-gather-protocols-file /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/Objects-normal/arm64/RunnerTests_const_extract_protocols.json -enable-bare-slash-regex -empty-abi-descriptor -validate-clang-modules-once -clang-build-session-file /Users/chrome-bot/Library/Developer/Xcode/DerivedData/ModuleCache.noindex/Session.modulevalidation -Xcc -working-directory -Xcc /Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos -resource-dir /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift -enable-anonymous-context-mangled-names -Xcc -fmodule-map-file\=/Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos/Pods/Headers/Public/in_app_purchase_storekit/in_app_purchase_storekit.modulemap -Xcc -fmodule-map-file\=/Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos/Pods/Headers/Public/integration_test/integration_test.modulemap -Xcc -fmodule-map-file\=/Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos/Pods/Headers/Public/shared_preferences_macos/shared_preferences_macos.modulemap -Xcc -ivfsstatcache -Xcc /Users/chrome-bot/Library/Developer/Xcode/DerivedData/SDKStatCaches.noindex/macosx14.0-23A334-edcf116c3e76c79d4aee70e3f10029bc.sdkstatcache -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/swift-overrides.hmap -Xcc -iquote -Xcc /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/RunnerTests-generated-files.hmap -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/RunnerTests-own-target-headers.hmap -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/RunnerTests-all-non-framework-target-headers.hmap -Xcc -ivfsoverlay -Xcc /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/all-product-headers.yaml -Xcc -iquote -Xcc /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/RunnerTests-project-headers.hmap -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/include -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/in_app_purchase_storekit/in_app_purchase_storekit.framework/Headers -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/shared_preferences_foundation/shared_preferences_foundation.framework/Headers -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/OCMock/OCMock.framework/Headers -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/in_app_purchase_storekit/in_app_purchase_storekit.framework/Headers -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Products/Debug/shared_preferences_foundation/shared_preferences_foundation.framework/Headers -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/DerivedSources-normal/arm64 -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/DerivedSources/arm64 -Xcc -I/Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/Runner.build/Debug/RunnerTests.build/DerivedSources -Xcc -DDEBUG\=1 -Xcc -DCOCOAPODS\=1 -Xcc -DCOCOAPODS\=1 -import-objc-header /Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos/RunnerTests/RunnerTests-Bridging-Header.h -module-name RunnerTests -disable-clang-spi -target-sdk-version 14.0 -target-sdk-name macosx14.0 -external-plugin-path /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk/usr/lib/swift/host/plugins\#/Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk/usr/bin/swift-plugin-server -external-plugin-path /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk/usr/local/lib/swift/host/plugins\#/Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk/usr/bin/swift-plugin-server -external-plugin-path /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/lib/swift/host/plugins\#/Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/bin/swift-plugin-server -external-plugin-path /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/local/lib/swift/host/plugins\#/Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/bin/swift-plugin-server -plugin-path /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/host/plugins -plugin-path /Volumes/Work/s/w/ir/cache/osx_sdk/xcode_15a240d/XCode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/local/lib/swift/host/plugins -index-store-path /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Index.noindex/DataStore -serialize-diagnostics-path /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/PrecompiledHeaders/RunnerTests-Bridging-Header-dmi8xq9h13gd.dia /Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos/RunnerTests/RunnerTests-Bridging-Header.h -index-store-path /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Index.noindex/DataStore -emit-pch -pch-output-dir /Users/chrome-bot/Library/Developer/Xcode/DerivedData/Runner-cqhwqvjiontyvddtpbfbwdhoqskw/Build/Intermediates.noindex/PrecompiledHeaders
<unknown>:0: error: module map file '/Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos/Pods/Headers/Public/in_app_purchase_storekit/in_app_purchase_storekit.modulemap' not found
<unknown>:0: error: module map file '/Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos/Pods/Headers/Public/integration_test/integration_test.modulemap' not found
<unknown>:0: error: module map file '/Volumes/Work/s/w/ir/x/w/packages/packages/in_app_purchase/in_app_purchase_storekit/example/macos/Pods/Headers/Public/shared_preferences_macos/shared_preferences_macos.modulemap' not found
Command SwiftGeneratePch emitted errors but did not return a nonzero exit code to indicate failure
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8749646386009370081/+/u/Run_package_tests/native_test/stdout

I believe the `OTHER_SWIFT_FLAGS` in the macOS project is set up incorrectly:

https://github.com/flutter/packages/blob/26bd37b35fdfb1aea96e83e71224e07aeb477b1f/packages/in_app_purchase/in_app_purchase_storekit/example/macos/Runner.xcodeproj/project.pbxproj#L770


From the original commit of the project:
https://github.com/flutter/plugins/pull/6517/files#diff-0baa9bfc86ab566e4f94f744d8fa57d187570868e84706c04b21b8ced0a6fedaR769

packages/in_app_purchase/in_app_purchase_storekit/example/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -773,7 +773,6 @@
 				MARKETING_VERSION = 1.0;
 				MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE;
 				MTL_FAST_MATH = YES;
-				OTHER_SWIFT_FLAGS = "$(inherited) -D COCOAPODS -Xcc -fmodule-map-file=\"${PODS_ROOT}/Headers/Public/in_app_purchase_storekit/in_app_purchase_storekit.modulemap\" -Xcc -fmodule-map-file=\"${PODS_ROOT}/Headers/Public/integration_test/integration_test.modulemap\" -Xcc -fmodule-map-file=\"${PODS_ROOT}/Headers/Public/shared_preferences_macos/shared_preferences_macos.modulemap\"";
```
jmagman, line 770: There are two more of these I believe, one for each configuration.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+#endif
+
+extension FlutterError: Error {}
+@objcMembers
+public class InAppPurchasePlugin: NSObject, FlutterPlugin, InAppPurchaseAPI {
+  // Properties
+  private(set) var productsCache: NSMutableDictionary = [:]
+  private(set) var paymentQueueDelegateCallbackChannel: FlutterMethodChannel?
+  private(set) var receiptManager: FIAPReceiptManager?
+  private(set) var paymentQueueDelegate: Any? = nil
```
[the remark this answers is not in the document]
LouiseHsu: For tests - the objc test file can't seem to find them if theyre only private

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+  // note - the type should be FIAPPaymentQueueDelegate, but this is only available >= iOS 13,
+  private var requestHandlers = Set<FIAPRequestHandler>()
+  private var handlerFactory: ((SKRequest) -> FIAPRequestHandler)?
+  public var registrar: FlutterPluginRegistrar?
+  public var paymentQueueHandler: FIAPaymentQueueHandler?
+  public var transactionObserverCallbackChannel: FlutterMethodChannel?
```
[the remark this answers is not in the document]
LouiseHsu: > these shouldn't be public. public means it can be called from outside of this module.
For tests - the objc test file can't seem to find them if theyre only private

hellohuanlin:
is it ready for another review?

LouiseHsu:
> is it ready for another review?

@hellohuanlin Yeah, please take a look when you can!

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+    self.handlerFactory = handlerFactory
+  }
+
+  public convenience init(registrar: FlutterPluginRegistrar) {
+    self.init(receiptManager: FIAPReceiptManager())
+    self.registrar = registrar
+
+    self.paymentQueueHandler = FIAPaymentQueueHandler(
+      queue: SKPaymentQueue.default(),
+      transactionsUpdated: { [weak self] (transactions: [SKPaymentTransaction]) -> Void in
```
hellohuanlin, line 72: nit: 
```
      transactionsUpdated: { [weak self] transactions in
```
here and below

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,412 @@
...
+      },
+      restoreCompletedTransactionsFinished: { () -> Void in
+        weakSelf!.restoreCompletedTransactionsFinished()
+      },
+      shouldAddStorePayment: { (payment: SKPayment, product: SKProduct) -> Bool in
+        return weakSelf!.shouldAddStorePayment(payment: payment, product: product)
+      },
+      updatedDownloads: { _ in
+        weakSelf!.updatedDownloads()
+      }, transactionCache: FIATransactionCache())
```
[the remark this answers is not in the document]
hellohuanlin: this

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+  public func canMakePaymentsWithError(_ error: AutoreleasingUnsafeMutablePointer<FlutterError?>)
+    -> NSNumber?
+  {
+    return SKPaymentQueue.canMakePayments() as NSNumber
+  }
+
+  public func transactionsWithError(_ error: AutoreleasingUnsafeMutablePointer<FlutterError?>)
+    -> [SKPaymentTransactionMessage]?
+  {
+    let transactions = self.paymentQueueHandler!.getUnfinishedTransactions()
```
hellohuanlin, line 109: is `paymentQueueHandler` guaranteed to be present?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+    var transactionMaps: [SKPaymentTransactionMessage] = []
+    transactionMaps = transactions.compactMap {
+      FIAObjectTranslator.convertTransaction(toPigeon: $0)
+    }
+    return transactionMaps
```
hellohuanlin, line 114: nit: 
can just be 
```
return transactions.compactMap { ... }
```

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
hellohuanlin: can combine this guard and the variable declaration, so that it's not accessible outside of the guard

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+            code: "storekit_invalid_payment_discount_object",
+            message:
+              "You have requested a payment and specified a payment discount with invalid properties. \(invalidError)",
+            details: paymentMap)
+          return
+        }
+      }
+    }
+
+    if !self.paymentQueueHandler!.add(payment) {
```
hellohuanlin, line 216: is `paymentQueueHandler` guaranteed to be non-nil?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+      error.pointee = FlutterError(
+        code: "storekit_duplicate_product_object",
+        message:
+          "There is a pending transaction for the same product identifier. Please either wait for it to be finished or finish it manually using `completePurchase` to avoid edge cases.",
+        details: paymentMap)
+      return
```
hellohuanlin, line 222: no need this `return`

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+    }
+  }
+
+  public func finishTransactionFinishMap(
+    _ finishMap: [String: Any], error: AutoreleasingUnsafeMutablePointer<FlutterError?>
+  ) {
+    let transactionIdentifier = (finishMap["transactionIdentifier"] ?? "") as? String
```
hellohuanlin, line 229: move it below `guard let productIdentifier`

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+    for transaction in pendingTransactions {
+      // If the user cancels the purchase dialog we won't have a transactionIdentifier.
+      // So if it is null AND a transaction in the pendingTransactions list has
+      // also a null transactionIdentifier we check for equal product identifiers.
+      if transaction.transactionIdentifier == transactionIdentifier
+        || (transactionIdentifier == ""
+          && transaction.transactionIdentifier == nil
+          && transaction.payment.productIdentifier == productIdentifier)
+      {
+        paymentQueueHandler!.finish(transaction)
```
hellohuanlin, line 246: Writing `!` all over the code makes it pretty hard to reason about, and can easily be overlooked, defeating the purpose of null safety. If this is guaranteed to be non nil, it's more readable to move the cast to the front and explain why:  

```
guard let paymentQueueHandler else { 
  fatalError("paymentQueueHandler cant be nil because reason 1 2 3") 
}
```

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+      paymentQueueHandler!.presentCodeRedemptionSheet()
+    #endif
+  }
+
+  public func retrieveReceiptDataWithError(
+    _ error: AutoreleasingUnsafeMutablePointer<FlutterError?>
+  ) -> String? {
+    var flutterError: FlutterError? = nil
+    let receiptData: String? = receiptManager.retrieveReceiptWithError(&flutterError)
+    if receiptData == nil {
```
hellohuanlin, line 270:
```
if let receiptData = receiptManager.retrieveReceiptWithError(&flutterError)
```

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+      error.pointee = flutterError
+      return nil
+    }
+    return receiptData
+  }
+
+  public func refreshReceiptReceiptProperties(
+    _ receiptProperties: [String: Any]?, completion: @escaping (FlutterError?) -> Void
+  ) {
+    var request: SKReceiptRefreshRequest
```
hellohuanlin, line 280: avoid `var` if it doesn't chnage

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+      // If receiptProperties is not nil, this call is for testing.
+      let properties = [
+        SKReceiptPropertyIsExpired: receiptProperties["isExpired"]!,
+        SKReceiptPropertyIsRevoked: receiptProperties["isRevoked"]!,
+        SKReceiptPropertyIsVolumePurchase: receiptProperties["isVolumePurchase"]!,
+      ]
+      request = getRefreshReceiptRequest(properties: properties)
+    } else {
+      request = getRefreshReceiptRequest(properties: nil)
+    }
```
hellohuanlin, line 291: this if/else can be
```
let properties = receiptProperties.map { ... }
let request = getRefreshReceiptRequest(properties: properties)
```

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+    #if os(iOS)
+      if #available(iOS 13.0, *) {
+        let messenger = registrar?.messenger()
+        paymentQueueDelegateCallbackChannel = FlutterMethodChannel(
+          name: "plugins.flutter.io/in_app_purchase_payment_queue_delegate",
+          binaryMessenger: messenger!)
+
+        paymentQueueDelegate = FIAPPaymentQueueDelegate(
+          methodChannel: paymentQueueDelegateCallbackChannel!)
+        paymentQueueHandler!.delegate = (paymentQueueDelegate as! any SKPaymentQueueDelegate)
```
hellohuanlin, line 333: iirc `any` keyword is quite recent. probably a few years after ios 13. can you double check?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+        paymentQueueHandler!.showPriceConsentIfNeeded()
+      }
+    #endif
+  }
+
+  @objc
+  public func handleTransactionsUpdated(_ transactions: [SKPaymentTransaction]) {
+    let maps = transactions.map { transaction in
+      FIAObjectTranslator.getMapFrom(transaction)
+    }
```
hellohuanlin, line 364: can be just 
```
let maps = transactions.map { FIAObjectTranslator.getMapFrom($0) }
```

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+    transactionObserverCallbackChannel!.invokeMethod("updatedTransactions", arguments: maps)
+  }
+
+  @objc
+  public func handleTransactionsRemoved(_ transactions: [SKPaymentTransaction]) {
+    let maps = transactions.map { transaction in
+      FIAObjectTranslator.getMapFrom(transaction)
+    }
```
hellohuanlin, line 372: same

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
hellohuanlin: this

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+  @objc
+  public convenience init(
+    receiptManager: FIAPReceiptManager,
+    handlerFactory: @escaping (SKRequest) -> FIAPRequestHandler
+  ) {
+    self.init(receiptManager: receiptManager)
+    self.handlerFactory = handlerFactory
+  }
+
+  public convenience init(registrar: FlutterPluginRegistrar) {
```
hellohuanlin, line 66: why do we have so many `init`? can you add a comment for each `init`

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/messages.g.h
```diff
@@ -1,7 +1,7 @@
 // Copyright 2013 The Flutter Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
-// Autogenerated from Pigeon (v16.0.4), do not edit directly.
+// Autogenerated from Pigeon (v16.0.5), do not edit directly.
```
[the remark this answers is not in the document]
LouiseHsu: no, it just gets regenerated. is it worth going down a version to avoid these comment changes?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,425 @@
...
+  import FlutterMacOS
+#endif
+
+public class InAppPurchasePlugin: NSObject, FlutterPlugin, InAppPurchaseAPI {
+  // Properties
+  private let receiptManager: FIAPReceiptManager
+  private var productsCache: NSMutableDictionary = [:]
+  private var paymentQueueDelegateCallbackChannel: FlutterMethodChannel?
+  private var paymentQueueDelegate: Any? = nil
+  // note - the type should be FIAPPaymentQueueDelegate, but this is only available >= iOS 13,
```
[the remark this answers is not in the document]
LouiseHsu: Nothing, I imagine. This is only used in registerPaymentQueueDelegateWithError or removePaymentQueueDelegateWithError, which both are ios13+ only

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/messages.g.h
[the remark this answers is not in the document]
hellohuanlin: no need to regenerate. just don't check-in the changes that's not related to this PR.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
LouiseHsu: Technically yes? The other two initializers that dont set paymentQueueHandler are either marked unavailable or available in tests only in the objc version. I just looked it up and it doesnt look like theres a nice swift version for this

LouiseHsu: nevermind, i saw your suggestion further below

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
LouiseHsu: these need to be not private so they can be overriden for testing purposes
Like here:
https://github.com/flutter/packages/blob/74de10af17274ab3d93a264b215ba19ec20e7ba6/packages/in_app_purchase/in_app_purchase_storekit/example/ios/RunnerTests/SwiftStubs.swift#L11

packages/in_app_purchase/in_app_purchase_storekit/pigeons/messages.dart
LouiseHsu: Sorry, its not generic, but you can't specify a type inside a NSDictionary to be nullable. (i dont think)
Specifically here -
https://github.com/flutter/packages/blob/74de10af17274ab3d93a264b215ba19ec20e7ba6/packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/messages.g.h#L268
What I really want is for this line in swift 
https://github.com/flutter/packages/blob/74de10af17274ab3d93a264b215ba19ec20e7ba6/packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift#L227
To be `finishMap: [String: String?]`. But the automatic conversion turns all NSStrings as `String` and not `String?`. And because I can't specify the NSString as extra nullable, I have to make it `Object?` (in pigeon) so it gets generated into `id` in objc, which then will turn into `Any`, and thus allow for nullability
Not sure if that makes sense lol

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
LouiseHsu: turns out i can just delete it lol

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+import StoreKit
+
+#if os(iOS)
+  import Flutter
+#elseif os(macOS)
+  import FlutterMacOS
+#endif
+
+public class InAppPurchasePlugin: NSObject, FlutterPlugin, InAppPurchaseAPI {
+  // Properties
```
hellohuanlin, line 15: no need this comment

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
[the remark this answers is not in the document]
hellohuanlin: > This is only gets set/used in registerPaymentQueueDelegateWithError or removePaymentQueueDelegateWithError, which both are ios13+ only

Can you put this in the comment? 

Also comments should be before the code that it's commenting, not after.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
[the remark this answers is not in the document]
hellohuanlin: maybe im missing something here - it's marked as resolved, but the code is not updated.

hellohuanlin (commented):
I started the review, but unless im mistaken, it looks like a few previous comments have not been addressed. can you make sure all previous comments are addressed before i continue?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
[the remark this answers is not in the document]
LouiseHsu: wtf i swear i added a comment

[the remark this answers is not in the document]
LouiseHsu: Is this not the comment? (https://github.com/flutter/packages/blob/b300519b60588d257abd6a655172d82e999875e3/packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift#L29)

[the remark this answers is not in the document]
hellohuanlin: Can you put the comment before the code?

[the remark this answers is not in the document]
hellohuanlin: Discussed offline - I didn't see the comment previously because the comment is after the code, so it looks like a comment for the `register` function

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+  }
+
+  public func transactionsWithError(_ error: AutoreleasingUnsafeMutablePointer<FlutterError?>)
+    -> [SKPaymentTransactionMessage]?
+  {
+
+    let transactions = getPaymentQueueHandler().getUnfinishedTransactions()
+    return transactions.compactMap {
+      FIAObjectTranslator.convertTransaction(toPigeon: $0)
+    }
```
hellohuanlin, line 116: nit: can just be 
```
return getPaymentQueueHandler()
  .getUnfinishedTransactions()
  .compactMap {
    ...
  }
```

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+  }
+
+  public func storefrontWithError(_ error: AutoreleasingUnsafeMutablePointer<FlutterError?>)
```
hellohuanlin, line 119: are these pigeon functions? can you either use `MARK` to denote the section of pigeon functions, or move pigeon protocol conformance to a class extension.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+
+  public func startProductRequestProductIdentifiers(
+    _ productIdentifiers: [String],
+    completion: @escaping (SKProductsResponseMessage?, FlutterError?) -> Void
+  ) {
+    let request = getProductRequest(withIdentifiers: Set(productIdentifiers))
+    let handler = handlerFactory(request)
+    requestHandlers.insert(handler)
+
+    handler.startProductRequest { [weak self] response, startProductRequestError in
```
hellohuanlin, line 136: this is a pretty long closure. can you check `weak self` first as the first line? In case of some race condition during the execution of this block

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+  ) {
+    getPaymentQueueHandler().stopObservingPaymentQueue()
+  }
+
+  public func registerPaymentQueueDelegateWithError(
+    _ error: AutoreleasingUnsafeMutablePointer<FlutterError?>
+  ) {
+    #if os(iOS)
+      if #available(iOS 13.0, *) {
+        let messenger = registrar?.messenger()
```
hellohuanlin, line 312: nit: `guard let messenger = registrar?.messenger()`. 
It's better to deal with nullability up front, so that it's non-null for the rest of code

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+  public func handleTransactionsRemoved(_ transactions: [SKPaymentTransaction]) {
+    let maps = transactions.map {
+      FIAObjectTranslator.getMapFrom($0)
+    }
+    transactionObserverCallbackChannel!.invokeMethod("removedTransactions", arguments: maps)
+  }
+
+  @objc
+  public func handleTransactionRestoreFailed(_ error: NSError) {
+    transactionObserverCallbackChannel!.invokeMethod(
```
hellohuanlin, line 364: can this be `nil`? can you go over all usage of `!` and make sure either remove it, or document why it's safe to force unwrap it. `!` should be very rarely used, and heavily documented if any.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+
+  func getProductRequest(withIdentifiers productIdentifiers: Set<String>) -> SKProductsRequest {
+    return SKProductsRequest(productIdentifiers: productIdentifiers)
+  }
+
+  func getRefreshReceiptRequest(properties: [String: Any]?) -> SKReceiptRefreshRequest {
+    return SKReceiptRefreshRequest(receiptProperties: properties)
+  }
+
+  // Private convenience methods
```
hellohuanlin, line 408: use `// MARK: - ` to separate logical sections of code, so it shows up in xcode

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
LouiseHsu: Like check if `weak self` exists?

hellohuanlin: you can do `guard let self = else else { return }`

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+    self.requestHandlers = Set<FIAPRequestHandler>()
+    self.handlerFactory = {
+      FIAPRequestHandler(request: $0)
+    }
+    super.init()
+  }
+
+  @objc
+  // TODO(louisehsu): this initializer is for tests only. Make this private after the tests are moved to swift.
+  public convenience init(
```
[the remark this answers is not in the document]
LouiseHsu: Interesting! Okay, i changed the designated initizalizers, but I cant make it into one init, since when it gets bridged to objc it forces the default value to be a normal param.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+    receiptManager: FIAPReceiptManager,
+    handlerFactory: @escaping (SKRequest) -> FIAPRequestHandler
+  ) {
+    self.init(receiptManager: receiptManager)
+    self.handlerFactory = handlerFactory
+  }
+
+  // This is the designated initializer
```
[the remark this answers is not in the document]
LouiseHsu: No this is the init that should actually get called - in the objc version this was marked with NS_DESIGNATED_INITIALIZER, while the current default init was marked as NS_UNAVAILABLE. Not sure what's a nicer way of doing this?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
LouiseHsu: Is using `?` preferable to `!` in most cases? or is it better to unwrap everytime?

hellohuanlin: > or is it better to unwrap everytime

both `?` and `!` are ways of "unwrapping" (since `Optional` type is just an enum wrapper). `!` will crash if it's nil, so we have to be extra careful, and we should avoid it if possible. 

For example, if `transactionObserverCallbackChannel` can be declared as non-optional, do that. Otherwise, add a comment explaining why it has to be an optional, and why we have to force unwrap it here.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,422 @@
...
+  // note - the type should be FIAPPaymentQueueDelegate, but this is only available >= iOS 13,
+  // FIAPPaymentQueueDelegateonly gets set/used in registerPaymentQueueDelegateWithError or removePaymentQueueDelegateWithError, which both are ios13+ only
+  private var paymentQueueDelegate: Any? = nil
+  private var requestHandlers = Set<FIAPRequestHandler>()
+  private var handlerFactory: ((SKRequest) -> FIAPRequestHandler)
+  // TODO(louisehsu): Once tests are migrated to swift, we can use @testable import, and make theses vars private again
+  @objc
+  public var registrar: FlutterPluginRegistrar?
+  @objc
+  public var paymentQueueHandler: FIAPaymentQueueHandler?
```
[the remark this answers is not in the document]
hellohuanlin: Can you try again after you make the other init as the designated init (as discussed in the other comment)?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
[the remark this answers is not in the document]
hellohuanlin: > in the objc version this was marked with NS_DESIGNATED_INITIALIZER

The objc version says the other one that takes receipt manager is the designated one tho: 

```
- (instancetype)initWithReceiptManager:(FIAPReceiptManager *)receiptManager
    NS_DESIGNATED_INITIALIZER;
```

Or am i missing something?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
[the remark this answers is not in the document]
hellohuanlin: You may also want to take `paymentQueueHandler` as another param for the designated init. We want other `init`s to call this designated `init`, so that `paymentQueueHandler` can be a `let` rather than a `var`. 

```
let p1: Int
let p2: Int
let p3: Int?
init(p1: Int, p2: Int, p3: Int?) {}

convenience init(p1: Int, p2: Int) {
  self.init(p1: p1, p2: p2, p3: nil)
}

convenience init(p1: Int, p3: Int?) {
  self.init(p1: p1, p2: 123, p3: p3)
}
```

This kind of setup avoids haven't to use `var` in fields that don't actually change.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
LouiseHsu: Okay, I left a comment for transactionObserverCallbackChannel and why it has to be optional

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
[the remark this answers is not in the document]
LouiseHsu: discussed offline - paymentQueueHandler cannot be a let var as it requires self to be initialized

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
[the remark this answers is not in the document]
LouiseHsu: You're right. Left some comments for the inits lol

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+#elseif os(macOS)
+  import FlutterMacOS
+#endif
+
+public class InAppPurchasePlugin: NSObject, FlutterPlugin, InAppPurchaseAPI {
+  private let receiptManager: FIAPReceiptManager
+  private var productsCache: NSMutableDictionary = [:]
+  private var paymentQueueDelegateCallbackChannel: FlutterMethodChannel?
+  // note - the type should be FIAPPaymentQueueDelegate, but this is only available >= iOS 13,
+  // FIAPPaymentQueueDelegateonly gets set/used in registerPaymentQueueDelegateWithError or removePaymentQueueDelegateWithError, which both are ios13+ only
```
hellohuanlin, line 19: * only* (space)

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+  private var requestHandlers = Set<FIAPRequestHandler>()
+  private var handlerFactory: ((SKRequest) -> FIAPRequestHandler)
+  // TODO(louisehsu): Once tests are migrated to swift, we can use @testable import, and make theses vars private again and remove all instances of @objc
+  @objc
+  public var registrar: FlutterPluginRegistrar?
+  @objc
+  // This property is optional, as it requires self to exist to be initialized.
+  public var paymentQueueHandler: FIAPaymentQueueHandler?
+  @objc
+  // This property is optional, as it is set when the plugin is registered.
```
hellohuanlin, line 30: this is set in the `init` tho.

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+      name: "plugins.flutter.io/in_app_purchase",
+      binaryMessenger: messenger)
+    let instance = InAppPurchasePlugin(registrar: registrar)
+    registrar.addMethodCallDelegate(instance, channel: channel)
+    registrar.addApplicationDelegate(instance)
+    SetUpInAppPurchaseAPI(messenger, instance)
+  }
+
+  @objc
+  public init(
```
hellohuanlin, line 50: can you add a comment on when to use this init?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+    receiptManager: FIAPReceiptManager,
+    handlerFactory: @escaping (SKRequest) -> FIAPRequestHandler = {
+      FIAPRequestHandler(request: $0)
+    }
+  ) {
+    self.receiptManager = receiptManager
+    self.requestHandlers = Set<FIAPRequestHandler>()
```
hellohuanlin, line 57: `requestHandlers` is already set when you declare the ivar: 

```
  private var requestHandlers = Set<FIAPRequestHandler>()
```

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+    self.handlerFactory = handlerFactory
+    super.init()
+  }
+
+  // This init is used for tests
+  @objc
+  public convenience init(receiptManager: FIAPReceiptManager) {
+    self.init(receiptManager: receiptManager, handlerFactory: { FIAPRequestHandler(request: $0) })
```
hellohuanlin, line 65: >   // This init is used for tests

Can we simply pass `FIAPRequestHandler(request: $0) }` in test?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  // This init gets called when the plugin gets registered
```
hellohuanlin, line 68: Nit: something like "used for plugin registration"

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+    #if os(iOS)
+      if #available(iOS 13.4, *) {
+        getPaymentQueueHandler().showPriceConsentIfNeeded()
+      }
+    #endif
+  }
+
+  @objc
+  public func handleTransactionsUpdated(_ transactions: [SKPaymentTransaction]) {
+    let maps = transactions.map {
```
hellohuanlin, line 358: nit: `let translatedTransactions = ...`

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+      FIAObjectTranslator.getMapFrom($0)
+    }
+    transactionObserverCallbackChannel?.invokeMethod("updatedTransactions", arguments: maps)
+  }
+
+  @objc
+  public func handleTransactionsRemoved(_ transactions: [SKPaymentTransaction]) {
+    let maps = transactions.map {
```
hellohuanlin, line 366: ditto

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+  public func restoreCompletedTransactionsFinished() {
+    transactionObserverCallbackChannel?.invokeMethod(
+      "paymentQueueRestoreCompletedTransactionsFinished", arguments: nil)
+  }
+
+  @objc
+  public func shouldAddStorePayment(payment: SKPayment, product: SKProduct) -> Bool {
+    productsCache[product.productIdentifier] = product
+    transactionObserverCallbackChannel?
+      .invokeMethod(
```
hellohuanlin, line 388: nit: you can reduce the indentation like
```
transactionObserverCallbackChannel?.invokeMethod(
  "shouldAddStorePayment",
  arguments: ...)
```

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,428 @@
...
+          "product": FIAObjectTranslator.getMapFrom(product),
+        ])
+    return false
+  }
+
+  public func updatedDownloads() {
+    NSLog("Received an updatedDownloads callback, but downloads are not supported.")
+  }
+
+  // MARK: - Methods to allow for testing
```
hellohuanlin, line 401: nit: Methods exposed for testing

hellohuanlin (commented):
Getting very close

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
LouiseHsu: Oh, thats true. Separately, do you think there's merit in keeping this init, if this is the designated init from the objc version `initWithRecieptManager:`? It might be a breaking change if this is removed.

LouiseHsu: tho im not sure how common it is to be initializing plugins manually?

hellohuanlin: if it's only used in tests (as stated in the comment), then it's not gonna be a breaking change. I could be wrong, but I don't think this is used by developers. Can you double check?

LouiseHsu: i did a quick search and couldnt find anything - so i just deleted it lol

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,423 @@
...
+  // note - the type should be FIAPPaymentQueueDelegate, but this is only available >= iOS 13,
+  // FIAPPaymentQueueDelegate only gets set/used in registerPaymentQueueDelegateWithError or removePaymentQueueDelegateWithError, which both are ios13+ only
+  private var paymentQueueDelegate: Any?
+  private var requestHandlers = Set<FIAPRequestHandler>()
+  private var handlerFactory: ((SKRequest) -> FIAPRequestHandler)
+  // TODO(louisehsu): Once tests are migrated to swift, we can use @testable import, and make theses vars private again and remove all instances of @objc
+  @objc
+  public var registrar: FlutterPluginRegistrar?
+  @objc
+  // This property is optional, as it requires self to exist to be initialized.
```
hellohuanlin, line 27: can you move comment before `@objc`?

packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift
```diff
@@ -0,0 +1,423 @@
...
+      name: "plugins.flutter.io/in_app_purchase",
+      binaryMessenger: messenger)
+    let instance = InAppPurchasePlugin(registrar: registrar)
+    registrar.addMethodCallDelegate(instance, channel: channel)
+    registrar.addApplicationDelegate(instance)
+    SetUpInAppPurchaseAPI(messenger, instance)
+  }
+
+  @objc
+  // This is the designated initializer
```
hellohuanlin, line 50: An init without `convenience` is designated initializer, so no need to explain this. Can you comment when should this init be used instead?

packages/in_app_purchase/in_app_purchase_storekit/pigeons/messages.dart
hellohuanlin: Using `Any` to represent an optional can be error-prone. 
Why do you want `[String:String?]`? Is there any issue with using `[String:String]`?

LouiseHsu: because the key (in this case `transactionidentifier`) is never nil, but there are cases where the value will be nil. but that is still an acceptable form, since it signals a cancelled transaction

hellohuanlin: Can you point me to the place where it's used and where it distinguishes the cancelled transaction vs no transaction?

LouiseHsu: https://github.com/flutter/packages/blob/74de10af17274ab3d93a264b215ba19ec20e7ba6/packages/in_app_purchase/in_app_purchase_storekit/darwin/Classes/InAppPurchasePlugin.swift#L237

hellohuanlin: are you referring to the comments below the code?

hellohuanlin: Not sure how `pendingTransactions` is related to the dictionary type. Can you explain how using `[String:String]` type would fail to work here?

LouiseHsu: Yea I meant the comments
Im not understanding the question i think? finishMap should be `[String:String?]` because `finishMap["transactionIdentifier"]` should be allowed to be null.

LouiseHsu: If it's just `[String:String]` then `finishMap["transactionIdentifier"] ` wouldn't be allowed to be null?

LouiseHsu: https://github.com/flutter/packages/blob/871a24b977c5e774cf9fc6a8c9f03d3338313125/packages/in_app_purchase/in_app_purchase_storekit/lib/src/store_kit_wrappers/sk_payment_transaction_wrappers.dart#L218

You can see on the dart side that the finishMap arg it gets called with has [String:String?] as well - I genuinely do not think it would be correct to make it [String:String]. Once this gets converted to using Swift pigeon, I can change it from `String:Any` to `String:String?`, so this should be temporary

hellohuanlin: > finishMap should be [String:String?] because finishMap["transactionIdentifier"] should be allowed to be null.
> If it's just [String:String] then finishMap["transactionIdentifier"]  wouldn't be allowed to be null, right?

Not really. `dict[key]` always returns optional. If type of `finishMap` is `[String:String]`, then the type of `finishMap["transactionIdentifier"]` would be `String?`. 

If you use `[String:String?]` as the type, then `finishMap["transactionIdentifier"]` would be `String??` (double optional).

hellohuanlin (approved):
LGTM! Thanks for making all the changes! Please make sure to manually test out if the feature still works as expected.

PR #6568, opened 2024-04-18
[camera] Convert iOS Obj-C->Dart calls to Pigeon

stuartmorgan:
Converts all of the Obj-C -> Dart calls to Pigeon, using the new suffix-based Pigeon API instantiation feature.

This required decentralizing some threading code slightly: since method channel calls only involve one method (due to the lack of strong types), a wrapper that automatically did thread bouncing was feasible, but with Pigeon it's not since a wrapper would have to duplicate the entire API surface, and that's more work than just doing the dispatches at the call site.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera_avfoundation/ios/Classes/CameraProperties.h
```diff
@@ -30,49 +32,24 @@ extern AVCaptureFlashMode FLTGetAVCaptureFlashModeForFLTFlashMode(FLTFlashMode m
 
 #pragma mark - exposure mode
 
-/// Represents camera's exposure mode. Mirrors ExposureMode in camera.dart.
-typedef NS_ENUM(NSInteger, FLTExposureMode) {
```
stuartmorgan, line 34: These enums are no longer needed because we can just use the Pigeon-generated enums instead.

packages/camera/camera_avfoundation/ios/Classes/CameraProperties.h
```diff
@@ -30,49 +32,24 @@ extern AVCaptureFlashMode FLTGetAVCaptureFlashModeForFLTFlashMode(FLTFlashMode m
...
-/// Gets a string representation of exposure mode.
-/// @param mode exposure mode
-extern NSString *FLTGetStringForFLTExposureMode(FLTExposureMode mode);
-
-/// Gets FLTExposureMode from its string representation.
-/// @param mode a string representation of the FLTExposureMode.
-extern FLTExposureMode FLTGetFLTExposureModeForString(NSString *mode);
+/// Gets FCPPlatformExposureMode from its string representation.
+/// @param mode a string representation of the exposure mode.
+extern FCPPlatformExposureMode FCPGetExposureModeForString(NSString *mode);
```
stuartmorgan, line 37: These are temporary methods; there are Dart->host calls that send exposure and focus modes, so until I convert those in a follow-up we need to be able to convert method channel strings to the Pigeon enum that replaced the ones removed here.

packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
```diff
@@ -169,25 +174,10 @@ - (void)handleMethodCallAsync:(FlutterMethodCall *)call result:(FlutterResult)re
...
-               @"exposureMode" : FLTGetStringForFLTExposureMode([_camera exposureMode]),
-               @"focusMode" : FLTGetStringForFLTFocusMode([_camera focusMode]),
-               @"exposurePointSupported" :
-                   @([_camera.captureDevice isExposurePointOfInterestSupported]),
-               @"focusPointSupported" : @([_camera.captureDevice isFocusPointOfInterestSupported]),
-             }];
+      _camera.dartAPI = [[FCPCameraEventApi alloc]
+          initWithBinaryMessenger:_messenger
+             messageChannelSuffix:[NSString stringWithFormat:@"%ld", cameraId]];
+      [_camera reportInitializationState];
```
stuartmorgan, line 180: I moved the actual channel invocation into `FLTCam` because it seemed strange to have it here when a) the camera instance already has the channel, and b) all of the values here come from the camera instance.

packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
```diff
@@ -41,37 +41,34 @@ + (void)registerWithRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
...
-            binaryMessenger:_messenger];
-  _deviceEventMethodChannel =
-      [[FLTThreadSafeMethodChannel alloc] initWithMethodChannel:methodChannel];
-}
-
-- (void)detachFromEngineForRegistrar:(NSObject<FlutterPluginRegistrar> *)registrar {
-  [UIDevice.currentDevice endGeneratingDeviceOrientationNotifications];
-}
-
-- (void)startOrientationListener {
```
stuartmorgan, line 69: I inlined this as a drive-by cleanup because it was only being called from one place, and that was `init`, which shouldn't be calling methods.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -959,17 +979,7 @@ - (void)setFlashModeWithResult:(FlutterResult)result mode:(NSString *)modeStr {
...
-  if (mode == FLTExposureModeInvalid) {
-    result(FlutterErrorFromNSError([NSError
-        errorWithDomain:NSCocoaErrorDomain
-                   code:NSURLErrorUnknown
-               userInfo:@{
-                 NSLocalizedDescriptionKey :
-                     [NSString stringWithFormat:@"Unknown exposure mode %@", modeStr]
-               }]));
-    return;
-  }
```
stuartmorgan, line 972: I'm now preventing this error from being possible on the Dart side (same below).

packages/camera/camera_avfoundation/lib/src/avfoundation_camera.dart
```diff
@@ -552,6 +536,38 @@ class AVFoundationCamera extends CameraPlatform {
     return Texture(textureId: cameraId);
   }
 
+  String _serializeFocusMode(FocusMode mode) {
```
stuartmorgan, line 539: I added these instead of continuing to use the ones from the platform interface package (the versions without the `_`) because having serialization in the platform interface package and deserialization in the implementation package is an anti-pattern since it means changes in one package can break internal communication in another. We created a lot of it by accident during the initial process of federating plugins; Pigeon migration has been eliminating it, but doing it now meant I could remove all the native code that required an "invalid" state for these enums, since I can guarantee within this package that we won't get a state at the native layer that we don't handle.

(These methods will become enum->enum conversion methods instead of enum->string conversion when I finish the Pigeon migration for this package.)

packages/camera/camera_avfoundation/test/avfoundation_camera_test.dart
```diff
@@ -361,77 +343,27 @@ void main() {
...
-          MethodCall('initialized', event.toJson()), cameraId);
-
-      // Assert
-      expect(await streamQueue.next, event);
-
-      // Clean up
-      await streamQueue.cancel();
-    });
-
-    test('Should receive resolution changes', () async {
```
stuartmorgan, line 384: This and the next test are testing functionality that was dead code. The Dart method call handler handled several messages that are never sent by the native code. It doesn't look like Android uses them either, so I'm guessing it was code mistakenly left in during refactoring in code review of a feature.

packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
```diff
@@ -93,9 +90,17 @@ - (void)orientationChanged:(NSNotification *)note {
 }
 
 - (void)sendDeviceOrientation:(UIDeviceOrientation)orientation {
-  [_deviceEventMethodChannel
-      invokeMethod:@"orientation_changed"
-         arguments:@{@"orientation" : FLTGetStringForUIDeviceOrientation(orientation)}];
+  __weak typeof(self) weakSelf = self;
+  dispatch_async(dispatch_get_main_queue(), ^{
+    [weakSelf.globalEventAPI
```
hellohuanlin, line 95: I vaguely remember there was an effort to make method channel thread safe. what's the progress on that and do we still need to manually dispatch to main in plugin code?

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraOrientationTests.m
```diff
@@ -9,44 +9,87 @@
...
+#pragma mark -
+
 @interface CameraOrientationTests : XCTestCase
 @end
 
 @implementation CameraOrientationTests
 
+// Ensure that the given queue and then the main queue have both cycled, to wait for any pending
+// async events that may have been bounced between them.
+- (void)waitForRoundTripWithQueue:(dispatch_queue_t)queue {
```
hellohuanlin, line 42: nice

packages/camera/camera_avfoundation/ios/Classes/CameraProperties.m
```diff
@@ -36,51 +36,29 @@ AVCaptureFlashMode FLTGetAVCaptureFlashModeForFLTFlashMode(FLTFlashMode mode) {
...
-    case FLTFocusModeInvalid:
-      // This value should never actually be used.
-      return nil;
-  }
-  return nil;
-}
-
-FLTFocusMode FLTGetFLTFocusModeForString(NSString *mode) {
+FCPPlatformFocusMode FCPGetFocusModeForString(NSString *mode) {
   if ([mode isEqualToString:@"auto"]) {
```
hellohuanlin, line 54: will these hard coded strings be gone eventually after pigeon work is done?

hellohuanlin (approved):
Went through a thorough review. Looks pretty clean.

packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
stuartmorgan: The current state is that replies are thread-safe, but calls that start on the host side still have to be done from the main thread. I don't think there's active work on addressing that part (although I believe it's still something we do eventually want to do).

packages/camera/camera_avfoundation/ios/Classes/CameraProperties.m
stuartmorgan: Yes, this is just for the remaining non-Pigeon Dart->host calls.

PR #6579, opened 2024-04-19
[url_launcher][web] Fix Link issues when Semantics are enabled

htoor3:
DRAFT PR - will add tests shortly.

This PR fixes an issue where Links break in Semantics mode (clicking links does nothing) because the Link's child semantics obscure the platform view. 

The fix:
- Clean up the extraneous semantics nodes being generated by `PlatformViewLink` so we can consolidate all of the link information in a single place.  Also exclude it from the focus tree to fix the "double tabbing" issue.
- Pass the link role and URI to the child semantics so that we can wrap the node in an `<a>` tag to enable the context menu and provide link preview info.  This requires a separate engine PR.
- Click events never bubbled back up when clicking the Link's child semantics nodes (due to `stopPropagation` being called in the engine), so we have to handle click events in the `capture` phase.
- Handle link trigger logic during hit test registration (not just during browser events).  

For keydown events, things remain the same:
1. User hits enter
2. Framework receives + handles event, executes `followLink`, registers a hit test view id
3. Browser receives event and triggers link behavior using the registered `viewId`
4. Cleanup 

For click events, things are slightly different depending on if the node that's clicked is a semantic node or a platform view link.

Clicking platform view links will behave similarly to keydown events.

Clicking the generated semantic node of a link (specifically its children) will behave as follows:
1. User clicks button
2. Browser receives + handles event. No hit test id is registered yet, so we just cache the click event for later 
3. Framework executes `followLink`, registers a hit test view id.  Triggers link behavior using the registered `viewId`
4. Cleanup 

Fixes https://github.com/flutter/flutter/issues/143164

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/url_launcher/url_launcher_web/lib/src/link.dart
+++ b/packages/url_launcher/url_launcher_web/lib/src/link.dart
@@ -236,7 +236,14 @@ class LinkViewController extends PlatformViewController {
   /// to only trigger the link if a hit test was registered and remains valid at
   /// the time the click handler executes.
   static void _onGlobalClick(html.MouseEvent event) {
-    triggeredMouseEvent = event;
+    final html.Element? targetElement = event.target as html.Element?;
+
+    // We only want to cache the mouse event if its target is a link, so that 
+    // we can later prevent its default behavior and handle how it should 
+    // open the link ourselves.
+    if(isLinkElement(targetElement) || isSemanticLinkElement(targetElement)) {
+      triggeredMouseEvent = event;
+    }
 
     if (followLinkTriggered) {
       final int? viewId = getViewIdFromTarget(event) ?? _hitTestedViewId;
@@ -439,4 +446,25 @@ bool isLinkElement(html.Element? element) {
   return element != null &&
       element.tagName == 'A' &&
       element.hasProperty(linkViewIdProperty.toJS).toDart;
+}
+
+/// Checks if the given [element] is a semantic Link. This is necessary because
+/// we need a way to distinguish click event targets that are links.
+bool isSemanticLinkElement(html.Element? element) {
+  return element != null &&
+      element.tagName == 'FLT-SEMANTICS' &&
+      hasLinkAncestor(element);
+}
+
+/// Checks if the given [element] has a link ancestor.
+bool hasLinkAncestor(html.Element? element) {
+  while (element != null) {
+    if(element.tagName == 'A') {
+      return true;
+    }
+
+    element = element.parentElement;
+  }
+
+  return false;
 }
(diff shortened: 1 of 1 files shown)


packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -222,11 +262,26 @@ class LinkViewController extends PlatformViewController {
   /// `click` from the browser.
   static void registerHitTest(LinkViewController controller) {
     _hitTestedViewId = controller.viewId;
+    followLinkTriggered = true;
+
+    if (triggeredMouseEvent != null) {
```
mdebbar, line 267: We should do the same for keyboard events (we probably don't need a `triggeredKeyboardEvent`, but we need a boolean at least).

Here's how I would do it, but feel free to adjust/modify it to how you see fit:
```dart
// Set to true in registerHitTest()
static bool followLinkTriggered = false;

// Set to true in _onGlobalClick() and _onGlobalKeydown()
static bool domEventTriggered = false;

// Set only in _onGlobalClick()
static html.Event? triggeredEvent;
```

packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -200,19 +209,50 @@ class LinkViewController extends PlatformViewController {
...
-    _instances[viewId]?._onDomClick(event);
-    // After the DOM click event has been received, clean up the hit test state
-    // so we can start fresh on the next event.
-    unregisterHitTest();
+    final html.Element? targetElement = event.target as html.Element?;
+
+    // We only want to cache the mouse event if its target is a link, so that 
+    // we can later prevent its default behavior and handle how it should 
+    // open the link ourselves.
+    if(isLinkElement(targetElement) || isSemanticLinkElement(targetElement)) {
```
htoor3, line 244: @mdebbar I had to add this to fix the issue of caching mouse events that aren't relevant to us (since the global handler captures every mouse event).  

I added `isSemanticLinkElement` which just checks if the target element is a semantic node that's wrapped within an `<a>` tag.  Is this robust enough to handle the semantic link case? Or should we try to associate a platform view's `viewId` to its children's generated semantic nodes?

packages/url_launcher/url_launcher_web/lib/src/link.dart
htoor3: Based on my testing, the keyboard event path doesn't deviate from hit test -> global keydown.  Do you think this boolean is necessary because there would be a case where the browser keydown handler executes before the framework handles it?

I had to be pretty careful about this in the click handler because we were running into issues where the dom event trigger was being set to true for arbitrary clicks (e.g. clicking elsewhere on the app).  This led to issues where we would trigger the link twice if the event path went hit test -> global click

stuartmorgan:
From triage: what's the status of this PR? Are you still planning on adding tests?

mdebbar:
This PR is not necessary anymore. I'm working on a bigger fix that will make this PR obsolete.

PR #6600, opened 2024-04-23
[pigeon] Moves all codec logic to singular custom codec

tarrinneal:
[pigeon] Moves all codec logic to singular custom codec.

Also fixes a few small codec related bugs that have cropped up over time.

fixes https://github.com/flutter/flutter/issues/147454
fixes https://github.com/flutter/flutter/issues/147127
fixes https://github.com/flutter/flutter/issues/147587
fixes https://github.com/flutter/flutter/issues/148065


packages/pigeon/platform_tests/test_plugin/windows/pigeon/core_tests.gen.cpp
```diff
@@ -2578,15 +2761,8 @@ void HostIntegrationCoreApi::SetUp(flutter::BinaryMessenger* binary_messenger,
...
-                  encodable_an_enum_arg.IsNull()
-                      ? 0
-                      : encodable_an_enum_arg.LongValue();
-              const auto an_enum_arg =
-                  encodable_an_enum_arg.IsNull()
-                      ? std::nullopt
-                      : std::make_optional<AnEnum>(
-                            static_cast<AnEnum>(an_enum_arg_value));
+              const auto* an_enum_arg = &(std::any_cast<const AnEnum&>(
+                  std::get<CustomEncodableValue>(encodable_an_enum_arg)));
```
stuartmorgan, line 2765: The bad variant access error is coming from this pattern: if the arg's value is null, then `encodable_an_enum_arg` isn't a `CustomEncodableValue` at all, it's a monotype (i.e., `IsNull()` is true). So you need to check `IsNull()` before doing a std::get instead of after, or you need to use a `get_if<CustomEncodableValue>(&encodable_an_enum_arg)` and then if that's null, pass null, and if it's not not, do the any-cast. (The thing that tripped me up in my first attempt to write a fix for this is that you can't `any_cast` a `CustomEncodableValue*` because of how implicit conversions work, so you'll need to do some gymnastics with a value variable like before I think.)

stuartmorgan: This is kind of clunky, but the only way I can find to do it in a single statement involves a pointer to a reference that I can't prove to myself has a lifetime beyond that statement, and I'd rather the code be longer than have undefined behavior...

```cpp
AnEnum an_enum_arg_value;
const AnEnum* an_enum_arg = nullptr;
if (!encodable_an_enum_arg.IsNull()) {
  an_enum_arg_value = std::any_cast<AnEnum>(std::get<CustomEncodableValue>(encodable_an_enum_arg));
  an_enum_arg = &an_enum_arg_value;
}
...EchoNullableEnum(an_enum_arg);
```

packages/pigeon/platform_tests/test_plugin/android/src/test/kotlin/com/example/test_plugin/AllDatatypesTest.kt
```diff
@@ -46,13 +53,16 @@ internal class AllDatatypesTest : TestCase() {
...
     assertEquals(firstTypes.aNullableObject, secondTypes.aNullableObject)
+    assertEquals(firstTypes.list, secondTypes.list)
+    assertEquals(firstTypes.boolList, secondTypes.boolList)
+    assertEquals(firstTypes.doubleList, secondTypes.doubleList)
+    assertEquals(firstTypes.intList, secondTypes.intList)
+    assertEquals(firstTypes.stringList, secondTypes.stringList)
+    assertEquals(firstTypes.map, secondTypes.map)
   }
 
-  @Test
```
stuartmorgan, line 55: Why don't we want this annotation on tests?

packages/pigeon/test/cpp_generator_test.dart
```diff
@@ -1911,8 +1868,8 @@ void main() {
       sink,
       dartPackageName: DEFAULT_PACKAGE_NAME,
     );
-    final String code = sink.toString();
-    expect(code, contains(' : public flutter::StandardCodecSerializer'));
+    // final String code = sink.toString();
+    // expect(code, contains(' : public flutter::StandardCodecSerializer'));
```
stuartmorgan, line 1872: ?

stuartmorgan (changes requested):
This is great! It's awesome to see how much complexity and branching was removed from all over the generators.

packages/pigeon/lib/pigeon_lib.dart
```diff
@@ -2249,14 +2268,16 @@ ${_argParser.usage}''';
       options = options.merge(PigeonOptions(
           objcOptions: (options.objcOptions ?? const ObjcOptions()).merge(
               ObjcOptions(
-                  headerIncludePath: path.basename(options.objcHeaderOut!)))));
+                  headerIncludePath: options.objcOptions?.headerIncludePath ??
+                      path.basename(options.objcHeaderOut!)))));
```
vashworth, line 2272: Is there a unit test that confirms this resolves https://github.com/flutter/flutter/issues/147587?

When I played with it when I discovered the issue, I also had to change a few other things: https://github.com/flutter/flutter/issues/147587#issuecomment-2086304905

packages/pigeon/pigeons/core_tests.dart
```diff
@@ -38,15 +46,21 @@ class AllTypes {
...
+  // ignore: strict_raw_type, always_specify_types
+  List list;
+  List<String?> stringList;
+  List<int?> intList;
+  List<double?> doubleList;
+  List<bool?> boolList;
+
+  // Maps
+  // ignore: strict_raw_type, always_specify_types
+  Map map;
```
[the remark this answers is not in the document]
tarrinneal: It's actually the beginning of a much large set of collection types that will be added in a future pr that will all have this naming convention.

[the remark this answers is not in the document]
tarrinneal: The list types already are all like this

packages/pigeon/platform_tests/test_plugin/android/src/test/kotlin/com/example/test_plugin/AllDatatypesTest.kt
tarrinneal: The class level annotation covers it.

stuartmorgan: From what I see in JUnit docs, it looks like `extends TestCase` is a legacy mode, and the `@Test` annotation is the preferred way for newer code. Is the IDE giving different feedback?

packages/pigeon/lib/generator_tools.dart
```diff
@@ -494,39 +504,42 @@ Map<TypeDeclaration, List<int>> getReferencedTypes(
...
-      .toList();
-  sortedNames.sort();
-  int enumeration = _minimumCodecFieldKey;
+Iterable<EnumeratedType> getEnumeratedTypes(Root root) sync* {
   const int maxCustomClassesPerApi = 255 - _minimumCodecFieldKey;
-  if (sortedNames.length > maxCustomClassesPerApi) {
+  if (root.classes.length + root.enums.length > maxCustomClassesPerApi) {
     throw Exception(
-        "Pigeon doesn't support more than $maxCustomClassesPerApi referenced custom classes per API, try splitting up your APIs.");
+        "Pigeon doesn't currently support more than $maxCustomClassesPerApi referenced custom classes per file.");
```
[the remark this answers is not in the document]
tarrinneal: yeah, if I don't get the fix ready to merge with this pr fast enough, it'll have to.

packages/pigeon/lib/generator_tools.dart
```diff
@@ -685,3 +698,25 @@ class OutputFileOptions<T> {
...
+  }
+  if (symbol == ' ' || symbol == '_' || symbol == '-') {
+    final List<String> words = text.split(symbol);
+    for (final String word in words) {
+      className += word.substring(0, 1).toUpperCase() + word.substring(1);
+    }
+  } else {
+    className = text.substring(0, 1).toUpperCase() + text.substring(1);
+  }
+  return className;
```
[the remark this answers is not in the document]
tarrinneal: totally forgot to come back and simplify this, thanks

packages/pigeon/lib/pigeon_lib.dart
tarrinneal: Still need to finish that part of the pr actually

tarrinneal: > Is there a unit test that confirms this resolves [flutter/flutter#147587](https://github.com/flutter/flutter/issues/147587)?


There is, but it bypasses the part of the code that was breaking this. 

I'll add one.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 20.0.0
+
+* Moves all codec logic to single custom codec per file.
+* **Breaking Change** Limits the number of total custom types to 126.
```
stuartmorgan, line 4: Maybe mention that if clients hit this limit, the workaround for now would be to split their interface definition into multiple files.

packages/pigeon/example/app/windows/runner/messages.g.cpp
```diff
@@ -101,39 +102,54 @@ MessageData MessageData::FromEncodableList(const EncodableList& list) {
...
-    case 128:
+    case 129:
       return CustomEncodableValue(MessageData::FromEncodableList(
           std::get<EncodableList>(ReadValue(stream))));
+    case 130: {
+      const auto& encodable_enum_arg = ReadValue(stream);
+      const int64_t enum_arg_value =
+          encodable_enum_arg.IsNull() ? 0 : encodable_enum_arg.LongValue();
+      return encodable_enum_arg.IsNull()
+                 ? CustomEncodableValue(std::nullopt)
```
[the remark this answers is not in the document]
stuartmorgan: This was marked as resolved without changes or comment; am I wrong about the type?

stuartmorgan (approved):
A couple minor things still, but LGTM

packages/pigeon/example/app/windows/runner/messages.g.cpp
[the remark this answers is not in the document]
tarrinneal: End up with this error when I do EncodableValue https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8746390072007433569/+/u/Run_package_tests/custom_package_tests/stdout?format=raw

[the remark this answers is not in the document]
stuartmorgan: Sorry, I hadn't fully swapped the C++ APIs back in. It should be `EncodableValue()` for the null case, not `EncodableValue(std:nullopt)` (which I'm guessing is what gave you that error?). Using `EncodableValue()` in the null case works for me locally.

PR #6601, opened 2024-04-23
[camera] Finish converting iOS to Pigeon

stuartmorgan:
Converts all remaining Dart->host communication in the iOS implementation to use Pigeon. Given the boilerplate nature of many of the changes, it seemed easiest to just do the remaining calls all at once now that the structure is in place.

Some high-level notes:
- Many methods used to send the `cameraId` without it ever being used on the native side, so the Pigeon versions do not send them.
- `ThreadSafeTextureRegistry` is removed because I discovered that it was masking a bug, so was more trouble than it was worth (see inline comments in PR).
- A number of enums have been removed in favor of using the Pigeon-generated enums to pass data from the plugin class to `FLTCam`.
- In many cases where the completion callback (previously `result`) was being passed to `FLTCam` in a call, only to have it always just call `result(nil)`, that's now done in the plugin class since it's easier to reason about completions being called when they aren't passed around. (Long term we should consider moving almost all of the rest out, and using `FlutterError*` out params that the plugin class passes to `completion`, but that is more surgery than I wanted to do in this PR.)

Completes the iOS portion of https://github.com/flutter/flutter/issues/117905

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
```diff
@@ -148,274 +138,355 @@ - (void)availableCamerasWithCompletion:
...
-      [_camera setFocusPointWithResult:result x:x y:y];
-    } else if ([@"pausePreview" isEqualToString:call.method]) {
-      [_camera pausePreviewWithResult:result];
-    } else if ([@"resumePreview" isEqualToString:call.method]) {
-      [_camera resumePreviewWithResult:result];
-    } else if ([@"setDescriptionWhileRecording" isEqualToString:call.method]) {
-      [_camera setDescriptionWhileRecording:(call.arguments[@"cameraName"]) result:result];
-    } else if ([@"setImageFileFormat" isEqualToString:call.method]) {
-      NSString *fileFormat = call.arguments[@"fileFormat"];
-      [_camera setImageFileFormat:FCPGetFileFormatFromString(fileFormat)];
```
stuartmorgan, line 258: This was never calling `result` (and it doesn't pass it in to the implementation), so that wasn't good :|

packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
```diff
@@ -148,274 +138,355 @@ - (void)availableCamerasWithCompletion:
...
+                                           (nonnull void (^)(FlutterError *_Nullable))completion {
+  __weak typeof(self) weakSelf = self;
+  dispatch_async(self.captureSessionQueue, ^{
+    [weakSelf.camera setDescriptionWhileRecording:cameraName withCompletion:completion];
+  });
+}
+
+- (void)disposeCamera:(NSInteger)cameraId
+           completion:(nonnull void (^)(FlutterError *_Nullable))completion {
+  [_registry unregisterTexture:cameraId];
```
stuartmorgan, line 415: This is what made me remove the thread abstraction around the registry—this was one of two meaningful uses of the wrapper, and it was wrong. The old code was calling this in the async handler, which means it was effectively in the `captureSessionQueue` dispatch block below. But it was internally moving the call back to the main thread, which while correct, *was not awaited* before continuing. That meant that unregistering the texture—which ensures that the engine won't keep asking for textures—was not reliably being done *before* tearing down the camera on the session queue.

It may have worked out later due to the way thread redispatch is done when actually doing the copy, but I didn't try to unwind that since it's clearly better to do in this order.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.h
```diff
@@ -54,26 +51,27 @@ NS_ASSUME_NONNULL_BEGIN
...
-- (void)setFocusModeWithResult:(FlutterResult)result mode:(NSString *)modeStr;
+- (void)startVideoRecordingWithCompletion:(void (^)(FlutterError *_Nullable))completion
+                    messengerForStreaming:(nullable NSObject<FlutterBinaryMessenger> *)messenger;
+- (void)stopVideoRecordingWithCompletion:(void (^)(NSString *_Nullable,
+                                                   FlutterError *_Nullable))completion;
+- (void)pauseVideoRecording;
+- (void)resumeVideoRecording;
+- (void)lockCaptureOrientation:(FCPPlatformDeviceOrientation)orientation;
+- (void)unlockCaptureOrientation;
+- (void)setFlashMode:(FCPPlatformFlashMode)mode
```
stuartmorgan, line 71: I fixed a bunch of weird/unidiomatic naming while touching these signatures.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.h
```diff
@@ -95,17 +93,26 @@ NS_ASSUME_NONNULL_BEGIN
...
-- (void)setFocusPointWithResult:(FlutterResult)result x:(double)x y:(double)y;
-- (void)setExposureOffsetWithResult:(FlutterResult)result offset:(double)offset;
+- (void)pausePreview;
+- (void)resumePreview;
+- (void)setDescriptionWhileRecording:(NSString *)cameraName
+                      withCompletion:(void (^)(FlutterError *_Nullable))completion;
+
+/// Sets the exposure point, in a (0,1) coordinate system.
+///
+/// If @c point is nil, the exposure point will reset to the center.
```
stuartmorgan, line 103: This logic was in CameraPlugin, but I moved it into this class now that it was easy to pass nil through using the new class.

packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
```diff
@@ -148,274 +138,355 @@ - (void)availableCamerasWithCompletion:
...
-      if (!reset) {
-        x = ((NSNumber *)call.arguments[@"x"]).doubleValue;
-        y = ((NSNumber *)call.arguments[@"y"]).doubleValue;
-      }
-      [_camera setExposurePointWithResult:result x:x y:y];
-    } else if ([@"getMinExposureOffset" isEqualToString:call.method]) {
-      result(@(_camera.captureDevice.minExposureTargetBias));
-    } else if ([@"getMaxExposureOffset" isEqualToString:call.method]) {
-      result(@(_camera.captureDevice.maxExposureTargetBias));
-    } else if ([@"getExposureOffsetStepSize" isEqualToString:call.method]) {
```
stuartmorgan, line 230: This method does not exist in the Pigeon version because hard-coding 0 as the return value is trivial to do in Dart.

packages/camera/camera_avfoundation/ios/Classes/CameraPlugin.m
hellohuanlin: good call!

hellohuanlin (approved):
I went through a pretty thorough review and everything looks very clean

PR #3234, opened 2023-02-21
[pigeon] Standardize host api error handling

xegrox:
* [java] Adds a `GeneratedApi.FlutterError` exception for passing custom error parameters (code, message, details).
* [kotlin] Adds a `FlutterError` exception for passing custom error parameters (code, message, details).
* [kotlin] Adds a `errorClassName` option in `KotlinOptions` for custom error class names.
* [java] Removes legacy try catch in async apis.
* Adds `FlutterError` handling integration tests for all platforms.

Fixes flutter/flutter#120861

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[...]

PR #6611, opened 2024-04-25
[pigeon][swift] Removes FlutterError in favor of PigeonError

bc-lee:
Swift has more strict type system than Objective-C. For example, protocol conformance must not be redundant[1]. Also, Both FlutterError and Swift.Error are public, extension `FlutterError` to `Swift.Error` is also public. If someone create a such extension in the plugin code, Flutter app can't create a such extension in the app code, which forces Plugin developers to use Objective-C when they want to use pigeon, instead of Swift.

To avoid this issue, this change makes pigeon to use another error type, named `PigeonError`, instead of `FlutterError`. By declaring `PigeonError` as internal visibility, their existence won't make compilation error even if `PigeonError` is declared in the app code.

Fixes https://github.com/flutter/flutter/issues/147371.
See also https://github.com/flutter/flutter/issues/137057#issuecomment-1775835986.

[1] https://docs.swift.org/swift-book/documentation/the-swift-programming-language/declarations/#Protocol-Conformance-Must-Not-Be-Redundant

## Pre-launch Checklist

- [x] I signed the [CLA].


hellohuanlin:
> Also, Both FlutterError and Swift.Error are public, extension FlutterError to Swift.Error is also public. 

I haven't reviewed it yet, but to clarify, i think our existing recommendation is to use **non-public** extension: 

```
extension FlutterError: Error {}
```

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -14,6 +14,32 @@ import Foundation
   #error("Unsupported platform.")
 #endif
 
+/// Error thrown by Pigeon. Encapsulates a code, message, and details.
+class PigeonError: Swift.Error {
```
hellohuanlin, line 18: Nit: `final class PigeonError: Error`

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -14,6 +14,32 @@ import Foundation
...
+
+  init(code: String, message: String?, details: Any?) {
+    self.code = code
+    self.message = message
+    self.details = details
+  }
+
+  var localizedDescription: String {
+    let detailsDescription: String
+    if let convertibleObject = details as? CustomStringConvertible {
```
hellohuanlin, line 31: does string interpolation already cover the `CustomStringConvertible` case?

packages/pigeon/lib/swift_generator.dart
```diff
@@ -19,16 +19,22 @@ class SwiftOptions {
...
     this.copyrightHeader,
+    this.swiftEmitErrorClass,
   });
 
   /// A copyright header that will get prepended to generated code.
   final Iterable<String>? copyrightHeader;
 
+  /// Whether to emit the PigeonError class in the generated code.
+  /// Defaults to true.
+  final bool? swiftEmitErrorClass;
```
hellohuanlin, line 30: nit: `swiftEmitPigeonErrorClass`

packages/pigeon/lib/swift_generator.dart
```diff
@@ -673,10 +681,10 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
   void _writeCreateConnectionError(Indent indent) {
     indent.newln();
     indent.writeScoped(
-        'private func createConnectionError(withChannelName channelName: String) -> FlutterError {',
+        'private func createConnectionError(withChannelName channelName: String) -> PigeonError {',
```
hellohuanlin, line 684: what happens to existing code that throws `FlutterError`? Do we support both? or do you completely swap it to `PigeonError`.

hellohuanlin (commented):
I haven't looked into the details. One key thing in my mind is that `FlutterError` should still be supported (not sure if this PR changes that). 

I think I agree with the general direction of creating `PigeonError`, but I'd like to hear from @stuartmorgan too.

stuartmorgan:
@hellohuanlin I thought we had determined that the extension was internal and wouldn't cause collisions; what did we miss? Is it not possible to give the extension the same visibility we would a new class?

hellohuanlin:
Just did a simple experiment with 2 modules: 

In ChildFramework: 

```
extension FlutterError: Error {}
```

In ParentFramework: 

```
import ChildFramework

extension FlutterError: Error {}
```

And confirmed that we got an error in ParentFramework: 

> Redundant conformance of 'FlutterError' to protocol 'Error'

So the extensions are indeed public.

packages/pigeon/lib/swift_generator.dart
bc-lee: If someone have a code that throws `FlutterError`, they are encouraged to change it to `PigeonError`. The whole point of this PR is that packages shouldn't use `FlutterError` to conform to the `Swift.Error` protocol. I have updated the README(`packages/pigeon/README.md`) to reflect this change.

packages/pigeon/example/app/ios/Runner/Messages.g.swift
bc-lee: Yeah. I thought I had to check multiple cases, but aparently the simple solution is enough. Done in c2beea180e17fa6ed3b3400206c0d5ae98ebff75.

packages/pigeon/lib/swift_generator.dart
bc-lee: Can you provide the names of open-source downstream projects that I can check to see if this PR breaks them?

tarrinneal:
It may be best to handle this similarly to the way we handle kotlin error classes. Each file has it's own unique error class.

packages/pigeon/lib/swift_generator.dart
stuartmorgan: > Ah that will be a breaking change and may cause chaos.

Compile time breaking changes in Pigeon are fine; we do them pretty frequently. Clients of Pigeon are fully in control of when they update their Pigeon generation, and Pigeon isn't a transitive dependency in package trees so there aren't any version conflict concerns.

hellohuanlin: oh i was referring to runtime breaking changes. let's say we have this old code: 

```
// custom code
func fooAPI() throws {
  throw FlutterError(...)
}

// generated code
func callTheFooAPI() {
  do {
    try fooAPI()
  } catch {
    if let e = error as? FlutterError {
      // encode error and send to dart side
    }
  }
}
```
Then when the customer bump the version, which doesn't handle the `FlutterError`: 
```
// custom code
func fooAPI() throws {
  throw FlutterError(...)
}

// generated code
func callTheFooAPI() {
  do {
    try fooAPI()
  } catch {
    if let e = error as? PigeonError { // <- here
      // encode error and send to dart side
    }
  }
}
```
Since no compile error, developers may not update their code.

bc-lee: Okay, I'll explain this using the source code of test codes in the pigeon package.

```dart
// https://github.com/flutter/packages/blob/dd01140f470e02b06d05ffa0213c7b26e89ae4c4/packages/pigeon/pigeons/core_tests.dart#L152
@HostApi()
abstract class HostIntegrationCoreApi {
...

  /// Returns an error, to test error handling.
  Object? throwError();
}
```

will be generated like this:
```swift
// https://github.com/flutter/packages/blob/dd01140f470e02b06d05ffa0213c7b26e89ae4c4/packages/pigeon/platform_tests/test_plugin/ios/Classes/CoreTests.gen.swift#L466
protocol HostIntegrationCoreApi {
...
  /// Returns an error, to test error handling.
  func throwError() throws -> Any?
}
...
// https://github.com/flutter/packages/blob/dd01140f470e02b06d05ffa0213c7b26e89ae4c4/packages/pigeon/platform_tests/test_plugin/ios/Classes/CoreTests.gen.swift#L701
    let throwErrorChannel = FlutterBasicMessageChannel(
      name:
        "dev.flutter.pigeon.pigeon_integration_tests.HostIntegrationCoreApi.throwError\(channelSuffix)",
      binaryMessenger: binaryMessenger, codec: codec)
    if let api = api {
      throwErrorChannel.setMessageHandler { _, reply in
        do {
          let result = try api.throwError()
          reply(wrapResult(result))
        } catch {
          reply(wrapError(error))
        }
      }
    } else {
      throwErrorChannel.setMessageHandler(nil)
    }
...
```
and current implementation of `throwError` function is as follows:

```swift
// https://github.com/flutter/packages/blob/dd01140f470e02b06d05ffa0213c7b26e89ae4c4/packages/pigeon/platform_tests/test_plugin/ios/Classes/TestPlugin.swift#L52
func throwError() throws -> Any? {
  throw FlutterError(code: "code", message: "message", details: "details")
}
```

Also, current implementation of `wrapError` function is as follows:
```swift
private func wrapError(_ error: Any) -> [Any?] {
  if let flutterError = error as? FlutterError {
    return [
      flutterError.code,
      flutterError.message,
      flutterError.details,
    ]
  }
  return [
    "\(error)",
    "\(type(of: error))",
    "Stacktrace: \(Thread.callStackSymbols)",
  ]
}
```

I believe the goal of `wrapError` is converting the error as a list with 3 elements, so that dart side decode it as a `FlutterError` object. We can throw what error we want in the dart side, and we just need to extract extra information such as `code`, `message`, and `details` from the error object.

Even we convert errors in the generated swift code to use `PigeonError`, the `wrapError` function will work as expected. I forgot to add a case for the conversion of `PigeonError` in the `wrapError` function, but it is easy to fix.

This PR's `wrapError` function will be like this:
```swift
private func wrapError(_ error: Any) -> [Any?] {
  if let flutterError = error as? FlutterError {
    return [
      flutterError.code,
      flutterError.message,
      flutterError.details,
    ]
  }
  if let pigeonError = error as? PigeonError {
    return [
      pigeonError.code,
      pigeonError.message,
      pigeonError.details,
    ]
  }
  return [
    "\(error)",
    "\(type(of: error))",
    "Stacktrace: \(Thread.callStackSymbols)",
  ]
}
```

Then, changes in the generated code won't make any problem in the dart side, even some older Swift code is throwing `FlutterError` object. Of course, as I wrote in the previous comment, implementations are encouraged to use `PigeonError` instead of `FlutterError` in the future to deal with the redundant protocol conformance issue.

In short, I believe that no compatibility issue will occur even after this PR is merged.

stuartmorgan: > oh i was referring to runtime breaking changes. let's say we have this old code:
> 
> ```
> // custom code
> func fooAPI() throws {
> [...]

Ah, right. I was thinking of the aync functions, where the type is part of the signature.

We definitely still need `FlutterError` handling in the `wrapError` function for runtime compatibility with synchronous code. That's a tiny amount of code though, luckily.

bc-lee:
> It may be best to handle this similarly to the way we handle kotlin error classes. Each file has it's own unique error class.

You're right. I don't need to create an option `swift_emit_pigeon_error_class` to control the error class generation. I will create a unique error class for each file.

bc-lee:
Failures in https://ci.chromium.org/ui/p/flutter/builders/try/Windows_x64%20windows-build_all_packages%20stable/5031/overview are not related to this change. I don't have permission to retry the build.

jmagman:
I re-ran the test, though it looks like you'll have to push another commit anyway to resolve the conflicts, which will re-run the tests.

bc-lee:
Hello, it seems that the code review has been stalled for 2 weeks and there is a merge conflict. Should I force-push to the latest main branch, or merge the main branch into my branch? Thanks.

jmagman:
You can merge main into your branch and resolve the merge conflicts.  Or rebase onto and resolve, up to you.  This won't be landable until the conflicts are resolved though.

bc-lee:
Merge conflict is now gone and unit tests are passing. Please take another look.

packages/pigeon/README.md
```diff
@@ -53,8 +53,7 @@ should be returned via the provided callback.
 To pass custom details into `PlatformException` for error handling, 
 use `FlutterError` in your Host API. [Example](./example/README.md#HostApi_Example).
 
-To use `FlutterError` in Swift you must first extend a standard error.
-[Example](./example/README.md#AppDelegate.swift).
+For swift, use `PigeonError` instead of `FlutterError`.
```
hellohuanlin, line 56: can you add an example here to replace existing example?

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -26,19 +26,50 @@ private func wrapError(_ error: Any) -> [Any?] {
...
 
-private func createConnectionError(withChannelName channelName: String) -> FlutterError {
-  return FlutterError(
+private func createConnectionError(withChannelName channelName: String) -> PigeonError {
+  return PigeonError(
     code: "channel-error", message: "Unable to establish connection on channel: '\(channelName)'.",
     details: "")
 }
 
+/// Error class for passing custom error details to Flutter.
```
hellohuanlin, line 49: "to Dart side".

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -26,19 +26,50 @@ private func wrapError(_ error: Any) -> [Any?] {
...
+  let message: String?
+  let details: Any?
+
+  init(code: String, message: String?, details: Any?) {
+    self.code = code
+    self.message = message
+    self.details = details
+  }
+
+  init(flutterError: FlutterError) {
```
hellohuanlin, line 61: why do we need this?

packages/pigeon/lib/swift_generator.dart
```diff
@@ -19,16 +19,21 @@ class SwiftOptions {
...
 
+  /// The name of the error class used for passing custom error parameters.
+  final String? errorClassName;
+
   /// Creates a [SwiftOptions] from a Map representation where:
   /// `x = SwiftOptions.fromList(x.toMap())`.
   static SwiftOptions fromList(Map<String, Object> map) {
     return SwiftOptions(
       copyrightHeader: map['copyrightHeader'] as Iterable<String>?,
+      errorClassName: map['errorClassName'] as String?,
```
hellohuanlin, line 36: is it open for developers to use a custom class name? why don't we just use `PigeonError`?

bc-lee: > is it open for developers to use a custom class name? why don't we just use PigeonError?

Generally, I believe calling pigeon once for each package (plugin, app, etc.) should suffice. In these scenarios, there's no need to concern ourselves with the generated class name. However, in this repo, pigeon is invoked multiple times for the same package, which results in multiple classes with the same name, ultimately causing a compilation error.

Kotlin encounters a similar problem but has a solution for it. This [Link](https://github.com/flutter/packages/pull/3234#issuecomment-1445022950) provides a background of the issue as well as its resolution. I have therefore applied the same solution to the generation of Swift code.

I believe overriding the error class name as an advanced usage, so I prefer not to integrate it with the command line option, similar to the approach taken by Kotlin.

hellohuanlin (approved):

packages/pigeon/example/app/ios/Runner/Messages.g.swift
hellohuanlin: Please remove if it's not used. I don't see why people still have to use `FlutterError`. Otherwise LGTM.

packages/pigeon/README.md
bc-lee: Done.

packages/pigeon/example/app/ios/Runner/Messages.g.swift
bc-lee: Done.

packages/pigeon/example/README.md
```diff
@@ -115,27 +115,25 @@ Future<bool> sendMessage(String messageText) {
 ### Swift
 
 This is the code that will use the generated Swift code to receive calls from Flutter.
-packages/pigeon/example/app/ios/Runner/AppDelegate.swift
+Unlike other languages, when throwing an error (both synchronous and asynchronous), use `PigeonError` instead of `FlutterError`, as `FlutterError` is not conforming to `Swift.Error`.
+Previously, a workaround was to declare an extension to `FlutterError` to conform to `Swift.Error`, but, as of Pigeon 19.0.0, `PigeonError` is provided for this purpose. Older code still using `FlutterError` will continue to work, but it is recommended to switch to `PigeonError` and remove the extension.
```
tarrinneal, line 119: `Unlike other languages, when throwing an error, use `PigeonError` instead of `FlutterError`,
as `FlutterError` does not conform to `Swift.Error`.

I don't think the second line is necessary.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 19.0.0
+* **Breaking Change** [swift] Do not use `FlutterError` on the generated code.
```
tarrinneal, line 2: Should be an empty line between these.

`* **Breaking Change** [swift] Removes `FlutterError` in favor of `PigeonError`.`

packages/pigeon/lib/swift_generator.dart
```diff
@@ -614,7 +621,7 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
     });
   }
 
-  void _writeWrapError(Indent indent) {
+  void _writeWrapError(SwiftOptions generatorOptions, Indent indent) {
     indent.newln();
     indent.write('private func wrapError(_ error: Any) -> [Any?] ');
     indent.addScoped('{', '}', () {
```
tarrinneal, line 627: @stuartmorgan I'm of the general opinion that leaving the legacy option here is unnecessary. What are your thoughts?

packages/pigeon/lib/swift_generator.dart
```diff
@@ -870,11 +891,48 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
...
     });
   }
+
+  void _writePigeonError(SwiftOptions generatorOptions, Indent indent) {
+    indent.newln();
+    indent.writeln(
+        '/// Error class for passing custom error details to Dart side.');
+    indent.writeln(
+        'final class ${_getErrorClassName(generatorOptions)}: Error {');
+    indent.nest(1, () {
```
tarrinneal, line 901: you should replace all the `nest` uses in this method with `writeScoped`

tarrinneal (changes requested):
Generally solid pr! I have a few nits and a general design question I'd like to run by @stuartmorgan before we can land this though. Thanks for putting the time into this!

packages/pigeon/lib/swift_generator.dart
stuartmorgan: That would silently change behavior of existing code, which we try to avoid. Breaks that require people to update to compile are fine, but this won't do that, so I think we should leave it.

We should put it after the expected codepath though.

packages/pigeon/example/README.md
bc-lee: Done.

packages/pigeon/CHANGELOG.md
bc-lee: Done.

tarrinneal: can you update the changelog itself to match my comment?

packages/pigeon/lib/swift_generator.dart
```diff
@@ -627,6 +634,16 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
...
+      indent.write(
+          'if let pigeonError = error as? ${_getErrorClassName(generatorOptions)} ');
+      indent.addScoped('{', '}', () {
+        indent.write('return ');
+        indent.addScoped('[', ']', () {
+          indent.writeln('pigeonError.code,');
+          indent.writeln('pigeonError.message,');
+          indent.writeln('pigeonError.details,');
+        });
+      });
```
tarrinneal, line 646: Can you move this section above the `if let flutterError` line (628)

bc-lee: Done.

packages/pigeon/lib/swift_generator.dart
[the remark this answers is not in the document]
tarrinneal: also, I'm not sure this is actually set up to do anything yet. Were you planning to make the class optional in the generator? (I may have just missed it though)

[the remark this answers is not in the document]
tarrinneal: which also means there should probably be a test for this as well.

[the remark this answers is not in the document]
bc-lee: I think you're looking older version of the code. This boolean field was removed in [800ba60a8](https://github.com/flutter/packages/pull/6611/commits/800ba60a802fff69afb75b547cb76dae3bfa8e15).

Current diff is here: https://github.com/flutter/packages/pull/6611/files#diff-bfdb6a91beb03a906435e77e0168117f3f3977ee4d6f8bcaa1724156ae4dc27c

packages/pigeon/lib/swift_generator.dart
```diff
@@ -671,17 +689,20 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
...
       _writeWrapResult(indent);
-      _writeWrapError(indent);
+      _writeWrapError(generatorOptions, indent);
     }
     if (hasFlutterApi) {
-      _writeCreateConnectionError(indent);
+      _writeCreateConnectionError(generatorOptions, indent);
     }
+
+    _writePigeonError(generatorOptions, indent);
```
tarrinneal, line 698: One last nit and we can land it, can you move the pigeon error above the other methods here? above `if (hasHostApi) {'

tarrinneal (approved):
one last thing! Thanks a bunch for putting this pr together!

packages/pigeon/lib/swift_generator.dart
bc-lee: Done.

bc-lee:
Should I put my changelog in the `NEXT` section or in the next version?

```
<<<<<<< feature/pigeon-swift-error
## 19.0.0

* **Breaking Change** [swift] Removes `FlutterError` in favor of `PigeonError`.
=======
## NEXT

* Updates minimum supported SDK version to Flutter 3.16/Dart 3.2.
>>>>>>> main
```

tarrinneal:
```## 19.0.0

* **Breaking Change** [swift] Removes `FlutterError` in favor of `PigeonError`.
* Updates minimum supported SDK version to Flutter 3.16/Dart 3.2.```

PR #6623, opened 2024-04-29
[go_router] Use the correct configuration to build the state passed to the `onExit`

ValentinVignal:
While working on https://github.com/flutter/packages/pull/6614, I notice some issues with the state that is given to the `onExit` (see the run https://github.com/flutter/packages/pull/6614/checks?check_run_id=24284539541)

This PR uses the correct configuration to build the state and pass it to the `onExit`

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/go_router/lib/src/delegate.dart
```diff
@@ -233,7 +233,7 @@ class GoRouterDelegate extends RouterDelegate<RouteMatchList>
           exitingMatches.length - 1,
           context: navigatorContext,
           matches: exitingMatches,
-          configuration: configuration,
+          // configuration: configuration,
```
chunhtai, line 236: ?!

packages/go_router/lib/src/delegate.dart
```diff
@@ -276,15 +274,14 @@ class GoRouterDelegate extends RouterDelegate<RouteMatchList>
...
-          configuration: configuration,
         );
       }
       return SynchronousFuture<bool>(false);
     }
 
     final FutureOr<bool> exitFuture = goRoute.onExit!(
       context,
-      match.buildState(_configuration, configuration),
+      match.buildState(_configuration, currentConfiguration),
```
chunhtai, line 284: so the difference is whether we use the new RouteMatchList after the change vs old RouteMatchList?

I think for redirect we use the new RouteMatchList. it seems to previous implementation is more inline with the redirect API no?

chunhtai: I maybe wrong here. do you have use case that using the old routematchList is more preferred? on the other hand, I plan to redesign redirect, so maybe we should just provide both before and after?

packages/go_router/lib/src/delegate.dart
ValentinVignal: Oh my ! Sorry about that. I removed it in [refactor: Remove commented code](https://github.com/flutter/packages/pull/6623/commits/560d3f44054795cc828101f3ad889830979d9853)

packages/go_router/lib/src/delegate.dart
ValentinVignal: Yes, I found one in https://github.com/flutter/packages/pull/6614. I was making `go_router_builder` use `onExit` in some tests failed https://github.com/flutter/packages/pull/6614/checks?check_run_id=24284539541.

The issue was when using `.go()`. The `onExit` had the state of the new page, not the previous page.

The test I wrote in this PR 

```dart
'It should provide the correct path parameters to the onExit callback during a go
```

fails without my changes.

> On the other hand, I plan to redesign redirect, so maybe we should just provide both before and after?

Sure, do you want me to change something then?

chunhtai: I will need to redesign the onExit in the future. It also has issue with popscope and cupertinobackswipe. Thing may become a lot different after I resolve all these issues.

After thinking about it more, I think we should just keep it simple for now. I think what you have is fine.

chunhtai (approved):
LGTM

ValentinVignal:
@hangyujin Is there something I need to do about this PR ?

hangyujin (approved):
LGTM

PR #6627, opened 2024-04-29
Skip podspec Swift Search Path validation if only swift file is Package.swift

vashworth:
Usually when an iOS plugin uses Swift files, it requires a workaround in the podspec to add Swift to the search paths. Part of the `podspec-check` command is validating this workaround is found. However, when the only Swift file is the `Package.swift` (Swift Package Manager manifest), skip this validation since having this file does not indicate the plugin uses Swift files.

Fixes https://github.com/flutter/flutter/issues/147548.

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
> Also, skip license check for Package.swift files.

I don't think we should skip the license here; it's authored code, not generated code. Even though it's not very interesting, it's still non-generated.

script/tool/lib/src/podspec_check_command.dart
```diff
@@ -171,7 +172,8 @@ class PodspecCheckCommand extends PackageLoopingCommand {
         return false;
       }
       final String filePath = entity.path;
-      return path.extension(filePath) == '.swift';
+      return entity.basename != 'Package.swift' &&
```
stuartmorgan, line 175: Shouldn't we check the path to make sure it's at the root of the module? In theory someone could have a `Package.swift` as an actual source file within the plugin (although I guess it's very unlikely that even if they did it would be the *only* Swift file).

vashworth: Sure, can do

vashworth:
> > Also, skip license check for Package.swift files.
> 
> I don't think we should skip the license here; it's authored code, not generated code. Even though it's not very interesting, it's still non-generated.

I was thinking it didn't need to be licensed since we don't license the podspec file. Also, Package.swift is very picky about what comments are at the top of the file. The first comment has to be the tools version.

So to license it, it would look like
```swift
// swift-tools-version: 5.9
// The swift-tools-version declares the minimum version of Swift required to build this package.

// found in the LICENSE file.

import PackageDescription

let package = Package(
...
```

Is that okay?

stuartmorgan:
That should be fine; it's similar to how `bash` script files have a `#!` comment before the license block section since that also has to be the very first line.

stuartmorgan (approved):
LGTM

PR #6673, opened 2024-05-06
[ci] Reduce the number of tasks in recipe CQ

stuartmorgan:
We only need enough recipe CQ coverage to cover the platforms and functionality in the recipes, which doesn't require running so many of our tests there (especially multiple shards of the same test). This reduces the number of tasks in the recipe CQ, while still keeping broad coverage of different kinds of functionality.

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -150,7 +150,6 @@ targets:
     recipe: packages/packages
     timeout: 60
     properties:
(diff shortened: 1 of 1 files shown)


stuartmorgan:
See also previous discussion in https://github.com/flutter/packages/pull/5226

tarrinneal (approved):

PR #6685, opened 2024-05-07
[go_router] guard context access in then clauses

goderbauer:
Prepares for a fix to the `use_build_context_synchronously` lint (https://dart-review.googlesource.com/c/sdk/+/365541) that will complain about these unsafe usages.

### DIFF
--- a/packages/go_router/lib/src/parser.dart
+++ b/packages/go_router/lib/src/parser.dart
@@ -72,7 +72,9 @@ class GoRouteInformationParser extends RouteInformationParser<RouteMatchList> {
       return debugParserFuture = _redirect(context, matchList)
           .then<RouteMatchList>((RouteMatchList value) {
         if (value.isError && onParserException != null) {
-          return onParserException!(context, value); // ignore: use_build_context_synchronously
+          // TODO(chunhtai): Figure out what to return if context is invalid.
+          // ignore: use_build_context_synchronously
+          return onParserException!(context, value);
         }
         return value;
       });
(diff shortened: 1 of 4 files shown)


packages/go_router/lib/src/parser.dart
```diff
@@ -72,7 +72,7 @@ class GoRouteInformationParser extends RouteInformationParser<RouteMatchList> {
       return debugParserFuture = _redirect(context, matchList)
           .then<RouteMatchList>((RouteMatchList value) {
         if (value.isError && onParserException != null) {
-          return onParserException!(context, value);
+          return onParserException!(context, value); // ignore: use_build_context_synchronously
```
goderbauer, line 75: @chunhtai Here and below I am actually not sure what should happen if the context has become invalid. Do you have a suggestion?

goderbauer:
cc @srawlins

srawlins (approved):
😁

packages/go_router/lib/src/parser.dart
chunhtai: if the context is invalid, it should just return a never complete future.

Something like `return Completer<RouteMatchList>().future`

chunhtai (approved):
LGTM, can you add a todo for the ignore?

goderbauer:
Reason for override:
* PRs that only affect local development of the package (e.g., changes to ignored lints).
* PRs that only affect unpublished parts of example apps.

PR #6629, opened 2024-04-30
[camera] Change default Android implementation from `camera_android` to `camera_android_camerax`

camsim99:
Breaking change: Changes default Android implementation from `camera_android` to `camera_android_camerax` via a breaking change and updates relevant documentation.

~<ins>**Note:**</ins> Planning to land this after https://github.com/flutter/packages/pull/6608, as it is a semi-crucial fix.~ Done :)

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ ] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[...]

PR #6697, opened 2024-05-08
[camerax] Make fixes required to swap camera_android_camerax for camera_android

camsim99:
Makes changes needed to land https://github.com/flutter/packages/pull/6629. Specifically:

- Fixes timing issue with `stopVideoRecording` such that the `Future` it returns will only complete when CameraX reports that the recording is finalized (via listening for the [finalized video recording event](https://developer.android.com/reference/androidx/camera/video/VideoRecordEvent.Finalize))
- Modifies `startVideoCapturing` such that the `Future` it returns will only complete when CameraX reports that video capturing has started (via listening for the [started video recording event](https://developer.android.com/reference/androidx/camera/video/VideoRecordEvent.Start))
- Adds empty implementation and TODO for implementing `setDescriptionWhileRecording`

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_android_camerax/README.md
+++ b/packages/camera/camera_android_camerax/README.md
@@ -37,6 +37,9 @@ use cases, the plugin behaves according to the following:
     video recording and image streaming is supported, but concurrent video recording, image
     streaming, and image capture is not supported.
 
+### `setDescriptionWhileRecording` is unimplemented [Issue #148013][148013]
+`setDescriptionWhileRecording`, used to switch cameras while recording video, is currently unimplemented.
+
 ### 240p resolution configuration for video recording
 
 240p resolution configuration for video recording is unsupported by CameraX,
@@ -64,11 +67,4 @@ For more information on contributing to this plugin, see [`CONTRIBUTING.md`](CON
 [6]: https://developer.android.com/media/camera/camerax/architecture#combine-use-cases
 [7]: https://developer.android.com/reference/android/hardware/camera2/CameraMetadata#INFO_SUPPORTED_HARDWARE_LEVEL_3
 [8]: https://developer.android.com/reference/android/hardware/camera2/CameraMetadata#INFO_SUPPORTED_HARDWARE_LEVEL_LIMITED
-[120462]: https://github.com/flutter/flutter/issues/120462
-[125915]: https://github.com/flutter/flutter/issues/125915
-[120715]: https://github.com/flutter/flutter/issues/120715
-[120468]: https://github.com/flutter/flutter/issues/120468
-[120467]: https://github.com/flutter/flutter/issues/120467
-[125371]: https://github.com/flutter/flutter/issues/125371
-[126477]: https://github.com/flutter/flutter/issues/126477
-[127896]: https://github.com/flutter/flutter/issues/127896
+[148013]: https://github.com/flutter/flutter/issues/148013
--- a/packages/camera/camera_android_camerax/README.md
+++ b/packages/camera/camera_android_camerax/README.md
@@ -38,7 +38,8 @@ use cases, the plugin behaves according to the following:
     streaming, and image capture is not supported.
 
 ### `setDescriptionWhileRecording` is unimplemented [Issue #148013][148013]
-`setDescriptionWhileRecording`, used to switch cameras while recording video, is currently unimplemented.
+`setDescriptionWhileRecording`, used to switch cameras while recording video, is currently unimplemented
+due to this not currently being supported by CameraX.
 
 ### 240p resolution configuration for video recording
 
--- a/packages/camera/camera_android_camerax/example/integration_test/integration_test.dart
+++ b/packages/camera/camera_android_camerax/example/integration_test/integration_test.dart
@@ -195,11 +195,8 @@ void main() {
     await controller.startVideoRecording();
     final int recordingStart = DateTime.now().millisecondsSinceEpoch;
 
-    const int recordDuration = 2000;
-    sleep(const Duration(milliseconds: recordDuration));
+    sleep(const Duration(seconds: 2));
 
-    final int preStopTime =
-        DateTime.now().millisecondsSinceEpoch - recordingStart;
     final XFile file = await controller.stopVideoRecording();
     final int postStopTime =
         DateTime.now().millisecondsSinceEpoch - recordingStart;
@@ -212,7 +209,6 @@ void main() {
     final int duration = videoController.value.duration.inMilliseconds;
     await videoController.dispose();
 
-    expect(duration, greaterThan(preStopTime));
     expect(duration, lessThan(postStopTime));
   });
 
--- a/packages/camera/camera_android_camerax/example/integration_test/integration_test.dart
+++ b/packages/camera/camera_android_camerax/example/integration_test/integration_test.dart
@@ -219,8 +219,8 @@ void main() {
     }
 
     final CameraController controller = CameraController(cameras[0],
-        mediaSettings: const MediaSettings(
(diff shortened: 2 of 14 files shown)


packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -122,7 +128,7 @@ class AndroidCameraCameraX extends CameraPlatform {
 
   /// The prefix used to create the filename for video recording files.
   @visibleForTesting
-  final String videoPrefix = 'MOV';
+  final String videoPrefix = 'REC';
```
camsim99, line 131: Changed this + the .temp --> .mp4 extension change to be consistent with camera_android, see [source](https://github.com/flutter/packages/blob/d670b2c38c8db0a773aa703e7d3f15682e05ad7f/packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java#L1250).

packages/camera/camera_android_camerax/README.md
```diff
@@ -64,11 +67,4 @@ For more information on contributing to this plugin, see [`CONTRIBUTING.md`](CON
 [6]: https://developer.android.com/media/camera/camerax/architecture#combine-use-cases
 [7]: https://developer.android.com/reference/android/hardware/camera2/CameraMetadata#INFO_SUPPORTED_HARDWARE_LEVEL_3
 [8]: https://developer.android.com/reference/android/hardware/camera2/CameraMetadata#INFO_SUPPORTED_HARDWARE_LEVEL_LIMITED
-[120462]: https://github.com/flutter/flutter/issues/120462
```
reidbaker, line 67: Intentional removal?

packages/camera/camera_android_camerax/example/integration_test/integration_test.dart
```diff
@@ -178,4 +179,86 @@ void main() {
...
+    await controller.prepareForVideoRecording();
+
+    await controller.startVideoRecording();
+    final int recordingStart = DateTime.now().millisecondsSinceEpoch;
+
+    sleep(const Duration(seconds: 2));
+
+    final XFile file = await controller.stopVideoRecording();
+    final int recordingTime =
+        DateTime.now().millisecondsSinceEpoch - recordingStart;
```
reidbaker, line 202: I think we can make this test a bit better by comparing on both sides of the stopRecording call. 


```suggestion
    // Test should pass with this at any non zero value. 
    sleep(const Duration(seconds: 2));
    
     final int preStopTime =
        DateTime.now().millisecondsSinceEpoch - recordingStart;
    final XFile file = await controller.stopVideoRecording();
    final int postStopTime =
        DateTime.now().millisecondsSinceEpoch - recordingStart;
        ....
        expect(duration, greaterThan(preStopTime));
        expect(duration, lessThan(postStopTime));
```

Note: 
We should be using a package like https://pub.dev/packages/system_clock because DateTime.now is vulnerable to the clock changing while the test is running but that feels out of scope.

packages/camera/camera_android_camerax/example/integration_test/integration_test.dart
```diff
@@ -178,4 +179,86 @@ void main() {
...
+    await controller.prepareForVideoRecording();
+
+    int startPause;
+    int timePaused = 0;
+
+    await controller.startVideoRecording();
+    final int recordingStart = DateTime.now().millisecondsSinceEpoch;
+    sleep(const Duration(milliseconds: 500));
+
+    await controller.pauseVideoRecording();
```
reidbaker, line 234: Possibly overkill but consider dropping the pause/"measure pause time" into either a loop or a method. When we were debugging having the blocks back to back made it harder to understand until you realized the blocks where supposed to be identical.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
reidbaker: Are we guaranteed that the file is an mp4? If not then maybe we should leave it .temp because having the wrong file extension is worse than having an unknown file extension.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -777,6 +783,15 @@ class AndroidCameraCameraX extends CameraPlatform {
     await _unbindUseCaseFromLifecycle(preview!);
   }
 
+  /// Sets the active camera while recording.
+  ///
+  /// Currently unsupported, so is a no-op.
+  @override
+  Future<void> setDescriptionWhileRecording(CameraDescription description) {
```
reidbaker, line 790: Does this mean we have a feature we know will silently stop working when we modify the default to camerax?

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -979,21 +993,22 @@ class AndroidCameraCameraX extends CameraPlatform {
           'Attempting to stop a '
               'video recording while no recording is in progress.');
     }
+
+    /// Stop the active recording and wiat for the video recording to be finalized.
```
reidbaker, line 997:
```suggestion
    /// Stop the active recording and wait for the video recording to be finalized.
```

reidbaker (approved):

packages/camera/camera_android_camerax/README.md
camsim99: Yes! There aren't referenced in the README anymore.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
camsim99: It does mean that :/ but the fact that it will be a breaking change can mitigate that fact; it will be noted in the README here but I can also add it to the camera/camera CHANGELOG and camera/camera README.

@stuartmorgan any thoughts on this? I realized recently that this was not implemented for camera_android_camerax.

stuartmorgan: I would definitely note the loss of functionality in the CHANGELOG for the `camera` breaking change.

This is a pretty new feature IIRC, and I would expect relatively niche, so losing it in a breaking change doesn't seem like a major issue. People will always have the possibility of using `camera_android` (or a fork of it, if/when we stop maintaining it) if they need it.

packages/camera/camera_android_camerax/example/integration_test/integration_test.dart
camsim99: The preStopTime check didn't quite work I assume because stopping video recording takes so long but I kept the naming of postStopTime for clarity

camsim99: I suppose we could also fix this by listening for the [`VideoRecordEvent.Start`](https://developer.android.com/reference/androidx/camera/video/VideoRecordEvent.Start) event to return. Let me give it a go!

camsim99: @reidbaker Ok so I added the fix but it didn't quite work still :/ the duration increased but guessing there still some lag time in starting the video. Still left it in though as it is a better approximation.

camsim99: Going to land this as-is since I've done all I can to remedy this issue. If we feel like we need this fixed, I can file a bug and escalate to CameraX.

camsim99: Filed https://github.com/flutter/flutter/issues/148138 for this.

PR #6700, opened 2024-05-09
[multicast_dns] Optimized Socket Binding: Always bind to 0.0.0.0 for simplicity and efficiency - #79772

biagiopietro:
I encountered this package long time ago (see linked issue below) and there were cases where it wasn't working.
After 3 years (yeah it's more time than expected) I managed to find the time to dust off `wireshark` and have look again.

# Preamble

Considering the following setup

image

Where Raspberry pi runs a `mDNS` service using the following `go`:

<details><summary>main.go</summary>
<p>

```go
package main

import (
	"fmt"
	"github.com/hashicorp/mdns"
	"os"
  "net/http"
)

func health(w http.ResponseWriter, r *http.Request) {
  fmt.Fprintf(w, "ok")
}

func main() {

	hostname, err := os.Hostname()

	if err != nil {
		panic(fmt.Sprintf("Error getting current hostname, description: %s", err.Error()))
	}

	info := []string{"mDNS get server"}

	service, err := mdns.NewMDNSService(hostname, "_test._tcp", "", "", 8080, nil, info)

	if err != nil {
		panic(fmt.Sprintf("Error while exporting the service, description: %s", err.Error()))
	}

	server, err := mdns.NewServer(&mdns.Config{Zone: service})

	if err != nil {
		panic(fmt.Sprintf("Error while setting the discover server up, description: %s", err.Error()))
	}

	defer server.Shutdown()
	
	http.HandleFunc("/", health)
	http.ListenAndServe(":8081",nil)
}
```
</p>
</details> 

Considering the following client (which I got from [here](https://pub.dev/packages/multicast_dns/example)):

<details><summary>client.dart</summary>
<p>

```dart
// found in the LICENSE file.

// Example script to illustrate how to use the mdns package to discover the port
// of a Dart observatory over mDNS.

// ignore_for_file: avoid_print

import 'package:multicast_dns/multicast_dns.dart';

Future<void> main() async {
  // Parse the command line arguments.

  const String name = '_test._tcp.local';
  final MDnsClient client = MDnsClient();
  // Start the client with default options.
  await client.start();

  // Get the PTR record for the service.
  await for (final PtrResourceRecord ptr in client
      .lookup<PtrResourceRecord>(ResourceRecordQuery.serverPointer(name))) {
    // Use the domainName from the PTR record to get the SRV record,
    // which will have the port and local hostname.
    // Note that duplicate messages may come through, especially if any
    // other mDNS queries are running elsewhere on the machine.
    await for (final SrvResourceRecord srv in client.lookup<SrvResourceRecord>(
        ResourceRecordQuery.service(ptr.domainName))) {
      // Domain name will be something like "io.flutter.example@some-iphone.local._dartobservatory._tcp.local"
      final String bundleId =
          ptr.domainName; //.substring(0, ptr.domainName.indexOf('@'));
      print('Dart observatory instance found at '
          '${srv.target}:${srv.port} for "$bundleId".');
    }
  }
  client.stop();

  print('Done.');
}
```

</p>
</details> 

# What happens 
When running the client script (`dart run client.dart` so with the latest package of `multicast_dns` package) as is, a list of sockets is created which are bind to port `5353` and `IP`s:

- `0.0.0.0`;
- `127.0.0.1`;
- `192.168.2.16`;
- `172.17.0.1`;

a list of interfaces (see list below) are _joined_ to the multicast socket which is bound to `0.0.0.0:5353`:

- `lo` (with address `127.0.0.1`);
- `wlan0` (with address `192.168.2.16`);
- `docker0` (with address `172.17.0.1`).

and eventually when `lookup` function is being called `QM`queries are being sent from `ALL` the sockets in the list; which means that for `0.0.0.0` the `IP` address chosen by the **operating system** will depend on various factors such as the routing table, the default network interface, and the specific configuration of the network interfaces on the machine. It could be sent from any of the `IP` addresses associated with the machine's network interfaces, including `IP` addresses assigned to physical network adapters or virtual interfaces.

Using `Wireshark`, I can see that 2 `QM` packets are being sent and I can see that `mDNS` service is responding to the client with proper packet but it seems that the socket opened at `0.0.0.0:5353` is not reading them at all even though the socket is still open.

```shell
Source	        Destination	Protocol	Length	Info
192.168.2.16	224.0.0.251	MDNS		    76	Standard query 0x0000 PTR _test._tcp.local, "QM" question
192.168.2.16	224.0.0.251	MDNS		    76	Standard query 0x0000 PTR _test._tcp.local, "QM" question
192.168.2.7	192.168.2.16	MDNS		   180	Standard query response 0x0000 PTR mdnsserv._test._tcp.local SRV 10 1 8080 mdnsserv A 127.0.1.1 TXT
192.168.2.7	192.168.2.16	MDNS		   180	Standard query response 0x0000 PTR mdnsserv._test._tcp.local SRV 10 1 8080 mdnsserv A 127.0.1.1 TXT
```

# First approach (not sure if it's RFC 6762 friendly)

I had the "_feeling_" that sending `QM` packets to `0.0.0.0:5353` and other interfaces on the same port would generate some _sort of unexpected behavior_ due to the nature of `0.0.0.0` which `IP` selections **depends on multiple factors**.

Therefore I tried initially to change the `incoming` socket (the one bound to `0.0.0.0`) from:

```dart
final RawDatagramSocket incoming = await _rawDatagramSocketFactory(
      listenAddress.address,
      selectedMDnsPort,
      reuseAddress: true,
      reusePort: true,
      ttl: 255,
    );
```
to
``` dart
final RawDatagramSocket incoming = await _rawDatagramSocketFactory(
      listenAddress.address,
      0,
      reuseAddress: true,
      reusePort: true,
      ttl: 255,
    );
```
which essentially delegates to `OS` to choose a **random** port (instead of forcing `5353`).

In this case the client managed to process correctly all the packages for discovering the `mDNS` service, indeed in `Wireshark` I could see:

```shell
Source	        Destination	Protocol	Length	Info
192.168.2.16	224.0.0.251	MDNS		    76	Standard query 0x0000 PTR _test._tcp.local, "QM" question
192.168.2.16	224.0.0.251	MDNS		    76	Standard query 0x0000 PTR _test._tcp.local, "QM" question
192.168.2.7	192.168.2.16	MDNS		   180	Standard query response 0x0000 PTR mdnsserv._test._tcp.local SRV 10 1 8080 mdnsserv A 127.0.1.1 TXT
192.168.2.7	192.168.2.16	MDNS		   180	Standard query response 0x0000 PTR mdnsserv._test._tcp.local SRV 10 1 8080 mdnsserv A 127.0.1.1 TXT
192.168.2.16	224.0.0.251	MDNS		    85	Standard query 0x0000 SRV mdnsserv._test._tcp.local, "QM" question
192.168.2.16	224.0.0.251	MDNS		    85	Standard query 0x0000 SRV mdnsserv._test._tcp.local, "QM" question
192.168.2.7	192.168.2.16	MDNS		   123	Standard query response 0x0000 SRV 10 1 8080 mdnsserv A 127.0.1.1
192.168.2.7	192.168.2.16	MDNS		   123	Standard query response 0x0000 SRV 10 1 8080 mdnsserv A 127.0.1.1
```
and on the client I could see the message:

```shell
Dart observatory instance found at mdnsserv:8080 for "mdnsserv._test._tcp.local"
```

⚠️ : Again, I'm not sure if it can be considered a _solution_ because I dunno is`RFC 6762` friendly, I checked some packages which implement `mDNS` clients and I saw some of them doing what I proposed. I would like to hear comments about it.

# Second approach (which it's what is presented in this PR)

After trying the first approach I realized that _maybe_ there is no need to open sockets on more interfaces (and therefore send `QM` messages) it maybe be enough to send and listen only on a socket bound to `0.0.0.0` since, again, listen on **ANY** `IP` and send packets from a selected `IP` address chosen by the `OS`.

Also in this case the client managed to process correctly all the packages for discovering the `mDNS` service, indeed in `Wireshark` I could see:

```shell
Source	        Destination	Protocol	Length	Info
192.168.2.16	224.0.0.251	MDNS		    76	Standard query 0x0000 PTR _test._tcp.local, "QM" question
192.168.2.7	192.168.2.16	MDNS		   180	Standard query response 0x0000 PTR mdnsserv._test._tcp.local SRV 10 1 8080 mdnsserv A 127.0.1.1 TXT
192.168.2.16	224.0.0.251	MDNS		    85	Standard query 0x0000 SRV mdnsserv._test._tcp.local, "QM" question
192.168.2.7	192.168.2.16	MDNS		   123	Standard query response 0x0000 SRV 10 1 8080 mdnsserv A 127.0.1.1
```
and on the client I could see the message:

```shell
Dart observatory instance found at mdnsserv:8080 for "mdnsserv._test._tcp.local"
```

# Third approach (It did not work but mentioning for completeness)
The idea here is to **don't send** `QM` packets via `0.0.0.0` but just **listen** on possible response/s since packets would be send via the following `IP`s and `0.0.0.0` should represent **ANY** `IP`.

- `127.0.0.1`;
- `192.168.2.16`;
- `172.17.0.1`.

# Fourth approach (It did not work but mentioning for completeness)

Another solution that I tried but unfortunately it did not work, was to put `0.0.0.0` as **last** item in the socket list so `QM` packets would be sent according to the following order:

- `127.0.0.1`;
- `192.168.2.16`;
- `172.17.0.1`;
- `0.0.0.0`.

<details><summary>multicast_dns.start() function</summary>
<p>

```dart
  Future<void> start({
    InternetAddress? listenAddress,
    NetworkInterfacesFactory? interfacesFactory,
    int mDnsPort = mDnsPort,
    InternetAddress? mDnsAddress,
  }) async {
    listenAddress ??= InternetAddress.anyIPv4;
    interfacesFactory ??= allInterfacesFactory;

    assert(listenAddress.address == InternetAddress.anyIPv4.address ||
        listenAddress.address == InternetAddress.anyIPv6.address);

    if (_started || _starting) {
      return;
    }
    _starting = true;

    final int selectedMDnsPort = _mDnsPort = mDnsPort;
    _mDnsAddress = mDnsAddress;

    // Listen on all addresses.
    final RawDatagramSocket incoming = await _rawDatagramSocketFactory(
      listenAddress.address,
      selectedMDnsPort,
      reuseAddress: true,
      reusePort: true,
      ttl: 255,
    );

    _mDnsAddress ??= incoming.address.type == InternetAddressType.IPv4
        ? mDnsAddressIPv4
        : mDnsAddressIPv6;

    final List<NetworkInterface> interfaces =
        (await interfacesFactory(listenAddress.type)).toList();

    for (final NetworkInterface interface in interfaces) {
      // Create a socket for sending on each adapter.
      final InternetAddress targetAddress = interface.addresses[0];
      // Join multicast on this interface.
      incoming.joinMulticast(_mDnsAddress!, interface);
    }

    // Can't send to IPv6 any address.
    if (incoming.address != InternetAddress.anyIPv6) {
      _sockets.add(incoming);
    } else {
      _toBeClosed.add(incoming);
    }

    incoming.listen((RawSocketEvent event) => _handleIncoming(event, incoming));
    _started = true;
    _starting = false;
  }
```
</p>
</details> 

The idea is indeed to let the first 3 `IP`s to send the `QM` packets which response should be _hopefully_ captured by the `incoming` socket before the socket on `0.0.0.0` would send the `QM` packet too.

# Wireshark filter

```shell
(ip.src==192.168.2.7 || ip.src==192.168.2.16) && udp.port eq 5353
```

# Related Issue
- It should resolves issue [#79772](https://github.com/flutter/flutter/issues/79772)

# Disclaimers

- I'm not expert in `flutter`/`dart`, I pulled the code and I tried to debug it with help of uncle `google` and `print()`;
- I don't have a huge expertise in networking but I _know_ how to play a bit with `Wireshark`, inspect the networks and craft packets.

## Pre-launch Checklist

- [X] I signed the [CLA].


jmagman (commented):
Apologies for the delay reviewing this.

@vashworth kindly tested this out against the `flutter` CLI and confirmed it didn't break wireless debugging for `flutter run` or `flutter attach`.

This change generally LGTM, though I would like a second opinion from @cbracken who might remember more about why the original code was written that way.

1. This needs a pubspec version bump and CHANGELOG update.  See https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version-and-changelog-updates
2. Can you add a test to https://github.com/flutter/packages/blob/main/packages/multicast_dns/test/client_test.dart?

cbracken:
Taking a look!

cbracken:
I did a bit of exploratory `git blame`ing to see if I could figure out the history of the code in question in case there's any additional useful info in log comments (there wasn't), but for the record:

The existing code was derived from the (long-dead, now) Dartino code and was added in
https://github.com/flutter/packages/commit/a062a2852223af630f1a14266009db072333c227

It's derived from older Dartino code from the Dart SDK. The most recent version of that can be viewed here:
https://github.com/dart-archive/sdk/tree/master/pkg/mdns

The Dartino mdns package was originally added in this commit:
https://github.com/dart-archive/sdk/commit/f0d30ec90e9c6c9c157fa583c74a5f8d5321891d

Looking at the actual logic in the context of the rest of the code now.

cbracken:
Thanks for sending the patch.

Thinking about this a bit, I agree that setting up a single socket on 0.0.0.0 should be sufficient for _listening_ on all network interfaces. As you note, for _sending_, we'll rely on the operating system's routing to determine which network interface is used for sending.

The current package doesn't really allow for interface-specific tuning/filtering so I don't think we're losing anything here and the code and resource management is significantly simpler.

The existing tests should cover the majority of the behaviour here, but as @jmagman says, please add a test that verifies the change in behaviour.

cbracken (commented):
looks good modulo comments above.

packages/multicast_dns/lib/multicast_dns.dart
```diff
@@ -130,30 +130,6 @@ class MDnsClient {
...
-      // Create a socket for sending on each adapter.
-      final InternetAddress targetAddress = interface.addresses[0];
-      final RawDatagramSocket socket = await _rawDatagramSocketFactory(
-        targetAddress,
-        selectedMDnsPort,
-        reuseAddress: true,
-        reusePort: true,
-        ttl: 255,
-      );
-      _sockets.add(socket);
```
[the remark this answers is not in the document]
biagiopietro: Thanks so much @cbracken and @jmagman for your time spent in this PR.
I applied the suggested changes. Please keep in mind that I'm not pro in `dart`/`flutter`, I did try my best for the tests. 
Any code suggestion is really appreciated.
Thanks again 🙇

packages/multicast_dns/lib/multicast_dns.dart
```diff
@@ -147,15 +147,8 @@ class MDnsClient {
...
-      socket.close();
-    }
-    _sockets.clear();
-
-    for (final RawDatagramSocket socket in _toBeClosed) {
-      socket.close();
-    }
-    _toBeClosed.clear();
+    _incomingIPv4?.close();
+    _incomingIPv6?.close();
```
cbracken, line 152: Set to null after closing, in case someone calls `start` again.

biagiopietro: `_incomingIPv4` and `_incomingIPv6` ?

packages/multicast_dns/test/client_test.dart
```diff
@@ -86,6 +86,66 @@ void main() {
...
+    final testCases = [
+      {
+        'name': 'IPv4',
+        'datagramSocketType': InternetAddress.anyIPv4,
+        'interfacePrefix': '2001:0db8:85a3:0000:0000:8a2e:7335:030'
+      },
+      {
+        'name': 'IPv6',
+        'datagramSocketType': InternetAddress.anyIPv6,
+        'interfacePrefix': '192.168.2.'
```
biagiopietro, line 100: I think I mess up IPs formats, I will switch them.

biagiopietro:
> Apologies for the delay reviewing this.
> 
> @vashworth kindly tested this out against the `flutter` CLI and confirmed it didn't break wireless debugging for `flutter run` or `flutter attach`.
> 
> This change generally LGTM, though I would like a second opinion from @cbracken who might remember more about why the original code was written that way.
> 
>     1. This needs a pubspec version bump and CHANGELOG update.  See https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version-and-changelog-updates
> 
>     2. Can you add a test to https://github.com/flutter/packages/blob/main/packages/multicast_dns/test/client_test.dart?

@jmagman is it the following command enough for accomplishing point 1? sorry if it's sounds like a stupid question but I ran it locally and it changed `changelog` and `pubspec` of almost all the packages.

```shell
dart run script/tool/bin/flutter_plugin_tools.dart update-release-info \
  --version=bugfix \
  --base-branch=fix-issue-79772 \
  --changelog="Optimized Socket Binding: Always bind to 0.0.0.0 for simplicity and efficiency."
```

jmagman:
> is it the following command enough for accomplishing point 1? sorry if it's sounds like a stupid question but I ran it locally and it changed `changelog` and `pubspec` of almost all the packages.
> 
> ```shell
> dart run script/tool/bin/flutter_plugin_tools.dart update-release-info \
>   --version=bugfix \
>   --base-branch=fix-issue-79772 \
>   --changelog="Optimized Socket Binding: Always bind to 0.0.0.0 for simplicity and efficiency."
> ```

You could edit the `pubspec.yaml` and `CHANGELOG.md` directly.

cc @stuartmorgan re: the `update-release-info` command, is something missing from the docs? https://github.com/flutter/packages/blob/main/script/tool/README.md#update-changelog-and-version

stuartmorgan:
> cc @stuartmorgan re: the `update-release-info` command, is something missing from the docs? https://github.com/flutter/packages/blob/main/script/tool/README.md#update-changelog-and-version

The docs say to use `upstream/main` as the base branch in general.

The command run above used this PR's branch as the base branch, which means it will find no changes. The repo tooling always interprets no changes as "assume everything changed" as a failsafe to prevent CI from silently running no tests if something goes wrong with diff computation. (Even if it didn't, that command still wouldn't have update the package, it would have just done nothing.)

biagiopietro:
Thanks for the explanation @stuartmorgan / @jmagman , 👇 did the trick:
```shell
➜  multicast_dns git:(fix-issue-79772) dart run ../../script/tool/bin/flutter_plugin_tools.dart update-release-info --current-package \
  --version=bugfix \
  --changelog="Optimized Socket Binding: Always bind to 0.0.0.0 for simplicity and efficiency."
```

biagiopietro:
finally all checks are green 😅

cbracken:
Taking another look! Thanks!

packages/multicast_dns/lib/multicast_dns.dart
```diff
@@ -219,11 +190,12 @@ class MDnsClient {
     final Stream<T> results = _resolver.addPendingRequest<T>(
         query.resourceRecordType, query.fullyQualifiedName, timeout);
 
-    // Send the request on all interfaces.
-    final List<int> packet = query.encode();
-    for (final RawDatagramSocket socket in _sockets) {
-      socket.send(packet, _mDnsAddress!, selectedMDnsPort);
+    if (_incomingIPv4 == null) {
+      throw StateError('Unable to send packet on null socket.');
```
cbracken, line 194: Won't this cause an exception to be thrown if the connection is over IPv6? 

For IPv6 in the old code, we'd just have no sockets to send over and not send anything, but now we'll throw an exception. I think you can just drop this check and everything should be fine, since you're using a null-safe dereference on the `send` call below.

cbracken (changes requested):
Just one comment, other than that looks good.

It may be worth adding an equivalent test for IPv6.

packages/multicast_dns/lib/multicast_dns.dart
cbracken: I'm curious why we didn't use equivalent logic for the IPv6 approach -- i.e. we never sent on the IPv6 equivalent of 0.0.0.0, only on the sockets bound to the individual network interfaces. I'll do a little digging to see if I can understand why.

biagiopietro: > Won't this cause an exception to be thrown if the connection is over IPv6?

correct!

> I think you can just drop this check and everything should be fine

Removed. I did not go deeper on the `IPV6` topic, I saw other MDNs packages (like [Hashicorp multicast mdns package](https://github.com/hashicorp/mdns/blob/e208a109b48c6cd841139c6913bddf0d7bb27a51/client.go#L184)) that they handle also `multicast` with `IPV6` but I did not want to go deeper in this PR. Maybe it's worth to open another issue for implementing it (or have a look if it's doable or not), what do you think? 🤔

biagiopietro: I guess this thread can be closed, right?

jmagman:
> Just one comment, other than that looks good.

@cbracken could you take another look at this when you get a chance?

packages/multicast_dns/lib/multicast_dns.dart
cbracken: I'm curious if you've had a chance to test this with IPv6. My worry is that the existing code was broken for IPv4 (and this fixes it), but this patch effectively removes IPv6 support altogether. That seems fine if it was previously broken, but if not, it's a regression for those users.

Other that phone users connecting over a cell connection, I suspect people with an IPv6-only connection are pretty few and far between so possibly not the end of the world, but still a regression.

cbracken: On balance, landing this is probably still a net win, but probably worth at least documenting the change.

cbracken: Thinking about it a little more I think there are two options here:
1. We add back the `_sockets` List (maybe as `_ipv6InterfaceSockets`) and the code that configures them as multicast -- but only when an `InternetAddress.anyIPv6` is passed to the constructor, and then the `send` code at the bottom similarly would either send on `_incomingIPv4` for IPv4, or the collection of IPv6 sockets for IPv6.
2. We update the constructor to specifically throw (maybe UnimplementedError) if `InternetAddress.anyIPv6` is passed in and document that IPv6 support has been removed, in the class/package docs.

Sorry to spot this at this point.

cbracken (commented):
Overall this looks great for IPv4. We'll want to either document the removal of IPv6 support (if the previous code didn't exist) or add it back (if it did).

packages/multicast_dns/lib/multicast_dns.dart
biagiopietro: Thanks again for the time spent on the review.

> Sorry to spot this at this point.

No need to say sorry, I'm just a guest here 😅 

I applied the changes, I put back `IPv6` logic and I updated tests accordingly.

(at this point, it's does not make sense to remove `IPv6` support.)

cbracken (approved):
Apologies for the delay while I was out on vacation and thank you so much for your patience in landing this improvement! Looks great to me.

jtmcdole:
Just a drive by comment; thanks for looking into this problem!

First observation: `0.0.0.0` is **not** a multicast address; its a signal to bind to any local address (unicast). The reuseaddr/reuseport I believe will not _share_ data with other listeners and its up to the OS to deliver the packets to {random|first} listener.

In the wireshark captures:

1. It looks like we are successfully sending out a QM - query multicast; the response should be made to the multicast address 
2. The responses look **unicast** and probably only delivered to the first listening socket.

```
192.168.2.7	192.168.2.16	MDNS		   180	Standard query response 0x0000 PTR mdnsserv._test._tcp.local SRV 10 1 8080 mdnsserv A 127.0.1.1 TXT
192.168.2.7	192.168.2.16	MDNS		   180	Standard query response 0x0000 PTR mdnsserv._test._tcp.local SRV 10 1 8080 mdnsserv A 127.0.1.1 TXT
```

Are there any other clients listening on your system (e.g. avahi)?

For direct-unicode messages, I believe it is OK to use a random listening port. Your first attempt looked fine. I believe this falls under 5.1 for one-shot multicast DNS queries and specifically called out:

> These queries are typically done using a high-numbered ephemeral UDP source port, but regardless of whether they are sent from a dynamic port or from a fixed port, these queries MUST NOT be sent using UDP source port 5353, since using UDP source port 5353 signals the presence of a fully compliant Multicast DNS querier, as described below.

Let me dig a little deeper?

jtmcdole:
caveat emptor: I'm not an mdns expert. 

> The class states that it's "one shot" but it has _never_ been so since the port is always 5353.

I missed the `incoming.joinMulticast()` at the bottom of the interface loop - so you're right on that point. We should update our documentation on that point. For it to be [5.1 One-Shot Multicast DNS Queries](https://datatracker.ietf.org/doc/html/rfc6762#section-5.1), we would be listening to a unicast only socket and send QU queries.

What's interesting and might just be a quirk of the mdns server - we get non-multicast responses back in your wireshark logs. Or maybe wireshark is just not showing the destination multicast address on received packets (I'm also not a wireshark expert)?

> I don't feel this should be changed in this PR (I would rather add the option to be "one shot" mode and keep as default what we have now 🤷‍♂️ ).

**Sounds good to me.** the rest of the PR is LGTM.

> Note: could you please point me to the comment where I state that 0.0.0.0 is a multicast address? It's an old PR 😅

A misreading of your description and our documentation.

jtmcdole (approved):

PR #6714, opened 2024-05-12
[go_router] New feature improve debug full path

hashirshoaeb:
This PR fixes https://github.com/flutter/flutter/issues/148121

- Replaced `=>` with `| `, `├─` and `└─ ` to improve readability
- It prints the Widget name for easy referencing
- Shell routes does not have their own paths for it is presented as ` (Shell route)` in the tree
- Prints the widget name of the routes it is building.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/go_router/lib/src/configuration.dart
+++ b/packages/go_router/lib/src/configuration.dart
@@ -550,7 +550,7 @@ class RouteConfiguration {
         sb.writeln('$decoration$path ($screenName)'
             '${route.name == null ? '' : ' (${route.name})'}');
       } else if (route is ShellRouteBase) {
-        sb.writeln('$decoration$parentFullpath (ShellRoute)');
+        sb.writeln('$decoration (ShellRoute)');
       }
       _debugFullPathsFor(route.routes, path, decoration, sb);
     }
--- a/packages/go_router/lib/src/configuration.dart
+++ b/packages/go_router/lib/src/configuration.dart
@@ -394,14 +394,13 @@ class RouteConfiguration {
           return prevMatchList;
         }
 
-        final List<RouteMatch> routeMatches = <RouteMatch>[];
+        final List<RouteMatchBase> routeMatches = <RouteMatchBase>[];
         prevMatchList.visitRouteMatches((RouteMatchBase match) {
-          if (match is RouteMatch) {
+          if (match.route.redirect != null) {
             routeMatches.add(match);
           }
           return true;
         });
-
         final FutureOr<String?> routeLevelRedirectResult =
             _getRouteLevelRedirect(context, prevMatchList, routeMatches, 0);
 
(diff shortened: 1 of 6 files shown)


packages/go_router/lib/src/configuration.dart
```diff
@@ -538,15 +538,35 @@ class RouteConfiguration {
...
+      final String decoration =
+          _getDecoration(parentDecoration, index, routes.length);
+      String path = parentFullpath;
       if (route is GoRoute) {
-        final String fullPath = concatenatePaths(parentFullpath, route.path);
-        sb.writeln('  => ${''.padLeft(depth * 2)}$fullPath');
-        _debugFullPathsFor(route.routes, fullPath, depth + 1, sb);
+        path = concatenatePaths(parentFullpath, route.path);
+        final String screenName =
+            route.builder?.runtimeType.toString().split('=> ').last ?? '';
```
chunhtai, line 549: Wouldn't this always be `Widget`?

maybe just use `GoRoute` and `GoRoute(redirectOnly)`

chunhtai (commented):
This is so so much better than the previous, Thanks a lot!!

packages/go_router/lib/src/configuration.dart
hashirshoaeb: screenName is the name of the widget. it will not always print `Widget`. Otherwise it was for no use. 
Just like in the `logging_test.dart` test it logs `/ (Text)`

Example from my app:
```
[GoRouter] Full paths for routes:
           ├─/ (SplashScreen)
           ├─/intro (IntroScreen)
           ├─ (ShellRoute)
           │ ├─/main/year ()
           │ │ └─/main/year/month (HomeScreen)
           │ └─/main/profile (ProfileScreen)
           │   ├─/main/profile/analytics (AnalyticsScreen)
           │   └─/main/profile/settings (SettingsScreen)
           │     └─/main/profile/settings/license (LicenseAgreementScreen)
           └─/journal (JournalsListScreen)
             ├─/journal/new (JournalCreateScreen)
             └─/journal/:id (JournalDetailScreen)
               └─/journal/:id/edit (JournalEditScreen)
``` 

There are edge cases: 
- screenName is only get printed, when user uses `GoRouter.builder`
- It does not print for `GoRouter.pageBuilder`
- It prints `(Never)` when `GoRouter.builder` throws `Unimplemented Error`

hashirshoaeb:
instead of using magic strings 
```dart
    final String newDecoration =
        parentDecoration.replaceAll('├─', '│ ').replaceAll('└─', '  ');
```

we can treat `String` as `list on enum`
```dart
    final newDecoration = parentDecoration.map((e) {
      switch (e) {
        case DecorType.branch:
          return DecorType.parentBranch;
        case DecorType.leaf:
          return DecorType.none;
        default:
          return e;
      }
    });
```
where DecorType is 
```dart
enum DecorType {
  parentBranch('| '),
  branch('├─'),
  leaf('└─'),
  none('  '),
  ;

  final String value;
  const DecorType(this.value);

  @override
  String toString() => value;
}
```

@chunhtai What are your thoughts, is it worth changing?

chunhtai:
yup that sounds better

chunhtai (approved):
LGTM

hashirshoaeb:
Just checking. If anything is pending on my side, just let me know. @chunhtai @hangyujin. Looking forward to the merge. 🙈

hangyujin:
LGTM, this looks very helpful!

hangyujin (approved):
LGTM

PR #6733, opened 2024-05-15
[interactive_media_ads] Adds initial Android implementation

bparrishMines:
Android implementation for https://github.com/flutter/flutter/issues/134228

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/interactive_media_ads/android/build.gradle
```diff
@@ -51,10 +51,16 @@ android {
...
+        implementation 'com.google.ads.interactivemedia.v3:interactivemedia:3.33.0'
         testImplementation 'junit:junit:4.13.2'
         testImplementation 'org.jetbrains.kotlin:kotlin-test'
+        testImplementation "org.mockito.kotlin:mockito-kotlin:4.1.0"
         testImplementation 'org.mockito:mockito-inline:5.1.0'
         testImplementation 'androidx.test:core:1.3.0'
+
+        // org.jetbrains.kotlin:kotlin-bom artifact purpose is to align kotlin stdlib and related code versions.
+        // See: https://youtrack.jetbrains.com/issue/KT-55297/kotlin-stdlib-should-declare-constraints-on-kotlin-stdlib-jdk8-and-kotlin-stdlib-jdk7
+        implementation(platform("org.jetbrains.kotlin:kotlin-bom:1.8.10"))
```
bparrishMines, line 63: This is the same line added to `camera_android_camerax`:  https://github.com/flutter/packages/blob/main/packages/camera/camera_android_camerax/android/build.gradle#L77

packages/interactive_media_ads/android/src/main/kotlin/dev/flutter/packages/interactive_media_ads/ProxyApiRegistrar.kt
```diff
@@ -0,0 +1,124 @@
...
+import io.flutter.plugin.common.BinaryMessenger
+
+open class ProxyApiRegistrar(binaryMessenger: BinaryMessenger, var context: Context) :
+    PigeonProxyApiRegistrar(binaryMessenger) {
+
+  // Added to be overriden for tests. The test implementation calls `callback` immediately, instead
+  // of waiting for the main thread to run it.
+  internal open fun runOnMainThread(callback: Runnable) {
+    Handler(Looper.getMainLooper()).post { callback.run() }
+  }
```
bparrishMines, line 19: This was added to the registrar to make it easier to handle and test callback methods on the correct thread.

packages/interactive_media_ads/android/src/main/kotlin/dev/flutter/packages/interactive_media_ads/VideoAdPlayerProxyApi.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  override fun setVolume(pigeon_instance: VideoAdPlayer, value: Long) {
+    (pigeon_instance as VideoAdPlayerImpl).savedVolume = value.toInt()
+  }
+
+  override fun setAdProgress(pigeon_instance: VideoAdPlayer, progress: VideoProgressUpdate) {
+    (pigeon_instance as VideoAdPlayerImpl).savedAdProgress = progress
+  }
```
bparrishMines, line 66: Custom methods added since `VideoAdPlayer.getVolume` and `VideoAdPlayer.getAdProgress` require a synchronous callback method.

packages/interactive_media_ads/lib/src/android/android_ad_display_container.dart
```diff
@@ -0,0 +1,241 @@
...
+    );
+  }
+
+  // Stops updating the IMA SDK the progress of the currently playing ad.
+  void _stopAdProgressTracking() {
+    _adProgressTimer?.cancel();
+    _adProgressTimer = null;
+  }
+
+  static ima.VideoView _setUpVideoView(
```
bparrishMines, line 161: I started a new style of setting callback methods, which can cause a circular reference, in a static method. I would forget the rule of not using non-local methods in callback methods and this help prevented me from accidentally calling a method/field in the encapsulating class.

packages/interactive_media_ads/android/src/main/kotlin/dev/flutter/packages/interactive_media_ads/AdErrorProxyApi.kt
```diff
@@ -0,0 +1,54 @@
...
+      AdError.AdErrorCode.COMPANION_AD_LOADING_FAILED -> AdErrorCode.COMPANION_AD_LOADING_FAILED
+      AdError.AdErrorCode.UNKNOWN_ERROR -> AdErrorCode.UNKNOWN_ERROR
+      AdError.AdErrorCode.VAST_EMPTY_RESPONSE -> AdErrorCode.VAST_EMPTY_RESPONSE
+      AdError.AdErrorCode.FAILED_TO_REQUEST_ADS -> AdErrorCode.FAILED_TO_REQUEST_ADS
+      AdError.AdErrorCode.VAST_ASSET_NOT_FOUND -> AdErrorCode.VAST_ASSET_NOT_FOUND
+      AdError.AdErrorCode.ADS_REQUEST_NETWORK_ERROR -> AdErrorCode.ADS_REQUEST_NETWORK_ERROR
+      AdError.AdErrorCode.INVALID_ARGUMENTS -> AdErrorCode.INVALID_ARGUMENTS
+      AdError.AdErrorCode.PLAYLIST_NO_CONTENT_TRACKING -> AdErrorCode.PLAYLIST_NO_CONTENT_TRACKING
+      AdError.AdErrorCode.UNEXPECTED_ADS_LOADED_EVENT -> AdErrorCode.UNEXPECTED_ADS_LOADED_EVENT
+    }
```
stuartmorgan, line 36: It looks like this doesn't handle the case where a new error code is added upstream. It seems like it would probably even fail to compile if that happened?

packages/interactive_media_ads/android/src/main/kotlin/dev/flutter/packages/interactive_media_ads/BaseDisplayContainerProxyApi.kt
```diff
@@ -0,0 +1,8 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package dev.flutter.packages.interactive_media_ads
+
+class BaseDisplayContainerProxyApi(override val pigeonRegistrar: ProxyApiRegistrar) :
```
stuartmorgan, line 7: Shouldn't all of the classes have class-level comments? https://developer.android.com/kotlin/style-guide#documentation

packages/interactive_media_ads/android/src/main/kotlin/dev/flutter/packages/interactive_media_ads/VideoAdPlayerProxyApi.kt
stuartmorgan: I think it's worth adding this as a code comment to these methods. It would be good to have it clarify why it's not calling `api.foo` in addition to saving the value.

packages/interactive_media_ads/lib/interactive_media_ads.dart
```diff
@@ -5,6 +5,15 @@
 export 'src/ad_display_container.dart';
 export 'src/ads_loader.dart';
 export 'src/ads_manager_delegate.dart';
-export 'src/platform_interface/ad_error.dart';
-export 'src/platform_interface/ad_event.dart';
-export 'src/platform_interface/ads_request.dart';
+export 'src/android/android_interactive_media_ads.dart'
+    show AndroidInteractiveMediaAds;
+export 'src/platform_interface/platform_interface.dart'
+    show
```
stuartmorgan, line 11: In some cases in the past we've regretted directly re-exposing platform interface APIs in the app-facing API (instead of having a second version and converting); have you audited these to make sure we're very unlikely to ever want to make any breaking changes to them, such as adding a required constructor param, or in having some method that implementations need but clients should never use?

packages/interactive_media_ads/lib/src/android/android_ad_display_container.dart
```diff
@@ -0,0 +1,241 @@
...
+    final WeakReference<AndroidAdDisplayContainer> weakThis =
+        WeakReference<AndroidAdDisplayContainer>(this);
+    _videoView = _setUpVideoView(weakThis);
+    _frameLayout.addView(_videoView);
+    _videoAdPlayer = _setUpVideoAdPlayer(weakThis);
+  }
+
+  // The duration between each update to the IMA SDK of the progress of the
+  // currently playing ad.
+  static const int _progressPollingMs = 250;
```
stuartmorgan, line 62: Could you explain in the comment where/how this number was derived? I.e., is it arbitrary, based on best practice docs, other?

packages/interactive_media_ads/lib/src/android/android_ad_display_container.dart
```diff
@@ -0,0 +1,241 @@
...
+
+  // ViewGroup used to create the `ima.AdDisplayContainer`.
+  late final ima.FrameLayout _frameLayout =
+      _androidParams._imaProxy.newFrameLayout();
+
+  // Handles ad playback.
```
stuartmorgan, line 68: Could you elaborate on the comments here and/or for `_videoAdPlayer`? After reading both comments it's not really clear to me what the responsibilities of each are.

packages/interactive_media_ads/lib/src/android/android_ad_display_container.dart
```diff
@@ -0,0 +1,241 @@
...
+
+  /// Callbacks that update the state of ad playback.
+  @internal
+  final Set<ima.VideoAdPlayerCallback> videoAdPlayerCallbacks =
+      <ima.VideoAdPlayerCallback>{};
+
+  // Handles ad playback callbacks from the IMA SDK.
+  late final ima.VideoAdPlayer _videoAdPlayer;
+
+  /// The native Android AdDisplayContainer.
```
stuartmorgan, line 80: Maybe explain the difference between a display container and a view (`_videoView`) here?

packages/interactive_media_ads/lib/src/android/android_ad_display_container.dart
```diff
@@ -0,0 +1,241 @@
...
+            .createAdDisplayContainerImaSdkFactory(
+          _frameLayout,
+          _videoAdPlayer,
+        );
+        params.onContainerAdded(this);
+      },
+    );
+  }
+
+  void _resetPlayer() {
```
stuartmorgan, line 120: This could use a comment about intended use, since it only resets a subset of the state.

packages/interactive_media_ads/lib/src/android/android_ad_display_container.dart
```diff
@@ -0,0 +1,241 @@
...
+                  _loadedAdMediaInfo!,
+                  currentProgress,
+                );
+              },
+            ),
+          ],
+        );
+      },
+    );
+  }
```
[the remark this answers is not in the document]
stuartmorgan: Maybe add this reference to the method comment?

packages/interactive_media_ads/lib/src/android/android_ad_display_container.dart
stuartmorgan: It's probably worth briefly commenting this pattern in the code each time it's used so that future people understand the benefits and aren't tempted to "simplify" the code by removing this (or at least not replicating it in new code).

packages/interactive_media_ads/lib/src/android/interactive_media_ads_proxy.dart
```diff
@@ -0,0 +1,130 @@
...
+  final VideoView Function({
+    required void Function(
+      VideoView,
+      MediaPlayer,
+      int,
+      int,
+    ) onError,
+    void Function(
+      VideoView,
+      MediaPlayer,
```
stuartmorgan, line 52: Wouldn't some of these be more readable without the trailing comma? Or would the inconsistency hurt readability?

packages/interactive_media_ads/pigeons/interactive_media_ads_android.dart
```diff
@@ -0,0 +1,717 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// // ignore_for_file: avoid_unused_constructor_parameters
```
stuartmorgan, line 5: Why is everything in this file commented out?

stuartmorgan (commented):
(I only briefly skimmed the tests

packages/interactive_media_ads/android/src/main/kotlin/dev/flutter/packages/interactive_media_ads/GeneratedInteractiveMediaAdsLibrary.kt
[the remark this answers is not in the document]
bparrishMines: Good catch. I forgot that we only did that because of the limitations of Java and Obj-c.

packages/interactive_media_ads/android/src/main/kotlin/dev/flutter/packages/interactive_media_ads/AdErrorProxyApi.kt
bparrishMines: My intention was for this to break compilation when a new enum is added. Since this uses a pinned version of the IMA SDK, we can add a new enum to the code generation whenever we bump the version.

We add an `unknown` enum in the `webview_flutter` implementations, but I assumed we did that because it uses a system library and not a pinned SDK.

packages/interactive_media_ads/lib/src/android/android_ad_display_container.dart
bparrishMines: It was the same value as the one in the android example: https://developers.google.com/interactive-media-ads/docs/sdks/android/client-side#6.-create-the-videoadplayeradapter-class. It doesn't give an explanation, but I added a comment to explain where the value is from.

packages/interactive_media_ads/lib/src/android/interactive_media_ads_proxy.dart
bparrishMines: Yea, it does make it more readable. I removed the trailing commas

packages/interactive_media_ads/pigeons/interactive_media_ads_android.dart
bparrishMines: This uses features from https://github.com/flutter/packages/pull/6371, so it would break tests without it commented out. But I'll go ahead and uncomment it for now so it can be reviewed while I still wait for https://github.com/flutter/packages/pull/6371 to land.

packages/interactive_media_ads/lib/interactive_media_ads.dart
bparrishMines: Does the fact that the platform interface, the implementations, and the app-facing API live in the same package affect this? This should ideally give us more flexibility to make private methods/constructors/subclasses if we need to.

packages/interactive_media_ads/android/src/main/kotlin/dev/flutter/packages/interactive_media_ads/AdErrorProxyApi.kt
stuartmorgan: > Since this uses a pinned version of the IMA SDK

It doesn't really though, because of [optimistic upgrades](https://docs.gradle.org/current/userguide/resolution_strategy_tuning.html#fail-version-conflict). Because of the way Gradle works, unless we want to make the entire build fail on conflicts (which would be a large hammer) we don't actually have any control over whether we get the version listed in build.gradle.

packages/interactive_media_ads/lib/interactive_media_ads.dart
stuartmorgan: I guess we could always move these classes to the app-facing directory and make a new class that the platform interface code used, if we need to make breaking changes later, since changing the platform interface wouldn't be public in this structure.

stuartmorgan (approved):
LGTM once the remaining comment about the enum fallback is resolved.

reidbaker:
This is on my review queue. Out of curiosity will the android team maintain the android implementation of this plugin like we do for some others or will this be owned by the ecosystem team?

stuartmorgan:
For now it'll be ecosystem; medium term we can discuss. Longer term, the goal is to hand off to the SDK owners.

bparrishMines:
@reidbaker I'm submitting this now since I have the LGTM from @stuartmorgan. But still feel free to review or skim when you are able to.

reidbaker:
Can you give me another day?

bparrishMines:
Ah, sorry. The autosubmit bot merged it quick after. The plugin is still in alpha/beta, so we can still make breaking changes after this publishes.

PR #6740, opened 2024-05-15
[file_selector_ios, image_picker_ios] Remove Swift Package Support

vashworth:
Plugins converted to Swift Package Manager that use a modulemap seem to be having issues with there being a `module.modulemap`. As a temporary solution to unblock people, we'll remove the `module.modulemap` and Swift Package support (`Package.swift`). 

Temporary solution for https://github.com/flutter/flutter/issues/148307.

## Pre-launch Checklist

- [x] I signed the [CLA].


loic-sharma (approved):
LGTM

stuartmorgan:
Do we have any idea yet why we don't see this in CI? We can land this as is, but it would be nice to have a test that we can ensure we'd don't regress later when re-landing support.

packages/file_selector/file_selector_ios/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.2+1
+
+* Removes Swift Package Manager compatibility.
```
stuartmorgan, line 3: How about "Temporarily remove Swift Package Manager compatibility to resolve issues with Cocoapods builds."?

vashworth:
> Do we have any idea yet why we don't see this in CI? We can land this as is, but it would be nice to have a test that we can ensure we'd don't regress later when re-landing support.

So far, seems to be older versions of Xcode are having this issue and we don't enforce Xcode 15 yet. So unfortunately impossible to test in CI since macOS 13 does not support Xcode 14 or lower.

stuartmorgan:
Have we actually validated that it's an Xcode 14 vs 15 thing? I thought that was still speculative based on the issue discussion (i.e., nobody seems to have indicated confirming that the same project works in 15 but not 14).

vashworth:
> Have we actually validated that it's an Xcode 14 vs 15 thing? I thought that was still speculative based on the issue discussion (i.e., nobody seems to have indicated confirming that the same project works in 15 but not 14).

I confirmed Xcode 14 breaks: https://ci.chromium.org/ui/p/flutter/builders/try.shadow/Mac_x64%20ios_build_all_packages%20master/484/overview

jmagman (approved):

vashworth:
> I think we can get rid of the manual modulemaps, that was a lot of work to `explicit module Test` working. We could just do the cheap thing and redeclare the methods we need in the tests.

Unfortunately, we're not the only ones that do this, though. I've seen other plugin authors copy our model to achieve the same. I would think we would want to provide a way for them to maintain it, but yeah the modulemaps are super annoying and I wish we could get rid of them. I'll think on it more.

Example: https://github.com/Baseflow/flutter-geolocator/blob/ca5469fe4212fe82d586bed93634d5e5a0e6154a/geolocator_apple/ios/Classes/GeolocatorPlugin.modulemap#L24

stuartmorgan:
Have we considered cherry picking an Xcode 15 requirement to 3.22?

vashworth:
> Have we considered cherry picking an Xcode 15 requirement to 3.22?

We could, but this issue isn't the only problem we're having with the double modulemaps. @loic-sharma also ran into an issue when trying to convert `camera_avfoundation`: https://gist.github.com/loic-sharma/27bd8e3ea0d9ab16cd4b8de03918ec4d

So I think we need to figure out a better way to handle modulemaps

PR #6781, opened 2024-05-22
[rfw] Adds support for `DecorationImage.filterQuality`.

goderbauer:
And hard-codes filter qualities in tests to prepare for https://github.com/flutter/flutter/pull/148799, which switches the default to `medium`. This hard-codes a filter quality in the test so the test passes no matter what the default is.


Hixie (approved):
LGTM

PR #6782, opened 2024-05-23
[ci] Start moving podspec check to ARM

stuartmorgan:
Now that `google_maps_flutter` supports ARM simulators, we should be able to move this task to ARM. This adds the ARM version in bringup mode, with a new name to reflect the fact that it is doing more that just checking podspecs now.

Once this propagates and is green, the x64 version will be removed.

Part of https://github.com/flutter/flutter/issues/141493


jmagman (approved):
LGTM!

PR #6792, opened 2024-05-24
[webview_flutter_web] Migrate to package:web.

ThomasAunvik:
* Migrates to `package:web` https://github.com/flutter/flutter/issues/139749
* Updates `HttpRequestFactory.request` to use `package:http` `BrowserClient`
* Updates `ìndex.html` in the example to use `flutter_bootstrap.js`
* Updates minimum dart sdk requirement to `^3.3.0`

Would appreciate help with completing the mock tests if in case it does not work.
(I am somehow stuck with 'loading...' when attempting to test with mockito with --platform chrome)

Integration tests from what i was able to test passes.

Migrated to using BrowserClient for web due to issues creating mock tests with `XMLHttpRequest` which is returned from `package:web`'s `HttpRequest.request` following error:

`Bad state: Interface type 'XMLHttpRequest' which is nether an enum, nor a class, nor a mixin. This case is unknown, please report a bug.`

## Pre-launch Checklist

- [x] I signed the [CLA].

Co-authored-by: David Iglesias <ditman@gmail.com>
Co-authored-by: Navaron Bracke <brackenavaron@gmail.com>


ThomasAunvik:
The tests that seem to be failing now are just the ones in the legacy/ folder, that we require to create a mock `HTMLIFrameElement`.

And these two calls that i do not understand what it is trying to do.
Based on the lsp, argThat just returns null and mockElement.src cannot be set to a null, can only be set to String.

https://github.com/flutter/packages/blob/a598bef16e6f00db6cc12f62c758c70f7b452607/packages/webview_flutter/webview_flutter_web/test/legacy/webview_flutter_web_test.dart#L84
https://github.com/flutter/packages/blob/a598bef16e6f00db6cc12f62c758c70f7b452607/packages/webview_flutter/webview_flutter_web/test/legacy/webview_flutter_web_test.dart#L178

Two legacy tests N-1 and N-2 are failing due to the `package:web` having a higher dart sdk requirement than the test has installed.
currently web version is set to version `^0.5.1`, it suggests it to be set to `^0.3.0`. Reading the changelog of the package, i don't see any breaking changes on it. So i assume it should be fine to lower the versioning?

ditman:
Thanks for the PR @ThomasAunvik! I'll take a look!

(Please take into account that next monday is a holiday in the US, so I won't be able to get to this until next tuesday at the earliest)

uldall:
I have been experimenting with this PR for a few days, and run into the issue that the created <iframe> doesn't have the `credentialless` attribute set, and so won't load if the iframe'ed document doesn't have the proper headers when being GET'ed. See:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/credentialless and https://developer.mozilla.org/en-US/docs/Web/Security/IFrame_credentialless

I think it would make sense to make it possible to specify whether this attribute should be set from Flutter code?

ThomasAunvik:
> https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/credentialless and https://developer.mozilla.org/en-US/docs/Web/Security/IFrame_credentialless

Looking at the compatability, firefox and safari is not yet supported, but an option to enable it would be beneficial once it is fully supported.

Otherwise the current solution for all browsers is that the flutter site has the `Cross-Origin-Embedder-Policy: credentialless` header set.

uldall:
> Otherwise the current solution for all browsers is that the hosted webapp has the `Cross-Origin-Embedder-Policy: credentialless` header set.

In most cases people don't own the sites that they iframe, and it can be very hard to convince a third party to add support for these headers.

> Looking at the compatability, firefox and safari is not yet supported, but an option to enable it would be beneficial once it is fully supported.

I don't see why we would wait. The WASM builds are already only supported in Chromium based browsers. But I also understand if you don't want to add it in this PR :)

ThomasAunvik:
> In most cases people don't own the sites that they iframe, and it can be very hard to convince a third party to add support for these headers.

Nevermind i got confused on how the cors weirdness works, as i did pain with it too. It would work on setting credentialless on the iframe.

It can be easily done by adding this right after `loadRequest`.
```
_webWebViewParams.iFrame.setAttribute('credentialless', 'true');
```
As the current `HTMLIFrameElement` doesn't have it implemented directly yet.

Would require me to update the `LoadRequestParams` to include a boolean for `credentialless`.

ThomasAunvik:
Here is a proposal on how setting the `credentialless` would be.
This is added under `WebWebViewController`

```dart
/// Sets the IFrame Credentialless
/// https://developer.mozilla.org/en-US/docs/Web/Security/IFrame_credentialless
void setIFrameCredentialless(bool flag) {
  _webWebViewParams.iFrame
      .setAttribute('credentialless', flag ? 'true' : 'false');
}
```

And to set it would be as such:
```dart
if(WebViewPlatform.instance is WebWebViewPlatform) {
    WebWebViewController(
        const PlatformWebViewControllerCreationParams(),
    )
    ..setIFrameCredentialless(true)
    ..loadRequest(
        LoadRequestParams(
            uri: Uri.parse('https://flutter.dev'),
        ),
    ))
}
```

`setIFrameCredentialless` is  would

uldall:
It seems that on other platforms these kind of platform specific parameters are set on the `WebViewControllerCreationParams` implementation. 

An example from the [Webkit impl](https://github.com/ThomasAunvik/flutter-packages/blob/0870dc84d364aad79fba7a57ef2984b01b073c4d/packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart#L52). Here some boolean parameters like `allowsInlineMediaPlayback` and `limitsNavigationsToAppBoundDomains` are set in the `WebKitWebViewControllerCreationParams`. Seems similar to our `credentialless` use case.

I am by no means well versed in the Flutter codebase, but this was just my best effort on some feedback. I will be going on vacation for a week, but will test the new branch when I am back :)

kevmoo:
What's the status here, friends?

uldall:
I managed to test the credentialless branch just now, and it works as expected and the contents of the iframe is loaded:
image

kevmoo:
@ditman would you take a look?

uldall:
I have been testing the PR quite a bit, and created this issue because Godot games being loaded in the iframe resulted in an error: https://github.com/flutter/flutter/issues/150119

However since then I have also started getting the error on other sites (for example https://flutter.dev/). It does however work great on a payment site that I use. (reepay.com). I am not sure what properties of the iframe'ed site makes it throw the error.

Maybe the error is unrelated to this PR, and probably a bug in the skwasm renderer?

ditman:
This is next on my radar, after this lands:

* https://github.com/flutter/packages/pull/6981

ditman:
@ThomasAunvik I've finally had some time to review this; insetead of adding a lot of back and forth I just **force pushed** (please, remember to `git pull --rebase` on your checkout!) the changes I thought appropriate to your branch.

Do you mind verifying that the plugin still does what you **need**? Tests seem to be passing, and the example app seems to work as expected, but the more use cases, the merrier!

packages/webview_flutter/webview_flutter_web/lib/src/http_request_factory.dart
```diff
@@ -11,20 +15,16 @@ class HttpRequestFactory {
...
   ///
-  /// If specified, `sendData` will send data in the form of a [ByteBuffer],
-  /// [Blob], [Document], [String], or [FormData] along with the HttpRequest.
+  /// The Future is completed when the [web.Response] is available.
   ///
-  /// If specified, [responseType] sets the desired response format for the
-  /// request. By default it is [String], but can also be 'arraybuffer', 'blob',
-  /// 'document', 'json', or 'text'. See also [HttpRequest.responseType]
-  /// for more information.
+  /// If specified, `sendData` will be sent as the `body` of the fetch.
```
navaronbracke, line 27:
```suggestion
  /// If specified, [sendData] will be sent as the `body` of the fetch.
```

packages/webview_flutter/webview_flutter_web/lib/src/http_request_factory.dart
```diff
@@ -55,27 +55,33 @@ class HttpRequestFactory {
   ///         // Do something with the response.
   ///     });
   ///
-  /// Note that requests for file:// URIs are only supported by Chrome extensions
+  /// Requests for file:// URIs are only supported by Chrome extensions
```
navaronbracke, line 58:
```suggestion
  /// Requests for `file://` URIs are only supported by Chrome extensions
```

packages/webview_flutter/webview_flutter_web/lib/src/http_request_factory.dart
```diff
@@ -55,27 +55,33 @@ class HttpRequestFactory {
...
+      ...?requestHeaders,
+    };
+    return web.window
+        .fetch(
+            url.toJS,
+            web.RequestInit(
+              method: method,
+              body: sendData?.toJS,
+              credentials: withCredentials ? 'include' : 'same-origin',
+              headers: headers.jsify()! as web.HeadersInit,
```
navaronbracke, line 83: I don't think we have support for the JS Map type yet, so perhaps we should add a TODO for that?

In my opinion we can eventually get rid of the `jsify()` in the long run?

Not sure how @ditman feels about doing that, though?

ditman: (Maybe something with `_createObjectLiteral`? [source](https://github.com/dart-lang/sdk/blob/60465149414572c8ca189d8f65fdb39795c4b97d/sdk/lib/js_interop/js_interop.dart#L107-L111))

srujzs: We could add a better constructor here if that's easier to read, but it's not going to do anything more complicated than just `jsify` :) e.g.

`JSObject.fromMap(Map m) => m.jsify() as JSObject;`

ditman: @srujzs yeah, I'd like to have a `jsify` in non-nullable-object so I don't have to `Map.jsify()!` and maybe a `T jsifyAs<T>()` that does the casting for me, so I could replace:

```dart
headers.jsify()! as web.HeadersInit
```

by

```dart
headers.jsifyAs<web.HeadersInit>();
```

(Not that it saves too much, though)

ditman:
Just verified [in CI](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8742866746488585009/+/u/Run_package_tests/l_Wasm__create_all_packages_app/stdout) that this package enables the `webview_flutter` package to compile to Wasm! 🎉 

```
Exluding the following plugins from the combined build:
  plugin_platform_interface
  camera
  google_maps_flutter
```

ThomasAunvik:
Works as expected.

Experiencing an issue when re-opening the same pdf blob data, iframe doesn't render properly, showing nothing just the background of the parent widget), until resizing the window. (on wasm only).
I'll have to test it a bit more.

ThomasAunvik:
Created a reproducible example: https://github.com/ThomasAunvik/flutter_webview_wasm_error/commit/f75fddbdf3b805328a66a9a87e6274e75e6f0414

Seems like this time I managed to make the iframe not appear at all when using `Navigator.of(context).push()`

Left is wasm on chrome, right is non-wasm using firefox.

https://github.com/flutter/packages/assets/9968151/7e46caa3-7e86-4a76-a0e2-0c678bf1f2fd

Unless it is an issue with this package alone, i don't see it to be needing to not merge it with this issue, if that is also just a lower level flutter issue.

ditman:
@ThomasAunvik I think you've encountered something like this:

* https://github.com/flutter/flutter/issues/149909

Platform Views in Wasm have some positioning issues (it can also be seen with the default app, I think)

ditman (approved):
Been here, seen this!

ditman:
Let's try to autoland this!

kevmoo:
Rerunning failures and adding back auto-submit!

packages/webview_flutter/webview_flutter_web/example/web/index.html
```diff
@@ -3,6 +3,7 @@
 Use of this source code is governed by a BSD-style license that can be
 found in the LICENSE file. -->
 <html>
+
```
kevmoo, line 6:
```suggestion
```

PR #6793, opened 2024-05-24
Amend package:web tweaks to allow package:web roll

srujzs:
https://github.com/dart-lang/web/commit/7604578eb538c471d438608673c037121d95dba5 adds a number of APIs in order to get a more consistent API surface. This includes disablePictureInPicture and disableRemotePlayback, which both take in a bool and not a JSBoolean. However, it can't be rolled into google3 as this package will now be broken due to the way extension type methods take a higher precedence over extension methods.

This CL aligns those methods with the package:web equivalent so that usages of these methods can be consistent. controlsList is also amended to take in a String so that if it's ever added to the Web IDL, it doesn't conflict with the package:web definition that will be added then.


srujzs:
Hmm, the bot isn't happy with my versioning here. Should I be doing a 2.3.0+1/2.3.1 and update the pubspec as well?

ditman:
> Hmm, the bot isn't happy with my versioning here. Should I be doing a 2.3.0+1/2.3.1 and update the pubspec as well?

@srujzs Yes please, do it `2.3.1` (`2.3.0+1` is not a valid pub version number, ask me how I know :P)

As for the test, didn't we have overrides for these properties in the test code as well? (to make them "readable"?)

ditman (commented):
> Hmm, the bot isn't happy with my versioning here. Should I be doing a 2.3.0+1/2.3.1 and update the pubspec as well?

@srujzs Yes please, do it `2.3.1` (`2.3.0+1` is not a valid pub version number, ask me how I know :P)

As for the tests, please check here, where we do something similar but for the integration tests of the package (I can help you running those locally if needed):

* https://github.com/flutter/packages/blob/main/packages/video_player/video_player_web/example/integration_test/pkg_web_tweaks.dart

srujzs:
> Yes please, do it 2.3.1 (2.3.0+1 is not a valid pub version number, ask me how I know :P)

Done!

> As for the tests, please check here, where we do something similar but for the integration tests of the package

Ah I forgot that this repo has two such files to tweak `package:web`. Done.

srujzs:
Friendly ping. :)

ditman (approved):
Thanks for the fix, ship it!

PR #6839, opened 2024-05-30
Reland SurfaceProducer migration.

jonahwilliams:
Relands the migration with fixes for the camera preivew. WIP due to upsidedown portrait.


packages/camera/camera_android_camerax/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.7.0
```
jonahwilliams, line 1: I did a minor version bump so that I can constraint the new package/camera version to depend on this on newer, since the rotation logic will change between 0.6.5+5 and 0.7.0

packages/camera/camera_android_camerax/example/lib/camera_preview.dart
```diff
@@ -63,10 +63,10 @@ class CameraPreview extends StatelessWidget {
 
   int _getQuarterTurns() {
     final Map<DeviceOrientation, int> turns = <DeviceOrientation, int>{
-      DeviceOrientation.portraitUp: 0,
-      DeviceOrientation.landscapeRight: 1,
-      DeviceOrientation.portraitDown: 2,
```
jonahwilliams, line 68: I'm not sure what to do about portraitDown. How is it simulated?

camsim99: Do you have a Samsung devices? I can't see it on my pixel but I have seen it on a Galaxy IIRC

jonahwilliams:
from discussion in chat, we need to split this into two PRs

packages/camera/camera_android_camerax/example/lib/camera_preview.dart
camsim99: I tested two Galaxys and they did not rotate to portraitDown :') It could have possibly been a Samsung tablet and I can definitely help test out different devices in the office, but I'm not sure, sorry.

jonahwilliams: maybe it will be the same as portraitUp?

camsim99: Oh I assumed it would be 1 because it should face opposite of portraitUp?

camsim99: Meaning that the change has just shifted what is considered the natural orientation

jonahwilliams: makes sense!

camsim99:
Taking this over in a separate PR!

PR #6840, opened 2024-05-31
[pigeon] removes restriction on number of custom types per file

tarrinneal:
removes restriction on number of custom types per file (except for gobject).

also fixes bug in objc with multiple enums in a class

resolves a couple other minor issues:
fixes https://github.com/flutter/flutter/issues/150385
fixes https://github.com/flutter/flutter/issues/150108
work toward https://github.com/flutter/flutter/issues/152916


packages/pigeon/lib/java_generator.dart
```diff
@@ -504,6 +568,71 @@ class JavaGenerator extends StructuredGenerator<JavaOptions> {
...
+        root,
+        indent,
+        overflowClass,
+        dartPackageName: dartPackageName,
+      );
+
+      indent.format('''
+static @Nullable Object fromList(@NonNull ArrayList<Object> ${varNamePrefix}list) {
+  $_overflowClassName wrapper = new $_overflowClassName();
+  Object type = ${varNamePrefix}list.get(0);
```
stuartmorgan, line 603: Couldn't these two locals just be inlined? They don't seem to add much here.

packages/pigeon/lib/java_generator.dart
```diff
@@ -504,6 +568,71 @@ class JavaGenerator extends StructuredGenerator<JavaOptions> {
...
+
+      indent.writeScoped('@Nullable Object unwrap() {', '}', () {
+        indent.format('''
+if (wrapped == null) {
+  return null;
+}
+    ''');
+        indent.writeScoped('switch (type) {', '}', () {
+          for (int i = totalCustomCodecKeysAllowed; i < types.length; i++) {
+            indent.writeln('case (${i - totalCustomCodecKeysAllowed}):');
```
stuartmorgan, line 619: Why the parens around the case value?

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -27,6 +27,8 @@ const DocumentCommentSpecification _docCommentSpec =
 
 String _codecName = 'PigeonCodec';
 
+const String _overflowClassName = '${varNamePrefix}CodecOverflow';
```
stuartmorgan, line 30: In general for all the generators, I don't think we should be using `varNamePrefix`; this isn't a variable name. We should do something class-appropriate in each language for this name.

We could even just make up a new convention, like `PigeonCodecReserved` as the start of the class name. I would be okay with a world where if someone names a class `PigeonCodecReserved*` *and* happened to pick a value of `*` that collided with one of our handful of generated classes, it would fail.

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -369,6 +405,76 @@ class KotlinGenerator extends StructuredGenerator<KotlinOptions> {
...
+
+      indent.writeScoped('fun unwrap(): Any? {', '}', () {
+        indent.format('''
+if (wrapped == null) {
+  return null
+}
+    ''');
+        indent.writeScoped('when (type) {', '}', () {
+          for (int i = totalCustomCodecKeysAllowed; i < types.length; i++) {
+            indent.writeScoped('${i - totalCustomCodecKeysAllowed} -> {', '}',
```
stuartmorgan, line 462: Since all branches here are a single statement, I don't think the `{}` are necessary.

packages/pigeon/lib/objc_generator.dart
```diff
@@ -598,7 +624,8 @@ class ObjcSourceGenerator extends StructuredGenerator<ObjcOptions> {
             valueGetter,
             prefix: generatorOptions.prefix,
           );
-          ivarValueExpression = 'enumBox.value';
+          ivarValueExpression =
+              'an${_enumName(field.type.baseName, prefix: generatorOptions.prefix, box: true)}.value';
```
stuartmorgan, line 628: You've hard-code "an" in a few places in this codec; "an" only works if the class starts with a vowel sound. If you are going to hard-code something it should be "a".

packages/pigeon/lib/pigeon_lib.dart
```diff
@@ -843,7 +843,15 @@ class GObjectGeneratorAdapter implements GeneratorAdapter {
   }
 
   @override
-  List<Error> validate(PigeonOptions options, Root root) => <Error>[];
+  List<Error> validate(PigeonOptions options, Root root) {
+    final List<Error> errors = <Error>[];
+    if (root.classes.length + root.enums.length > totalCustomCodecKeysAllowed) {
+      errors.add(Error(
```
stuartmorgan, line 849: Please add a TODO with an issue reference.

packages/pigeon/lib/swift_generator.dart
```diff
@@ -135,33 +137,55 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
...
+      indent.writeln('case ${customType.enumeration}:');
+      indent.nest(1, () {
+        if (customType.type == CustomTypes.customEnum) {
+          indent.writeln('var enumResult: ${customType.name}? = nil');
+          indent.writeln(
+              'let enumResultAsInt: Int? = nilOrValue(self.readValue() as? Int)');
+          indent.writeScoped('if let enumResultAsInt = enumResultAsInt {', '}',
+              () {
+            indent.writeln(
+                'enumResult = ${customType.name}(rawValue: enumResultAsInt)');
```
stuartmorgan, line 153: Optional: it doesn't look like we need `enumResult` at all; we could `return` the RHS here, and `return nil` below. One line shorter per enum, and half as many boilerplate variables.

packages/pigeon/pigeons/core_tests.dart
```diff
@@ -12,6 +12,20 @@ enum AnEnum {
   fourHundredTwentyTwo,
 }
 
+enum AnotherEnum {
```
stuartmorgan, line 15: I think it's worth commenting that this is because enums often require extra logic in generators, and having two enums ensures that the logic can be replicated without collisions.

packages/pigeon/tool/shared/test_runner.dart
```diff
@@ -18,31 +18,67 @@ Future<void> runTests(
...
+  }
+
+  await _runTests(testsToRun, ciMode: ciMode);
+
+  if (overflow) {
+    await _runGenerate(baseDir, overflow: true);
+
+    await _runTests(testsToRun
+        .where((String test) =>
+            test.contains('integration') && !test.contains('linux'))
```
stuartmorgan, line 39: This needs a comment with the TODO reference.

packages/pigeon/tool/shared/test_runner.dart
```diff
@@ -18,31 +18,67 @@ Future<void> runTests(
   bool runFormat = false,
   bool runGeneration = true,
   bool ciMode = false,
+  bool overflow = false,
```
stuartmorgan, line 21: I would call this one `inculdeOverflow`. The other instances of this name all seem to be a switch controlling which of the two modes the method is in, but here the modes are "without" and "without and with".

stuartmorgan (changes requested):

packages/pigeon/lib/kotlin_generator.dart
tarrinneal: I mean, there really isn't a reason to have a strangely shaped prefix anyway. We can disallow any prefix. I just didn't want to block people from using `pigeon` as a prefix, since I thought that was probably fairly common. I'll adjust the prefix to something simpler and less ugly and add it to the block list. 

This will be a breaking change though.

packages/pigeon/lib/java_generator.dart
tarrinneal: If I changed the class signature to not be shared with data classes.

packages/pigeon/lib/cpp_generator.dart
```diff
@@ -23,6 +23,24 @@ const String _standardCodecSerializer = 'flutter::StandardCodecSerializer';
 /// The name of the codec serializer.
 const String _codecSerializerName = 'PigeonCodecSerializer';
 
+const String _overflowClassName = '${varNamePrefix}CodecOverflow';
```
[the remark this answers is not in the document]
tarrinneal: I changed the prefix, let me know if that is enough or if I need to `internal`ize it still.

packages/pigeon/lib/cpp_generator.dart
```diff
@@ -934,22 +1055,33 @@ class CppSourceGenerator extends StructuredGenerator<CppOptions> {
...
-              indent.writeln('stream->WriteByte(${customType.enumeration});');
-              if (customType.type == CustomTypes.customClass) {
-                indent.writeln(
-                    'WriteValue(EncodableValue(std::any_cast<${customType.name}>(*custom_value).ToEncodableList()), stream);');
-              } else if (customType.type == CustomTypes.customEnum) {
+              indent.writeln('stream->WriteByte($enumeration);');
+              if (enumeration == maximumCodecFieldKey) {
                 indent.writeln(
-                    'WriteValue(EncodableValue(static_cast<int>(std::any_cast<${customType.name}>(*custom_value))), stream);');
+                    'const auto& wrap = $_overflowClassName(${customType.enumeration - maximumCodecFieldKey}, $encodeString);');
```
stuartmorgan, line 1081: This one would be better as `auto` rather than `auto&`. The latter relies on subtle lifetime extension rules, and doesn't add anything.

packages/pigeon/lib/cpp_generator.dart
```diff
@@ -934,22 +1055,33 @@ class CppSourceGenerator extends StructuredGenerator<CppOptions> {
...
-          for (final EnumeratedType customType in customTypes) {
+          for (final EnumeratedType customType in enumeratedTypes) {
+            final String encodeString = customType.type ==
+                    CustomTypes.customClass
+                ? 'std::any_cast<${customType.name}>(*custom_value).ToEncodableList()'
+                : 'static_cast<int>(std::any_cast<${customType.name}>(*custom_value))';
+            final String valueString = customType.enumeration <
+                    maximumCodecFieldKey
+                ? encodeString
+                : 'std::any_cast<$_overflowClassName>(wrap).ToEncodableList()';
```
stuartmorgan, line 1070: Per offline discussion, this shouldn't be necessary, and I'm surprised it compiles since `wrap` isn't a `std::any`. This string *should* just be `'wrap.ToEncodableList()'` unless I've missed something.

packages/pigeon/lib/cpp_generator.dart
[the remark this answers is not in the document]
stuartmorgan: It would be more idiomatic C++ to put things that have to be in the header but aren't intended for direct use in an `internal` namespace, but it's not a big deal, so we can leave this as-is.

packages/pigeon/lib/objc_generator.dart
```diff
@@ -15,6 +15,24 @@ const String _docCommentPrefix = '///';
...
+    name: 'wrapped',
+    type: const TypeDeclaration(baseName: 'Object', isNullable: true));
+final List<NamedType> _overflowFields = <NamedType>[
+  _overflowInt,
+  _overflowObject,
+];
+final Class _overflowClass =
+    Class(name: _overflowClassName, fields: _overflowFields);
+final EnumeratedType _enumeratedOverflow = EnumeratedType(
+    _overflowClassName, 255, CustomTypes.customClass,
```
stuartmorgan, line 33: `maximumCodecFieldKey`?

packages/pigeon/lib/objc_generator.dart
```diff
@@ -624,34 +745,41 @@ class ObjcSourceGenerator extends StructuredGenerator<ObjcOptions> {
...
+      indent.writeScoped('switch (type) {', '}', () {
+        for (final EnumeratedType customType in enumeratedTypes) {
+          if (customType.enumeration < maximumCodecFieldKey) {
+            indent.write('case ${customType.enumeration}: ');
+            _writeCodecDecode(
+                indent, customType, generatorOptions.prefix ?? '');
+          }
+        }
+        if (root.requiresOverflowClass) {
+          indent.write('case 255: ');
```
stuartmorgan, line 777: `maximumCodecFieldKey`?

packages/pigeon/lib/cpp_generator.dart
```diff
@@ -21,7 +21,25 @@ const DocumentCommentSpecification _docCommentSpec =
...
+    name: 'wrapped',
+    type: const TypeDeclaration(baseName: 'Object', isNullable: false));
+final List<NamedType> _overflowFields = <NamedType>[
+  _overflowType,
+  _overflowObject,
+];
+final Class _overflowClass =
+    Class(name: _overflowClassName, fields: _overflowFields);
+final EnumeratedType _enumeratedOverflow = EnumeratedType(
+    _overflowClassName, 255, CustomTypes.customClass,
```
stuartmorgan, line 41: `maximumCodecFieldKey`?

packages/pigeon/lib/cpp_generator.dart
```diff
@@ -894,33 +1021,27 @@ class CppSourceGenerator extends StructuredGenerator<CppOptions> {
...
+        for (final EnumeratedType customType in enumeratedTypes) {
+          if (customType.enumeration < maximumCodecFieldKey) {
+            indent.write('case ${customType.enumeration}: ');
+            indent.nest(1, () {
+              _writeCodecDecode(indent, customType, 'ReadValue(stream)');
+            });
+          }
+        }
+        if (root.requiresOverflowClass) {
+          indent.write('case 255:');
```
stuartmorgan, line 1036: `maximumCodecFieldKey`?

packages/pigeon/lib/objc_generator.dart
```diff
@@ -197,6 +215,15 @@ class ObjcHeaderGenerator extends StructuredGenerator<ObjcOptions> {
       indent,
       dartPackageName: dartPackageName,
     );
+    if (root.requiresOverflowClass) {
```
stuartmorgan, line 218: For Objective-C, I think you could actually move this to the end of `writeGeneralUtilities` in the implementation file, because you don't need to reference it anywhere in the header.

You'd need to extract the body of the header `writeDataClass` to a file-level `_writeDataClassDeclaration` so you could use it from the implementation file, but I think that would be straightforward?

packages/pigeon/lib/swift_generator.dart
```diff
@@ -263,7 +284,99 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
...
+        name: 'wrapped',
+        type: const TypeDeclaration(baseName: 'Object', isNullable: true));
+    final List<NamedType> overflowFields = <NamedType>[
+      overflowInt,
+      overflowObject,
+    ];
+    final Class overflowClass =
+        Class(name: _overflowClassName, fields: overflowFields);
+    indent.newln();
+    indent.writeln('// swift-format-ignore: TypeNamesShouldBeCapitalized');
```
stuartmorgan, line 310: This can be removed.

packages/pigeon/lib/swift_generator.dart
```diff
@@ -263,7 +284,99 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
...
+      writeClassEncode(
+        generatorOptions,
+        root,
+        indent,
+        overflowClass,
+        dartPackageName: dartPackageName,
+      );
+
+      indent.format('''
+// swift-format-ignore: AlwaysUseLowerCamelCase
```
stuartmorgan, line 322: Can this be removed, or is it unhappy about the internal `_`?

packages/pigeon/lib/swift_generator.dart
```diff
@@ -263,7 +284,99 @@ class SwiftGenerator extends StructuredGenerator<SwiftOptions> {
...
+        type: const TypeDeclaration(baseName: 'Object', isNullable: true));
+    final List<NamedType> overflowFields = <NamedType>[
+      overflowInt,
+      overflowObject,
+    ];
+    final Class overflowClass =
+        Class(name: _overflowClassName, fields: overflowFields);
+    indent.newln();
+    indent.writeln('// swift-format-ignore: TypeNamesShouldBeCapitalized');
+    _writeDataClassSignature(indent, overflowClass);
```
stuartmorgan, line 311: Can this be private (by adding a `{bool private = false}` param to add the private keyword)?

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -369,6 +410,75 @@ class KotlinGenerator extends StructuredGenerator<KotlinOptions> {
...
+    final List<NamedType> overflowFields = <NamedType>[
+      overflowInt,
+      overflowObject,
+    ];
+    final Class overflowClass = Class(
+        name:
+            '${generatorOptions.fileSpecificClassNameComponent}$_overflowClassName',
+        fields: overflowFields);
+
+    indent.writeln('@Suppress("ClassName")');
```
stuartmorgan, line 435: Remove?

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -369,6 +410,75 @@ class KotlinGenerator extends StructuredGenerator<KotlinOptions> {
...
+        generatorOptions,
+        root,
+        indent,
+        overflowClass,
+        dartPackageName: dartPackageName,
+      );
+
+      indent.format('''
+companion object {
+  @Suppress("LocalVariableName")
```
stuartmorgan, line 448: Can this be removed, or it needed still?

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -369,6 +410,75 @@ class KotlinGenerator extends StructuredGenerator<KotlinOptions> {
...
+      overflowInt,
+      overflowObject,
+    ];
+    final Class overflowClass = Class(
+        name:
+            '${generatorOptions.fileSpecificClassNameComponent}$_overflowClassName',
+        fields: overflowFields);
+
+    indent.writeln('@Suppress("ClassName")');
+    _writeDataClassSignature(indent, overflowClass);
```
stuartmorgan, line 436: As with Swift, can we add `private`?

stuartmorgan (changes requested):

packages/pigeon/lib/pigeon_lib.dart
```diff
@@ -909,9 +919,54 @@ List<Error> _validateAst(Root root, String source) {
   final List<String> customClasses =
       root.classes.map((Class x) => x.name).toList();
   final Iterable<String> customEnums = root.enums.map((Enum x) => x.name);
+  for (final Enum enumDefinition in root.enums) {
```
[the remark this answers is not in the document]
tarrinneal: I did consider it, I figure it's better to just have a universal ban list and keep it simple.

packages/pigeon/lib/kotlin_generator.dart
tarrinneal: I added the logic, but forgot to use it...

stuartmorgan (approved):
LGTM!

PR #6871, opened 2024-06-05
[google_adsense] (Experimental) AdSense plugin for Flutter Web

sokoloff06:
Google Adsense plugin should allow Flutter Web developers to integrate Adsense more easily and monetize web project.

https://github.com/flutter/flutter/issues/40376

## Pre-launch Checklist

- [x] I signed the [CLA].


noobcoder95:
is this work? when will this codes included in flutter?

zhuhuihuihui:
What's the release timeline for this PR?

packages/google_adsense/lib/src/ad_unit_widget_web.dart
```diff
@@ -0,0 +1,201 @@
...
+/// Widget displaying an ad unit
+class AdUnitWidgetWeb extends AdUnitWidget {
+  // TODO(sokoloff06): consider builder?
+  /// Constructs [AdUnitWidgetWeb]
+  AdUnitWidgetWeb({
+    required String adClient,
+    required String adSlot,
+    required bool isAdTest,
+    required Map<String, dynamic> additionalParams,
+    String? cssText,
```
ditman, line 24: What do you think about making these values a configuration object instead? So we end up with:

```dart
AdUnitWidget(
  AdUnitConfig(
    adClient: "my_client_id",
    adSlot: "slot_name",
    additionalParams: AdUnitAdditionalParams(
      fullWidthResponsive: true,
      adFormat: AdFormat.SOME_TYPE,
    ),
  )
);
```

packages/google_adsense/lib/src/ad_unit_widget_web.dart
```diff
@@ -0,0 +1,201 @@
...
+        Map<String, String>.of(<String, String>{
+      AdUnitParams.AD_CLIENT: 'ca-pub-$_adClient',
+      'adSlot': _adSlot,
+    });
+    if (_isAdTest) {
+      dataAttrs.addAll(<String, String>{_AD_TEST_KEY: 'on'});
+    }
+    for (final String key in dataAttrs.keys) {
+      _insElement.dataset
+          .setProperty(key as JSString, dataAttrs[key]! as JSString);
```
ditman, line 46: Casting `as JSString` is (maybe?) going to break when we use this in WASM, try using `.toJS` instead from the new JS-interop:

```dart
  .setProperty(key.toJS, dataAttrs[key]!.toJS);
```

(See the docs: **JavaScript interop > JS Types > [Conversions](https://dart.dev/interop/js-interop/js-types#conversions)**)

packages/google_adsense/lib/src/ad_unit_widget_web.dart
```diff
@@ -0,0 +1,201 @@
...
+  @override
+  bool get wantKeepAlive => true;
+
+  @override
+  Widget build(BuildContext context) {
+    super.build(context);
+    return SizedBox(
+      height: adHeight,
+      child: HtmlElementView.fromTagName(
+          tagName: 'div', onElementCreated: onElementCreated),
```
ditman, line 96: Do we need a container `div` element, or could this be the `ins` directly?

packages/google_adsense/lib/src/ad_unit_widget_web.dart
```diff
@@ -0,0 +1,201 @@
...
+    // Using Resize observer to detect element attached to DOM
+    web.ResizeObserver((JSArray<web.ResizeObserverEntry> entries,
+            web.ResizeObserver observer) {
+      for (final web.ResizeObserverEntry entry in entries.toDart) {
+        final web.Element target = entry.target;
+        if (target.isConnected) {
+          // First time resized since attached to DOM -> attachment callback from Flutter docs by David
+          if (!(target as web.HTMLElement)
+              .dataset
+              .getProperty('attached' as JSString)
```
ditman, line 120: Why do we need this `'attached'` flag? In case the widget rebuilds, we'll get a brand new `element`, so I don't think the `target` will ever repeat?

packages/google_adsense/lib/src/ad_unit_widget_web.dart
```diff
@@ -0,0 +1,201 @@
...
+          }
+        }
+      }
+    }.toJS)
+        .observe(
+            widget._insElement,
+            web.MutationObserverInit(
+                attributes: true,
+                attributeFilter:
+                    <String>['data-ad-status'].jsify()! as JSArray<JSString>));
```
ditman, line 157: This line is a bit odd, you shouldn't need all these conversions or the as cast I think!

packages/google_adsense/lib/src/ad_unit_widget_web.dart
```diff
@@ -0,0 +1,201 @@
...
+    if (isLoaded) {
+      log('Ad is loaded');
+    } else {
+      log('Ad is loading');
+    }
+    return isLoaded;
+  }
+
+  bool isFilled(web.HTMLElement target) {
+    final JSAny? adStatus = target.dataset.getProperty('adStatus' as JSString);
```
ditman, line 177: Since this is Dart code, you don't need to to use "JSAny", you can:

```dart
final String? adStatus = target.dataset.getProperty('adStatus'.toJS);
```

(if the above complains, throw a `.toDart` so you convert the JS String to a proper Dart string)

packages/google_adsense/lib/src/ad_unit_widget_web.dart
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  void onElementAttached(web.HTMLElement element) {
+    log('Element ${element.id} attached with style: height=${element.offsetHeight} and width=${element.offsetWidth}');
+    adsbygoogle.loadAds();
+  }
+
+  bool isLoaded(web.HTMLElement target) {
+    final bool isLoaded =
+        target.dataset.getProperty('adStatus' as JSString).isDefinedAndNotNull;
```
ditman, line 167: There's a bunch of properties that we're setting/getting that should probably be in a nice enum/class, for example: `'adStatus'`, `'data-ad-status'` or `'attached'` instead of being hardcoded.

packages/google_adsense/lib/src/ad_unit_widget_web.dart
ditman: Also, if this were an `enum`, the compiler would help you check that your switch is checking all possible values!

packages/google_adsense/lib/src/ad_unit_widget_web.dart
```diff
@@ -0,0 +1,201 @@
...
+          log('Ad unfilled!');
+          return false;
+        }
+      default:
+        log('No data-ad-status attribute found');
+        return false;
+    }
+  }
+
+  void updateWidgetHeight(int newHeight) {
```
ditman, line 195: Can we detect that the ad unit did *not* load, and log to the console that it didn't (and maybe remove the platform view from the DOM?)

packages/google_adsense/lib/src/adsense_stub.dart
```diff
@@ -0,0 +1,37 @@
...
+      Map<String, dynamic> adUnitParams = const <String, dynamic>{},
+      String? cssText}) {
+    throw UnsupportedError('Only supported on web');
+  }
+
+  /// Only for use in tests
+  static void resetForTesting() {
+    _instance = null;
+  }
+}
```
ditman, line 37: I think this file should expose a global:

`final AdSense adSense = AdSense();`

So people can use this as:

```dart
  adSense.initialize(...);
  adSense.adUnit(...);
```

Instead of always having to do: `AdSense().things()`

packages/google_adsense/lib/src/adsense_stub.dart
```diff
@@ -0,0 +1,37 @@
...
+
+  Adsense._internal() {
+    _instance = this;
+  }
+
+  static Adsense? _instance = Adsense._internal();
+
+  /// Initialization API. Should be called ASAP, ideally in the main method of your app.
+  void initialize(String adClient) {
+    throw UnsupportedError('Only supported on web');
```
ditman, line 20: Since we always throw when calling `initialize` outside of the web, I'm not sure it makes sense to have this split API. I think this should be a pure web package, and users would need to use it by conditionally importing it wherever they need it, for example having a:

```
Widget getAds(config);
```

That has separate implementations, and it calls this package when `kIsWeb`.

Removing the stubs from the conditional exports of this package would simplify it a lot!

(See the sign-in button example [here](https://github.com/flutter/packages/blob/main/packages/google_sign_in/google_sign_in/example/lib/src/sign_in_button.dart))

packages/google_adsense/lib/src/adsense_web.dart
```diff
@@ -0,0 +1,66 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:web/web.dart' as web;
+import 'ad_unit_widget.dart';
+import 'ad_unit_widget_web.dart';
+
+/// Main class to work with the library
+class Adsense {
```
ditman, line 10: I just noticed: Is it `Adsense` or `AdSense`? I think it's the latter, if so, the class names should be `AdSenseSomething`

packages/google_adsense/lib/src/adsense_web.dart
```diff
@@ -0,0 +1,66 @@
...
+        adSlot: adSlot,
+        adClient: adClient.isNotEmpty ? adClient : _adClient,
+        isAdTest: isAdTest,
+        additionalParams: adUnitParams,
+        cssText: cssText);
+  }
+
+  void _addMasterScript(String adClient) {
+    final web.HTMLScriptElement adsbygoogle = web.HTMLScriptElement();
+    adsbygoogle.innerText = 'adsbygoogle = window.adsbygoogle || [];';
```
ditman, line 52: We need to use TrustedTypes for this `adsbygoogle` so we can safely set the `innerText`, and also for the `script` below, to set its `src`.

Example: [code](https://github.com/flutter/packages/blob/main/packages/google_identity_services_web/lib/src/js_loader.dart), [TrustedHTML docs](https://developer.mozilla.org/en-US/docs/Web/API/TrustedHTML), [TrustedScriptURL docs](https://developer.mozilla.org/en-US/docs/Web/API/TrustedScriptURL).

packages/google_adsense/lib/src/adsense_web.dart
```diff
@@ -0,0 +1,66 @@
...
+    final web.HTMLScriptElement script =
+        web.document.createElement('script') as web.HTMLScriptElement
+          ..async = true
+          ..crossOrigin = 'anonymous';
+    script.src = _url + adClient;
+    (web.document.head ?? web.document).appendChild(adsbygoogle);
+    (web.document.head ?? web.document).appendChild(script);
```
ditman, line 59: `web.document` is a bad default: calling `appendChild` on it is going to fail (try it in the JS console: `document.appendChild(document.createElement('div'))`.

In this case, I'd do:

```dart
web.document.head!.appendChild...
```

Instead of defaulting to `document`.

packages/google_adsense/lib/src/adsense_web.dart
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /// Only for use in tests
+  static void resetForTesting() {
+    _instance = null;
+  }
```
ditman, line 65: If we expose an `adSense` instance for users to use, tests can just create as many new `AdSense()` instances as they want, without having to `resetForTesting`.

packages/google_adsense/lib/src/js_interop/adsbygoogle.dart
```diff
@@ -0,0 +1,30 @@
...
+/// Binding to the `adsbygoogle` JS global.
+///
+/// See: https://support.google.com/adsense/answer/9274516?hl=en&ref_topic=28893&sjid=11495822575537499409-EU
+@JS('adsbygoogle')
+external AdsByGoogle get adsbygoogle;
+
+/// The Dart definition of the `adsbygoogle` global.
+@JS()
+@staticInterop
+abstract class AdsByGoogle {}
```
ditman, line 16: This style of JS-interop is a little bit outdated, this is now done with "Extension Types".

As an example, see how the "console" type (and object) is defined in `package:web`

* https://github.com/dart-lang/web/blob/d3a61aa9b4e43cdaf949a69c159038cdecf5681a/web/lib/src/dom/console.dart

packages/google_adsense/lib/src/js_interop/adsbygoogle.dart
```diff
@@ -0,0 +1,30 @@
...
+
+/// The `adsbygoogle` methods mappings
+extension AdsByGoogleExtension on AdsByGoogle {
+  /// Replacement of part of the adUnit code:
+  /// <script>
+  // (adsbygoogle = window.adsbygoogle || []).push({});
+  // </script>
+  external void push(JSObject? params);
```
ditman, line 24: In this case, since `adsbygoogle` is a "JSArray", if your JS interop type is like:

```
@JS('adsbygoogle')
extension type AdsByGoogle._(JSArray _) implements JSArray {
```

You may get the "push" method for free (with the exact same definition that you have).

`push` can also be made "private", so Dart users only see the `requestAd()` method, like this:

```
  @JS('push')
  external void _push(JSObject? params);
```

(Then `requestAd()` would call `_push` instead of `push`, of course).

packages/google_adsense/lib/src/js_interop/adsbygoogle.dart
```diff
@@ -0,0 +1,30 @@
...
+
+  /// Convenience method for invoking push() with an empty object
+  void loadAds() {
```
ditman, line 27: Should this be named: `requestAd()`?

packages/google_adsense/lib/src/js_interop/adsbygoogle.dart
```diff
@@ -0,0 +1,30 @@
...
+    push(<String, String>{}.jsify() as JSObject?);
```
ditman, line 28: (You shouldn't have to "as" after a call to ".jsify")

packages/google_adsense/test/adsense_apis_test.dart
```diff
@@ -0,0 +1,21 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@TestOn('vm')
```
ditman, line 5: This code should never be executed in the `vm`, so this test can possibly be removed.

packages/google_adsense/test/adsense_stub_test.dart
```diff
@@ -0,0 +1,56 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@TestOn('vm')
```
ditman, line 5: IMO Stubs should be removed, so this test can go as well.

packages/google_adsense/test/adsense_web_test.dart
```diff
@@ -0,0 +1,84 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Run locally with flutter test --platform=chrome
+
+@TestOn('chrome')
```
ditman, line 7: This is a good test, but if you want to, you can move it to the `example/integration_test`, so all tests stay together.

What I do in these cases, is I create a bogus "vm" test that tells users that "tests live elsewhere", see:

* https://github.com/flutter/packages/tree/main/packages/google_sign_in/google_sign_in_web/test

ditman (commented):
Thanks for the code @sokoloff06!

I left a bunch of comments, but I think all should be easy to fix/actionable.

Do message me if you need any help with any of the fixes (esp. on the JS-interop stuff, that is in flux)!

packages/google_adsense/README.md
```diff
@@ -0,0 +1,60 @@
...
+void main() {
+  Adsense().initialize('your_ad_client_id');
+  runApp(const MyApp());
+}
+
+```
+#### Display AdUnitWidget
+<?code-excerpt "example/lib/main.dart (adUnit)"?>
+```dart
+Adsense().adUnit(
```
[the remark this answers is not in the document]
sokoloff06: @ditman agree, thanks for flagging! Will fix so API will be `AdSense.instance`

packages/google_adsense/lib/src/adsense_stub.dart
sokoloff06: @ditman thanks! In the case when the consuming app is cross-platform and ads are only implemented on Web, is there a way to import only web package? 

I couldn't find a way to do it, it seems like developers need to import "something" on other platforms, even if they only want web package. Happy to sync on that too

packages/google_adsense/README.md
```diff
@@ -0,0 +1,60 @@
...
+    adSlot: 'your_ad_slot_id',
+    isAdTest: true,
+    adUnitParams: <String, dynamic>{
+      AdUnitParams.AD_FORMAT: 'auto',
+      AdUnitParams.FULL_WIDTH_RESPONSIVE: true,
+    },
```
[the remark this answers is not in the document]
sokoloff06: I created `AdUnitParams` class just to provide String keys for **some** of the common parameters that developers might want to pass to adUnitParams argument. There might be more of those

But, I do think there is no need for dynamic, it can be strongly typed String. Under the hood these will simply become data attributes on the html element. E.g.  `AdUnitParams.AD_FORMAT: 'auto'` becomes `data-ad-format=auto`

packages/google_adsense/lib/src/ad_unit_params.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Contains adUnitParams keys constants for convenience and documentation
+class AdUnitParams {
```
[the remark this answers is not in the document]
sokoloff06: Hi @ditman! Going over it again, Im not sure camelCase naming will bring any value (because there is not pure JS API for AdSense, rather data-attributes in the html tag). At the same time SCREAMING_SNAKE_CASE suggests that these are constants so Im not changing this for now

packages/google_adsense/lib/src/ad_unit_widget_web.dart
sokoloff06: I think I got rid of all as JS* casts, will test WASM compatibility as well

packages/google_adsense/example/README.md
```diff
@@ -0,0 +1,10 @@
+# google_adsense_example
+
+An example demonstrating google_adsense Flutter plugin usage
+
+## Screenshots
+
+![Screenshot of the test app showing an ad on mobile](../example/images/mobile_screenshot.png)
```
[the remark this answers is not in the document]
sokoloff06: Great idea, added, thanks!

packages/google_adsense/lib/src/adsense_stub.dart
sokoloff06: I have changed it to be Adsense.instance. Let's discuss if global is a better approach

packages/google_adsense/README.md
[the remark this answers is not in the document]
sokoloff06: As discussed, will export as global

packages/google_adsense/example/integration_test/test_ad.html
```diff
@@ -0,0 +1,4477 @@
+<!-- Copyright 2013 The Flutter Authors. All rights reserved.
+Use of this source code is governed by a BSD-style license that can be
+found in the LICENSE file. -->
+
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
sokoloff06: Option 1:
Mock Adsense.push() method that would
1. Inject custom JS on the setup or Use data-url instead of real AdSense SDK to load "fake" implementation
- with 100ms network simulation:
2. Return/render something smaller (div filled with color/image?)
3. Response as unfilled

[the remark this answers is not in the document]
sokoloff06: Assert the size of the platform view (0 or size of the ad)

packages/google_adsense/lib/src/ad_unit_widget_web.dart
sokoloff06: const as JSString
enum of ore than 1

packages/google_adsense/lib/src/adsense_web.dart
sokoloff06: Use js interop to work directly on window but use trusted Types/source for script src

packages/google_adsense/lib/src/adsense_web.dart
sokoloff06: We don't need singleton if we use global, so we can recreate adsense instances for tests

packages/google_adsense/test/adsense_web_test.dart
sokoloff06: Move all tests to integrations tests
Use vm tests to check for throwing

packages/google_adsense/lib/src/js_interop/adsbygoogle.dart
sokoloff06: @ditman Interestingly when using the new interop style Im getting the following error from time to time:
`Instance of 'LegacyJavaScriptObject': type 'LegacyJavaScriptObject' is not a subtype of type JSArray<JSObject?>`
Im not quite sure what is causing it and it is not happening all the time

packages/google_adsense/lib/src/ad_unit_widget_web.dart
sokoloff06: Created AdUnitConfiguration class and updated the API

packages/google_adsense/lib/src/ad_unit_widget_web.dart
sokoloff06: Compiles and runs with WASM succesfully! See adsense-test.visokolov.com

packages/google_adsense/lib/src/ad_unit_widget_web.dart
sokoloff06: @ditman I tried to remove the wrapping div element but noticed that in such case some developer custom css rules can break ad behavior. 
E.g. When I applied a 5px border:

- Without wrapping div, right border is out of the viewport
image

- With wrapping div it is processed correctly:
image

I don't know how important it is, I also played with paddins and margins and they caused overflows in both cases. I have a feeling there might be some conflict between Flutter platform view and adsense logic. 

Would love to hear your thoughts on this

packages/google_adsense/lib/src/adsense_web.dart
sokoloff06: Added Trusted Types with heavy re-use of [your package_web extensions](https://github.com/flutter/packages/blob/main/packages/google_identity_services_web/lib/src/js_interop/package_web_tweaks.dart), thanks for those! Hope eventually it will be part of the web package

Would you taking a look, to see if I haven't missed anything

Tananga:
🥳🥳🥳🥳🥳

packages/google_adsense/lib/src/ad_unit_widget_web.dart
sokoloff06: Agreeed to remove cssText property for now and document that Widget styling should be done on Flutter side

sokoloff06: @ditman I wll also keep the wrapping div for now I think as it make widget creation code a little more readable.

ditman:
(Just verified that web integration tests are running Just Fine, [here](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8729722350423670737/+/u/Run_package_tests/drive_examples/stdout))

ditman (approved):
LGTM, thanks @sokoloff06!

Tananga:
Great job!

PR #6900, opened 2024-06-10
[flutter_markdown] Fix Scrollbars not showing up when multiple code blocks are in the same markdown.

apwilson:
If the same ScrollController is used for multiple Scrollbars an exception is thrown.


domesticmouse:
Is there a way to add a test for this change?

PR #6904, opened 2024-06-10
Ensure each code block specified in the markdown uses its own ScrollController.

apwilson:
Currently all code blocks share the same ScrollController causing an exception when they appear on screen together.  This fixes that situation.


domesticmouse (approved):

domesticmouse:
PTAL @stuartmorgan

PR #6908, opened 2024-06-11
Refactor `VideoPlayer` to be less exposed to `EventChannel` & related

matanlurey:
Part of https://github.com/flutter/flutter/issues/148417.

These are some proposed refactors to limit how much work `VideoPlayer` is doing, so I can better understand what needs to be reset (or not) internally. Specifically, `VideoPlayer` no longer knows what an `EventChannel` or `EventSink` is, and does not need to manage the lifecycle (it stores a `private final VideoPlayerCallbacks` instead), and instead there is a `VideoPlayerCallbacks` interface that does all that.

I'm totally open to:
- Landing this as-is (+/- nits) and making minor improvements in follow-up PRs
- Making more significant changes to this PR and then landing it
- Not landing this PR at all because it doesn't follow the approach the folks who maintain the plugin prefer

Also happy to chat in VC/person about any of the changes.

### DIFF
--- a/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
+++ b/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
@@ -28,12 +28,7 @@ import androidx.media3.datasource.DefaultDataSource;
 import androidx.media3.datasource.DefaultHttpDataSource;
 import androidx.media3.exoplayer.ExoPlayer;
 import androidx.media3.exoplayer.source.DefaultMediaSourceFactory;
-import io.flutter.plugin.common.EventChannel;
 import io.flutter.view.TextureRegistry;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 final class VideoPlayer {
@@ -48,9 +43,7 @@ final class VideoPlayer {
 
   private final TextureRegistry.SurfaceTextureEntry textureEntry;
 
-  private QueuingEventSink eventSink;
-
-  private final EventChannel eventChannel;
+  private final VideoPlayerCallbacks events;
 
   private static final String USER_AGENT = "User-Agent";
 
@@ -62,13 +55,13 @@ final class VideoPlayer {
 
   VideoPlayer(
       Context context,
-      EventChannel eventChannel,
+      VideoPlayerCallbacks events,
       TextureRegistry.SurfaceTextureEntry textureEntry,
       String dataSource,
       String formatHint,
       @NonNull Map<String, String> httpHeaders,
       VideoPlayerOptions options) {
-    this.eventChannel = eventChannel;
+    this.events = events;
     this.textureEntry = textureEntry;
     this.options = options;
 
@@ -86,24 +79,23 @@ final class VideoPlayer {
     exoPlayer.setMediaItem(mediaItem);
     exoPlayer.prepare();
 
-    setUpVideoPlayer(exoPlayer, new QueuingEventSink());
+    setUpVideoPlayer(exoPlayer);
   }
 
   // Constructor used to directly test members of this class.
   @VisibleForTesting
   VideoPlayer(
       ExoPlayer exoPlayer,
-      EventChannel eventChannel,
+      VideoPlayerCallbacks events,
       TextureRegistry.SurfaceTextureEntry textureEntry,
       VideoPlayerOptions options,
-      QueuingEventSink eventSink,
       DefaultHttpDataSource.Factory httpDataSourceFactory) {
-    this.eventChannel = eventChannel;
+    this.events = events;
     this.textureEntry = textureEntry;
     this.options = options;
     this.httpDataSourceFactory = httpDataSourceFactory;
 
-    setUpVideoPlayer(exoPlayer, eventSink);
+    setUpVideoPlayer(exoPlayer);
   }
 
   @VisibleForTesting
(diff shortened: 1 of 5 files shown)


packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerCallbacks.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Callbacks representing events invoked by {@link VideoPlayer}.
+ *
+ * <p>In the actual plugin, this will always be {@link VideoPlayerEventCallbacks}, which creates the
+ * expected events to send back through the plugin channel. In tests methods can be overridden in
+ * order to assert results.
+ */
+interface VideoPlayerCallbacks {
+  void onInitialized(
```
gaaclarke, line 18: nit: `rotationCorrectionInDegrees = 0` and `rotationCorrectionInDegrees = null` have the same semantic meaning, right?  If so, no need to make this nullable.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerEventCallbacks.java
```diff
@@ -0,0 +1,96 @@
...
+
+          @Override
+          public void onCancel(Object arguments) {
+            eventSink.setDelegate(null);
+          }
+        });
+    return VideoPlayerEventCallbacks.withSink(eventSink);
+  }
+
+  static VideoPlayerEventCallbacks withSink(EventChannel.EventSink eventSink) {
```
gaaclarke, line 34: `@VisibleForTesting`?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerEventCallbacks.java
```diff
@@ -0,0 +1,96 @@
...
+  public void onBufferingUpdate(long bufferedPosition) {
+    // iOS supports a list of buffered ranges, so we send as a list with a single range.
+    Map<String, Object> event = new HashMap<>();
+    event.put("values", Collections.singletonList(bufferedPosition));
+    eventSink.success(event);
+  }
+
+  @Override
+  public void onBufferingEnd() {
+    Map<String, Object> event = new HashMap<>();
```
gaaclarke, line 73: nit: All of these events are allocating new HashMap with the same value.  They could be private constants to avoid generating garbage.  None of them seem that chatty though.

gaaclarke (approved):
lgtm modulo some nitpicks

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerCallbacks.java
matanlurey: Good point! Done.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerEventCallbacks.java
matanlurey: Done!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerEventCallbacks.java
matanlurey: I'm going to leave it as-is for now, hopefully we can use `Map.of` soonish?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerCallbacks.java
```diff
@@ -0,0 +1,31 @@
...
+import androidx.annotation.Nullable;
+
+/**
+ * Callbacks representing events invoked by {@link VideoPlayer}.
+ *
+ * <p>In the actual plugin, this will always be {@link VideoPlayerEventCallbacks}, which creates the
+ * expected events to send back through the plugin channel. In tests methods can be overridden in
+ * order to assert results.
+ */
+interface VideoPlayerCallbacks {
```
camsim99, line 17: Maybe link the `Player.Listener` [docs](https://developer.android.com/reference/androidx/media3/common/Player.Listener) to the callbacks that directly map to one of those events?

camsim99 (approved):
LGTM! Just left some suggestions :)

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -180,30 +161,20 @@ public void onPlayerError(@NonNull final PlaybackException error) {
               // See https://exoplayer.dev/live-streaming.html#behindlivewindowexception-and-error_code_behind_live_window
               exoPlayer.seekToDefaultPosition();
               exoPlayer.prepare();
-            } else if (eventSink != null) {
-              eventSink.error("VideoError", "Video player had error " + error, null);
+            } else {
+              events.onError("VideoPlayer", "Video player had error " + error, null);
```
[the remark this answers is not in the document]
matanlurey: Oh, nope, that was an actual mistake - I'll also add a test before I resolve this comment.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerCallbacks.java
matanlurey: Done!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -48,9 +43,7 @@ final class VideoPlayer {
 
   private final TextureRegistry.SurfaceTextureEntry textureEntry;
 
-  private QueuingEventSink eventSink;
-
-  private final EventChannel eventChannel;
+  private final VideoPlayerCallbacks events;
```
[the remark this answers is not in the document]
matanlurey: Done!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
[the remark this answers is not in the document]
matanlurey: Done and test added!

PR #6922, opened 2024-06-13
Move `Player.Listener` impl, remove `@VisibleForTesting isInitialized`.

matanlurey:
Similar to https://github.com/flutter/packages/pull/6908, as part of https://github.com/flutter/flutter/issues/148417.

In this PR, I've removed the concept of the class being "initialized" or not - the only thing "initialized" means is "for a given instance of `ExoPlayer`, has received the `'initialized'` event. As a result I removed the quasi-public API that was used for testing only and replaced it with observing what the real production instance does (`Player.STATE_READY`).

After this PR, I'll likely do one more pass around the constructors - the constructor that takes an `ExoPlayer` that is marked `@VisibleForTesting` _also_ doesn't make sense once we'll support suspending/resuming video players, so it will need to get reworked (probably into taking a factory method).


packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+    events.onInitialized(width, height, exoPlayer.getDuration(), rotationCorrection);
+  }
+
+  @Override
+  public void onPlaybackStateChanged(final int playbackState) {
+    switch (playbackState) {
+      case Player.STATE_BUFFERING:
+        setBuffering(true);
```
gmackall, line 69: This is a slight behavior change, right?

Before we would always invoke `videoPlayerEvents.onBufferingUpdate(exoPlayer.getBufferedPosition())`:
```
setBuffering(true);
sendBufferingUpdate();
```
but now we only invoke it if we pass the check at the start of the slightly modified `setBuffering`.

I don't know enough about `video_player` to know if that case (we are in the case of this switch `Player.STATE_BUFFERING` while `isBuffering` is already true) can happen/matters, so feel free to ignore if it isn't important.

gmackall (commented):
Mostly LGTM just one question about the `Player.STATE_BUFFERING` case

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
matanlurey: Not only was this a good catch, but you caught a bug missed in the previous refactor.

I've fixed this and added a test to catch regressions.

gmackall (approved):

matanlurey:
@camsim99 I am going to add auto-submit to fix the previous regression, but please leave comments still and I promise I'll address in a follow-up PR.

PR #6923, opened 2024-06-13
[many] More v1 embedding deletion that was missed in https://github.com/flutter/packages/pull/6494

gmackall:
There were some additional v1 Android embedding references missed by https://github.com/flutter/packages/pull/6494.

This PR aims to remove those missed references.

I built the android example app of each plugin affected here on the [deletion branch](https://github.com/flutter/engine/pull/52022) to be completely sure:
- [x] `video_player`
- [x] `google_sign_in` (builds, but only with the fixes to `espresso` applied).
- [x] `espresso`
- [x] `in_app_purchase`
- [x] `image_picker` (builds, but only with the fixes to both `espresso` and `video_player` applied)

Final testing, I tested that the `all_packages` app builds on the v1 embedding deletion branch: 
```
$ ./.ci/scripts/create_all_packages_app.sh && cd all_packages
...
$ flutter build apk --debug --local-engine-src-path=/Users/mackall/development/engine/src --local-engine=android_debug_arm64 --local-engine-host=host_debug

Running Gradle task 'assembleDebug'...                              4.8s
✓ Built build/app/outputs/flutter-apk/app-debug.apk
```

🎉 

Linux repo checks are failing 
```
The following packages had errors:
  packages/google_maps_flutter/google_maps_flutter:
    Missing CHANGELOG change
  packages/palette_generator:
    Missing CHANGELOG change
  packages/quick_actions/quick_actions:
    Missing CHANGELOG change
  packages/webview_flutter/webview_flutter:
    Missing CHANGELOG change
```
The only changes to these packages are `xml` changes to example apps which won't cause build failures even when the v1 embedding is deleted (they will just be unused). So I believe these changes should be version exempt.

## Pre-launch Checklist

- [x] I signed the [CLA].


gmackall:
cc @stuartmorgan for a versioning question (no rush) - In a PR like this, where a fix to `google_sign_in` depends on a fix to `espresso`, will I need to land the fix to the latter first, and then put up a follow up PR to fix the former?

gmackall:
I tried to keep the commit history meaningful to help review - the `google_sign_in` removal ([1](https://github.com/flutter/packages/pull/6923/commits/2611aa08db6f78134d7039a70cbf43f0a3a8ed88),[2](https://github.com/flutter/packages/pull/6923/commits/0173fc3f1dad67cf1863341b474b7d7fb1743a11)) and the `espresso` removal ([1](https://github.com/flutter/packages/pull/6923/commits/53f8f88c11c2ac00cc78dbcd91b72099de5ced75)) were the non-trivial ones.

stuartmorgan:
> cc @stuartmorgan for a versioning question (no rush) - In a PR like this, where a fix to `google_sign_in` depends on a fix to `espresso`, will I need to land the fix to the latter first, and then put up a follow up PR to fix the former?

What's the dependency exactly?

- If it's just that both need to have the latest version for the v1 complaint not to trigger, not necessarily (it can help avoid issues for people who don't `flutter pub upgrade`, but that's likely to be necessary for most people for the v1 removals anyway).
- If `google_sign_in`'s tests now actually require a higher version of `espresso` to successfully run, then that should be expressed in `pubspec.yaml`, which will require the `espresso` change to be published first.

packages/espresso/android/src/main/java/androidx/test/espresso/flutter/internal/protocol/impl/DartVmServiceUtil.java
```diff
@@ -71,25 +70,15 @@ public static String getDartIsolateId(View flutterView) {
...
-          ((io.flutter.embedding.android.FlutterView) flutterView).getAttachedFlutterEngine();
-      if (flutterEngine == null) {
-        throw new FlutterProtocolException(
-            String.format(
-                "No Flutter engine attached to the Flutter view [id: %d].", flutterView.getId()));
-      }
-      return flutterEngine.getDartExecutor();
-    } else {
+  public static DartExecutor getDartExecutor(@Nonnull FlutterView flutterView) {
+    // TODO(gmackall): getAttachedFlutterEngine() is marked as @visibleForTesting - determine if
```
reidbaker, line 74: Not a fan of Visible for testing methods being used in production code.

packages/espresso/android/src/main/java/androidx/test/espresso/flutter/matcher/FlutterMatchers.java
```diff
@@ -99,8 +99,7 @@ public void describeTo(Description description) {
     @SuppressWarnings("deprecation")
```
reidbaker, line 99: Is this suppressed warning still needed? I think not since you removed the deprecated class.

reidbaker (approved):

packages/espresso/android/src/main/java/androidx/test/espresso/flutter/matcher/FlutterMatchers.java
gmackall: No it isn't, thanks!

packages/espresso/android/src/main/java/androidx/test/espresso/flutter/internal/protocol/impl/DartVmServiceUtil.java
gmackall: Agreed, but I suppose given that this is the `espresso` plugin this is indeed actually used for testing?

reidbaker: Yeah you are correct this is test code.

gmackall:
> * If `google_sign_in`'s tests now actually require a higher version of `espresso` to successfully run, then that should be expressed in `pubspec.yaml`, which will require the `espresso` change to be published first.

It will require a newer `espresso` version to run when https://github.com/flutter/engine/pull/52022 lands, but does not at this time

gmackall:
> > * If `google_sign_in`'s tests now actually require a higher version of `espresso` to successfully run, then that should be expressed in `pubspec.yaml`, which will require the `espresso` change to be published first.
> 
> It will require a newer `espresso` version to run when [flutter/engine#52022](https://github.com/flutter/engine/pull/52022) lands, but does not at this time

I believe this means it should be safe to land in the same PR

Also, separately from the versioning question would also appreciate your review, when you have a chance @stuartmorgan (not a rush just a clarification that the review request wasn't only for versioning question 🙂 )

stuartmorgan:
> I believe this means it should be safe to land in the same PR

Agreed, I don't expect any issue given that. And if it weren't safe, I would expect presubmits to fail, since the native tests in CI are not using path-based deps.

stuartmorgan (approved):
LGTM modulo making `format` happy.

stuartmorgan:
Overriding CHANGELOG check: the manifest XML changes are not interesting to clients, so don't need to be mentioned.

PR #6952, opened 2024-06-19
Fix breakages introduced by `SplashFactory`

nate-thegrate:
Preparing for:
- https://github.com/flutter/flutter/pull/149963


nate-thegrate:
I realized that instead of changing the type to `StatelessElement`, we could just use `Element`, and we'd be able to land it now without breakages!

goderbauer (approved):
LGTM

PR #6962, opened 2024-06-20
[shared_preferences] Platform interface for new shared preferences async

tarrinneal:
Platform interface portion of https://github.com/flutter/packages/pull/5210

Adds new async api

## Pre-launch Checklist

- [x] I signed the [CLA].


tarrinneal:
@stuartmorgan this pr is ready to go now that the other has approval, I've moved all new changes over to here for the PI

packages/shared_preferences/shared_preferences_platform_interface/README.md
```diff
@@ -1,17 +1,26 @@
 # shared_preferences_platform_interface
 
-A common platform interface for the [`shared_preferences`][1] plugin.
+A common platform interface for the [`shared_preferences`][1] plugins.
```
stuartmorgan, line 3: It's still just one plugin; plugin is shorthand for plugin package, so refers to the whole package rather than the classes.

packages/shared_preferences/shared_preferences_platform_interface/README.md
```diff
@@ -1,17 +1,26 @@
...
 plugin, as well as the plugin itself, to ensure they are supporting the
 same interface.
 
 # Usage
 
-To implement a new platform-specific implementation of `shared_preferences`, extend
-[`SharedPreferencesPlatform`][2] with an implementation that performs the
-platform-specific behavior, and when you register your plugin, set the default
-`SharedPreferencesLoader` by calling the `SharedPreferencesPlatform.loader` setter.
+To implement a new platform-specific implementation of `shared_preferences`,
```
stuartmorgan, line 11: I don't think we want to describe this is something where people should make an individual decision by file; instead I would say:

```
To implement a new platform-specific implementation of `shared_preferences`, extend
[`SharedPreferencesPlatform`][2] and [`SharedPreferencesAsyncPlatform`][3] with
implementations that perform the platform-specific behaviors, and when you register
your plugin, set the default `SharedPreferencesStorePlatform` and
`SharedPreferencesAsyncPlatform` by calling the `SharedPreferencesPlatform.instance`
and `SharedPreferencesAsyncPlatform.instance` setters.

Please note that the plugin tooling only registers the native and/or Dart classes
listed in your package's `pubsec.yaml`, so if you intend to implement more than
one class, you will need to manually register the second class
(as can be seen in the Android and iOS implementations).
```

packages/shared_preferences/shared_preferences_platform_interface/lib/in_memory_shared_preferences_async.dart
```diff
@@ -0,0 +1,158 @@
...
+  ) async {
+    return _data[key] as int?;
+  }
+
+  @override
+  Future<List<String>?> getStringList(
+    String key,
+    SharedPreferencesOptions options,
+  ) async {
+    final List<Object?>? data = _data[key] as List<Object?>?;
```
stuartmorgan, line 141: `final List<Object>? data = _data[key] as List<Object>?;`

Whatever type the list has should be non-nullable, so I believe that will work at runtime. (In fact, for this particular implementation I think just `return _data[key] as List<String>?` would be fine since this is staying within Dart, so the value should never lose the original `List<String>` runtime type.)

stuartmorgan (approved):
LGTM with minor comments!

PR #6965, opened 2024-06-21
[shared_preferences] Platform implementations of async api

tarrinneal:
part 2 of https://github.com/flutter/packages/pull/5210


tarrinneal:
@reidbaker any chance you can help me diagnose this gradle issue?

stuartmorgan:
I *believe* that ("`Unsupported class file major version`") is the result of something requiring a newer Java version than is available on the machine?

reidbaker:
> I _believe_ that ("`Unsupported class file major version`") is the result of something requiring a newer Java version than is available on the machine?

Yes that is correct.

reidbaker:
Unsupported class file major version 65 means you are using java 21 and should be using something lower [more detail here](https://stackoverflow.com/questions/77974590/unsupported-class-file-major-version-65-with-java-21-workaround) 

Looking at your changes I dont see anything that should cause this failure. @christopherfujino do you see something about this machine config that would cause it to run java 21 instead of the lower java versions we expect.

stuartmorgan:
The machine should definitely be using 17: https://github.com/flutter/packages/blob/3144ac0576e30ab73653ec3fbced4e598882f4bf/.ci.yaml#L32

Could it be something precompiled that used 21?

packages/shared_preferences/shared_preferences_foundation/lib/shared_preferences_foundation.dart
```diff
@@ -2,106 +2,236 @@
...
+}
+
+/// Options for the Foundation specific SharedPreferences plugin.
+@immutable
+class SharedPreferencesFoundationOptions extends SharedPreferencesOptions {
+  /// Creates a new instance with the given options.
+  SharedPreferencesFoundationOptions({
+    this.suiteName,
+  }) {
+    if (Platform.isIOS && !(suiteName?.startsWith('group.') ?? true)) {
```
stuartmorgan, line 215: Let's comment why here, so it's obvious why we're restricting the API. Something like:

```dart
// Ensure that use of suite is compliant with required reason API category 1C8F.1; see
// https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api
```

packages/shared_preferences/shared_preferences_android/lib/shared_preferences_android.dart
```diff
@@ -4,99 +4,194 @@
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
-import 'package:shared_preferences_platform_interface/shared_preferences_platform_interface.dart';
+import 'package:shared_preferences_platform_interface/shared_preferences_async_platform_interface.dart';
 import 'package:shared_preferences_platform_interface/types.dart';
 
-import 'src/messages.g.dart';
+import 'legacy_shared_preferences_android.dart';
```
stuartmorgan, line 10: Oops, I missed this in all of the implementations: you need to export this file, otherwise it's a breaking change for the package.

Let's instead do the more common thing we do with multiple files that get exported: make the top-level file just an export file, then put the actual implementations in src. See `url_launcher/lib/url_launcher.dart` for example.

stuartmorgan: Wait, but you also renamed the Dart classes, which is also breaking. So you'll have to undo the name changes, or major-version-bump all of the implementation packages.

stuartmorgan (changes requested):

christopherfujino:
> Unsupported class file major version 65 means you are using java 21 and should be using something lower [more detail here](https://stackoverflow.com/questions/77974590/unsupported-class-file-major-version-65-with-java-21-workaround)
> 
> Looking at your changes I dont see anything that should cause this failure. @christopherfujino do you see something about this machine config that would cause it to run java 21 instead of the lower java versions we expect.

@gmackall I vaguely recall us digging into some configs, and realizing that--regardless of what we last installed via CIPD, the tooling would choose the newest version installed. Do you remember the context, and if that's what's happening here?

reidbaker:
At one point there was a bug where different parts of the tooling had different java evaluations. But I think @andrewkolos fixed all known instances and refactored the logic to be shared.

gmackall:
> > Unsupported class file major version 65 means you are using java 21 and should be using something lower [more detail here](https://stackoverflow.com/questions/77974590/unsupported-class-file-major-version-65-with-java-21-workaround)
> > Looking at your changes I dont see anything that should cause this failure. @christopherfujino do you see something about this machine config that would cause it to run java 21 instead of the lower java versions we expect.
> 
> @gmackall I vaguely recall us digging into some configs, and realizing that--regardless of what we last installed via CIPD, the tooling would choose the newest version installed. Do you remember the context, and if that's what's happening here?

Hmm, I'm not sure I remember that discussion, but regardless - there should only be one Java version installed according to the `.ci.yaml` link that @stuartmorgan provided (unless something else is somehow installing a Java version). 

Also, it looks like the latest failures are just test failures now, not the same Gradle error?

```
io.flutter.plugins.sharedpreferences.SharedPreferencesTest > testClearWithAllowList FAILED
    java.lang.ClassCastException at SharedPreferencesTest.kt:152
```

tarrinneal:
I updated the kotlin version to the version that is often forced using the `bom` method. That resolved the gradle issues

packages/shared_preferences/shared_preferences_android/lib/shared_preferences_android.dart
tarrinneal: the "file must be exported or else it's a breaking change" issue could probably use a test at some point

packages/shared_preferences/shared_preferences_foundation/darwin/shared_preferences_foundation/Sources/shared_preferences_foundation/SharedPreferencesPlugin.swift
```diff
@@ -54,18 +57,130 @@ public class SharedPreferencesPlugin: NSObject, FlutterPlugin, UserDefaultsApi {
...
+  {
+    return try SharedPreferencesPlugin.getAllPrefs(allowList: allowList, options: options)
+  }
+
+  func set(key: String, value: Any, options: SharedPreferencesPigeonOptions) throws {
+    try SharedPreferencesPlugin.getUserDefaults(options: options).set(value, forKey: key)
+  }
+
+  func getValue(key: String, options: SharedPreferencesPigeonOptions) throws -> Any? {
+    return try SharedPreferencesPlugin.getUserDefaults(options: options).object(forKey: key)
```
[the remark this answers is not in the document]
tarrinneal: That case is actually covered by the current system, I tested it.

packages/shared_preferences/shared_preferences_android/android/build.gradle
```diff
@@ -2,13 +2,15 @@ group 'io.flutter.plugins.sharedpreferences'
 version '1.0-SNAPSHOT'
 
 buildscript {
+    ext.kotlin_version = '1.8.10'
```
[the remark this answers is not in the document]
tarrinneal: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8741408197539124465/+/u/Run_package_tests/native_unit_tests/stdout?format=raw

Integration tests run fine though, only unit tests fail. There may be some other underlying issue.

stuartmorgan:
@tarrinneal I played around with a bit locally, and if I add the `implementation(platform("org.jetbrains.kotlin:kotlin-bom:1.8.22"))` workaround to `android/build.gradle`, drive-examples works (which suggests to me that clients should be fine), then if I add the `kotlin_version` bump to `1.8.10` *only* in `example/android/build.gradle` (which doesn't affect clients), unit tests compiled and passed.

@reidbaker Does that seem like a reasonable thing to do? I'm still not clear why this example needs a different `kotlin_version` when none of the others do, but maybe it's because the unit tests are Kotlin?

reidbaker:
> @tarrinneal I played around with a bit locally, and if I add the `implementation(platform("org.jetbrains.kotlin:kotlin-bom:1.8.22"))` workaround to `android/build.gradle`, drive-examples works (which suggests to me that clients should be fine), then if I add the `kotlin_version` bump to `1.8.10` _only_ in `example/android/build.gradle` (which doesn't affect clients), unit tests compiled and passed.
> 
> @reidbaker Does that seem like a reasonable thing to do? I'm still not clear why this example needs a different `kotlin_version` when none of the others do, but maybe it's because the unit tests are Kotlin?

It seems weird to me that example/android/build.gradle cant use kotlin 1.8.22. The only reason for that would be if the dependencies used in example/android/app/build.gradle were too old to use a newer version of kotlin. 

In general I think it is ok to have the example app have whatever version of kotlin but it seems to me like they should be aligned.

stuartmorgan:
> It seems weird to me that example/android/build.gradle cant use kotlin 1.8.22.

To clarify, maybe they can use 1.8.22. What doesn't work is continuing to use 1.7.10, even though that's what every other package we have that uses the bom workaround thing is doing.

stuartmorgan:
And to put some numbers to that:
- 7 of our plugins currently use the `implementation(platform("org.jetbrains.kotlin:kotlin-bom:1.8.22"))` trick, with either 1.8.10 or 1.8.22
- **0** of our plugins (other than a Pigeon internal test app) set `ext.kotlin_version` to something newer than `1.7.10`.

So in seven other plugins, just adding the bom line seems to be enough, but here we also need to bump the example app's Kotlin version.

reidbaker:
IIRC there was an issue fixed in kotlin 1.8.* that fixed situations where when you had multiple kotlin versions between the app and libraries that you would have duplicate files and the work around was to add the kotlin bom. 

@gmackall might remember the exact versions.

tarrinneal:
Any ideas why the bom isn't working for this pr?

stuartmorgan:
Since it works if we update Kotlin only in the example app, I would just add that to unblock the PR. We have the build-all app ensuring that it works in an app that doesn't set a higher Kotlin version.

stuartmorgan (approved):
LGTM!

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
reidbaker, line 1: Are you stuck on legacy as the prefix? The meaning of legacy can change over time so v1 might be a better prefix.

reidbaker (approved):

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
tarrinneal: I think legacy is pretty fitting, as it is now legacy code. It implies future intent to deprecate the code

reidbaker: The issue is legacy assumes 2 states, Legacy and Current and does not account for a 3rd state to exist at the same time.

stuartmorgan: This is something I don't remember to think about until you raise it in reviews, but I'm trying to internalize it into my normal review process :)

I agree it would have been better not to use that name, but in this case I think it'll end up okay, because this was a big change to an API surface that we very explicitly don't want a lot of churn to, and hasn't been fundamentally changed like this in the entire life of the plugin. It's a very different case than, e.g., Android older API version codepaths, where this has come up before, where we can easily get new ones every year.

reidbaker: Yep I did not view it as a blocker which was why it came paired with an approval. It was just the dismissal that "I think legacy is pretty fitting, as it is now legacy code." that I wanted to expand upon since that misses my entire reason for commenting on the name. 

I should have linked to the style guide. https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-newold-modifiers-in-code

tarrinneal: This code, and the names specifically, can also change any time without breaking anything, as this is just the native platform implementation. If there does end up being another massive overhaul before it's deprecated, we can just change it to something else at that time.

tarrinneal: There's always `LegacyLegacy` ;)

reidbaker: yes there is hahaha 😭

tarrinneal: >  It was just the dismissal that "I think legacy is pretty fitting, as it is now legacy code." that I wanted to expand upon since that misses my entire reason for commenting on the name.

Totally fair, I was missing the expanded context so I appreciate the comment :)

PR #6362, opened 2024-03-21
[camera] Camera web mobile orientation

kristoffer-zliide:
Various orientation-related fixes for camera in mobile browsers, including https://github.com/flutter/flutter/issues/143288 and https://github.com/flutter/flutter/issues/117430. The improved video constraints might also fix the zoom behavior observed in https://github.com/flutter/flutter/issues/112280.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [ ] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #6976, opened 2024-06-24
[camera] Camera web mobile orientation

kristoffer-zliide:
Various orientation-related fixes for camera in mobile browsers, including https://github.com/flutter/flutter/issues/143288 and https://github.com/flutter/flutter/issues/117430. The improved video constraints might also fix the zoom behavior observed in https://github.com/flutter/flutter/issues/112280. Previously closed as https://github.com/flutter/packages/pull/6362.

Fixes https://github.com/flutter/flutter/issues/143288

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/lib/src/camera_controller.dart
+++ b/packages/camera/camera/lib/src/camera_controller.dart
@@ -172,6 +172,8 @@ class CameraValue {
     CameraDescription? description,
     Optional<DeviceOrientation>? previewPauseOrientation,
   }) {
+    assert((isRecordingVideo ?? false) ==
+        (recordingOrientation?.isPresent ?? false));
     return CameraValue(
       isInitialized: isInitialized ?? this.isInitialized,
       errorDescription: errorDescription,
--- a/packages/camera/camera/lib/src/camera_preview.dart
+++ b/packages/camera/camera/lib/src/camera_preview.dart
@@ -55,27 +55,33 @@ class CameraPreview extends StatelessWidget {
   }
 
   bool _isLandscape() {
-    return <DeviceOrientation>[
-      DeviceOrientation.landscapeLeft,
-      DeviceOrientation.landscapeRight
-    ].contains(_getApplicableOrientation());
+    switch (_getApplicableOrientation()) {
+      case DeviceOrientation.landscapeLeft:
+      case DeviceOrientation.landscapeRight:
+        return true;
+      case DeviceOrientation.portraitUp:
+      case DeviceOrientation.portraitDown:
+        return false;
+    }
   }
 
   int _getQuarterTurns() {
-    final Map<DeviceOrientation, int> turns = <DeviceOrientation, int>{
-      DeviceOrientation.portraitUp: 0,
-      DeviceOrientation.landscapeRight: 1,
-      DeviceOrientation.portraitDown: 2,
-      DeviceOrientation.landscapeLeft: 3,
-    };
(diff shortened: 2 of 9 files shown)


packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -172,6 +172,8 @@ class CameraValue {
     CameraDescription? description,
     Optional<DeviceOrientation>? previewPauseOrientation,
   }) {
+    assert((isRecordingVideo ?? false) ==
```
bparrishMines, line 175: Is this assert in `copyWith` necessary? This is used by other methods to freely change state, so I don't think we should be asserting anything here. Potentially this check is better in the method that changes it.

packages/camera/camera/lib/src/camera_preview.dart
```diff
@@ -44,38 +44,46 @@ class CameraPreview extends StatelessWidget {
   }
 
   Widget _wrapInRotatedBox({required Widget child}) {
-    if (kIsWeb || defaultTargetPlatform != TargetPlatform.android) {
-      return child;
+    if (defaultTargetPlatform == TargetPlatform.android) {
+      if (kIsWeb) {
```
bparrishMines, line 48: Doesn't this only check if the app was compiled for web?: https://api.flutter.dev/flutter/foundation/kIsWeb-constant.html

This should cause this to always be false if `defaultTargetPlatform == TargetPlatform.android` is true.

stuartmorgan:
@kristoffer-zliide Are you still planning on updating this based on the review comments above?

harryterkelsen:
Fixes https://github.com/flutter/flutter/issues/143288

stuartmorgan:

packages/camera/camera/lib/src/camera_controller.dart
kristoffer-zliide: @stuartmorgan [needs this invariant to hold](https://github.com/flutter/packages/commit/8e8091365ea1e3a4eeb141a39cc35a8f4e71a8ee#diff-811158e46b3c8346bd837b73d34dd132a06f8246822b254420ec7df586550789R80), so I thought rather do an assert when at the time invariant is broken, rather than have `CameraValues` floating around that will cause crashes some time in the future. That would certainly helped me at least.

packages/camera/camera/lib/src/camera_preview.dart
kristoffer-zliide: I think this misunderstanding is likely a large part of the issue here. The assumption appears to be that there exists only iOS, Android, (Desktop) web, when in fact there are iOS, Android, iOS web, Android web, desktop web, etc. See the [documentation](https://api.flutter.dev/flutter/foundation/defaultTargetPlatform.html). The commit message for this change says "[camera] Only native Android needs to have its preview rotated." That was my assumption.

kristoffer-zliide:
You're welcome, @stuartmorgan! I _totally_ understand that. Good luck with your project and its 12.500+ open issues.

stuartmorgan:
> Good luck with your project and its 12.500+ open issues.

We have not found that leaving PRs open indefinitely when the authors haven't responded to questions or comments in several months helps reduce the issue count. It just makes the number of open PRs go up over time, making it harder to focus on active PRs, thus our policy of closing PRs when we don't receive any response from the author for an extended period.

PR #6982, opened 2024-06-25
Final refactor of `video_player_android` before `SurfaceProducer#setCallback`.

matanlurey:

Follows the patterns of some of the previous PRs, i.e.

- https://github.com/flutter/packages/pull/6922
- https://github.com/flutter/packages/pull/6908

The changes in this PR are _mostly_ tests, it was extremely difficult to just add more tests to the already very leaky `VideoPlayer` abstraction which had lots of `@VisibleForTesting` methods and other "holes" to observe state. This PR removes all of that, and adds test coverage where it was missing.

Namely it:

- Adds a new class, `VideoAsset`, that builds and configures the media that `ExoPlayer` uses.
- Removes all "testing" state from `VidePlayer`, keeping it nearly immutable.
- Added tests for most of the classes I've added since, which were mostly missing.

That being said, this is a large change. I'm happy to sit down with either of you and walk through it.

---

Opening as a draft for the moment, since there is a pubspec change needing I want to handle first.

### DIFF
--- a/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoAsset.java
+++ b/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoAsset.java
@@ -69,15 +69,15 @@ abstract class VideoAsset {
   /** Streaming formats that can be provided to the video player as a hint. */
   enum StreamingFormat {
     /** Default, if the format is either not known or not another valid format. */
-    Unknown,
+    UNKNOWN,
 
     /** Smooth Streaming. */
-    Smooth,
+    SMOOTH,
 
     /** MPEG-DASH (Dynamic Adaptive over HTTP). */
-    DynamicAdaptive,
+    DYNAMIC_ADAPTIVE,
 
     /** HTTP Live Streaming (HLS). */
-    HttpLive
+    HTTP_LIVE
   }
 }
--- a/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoAsset.java
+++ b/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoAsset.java
@@ -1,7 +1,6 @@
 package io.flutter.plugins.videoplayer;
 
 import android.content.Context;
-
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.OptIn;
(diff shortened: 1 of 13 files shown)


packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoAssetTest.java
```diff
@@ -0,0 +1,109 @@
...
+    DefaultHttpDataSource.Factory httpFactory = mock(DefaultHttpDataSource.Factory.class);
+    when(httpFactory.setUserAgent(anyString())).thenReturn(httpFactory);
+    when(httpFactory.setAllowCrossProtocolRedirects(anyBoolean())).thenReturn(httpFactory);
+    when(httpFactory.setDefaultRequestProperties(anyMap())).thenReturn(httpFactory);
+    return httpFactory;
+  }
+
+  @Test
+  public void remoteVideoByDefaultSetsUserAgentAndCrossProtocolRedirects() {
+    RemoteVideoAsset asset =
```
camsim99, line 64: Should the tests constructing `RemoteVideoAsset`s be using `VideoAsset.fromRemoteUrl`?

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoAssetTest.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Test
+  public void localVideoCreatesMediaItem() {
+    VideoAsset asset = VideoAsset.fromAssetUrl("asset:///asset-key");
+    MediaItem mediaItem = asset.getMediaItem();
+
+    assert mediaItem.localConfiguration != null;
+    assertEquals(mediaItem.localConfiguration.uri, Uri.parse("asset:///asset-key"));
+  }
```
camsim99, line 52: Do you think it's helpful to test that the type of `asset. getMediaSourceFactory` is what we expect?

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoAssetTest.java
```diff
@@ -0,0 +1,109 @@
...
+  public void remoteVideoOverridesUserAgentIfProvided() {
+    Map<String, String> headers = new HashMap<>();
+    headers.put("User-Agent", "FantasticalVideoBot");
+
+    RemoteVideoAsset asset =
+        new RemoteVideoAsset(
+            "https://flutter.dev/video.mp4", VideoAsset.StreamingFormat.Unknown, headers);
+
+    DefaultHttpDataSource.Factory mockFactory = mockHttpFactory();
+    asset.getMediaSourceFactory(ApplicationProvider.getApplicationContext(), mockFactory);
```
camsim99, line 86: Do you think it'd be helpful to test the overridden behavior of `getMediaSourceFactory` for `RemoteVideoAsset`s?

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java
```diff
@@ -5,355 +5,176 @@
...
+ * Unit tests for {@link VideoPlayer}.
+ *
+ * <p>This test suite <em>narrowly verifies</em> that {@link VideoPlayer} interfaces with the {@link
+ * ExoPlayer} interface <em>exactly</em> as it did when the test suite was created. That is, if the
+ * behavior changes, this test will need to change. However, this suite should catch bugs related to
+ * <em>"this is a safe refactor with no behavior changes"</em>.
+ *
+ * <p>It's hypothetically possible to write better tests using {@link
+ * androidx.media3.test.utils.FakeMediaSource}, but you really need a PhD in the Android media APIs
+ * in order to figure out how to set everything up so the player "works".
```
camsim99, line 39: Your disclaimers are the best lol!

camsim99 (approved):
Looks good overall! Left some questions and a nit.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerEventCallbacks.java
```diff
@@ -97,6 +97,7 @@ public void onError(@NonNull String code, @Nullable String message, @Nullable Ob
   @Override
   public void onIsPlayingStateUpdate(boolean isPlaying) {
     Map<String, Object> event = new HashMap<>();
+    event.put("event", "isPlayingStateUpdate");
```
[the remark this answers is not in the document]
matanlurey: This was actually a regression in a previous refactor, we just haven't made another release since they have been internal refactors only. In other words, all events contain an `event` key-value pair and I made a mistake!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoAsset.java
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Returns the configured media source factory, if needed for this asset type.
+   *
+   * @param context application context.
+   * @return configured factory, or {@code null} if not needed for this asset type.
+   */
+  abstract MediaSource.Factory getMediaSourceFactory(Context context);
+
+  /** Streaming formats that can be provided to the video player as a hint. */
+  enum StreamingFormat {
```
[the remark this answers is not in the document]
matanlurey: Thank you! Done.

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoAssetTest.java
matanlurey: Because of the wrapping, I don't think it is possible.

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoAssetTest.java
matanlurey: Good question. It's intentionally using this constructor so that we have access to the `@VisibleForTesting` method, but I think it could be made as or more clear by casting. Let me make that change and add a comment.

(Great suggestion btw, consistency is king)

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoAssetTest.java
matanlurey: So in production, there is a single method, `getMediaSourceFactory(Context)`, which always uses a new `DefaultHttpDataSource.Factory` (the same behavior before this patch).

For testing, we added a way to _not_ call `new DefaultHttpDataSource.Factory`, as we'd otherwise we'd not be able to assert that the HTTP factory is configured as expected.

I suspect I could assert that the factory makes a _real_ HTTP connection by default? I will leave a comment for the next person, and otherwise at least assert the returned object is real.

matanlurey (commented):
Thanks for the thoughtful review!

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoAssetTest.java
camsim99: Ah I see makes sense!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -10,98 +10,59 @@
...
-  public void configureHttpDataSourceFactory(@NonNull Map<String, String> httpHeaders) {
-    final boolean httpHeadersNotEmpty = !httpHeaders.isEmpty();
-    final String userAgent =
-        httpHeadersNotEmpty && httpHeaders.containsKey(USER_AGENT)
-            ? httpHeaders.get(USER_AGENT)
-            : "ExoPlayer";
-
-    unstableUpdateDataSourceFactory(
-        httpDataSourceFactory, httpHeaders, userAgent, httpHeadersNotEmpty);
+    setUpVideoPlayer(exoPlayer);
```
gmackall, line 65: not worth blocking anything but if you agree and want to change:

Seems a bit odd to me to do half the setup of `exoPlayer` out of `setUpVideoPlayer` and half of it in.

gmackall (commented):
LGTM

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
matanlurey: Agreed. This will change (for the better) in the next PR.

PR #6992, opened 2024-06-26
[pigeon] Implement equals for Java data classes

stuartmorgan:
Adds implementations of `equals` and `hashCode` to Java data classes. This is frequently useful for native unit tests of plugins using Pigeon (e.g., when using a mock FlutterApi implementation to check that the expected call is being made with the right arguments).

Part of https://github.com/flutter/flutter/issues/118087

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/pigeon/platform_tests/alternate_language_test_plugin/android/src/main/java/com/example/alternate_language_test_plugin/CoreTests.java
+++ b/packages/pigeon/platform_tests/alternate_language_test_plugin/android/src/main/java/com/example/alternate_language_test_plugin/CoreTests.java
@@ -26,6 +26,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 /** Generated class from Pigeon. */
 @SuppressWarnings({"unused", "unchecked", "CodeBlock2Expr", "RedundantSuppression", "serial"})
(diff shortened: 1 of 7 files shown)


packages/pigeon/platform_tests/alternate_language_test_plugin/android/src/test/java/com/example/alternate_language_test_plugin/AllDatatypesTest.java
```diff
@@ -168,7 +179,7 @@ public void hasValues() {
             .setBoolList(Arrays.asList(new Boolean[] {true, false}))
             .setDoubleList(Arrays.asList(new Double[] {0.5, 0.25, 1.5, 1.25}))
             .setIntList(Arrays.asList(new Long[] {1l, 2l, 3l, 4l}))
-            .setList(Arrays.asList(new int[] {1, 2, 3, 4}))
+            .setList(Arrays.asList(genericList))
```
stuartmorgan, line 182: This change is because an inherent limitation in Java `equals` implementations is that `List` equality only works if each element in the list has a useful implementation of equals, and (as demonstrated in the new generator code) arrays don't have that. So having the generic list's value be a nested *array* makes equality not work. Fixing that would require a bunch of special code generation—we would have to do custom equality checks of all collections to call out to `Arrays.equals` for certain types—for a case that I'm skeptical will ever matter in practice, and is a common behavior of the language anyway.

Since the specific type here didn't actually matter, I opted for just changing to something with a useful `equals` so that the test will pass.

tarrinneal:
Looks like you need to gen/format a file

packages/pigeon/platform_tests/alternate_language_test_plugin/android/src/main/java/com/example/alternate_language_test_plugin/CoreTests.java
```diff
@@ -318,6 +319,58 @@ public void setMap(@NonNull Map<Object, Object> setterArg) {
...
+              list,
+              stringList,
+              intList,
+              doubleList,
+              boolList,
+              map);
+      __pigeon_result = 31 * __pigeon_result + Arrays.hashCode(aByteArray);
+      __pigeon_result = 31 * __pigeon_result + Arrays.hashCode(a4ByteArray);
+      __pigeon_result = 31 * __pigeon_result + Arrays.hashCode(a8ByteArray);
+      __pigeon_result = 31 * __pigeon_result + Arrays.hashCode(aFloatArray);
```
tarrinneal, line 370: Should we make this a helper?

tarrinneal (approved):
a questionable nit, make the gen code cleaner, but the generator messier?

packages/pigeon/platform_tests/alternate_language_test_plugin/android/src/main/java/com/example/alternate_language_test_plugin/CoreTests.java
stuartmorgan: Doing it inline like this seems to be the stand way I've seen it in other Java code. We'd still need the assignment, so we'd be adding a function call just to abstract away `31 * __pigeon_result +` which seems like a lot of overhead.

PR #6995, opened 2024-06-26
[tools] Fix vm test requirement

stuartmorgan:
The logic for handling Dart unit test `test_on` directives was incorrect, causing it to skip packages that required vm testing, even when run in vm mode. This PR:
- Adds the missing tests for false negatives.
- Reworks the logic to explicitly fail for anything that isn't one of the exact patterns we are expecting, to make it much harder to re-introduce a bug like this in the future.

### DIFF
--- a/script/tool/lib/src/dart_test_command.dart
+++ b/script/tool/lib/src/dart_test_command.dart
@@ -11,6 +11,15 @@ import 'common/plugin_utils.dart';
 import 'common/pub_utils.dart';
 import 'common/repository_package.dart';
 
+const int _exitUnknownTestPlatform = 3;
+
+enum _TestPlatform {
+  // Must run in the command-line VM.
+  vm,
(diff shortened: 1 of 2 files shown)


stuartmorgan:
Hopefully we won't find latent test failures that this was hiding :|

tarrinneal:
looks like there's more non-test test failures

tarrinneal (approved):

stuartmorgan:
The package dependencies in `flutter_migrate` had become stale since we weren't running the tests and getting failures on rolls the way we did with other packages.

stuartmorgan:
Verified in CI output that this does fix the thing that caused us to find this in the first place:

```
packages/pigeon - ran
```

(It was incorrectly `skipped` before.)

stuartmorgan:
The macOS infra flake is getting pretty bad :(

PR #6998, opened 2024-06-27
Update navigation.md

soomrozaid:
Added missing trailing single-quote.

### DIFF
--- a/packages/go_router/doc/navigation.md
+++ b/packages/go_router/doc/navigation.md
@@ -12,7 +12,7 @@ build(BuildContext context) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


stuartmorgan (changes requested):
Thanks for the contribution!

The right way to fix this is to change this file to [use code excerpts](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#readme-code), so that we know the code is correct because it's actual code, as that addresses the underlying problem that led to the typo even being possible, instead of this one symptom.

chunhtai:
closing this pr for now, feel free to reopen when you have time to work on this pr again

PR #7003, opened 2024-06-28
150916: Uri APIs should assert when hasScheme is false

moshe5745:
Assert the uri for having hasSchema in canLaunchUrl API.

150916


stuartmorgan:
Thanks for the contribution, but as noted in the issue adding a case where the API throws is a breaking change, and that would be done as part of a larger set of other breaking changes which we don't have planned for the near future.

(In the future, please don't delete the checklist that's in the PR description; it's there because those steps are required.)

moshe5745:
So how are all the breaking changes merged? Is it a combination of multiple PRs that are merged together or one big merge?

stuartmorgan:
See https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#breaking-changes

PR #7014, opened 2024-06-28
[ci] Adds @matanlurey to some Android CODEOWNERS until Impeller is enabled.

matanlurey:
I.e. until https://github.com/flutter/flutter/issues/151018 is completed.

I know this is not foolproof, but I'm hoping it at least keeps me on top of whatever needs to change in the meantime.


ditman (approved):
LGTM!

PR #7018, opened 2024-06-30
[path_provider_windows] add getExternalStoragePaths()

squarebrakets:
This pr allows to get list of drives in a windows system using getExternalStorageDirectories();

https://github.com/flutter/flutter/issues/151059
## Pre-launch Checklist

- [X] I signed the [CLA].


packages/path_provider/path_provider_windows/lib/src/path_provider_windows_real.dart
```diff
@@ -262,4 +262,59 @@ class PathProviderWindows extends PathProviderPlatform {
     }
     return directory.path;
   }
+
+  @override
+  Future<List<String>?> getExternalStoragePaths(
+      {StorageDirectory? type}) async {
+    if (type == null) {
+      return _getLogicalDrives();
```
stuartmorgan, line 270: This doesn't seem to correspond to the Android implementation, which is the de-facto standard for the behavior of the method (since it has previously been Android-only). From the Android docs:

> Returns absolute paths to application-specific directories on all shared/external storage devices where the application can place persistent files it owns. These files are internal to the application, and not typically visible to the user as media.

This does not describe a top-level drive.

> Shared storage devices returned here are considered a stable part of the device, including physical media slots under a protective cover. The returned paths do not include transient devices, such as USB flash drives connected to handheld devices.

This does not appear to be true of the implementation here.

stuartmorgan (commented):
Thanks for the contribution!

Per comments below, it's not clear to me that the implementation here matches the expected behavior of this method.

packages/path_provider/path_provider_windows/lib/src/path_provider_windows_real.dart
squarebrakets: I assumed getExternalStorageDirectories() is to list storage drive roots to directly list users Storage. Last time I checked 
by default this function returns Phone and SD card roots. With proper permission this allows to list SD Card and Phone storage.  According to.
 https://github.com/flutter/packages/blob/d2705fb82b2605633e8528c053148128cc202327/packages/path_provider/path_provider/lib/path_provider.dart#L186-L188

So to reflect that here getLogicalDrive is being used since default root directory is a C drive. User may opt to add custom partiion to their existing drive or may have multiple drives internally insided the machine or a SD card slot similarly to Phones' SIM/SD card slot.

squarebrakets: If the androids behavior I have seen is unexpected then getLogicalDrive has to be removed

cbracken (changes requested):
It would be useful to understand the underlying problem you're trying to solve with this PR. The idea of "external storage paths" as it exists on Android doesn't really map well to Windows.

If you wouldn't mind describing what you're trying to use this for, perhaps there's an alternate solution.

squarebrakets:
> It would be useful to understand the underlying problem you're trying to solve with this PR. The idea of "external storage paths" as it exists on Android doesn't really map well to Windows.
> 
> If you wouldn't mind describing what you're trying to use this for, perhaps there's an alternate solution.

I realize this wouldn't be straight forward to just list drives with getExternalStorageDirs(). So I am closing this. If however removing getLogicalDrives function leads to any meaningful change I will reopen this pull.

squarebrakets:
> It would be useful to understand the underlying problem you're trying to solve with this PR. The idea of "external storage paths" as it exists on Android doesn't really map well to Windows.
> 
> If you wouldn't mind describing what you're trying to use this for, perhaps there's an alternate solution.
I realize LogicalDrives deosn't really map to Androids storage media. So removing getLogicalDrives

stuartmorgan:
> > It would be useful to understand the underlying problem you're trying to solve with this PR. The idea of "external storage paths" as it exists on Android doesn't really map well to Windows.
> > If you wouldn't mind describing what you're trying to use this for, perhaps there's an alternate solution.
> 
> I realize LogicalDrives deosn't really map to Androids storage media. So removing getLogicalDrives

The use case you provided in the associated issue is "Manage users drives from a flutter app"; that's the part of the PR you removed, so it's still not clear what the use case is here. Please provide (in the issue) the use case that is addressed by this PR, and an explanation of why `getPath` doesn't handle that case.

squarebrakets:
> > > It would be useful to understand the underlying problem you're trying to solve with this PR. The idea of "external storage paths" as it exists on Android doesn't really map well to Windows.
> > > If you wouldn't mind describing what you're trying to use this for, perhaps there's an alternate solution.
> > 
> > 
> > I realize LogicalDrives deosn't really map to Androids storage media. So removing getLogicalDrives
> 
> The use case you provided in the associated issue is "Manage users drives from a flutter app"; that's the part of the PR you removed, so it's still not clear what the use case is here. Please provide (in the issue) the use case that is addressed by this PR, and an explanation of why `getPath` doesn't handle that case.

 getPath would handle these cases.  So closing this.

PR #7039, opened 2024-07-01
[camera] Clean up `maxDuration` code

stuartmorgan:
`maxVideoDuration` was added to the platform interface a long time ago in preparation for adding that feature, but the other parts were never landed. The previous state was:
- It has never been implemented for iOS or Android
- It has never been settable from the app-facing package, so is always null unless someone uses the platform interface directly, which we don't consider a supported use case.
- It cannot be implemented in the CameraX Android implementation.
- It was implemented for Windows and web because when those platforms were added much later, nobody realized that the parameter was unused.

There is no compelling need for this feature, as clients of the plugin can simply set their own timer to stop recording. Given that, rather than leave the confusing partial state, this marks the option as deprecated at the platform interface layer and warns implementers that it can be ignored. It also removes the implementations from Windows and web in order to reduce implementation complexity, since that code was not reachable from the app-facing API.

This does not consider the Windows and web changes to be breaking, even though they arguably could be, because we do not expect clients to be calling platform interface methods directly.

Fixes https://github.com/flutter/flutter/issues/150959

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
(CI will make me split out the platform interface change, but I wanted to start with a combo PR to provide the whole picture.)

packages/camera/camera_android_camerax/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## 0.6.5+6
```
camsim99, line 1:
```suggestion
## 0.6.5+7
```

camsim99 (approved):
Android LGTM!

packages/camera/camera_android_camerax/CHANGELOG.md
stuartmorgan: Oops, adjusted some minor version changes back down to bugfix manually, and messed this one up!

bparrishMines (approved):
LGTM 

I'm not recently familiar with c++, but the code here looks fine to me.

I'm in support of removing it because it is in a partial state, it's not in the app-facing, and we don't have any current plans to add full support for it. 

But I do think this is a feature that should be supported in the future. Although you can set a `Timer` in Dart, its not really possible to get a video of an exact length by using it. Mostly because starting and stopping a camera are typically done on their own threads. It also looks like camerax and iOS do support this feature (but seemingly are not compatible with our current implementations):

https://developer.android.com/reference/androidx/camera/video/OutputOptions#getDurationLimitMillis()
https://developer.apple.com/documentation/avfoundation/avcapturefileoutput/1387390-maxrecordedduration

stuartmorgan:
> Although you can set a `Timer` in Dart, its not really possible to get a video of an exact length by using it. Mostly because starting and stopping a camera are typically done on their own threads.

I'd want to get user feedback about cases where that's actually a concern; I would not expect that sub-second variation in max times would matter in real world usage.

hellohuanlin (approved):

PR #7040, opened 2024-07-01
[google_maps_flutter] Moves Java->Dart calls to Pigeon

stuartmorgan:
Converts all of the Java->Dart calls from raw method channels to Pigeon. This finishes conversion of all of the invocations in `google_maps_flutter_android`

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/CirclesController.java
```diff
@@ -67,7 +67,15 @@ boolean onCircleTap(String googleCircleId) {
...
-    methodChannel.invokeMethod("circle#onTap", Convert.circleIdToJson(circleId));
+    flutterApi.onCircleTap(
+        circleId,
+        new Messages.VoidResult() {
+          @Override
+          public void success() {}
+
+          @Override
+          public void error(@NonNull Throwable error) {}
+        });
```
stuartmorgan, line 78: These are because we don't currently have a way in Pigeon to ignore responses. Longer term I think we'll want to convert this over to event channels though; I'm not sure why maps wasn't written using them in the first place, since a stream of one-way calls without responses are the use cases event channels were written for IIUC.

(I'm not doing that conversion as part of this because we don't have event channel support for Pigeon yet; probably in Q3 sometime.)

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -426,29 +424,6 @@ static Object polylineIdToJson(String polylineId) {
...
-  static Object circleIdToJson(String circleId) {
-    if (circleId == null) {
-      return null;
-    }
-    final Map<String, Object> data = new HashMap<>(1);
-    data.put("circleId", circleId);
-    return data;
-  }
-
-  static Map<String, Object> tileOverlayArgumentsToJson(
```
stuartmorgan, line 438: I can change this to stay here and do a Pigeon build if you prefer, but since we have to pass all the arguments individually anyway (vs passing a single object to convert) it seemed easier to just inline it.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -464,36 +439,6 @@ static LatLng latLngFromPigeon(Messages.PlatformLatLng latLng) {
     return new LatLng(latLng.getLatitude(), latLng.getLongitude());
   }
 
-  static Object clusterToJson(String clusterManagerId, Cluster<MarkerBuilder> cluster) {
```
stuartmorgan, line 467: This is pure removal because the Pigeon version (just below) was added in a previous PR.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -182,10 +190,6 @@ void init() {
     mapView.getMapAsync(this);
   }
 
-  private CameraPosition getCameraPosition() {
```
stuartmorgan, line 185: Opportunistic cleanup; the IDE flagged that this was unused.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/TileProviderController.java
```diff
@@ -66,37 +65,43 @@ Tile getTile() {
         return TileProvider.NO_TILE;
       }
       try {
-        return Convert.interpretTile(result);
+        if (result == null) {
```
stuartmorgan, line 68: Opportunistic safety fix; the IDE was flagging that we weren't handling null `result` here.

packages/google_maps_flutter/google_maps_flutter_android/test/google_maps_flutter_android_test.dart
```diff
@@ -492,17 +474,121 @@ void main() {
...
+    callbackHandler.onMarkerDragStart(dragStartId, fakePosition);
+    callbackHandler.onMarkerDrag(dragId, fakePosition);
+    callbackHandler.onMarkerDragEnd(dragEndId, fakePosition);
+
+    expect((await markerDragStartStream.next).value.value, equals(dragStartId));
+    expect((await markerDragStream.next).value.value, equals(dragId));
+    expect((await markerDragEndStream.next).value.value, equals(dragEndId));
+  });
+
+  test('markers send tap events to correct stream', () async {
```
stuartmorgan, line 487: Longer because I added a bunch of missing test coverage.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/CirclesController.java
Rexios80, line 1: @stuartmorgan I'm working on the heatmaps PR and maybe I'm missing something, but I believe the `getCircleId` method below is invalid. It should be:

```java
private static String getCircleId(Messages.PlatformCircle circle) {
    Map<String, Object> circleMap = (Map<String, Object>) circle.getJson();
    return (String) circleMap.get("circleId");
}
```

The equivalent methods in other controllers might also be invalid.

Rexios80: It looks like `changeCircle` might also be invalid:

```java
Convert.interpretCircleOptions(circle.getJson(), circleController);
```

Notice the `getJson()` call. Again I might just be misreading this code, but did the tests pass on this? Is this missed in coverage somehow?

stuartmorgan: It looks like you're talking about the previous PR, not this one. `changeCircles` looks wrong, I'll have a PR shortly to fix it, and check the others (I'm sure I did at least several of the other ones correctly, but I'll check them all).

Nothing lands that doesn't pass our CI, so there's probably missing coverage here. And unfortunately because all the JSON code is untyped, the compile doesn't catch this kind of bug—which is a motivation for the change in general, but unfortunately can still bite us during the transition.

stuartmorgan: https://github.com/flutter/packages/pull/7056 addresses this.

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/CirclesController.java
stuartmorgan: Now that I think of it, I'm not sure when event channels were added, so this code may just predate them.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/CirclesController.java
```diff
@@ -67,7 +67,15 @@ boolean onCircleTap(String googleCircleId) {
...
     }
-    methodChannel.invokeMethod("circle#onTap", Convert.circleIdToJson(circleId));
+    flutterApi.onCircleTap(
+        circleId,
+        new Messages.VoidResult() {
+          @Override
+          public void success() {}
+
+          @Override
+          public void error(@NonNull Throwable error) {}
```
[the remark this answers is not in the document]
stuartmorgan: The main reason I didn't add it in this PR is that I know we have a class of issues around lifetime mismatch between native and Dart in Google Maps, and I worry that just adding logging here without deeply investigating those first would cause a bunch of spam where the call fails only because the instance has been eliminated on the Dart side, in which case silently dropping them is better than logging.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/PolylinesController.java
```diff
@@ -70,7 +71,15 @@ boolean onPolylineTap(String googlePolylineId) {
     if (polylineId == null) {
       return false;
     }
-    methodChannel.invokeMethod("polyline#onTap", Convert.polylineIdToJson(polylineId));
+    flutterApi.onPolylineTap(
+        polylineId,
+        new Messages.VoidResult() {
```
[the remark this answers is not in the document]
stuartmorgan: Done!

PR #7048, opened 2024-07-02
Update espresso dependencies

reidbaker:
- **Update espresso dependencies**
fixes #151188

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/espresso/CHANGELOG.md
+++ b/packages/espresso/CHANGELOG.md
@@ -1,7 +1,8 @@
-## 0.3.1
+## 0.4.0
 
 * Updates androidx.test.espresso version to 3.6.1.
 * Updates androidx.test to 1.6.1.
+* Removes androidx.test.annotation.ExperimentalTestApi.
 
 ## 0.3.0+10
 
--- a/packages/espresso/CHANGELOG.md
+++ b/packages/espresso/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.3.1
+
+* Updates espresso version to 3.6.1
+* Updates androidx.test to 1.6.1
+
 ## 0.3.0+10
 
 * Removes additional references to v1 Android embedding.
--- a/packages/espresso/CHANGELOG.md
+++ b/packages/espresso/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.3.1
 
-* Updates espresso version to 3.6.1
-* Updates androidx.test to 1.6.1
+* Updates androidx.test.espresso version to 3.6.1.
+* Updates androidx.test to 1.6.1.
 
 ## 0.3.0+10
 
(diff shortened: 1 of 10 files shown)


packages/espresso/example/android/app/build.gradle
```diff
@@ -62,28 +62,28 @@ dependencies {
     implementation "androidx.multidex:multidex:2.0.1"
 
     // Core library
-    api 'androidx.test:core:1.2.0'
+    api 'androidx.test:core:1.6.1'
```
stuartmorgan, line 65: IIRC this is `api` because it's actually re-exposed as part of the public (native) API surface, so if anything here is breaking we'd want to version accordingly (although since it's a minor version change, maybe it's not breaking?)

packages/espresso/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.3.1
+
+* Updates espresso version to 3.6.1
```
stuartmorgan, line 3: Nit: these are missing periods at the end.

stuartmorgan, line 3: Maybe call this `androidx.test.espresso` instead of just `espresso` to make it less confusing, since `espresso` is also the package name?

stuartmorgan (approved):
LGTM with nits

packages/espresso/example/android/app/build.gradle
reidbaker: This value has not change in 5 years since the creation of this plugin. 
I am going to assume it is the most breaking change we can make.

I will update this to have more obviously breaking version number once I get the tests passing. 
https://github.com/flutter/packages/commit/dcace54ce86539c9535c4890759dd8a538624fbc#diff-af20d9c2b386a0c999752ae955c2295514c30f75e7aa6255a47e2c6114340ef9

gmackall (approved):

reidbaker:
https://developer.android.com/jetpack/androidx/releases/test Annotation 1.1.0-alpha03 "unhide ExperimentalTestApi"

packages/espresso/android/src/main/java/androidx/test/espresso/flutter/EspressoFlutter.java
```diff
@@ -100,7 +99,6 @@ private WidgetInteraction(
      * @param widgetActions one or more actions that shall be performed. Cannot be {@code null}.
      * @return this interaction for further perform/verification calls.
      */
-    @ExperimentalTestApi()
```
reidbaker, line 103: I am struggling to find the original documentation for this annotation. There is an identically named class in compose that is exposed but I think it is for compose related apis. 

This annotation either is marking something experimental from us which we dont need anymore or it was required as an opt in feature and all the features have graduated.

reidbaker:
Not going to merge before a US holiday weekend. Will merge on monday

stuartmorgan (approved):
Still LGTM

PR #7050, opened 2024-07-03
[video_player] Fix appleos crash when set negative video composition

lxhlzyh:
iOS/macOS crash when set negative video composition, we should catch this exception.

Fixes https://github.com/flutter/flutter/issues/151031

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
+++ b/packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
@@ -291,7 +291,14 @@ NS_INLINE CGFloat radiansToDegrees(CGFloat radians) {
                 [self getVideoCompositionWithTransform:self->_preferredTransform
                                              withAsset:asset
                                         withVideoTrack:videoTrack];
-            item.videoComposition = videoComposition;
+            // Invalid values of video composition will throws an exception
+            // (https://github.com/flutter/flutter/issues/151031).
+            // When there is a problem with the parameters of video composition, set nil.
+            @try {
+              item.videoComposition = videoComposition;
+            } @catch (NSException *exception) {
+              item.videoComposition = nil;
+            }
           }
         };
         [videoTrack loadValuesAsynchronouslyForKeys:@[ @"preferredTransform" ]
--- a/packages/video_player/video_player_avfoundation/CHANGELOG.md
+++ b/packages/video_player/video_player_avfoundation/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.6.2
+
+* Fixes the iOS/macOS crash when set negative video composition.
+
 ## 2.6.1
 
 * Adds files to make include directory permanent.
--- a/packages/video_player/video_player_avfoundation/pubspec.yaml
+++ b/packages/video_player/video_player_avfoundation/pubspec.yaml
@@ -2,7 +2,7 @@ name: video_player_avfoundation
 description: iOS and macOS implementation of the video_player plugin.
 repository: https://github.com/flutter/packages/tree/main/packages/video_player/video_player_avfoundation
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+video_player%22
-version: 2.6.1
+version: 2.6.2
 
 environment:
   sdk: ^3.2.3


packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -291,7 +291,14 @@ - (instancetype)initWithPlayerItem:(AVPlayerItem *)item
                 [self getVideoCompositionWithTransform:self->_preferredTransform
                                              withAsset:asset
                                         withVideoTrack:videoTrack];
-            item.videoComposition = videoComposition;
+            // Invalid values of video composition will throws an exception
+            // (https://github.com/flutter/flutter/issues/151031).
+            // When there is a problem with the parameters of video composition, set nil.
```
stuartmorgan, line 296: Where is the invalid video composition value coming from? If it's user input, this is very late in the code to be enforcing anything about it.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -291,7 +291,14 @@ - (instancetype)initWithPlayerItem:(AVPlayerItem *)item
...
+            @try {
+              item.videoComposition = videoComposition;
+            } @catch (NSException *exception) {
```
stuartmorgan, line 299: Unless it's impossible to know in advance that a value is invalid for some reason, creating an exception and then catching it is not correct behavior. Invalid arguments are a programming error, not a runtime error, so the plugin code should never be calling this with invalid arguments in the first place.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -291,7 +291,14 @@ - (instancetype)initWithPlayerItem:(AVPlayerItem *)item
...
+              item.videoComposition = nil;
```
stuartmorgan, line 300: Errors, especially if they are ultimately caused by input from plugin clients, should never be silently ignored.

stuartmorgan (changes requested):
Thanks for the contribution!

> - [x] I added new tests to check the change I am making, or this PR is [test-exempt](https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests).

Where is this new test?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
lxhlzyh: In this plugin, invalid parameters may be caused by rederSize of videoCompositiom, but there are multiple cases in the documentation. However if we know that the parameter is abnormal, should we still let it be created successfully? I don't think setting videoComposition is necessary for the creation process.

lxhlzyh:
> 
> > * [x]  I added new tests to check the change I am making, or this PR is [test-exempt](https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests).
> 
> Where is this new test?
The problem was found in the crash log, so there was no video source to write a replay test, I don't know what kind of test to write.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
lxhlzyh: I agree, as i said earlier,  if we know that the parameter is abnormal, should we still let it be created successfully? If the creation process is to continue, there should be no other way than to set fake data and not set it at all. Or just return an error back to flutter?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
lxhlzyh: However, we should check the validity of videoComposition in this function. Still, we want to determine if the creation process should be interrupted.https://github.com/flutter/packages/blob/754de1918a339270b70971b6841cf1e04dd71050/packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m#L211

stuartmorgan: > there are multiple cases in the documentation

Please provide a link to the documentation that you are referring to here, so that we are all working from the same information.

> However if we know that the parameter is abnormal, should we still let it be created successfully? I don't think setting invalid videoComposition is necessary for the creation process.

Please see my other unanswered question about whether the invalid values are coming from. Until there's an answer to that, there's no way to answer questions about how they should be handled. Bad values that come from plugin clients are very different from bad values that come from uncontrolled external video date, both of which are very different from bad values coming from a programming error in the plugin itself.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
stuartmorgan: As above; we cannot answer this question without more information about the cause of the error.

stuartmorgan:
> The problem was found in the crash log report, so there was no video source to write a replay test, I don't know what kind of test to write.

If we don't know enough to construct a plausible test case that would reproduce the exception in a test, then we don't know enough to fix the issue.

If we do, then then the test should do that.

lxhlzyh:
> > The problem was found in the crash log report, so there was no video source to write a replay test, I don't know what kind of test to write.
> 
> If we don't know enough to construct a plausible test case that would reproduce the exception in a test, then we don't know enough to fix the issue.
> 
> If we do, then then the test should do that.
> 
> Either way, please don't check a checkbox for something you haven't actually done, as it makes things more difficult for reviewers. The checklist should reflect the actual state of the PR.

Thank you, i miss something.

stuartmorgan:
From triage: @lxhlzyh are you still planning on updating this PR per the discussion above?

stuartmorgan:

PR #7073, opened 2024-07-08
[path_provider] Remove `win32`

stuartmorgan:
Per https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#dependencies, we generally do not want third-party dependencies. `path_provider` in particular is a key part of the ecosystem (well over 1,000 packages depend directly on it, and transitively it's probably several times that at least), and thus the maintenance and security considerations are particularly acute.

This eliminates the dependency on `win32`, a large third-party dependency, in favor of direct FFI code written from scratch using the official Win32 reference documentation from Microsoft as the source. The only behavioral change that should result here is that the exceptions thrown in failure cases have changed, but they were never documented, and were entirely platform-specific, so it's relatively unlikely that people will be broken by that. (As noted in a TODO, the longer term solution is to provide real exceptions for this package, and use those across platforms.)

Fixes https://github.com/flutter/flutter/issues/130940

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/path_provider/path_provider_windows/lib/src/folders.dart
+++ b/packages/path_provider/path_provider_windows/lib/src/folders.dart
@@ -10,7 +10,7 @@
 /// method in the [PathProvidersWindows] class to retrieve a known folder from
 /// Windows.
 // These constants come from
-// https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid
+// https://learn.microsoft.com/windows/win32/shell/knownfolderid
 class WindowsKnownFolder {
   /// The file system directory that is used to store administrative tools for
   /// an individual user. The MMC will save customized consoles to this
--- a/packages/path_provider/path_provider_windows/lib/src/folders.dart
+++ b/packages/path_provider/path_provider_windows/lib/src/folders.dart
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'package:win32/win32.dart';
-
 // ignore_for_file: non_constant_identifier_names
 
 // ignore: avoid_classes_with_only_static_members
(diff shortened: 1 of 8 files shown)


packages/path_provider/path_provider_windows/lib/src/folders.dart
```diff
@@ -2,242 +2,249 @@
...
-
 // ignore_for_file: non_constant_identifier_names
 
 // ignore: avoid_classes_with_only_static_members
 /// A class containing the GUID references for each of the documented Windows
 /// known folders. A property of this class may be passed to the `getPath`
 /// method in the [PathProvidersWindows] class to retrieve a known folder from
 /// Windows.
+// These constants come from
+// https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid
```
loic-sharma, line 13: You should omit the language code from Microsoft Learn URLs so that it redirects to the user's preferred language if possible.

```suggestion
// https://learn.microsoft.com/windows/win32/shell/knownfolderid
```

packages/path_provider/path_provider_windows/lib/src/guid.dart
```diff
@@ -0,0 +1,51 @@
...
+
+import 'dart:ffi';
+import 'dart:typed_data';
+
+/// Representation of the Win32 GUID struct.
+// For the layout of this struct, see
+// https://learn.microsoft.com/en-us/windows/win32/api/guiddef/ns-guiddef-guid
+@Packed(4)
+base class GUID extends Struct {
+  /// Navite Data1 field.
```
loic-sharma, line 13: Here and below:

```suggestion
  /// Native Data1 field.
```

packages/path_provider/path_provider_windows/lib/src/guid.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:ffi';
+import 'dart:typed_data';
+
+/// Representation of the Win32 GUID struct.
+// For the layout of this struct, see
+// https://learn.microsoft.com/en-us/windows/win32/api/guiddef/ns-guiddef-guid
```
loic-sharma, line 10:
```suggestion
// https://learn.microsoft.com/windows/win32/api/guiddef/ns-guiddef-guid
```

packages/path_provider/path_provider_windows/lib/src/path_provider_windows_real.dart
```diff
@@ -144,9 +146,9 @@ class PathProviderWindows extends PathProviderPlatform {
         pathPtrPtr,
       );
 
-      if (FAILED(hr)) {
+      if (hr != S_OK) {
```
loic-sharma, line 149: Windows can set bits in the `HRESULT`, a successful status might not be equal to `S_OK`.

I would consider adding a `FAILED` helper whose implementation is `hr < 0`.

See:
1. https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-failed
2. https://stackoverflow.com/a/12852668

packages/path_provider/path_provider_windows/lib/src/guid.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  /// Navite Data4 field.
+  // This should be an eight-element byte array, but there's no such annotation.
+  @Uint64()
+  external int data4;
+
+  /// Parses a GUID string, with optional enclosing "{}"s and optional "-"s,
+  /// into data.
+  void parse(String guid) {
+    final String hexOnly = guid.replaceAll(RegExp(r'[{}-]'), '');
```
loic-sharma, line 33: `package:win32`'s `GUID.parse` requires the `{}`s and `-`s. Should we do that as well?

loic-sharma (approved):
LGTM, nice work!

packages/path_provider/path_provider_windows/lib/src/guid.dart
stuartmorgan: I'm removing those symbols anyway to make the implementation simple, and since https://en.wikipedia.org/wiki/Universally_unique_identifier#Textual_representation says they are all valid I figured I would just accept all of them and relaxed the API accordingly. I didn't look at the `win32` implementation so I don't know if there's a reason they are stricter; the only one I could think of would be efficiency, and I'm pretty confident that it's never going to matter in this use case. (If efficiency were ever actually important the way to go would be to make an API that's not string-based in the first place, and constants that built the GUIDs directly.)

packages/path_provider/path_provider_windows/lib/src/folders.dart
stuartmorgan: Done throughout.

packages/path_provider/path_provider_windows/lib/src/guid.dart
stuartmorgan: Done.

packages/path_provider/path_provider_windows/lib/src/path_provider_windows_real.dart
stuartmorgan: Done, thanks for catching that.

packages/path_provider/path_provider_windows/lib/src/guid.dart
loic-sharma: Sounds good to me, thanks for the thoughtful reply

PR #7079, opened 2024-07-08
[google_maps_flutter] Semi-convert remaining iOS host API calls to Pigeon

stuartmorgan:
Does a "shallow" Pigeon conversion of the remaining host API calls in the iOS implementation. All of the actual method calls are now Pigeon, but for the most part the data structures are not yet converted, and the Pigeon representations of the map objects are instead placeholders that currently just wrap the existing JSON serialization. This is done for a few reasons:
- Keeps the incremental PR relatively small and easy to understand.
- Quickly gets us to a state where any new APIs added will automatically use Pigeon, reducing further accumulation of technical debt.
- Avoids duplication of handling code until https://github.com/flutter/flutter/issues/150631 is resolved. As noted in a TODO added in this PR, almost all of the data structures that are passed in these methods are also passed through the PlatformView factory constructor, and Pigeon doesn't yet support using the Pigeon codec in non-Pigeon-generated code, so we would need to have both the structured *and* JSON handler for all of these objects if we converted them now.

Future PRs will be able to incrementally convert each map object to a structured form. Converting the Flutter APIs (Java->Dart) will also be done in a follow-up, to limit the scope of this PR.

This is the iOS equivalent of https://github.com/flutter/packages/pull/6980, and the Dart code is largely the same.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/FLTGoogleMapJSONConversionsConversionTests.m
```diff
@@ -188,18 +183,18 @@ - (void)testMapViewTypeFromTypeValue {
   XCTAssertEqual(kGMSTypeNone, [FLTGoogleMapJSONConversions mapViewTypeFromTypeValue:@5]);
 }
 
-- (void)testCameraUpdateFromChannelValueNewCameraPosition {
+- (void)testcameraUpdateFromArrayNewCameraPosition {
```
jmagman, line 186: nit
```suggestion
- (void)testCameraUpdateFromArrayNewCameraPosition {
```

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/FLTGoogleMapJSONConversionsConversionTests.m
```diff
@@ -213,10 +208,10 @@ - (void)testCameraUpdateFromChannelValueNewCameraPosition {
 // verified.
 //
 // The code in below test uses the 2nd approach.
-- (void)skip_testCameraUpdateFromChannelValueNewLatLong {
+- (void)skip_testcameraUpdateFromArrayNewLatLong {
```
jmagman, line 211:
```suggestion
- (void)skip_testCameraUpdateFromArrayNewLatLong {
```

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
```diff
@@ -503,19 +337,23 @@ - (void)setMyLocationButtonEnabled:(BOOL)enabled {
   self.mapView.settings.myLocationButton = enabled;
 }
 
+/// Sets the map style, returing any error string as well as storing that error in `mapStyle` for
+/// later access.
 - (NSString *)setMapStyle:(NSString *)mapStyle {
```
jmagman, line 342: Is the return string needed any more since you added the `styleError` setter side-effect?

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/messages.g.h
```diff
@@ -42,6 +157,109 @@ NS_ASSUME_NONNULL_BEGIN
...
+/// @return `nil` only when `error != nil`.
+- (nullable NSNumber *)
+    isShowingInfoWindowForMarkerWithIdentifier:(NSString *)markerId
+                                         error:(FlutterError *_Nullable *_Nonnull)error;
+/// Sets the style to the given map style string, where an empty string
+/// indicates that the style should be cleared.
+///
+/// If there was an error setting the style, such as an invalid style string,
+/// returns the error message.
+- (nullable NSString *)setStyleStyle:(NSString *)style
```
jmagman, line 240: Not sure how to avoid this in the generated code, but `setStyleStyle:style` isn't the most clear...

stuartmorgan: Oops, missed that since I let Xcode auto-create all the protocol conformance stubs for me. It's easy to fix in `messages.dart`.

(I'll also file a Pigeon issue, because I thought we'd special-cased the logic to avoid duplication in the default name when it's the same as the end of the method name.)

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
stuartmorgan: Technically no, but I think it's easier to understand `return [self.controller setMapStyle:style];` in the Pigeon method implementation than:

```objc
[self.controller setMapStyle:style];
// This value has to be read after the call above because it's set as a side effect of that call.
return self.styleError;
```

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/FLTGoogleMapJSONConversionsConversionTests.m
stuartmorgan: Fixed throughout.

jmagman (approved):

PR #7085, opened 2024-07-09
[camera] Fix iOS torch mode regression

stuartmorgan:
Fixes a regression from the Pigeon conversion, where converting from a Pigeon flash mode to an OS flash mode now fails instead of returning a magic sentinel value, so the logic to handle it needed to be adjusted, but wasn't in the conversion.

Fixes https://github.com/flutter/flutter/issues/151453

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
I'm not thrilled at adding more OCMock, but it's *everywhere* in the camera tests right now, including other tests that inject a mock capture device, so I'm following the existing patterns. Eventually we need to get this whole plugin off of OCMock, but that'll be a larger refactor that we should do separately from behavioral changes.

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
```diff
@@ -374,9 +374,8 @@ - (void)captureToFileWithCompletion:(void (^)(NSString *_Nullable,
     extension = @"jpg";
   }
 
-  AVCaptureFlashMode avFlashMode = FCPGetAVCaptureFlashModeForPigeonFlashMode(_flashMode);
-  if (avFlashMode != -1) {
-    [settings setFlashMode:avFlashMode];
+  if (_flashMode != FCPPlatformFlashModeTorch) {
```
hellohuanlin, line 377: Is this because iOS doesn't support taking picture in torch mode? Maybe add a comment?

hellohuanlin (approved):

packages/camera/camera_avfoundation/ios/Classes/FLTCam.m
stuartmorgan: It's because torch mode means the flash is forced to on all of the time even when not taking a picture, so there is no picture-level setting for it.

stuartmorgan: Added a comment in the code.

PR #7089, opened 2024-07-09
Test iOS and macOS on Xcode 16 beta 3

jmagman:
Test to see what's failing in beta 3
https://github.com/flutter/flutter/issues/148901

Hit https://github.com/flutter/flutter/issues/151467 and https://github.com/flutter/flutter/issues/150943


harishSignalz:
Xcode 15.4 will support Minimum deployment 12 but this library starts from 9 can you make it minimum os version 12 for iOS?

jmagman:
@harishSignalz can you file an issue describing the problem in full?  This PR isn't a good place to track whatever you're seeing.

PR #7100, opened 2024-07-11
[ios]Fix compile error when conforming UIApplication to Launcher due to MainActor annotation

hellohuanlin:
iOS 18 Beta 3 added `@MainActor @Sendable` for openURL's completion handler: 

```
    @available(iOS 10.0, *)
    open func open(_ url: URL, options: [UIApplication.OpenExternalURLOptionsKey : Any] = [:], completionHandler completion: (@MainActor @Sendable (Bool) -> Void)? = nil)
```

The addition of `@MainActor` is the one that caused the compile error. It was not there for Beta 1. 

This PR we simply use a `DefaultLauncher` wrapper. As expected, passing in a **non-isolated closure** to a function that expects a **main-actor isolated closure** is fine, since non-isolated closure can be called in any isolation context; plus the minimal concurrency checking also silences the Sendable warning.
 
But we should check again in newer betas if we can revert this change. Because this compile error forces some libraries to update for swift concurrency, which seems to be against the whole idea of having 3 different levels of concurrency checking. So I suspect that Apple is still hashing things out and may change it back (either by removing `@MainActor` from signature, or compiler change to allow this conformance under minimal checking) 

Fixes https://github.com/flutter/flutter/issues/151467

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/Launcher.swift
+++ b/packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/Launcher.swift
@@ -24,12 +24,12 @@ final class DefaultLauncher: Launcher {
   func canOpenURL(_ url: URL) -> Bool {
     return UIApplication.shared.canOpenURL(url)
   }
-  
+
   func open(
     _ url: URL,
     options: [UIApplication.OpenExternalURLOptionsKey: Any],
-    completionHandler completion: ((Bool) -> Void)?)
-  {
+    completionHandler completion: ((Bool) -> Void)?
+  ) {
     UIApplication.shared.open(
       url,
       options: options,
--- a/packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/Launcher.swift
+++ b/packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/Launcher.swift
@@ -18,5 +18,21 @@ protocol Launcher {
     completionHandler completion: ((Bool) -> Void)?)
 }
 
-/// Launcher is intentionally a direct passthroguh to UIApplication.
-extension UIApplication: Launcher {}
+// TODO(hellohuanlin): This wrapper is a workaround for iOS 18 Beta 3 where completionHandler is annotated with @MainActor @Sendable, resulting in compile error when conforming UIApplication to Launcher. We should try again in newer betas.
+/// A default URL launcher.
+final class DefaultLauncher: Launcher {
+  func canOpenURL(_ url: URL) -> Bool {
+    return UIApplication.shared.canOpenURL(url)
+  }
+  
+  func open(
+    _ url: URL,
+    options: [UIApplication.OpenExternalURLOptionsKey: Any],
+    completionHandler completion: ((Bool) -> Void)?)
+  {
+    UIApplication.shared.open(
+      url,
+      options: options,
+      completionHandler: completion)
+  }
+}
--- a/packages/url_launcher/url_launcher_ios/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher_ios/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.3.1
+
+* Fixes a compile error when comforming UIApplication to Launcher in iOS 18 Beta 3.
+
 ## 6.3.0
 
 * Adds Swift Package Manager compatibility.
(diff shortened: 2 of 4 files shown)


stuartmorgan:
If we were to just add the annotations to `Launcher` instead, would compiling break in older versions of Xcode?

stuartmorgan:
test-exempt: Cannot be tested with the tooling currently in CI; will automatically be tested after the next major Xcode update in CI.

packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/Launcher.swift
```diff
@@ -18,5 +18,21 @@ protocol Launcher {
     completionHandler completion: ((Bool) -> Void)?)
 }
 
-/// Launcher is intentionally a direct passthroguh to UIApplication.
-extension UIApplication: Launcher {}
+// TODO(hellohuanlin): This wrapper is a workaround for iOS 18 Beta 3 where completionHandler is annotated with @MainActor @Sendable, resulting in compile error when conforming UIApplication to Launcher. We should try again in newer betas.
+/// A default URL launcher.
+final class DefaultLauncher: Launcher {
```
jmagman, line 23: @hellohuanlin `Launcher` is documented to only exist for testing purposes.  
https://github.com/flutter/packages/blob/01527175e38e7ecd9ec180a1f0781580adb82dd5/packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/Launcher.swift#L7-L9

I know you [reviewed the PR](https://github.com/flutter/packages/pull/4753/files#diff-c38ad4a9a3711ebe154cfb06c02721450bbd4d80beedb7115b15c58f31d26deeR136), based on [your `DeviceProtocol` example](https://docs.google.com/document/d/1XsaulkJA6_ZSpM7chkQLhQY25sQqhEMiqHMYzw8H85o/edit?resourcekey=0-_cUjF1c0iBvRLKfV-3gK5A#heading=h.5t23pi6mihh3), that's the idiomatic way you suggest doing this, or is there another way?

hellohuanlin: The idiomatic way is (as shown in wwdc's demo code a few years ago): 
  
(1) for real dependency, extend system type to conform to our protocol (e.g. `extension UIApplication: Launcher`)
(2) for fake dependency, define a custom type that conforms to the same protocol (e.g. `class MockLauncher: Launcher`)

In this case, (1) doesn't work anymore. So we have to define the custom type (e.g. `class DefaultLauncher: Launcher`). 

However, I highly suspect Apple may fix it in newer betas, so that we can revert this PR. As discussed in the PR description, Apple seems to suggest that if a module is not ready to adopt swift concurrency, it should just work as before. So they are not supposed to break us.

hellohuanlin:
> If we were to just add the annotations to `Launcher` instead, would compiling break in older versions of Xcode?

Yes it breaks in older Xcode.

stuartmorgan (approved):
LGTM

stuartmorgan:
(modulo the Swift formatter CI failure)

jmagman (approved):

ciriousjoker:
This seems to be not live on pub.dev yet, the last version was 6.3.0 from 3 months ago. When will this be released?

jmagman:
> This seems to be not live on pub.dev yet, the last version was 6.3.0 from 3 months ago. When will this be released?

@ciriousjoker url_launcher_ios is on 6.3.1 https://pub.dev/packages/url_launcher_ios/changelog

packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/Launcher.swift
ejdrian313: looks like ios18 and xcode update break it...

stuartmorgan:
@ahmedmahershaaban All you need to do is run `flutter pub ugrade`. No pubspec changes are necessary.

xiaoliang-cn:
> @ahmedmahershaaban All you need to do is run `flutter pub ugrade`. No pubspec changes are necessary.
Didn't work

stuartmorgan:
@xiaoliang-cn If you have a situation where `flutter pub upgrade` doesn't give you the latest compatible version of your dependencies, please file an issue with details.

PR #7114, opened 2024-07-12
Modernize all `index.html` files to support WASM

Rexios80:
Fixes https://github.com/flutter/flutter/issues/151663 in preparation for https://github.com/flutter/flutter/issues/151664


Rexios80:
```
No CHANGELOG change found.
If this PR needs an exemption from the standard policy of listing all changes in the CHANGELOG,
comment in the PR to explain why the PR is exempt, and add (or ask your reviewer to add) the
"override: no changelog needed" label.
```

Since none of this affects code that consumers actually run when using these packages in their projects, we shouldn't need CHANGELOG entries

kevmoo:
> ```
> No CHANGELOG change found.
> If this PR needs an exemption from the standard policy of listing all changes in the CHANGELOG,
> comment in the PR to explain why the PR is exempt, and add (or ask your reviewer to add) the
> "override: no changelog needed" label.
> ```
> 
> Since none of this affects code that consumers actually run when using these packages in their projects, we shouldn't need CHANGELOG entries

Sure, but there are robots involved here. Read the docs to see what to do in this case. Or ask in Discord...

Rexios80:
Is that text from the CI failure not the documentation? I guess I forgot to explicitly ask someone to add the override tag. Is there something else I'm missing?

Rexios80:
Also there might be a bug in `flutter_bootstrap.js` that's causing integration tests to not report their status and hang forever but I'm not exactly sure what's going on. We're having the same issue in this PR: https://github.com/flutter/packages/issues/7012

stuartmorgan:

CHANGELOG override: mechanical change to example plumbing to bring it up to current standard.

ditman:
This new bootstrap style + canvaskit are introducing some race conditions that flutter drive is having trouble handling. I'm trying to work on a more permanent solution here, but still haven't figured it out yet:

* https://github.com/flutter/packages/pull/7115

packages/camera/camera/example/web/index.html
```diff
@@ -23,17 +23,8 @@
...
-      window.addEventListener('load', function () {
-        navigator.serviceWorker.register('flutter_service_worker.js');
-      });
-    }
   </script>
-  <script src="main.dart.js" type="application/javascript"></script>
+  <script src="flutter_bootstrap.js" async></script>
 </body>
 
 </html>
```
domesticmouse, line 30: nit: add trailing NL at EOF

domesticmouse (approved):

chunhtai (approved):
LGTM for go_router

ditman:
Linux_web web_platform_tests should improve a lot by rebasing this PR with `main` after [this PR](https://github.com/flutter/packages/pull/7146) lands.

kevmoo:
Guessing that remaining failure is UNRELATED. I'll try rerunning.

ditman (approved):
LGTM!

PR #7143, opened 2024-07-16
[video_player_avfoundation, camera_avfoundation] never overwrite but only upgrade audio session category

misos1:
Setting category of `AVAudioSession` was moved into a wrapper which ensures that only changes which do not disable the ability to play in silent mode (`play`) and ability to record (`record`) are considered. If either the new category or old category is `play`/`record` then also the set category will be `play`/`record`. This currently means that the video player will not overwrite `PlayAndRecord` with `Playback` which causes inability to record audio by camera. Options are treated selectively so the video player will no longer overwrite flags set by camera like `DefaultToSpeaker` and camera will not overwrite `MixWithOthers` set by video player `setMixWithOthers`. It will also only change category or options if there is change to prevent lags every time is constructed `VideoPlayerController` with non-null `videoPlayerOptions` which always causes call to `setMixWithOthers`.

Test `testSeekToWhilePausedStartsDisplayLinkTemporarily` is failing on the main branch on the tested device:
``` objc
XCTAssertEqual([player position], 1234); // (([player position]) equal to (1234)) failed: ("0") is not equal to ("1234")
```

https://github.com/flutter/flutter/issues/131553

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1302,9 +1303,47 @@ - (BOOL)setupWriterForPath:(NSString *)path {
...
+// ability to record audio but never disables it, that could affect other plugins
+// which depend on this global state, only change category or options if there is
+// change to prevent unnecessary route changes which can cause lags
+// https://github.com/flutter/flutter/issues/131553
+static void upgradeAudioSessionCategory(AVAudioSessionCategory category,
+                                        AVAudioSessionCategoryOptions options,
+                                        AVAudioSessionCategoryOptions clearOptions) {
+  if (!NSThread.isMainThread) {
+    dispatch_sync(dispatch_get_main_queue(), ^{
+      upgradeAudioSessionCategory(category, options, clearOptions);
```
hellohuanlin, line 1318: haven't looked into the details. Does this have to be on main queue? Can it be on capture session queue?

misos1: Audio session category is global state, it is changed based on old state so it must be synchronized with other plugins changing it.

misos1: There is needed `dispatch_sync`. Category must be `record` before adding inputs/outputs to capture session otherwise there would be no audio. It is called on the capture session queue so not on the main thread. Seems also `reportErrorMessage` is currently called only on capture session queue but there is that `FLTEnsureToRunOnMainQueue` anyway. If there was similar method which used `dispatch_sync` it would not make any difference it would just be `upgradeAudioSessionCategory->FLTEnsureToRunOnMainQueueSync->dispatch_sync` instead of `upgradeAudioSessionCategory->dispatch_sync->upgradeAudioSessionCategory` so the same depth in stack. Can it be even called recursion if that second call is in another thread on another stack? I think it would be ok to change it to just `dispatch_sync` as when this is called from the main thread it will crash with an unhandled exception instead of deadlock so that would be obvious for someone who did such a change. Btw are there better options on how to synchronize between independent plugins? Ideally it could be called on the original thread but with some mutex shared between these plugins (and something like `NSDistributedLock` seems like overkill).

misos1:
I have no way how to test `AVAudioSessionCategoryOptionOverrideMutedMicrophoneInterruption` so this one was not added although it might prevent some errors for which `camera` may not be prepared: "Attempting to start audio input after the system mutes the microphone results in an error.".

misos1:
I removed `AVAudioSessionCategoryOptionAllowBluetooth` which basically just enables and prefers HFP profile over A2DP. This is not implicitly present with the playback category and has unintended consequences as output is not stereo and lower quality so if there ever is HFP support it should be opt-in from the dart side.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1220,9 +1223,7 @@ - (BOOL)setupWriterForPath:(NSString *)path {
     return NO;
   }
 
-  if (_mediaSettings.enableAudio && !_isAudioSetup) {
-    [self setUpCaptureSessionForAudio];
-  }
+  [self setUpCaptureSessionForAudio];
```
hellohuanlin, line 1226: nit: `setUpCaptureSessionForAudioIfNeeded` since you moved condition inside

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1320,6 +1362,12 @@ - (void)setUpCaptureSessionForAudio {
   // Setup the audio output.
   _audioOutput = [[AVCaptureAudioDataOutput alloc] init];
 
+  upgradeAudioSessionCategory(AVAudioSessionCategoryPlayAndRecord,
+                              AVAudioSessionCategoryOptionDefaultToSpeaker |
+                                  AVAudioSessionCategoryOptionAllowBluetoothA2DP |
+                                  AVAudioSessionCategoryOptionAllowAirPlay,
+                              0);
```
hellohuanlin, line 1369: Do we only pass `0` here? Should we just remove this parameter?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1302,9 +1303,50 @@ - (BOOL)setupWriterForPath:(NSString *)path {
   return YES;
 }
 
+// configure application wide audio session manually to prevent overwriting
+// flag MixWithOthers by capture session, only change category if it is considered
+// as upgrade which means it can only enable ability to play in silent mode or
```
hellohuanlin, line 1308: > enable ability to play in silent mode

is it desired behavior?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1302,9 +1303,50 @@ - (BOOL)setupWriterForPath:(NSString *)path {
...
+// ability to record audio but never disables it, that could affect other plugins
+// which depend on this global state, only change category or options if there is
+// change to prevent unnecessary route changes which can cause lags
```
hellohuanlin, line 1311: > unnecessary route changes

What is a "route change"?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1302,9 +1303,50 @@ - (BOOL)setupWriterForPath:(NSString *)path {
   return YES;
 }
 
+// configure application wide audio session manually to prevent overwriting
+// flag MixWithOthers by capture session, only change category if it is considered
+// as upgrade which means it can only enable ability to play in silent mode or
+// ability to record audio but never disables it, that could affect other plugins
```
hellohuanlin, line 1309: > affect other plugins

can you explicitly mention video player as an example?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: It looked like a good idea during development as there is this same function also in the player which uses that argument and it was easier to work on it in a single place and then copy it into the other plugin. One time this argument was actually used when I tried to use `AVAudioSessionCategoryOptionAllowBluetooth` it looked like a good idea to remove that option when the camera was no longer used.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: System itself is doing this at some point at or after `AVCaptureSession` `addInput` or `addOutput` when `automaticallyConfiguresApplicationAudioSession` is YES which is default, this behaviour was there also before.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: https://developer.apple.com/documentation/avfaudio/responding_to_audio_route_changes?language=objc

When category changes it can cause route change seems:

https://developer.apple.com/documentation/avfaudio/avaudiosessionroutechangereason/avaudiosessionroutechangereasoncategorychange?language=objc

Anyway the effect of setting `AVAudioSessionCategoryPlayAndRecord` multiple times even when the category is unchanged and there are even no headphones causes half of second or longer silence on my device. Also it causes the video_player plugin to pause and must be manually called `play` again.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1331,6 +1368,14 @@ - (void)setUpCaptureSessionForAudio {
   // Setup the audio output.
   _audioOutput = [[AVCaptureAudioDataOutput alloc] init];
 
+  dispatch_sync(dispatch_get_main_queue(), ^{
+    upgradeAudioSessionCategory(AVAudioSessionCategoryPlayAndRecord,
+                                AVAudioSessionCategoryOptionDefaultToSpeaker |
+                                    AVAudioSessionCategoryOptionAllowBluetoothA2DP |
+                                    AVAudioSessionCategoryOptionAllowAirPlay,
```
hellohuanlin, line 1375: can you explain why these options?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1331,6 +1368,14 @@ - (void)setUpCaptureSessionForAudio {
...
+                                0);
```
hellohuanlin, line 1376: I still think it's better not to pass this param. It's making the function a bit complicated to read.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1313,9 +1314,45 @@ - (BOOL)setupWriterForPath:(NSString *)path {
...
+// flag MixWithOthers by capture session, only change category if it is considered
+// as upgrade which means it can only enable ability to play in silent mode or
+// ability to record audio but never disables it, that could affect other plugins
+// which depend on this global state, only change category or options if there is
+// change to prevent unnecessary lags and silence
+// https://github.com/flutter/flutter/issues/131553
+static void upgradeAudioSessionCategory(AVAudioSessionCategory category,
+                                        AVAudioSessionCategoryOptions options,
+                                        AVAudioSessionCategoryOptions clearOptions) {
+  if (category == nil) {
```
hellohuanlin, line 1328: can you explain what `category` means here? is it the category that we want to upgrade to? or the category that we want to upgrade from?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1313,9 +1314,45 @@ - (BOOL)setupWriterForPath:(NSString *)path {
...
+    category = AVAudioSession.sharedInstance.category;
+  }
+  NSSet *playCategories = [NSSet
+      setWithObjects:AVAudioSessionCategoryPlayback, AVAudioSessionCategoryPlayAndRecord, nil];
+  NSSet *recordCategories =
+      [NSSet setWithObjects:AVAudioSessionCategoryRecord, AVAudioSessionCategoryPlayAndRecord, nil];
+  NSSet *categories = [NSSet setWithObjects:category, AVAudioSession.sharedInstance.category, nil];
```
hellohuanlin, line 1335: what does this `categories` mean? why are we combining the set? Can you add some comments to explain?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: DefaultToSpeaker is what it was setting also before with automatic session configuration, so better to not change that behaviour. It is also implicit default for `AVAudioSessionCategoryPlayback` along with AllowBluetoothA2DP and AllowAirPlay which cannot be turned off (it is not set by default for PlayAndRecord). This category uses the video player so it is better to not change player behaviour by just initializing the camera (with audio).

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: Yes, category to which it should upgrade.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: Categories which should be combined. Upgrading means we need to combine categories so the new category has play flavour if one of them had it, and has record when one of them had it. Test whether any of these categories has play is done by intersection which means if anything in the first set is present the second set.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: Or you mean that better would be to have it as an objc selector rather than C function?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
hellohuanlin: can you rename this to `requestedCategory`? Also when will we need to pass a `nil` category param? What does it mean by requesting a nil category? and why are we assigning `category = AVAudioSession.sharedInstance.category`  when it's nil

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
hellohuanlin: Can you rename this to `requiredFinalCategories` or something similar

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
hellohuanlin: Can you add the comments explaining this?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
hellohuanlin: I don't think it's possible to share video player and camera code. Our plugins are not setup to do so. 

I meant it seems we only pass `0` and never other values to this param. So do we really need this param?

hellohuanlin (commented):
Thanks for the comments

misos1:
> Also when will we need to pass a nil category param?

It means it should not be changed, used in player, `AVAudioSession.sharedInstance.category` could not be passed as argument when there was thread switch inside function, now that changed so I can remove that check and make player pass `AVAudioSession.sharedInstance.category` instead of nil.

> I meant it seems we only pass 0 and never other values to this param. So do we really need this param?

No, not in camera at the moment.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCam.h
```diff
@@ -113,7 +113,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)startImageStreamWithMessenger:(NSObject<FlutterBinaryMessenger> *)messenger;
 - (void)stopImageStream;
 - (void)setZoomLevel:(CGFloat)zoom withCompletion:(void (^)(FlutterError *_Nullable))completion;
-- (void)setUpCaptureSessionForAudio;
+- (void)setUpCaptureSessionForAudioIfNeeded;
```
stuartmorgan, line 116: Is this called outside the implementation? If not it should be moved to the private category in the `.m` file.

packages/video_player/video_player_avfoundation/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 2.6.2
 
+* Fixes audio recorded only with first recording.
```
stuartmorgan, line 3: This description doesn't make sense in the context of this package, which does not record anything. Please describe the fix in terms of what is changing generally, not the specific effect it has on a specific test app that readers of the changelog won't have context on.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -705,10 +705,46 @@ - (int64_t)onPlayerSetup:(FVPVideoPlayer *)player frameUpdater:(FVPFrameUpdater
   return textureId;
 }
 
+// this same function is also in camera_avfoundation
```
stuartmorgan, line 708: This is the least important information in the comment, so should be last as a final note.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -705,10 +705,46 @@ - (int64_t)onPlayerSetup:(FVPVideoPlayer *)player frameUpdater:(FVPFrameUpdater
...
+// do not overwrite PlayAndRecord with Playback which causes inability to record
+// audio, do not overwrite all options, only change category if it is considered
+// as upgrade which means it can only enable ability to play in silent mode or
+// ability to record audio but never disables it, that could affect other plugins
+// which depend on this global state, only change category or options if there is
+// change to prevent unnecessary lags and silence
+// https://github.com/flutter/flutter/issues/131553
```
stuartmorgan, line 715: All the comments in the other file apply here as well.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
stuartmorgan: And it's not actually true, because they are slightly different due to `clearOptions`. So this should just be removed.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -184,6 +184,8 @@ - (instancetype)initWithMediaSettings:(FCPPlatformMediaSettings *)mediaSettings
   _videoFormat = kCVPixelFormatType_32BGRA;
   _inProgressSavePhotoDelegates = [NSMutableDictionary dictionary];
   _fileFormat = FCPPlatformImageFileFormatJpeg;
+  _videoCaptureSession.automaticallyConfiguresApplicationAudioSession = NO;
+  _audioCaptureSession.automaticallyConfiguresApplicationAudioSession = NO;
```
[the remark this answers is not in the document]
misos1: Default is YES:

https://developer.apple.com/documentation/avfoundation/avcapturesession/1620477-automaticallyconfiguresapplicati?language=objc

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCam.h
misos1: It is also called from `prepareForVideoRecordingWithCompletion` in CameraPlugin.m.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
misos1: Before it was the same (it was removed due to a comment about it not being used in camera). It was supposed to be in a single place but I could not find a way to share source files between plugins so I put it twice for the time being until it is possibly moved elsewhere into a single place. I think such a comment makes sense because without this context making certain changes to it in one plugin without changing it also in other plugins can result in it no longer working as intended (like simple change in camera to call it on capture session queue instead of main, that change may seem reasonable without context). Maybe I should change it a little like "..similar function..." and that these two need to be "synchronized" resp. on the same queue.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1331,6 +1365,19 @@ - (void)setUpCaptureSessionForAudio {
   // Setup the audio output.
   _audioOutput = [[AVCaptureAudioDataOutput alloc] init];
 
+  dispatch_block_t block = ^{
+    // Setup options implicit to AVAudioSessionCategoryPlayback to not disturb video_player.
```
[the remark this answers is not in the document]
misos1: This is to explain why there are such options: https://github.com/flutter/packages/pull/7143#discussion_r1733292433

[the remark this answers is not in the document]
misos1: Comment on the function does not cover this.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
stuartmorgan: The comments and implementation strategy seem overly focused on these two plugins to me. There could be 50 plugins in the ecosystem that set global audio state, only two of which we happen to control, so any solution that relies on these two plugins being exactly synchronized doesn't scale.

I would like the solution (and its comments) here to focus on things that are generally true:
- `AVAudioSession` is not explicitly documented as being thread safe, so we should only use it from the main thread. We can mention that other plugin or application code could be using `AVAudioSession.sharedInstance` as well, and that we have to hope that they are doing the same.
- We should upgrade rather than stomp the state because other code may be setting the global state as well, and we don't want to remove options they may have set.

Neither of those statements require exact synchronization with other code, and both generalize to the entire ecosystem. (Although as mentioned earlier, the real ecosystem-friendly solution is to move away from doing this in individual plugins at all, and leaving it for apps to do.)

packages/camera/camera_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.9.17+5
+
+* Fixes changing global audio session category to be collision free.
```
hellohuanlin, line 3: nit: ...collision free *across plugins*

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1313,9 +1314,42 @@ - (BOOL)setupWriterForPath:(NSString *)path {
   return YES;
 }
 
-- (void)setUpCaptureSessionForAudio {
+// This function, although slightly modified, is also in video_player_avfoundation.
+// Both need to do the same thing and run on the same thread.
```
hellohuanlin, line 1318: ...run on the *main* thread?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -1331,6 +1365,19 @@ - (void)setUpCaptureSessionForAudio {
...
+    // Setup options implicit to AVAudioSessionCategoryPlayback to not disturb video_player.
+    upgradeAudioSessionCategory(AVAudioSessionCategoryPlayAndRecord,
+                                AVAudioSessionCategoryOptionDefaultToSpeaker |
+                                    AVAudioSessionCategoryOptionAllowBluetoothA2DP |
+                                    AVAudioSessionCategoryOptionAllowAirPlay);
+  };
+  if (!NSThread.isMainThread) {
+    dispatch_sync(dispatch_get_main_queue(), block);
+  } else {
+    block();
```
hellohuanlin, line 1378: is `setUpCaptureSessionForAudioIfNeeded` already guaranteed to be on background? Can you double check the caller of this function? If it's already on background, we can simply do 
```
NSAssert(!NSThread.isMainThread);
dispatch_sync(dispatch_get_main_queue()) {
  // your logic here
};

File: packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m (hunk 16 above)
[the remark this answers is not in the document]
Comment (hellohuanlin) [reply packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m:1369]: How about `Upgrade options requested by camera without conflicting other plugins like video_player`.

**Review by hellohuanlin — COMMENTED** [review]
Sorry for the delay. It's looking good. Just a few questions.

File: packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m (hunk 13 above)
Comment (misos1) [reply packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m]: It is not enough to just use `AVAudioSession` on the main thread, something like this would not work well:

``` dart
dispatch_sync(main_queue, () => old_options = AVAudioSession.sharedInstance.categoryOptions);
new_options = old_options | AVAudioSessionCategoryOptionMixWithOthers;
dispatch_sync(main_queue, () => AVAudioSession.sharedInstance.categoryOptions = new_options);
```

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: It is called on the main thread in tests if I remember correctly https://github.com/flutter/packages/pull/7143#discussion_r1723565703.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
misos1: > AVAudioSession is not explicitly documented as being thread safe

Btw `category` and `categoryOptions` are atomic properties and I doubt that capture session is changing them on the main thread when `automaticallyConfiguresApplicationAudioSession` is true.

misos1: Just for curiosity, here it is changed by capture session and it is not possible that it happened on the main thread:

``` objc
      dispatch_async(dispatch_get_main_queue(), ^{
        NSLog(@"%@", AVAudioSession.sharedInstance.category);
        dispatch_sync(_captureSessionQueue, ^{
          [_audioCaptureSession addOutput:_audioOutput];
        });
        NSLog(@"%@", AVAudioSession.sharedInstance.category);
      });
```

```
AVAudioSessionCategoryPlayback
AVAudioSessionCategoryPlayAndRecord
```

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
hellohuanlin: is it only called on main thread in tests? Can you give an example of such failure? I'd like to see if it's possible to run on background thread in those tests

misos1: Try to run with this:

``` objc
  //if (!NSThread.isMainThread) {
    dispatch_sync(dispatch_get_main_queue(), block);
  //} else {
    //block();
  //}
```

And you will get this inside xcode:

```
Thread 1: EXC_BREAKPOINT (code=1, subcode=0x18082dc74)
```

When running outside of xcode with debugging there will be probably just crash, you can add `NSLog(@"%@", NSThread.currentThread);` to see that it runs on main thread `<_NSMainThread: 0x2804b0080>{number = 1, name = main}`.

You will probably need to run `flutter pub upgrade` first as without that I am getting this error during `flutter run` (win32 on macos?):

```
Could not build the precompiled application for the device.
Error (Xcode): ../../../../../../../../.pub-cache/hosted/pub.dev/win32-5.2.0/lib/src/guid.dart:32:9: Error: Type 'UnmodifiableUint8ListView' not found.
```

hellohuanlin: > It is called on the main thread in tests if I remember correctly https://github.com/flutter/packages/pull/7143#discussion_r1723565703.

> When running outside of xcode with debugging there will be probably just crash, you can add NSLog(@"%@", NSThread.currentThread); to see that it runs on main thread <_NSMainThread: 0x2804b0080>{number = 1, name = main}.

So  `setUpCaptureSessionForAudioIfNeeded` is called on main thread somewhere (not just in test). Could you find the place it's called on main? The session setup shouldn't happen on main thread.

misos1: No I did not mean that I observed something like that. I wrote that because I think that native tests can be run also outside of xcode.

hellohuanlin: hmmm im confused, maybe im missing something. `setUpCaptureSessionForAudioIfNeeded` should only be called on background thread, so the same applies when testing this function, it should only be tested on background thread as well.

misos1: Seems `FLTCam` tries to act like something independent and almost unaware of what queue it actually should run on and dispatches to the session queue are made in `CameraPlugin`. Tests are testing `FLTCam` not `CameraPlugin`. There are for example many calls to `startVideoRecordingWithCompletion` from tests. Now for example `reportErrorMessage` is called only from functions which are always called on session queue (outside of tests) so it itself is called on only on session queue but it uses `FLTEnsureToRunOnMainQueue` which tests whether we are on main queue or not, this suggests that it is probably intended that tests use `FLTCam` on main thread.

hellohuanlin: > but it uses FLTEnsureToRunOnMainQueue which tests whether we are on main queue or not, this suggests that it is probably intended that tests use FLTCam on main thread.

No, `FLTEnsureToRunOnMainQueue` doesn't test whether we are on main queue or not. Instead, it dispatches to main queue if it's not already on main: https://github.com/flutter/packages/blob/45bcc3210fcc24011bc7f88de1ca61528d9e5cbc/packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/QueueUtils.m

misos1: By "it tests" I meant this: `if (!NSThread.isMainThread) {`

hellohuanlin: Gotcha I misunderstood you. 

Right now we have the tests and production code calling separate paths: 
```
if background {
  // production code runs this
} else {
  // test code runs this
}
```

Is it hard to modify the tests to ensure session is setup on background? Something like: 

```

testFoo() {
  let expectation = ...
  sessionQueue.async {
    test here
    expectation.fulfill()
  }
  waitForExpectation()
}
```

misos1: This work is already done, in `CameraPlugin`, all that would be needed to do is to call functions from this class instead of `FLTCam`. But I do not think that there is any problem. I suppose these tests are supposed to be unit tests for `FLTCam` class and there is mostly not even used real (not mocked) `AVFoundation`.

hellohuanlin: > This work is already done, in CameraPlugin, all that would be needed to do is to call functions from this class instead of FLTCam. 

What work is already done?

misos1: About dispatching on the session queue.

hellohuanlin: Sorry im very confused. Does the test already dispatch to session queue? If it does, why are you still getting the crash? Otherwise, we should do it, so the test can pass.

misos1: It is not just a single test or my test, I guess most of the tests do that, I saw some 7 calls to just `startVideoRecordingWithCompletion`.

hellohuanlin: `CameraPlugin` and `FLTCam` are different classes and each has their own tests. Is the `FLTCam` the one fails? If so, we should dispatch to the background thread in the tests. 

The problem is that right now you have a code branch that is only run by tests but not in production, which defeats the purpose of tests.

misos1: Btw I do not see a single native test calling for example `startVideoRecordingWithStreaming`.

misos1: It also branches differently in most tests than in production.

misos1: So if there is a problem let's do it similarly, I will add the function `FLTEnsureToRunOnMainQueueSync` which will be used here.

hellohuanlin: > It also branches differently in most tests than in production.

I haven't touched this plugin for a while. Can you point me to the code? IIRC both of the branches are called in production.

misos1: > I haven't touched this plugin for a while. Can you point me to the code? IIRC both of the branches are called in production.

But not on all places where it is used like I already mentioned earlier about `reportErrorMessage`.

> The problem is that right now you have a code branch that is only run by tests but not in production, which defeats the purpose of tests.

I suppose you meant that in reverse that the problem is that there is a branch which is called in production but not in tests, tests should cover what is in production not vice versa. And btw this is not entirely true, there is an integration test which also calls that second branch on the session queue so both are actually covered already.

hellohuanlin: > I suppose you meant that in reverse that the problem is that there is a branch which is called in production but not in tests

No. 
In this code: 

```
  if (!NSThread.isMainThread) {
    dispatch_sync(dispatch_get_main_queue(), block);
  } else {
    block(); 
  }
```
The `if` branch is only run in production, and `else` is only run in test right? Or am i mistaken?

misos1: Both are run in tests. In production only one of them.

misos1: Oh sorry I confused `camera` with `video_player`. I am not sure whether there is such a test which tests with a real camera device but even if yes this has some limitations of course, like it probably cannot be tested automatically very well. So yes basically you are right. So what about that sync version of `FLTEnsureToRunOnMainQueue`?

misos1: Or maybe I can just simply modify my own test to use a background queue and then both branches will be covered.

hellohuanlin: > So what about that sync version of FLTEnsureToRunOnMainQueue?

This would be the same problem right? I think the right approach is the change the test, so that it dispatch to the background thread. The non-test code itself should assert that it's on the background thread.

misos1: > This would be the same problem right?

Actually not, a separate function can be separately tested like is done with that one.

> The non-test code itself should assert that it's on the background thread.

This is actually by "definition".

hellohuanlin: > Actually not, a separate function can be separately tested like is done with that one.

The problem here is not just the test coverage, but it's that you have a branch in code that is only called by tests. 

> This is actually by "definition".

What do you mean? To be clear, I was suggesting: 

```
NSAssert(!NSThread.isMainThread);
dispatch_sync(dispatch_get_main_queue()) {
  // your logic here
};
```

misos1: But that same situation is also with `FLTEnsureToRunOnMainQueue`. It could be removed in several places where it has meaning only for tests. And if this is somehow fine because that function is maybe also used on places where it may run on the main thread in production then I have probably bad luck that I need sync dispatch instead of async. I think it is possible to issue `dispatch_async` and then wait until it finishes but that would feel somewhat weird. This looks to me more like an artificial problem. And I do not think that refactoring most of the tests is in scope of this PR which is intended to fix a bug.

> The problem here is not just the test coverage, but it's that you have a branch in code that is only called by tests.

How exactly is this a problem anyway? There are many other "adaptations" only because of tests, like interfaces and factories just to simplify testing, in both `camera` and `video_player`.

hellohuanlin: > But that same situation is also with FLTEnsureToRunOnMainQueue.

Could you explain the problem with FLTEnsureToRunOnMainQueue? Maybe a concrete example? 

> How exactly is this a problem anyway? There are many other "adaptations" only for tests, like interfaces and factories just to simplify testing, in both camera and video_player.

I don't know why you are comparing with interfaces and factories. 

Imagine you have this code: 

```
func setupSession() {
  if is background thread {
    Logic A
  } else {
    Logic B
  }
}
```
Here your production code only calls Logic A, and test code only calls Logic B. Then the test isn't really providing any confidence right? 

What I suggest is: 

```
func setupSession() { 
  Assert(session must be setup on background);
  Logic A;
}
```

Then in the test, you can do: 

```
func testSetupSession() {
  dispatch_to_background {
    setupSession();
  }
}
```
This way, the test code actually covers Logic A, providing us with confidence in production.

misos1: Your assertion will fail because other tests call `setupSession` on the main thread. And would this not cause a deadlock due to `dispatch_sync` waiting for block dispatched on main queue while also `waitForExpectation` waiting on main thread until `dispatch_sync` is done and expectation fulfilled? https://github.com/flutter/packages/pull/7143/#discussion_r1876564910

```
testFoo() {
  let expectation = ...
  sessionQueue.async {
    dispatch_sync(main_queue(), ...)
    expectation.fulfill()
  }
  waitForExpectation()
}
```

What I proposed does not have these problems and tests also cover production logic as I wrote.

hellohuanlin: > And would this not cause a deadlock due to dispatch_sync waiting for block dispatched on main queue while also waitForExpectation is waiting on main thread until dispatch_sync is done and expectation fulfilled?

Oh I overlooked the part that it's `dispatch_sync` to main, which leads to deadlock in test. 

> What I proposed does not have such problems and tests also cover production logic as I wrote.

It would be the same problem, that `dispatch_sync(main)` is inherently not testable - You simply moved the same problem to `FLTEnsureToRunOnMainQueueSync` (i.e. the `dispatch_sync` branch of this helper would not be testable for the same reason). So I'd accept the current code as it is.

hellohuanlin (approved):

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: > It would be the same problem

Oh I did not think of that, but as you wrote that, I realized that means that also the async test should be untestable, maybe it actually waits in some non blocking way. But anyway I did not see that documented anywhere, maybe that async test is not 100% clean then?

hellohuanlin: > I realized that means that also the async test should be untestable

`dispatch_async` is different and it can be tested. The problem with `dispatch_sync` is that it causes deadlock when dispatching to the same queue.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
misos1: Is this the last change request still not fulfilled after the latest commits? Should that comment be further rewritten to not explicitly mention the other plugin but something more general?

stuartmorgan (approved):
LGTM!

Thanks for your patience on this review; I'm still catching up on reviews after being out of office for several weeks

PR #7148, opened 2024-07-17
[rfw] Increase tolerance for material widget tests

victoreronmosele:
This PR introduces a tolerance for rfw material widget tests to fix the golden test error in https://github.com/flutter/flutter/issues/151611.

Part of a fix for https://github.com/flutter/flutter/issues/151659.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/rfw/test/material_widgets_test.dart
+++ b/packages/rfw/test/material_widgets_test.dart
@@ -26,7 +26,9 @@ void main() {
 
   setUpAll(() {
     setUpTolerantComparator(
-        testPath: 'test/material_widget_test.dart', precisionTolerance: 0.01);
+      testPath: 'test/material_widget_test.dart',
+      precisionTolerance: 0.00002,
+    );
   });
 
   testWidgets('Material widgets', (WidgetTester tester) async {
--- a/packages/rfw/test/material_widgets_test.dart
+++ b/packages/rfw/test/material_widgets_test.dart
@@ -24,10 +24,12 @@ void main() {
       ..update(materialName, createMaterialWidgets());
   }
 
-  testWidgets('Material widgets', (WidgetTester tester) async {
+  setUpAll(() {
     setUpTolerantComparator(
         testPath: 'test/material_widget_test.dart', precisionTolerance: 0.01);
+  });
 
+  testWidgets('Material widgets', (WidgetTester tester) async {
     final Runtime runtime = setupRuntime();
     final DynamicContent data = DynamicContent();
     final List<String> eventLog = <String>[];
--- a/packages/rfw/test/material_widgets_test.dart
+++ b/packages/rfw/test/material_widgets_test.dart
@@ -9,6 +9,8 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:rfw/formats.dart' show parseLibraryFile;
 import 'package:rfw/rfw.dart';
 
+import 'tolerant_comparator.dart'
+    if (dart.library.js_interop) 'tolerant_comparator_web.dart';
 import 'utils.dart';
 
 void main() {
--- a/packages/rfw/test/material_widgets_test.dart
+++ b/packages/rfw/test/material_widgets_test.dart
@@ -23,6 +23,9 @@ void main() {
   }
 
   testWidgets('Material widgets', (WidgetTester tester) async {
+    setUpTolerantComparator(
+        testPath: 'test/material_widget_test.dart', precisionTolerance: 0.01);
+
     final Runtime runtime = setupRuntime();
     final DynamicContent data = DynamicContent();
     final List<String> eventLog = <String>[];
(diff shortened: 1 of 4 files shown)


packages/rfw/test/material_widgets_test.dart
```diff
@@ -23,6 +25,9 @@ void main() {
   }
 
   testWidgets('Material widgets', (WidgetTester tester) async {
+    setUpTolerantComparator(
+        testPath: 'test/material_widget_test.dart', precisionTolerance: 0.01);
```
Hixie, line 29: how much is 0.01?

Hixie:
Do we understand what kinds of changes are causing failures here? Is it anti-aliasing? Blurs? Something else?

packages/rfw/test/material_widgets_test.dart
victoreronmosele: `0.01` is `4800px` out of the `480000px` in the `material_test.scaffold.png` file. 

The actual diff from the failing test is `4px` so a more precise value for the tolerance is `0.00000833333`.

victoreronmosele:
> Do we understand what kinds of changes are causing failures here? Is it anti-aliasing? Blurs? Something else?

I can't tell which one it is, but I generated the test failure artifacts in a GitHub Action workflow run [here](https://github.com/victoreronmosele/rfw_golden_tests_workflow/actions/runs/10228220835/artifacts/1772200192). 

---

Also, I pushed an update to the PR to set up the tolerant comparator for all the tests in the file since the failure was occurring in multiple tests.

Hixie:
I see nothing different in those diffs at all. :-/

packages/rfw/test/material_widgets_test.dart
Hixie: Meaning there are four pixels that have a non-identical match to the control?

Hixie: I worry that allowing four pixels to be just flat out wrong means that if we start incorrectly rendering a dot somewhere, we won't notice it. "Tolerance" makes it sound like we'd accept it if the four pixels happened to be slightly off in color (e.g. if 0xFFFEFEFE became 0xFFAFAFA or something), not that we'd accept wildly wrong renderings (e.g. 0xFF00FF00 becoming 0xFFFF0000).

victoreronmosele: True. What would be a better fix for the issue? Updating the control images? 

Also, I implemented the tolerance based on this comment in the original issue: https://github.com/flutter/flutter/issues/151611#issuecomment-2223874000.

victoreronmosele:
Cropping and zooming in on the images show some changes in the diffs.

See for `material_test.overflow_bar_resembles_button_bar_masterImage.png`:

| Isolated Diff | Masked Diff |
|---------------|-------------|
| Isolated Diff | Masked Diff |
| **Master Image** | **Test Image** |
| Master Image | Test Image |

Original `material_test.overflow_bar_resembles_button_bar_masterImage.png`:

material_test overflow_bar_resembles_button_bar_masterImage

packages/rfw/test/material_widgets_test.dart
Hixie: Updating the control images is a short-term fix, it doesn't really address the underlying issue (which is that skia and impeller don't guarantee pixel-identical rendering across versions for things like antialiasing and blurs, but the test is assuming they do).

I don't have a good solution. Setting the tolerance to something like 0.00002 (so if I understood correctly, allowing about 10 pixels to drift from the control) might be ok I guess? In the absence of any better ideas...

victoreronmosele: Yes, 10px away from the control. 

I'll update the tolerance.

victoreronmosele: Updated now in 0fd9943.

stuartmorgan:
The referenced issue is closed; is this obsolete, or is changing the allowed diff still something we want to do?

victoreronmosele:
@stuartmorgan, the relevant issue is https://github.com/flutter/flutter/issues/151659. It is still open. 

https://github.com/flutter/flutter/issues/151659 says to fix and enable the rfw tests. 

This PR is doing the fix. 

If this gets merged, another PR should be raised to enable the actual tests in the [flutter/tests](https://github.com/flutter/tests) repo.

stuartmorgan:
From triage: @Hixie, what is the next step here from a reviewer standpoint?

Hixie (approved):
Sorry, this fell of my radar.

I guess LGTM, but I'm a bit skeptical of this approach. I would prefer to find ways to make Flutter not change rendering across versions (e.g. the way we do for shadows, where in golden tests the shadows become big black rectangles).

stuartmorgan (approved):

PR #7156, opened 2024-07-17
[tool] Add a package-level pre-publish hook

stuartmorgan:
Adds the ability for a package to specify a script that should be run before publishing. To minimize the chance of such a script breaking things only in the post-submit `release` step, if the script is present it will also be run during `publish-check`.

These should be used with caution since they can cause the published artifacts to be different from that is checked in, but in the intended use case of extension builds this risk is far preferable to the risks associated with checking in binaries that were built on local, ad-hoc basis. (Longer term, we may need an alternate solution however, as generating artifacts in CI can have its own supply chain validation issues.)

Also does some minor refactoring to custom test script code to make it follow the same pattern as this new code.

Fixes https://github.com/flutter/flutter/issues/150210

## Pre-launch Checklist

- [x] I signed the [CLA].


bparrishMines (approved):
LGTM

Code looks good to me, but I do have some thoughts on the use case. It looks like this is mainly being done to build binaries/artifacts before publishing. Could it be possible that building could be needed outside of publishing (as in for testing)? Should the script be called something like `generate.dart` instead so that other commands could potentially make use of it?

stuartmorgan:
> Could it be possible that building could be needed outside of publishing (as in for testing)?

In general, I think special script entry points should be clearly named, and then if we have code we want to run from multiple entry points the best option is to factor that out into code that the different entry point files can call into.

PR #7162, opened 2024-07-18
[tool] Adds `--wasm` flag to the `drive-examples` command

Rexios80:
I was unsure if we wanted to add a new platform option for this or add a `--wasm` flag to the command. Adding a `--wasm` flag seems more complicated since we would probably want to add extra handling around proper use of the flag, whereas adding a `web-wasm` platform should require no additional checks.

This is in preparation for https://github.com/flutter/flutter/issues/151664


stuartmorgan:
> I was unsure if we wanted to add a new platform option for this or add a `--wasm` flag to the command.

Currently we have a 1:! correspondence with `flutter` build platforms; I would prefer not to change that.

Rexios80:
@stuartmorgan How about this?

stuartmorgan (approved):
I'm not familiar with the web args to `drive`, so @ditman should check that part, but structurally this all LGTM!

ditman (approved):
`--wasm --web-renderer=canvaskit` is some times useful, but not that much for app developers (@eyebrowsoffire said), so this LGTM! Let's land it so we can look at it in the maps migration PR! :)

Rexios80:
> `--wasm --web-renderer=canvaskit` is some times useful, but not that much for app developers (@eyebrowsoffire said), so this LGTM! Let's land it so we can look at it in the maps migration PR! :)

Do you have any insight on what this combination does? Does it compile the WASM version but run the canvaskit one?

eyebrowsoffire:
> > `--wasm --web-renderer=canvaskit` is some times useful, but not that much for app developers (@eyebrowsoffire said), so this LGTM! Let's land it so we can look at it in the maps migration PR! :)
> 
> 
> 
> Do you have any insight on what this combination does? Does it compile the WASM version but run the canvaskit one?

It compiles the dart code to webassembly but uses the canvaskit renderer. By default, `--wasm` uses the skwasm renderer, which is specially optimized to work with the dart2wasm build. The `--wasm --web-renderer=canvaskit` is useful sometimes for me to see if an issue is due to the compile target (dart2wasm) or due to a bug in the skwasm renderer. But most of the time app developers aren't going to care about that.

PR #6825, opened 2024-05-29
[go_router] Add support for relative routes

ThangVuNguyenViet:
Add supports for relative routes by allowing going to a path relatively, like go('./$path')

This PR doesn't fully resolve any issue, but it's mandatory to further add examples & tests for `TypedRelativeGoRoute` (see [#7174](https://github.com/flutter/packages/pull/6823)), which will resolves [#108177](https://github.com/flutter/flutter/issues/108177)

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version
[...]

PR #7174, opened 2024-07-19
[go_router_builder] Add support for relative routes

ThangVuNguyenViet:
Add supports for relative routes by allowing going to a path relatively, like go('./$path')

This PR is a 2nd part of #6825  to fully resolves https://github.com/flutter/flutter/issues/108177, which allow users to use `TypedRelativeGoRoute` to construct the route relatively. 

Example: 

```dart
import 'package:go_router/go_router.dart';

const TypedRelativeGoRoute<RelativeRoute> relativeRoute =
    TypedRelativeGoRoute<RelativeRoute>(
  path: 'relative-route',
  routes: <TypedRoute<RouteData>>[
    TypedRelativeGoRoute<InnerRelativeRoute>(path: 'inner-relative-route')
  ],
);

@TypedGoRoute<Route1>(
  path: 'route-1',
  routes: <TypedRoute<RouteData>>[relativeRoute],
)
class Route1 extends GoRouteData {
  const Route1();
}

@TypedGoRoute<Route2>(
  path: 'route-2',
  routes: <TypedRoute<RouteData>>[relativeRoute],
)
class Route2 extends GoRouteData {
  const Route2();
}

class RelativeRoute extends GoRouteData {
  const RelativeRoute();
}

class InnerRelativeRoute extends GoRouteData {
  const InnerRelativeRoute();
}
```

Basically the added TypedRelativeGoRoute allows us to construct the route tree above. If we replace it with the existing `TypedGoRoute` it will declare multiple extensions of a same thing and produce build time error

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/go_router/lib/src/route_data.dart
+++ b/packages/go_router/lib/src/route_data.dart
@@ -390,6 +390,28 @@ class TypedGoRoute<T extends GoRouteData> extends TypedRoute<T> {
   final List<TypedRoute<RouteData>> routes;
 }
 
+/// A superclass for each typed go route descendant
+@Target(<TargetKind>{TargetKind.library, TargetKind.classType})
+class TypedRelativeGoRoute<T extends GoRouteData> extends TypedRoute<T> {
+  /// Default const constructor
+  const TypedRelativeGoRoute({
+    required this.path,
+    this.routes = const <TypedRoute<RouteData>>[],
+  });
+
+  /// The relative path that corresponds to this route.
+  ///
+  /// See [GoRoute.path].
+  ///
+  ///
+  final String path;
+
+  /// Child route definitions.
+  ///
+  /// See [RouteBase.routes].
+  final List<TypedRoute<RouteData>> routes;
+}
+
 /// A superclass for each typed shell route descendant
 @Target(<TargetKind>{TargetKind.library, TargetKind.classType})
 class TypedShellRoute<T extends ShellRouteData> extends TypedRoute<T> {
--- a/packages/go_router/CHANGELOG.md
+++ b/packages/go_router/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 14.2.1
+- Allows going to a path relatively by prefixing `./`
+- Adds `TypedRelativeGoRoute`
+
 ## 14.2.0
 
 - Added proper `redirect` handling for `ShellRoute.$route` and `StatefulShellRoute.$route` for proper redirection handling in case of code generation.
--- a/packages/go_router/example/lib/go_relative.dart
+++ b/packages/go_router/example/lib/go_relative.dart
@@ -95,7 +95,7 @@ class DetailsScreen extends StatelessWidget {
             onPressed: () {
               context.goRelative('settings');
             },
-            child: const Text('go to settings'),
+            child: const Text('Go to the Settings screen'),
           ),
         ],
       )),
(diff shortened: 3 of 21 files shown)


ThangVuNguyenViet:
I temporarily added the dependency_overrides until the go_router PR gets merged. However, imho that should be there to avoid this type of PR split anyway

packages/go_router/lib/src/route_data.dart
```diff
@@ -390,6 +390,28 @@ class TypedGoRoute<T extends GoRouteData> extends TypedRoute<T> {
   final List<TypedRoute<RouteData>> routes;
 }
 
+/// A superclass for each typed go route descendant
+@Target(<TargetKind>{TargetKind.library, TargetKind.classType})
+class TypedRelativeGoRoute<T extends GoRouteData> extends TypedRoute<T> {
```
chunhtai, line 395: I can't see the difference between this  and `TypedGoRoute` it seems to me that if user uses go_router_builder they are not using url string to navigate so they don't really need relative routing?

thangmoxielabs: Well they won't be using './$path' directly, but they'll use `route.goRelative()`, which uses that under the hood.

Not sure if I got your question right thou..

chunhtai: I meant they could define a TypedGoRoute directly and do GoRouteData.go, why would they want to define a TypedRelativeRoute ?

thangmoxielabs: They can't define GoRoute in multiple places. Take the example in the description, the relativeRoute was declared in a variable and passed in 2 routes. Had that been normal GoRoute, it would've produced 2 extensions with the same name, of which locations are different, and results in build time error

chunhtai: This seems to be two different feature right?

1. is to be able to do `go('./relative')`
2. is to be able to reuse the same RouteData with same relative path in different sub route.

and 2 needs to build on top of 1.
If so I think we should separate the 2 into a new pr to make review simpler.

chunhtai: oh looks like you already did https://github.com/flutter/packages/pull/6825.

thangmoxielabs: okay... so do I need to adjust anything?

ThangVuNguyenViet: okay... so do I need to change anything else?

chunhtai: I would suggest removing the TypeRelativeRoute from https://github.com/flutter/packages/pull/6825 and do all the TypeRelativeRoute thing in a separate pr

ThangVuNguyenViet: TypedRelativeRoute needs to be landed in order for this PR to work. The reason this PR is working right now is because of the `dependency_overrides`, which I meant to remove after the #6825 is merged. Can you suggest how do I remove the TypedRelativeRoute? Do I create a PR with only `TypedRelativeRoute` declaration, wait for it to get merged and then continue with this PR?

Imo it's feels weird to me that an annotation of a builder package belongs to the main package and not in the builder package, but maybe there is a technical reason for that that I don't know

chunhtai: No I meant the TypedRelativeRoute should be separate from https://github.com/flutter/packages/pull/6825 since they are not related. and then we can review and merge https://github.com/flutter/packages/pull/6825 first before we review TypedRelativeRoute separately.

chunhtai:
I am going to close this pr since inactive, feel free to reopen when you have time

PR #7194, opened 2024-07-22
[video_player_avfoundation] send video load failure even when eventsink was initialized late

misos1:
Event `initialized` is sent from `onListenWithArguments` by calling `setupEventSinkIfReadyToPlay` if event sink was `nil` during `observeValueForKeyPath`. This change also sends failure in a similar way. Adds more details to the error message returned by `VideoPlayerController.initialize()`.

- https://github.com/flutter/flutter/issues/151475
- https://github.com/flutter/flutter/issues/147707
- https://github.com/flutter/flutter/issues/56665 - now error message may look like: `Failed to load video: Operation Stopped: The server is not correctly configured.: The operation couldn’t be completed. (CoreMediaErrorDomain error -12939 - byte range and no content length - error code is 200)`

Tests `testSeekToWhilePausedStartsDisplayLinkTemporarily` and `testSeekToWhilePlayingDoesNotStopDisplayLink` are failing on the main branch on ios. Seems position is correctly set to 1234 in `seekTo` completion handler but right after `[mockDisplayLink setRunning:YES]` is set again to 0 and after some time back to 1234 so those two tests sometimes fail (seems more often when running all tests).

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -406,9 +400,46 @@ - (void)updatePlayingState {
   _displayLink.running = _isPlaying || self.waitingForFrame;
 }
 
+- (void)sendFailedToLoadVideoEvent:(NSError *)error {
+  if (!NSThread.isMainThread) {
```
hellohuanlin, line 404: it looks like `sendFailedToLoadVideoEvent` is only called from background queue. Can this function simply be: 
```
- (void)sendFailedToLoadVideoEventOnMainQueue:(NSError *)error {
  dispatch_async(dispatch_get_main_queue(), ^{
    // all logic here
  }
}
```

misos1: It is also called from `setupEventSinkIfReadyToPlay` and `observeValueForKeyPath`.

hellohuanlin:
I have a hard time understanding this in the PR description: 

> Sends also failure similarly as is sent event initialized from onListenWithArguments through setupEventSinkIfReadyToPlay if event sink was not initialized during observeValueForKeyPath. 

Could you rephrase a bit? That'd be helpful for a review. Thanks!

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -345,13 +345,7 @@ - (void)observeValueForKeyPath:(NSString *)path
...
     switch (item.status) {
       case AVPlayerItemStatusFailed:
-        if (_eventSink != nil) {
-          _eventSink([FlutterError
-              errorWithCode:@"VideoError"
-                    message:[@"Failed to load video: "
-                                stringByAppendingString:[item.error localizedDescription]]
-                    details:nil]);
-        }
+        [self sendFailedToLoadVideoEvent:item.error];
```
hellohuanlin, line 348: nit: `sendFailedToLoadVideoEventIfNeeded` since you moved the check inside

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -406,9 +400,46 @@ - (void)updatePlayingState {
...
+  }
+  __block NSString *message = @"Failed to load video";
+  void (^add)(NSString *) = ^(NSString *detail) {
+    if (detail != nil) {
+      message = [message stringByAppendingFormat:@": %@", detail];
+    }
+  };
+  NSError *underlyingError = error.userInfo[NSUnderlyingErrorKey];
+  // add more details to error message
+  // https://github.com/flutter/flutter/issues/56665
```
hellohuanlin, line 421: i dont think we need this comment

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -406,9 +400,46 @@ - (void)updatePlayingState {
...
+  _eventSink([FlutterError errorWithCode:@"VideoError" message:message details:nil]);
+}
+
 - (void)setupEventSinkIfReadyToPlay {
   if (_eventSink && !_isInitialized) {
     AVPlayerItem *currentItem = self.player.currentItem;
+    // if status is Failed here then this was called from onListenWithArguments which means
+    // _eventSink was nil when was called observeValueForKeyPath with updated Failed status
+    // and error was not sent and never will be so it is needed to send it here to have
+    // future returned by initialize resolved instead of in state of infinite waiting,
```
hellohuanlin, line 435: Can you separate these 4 lines into multiple sentences? I have a hard time understanding this.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -417,13 +448,24 @@ - (void)setupEventSinkIfReadyToPlay {
...
+            [self performSelector:_cmd
+                         onThread:NSThread.mainThread
+                       withObject:self
+                    waitUntilDone:NO];
+            break;
           // Cancelled, or something failed.
-          return;
+          case AVKeyValueStatusFailed:
+            // if something failed then future should result in error
+            [self sendFailedToLoadVideoEvent:error];
```
hellohuanlin, line 464: can you also specify which thread this is on

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
misos1: I would not say "if needed" sounds right here. Meaning of that check is rather like "if can" or "if possible".

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
misos1: On some AVFoundation background queue (this is also mentioned in a comment a few lines above):

https://developer.apple.com/documentation/avfoundation/avasynchronouskeyvalueloading/1387321-loadvaluesasynchronouslyforkeys

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
hellohuanlin: can remove this link.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -406,9 +400,38 @@ - (void)updatePlayingState {
...
   if (_eventSink && !_isInitialized) {
     AVPlayerItem *currentItem = self.player.currentItem;
+    // if status is Failed this was probably called from onListenWithArguments, which means
+    // _eventSink was initialized to non-nil probably only after observeValueForKeyPath already
+    // observed Failed status and it did not send error, therefore we need to send it here
+    // see comment in onListenWithArguments
+    // https://github.com/flutter/flutter/issues/151475
+    // https://github.com/flutter/flutter/issues/147707
+    if (currentItem.status == AVPlayerItemStatusFailed) {
+      [self sendFailedToLoadVideoEvent:currentItem.error];
```
hellohuanlin, line 432: can we move this to the caller, before `setupEventSinkIfReadyToPlay` is called?

hellohuanlin: or can you update the function name to `setupEventSinkIfReadyToPlayOrReportFailure`

misos1: There is also 4th place where this function is called masked by `_cmd` here `self performSelector:_cmd`, I would rather not put that check outside of that function.

misos1: Actually that function does not set up `_eventsink`, I would say it is set up by `onListenWithArguments`. This function just sends an event through that event sink.

misos1: On the other hand if status becomes `failed` on that second call it should be already caught in the key observer where the event sink is not nil and so it would send an error from there.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -417,13 +429,27 @@ - (void)setupEventSinkIfReadyToPlay {
...
+                         onThread:NSThread.mainThread
+                       withObject:self
+                    waitUntilDone:NO];
+            break;
           // Cancelled, or something failed.
-          return;
+          case AVKeyValueStatusFailed: {
+            // if something failed then future should result in error
+            dispatch_async(dispatch_get_main_queue(), ^{
+              [self sendFailedToLoadVideoEvent:error];
```
hellohuanlin, line 446: is this code the key change that fixed the bug?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -587,6 +613,13 @@ - (FlutterError *_Nullable)onListenWithArguments:(id _Nullable)arguments
   // This line ensures the 'initialized' event is sent when the event
   // 'AVPlayerItemStatusReadyToPlay' fires before _eventSink is set (this function
   // onListenWithArguments is called)
+  // and also send error in similar case with 'AVPlayerItemStatusFailed'
```
hellohuanlin, line 616: by "similar case", do you mean AVPlayerItemStatusReadyToPlay fires before _eventSink is set? Can you be more explicit here in the comment? (Sorry I have some trouble understanding this bug fix)

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
misos1: Seems no, I think it is that change in `onListenWithArguments`. But without that change it seems also this change works as I now tested it with "Code sample 1" from https://github.com/flutter/flutter/issues/151475.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
misos1: Yes I was following up with that comment above. But this time with `AVPlayerItemStatusFailed` instead of `AVPlayerItemStatusReadyToPlay`. If anything fires before is event sink set then it could not be sent. Would that comment be better if I added "in similar case as above"?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -587,6 +607,13 @@ - (FlutterError *_Nullable)onListenWithArguments:(id _Nullable)arguments
...
+  // https://github.com/flutter/flutter/issues/151475
+  // https://github.com/flutter/flutter/issues/147707
+  if (self.player.currentItem.status == AVPlayerItemStatusFailed) {
```
hellohuanlin, line 613: So this is the part that actually fix the bug? Can you add a comment explaining how we ended up with Failed state here?

hellohuanlin (commented):
Sorry for the delay.

stuartmorgan:
@hellohuanlin Please add tarrinneal (the ecosystem-level CODEOWNER) as the secondary reviewer once this has passed iOS review; the iOS team review is primary since this is internal to the iOS implementation.

hellohuanlin (approved):
LGTM after the nit on comment

tarrinneal (approved):

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
misos1: Sorry I did not manage to proceed with this. Yes when `AVPlayerItemStatusFailed` fires before `_eventSink` is set. `AVPlayerItem` is "started" to do its things and can asynchronously become `ReadyToPlay` or `Failed` before the dart side starts to listen where `onListenWithArguments` is called. My comment here follows up that comment before which tries to explain this but that previous fix connected with that comment omitted `Failed` state as possible state here which this line fixes (if player item can end up with success here it can also end up with failure, so that previous comment should explain it).

stuartmorgan:
Thanks, I closed one, and left a comment in another that is likely fixed but needs re-testing (unless you've already tested that one? If so we can close it). For the third it's not clear to me whether it's fully resolved.

PR #7198, opened 2024-07-22
[shared_preferences] Documented return values for remove and set methods.

RamonFarizel:
This PR adds documentation about the `remove` and `set` methods.
Those methods are basically following the `get` methods patterns, they throw an exception when the type is not expected.
Also, returns true when the operation succeeded.

List which issues are fixed by this PR. You must list at least one issue.
fixes:  [#146070](https://github.com/flutter/flutter/issues/146070)

## Pre-launch Checklist

- [X] I signed the [CLA].


tarrinneal:
All of the changelogs and pubspecs changes for the sub packages (android, foundation, etc) should be reverted. Only the subpackages with changes need them.

tarrinneal:
I'm of the opinion that these changes to the doc comments aren't necessary, as the methods are self documenting.

You can't call `setString` with anything that isn't a string. Same with the other set methods.

The "returns true" comments are unfortunately not exactly correct, as some platforms don't actually have any response that verifies success. If I were to design this api myself today, I wouldn't have any bool returns for that reason.

@stuartmorgan what do you think?

RamonFarizel:
Thanks for your thoughts, folks.

So if we are moving with updating the docs, what do you think is the best approach? 
1. Should we have this message repeated for all set and remove methods?
2. Only a single message mentioning that this is useful for all those methods?

stuartmorgan:
Unless something is part of the comment directly on a method, it won't show up in IDE tools, and it will be harder to find even when looking at the header manually, so far fewer people would see anything added anywhere but to each method.

RamonFarizel:
Okay, I've implemented the changes.

PR #7207, opened 2024-07-24
[google_maps_flutter] Converts map configuration and platform view creation params to Pigeon

stuartmorgan:
For both Android and iOS, convert the platform view creation parameters to a typed structure using Pigeon, via the ability to set a custom codec for the parameter, instead of a JSON dictionary, as well as converting the map configuration structure used in both those params and `updateMapConfiguration`.

This change removes the last high-level use of non-Pigeon code in the plugin. The remaining uses of JSON are now internal to individual maps objects, which means that it should now be the case that all future changes to this plugin use Pigeon structures rather than JSON. (The individual objects don't change often, if at all, and it would be reasonable if someone does want to add a new parameter to one of those objects to require converting that specific object, as that would not require sweeping changes or knowledge of how to do a more holistic Pigeon conversion.)

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
I put both platforms together this time since the scope is smaller than the previous PRs, and the structure of the change is the same between the two, but you can each just review your platform's half.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -863,52 +869,200 @@ PlatformTile _platformTileFromTile(Tile tile) {
...
+    style: config.style,
+  );
+}
+
+// For supporting the deprecated updateMapOptions API.
+PlatformMapConfiguration _platformMapConfigurationFromOptionsJson(
+    Map<String, Object?> options) {
+  // All of these hard-coded values and structures come from
+  // google_maps_flutter_platform_interface/lib/src/types/utils/map_configuration_serialization.dart
+  // to support this legacy API that relied on cross-package magic strings.
```
stuartmorgan, line 968: (This applies to Android and iOS.)

This method is gross in that it uses magic strings from another package, but the good news is that it is only in the codepath for a platform interface API that we don't use in recent versions of the app-facing package. At some point we could just decide to do a breaking change to the *implementation* packages and remove this, without needing to do a breaking change to the app-facing or platform interface packages. I didn't want to do that in this PR though as it requires some extra steps.

(And while this looks like new gross code, it isn't conceptually: it's just moving from native to Dart, and moving from *every* codepath to a codepath that will in practice almost never be used.)

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -863,52 +869,200 @@ PlatformTile _platformTileFromTile(Tile tile) {
...
+    Object? zoomPrefsJson) {
+  if (zoomPrefsJson == null) {
+    return null;
+  }
+  // See `MinMaxZoomPreference.toJson`.
+  final List<double?> minMaxZoom =
+      (zoomPrefsJson as List<Object?>).cast<double?>();
+  return PlatformZoomRange(min: minMaxZoom[0], max: minMaxZoom[1]);
+}
+
```
stuartmorgan, line 1065: (This applies to Android and iOS.)

The reason there's a bunch of added conversion code without any corresponding removal of JSON versions is that the JSON serialization code is in `google_maps_flutter_platform_interface`. That's ugly technical debt, because having the serialization and deserialization in different packages is a bad idea, and means the entire JSON surface is effective part of the public API surface of the platform interface package, and we therefore can't do *anything* to that code other than add things to dictionaries.

So it looks like it's adding complexity, but is actually reducing technical debt by moving the serialization into the same package as the deserialization where it belongs. (There's actually a TODO about this at the top of the file.)

packages/google_maps_flutter/google_maps_flutter_ios/example/shared/maps_example_dart/lib/example_google_map.dart
```diff
@@ -214,15 +214,13 @@ class ExampleGoogleMap extends StatefulWidget {
     this.onMapCreated,
     this.gestureRecognizers = const <Factory<OneSequenceGestureRecognizer>>{},
     this.compassEnabled = true,
-    this.mapToolbarEnabled = true,
```
stuartmorgan, line 217: These changes to the iOS example are opportunistic cleanup; these have always been Android-only, and just got copypasta'd here when I federated the implementations out of the original shared implementation.

stuartmorgan:
I'm not sure what's going on with that Android test; there's not much info in FTL output, and I haven't been able to repro locally. I'll keep investigating that in parallel with review.

stuartmorgan:
> there's not much info in FTL output

Turns out there was, it was just buried in all the other log output.

It found a real bug where I missed a change I needed to make in Convert.java (which the compiler didn't flag because it's code casts `Object` to other things; the usual problem of dealing with pre-Pigeon code) and I had missed that branch when adding unit tests. Pushed a fix and a unit test that exercised that specific section of code.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
reidbaker: Thank you for adding a link to where the magic strings come from

reidbaker (approved):
Android LGTM

PR #7067, opened 2024-07-08
[camera_windows] Support image streams on Windows platform

jlundOverlay:
Adds support to the windows camera implementation to allow streaming of frames.

Fixes https://github.com/flutter/flutter/issues/97542

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

PR #7220, opened 2024-07-25
[camera] Support image streams on Windows platform

jlundOverlay:
Adds support to the to the camera plugin for streaming frames on windows as implemented in https://github.com/flutter/packages/pull/7067

Fixes https://github.com/flutter/flutter/issues/97542

## Pre-launch Checklist

- [x] I signed the [CLA].


jlundOverlay:
Are we better off just removing the assertion all together, seems redundant to limit what non federated packages can do?

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -486,7 +486,8 @@ class CameraController extends ValueNotifier<CameraValue> {
   // TODO(bmparr): Add settings for resolution and fps.
   Future<void> startImageStream(onLatestImageAvailable onAvailable) async {
     assert(defaultTargetPlatform == TargetPlatform.android ||
-        defaultTargetPlatform == TargetPlatform.iOS);
+        defaultTargetPlatform == TargetPlatform.iOS ||
+        defaultTargetPlatform == TargetPlatform.windows);
```
loic-sharma, line 490: Could you also update the comment above, on line 483?

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -522,7 +523,8 @@ class CameraController extends ValueNotifier<CameraValue> {
   /// platforms won't be supported in current setup).
   Future<void> stopImageStream() async {
     assert(defaultTargetPlatform == TargetPlatform.android ||
-        defaultTargetPlatform == TargetPlatform.iOS);
+        defaultTargetPlatform == TargetPlatform.iOS ||
+        defaultTargetPlatform == TargetPlatform.windows);
```
loic-sharma, line 527: Could you also update the comment above on line 522?

loic-sharma (approved):
LGTM.

For second reviewer: don't merge until https://github.com/flutter/packages/pull/7067 lands and `camera_windows 0.2.5` is published.

loic-sharma:
> Are we better off just removing the assertion all together, seems redundant to limit what non federated packages can do?

Excellent question. Personally, I think it's good to keep these checks. This tells app developers if they use an API that's unsupported on a specific platform, which is a common mistake. If a federated platform package wants to add a feature, they can ask us to remove a particular check to allowlist them.

bparrishMines (approved):
LGTM (it also looks like the windows implementation has landed)

> Are we better off just removing the assertion all together, seems redundant to limit what non federated packages can do?

Typically you would be correct, but the camera plugin still uses a [default implementation](https://github.com/flutter/packages/blob/main/packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart#L28) for `CameraPlatform`. For newer plugins, we don't add this check since the platform interface would throw an error  by default if the method wasn't supported.

cbracken:
Are we good to land this now?

stuartmorgan:
@jlundOverlay Are you still planning on updating this PR?

jlundOverlay:
Have fixed the merge conflicts, should be good to go.

stuartmorgan (changes requested):
Looking at this PR in more detail: it does not appear that the bug I flagged [here](https://github.com/flutter/packages/pull/7067#discussion_r1693929891) was ever fixed. We should not enable this feature when it will fail to send responses and violate the contract of the engine APIs it uses.

stuartmorgan:
> > Are we better off just removing the assertion all together, seems redundant to limit what non federated packages can do?
> 
> Excellent question. Personally, I think it's good to keep these checks. This tells app developers if they use an API that's unsupported on a specific platform, which is a common mistake. If a federated platform package wants to add a feature, they can ask us to remove a particular check to allowlist them.

We should fix this via [this pattern](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#api-support-queries) which is our preferred approach; the hard-coded platform checks are a legacy of pre-federation approaches. Rather than update the hard-coding, the much better fix here would be to add a new platform interface method to query for streaming support, update the supported platforms to return `true` for it, and then update the app-facing package to require those minimum versions of the implementation packages (to avoid regressions) and gate on that check instead of platform.

stuartmorgan:
@jlundOverlay Are you still planning on updating this to address the issue with missing responses?

If not, we should revert the underlying PR, as that code is not correct, and would cause serious issues if ever exposed.

yoongyo:
Please keep making progress.

gamelaster:
@yoongyo it seems as the original author of patch was doing that as part of their job, and maybe they don't work anymore. I think it would be wise to look for someone, who can continue working on this.

Of course, writing this is easy, but I already tried to put this into my job's todo-list, although, in the end, we will not use this feature :frowning_face: .

stuartmorgan:
Closing in light of https://github.com/flutter/packages/pull/7951 and the lack of response here. If you decided to revisit this feature in the future, please feel free to open a new PR!

PR #7233, opened 2024-07-29
[go_router] Support for redirection by history operations in the browser

koji-1009:
In the redirection process by the browser's backward/forward operation, it is necessary to give the browser an instruction to “execute the redirection”.

Fixes https://github.com/flutter/flutter/issues/149181

https://github.com/user-attachments/assets/6b62e36b-eb61-4d2b-90cf-e8b5331b1cd4

## Pre-launch Checklist

- [x] I signed the [CLA].


chunhtai (commented):
What is the expected behavior is? if you use push and go through redirect, you will push another Home page on top of the current home page, and I think if that is the expected behavior, the fix should be in the `GoRouteInformationParser.parseRouteInformationWithDependencies` to make sure go all the imperative route go through redirect. However, I think you should probably use `go` instead of push to go to the auth page.

If you would like to do some other complex thing such as pushing auth route automatically only when user not log in. You will probably need http://flutter.dev/go/go-router-redirect

koji-1009:
What I wanted to fix was the difference in behavior between `context.push` and browser's forward button. Both route level redirects fire, but when browser forward is used, it transitions to the page I want to guard against.

https://github.com/user-attachments/assets/ddb09408-af28-4b83-9289-ed90ab6cbafa

Is this expected behavior?
If it is expected, I think we can close the PR here, along with https://github.com/flutter/flutter/issues/149181. Thanks.

chunhtai:
I think the current behavior is wrong. but the fix should probably be in GoRouteInformationParser.parseRouteInformationWithDependencies to make sure go all the imperative route go through redirect

koji-1009:
Thanks for the reply.
Reflecting your advice, I think the code would look something like this.

```dart
if (state is! RouteInformationState) {
  // This is a result of browser backward/forward button or state
  // restoration. In this case, the route match list is already stored in
  // the state.
  final RouteMatchList baseMatchList =
      _routeMatchListCodec.decode(state as Map<Object?, Object?>);
  final Uri? newUri = baseMatchList.matches
      .whereType<ImperativeRouteMatch>()
      .lastOrNull
      ?.matches
      .uri;
  final RouteMatchList matchList;
  if (newUri != null) {
    matchList = RouteMatchList(
      matches: baseMatchList.matches,
      uri: newUri,
      pathParameters: const <String, String>{},
    );
  } else {
    matchList = baseMatchList;
  }

  return debugParserFuture = _redirect(context, matchList)
      .then<RouteMatchList>((RouteMatchList value) {
    if (value.isError && onParserException != null) {
      // TODO(chunhtai): Figure out what to return if context is invalid.
      // ignore: use_build_context_synchronously
      return onParserException!(context, value);
    }
    return value;
  });
}
```

At this time the test case is updated.

```dart
testWidgets(
    'GoRouteInformationParser can restore full route matches if optionURLReflectsImperativeAPIs is true',
    (WidgetTester tester) async {
  final GlobalKey<NavigatorState> navKey = GlobalKey<NavigatorState>();
  final List<GoRoute> routes = <GoRoute>[
    GoRoute(
      path: '/',
      builder: (_, __) => const Placeholder(),
      routes: <GoRoute>[
        GoRoute(
          path: 'abc',
          builder: (_, __) => const Placeholder(),
        ),
      ],
    ),
  ];
  GoRouter.optionURLReflectsImperativeAPIs = true;
  final GoRouter router =
      await createRouter(routes, tester, navigatorKey: navKey);

  // Generate RouteMatchList with imperative route match
  router.go('/abc');
  await tester.pumpAndSettle();
  router.push('/');
  await tester.pumpAndSettle();
  final RouteMatchList matchList = router.routerDelegate.currentConfiguration;
  expect(matchList.uri.toString(), '/abc');
  expect(matchList.matches.length, 3);

  final RouteInformation restoredRouteInformation =
      router.routeInformationParser.restoreRouteInformation(matchList)!;
  expect(restoredRouteInformation.uri.path, '/');

  // Can restore back to original RouteMatchList.
  final RouteMatchList parsedRouteMatch = await router.routeInformationParser
      .parseRouteInformationWithDependencies(
          restoredRouteInformation, navKey.currentContext!);
  expect(parsedRouteMatch.uri.toString(), '/'); // <- here, change '/abc` to `/`
  expect(parsedRouteMatch.matches.length, 3);

  GoRouter.optionURLReflectsImperativeAPIs = false;
});
```

I thought it would have less impact if the redirect loop was only allowed once, so I created this PR. 
However, I agree that getting the `parseRouteInformationWithDependencies` behavior right would be good. If I am not mistaken, I would like to modify it as advised.

packages/go_router/lib/src/parser.dart
```diff
@@ -67,8 +67,26 @@ class GoRouteInformationParser extends RouteInformationParser<RouteMatchList> {
       // This is a result of browser backward/forward button or state
       // restoration. In this case, the route match list is already stored in
       // the state.
-      final RouteMatchList matchList =
+      final RouteMatchList baseMatchList =
           _routeMatchListCodec.decode(state as Map<Object?, Object?>);
+      final Uri? newUri = baseMatchList.matches
```
chunhtai, line 72: If there is a imperative route match, should we only redirect that imperative route match instead of redirect the entire baseMatchList?

packages/go_router/lib/src/parser.dart
```diff
@@ -67,8 +67,26 @@ class GoRouteInformationParser extends RouteInformationParser<RouteMatchList> {
...
+          .whereType<ImperativeRouteMatch>()
+          .lastOrNull
+          ?.matches
+          .uri;
+      final RouteMatchList matchList;
+      if (newUri != null) {
+        // If the last route is an imperative route and has a new uri,
+        // set the new uri to redirect.
+        matchList = RouteMatchList(
```
chunhtai, line 81: we should probably use `configuration.findMatch`

koji-1009:
Sorry, I reread from the flutter/engine implementation and changed the fix.

---

When the browser's backward/forward button is pressed, flutter monitors PopStateEvent and detects the URL change. At this time, it is not possible to determine whether the change is backward or forward.

https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event
https://github.com/flutter/engine/blob/3.24.0/lib/web_ui/lib/ui_web/src/ui_web/navigation/platform_location.dart#L97

For this reason, it was not reasonable to separate the process by “imperative route match” as you pointed out.

---

On the other hand, if no redirection is performed during the popstate callback, the URL will be changed by the browser. For this reason, redirection processing triggered by the backward and forward buttons must be performed even if the URLs before and after the transition are the same. (During the popstate callback, the URL is before the change.)

Changed to set a flag when popstate events by backward and forward are detected. Since the redirection process is a recursive process, the flag is set to false when a recursive call is made, so that the process does not loop.

chunhtai:
some update, sorry for the late reply, I will take a look this week

packages/go_router/lib/src/configuration.dart
```diff
@@ -347,14 +347,17 @@ class RouteConfiguration {
...
+    required List<RouteMatchList> redirectHistory,
+    bool forceRedirect = false,
+  }) {
     FutureOr<RouteMatchList> processRedirect(RouteMatchList prevMatchList) {
       final String prevLocation = prevMatchList.uri.toString();
       FutureOr<RouteMatchList> processTopLevelRedirect(
           String? topRedirectLocation) {
         if (topRedirectLocation != null &&
-            topRedirectLocation != prevLocation) {
+            (forceRedirect || topRedirectLocation != prevLocation)) {
```
chunhtai, line 360: I am not entirely sure why we need this flag. if the location is the same wouldn't the newMatch be the same after `_getNewMatches`

chunhtai (commented):
I am slightly confused at what the pr is trying to achieve. It looks like we force browser backward and forward to go through a redirect in go_router, which I am not quite understand why.

koji-1009:
@chunhtai 
I commented at https://github.com/flutter/packages/pull/7233#issuecomment-2295276032 why I made the change.
Is this reason insufficient or unsatisfactory?

> If no redirect is performed during the popstate callback, the URL will be changed by the browser.
> In the implementation before the change, the redirect process is skipped if the paths before and after the redirect are identical. Therefore, the redirect process triggered by the back and forward buttons must be executed even if the URLs before and after the transition are the same. (During the popstate callback, the URL is before the change).

chunhtai:
> If no redirect is performed during the popstate callback, the URL will be changed by the browser.
I don't think a redirect is needed for this case, unless there is a bug. This is the part where go_router updates browser url
https://github.com/flutter/packages/blob/d83d021f70d183ce713d1a3a11da1bbfdd3f1448/packages/go_router/lib/src/information_provider.dart#L120

I imagine for the case the redirect don't change url, this method should call be called with
```
    SystemNavigator.routeInformationUpdated(
      uri: routeInformation.uri,
      state: routeInformation.state,
      replace: replace, // This needs to be true
    );
```

koji-1009:
The problem occurs in the case “When I do a backward/forward from one path (e.g. `/`) to another (e.g. `/user` ), I want to redirect from the other path (`/user`) to the first path (`/`).”.
As shown in the video, there is a problem with the current version of this process. https://github.com/flutter/packages/pull/7233#issuecomment-2257226855

---

I am a little confused, is the async redirect method resolved when `RouteReportsNewRouteInformation` is called?  Until the redirect method is resolved, it is not possible to determine if the redirect URL is null, same as current, or different from current.

chunhtai:
RouteReportsNewRouteInformation is called whenever Router rebuilds. Unless it doesn't rebuild when the backward and forward button is pressed, this can happen if RouterDelegate think the RouteMatchList is already up to date.

In this case, I think we should probably do a one off call to RouteReportsNewRouteInformation

PR #7267, opened 2024-08-01
[all] Cleanup unformatted files

Rexios80:
The format command always changes these even though I haven't touched them

https://github.com/flutter/flutter/issues/152651

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/QueueUtils.h
+++ b/packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/QueueUtils.h
@@ -7,7 +7,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /// Queue-specific context data to be associated with the capture session queue.
-extern const char* FLTCaptureSessionQueueSpecific;
+extern const char *FLTCaptureSessionQueueSpecific;
 
 /// Ensures the given block to be run on the main queue.
 /// If caller site is already on the main queue, the block will be run
--- a/packages/camera/camera_windows/windows/capture_controller.cpp
+++ b/packages/camera/camera_windows/windows/capture_controller.cpp
@@ -39,7 +39,7 @@ CaptureControllerImpl::CaptureControllerImpl(
     : capture_controller_listener_(listener),
       media_settings_(
           PlatformMediaSettings(PlatformResolutionPreset::kMax, true)),
-      CaptureController(){};
+      CaptureController() {};
 
 CaptureControllerImpl::~CaptureControllerImpl() {
   ResetCaptureController();
--- a/packages/camera/camera_windows/windows/test/mocks.h
+++ b/packages/camera/camera_windows/windows/test/mocks.h
@@ -130,7 +130,7 @@ class MockCameraFactory : public CameraFactory {
 class MockCamera : public Camera {
  public:
   MockCamera(const std::string& device_id)
-      : device_id_(device_id), Camera(device_id){};
+      : device_id_(device_id), Camera(device_id) {};
 
   ~MockCamera() = default;
 
@@ -217,7 +217,7 @@ class MockCamera : public Camera {
 
 class MockCaptureControllerFactory : public CaptureControllerFactory {
  public:
-  MockCaptureControllerFactory(){};
+  MockCaptureControllerFactory() {};
   virtual ~MockCaptureControllerFactory() = default;
 
   // Disallow copy and move.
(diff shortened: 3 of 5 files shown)


packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/QueueUtils.h
```diff
@@ -7,7 +7,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /// Queue-specific context data to be associated with the capture session queue.
-extern const char* FLTCaptureSessionQueueSpecific;
+extern const char *FLTCaptureSessionQueueSpecific;
```
stuartmorgan, line 10: This probably depends on the exact version of clang-format, due to changes in the logic for detecting whether a file is C++ or Obj-C.

packages/camera/camera_windows/windows/capture_controller.cpp
```diff
@@ -39,7 +39,7 @@ CaptureControllerImpl::CaptureControllerImpl(
     : capture_controller_listener_(listener),
       media_settings_(
           PlatformMediaSettings(PlatformResolutionPreset::kMax, true)),
-      CaptureController(){};
+      CaptureController() {};
```
stuartmorgan, line 42: All of the rest of these are minor differences in how clang-format handles this code, but the easy solution to these is to fix them all to not have the stray `;` that shouldn't be there and is causing it to have non-trivial format analysis. Every one of these should have stable formatting across versions once the incorrect `;` is gone.

stuartmorgan (changes requested):

stuartmorgan:
(FYI, the CI runs the same format command that you are running locally, and fails on diffs, so there are never files checked in that `format` touches but that aren't formatted.)

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/QueueUtils.h
Rexios80: I think I'm using the latest clang-format version. What version is CI using? Or has this file just not been touched in a really long time?

packages/camera/camera_windows/windows/capture_controller.cpp
Rexios80: I removed the `;` and the formatter still adds the space. Is that the expected behavior?

Rexios80:
> (FYI, the CI runs the same format command that you are running locally, and fails on diffs, so there are never files checked in that format touches but that aren't formatted.)

I figured that was the case. Is CI using an old clang-format version? Or does it only check files that changed in the PR.

stuartmorgan:
> Is CI using an old clang-format version?

It's using a specific git revision specified in the CI config; I don't know the exact clang version it corresponds to.

> Or does it only check files that changed in the PR.

Like almost all the repo commands, in CI it runs on any touched packages, or on all packages for repo-level changes (such as the daily Flutter roll).

packages/camera/camera_windows/windows/capture_controller.cpp
stuartmorgan: Yes, the space is correct.

Rexios80:
> It's using a specific git revision specified in the CI config; I don't know the exact clang version it corresponds to.

What is the reason for this? Should we update it?

Should we update the documentation on how to setup the correct version of clang-format? I've accidentally checked in these formatting changes a few times and goofed up PRs.

stuartmorgan:
> > It's using a specific git revision specified in the CI config; I don't know the exact clang version it corresponds to.
> 
> What is the reason for this?

The reason for using specific builds in CI is to ensure repeatable behavior in CI.

> Should we update it?

If there's a newer CIPD package already available, I'd be happy to review a PR to update it. It was current as of about a year ago, and there hasn't been a need to update it.

> Should we update the documentation on how to setup the correct version of clang-format? I've accidentally checked in these formatting changes a few times and goofed up PRs.

It's not-trivial to set up an identical version locally, and this doesn't come up all that often. There's an issue tracking doing this IIRC, but it hasn't been a priority. I'm happy to review a PR for that as well.

Rexios80:
We should definitely do _something_ about this. It's not a great developer experience (especially for new contributors) to have untouched files change by running the format command.

Rexios80:
@stuartmorgan See https://github.com/flutter/packages/pull/7271

stuartmorgan:
> We should definitely do _something_ about this. It's not a great developer experience (especially for new contributors) to have untouched files change by running the format command.

If clang is updated to a newer version that what all the contributors who don't currently have that problem are using, then they will start having that problem. There's no solution that will work for everyone unless the repo tooling manages a `clang-format` version itself (which is a lot of extra complexity).

PR #7268, opened 2024-08-01
[shared_preferences] fix stringlist bug

tarrinneal:
I tested this a few different ways and there was never any savings by not casting when the list is already `List<String>`.

fixes https://github.com/flutter/flutter/issues/152607

### DIFF
--- a/packages/shared_preferences/shared_preferences/CHANGELOG.md
+++ b/packages/shared_preferences/shared_preferences/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.3.1
+
+* Fixes `getStringList` bug with List<Object> cast exception.
+
 ## 2.3.0
 
 * Adds `SharedPreferencesAsync` and `SharedPreferencesWithCache` APIs.
--- a/packages/shared_preferences/shared_preferences/CHANGELOG.md
+++ b/packages/shared_preferences/shared_preferences/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 2.3.1
 
-* Fixes `getStringList` bug with `List<Object>` cast exception.
+* Fixes `getStringList` bug with `List<Object?>` cast exception.
 
 ## 2.3.0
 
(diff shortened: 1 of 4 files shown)


packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_test.dart
```diff
@@ -587,6 +587,17 @@ void main() {
         expect(preferences.getStringList(listKey), testList);
       });
 
+      testWidgets('get StringList handles List<Object>',
+          (WidgetTester _) async {
+        final (
+          SharedPreferencesWithCache preferences,
+          Map<String, Object?> cache
+        ) = await getPreferences();
+        final List<Object> listObject = <Object>['one', 'two'];
```
stuartmorgan, line 596: The actual type in the exception was `List<Object?>`, so this should be using that as the type even though none of the elements are null.

packages/shared_preferences/shared_preferences/lib/src/shared_preferences_async.dart
```diff
@@ -307,9 +307,8 @@ class SharedPreferencesWithCache {
       throw ArgumentError(
           '$key is not included in the PreferencesFilter allowlist');
     }
-    final List<String>? list = _cache[key] as List<String>?;
     // Make a copy of the list so that later mutations won't propagate
-    return list?.toList();
+    return (_cache[key] as List<Object>?)?.cast<String>().toList();
```
stuartmorgan, line 311: Same for the `as` cast here. I believe this will actually still fail in the real case.

packages/shared_preferences/shared_preferences/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.3.1
+
+* Fixes `getStringList` bug with List<Object> cast exception.
```
stuartmorgan, line 3: Nit: backticks around the type as well.

stuartmorgan (changes requested):

stuartmorgan (approved):
LGTM

PR #7294, opened 2024-08-02
Roll to 9d5ede0f5, disabling impeller for google_maps_flutter_android

stuartmorgan:
Attempts to unblock the flutter/flutter->flutter/packages roll, which is many days behind now, by turning off impeller in `google_maps_flutter_android` temporarily until https://github.com/flutter/flutter/issues/152691 is fixed, so that we can roll past the commit that enabled it by default.

The alternative would be turn off device testing of `google_maps_flutter*`, but we've never been able to turn that test on for emulators so that would remove all `google_maps_flutter*` on Android, which is worse from the standpoint of things we need to test in this repo. There are a number of non-trivial changes happening to that package, so we need integration test coverage.


script/tool/test/drive_examples_command_test.dart
```diff
@@ -1023,6 +1023,7 @@ void main() {
                   'test',
                   '-d',
                   _fakeAndroidDevice,
+                  '--no-enable-impeller',
```
reidbaker, line 1026: Can you add a comment to remove this when https://github.com/flutter/flutter/issues/152691 is resolved?

reidbaker (approved):
I would much prefer the engine to turn off impeller by default.

script/tool/test/drive_examples_command_test.dart
stuartmorgan: The comment is on the code; the test will fail once that is fixed. I'm also planning to file a separate issue just tracking the removal of the line to make very sure we don't forget.

stuartmorgan:
> I would much prefer the engine to turn off impeller by default.

I really want to avoid going into the weekend with the roller still stuck and getting further and further behind, which is why I decided to do this now; I'm happy to revert this next week if we have a fix for the crash or if the engine disables impeller, but in the intervening time I can coax some rolls through the macOS flake.

stuartmorgan:
🤦🏻 This is the wrong command; FTL doesn't use command. I need to move this to the FTL side.

script/tool/test/drive_examples_command_test.dart
reidbaker: Facepalm

reidbaker (approved):

tarrinneal (approved):

stuartmorgan:
test-exempt: rolling dependencies and configuration change

PR #7326, opened 2024-08-06
[google_maps_flutter_android] Convert Circle and Marker to Pigeon

yaakovschectman:
Utilize static typing for pigeon message objects for Marker and Circle in `google_maps_flutter_android`.

Draft for now to run tests.

[*List which issues are fixed by this PR. You must list at least one issue.*](https://github.com/flutter/flutter/issues/152925)

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/marker.dart
```diff
@@ -64,7 +64,8 @@ class InfoWindow {
     );
   }
 
-  Object _toJson() {
+  /// Converts this object to something serializable in JSON.
+  Object toJson() {
```
stuartmorgan, line 68: This object only has three fields and they are all simple; it would be much easier to include it in this Pigeon conversion than to make this a multi-package PR. (Also, exposing more of the JSON surface publicly is not desirable, as this is all legacy code that we are going to want to remove someday.)

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.13.1
+
+* Convert `PlatformCircle` and `PlatformMarker` to pigeon.
```
stuartmorgan, line 3: Converts

See https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -708,7 +722,21 @@ class GoogleMapsFlutterAndroid extends GoogleMapsFlutterPlatform {
     // This cast is not ideal, but the Java code already assumes this format.
     // See the TODOs at the top of this file and on the 'json' field in
     // messages.dart.
```
stuartmorgan, line 724: This TODO is resolved by the conversion.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -46,14 +46,27 @@ class PlatformCameraUpdate {
...
+  PlatformCircle({
+    required this.circleId,
+    required this.center,
+    this.consumeTapEvents = false,
+    this.fillColor = 0x00000000,
+    this.strokeColor = 0xFF000000,
+    this.visible = true,
+    this.strokeWidth = 10,
+    this.zIndex = 0.0,
+    this.radius = 0,
```
stuartmorgan, line 58: Optional: you can generally just make everything `required` in a Pigeon definition for a class like this that corresponds pretty directly to an existing class, because in practice there are usual no calls that don't provide every field due to the nature of how we use them.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -75,16 +88,45 @@ class PlatformClusterManager {
...
+    this.consumeTapEvents = false,
+    this.draggable = false,
+    this.flat = false,
+    this.icon = const <Object>['defaultMarker'],
+    this.infoWindow = const <String?, Object>{'anchor': <Object>[0.5, 0.0]},
+    required this.position,
+    this.rotation = 0.0,
+    this.visible = true,
+    this.zIndex = 0.0,
+    this.clusterManagerId,
```
stuartmorgan, line 114: Same note here.

stuartmorgan (changes requested):
A few initial notes.

I'll update and land https://github.com/flutter/packages/pull/7207 tomorrow now that the other PRs it was going to collide with have landed; that will make the diffs here (and in follow-up object conversion PRs) much easier to review because you'll be able to actually replace the JSON codepaths instead of duplicating them, so it will be very easy to see what's the same and what's changing.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
yaakovschectman: At least one case where they were not all provided is the Java unit tests prior to this PR. Technically there was only one field at that time, the `json`, but that json field only had the fields specified that were actually tested by the unit tests. If any of the non-null fields are not provided to the `Builder` nested class, `build` will throw an exception. I was hoping that providing default values would mean that the Builder would use them and not require them all to be specified, avoiding this problem, but it appears that the generated Java code does not use the defaults anyway.

stuartmorgan: Yes, default values haven't been implemented for any of the just languages. Usually for unit tests we just make a helper that builds the object with mostly default values.

stuartmorgan:
Okay, https://github.com/flutter/packages/pull/7207 has landed, so you can merge in `main` here, and then delete the JSON codepaths for anything you've converted. And for future PRs you'll be able to alter then in place instead of duplicating first.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -659,68 +661,35 @@ static void interpretMapConfiguration(
...
+    sink.setZIndex(marker.getZIndex().floatValue());
   }
 
   private static void interpretInfoWindowOptions(
-      MarkerOptionsSink sink, Map<String, Object> infoWindow) {
-    String title = (String) infoWindow.get("title");
-    String snippet = (String) infoWindow.get("snippet");
+      MarkerOptionsSink sink, Messages.PlatformInfoWindow infoWindow) {
+    String title = infoWindow.getTitle();
+    String snippet = infoWindow.getSnippet();
```
stuartmorgan, line 685: Nit: now that this is trivial and self-documenting, it could just go inline in the `setInfoWindowText` call

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkersController.java
```diff
@@ -169,15 +169,12 @@ public void onClusterItemRendered(MarkerBuilder markerBuilder, Marker marker) {
     }
   }
 
-  private void addJsonMarker(Map<String, ?> marker) {
+  private void addMarker(Messages.PlatformMarker marker) {
     if (marker == null) {
```
stuartmorgan, line 173: Can this be removed, and the param marked as `@NonNull`?

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ClusterManagersControllerTest.java
```diff
@@ -161,12 +161,26 @@ public void RemoveClusterManagers() {
         () -> controller.getClustersWithClusterManagerId(clusterManagerId));
   }
 
-  private Map<String, Object> createMarkerData(
+  private Messages.PlatformMarker createMarkerData(
       String markerId, List<Double> location, String clusterManagerId) {
     Map<String, Object> markerData = new HashMap<>();
```
stuartmorgan, line 166: This appears to all be dead code now.

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ClusterManagersControllerTest.java
```diff
@@ -161,12 +161,26 @@ public void RemoveClusterManagers() {
         () -> controller.getClustersWithClusterManagerId(clusterManagerId));
   }
 
-  private Map<String, Object> createMarkerData(
+  private Messages.PlatformMarker createMarkerData(
```
stuartmorgan, line 164: `createPigeonMarker` or `createPlatformMarker`.

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/MarkersControllerTest.java
```diff
@@ -49,6 +46,26 @@ public class MarkersControllerTest {
   private AssetManager assetManager;
   private final float density = 1;
 
+  private static Messages.PlatformMarker.Builder defaultBuilder() {
```
stuartmorgan, line 49: `defaultMarkerBuilder`

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/MarkersControllerTest.java
```diff
@@ -183,17 +182,10 @@ public void controller_AddChangeAndRemoveMarkerWithClusterManagerId() {
...
-    final Map<String, Object> markerOptions2 = new HashMap<>();
-    markerOptions2.put("markerId", googleMarkerId);
-    markerOptions2.put("position", location2);
-    markerOptions2.put("clusterManagerId", clusterManagerId);
-    final List<Messages.PlatformMarker> updatedMarkers =
-        Collections.singletonList(
-            new Messages.PlatformMarker.Builder().setJson(markerOptions2).build());
+
+    builder.setPosition(
+        new Messages.PlatformLatLng.Builder().setLatitude(3.3).setLongitude(4.4).build());
```
stuartmorgan, line 187: These should use `latLng2` accessors, rather than duplicating magic numbers.

stuartmorgan (changes requested):
Nice! It's great to see so much fragile Java code being replaced.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -112,6 +112,8 @@ private static BitmapDescriptor toBitmapDescriptor(
         }
         final Map<?, ?> byteData = toMap(data.get(1));
         return getBitmapFromBytes(byteData, density, new BitmapDescriptorFactoryWrapper());
+      case "null":
```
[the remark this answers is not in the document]
yaakovschectman: Basically, because the tests need to be able to construct Markers, which means every non-nullable field must have a value assigned, and it therefore must be able to assign a non-null `BitmapDescriptor`. I initially tried using `defaultMarker`, but that apparently does not work until the `BitmapDescriptorFactory` is initialized in some way, which is not done in the tests. Do you have any preferable alternatives in mind?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
yaakovschectman: The reason I did not do so is because `title` is also checked for `null`, and thought assignment would be neater than calling `getTitle` twice in a row. Let me know if you disagree with that assessment.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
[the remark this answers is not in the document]
stuartmorgan: Can you make `BitmapDescriptorFactory` an argument to the controller class that you are testing, and then DI a mock version that returns mock `BitmapDescriptiors`?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
stuartmorgan: This line is `getSnippet`, not `getTitle`.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -84,16 +97,58 @@ class PlatformClusterManager {
   final String identifier;
 }
 
+/// Pigeon equivalent of the Offset class.
+class PlatformOffset {
```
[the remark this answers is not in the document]
yaakovschectman: `PlatformPoint` has integer `x` and `y`. `Offset` uses floating point values.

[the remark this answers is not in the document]
stuartmorgan: 👍🏻 Sorry, I thought I checked the types before leaving that comment.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -659,68 +661,35 @@ static void interpretMapConfiguration(
...
-      MarkerOptionsSink sink, Map<String, Object> infoWindow) {
-    String title = (String) infoWindow.get("title");
-    String snippet = (String) infoWindow.get("snippet");
+      MarkerOptionsSink sink, Messages.PlatformInfoWindow infoWindow) {
+    String title = infoWindow.getTitle();
+    String snippet = infoWindow.getSnippet();
     // snippet is nullable.
     if (title != null) {
       sink.setInfoWindowText(title, snippet);
     }
```
reidbaker, line 689:
```suggestion
    if (title != null) {
      sink.setInfoWindowText(title, infoWindow.getSnippet());
    }
```

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
reidbaker: I added a suggestion based on this thread.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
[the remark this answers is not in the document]
reidbaker: In the places where this is used we call toFloat should we be storing them as floats or not discarding information?

stuartmorgan:
> In the places where this is used we call toFloat should we be storing them as floats or not discarding information?

@reidbaker For some reason I can't reply directly to that comment inline, but `float` isn't a Dart type, so there's no way to have the possibility of precision loss somewhere between an idiomatic Dart API and the Java SDK that wants floats.

reidbaker:
I feel like I should have known that but thanks for the explanation. Ignore my comment then.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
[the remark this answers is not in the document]
yaakovschectman: It looks like I will need to do so in two test files, as well as create and encode a minimal bitmap byte object.

stuartmorgan:
Is this ready for me to re-review, or still in progress?

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
reidbaker: @yaakovschectman did you and stuart agree not to make these values required?

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -84,16 +97,58 @@ class PlatformClusterManager {
...
-  /// internal implementation details of that method.
-  // TODO(stuartmorgan): Replace this with structured data. This exists only to
-  //  allow incremental migration to Pigeon.
-  final Map<String?, Object?> json;
+  final double alpha;
+  final PlatformOffset anchor;
+  final bool consumeTapEvents;
+  final bool draggable;
+  final bool flat;
+  final Object icon;
```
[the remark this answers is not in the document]
reidbaker: This is still missing a note and TODO

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
reidbaker: Should there be a note here? (my guess is no since id is a string but thought it was worth asking.

reidbaker (approved):

yaakovschectman:
@reidbaker If I am interpreting correctly, Stuart stated his note on whether the fields are required was optional, and approved the PR. I do not see why we would need a note for `clusterManagerId`, or what that note would be.

stuartmorgan:
> This is still missing a note and TODO

Oops, I missed that in my re-review. Thanks for the catch!

> If I am interpreting correctly, Stuart stated his note on whether the fields are required was optional, and approved the PR.

Yes, I'm fine with there being default values. I have a very slight preference for `required` since it means we can't forget to carry values over, but the real solution to that is unit tests anyway.

stuartmorgan:
@reidbaker For the ID, are you asking because the Dart code uses specialized wrapper objects for each ID type? If so, I've been ignoring that at the Pigeon level since I think it's more busywork than benefit at the level of a transfer protocol, and just using strings directly. The objects have more value at the client-code level where things are being stored, passed around, etc. and could plausibly be passed back to the wrong APIs later by buggy client code.

reidbaker:
> @reidbaker For the ID, are you asking because the Dart code uses specialized wrapper objects for each ID type? If so, I've been ignoring that at the Pigeon level since I think it's more busywork than benefit at the level of a transfer protocol, and just using strings directly. The objects have more value at the client-code level where things are being stored, passed around, etc. and could plausibly be passed back to the wrong APIs later by buggy client code.

I was asking because you had a comment saying "same here" and I was trying to figure out what comment applied.

stuartmorgan:
> > @reidbaker For the ID, are you asking because the Dart code uses specialized wrapper objects for each ID type? If so, I've been ignoring that at the Pigeon level since I think it's more busywork than benefit at the level of a transfer protocol, and just using strings directly. The objects have more value at the client-code level where things are being stored, passed around, etc. and could plausibly be passed back to the wrong APIs later by buggy client code.
> 
> I was asking because you had a comment saying "same here" and I was trying to figure out what comment applied.

Oh, I see it inline now. I still don't understand how GitHub displays responses sometimes 😐 

I should have been clearer about what the "same" was; I meant it to refer to not needing default values for all the fields in that object, not about the comment. Sorry for the confusion!

PR #7331, opened 2024-08-07
[ci] Update repository for the release of Flutter 3.24

stuartmorgan:
Updates the repo for 3.24 per https://github.com/flutter/flutter/blob/master/docs/ecosystem/release/Updating-Packages-repo-for-a-stable-release.md

Also fixes some packages to list 3.22 as their minimum version, which hadn't already been done only because we were not previously doing the N-1 test for 3.19 as intended, so didn't notice cases where a package claimed 3.19 support but required a 3.22+ dependency.


.ci.yaml
```diff
@@ -292,21 +292,21 @@ targets:
     timeout: 30
     properties:
       target_file: analyze_legacy.yaml
-      channel: "3.22.0"
+      channel: "3.22.3"
```
stuartmorgan, line 295: This change looks wrong, but that's only because I apparently messed this up last time. It should have been 3.19.0 prior to this PR, which means we've been analyzing with N-2, but not actually with N-1, for the past stable cycle 😬

packages/url_launcher/url_launcher/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## NEXT
+
+* Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
```
yjbanov, line 3: @mdebbar @harryterkelsen Do we need to add something about the recent a11y fixes?

tarrinneal (approved):
Number go up!

yjbanov:
I don't know enough about the release process for packages, but updating SDK constraints without bumping the package version feels wrong. Unless the latest package versions haven't been published yet, I guess.

stuartmorgan:
> updating SDK constraints without bumping the package version feels wrong

Could you elaborate on why? We do this every time, and it's [our documented policy](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version).

stuartmorgan:
Version override: per [repo policy](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version)

stuartmorgan:
Looks like a bunch of web example apps had incorrectly-tight `web` constraints for their SDK constraint, which we didn't catch due to the N-1 testing mistake. I'm not sure if they can be relaxed, so I'm going to drop N-2 support for all of those packages.

yjbanov:
> Looks like a bunch of web example apps had incorrectly-tight `web` constraints for their SDK constraint, which we didn't catch due to the N-1 testing mistake. I'm not sure if they can be relaxed, so I'm going to drop N-2 support for all of those packages.

cc @kevmoo To confirm this is the right move.

yjbanov:
> > updating SDK constraints without bumping the package version feels wrong
> 
> Could you elaborate on why? We do this every time, and it's [our documented policy](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version).

To be clear, I don't mean updating SDK constraints in `git` is wrong. I only mean it would be wrong to publish a package to pub.dev under the same version but with different constraints. The pub client could cache the package and believe that nothing's changed, because the version is the same, while there's different SDK constraints lurking in the ecosystem under the same package version.

stuartmorgan:
> cc @kevmoo To confirm this is the right move.

If someone really wants to re-add support for 3.19 in those packages, it can always be done in a follow-up PR.

> I only mean it would be wrong to publish a package to pub.dev under the same version but with different constraints.

It would also be impossible, so I'm not following.

packages/url_launcher/url_launcher/CHANGELOG.md
mdebbar: The a11y fixes are still in master, not in a stable release yet. Not sure what you had in mind here.

yjbanov:
> It would also be impossible, so I'm not following.

We're good then.

yjbanov (approved):

PR #7356, opened 2024-08-09
[google_sign_in_ios] Adds Swift Package Manager support

loic-sharma:
Adds Swift Package Manager support to `google_sign_in_ios`.

Fixes https://github.com/flutter/flutter/issues/146904

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_sign_in/google_sign_in_ios/example/ios/Podfile
```diff
@@ -26,17 +26,14 @@ end
 require File.expand_path(File.join('packages', 'flutter_tools', 'bin', 'podhelper'), flutter_root)
 
 # Suppress warnings from transitive dependencies that cause analysis to fail.
-pod 'AppAuth', :inhibit_warnings => true
-pod 'GTMAppAuth', :inhibit_warnings => true
+inhibit_all_warnings!
```
loic-sharma, line 29: These lines inhibit warnings but also add `AppAuth` and `GTMAppAuth` as CocoaPod dependencies:

```
pod 'AppAuth', :inhibit_warnings => true
pod 'GTMAppAuth', :inhibit_warnings => true
```

If the SwiftPM feature is enabled, this caused duplicate modules and the build would fail.

This is a temporary workaround. In the future, we'll remove CocoaPod integration from these example apps. That's tracked by https://github.com/flutter/flutter/issues/148021.

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Package.swift
```diff
@@ -0,0 +1,46 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import PackageDescription
+
+let package = Package(
+  name: "google_sign_in_ios",
+  platforms: [
+    .iOS("12.0"),
+    .macOS("10.15")
```
loic-sharma, line 13: This is bumped up to macOS 10.15. Required to depend on [GoogleSignIn](https://github.com/google/GoogleSignIn-iOS/blob/65fb3f1aa6ffbfdc79c4e22178a55cd91561f5e9/Package.swift#L26)

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Package.swift
```diff
@@ -0,0 +1,46 @@
...
+  ],
+  products: [
+    .library(name: "google-sign-in-ios", targets: ["google_sign_in_ios"])
+  ],
+  dependencies: [
+    // AppAuth and GTMSessionFetcher are GoogleSignIn transitive dependencies.
+    // Depend on versions which define modules.
```
loic-sharma, line 20: This mirrors: https://github.com/flutter/packages/blob/f7b12561aeb6be92fc973441ae0f26cfaa8df8f4/packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios.podspec#L19-L23

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Package.swift
stuartmorgan: We [already require 10.15](https://github.com/flutter/packages/blob/4b00343963a1529598e602f0e3954c9fee9ec7d5/packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios.podspec#L28) (which is good, because otherwise this would be very hard to roll out in a non-disruptive way).

stuartmorgan:
From triage: still blocked on upstream issues.

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Package.swift
```diff
@@ -0,0 +1,52 @@
...
+      ],
+      resources: [
+        .process("Resources")
+      ],
+      cSettings: [
+        .headerSearchPath("include/google_sign_in_ios"),
+
+        // Ignore AppAuth's warnings.
+        // https://github.com/openid/AppAuth-iOS/issues/703
+        //.unsafeFlags(["-w"]),
```
loic-sharma, line 48: TODO remove?

script/configs/exclude_xcode_deprecation.yaml
```diff
@@ -1,2 +1,4 @@
 # TODO(louisehsu): Remove deprecation check when StoreKit 2 is adopted. https://github.com/flutter/flutter/issues/116383
 - in_app_purchase_storekit
+# Depends on AppAuth-iOS which uses deprecated APIs.
+- google_sign_in_ios
```
loic-sharma, line 4: Add a new line

loic-sharma:
This is failing `xcode-analyze` and `native-test` because our transitive dependency AppAuth has warnings, and we treat warnings as errors. This isn't a problem when using CocoaPods as [we've configured it to ignore these warnings](https://github.com/flutter/packages/blob/bc0c22d57910047fff7569ccb4413cf031304c7d/packages/google_sign_in/google_sign_in_ios/example/ios/Podfile#L28-L30). Unfortunately, SwiftPM doesn't let suppress warnings in our dependencies.

Our options are to either allow packages to opt-out of warnings as errors, or, fix the warnings in the upstream dependency: https://github.com/openid/AppAuth-iOS/pull/888

stuartmorgan:
I'm okay with temporarily turning off warnigs-as-errors for this package, but we should have a tracking issue with context for turning it back on.

Ideally we should also reference an SPM feature request for the ability to do what we used to do with Cocoapods. Having the ability to use different warning settings for code you control and code you don't is a fairly common feature in build systems.

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios.podspec
```diff
@@ -12,13 +12,13 @@ Enables Google Sign-In in Flutter apps.
...
-  s.public_header_files = 'Classes/**/*.h'
-  s.module_map = 'Classes/FLTGoogleSignInPlugin.modulemap'
+  s.source_files = 'google_sign_in_ios/Sources/google_sign_in_ios/**/*.{h,m}'
+  s.public_header_files = 'google_sign_in_ios/Sources/google_sign_in_ios/include/**/*.h'
+  s.module_map = 'google_sign_in_ios/Sources/google_sign_in_ios/include/FLTGoogleSignInPlugin.modulemap'
 
   # AppAuth and GTMSessionFetcher are GoogleSignIn transitive dependencies.
   # Depend on versions which defines modules.
-  s.dependency 'AppAuth', '>= 1.7.4'
+  s.dependency 'AppAuth', '>= 1.7.6'
```
loic-sharma, line 21: This fixes some but not all AppAuth-iOS warnings

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Package.swift
stuartmorgan: Do we really need this? Isn't not defining modules something that would only apply to CocoaPods, not SPM?

stuartmorgan:
> Ideally we should also reference an SPM feature request for the ability to do what we used to do with Cocoapods. Having the ability to use different warning settings for code you control and code you don't is a fairly common feature in build systems.

Did we find/file an issue for this that we can reference in the comments?

vashworth:
> > Ideally we should also reference an SPM feature request for the ability to do what we used to do with Cocoapods. Having the ability to use different warning settings for code you control and code you don't is a fairly common feature in build systems.
> 
> Did we find/file an issue for this that we can reference in the comments?

Looks like you can potentially use an `unsafeFlag`, however, ["use of unsafe flags makes the products containing this target ineligible for use by other packages"](https://developer.apple.com/documentation/packagedescription/swiftsetting/unsafeflags(_:_:)). I wonder if we could use it conditionally only during testing.

https://stackoverflow.com/a/64842804

stuartmorgan:
It's not clear to me that that SO answer address our (and the questioner's) use case; it's showing suppressing warnings on a target being defined as part of the package, not on a dependency.

Either way though, "I want warnings in my code, but not in code I don't control" is a pretty standard thing to want, and a use case we have a demonstrated need for, so we should request SPM fully (not only via unsafe flags) support it.

stuartmorgan:
To clarify, I don't consider this as blocking landing; if we do have a reference we can add, great, if not we can land without it but I think it's important that we provide feedback on ways that SPM doesn't work for us.

loic-sharma:
I did try various configurations of `unsafeFlags`, `cSettings`, `swiftSettings` to suppress dependency warnings. These worked locally on my machine. However, I could not get these suppressions to work in our CI. I was unable to reproduce locally, even if I used the same version of Xcode as our CI... It's not clear to me what's going wrong on our CI.

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Package.swift
loic-sharma: Good question, let me fiddle with this

vashworth (approved):
LGTM

PR #7394, opened 2024-08-12
[camera_avfoundation] enable more than 30 fps

misos1:
Camera plugin was crashing when I tried to set fps to 60 on most media presets (except maybe on 1280x720, although tested device supports 60 fps for up to 1920x1440 and can do 240 fps on 1280x720) because when is `activeVideoMinFrameDuration` and `activeVideoMaxFrameDuration` set to fps outside of what active format supports it throws exception. Now it tries to find a format which supports fps closest to wanted fps and clamps it if it cannot be set to exact value to prevent crashes. It searches for formats with the exact same resolution. For example in format list it can be like "1920x1080 { 3- 30 fps}", "1920x1080 { 3- 60 fps}" and "1920x1080 { 6-120 fps}, but when setting `sessionPreset` then "1920x1080 { 3- 30 fps}" is selected by default. On the tested device there are 2 "media subtypes" `420f` and `420v` for each format where the first is denoted as "supports wide color" in debug description and the system has tendency to choose this one. So it tries to preserve the media subtype to what is preferred by the system and this is also added to `highestResolutionFormatForCaptureDevice` (with lower priority than max resolution/fps). Also there was nested `lockForConfiguration` and `unlockForConfiguration` when using `ResolutionPreset.max` together with setting up fps.

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
@misos1 Could you split this into two different PRs? Having non-trivial changes to two unrelated plugins in the same PR makes review much more difficult.

packages/camera/camera_avfoundation/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 0.9.17+4
 
+* Adds possibility to use any supported FPS and fixes crash when using unsupported FPS.
```
hellohuanlin, line 3: *Adds ability to...

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -211,20 +211,50 @@ - (instancetype)initWithMediaSettings:(FCPPlatformMediaSettings *)mediaSettings
...
+      // find the format which frame rate ranges are closest to the wanted frame rate
+      CMVideoDimensions targetRes = self.videoDimensionsForFormat(_captureDevice.activeFormat);
+      double targetFrameRate = _mediaSettings.framesPerSecond.doubleValue;
+      FourCharCode preferredSubType =
+          CMFormatDescriptionGetMediaSubType(_captureDevice.activeFormat.formatDescription);
+      AVCaptureDeviceFormat *bestFormat = _captureDevice.activeFormat;
+      double bestFrameRate = [self frameRateForFormat:bestFormat closestTo:targetFrameRate];
+      double minDistance = fabs(bestFrameRate - targetFrameRate);
+      int bestSubTypeScore = 1;
+      for (AVCaptureDeviceFormat *format in _captureDevice.formats) {
```
hellohuanlin, line 237: can you also move this to a helper function?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -211,20 +211,50 @@ - (instancetype)initWithMediaSettings:(FCPPlatformMediaSettings *)mediaSettings
   [_motionManager startAccelerometerUpdates];
 
   if (_mediaSettings.framesPerSecond) {
+    [_mediaSettingsAVWrapper beginConfigurationForSession:_videoCaptureSession];
+
+    // Possible values for presets are hard-coded in FLT interface having
```
stuartmorgan, line 216: Why is all this code moving out of the locked section; is that safe?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -211,20 +211,50 @@ - (instancetype)initWithMediaSettings:(FCPPlatformMediaSettings *)mediaSettings
...
-      // corresponding AVCaptureSessionPreset counterparts.
-      // If _resolutionPreset is not supported by camera there is
-      // fallback to lower resolution presets.
-      // If none can be selected there is error condition.
-      if (![self setCaptureSessionPreset:_mediaSettings.resolutionPreset withError:error]) {
-        [_videoCaptureSession commitConfiguration];
-        [_captureDevice unlockForConfiguration];
-        return nil;
+      // find the format which frame rate ranges are closest to the wanted frame rate
+      CMVideoDimensions targetRes = self.videoDimensionsForFormat(_captureDevice.activeFormat);
```
stuartmorgan, line 229: You've added several calls to `self` in this init method. The fact that there was already such a call is [a problem](https://google.github.io/styleguide/objcguide.html#avoid-messaging-the-current-object-within-initializers-and--dealloc), and that should not be made worse by adding more.

If there are general, stateless utilities that you want to call from init, they should be turned into freestanding static functions.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -211,20 +211,50 @@ - (instancetype)initWithMediaSettings:(FCPPlatformMediaSettings *)mediaSettings
...
-      // Possible values for presets are hard-coded in FLT interface having
-      // corresponding AVCaptureSessionPreset counterparts.
-      // If _resolutionPreset is not supported by camera there is
-      // fallback to lower resolution presets.
-      // If none can be selected there is error condition.
-      if (![self setCaptureSessionPreset:_mediaSettings.resolutionPreset withError:error]) {
-        [_videoCaptureSession commitConfiguration];
-        [_captureDevice unlockForConfiguration];
-        return nil;
+      // find the format which frame rate ranges are closest to the wanted frame rate
```
stuartmorgan, line 228: Comments need to follow standard sentence rules. https://google.github.io/styleguide/objcguide.html#comments

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
stuartmorgan, line 229: Please don't abbreviate variable names. https://google.github.io/styleguide/objcguide.html#naming

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -211,20 +211,50 @@ - (instancetype)initWithMediaSettings:(FCPPlatformMediaSettings *)mediaSettings
...
+        CMVideoDimensions res = self.videoDimensionsForFormat(format);
+        if (res.width != targetRes.width || res.height != targetRes.height) {
+          continue;
+        }
+        double frameRate = [self frameRateForFormat:format closestTo:targetFrameRate];
+        double distance = fabs(frameRate - targetFrameRate);
+        FourCharCode subType = CMFormatDescriptionGetMediaSubType(format.formatDescription);
+        int subTypeScore = subType == preferredSubType ? 1 : 0;
+        if (distance < minDistance ||
+            (distance == minDistance && subTypeScore > bestSubTypeScore)) {
```
stuartmorgan, line 247: I'm confused by this condition. `bestSubTypeScore` is initialized to 1, and `subTypeScore` is always either 0 or 1. How could this last piece of the condition ever be true?

I'm not clear on what this score even is; this whole algorithm needs some explanatory comment at the beginning.

packages/camera/camera_avfoundation/CHANGELOG.md
stuartmorgan, line 3: Where is the new test that triggers the codepath that used to crash?

stuartmorgan (changes requested):

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: if `distance < minDistance` then `bestSubTypeScore` can be overwritten by 0. This line explains it: `subTypeScore = subType == preferredSubType ? 1 : 0`, the score of the subtype is 1 when it is the preferred subtype and 0 otherwise. This is how to choose by minimum distance and then from formats which have equal distance choose that one with preferred subtype, it gives to subtypes some order. Instead of `subTypeScore > bestSubTypeScore` there could be something like `bestSubType != preferredSubType && subType == preferredSubType` but that makes it less clear as it deviates from simple way how to order something by multiple properties `currentA < otherA || (currentA == otherA && currentB < otherB)`.

stuartmorgan: > but that makes it less clear

I strongly disagree that abstracting the thing that matters—whether the current type is a preferred type—behind a completely arbitrary numeric value makes this code easier to understand. The code you have there that you described as "less clear" makes sense to me when I read it, whereas I did not understand what this score was, or why something that could only have two possible values was being called a score in the first place.

packages/camera/camera_avfoundation/CHANGELOG.md
misos1: When the real `captureDevice.activeVideoMinFrameDuration` (or ...Max...) is set to value outside of what is supported by active format it immediately throws an objc exception. For tests, mocked wrappers are used so this cannot be triggered. But the test `testSettings_ShouldSelectFormatWhichSupports60FPS` tries to set 60 fps which actually triggered this crash on my device for media presets other than 1280x720.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: I did not see anything saying that the order of `beginConfiguration` and `lockForConfiguration` is important or that setting the session preset must be under that lock. There is example where it is used in that order: https://developer.apple.com/documentation/avfoundation/avcapturedevice/1389221-activeformat?language=objc

I moved `setCaptureSessionPreset` outside of lock because it can itself call `lockForConfiguration` and `unlockForConfiguration` (and it is also called elsewhere even without `beginConfiguration`).

misos1: Seems `lockForConfiguration` is actually recursive from my observation (I could not get mentioned exceptions until now) as locking twice and unlocking once does not cause exceptions but unlocking twice yes. So I can revert this change (?).

Also it seems not all `commitConfiguration` calls go through `_mediaSettingsAVWrapper`.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -251,6 +254,55 @@ - (instancetype)initWithMediaSettings:(FCPPlatformMediaSettings *)mediaSettings
   return self;
 }
 
+static void selectBestFormatForRequestedFrameRate(
```
stuartmorgan, line 257: These helpers should be defined above their call sites; I'm suprised this is even compiling.

stuartmorgan, line 257: Per the Obj-C style guide, all declarations require declaration comments explaining them.

In this case, that should include covering what "select" means exactly, given that this doesn't return anything. Presumably it's modifying one or more arguments, but I can't tell without reading the implementation.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -251,6 +254,55 @@ - (instancetype)initWithMediaSettings:(FCPPlatformMediaSettings *)mediaSettings
...
+      bestSubType = subType;
+    }
+  }
+  if (![bestFormat isEqual:captureDevice.activeFormat]) {
+    captureDevice.activeFormat = bestFormat;
+  }
+  mediaSettings.framesPerSecond = @(bestFrameRate);
+}
+
+static double bestFrameRateForFormat(AVCaptureDeviceFormat *format, double targetFrameRate) {
```
stuartmorgan, line 292: This also needs a comment (e.g., explaining what "best" means).

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -251,6 +254,55 @@ - (instancetype)initWithMediaSettings:(FCPPlatformMediaSettings *)mediaSettings
...
+    CMVideoDimensions resolution = videoDimensionsForFormat(format);
+    if (resolution.width != targetResolution.width ||
+        resolution.height != targetResolution.height) {
+      continue;
+    }
+    double frameRate = bestFrameRateForFormat(format, targetFrameRate);
+    double distance = fabs(frameRate - targetFrameRate);
+    FourCharCode subType = CMFormatDescriptionGetMediaSubType(format.formatDescription);
+    if (distance < minDistance || (distance == minDistance && subType == preferredSubType &&
+                                   bestSubType != preferredSubType)) {
```
stuartmorgan, line 279: Here and in the very similar code below, I think you could improve the readability of the complex condition by extracting a local variable just before this `if`:

```objc
BOOL isBetterTypeMatch = subType == preferredSubType && bestSubType != preferredSubType;
```

Then you can use that in the `if` condition. (Yes, this reduces short-circuiting, but this isn't an expensive calculation.)

stuartmorgan (changes requested):

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: Or maybe `subType == preferredSubType` can be extracted into variable `subTypeIsPreferred` and instead of `bestSubType` will be stored `bestSubTypeIsPreferred` then that condition would become `distance == minDistance && subTypeIsPreferred && !bestSubTypeIsPreferred` (which is actually similar to that original with scores just there is bool instead of int and `a && !b` instead of `a > b`).

stuartmorgan: Agreed, that version definitely addresses my readability concern.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -153,6 +153,60 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
...
+      bestFrameRate = frameRate;
+      minDistance = distance;
+    }
+  }
+  return bestFrameRate;
+}
+
+// Finds format with same resolution as current activeFormat for which bestFrameRateForFormat
+// returned frame rate closest to mediaSettings.framesPerSecond. Preferred are formats with the
+// same subtype as current activeFormat. Sets this format as activeFormat and also updates
```
stuartmorgan, line 173: How about "Sets this format as the activeFormat in captureDevice and [...]"; I wasn't sure what `activeFormat` was referring to without reading the implementation.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -153,6 +153,60 @@ - (instancetype)initWithCameraName:(NSString *)cameraName
...
+    BOOL isSubTypePreferred = subType == preferredSubType;
+    if (distance < minDistance ||
+        (distance == minDistance && isSubTypePreferred && !isBestSubTypePreferred)) {
+      bestFormat = format;
+      bestFrameRate = frameRate;
+      minDistance = distance;
+      isBestSubTypePreferred = isSubTypePreferred;
+    }
+  }
+  if (![bestFormat isEqual:captureDevice.activeFormat]) {
```
stuartmorgan, line 204: Is there evidence in docs or actual testing that this check is necessary? Standard convention for setters in Obj-C is for them to explicitly handle set-to-same-value as a no-op, so I would expect that we could just unconditionally call the setter and get exactly the same behavior.

stuartmorgan (approved):
Minor nits, but otherwise the code LGTM, thanks!

I'm not very familiar with the camera APIs, so I'll defer to @hellohuanlin to review the behavior itself.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -537,23 +580,32 @@ - (BOOL)setCaptureSessionPreset:(FCPPlatformResolutionPreset)resolutionPreset
         return NO;
       }
   }
+  CMVideoDimensions size = self.videoDimensionsForFormat(_captureDevice.activeFormat);
+  _previewSize = CGSizeMake(size.width, size.height);
   _audioCaptureSession.sessionPreset = _videoCaptureSession.sessionPreset;
   return YES;
 }
 
 /// Finds the highest available resolution in terms of pixel count for the given device.
```
hellohuanlin, line 589: Nit: "... if same pixel count, use the subtype as the tie breaker."

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: I will need to check this, I think it changes `sessionPreset` to `AVCaptureSessionPresetInputPriority` even if format is the same as it was, please wait with merging until I can get into this and other comments left.

misos1: Yes it changes `sessionPreset` so the idea is to not change it if the format does not need any change. Although maybe it actually should be changed to prevent the session from changing to an active format which no longer supports fps set by the client: "When a client sets the session preset to anything other than AVCaptureSessionPresetInputPriority, the session resumes responsibility for configuring inputs and outputs, and is free to change its inputs' activeFormat as needed."

stuartmorgan:
Sorry, this slipped through the cracks on getting landed.

(@hellohuanlin Please make sure to add the `autosubmit` label when doing the last review on a non-member PR.)

PR #7423, opened 2024-08-15
Serve benchmarks with COOP/COEP headers in wasm mode.

eyebrowsoffire:
We need `crossOriginIsolated` for skwasm to work. Otherwise, this just falls back and benchmarks JS + canvaskit.

### DIFF
--- a/packages/web_benchmarks/CHANGELOG.md
+++ b/packages/web_benchmarks/CHANGELOG.md
@@ -1,6 +1,9 @@
-## NEXT
+## 2.0.2
 
 * Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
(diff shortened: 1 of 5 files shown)


yjbanov:
Can we test this by doing a simple `dart:io` HTTP fetch using a Dart test and checking for headers?

eyebrowsoffire:
> Can we test this by doing a simple `dart:io` HTTP fetch using a Dart test and checking for headers?

I think I'm actually going to add some data to the benchmark payload that says some info about the environment, including whether it is wasm. Then I'll go ahead and confirm it matches that in the tests. I think that's a bit more comprehensive and gives me a better level of confidence than just inspecting the HTTP headers.

kenzieschmoll (approved):

PR #7447, opened 2024-08-19
[camera_windows] Set device media type for video preview explicitly

dasyad00:
Fixes preview aspect ratio on Windows on some webcams (See https://github.com/flutter/flutter/issues/122966).

Before (main branch 7c1a05c8b12274e940e39adcb42896a32f57df32):

image

After:

image

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_windows/CHANGELOG.md
+++ b/packages/camera/camera_windows/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.2.6+1
+
+* Fixes black bars on camera preview [#122966](https://github.com/flutter/flutter/issues/122966).
+
 ## 0.2.6
 
 * Reverts streaming frame support, as the implementation was incorrect and never
--- a/packages/camera/camera_windows/windows/capture_controller.cpp
+++ b/packages/camera/camera_windows/windows/capture_controller.cpp
@@ -467,10 +467,11 @@ HRESULT CaptureControllerImpl::FindBaseMediaTypes() {
     return hr;
   }
 
-  return FindBaseMediaTypes(source.Get());
+  return FindBaseMediaTypesForSource(source.Get());
 }
 
-HRESULT CaptureControllerImpl::FindBaseMediaTypes(IMFCaptureSource* source) {
+HRESULT CaptureControllerImpl::FindBaseMediaTypesForSource(
+    IMFCaptureSource* source) {
   // Find base media type for previewing.
   if (!FindBestMediaType(
           (DWORD)MF_CAPTURE_ENGINE_PREFERRED_SOURCE_STREAM_FOR_VIDEO_PREVIEW,
@@ -579,7 +580,7 @@ void CaptureControllerImpl::StartPreview() {
 
   if (!base_preview_media_type_) {
     // Enumerates mediatypes and finds media type for video capture.
-    hr = FindBaseMediaTypes(source.Get());
+    hr = FindBaseMediaTypesForSource(source.Get());
     if (FAILED(hr)) {
       return OnPreviewStarted(GetCameraResult(hr),
                               "Failed to initialize video preview");
--- a/packages/camera/camera_windows/windows/capture_controller.cpp
+++ b/packages/camera/camera_windows/windows/capture_controller.cpp
@@ -575,7 +575,7 @@ void CaptureControllerImpl::StartPreview() {
   hr = capture_engine_->GetSource(&source);
   if (FAILED(hr)) {
     return OnPreviewStarted(GetCameraResult(hr),
-                            "Failed to initialize video preview");
+                            "Failed to get capture engine source");
   }
 
   if (!base_preview_media_type_) {
@@ -592,7 +592,7 @@ void CaptureControllerImpl::StartPreview() {
       base_preview_media_type_.Get());
   if (FAILED(hr)) {
     return OnPreviewStarted(GetCameraResult(hr),
-                            "Failed to initialize video preview");
+                            "Failed to set video preview output format");
   }
 
   texture_handler_->UpdateTextureSize(preview_frame_width_,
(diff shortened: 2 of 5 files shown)


stuartmorgan:
Thanks for the contribution! This will need a test, per the PR checklist, before it can be reviewed.

dasyad00:
Since the bug arose due to a missing function call to Microsoft's Media Foundation API, the only test I can reasonably add is to verify the function is now called.

stuartmorgan:
From triage: @cbracken it looks like this should be ready for review.

cbracken (approved):

Looks great -- thanks for your contribution!

packages/camera/camera_windows/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## NEXT
+## 0.2.6
```
stuartmorgan, line 1: For versions <1, Dart convention is to use the build number for bugfixes, so this should be `0.2.5+1`.

packages/camera/camera_windows/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
...
 
 * Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
+* Fix black bars on camera preview [#122966](https://github.com/flutter/flutter/issues/122966)
```
stuartmorgan, line 4: "Fixes". Also, this is missing a period. Please see the CHANGELOG style guide linked from the PR checklist.

packages/camera/camera_windows/windows/capture_controller.cpp
```diff
@@ -479,6 +479,13 @@ HRESULT CaptureControllerImpl::FindBaseMediaTypes() {
     return E_FAIL;
   }
 
+  hr = source->SetCurrentDeviceMediaType(
```
stuartmorgan, line 482: It seems extremely surprising to have a method called `FindBaseMediaTypes()` have the side-effect of *setting* a media type. Why is this method the place where this is called?

stuartmorgan (changes requested):

packages/camera/camera_windows/CHANGELOG.md
dasyad00: I have fixed the versioning in 4ffbedd

packages/camera/camera_windows/CHANGELOG.md
dasyad00: The changelog should now be following the style guide as of 4ffbedd

packages/camera/camera_windows/windows/capture_controller.cpp
dasyad00: [`SetCurrentDeviceMediaType`](https://learn.microsoft.com/en-us/windows/win32/api/mfcaptureengine/nf-mfcaptureengine-imfcapturesource-setcurrentdevicemediatype) is a class function of `IMFCaptureSource` which was conveniently already accessible in the function scope `FindBaseMediaType`. For the bug fix, the function only needs to be called after `base_preview_media_type_` is set, which is when `FindBestMediaType` is called with it.

We can alternatively move it inside the function `CaptureControllerImpl::StartPreview` after `base_preview_media_type_` is set, but then we would need to prepare a `IMFCaptureSource` object for this scope.

stuartmorgan: > For the bug fix, the function only needs to be called after `base_preview_media_type_` is set, which is when `FindBestMediaType` is called with it.

My primary concern is the maintainability of this code in the long term, rather than minimizing the diff. Having a method indicate that it gets a value have a side effect of changing a value is the kind of thing that makes maintaining code correctly a lot harder over the longer term.

> We can alternatively move it inside the function `CaptureControllerImpl::StartPreview`

That sounds like a much clearer location for setting values related to displaying the preview.

stuartmorgan:
@dasyad00 Are you still planning on updating this to address the remaining feedback?

dasyad00:
@stuartmorgan Sorry, I was occupied with other things lately. I have just pushed the refactor commit.

packages/camera/camera_windows/windows/capture_controller.h
```diff
@@ -180,6 +180,7 @@ class CaptureControllerImpl : public CaptureController,
   // Enumerates video_sources media types and finds out best resolution
   // for preview and video capture.
   HRESULT FindBaseMediaTypes();
+  HRESULT FindBaseMediaTypes(IMFCaptureSource* source);
```
stuartmorgan, line 183: Please give these different names, and their own declaration comments explaining what they do. Function overloading should only be used when the methods do exactly the same thing, just with different input types.

The new method could be called `FindBaseMediaTypesForSource`.

packages/camera/camera_windows/windows/capture_controller.cpp
```diff
@@ -566,15 +570,30 @@ void CaptureControllerImpl::StartPreview() {
 
   HRESULT hr = S_OK;
 
+  ComPtr<IMFCaptureSource> source;
+  hr = capture_engine_->GetSource(&source);
+  if (FAILED(hr)) {
+    return OnPreviewStarted(GetCameraResult(hr),
+                            "Failed to initialize video preview");
```
stuartmorgan, line 577: Please don't use the same error message for three different failures; we want to be able to distinguish what failed. Each case in the method should describe what specifically failed.

stuartmorgan (changes requested):
Thanks, the refactoring makes for a much clearer flow! Just minor comments.

packages/camera/camera_windows/windows/capture_controller.h
dasyad00: I have renamed the method in ff678d6e23138246d2d7553d3093b926186d2405

packages/camera/camera_windows/windows/capture_controller.cpp
dasyad00: I have clarified the error messages in 79610aa2c52f6971f4e26ab1ada6dd8be04fb6f0

PR #7497, opened 2024-08-26
Revert "[video_player] Relands #6456: Uses SurfaceProducer, this time with setCallback for suspend/resume lifecycles"

LinXunFeng:
Reverts https://github.com/flutter/packages/pull/6989
Fixes https://github.com/flutter/flutter/issues/154054 , https://github.com/flutter/flutter/issues/154050


matanlurey:
Thanks. I agree we should land this revert and roll forward with the SDK constraint set properly.

stuartmorgan:
I've retracted 2.7.0 as well.

PR #7507, opened 2024-08-26
[google_maps_flutter_android] Convert `PlatformCameraUpdate` to pigeon.

yaakovschectman:
Because this also converts the platform interface Camera type, this will need to be a federated change, unless we come up with a better idea of how to convert that class.

https://github.com/flutter/flutter/issues/152928

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -339,37 +339,50 @@ static CameraPosition toCameraPosition(Object o) {
...
-        throw new IllegalArgumentException("Cannot interpret " + o + " as CameraUpdate");
+  static CameraUpdate cameraUpdateFromPigeon(Messages.PlatformCameraUpdate update, float density) {
+    Messages.PlatformNewCameraPosition newCameraPosition = update.getNewCameraPosition();
+    if (newCameraPosition != null) {
+      return CameraUpdateFactory.newCameraPosition(
+          cameraPositionFromPigeon(newCameraPosition.getCameraPosition()));
+    }
+    Messages.PlatformNewLatLng newLatLng = update.getNewLatLng();
+    if (newLatLng != null) {
+      return CameraUpdateFactory.newLatLng(latLngFromPigeon(newLatLng.getLatLng()));
```
reidbaker, line 350: I may not be following the types here but is it possible to make this CameraUpdateFactory.newLatLng(latLngFromPigeon(newLatLng); It seems, based on names alone that we should not need so many unboxing/reboxing of lat long data.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -339,37 +339,50 @@ static CameraPosition toCameraPosition(Object o) {
...
-        return CameraUpdateFactory.zoomIn();
-      case "zoomOut":
-        return CameraUpdateFactory.zoomOut();
-      case "zoomTo":
-        return CameraUpdateFactory.zoomTo(toFloat(data.get(1)));
-      default:
-        throw new IllegalArgumentException("Cannot interpret " + o + " as CameraUpdate");
+  static CameraUpdate cameraUpdateFromPigeon(Messages.PlatformCameraUpdate update, float density) {
+    Messages.PlatformNewCameraPosition newCameraPosition = update.getNewCameraPosition();
+    if (newCameraPosition != null) {
```
reidbaker, line 344: I see how this code is a conversion from what was here before but I think the if statement make it harder to read what should happen. Before the toCameraUpdate method either updated the camera position, the lat/lng or the bounds. Now that is much harder to understand. 

Try restructuring this method to make it easier for a human reader and/or adding java doc comments.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -796,6 +796,48 @@ class GoogleMapsFlutterAndroid extends GoogleMapsFlutterPlatform {
     return PlatformTileOverlay(
         json: tileOverlay.toJson() as Map<String, Object?>);
   }
+
+  static PlatformCameraUpdate _platformCameraUpdateFromCameraUpdate(
+      CameraUpdate update) {
+    if (update is NewCameraPosition) {
```
reidbaker, line 802: You might have discussed this with @stuartmorgan if so feel free to resolve but can cameraUpdate include an enum of the type of update instead of these long object type checks?

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/camera.dart
```diff
@@ -155,49 +147,131 @@ class CameraUpdate {
...
   static CameraUpdate zoomTo(double zoom) {
-    return CameraUpdate._(<Object>['zoomTo', zoom]);
+    return ZoomTo(zoom);
   }
 
-  final Object _json;
-
   /// Converts this object to something serializable in JSON.
-  Object toJson() => _json;
+  Object toJson();// => _json;
```
stuartmorgan, line 182: The commented out code should be removed.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/camera.dart
```diff
@@ -155,49 +147,131 @@ class CameraUpdate {
...
+}
+
+/// Defines a camera move to a new position.
+class NewCameraPosition extends CameraUpdate {
```
stuartmorgan, line 186: I think it would probably be easier to keep track of all of these classes if they had a `CameraUpdate` prefix.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
stuartmorgan: This was discussed a bit for the Pigeon classes (where I don't think the enum probably helps us), but not for the Dart code.

I think for Dart, where we will be creating a new permanent public API surface, it would be better to add a type enum, and to make all of these classes private, so that we're not locking ourselves into a specific class structure forever.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -40,17 +40,79 @@ class PlatformCameraPosition {
...
+
+  final PlatformNewCameraPosition? newCameraPosition;
+  final PlatformNewLatLng? newLatLng;
+  final PlatformNewLatLngBounds? newLatLngBounds;
+  final PlatformNewLatLngZoom? newLatLngZoom;
+  final PlatformScrollBy? scrollBy;
+  final PlatformZoomBy? zoomBy;
+  final bool zoomIn;
+  final bool zoomOut;
+  final PlatformZoomTo? zoomTo;
```
stuartmorgan, line 63: This isn't the design we discussed in the issue; what's the reason for the change?

yaakovschectman: If the change you are referring to is `PlatformCameraUpdate` containing members instead of each option extending it as a base class, this is due to the limitations of Pigeon, as we discussed privately. If each of the individual subclasses extend `PlatformCameraUpdate`, but are not reachable by the exposed APIs, they will not be converted by Pigeon at all, and therefore be unusable. The only other alternative I can think of that still keeps the subtypes reachable is to overload `animate/moveCamera` for each of the subtypes, which does not sound very preferable.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
yaakovschectman: `newLatLng` is an instance of `PlatformNewLatLng`, which represents the camera move update to the new position. `latLngFromPigeon` accepts a `PlatformLatLng`, which is just a boxed pair of numeric coordinates, not a `PlatformNewLatLng`. I intend to add a `CameraUpdate` prefix to the relevant classes, including `PlatformNewLatLng`, as per Stuart's comment, which I believe will make this more clear.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
stuartmorgan: Oh, I didn't realize that was the context for that question.

@tarrinneal How hard would it be to generate all classes in the current implementation? I never remember if it works this way because it used to have to when Pigeon used mirrors, and it just stayed that way, or if it had to change to this way when we moved *off* of mirrors.

tarrinneal: I'm not sure I'm understanding the question here. @stuartmorgan

yaakovschectman: @tarrinneal I believe he is basically asking the following. Suppose there is some pigeon defined class `X` which is exposed via APIs and is therefore converted by the code generation when running pigeon. Is there a way to define multiple subclasses of `X` that will also be converted by code gen without defining new functions that explicitly use them?

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
yaakovschectman: If these classes were private, would that not then make their members inaccessible in this file?

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
stuartmorgan: We chatted about this briefly in triage yesterday, and @tarrinneal believes this would be easy to fix in Pigeon, in which case we should probably do that and then update this PR accordingly.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
stuartmorgan: 🤦🏻 

Sorry, I was thinking about mirroring the Java SDK's approach, but there they only need to consume the properties internally to the SDK.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
yaakovschectman: Okay, so how would we do that? By "fix in Pigeon", do you mean pushing a change to Pigeon itself?

tarrinneal: I'll push the pigeon fix asap

tarrinneal: https://github.com/flutter/packages/pull/7529

yaakovschectman: @stuartmorgan It sounds like what Tarrin has fixed in Pigeon does not actually resolve this, as inheritance is not supported in Pigeon in any capacity.

tarrinneal: Adding inheritance is a much bigger project I don't think that is something that can ready in a reasonable time frame for this pr.

reidbaker: Could we or do we want to bypass inheritance and say that the cameraUpdate object contains an Object and an a defined enum for the type. Then in methods that take cameraUpdate switch on an enum and cast.

stuartmorgan: > I think I am still blocked. It looks like Pigeon's generated code does not preserve inheritance between classes

This is why the issue describes using `Object` as the item type. Inheritance would be nice, but shouldn't be blocking.

> Could we or do we want to bypass inheritance and say that the cameraUpdate object contains an Object and an a defined enum for the type. Then in methods that take cameraUpdate switch on an enum and cast.

We don't even need the enum; we can still do type detection on the class without inheritance.

We could do it if you feel strongly, but I think it's just adding duplicate information. Checking the actual type before casting to that type seems much better to me than checking an enum and then casting.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -340,37 +340,56 @@ static CameraPosition toCameraPosition(Object o) {
...
+    if (cameraUpdate instanceof Messages.PlatformCameraUpdateZoomTo) {
+      Messages.PlatformCameraUpdateZoomTo zoomTo =
+          (Messages.PlatformCameraUpdateZoomTo) cameraUpdate;
+      return CameraUpdateFactory.zoomTo(zoomTo.getZoom().floatValue());
+    }
+    if (cameraUpdate instanceof Messages.PlatformCameraUpdateZoom) {
+      Messages.PlatformCameraUpdateZoom zoom = (Messages.PlatformCameraUpdateZoom) cameraUpdate;
+      return (zoom.getOut()) ? CameraUpdateFactory.zoomOut() : CameraUpdateFactory.zoomIn();
+    }
+    throw new IllegalArgumentException("PlatformCameraUpdate must have one non-null member.");
```
reidbaker, line 392: This error message does not make sense to me since the throw is not related to a value being null.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -494,6 +513,14 @@ static Point pointFromPigeon(Messages.PlatformPoint point) {
     return new Point(point.getX().intValue(), point.getY().intValue());
   }
 
+  @Nullable
+  static Point pointFromPigeon(@Nullable Messages.PlatformOffset point, float density) {
+    if (point == null) {
+      return null;
+    }
+    return new Point((int) (point.getDx() * density), (int) (point.getDy() * density));
```
reidbaker, line 521: why should this be a truncation (cast to int) instead of a round?

reidbaker (approved):

reidbaker:
Please hold on @stuartmorgan's approval as well.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
yaakovschectman: The previous implementation replaced by this PR called `toPoint`, which called `toPixels`, which performed a cast to int. I did not intend this PR to change any implementation details beyond how the data is structured.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
yaakovschectman: It made sense before, but now I need to update it.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/camera.dart
```diff
@@ -110,42 +110,59 @@ class CameraPosition {
       'CameraPosition(bearing: $bearing, target: $target, tilt: $tilt, zoom: $zoom)';
 }
 
+/// Indicates which type of camera update this instance represents.
+enum CameraUpdateType {
```
stuartmorgan, line 114: Ideally we would make CameraUpdate a sealed class so that we could just do exhaustive checked switches over the classes themselves, but since that would technically be a breaking change this works.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -799,6 +799,60 @@ class GoogleMapsFlutterAndroid extends GoogleMapsFlutterPlatform {
...
+                    : _platformOffsetFromOffset(update.focus!)));
+      case CameraUpdateType.zoomIn:
+        update as CameraUpdateZoomIn;
+        return PlatformCameraUpdate(
+            cameraUpdate: PlatformCameraUpdateZoom(out: false));
+      case CameraUpdateType.zoomOut:
+        update as CameraUpdateZoomOut;
+        return PlatformCameraUpdate(
+            cameraUpdate: PlatformCameraUpdateZoom(out: true));
+      case CameraUpdateType.scrollBy:
```
stuartmorgan, line 849: This is the only case currently covered by Dart unit tests. It was okay to only test one update type when the details of how they were serialized was up to a different package, but now that there is logic for each case within this package, the unit tests need to be updated so there's coverage of all of these branches.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/camera.dart
```diff
@@ -110,42 +110,59 @@ class CameraPosition {
...
+  zoomIn,
+  /// Zoom out
+  zoomOut,
+}
+
 /// Defines a camera move, supporting absolute moves as well as moves relative
 /// the current position.
-class CameraUpdate {
-  const CameraUpdate._(this._json);
+abstract class CameraUpdate {
```
stuartmorgan, line 137: All of the changes here need Dart unit tests. I would recommend one new test for each factory that validates:
- that the class is as expected,
- that the `updateType` is as expected, and
- that the properties were passed through correctly.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -340,37 +340,56 @@ static CameraPosition toCameraPosition(Object o) {
     return builder.build();
   }
 
-  static CameraUpdate toCameraUpdate(Object o, float density) {
-    final List<?> data = toList(o);
-    switch (toString(data.get(0))) {
-      case "newCameraPosition":
-        return CameraUpdateFactory.newCameraPosition(toCameraPosition(data.get(1)));
```
stuartmorgan, line 347: toCameraPosition appears to be unused now, so is dead code that should be removed.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -340,37 +340,56 @@ static CameraPosition toCameraPosition(Object o) {
...
-      case "newLatLng":
-        return CameraUpdateFactory.newLatLng(toLatLng(data.get(1)));
-      case "newLatLngBounds":
-        return CameraUpdateFactory.newLatLngBounds(
-            toLatLngBounds(data.get(1)), toPixels(data.get(2), density));
```
stuartmorgan, line 352: toLatLngBounds is also dead code.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -340,37 +340,56 @@ static CameraPosition toCameraPosition(Object o) {
...
-        return CameraUpdateFactory.newLatLngZoom(toLatLng(data.get(1)), toFloat(data.get(2)));
-      case "scrollBy":
-        return CameraUpdateFactory.scrollBy( //
-            toFractionalPixels(data.get(1), density), //
-            toFractionalPixels(data.get(2), density));
-      case "zoomBy":
-        if (data.size() == 2) {
-          return CameraUpdateFactory.zoomBy(toFloat(data.get(1)));
-        } else {
-          return CameraUpdateFactory.zoomBy(toFloat(data.get(1)), toPoint(data.get(2), density));
```
stuartmorgan, line 363: toPoint is also dead code, which cascades to toPixels and toFractionalPixels being dead code as well.

I believe if you edit this file in Android Studio, the "problems" pane will flag anything that is `static` and unused.

stuartmorgan: (As a general note, I would expect that when the Pigeon conversion of `google_maps_flutter_android` is complete, there would be no methods in this file that take an `Object` as their parameter, so it would be good to audit to make sure that dead code is removed, and anything left already has a tracking bug. Per [this comment](https://github.com/flutter/flutter/issues/152926) I'm not sure we have all the remaining work tracked in issues.)

packages/google_maps_flutter/google_maps_flutter/example/pubspec.yaml
```diff
@@ -33,3 +33,8 @@ flutter:
   uses-material-design: true
   assets:
     - assets/
+
+# FOR TESTING AND INITIAL REVIEW ONLY. DO NOT MERGE.
+# See https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins
+dependency_overrides:
```
stuartmorgan, line 39: The changes to `google_maps_flutter` can be reverted.

stuartmorgan (commented):
Some comments, but nothing structural at this point. Feel free to split `google_maps_flutter_platform_interface` out into its own PR and just addressing the testing in that PR.

stuartmorgan (changes requested):
(I forgot to switch from Comment to Request Changes to reflect the comments.)

yaakovschectman:
> Some comments, but nothing structural at this point. Feel free to split `google_maps_flutter_platform_interface` out into its own PR and just addressing the testing in that PR.

@stuartmorgan I do not think we can address the testing in the PR that updates the platform interface. The tests would be in the android package, and no PR can land that refers to changes in the platform interface until the platform interface's package's PR lands. Perhaps I have misunderstood what you mean.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -40,17 +40,66 @@ class PlatformCameraPosition {
...
-  /// The update data, as JSON. This should only be set from
-  /// CameraUpdate.toJson, and the native code must interpret it according to the
-  /// internal implementation details of the CameraUpdate class.
-  // TODO(stuartmorgan): Update the google_maps_platform_interface CameraUpdate
-  //  class to provide a structured representation of an update. Currently it
-  //  uses JSON as its only state, so there is no way to preserve structure.
-  //  This wrapper class exists as a placeholder for now to at least provide
-  //  type safety in the top-level call's arguments.
-  final Object json;
+  PlatformCameraUpdate({required this.cameraUpdate});
```
stuartmorgan, line 43: Making this a named argument vs a positional argument doesn't seem to add any clarity, and it makes the already-verbose code using this class even more verbose. Having the single argument just be positional would be clearer in this case IMO.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -40,17 +40,66 @@ class PlatformCameraPosition {
...
+
+  /// This Object shall be any of the below classes prefixed with
+  /// PlatformCameraUpdate. Each such class represents a different type of
+  /// camera update, and each holds a different set of data, preventing the
+  /// use of a single unified class. Pigeon does not support inheritance, which
+  /// prevents a more strict type bound.
```
stuartmorgan, line 49: You can reference https://github.com/flutter/flutter/issues/117819 here.

packages/google_maps_flutter/google_maps_flutter_android/test/google_maps_flutter_android_test.dart
```diff
@@ -104,7 +104,7 @@ void main() {
     expect(bounds, expectedBounds);
   });
 
-  test('moveCamera calls through', () async {
+  test('moveCamera calls through scrollBy', () async {
```
stuartmorgan, line 107: I'm confused by the naming changes here; the API being called is `moveCamera` at both layers; nothing called `scrollBy` is being called in either method.

You could say "calls through with expected parameters" if you want to be more explicit about what's being verified.

packages/google_maps_flutter/google_maps_flutter_android/test/google_maps_flutter_android_test.dart
```diff
@@ -774,6 +782,164 @@ void main() {
     },
   );
 
+  test('moveCamera calls through newCameraPosition', () async {
```
stuartmorgan, line 785: `moveCamera calls though with expected newCameraPosition`; similarly with the methods below.

stuartmorgan (changes requested):
Thanks for the expanded testing and Java cleanup! Mostly looks great, just a few comments.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
reidbaker: I believe that this is a logic change. toPixels calls toFractionalPixels which multiplies a float by the density. 
```
  private static float toFractionalPixels(Object o, float density) {
    return toFloat(o) * density;
  }
```

In none of the code I traced did I see us dropping data by an int cast.

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -494,16 +499,16 @@ static Point pointFromPigeon(Messages.PlatformPoint point) {
...
   }
 
-  static Messages.PlatformPoint pointToPigeon(Point point) {
-    return new Messages.PlatformPoint.Builder().setX((long) point.x).setY((long) point.y).build();
-  }
-
-  private static LatLngBounds toLatLngBounds(Object o) {
-    if (o == null) {
+  @Nullable
+  static Point pointFromPigeon(@Nullable Messages.PlatformOffset point, float density) {
```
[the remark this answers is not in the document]
stuartmorgan: I believe I established that naming convention in this file. The problem is that during this transition there have been a bunch of cases where there are two different version of all the methods: one that converts method channel primitives, for the unconverted code, and one that converts Pigeon objects.

In that context, the fact that the source object is Pigeon is very important. Especially because the old methods take `Object` which means you can accidentally pass a Pigeon object to them (which will then crash).

I would have no objection to mass renaming these when the conversion is done and the old versions don't exist any more. Or we could mass rename the remaining old methods now to something that clearly identifies them, and then mass rename the Pigeon versions to use the generic names.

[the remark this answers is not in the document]
reidbaker: That makes sense and makes me glad I asked the question.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
yaakovschectman: `toPixels` converts the result of `toFractionalPixels` to int via (int) cast: https://github.com/flutter/packages/pull/7507/files#diff-26752280ec885e4465baf81a4771b5a7e6fd1041d333c9fedd712bcc5c3916c7L534.  
`toPoint` calls `toPixels`: https://github.com/flutter/packages/pull/7507/files#diff-26752280ec885e4465baf81a4771b5a7e6fd1041d333c9fedd712bcc5c3916c7L568.  
The `zoomBy` case previously called `toPoint`: https://github.com/flutter/packages/pull/7507/files#diff-26752280ec885e4465baf81a4771b5a7e6fd1041d333c9fedd712bcc5c3916c7L363.  
The new `zoomBy` case calls `pointFromPigeon`: https://github.com/flutter/packages/pull/7507/files#diff-26752280ec885e4465baf81a4771b5a7e6fd1041d333c9fedd712bcc5c3916c7L363.

reidbaker: I am mistaken, resolving.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
yaakovschectman: As far as I can tell, Pigeon will convert positional arguments to named arguments in the output file, anyway. I tried making it positional in `messages.dart`, but after running Pigeon, the generated code does not change. Do you think it would still be useful for this file to use a positional argument?

stuartmorgan: Oh, for some I thought we preserved the argument structure on the Dart side. In that case it doesn't matter.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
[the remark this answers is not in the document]
yaakovschectman: @stuartmorgan Do you have any remaining concerns?

stuartmorgan (approved):
LGTM

PR #7511, opened 2024-08-26
Remove matan from codeowners

matanlurey:
Now that the patches have landed there isn't a lot of value in me getting notified about every change to each package. I'm of course still happy to help, but I'll have to be reached in a different way than the `CODEOWNERS` mechanism :)


jonahwilliams (approved):
RSLGTM

PR #7529, opened 2024-08-28
[pigeon] allow gen of unused classes

tarrinneal:
changes pigeon_lib to allow generation of classes that aren't referenced in any API.

### DIFF
--- a/packages/pigeon/pigeons/core_tests.dart
+++ b/packages/pigeon/pigeons/core_tests.dart
@@ -25,16 +25,6 @@ class UnusedClass {
   Object? aField;
 }
 
-class SimpleClass {
-  SimpleClass({
-    this.aString,
-    this.aBool = true,
-  });
-
-  String? aString;
-  bool aBool;
-}
-
 /// A class containing all supported types.
 class AllTypes {
   AllTypes({
--- a/packages/pigeon/pigeons/core_tests.dart
+++ b/packages/pigeon/pigeons/core_tests.dart
@@ -18,6 +18,13 @@ enum AnotherEnum {
   justInCase,
 }
 
+// This exists to show that unused data classes still generate.
+class UnusedClass {
+  UnusedClass({this.aField});
+
+  Object? aField;
+}
+
 class SimpleClass {
   SimpleClass({
     this.aString,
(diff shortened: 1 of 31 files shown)


yaakovschectman:
Will all classes now be converted, or do they need to be marked in some way?

tarrinneal:
> Will all classes now be converted, or do they need to be marked in some way?

All classes in the pigeon definition file will be generated.

packages/pigeon/pigeons/core_tests.dart
```diff
@@ -18,6 +18,13 @@ enum AnotherEnum {
...
 }
 
+// This exists to show that unused data classes still generate.
+class UnusedClass {
+  UnusedClass({this.aField});
+
+  Object? aField;
+}
+
 class SimpleClass {
```
stuartmorgan, line 28: Based on the output diffs, it looks like this class wasn't being generated before. That suggests that it should be unused, and thus removed.

stuartmorgan (commented):
Per offline discussion, it would be good to add native unit tests that just instantiate the unused class, so that we have confirmation that it's being generated for each language.

packages/pigeon/pigeons/core_tests.dart
reidbaker: Or it should be used to validate that unused classes are generated? https://github.com/flutter/packages/pull/7529#pullrequestreview-2267160848

reidbaker (approved):

packages/pigeon/platform_tests/alternate_language_test_plugin/android/src/main/java/com/example/alternate_language_test_plugin/AlternateLanguageTestPlugin.java
```diff
@@ -688,6 +688,10 @@ public void error(Throwable error) {
         };
     flutterSmallApiOne.echoString(aString, resultCallbackOne);
   }
+
+  public CoreTests.UnusedClass testIfUnusedClassIsGenerated() {
```
stuartmorgan, line 692: Looks like the linter wants a `@NonNull` on this.

stuartmorgan (approved):
LGTM, thanks!

PR #7547, opened 2024-08-29
[pigeon] adds support for non nullable types in collections

tarrinneal:
[pigeon] adds support for non nullable types in collections

fixes https://github.com/flutter/flutter/issues/97848

### DIFF
--- a/packages/pigeon/CHANGELOG.md
+++ b/packages/pigeon/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 22.4.0
+
+* Adds support for nullable types in collections.
+
 ## 22.3.0
 
 * Adds support for enums and classes in collections.
--- a/packages/pigeon/CHANGELOG.md
+++ b/packages/pigeon/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 22.4.0
 
-* Adds support for nullable types in collections.
+* Adds support for non-nullable types in collections.
 
 ## 22.3.0
 
--- a/packages/pigeon/lib/objc_generator.dart
+++ b/packages/pigeon/lib/objc_generator.dart
@@ -297,7 +297,7 @@ class ObjcHeaderGenerator extends StructuredGenerator<ObjcOptions> {
       final _ObjcType returnType = _objcTypeForDartType(
         generatorOptions.prefix, func.returnType,
         // Nullability is required since the return must be nil if NSError is set.
-        forceNullability: true,
+        forceBox: true,
       );
       final String callbackType =
           _callbackForType(func.returnType, returnType, generatorOptions);
@@ -336,7 +336,7 @@ class ObjcHeaderGenerator extends StructuredGenerator<ObjcOptions> {
         func.returnType,
         // Nullability is required since the return must be nil if NSError is
         // set.
-        forceNullability: true,
+        forceBox: true,
       );
 
       String? lastArgName;
@@ -1072,7 +1072,7 @@ static FlutterError *createConnectionError(NSString *channelName) {
       final _ObjcType returnType = _objcTypeForDartType(
         generatorOptions.prefix, func.returnType,
         // Nullability is required since the return must be nil if NSError is set.
-        forceNullability: true,
+        forceBox: true,
       );
       final Iterable<String> selectorComponents =
           _getSelectorComponents(func, lastSelectorComponent);
@@ -1187,7 +1187,7 @@ void _writeMethod(
     languageOptions.prefix,
     func.returnType,
     // Nullability is required since the return must be nil if NSError is set.
-    forceNullability: true,
+    forceBox: true,
   );
   final String callbackType =
       _callbackForType(func.returnType, returnType, languageOptions);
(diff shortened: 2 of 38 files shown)


packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 22.4.0
+
+* Adds support for nullable types in collections.
```
stuartmorgan, line 3: non-nullable

packages/pigeon/lib/objc_generator.dart
```diff
@@ -1444,7 +1444,8 @@ String _flattenTypeArguments(String? classPrefix, List<TypeDeclaration> args) {
       return _enumName(e.baseName,
           prefix: classPrefix, box: true, suffix: ' *');
     }
-    return _objcTypeForDartType(classPrefix, e).toString();
+    return _objcTypeForDartType(classPrefix, e, forceNullability: true)
```
stuartmorgan, line 1447: Can you put a comment in the code about why `forceNullability` is true here? (I assume the answer is because "nullable" corresponds to "boxed", and a collection can only have boxed types, but if so that's non-obvious because of the naming so warrants a comment. And if not I don't understand why, so warrants a comment 🙂 )

packages/pigeon/platform_tests/shared_test_plugin_code/lib/test_types.dart
```diff
@@ -54,20 +54,37 @@ void compareAllNullableTypesWithoutRecursion(
       true);
   expect(listEquals(allNullableTypesOne.enumList, allNullableTypesTwo.enumList),
       true);
+  // expect(
+  //     listEquals(
+  //         allNullableTypesOne.objectList, allNullableTypesTwo.objectList),
+  //     true);
```
stuartmorgan, line 60: Why are all of these new tests commented out?

packages/pigeon/platform_tests/shared_test_plugin_code/lib/test_types.dart
```diff
@@ -183,40 +233,141 @@ void compareAllClassesWrapper(
...
+  //   compareAllTypes(
+  //       allTypesOne.classList[i], allTypesTwo.classList[i]);
+  // }
+  //
+  // classMap: check each key and value, call each compare method for every class
+  //
+  // for (int i = 0; i < (allTypesOne.nullableClassList.length); i++) {
+  //   compareAllNullableTypesWithoutRecursion(
+  //       allTypesOne.nullableClassList[i], allTypesTwo.nullableClassList[i]);
+  // }
```
stuartmorgan, line 246: We definitely shouldn't be replacing existing tests with commented out tests.

packages/pigeon/platform_tests/shared_test_plugin_code/lib/test_types.dart
```diff
@@ -416,24 +567,23 @@ final AllTypes genericAllTypes = AllTypes(
...
+  // mapList: nonNullMapList,
+  map: nonNullMap,
+  stringMap: nonNullStringMap,
+  intMap: nonNullIntMap,
+  // doubleMap: nonNullDoubleMap,
+  // boolMap: nonNullBoolMap,
+  enumMap: nonNullEnumMap,
+  objectMap: nonNullMap,
+  // listMap: nonNullListMap,
+  // mapMap: nonNullMapMap,
```
stuartmorgan, line 586: I either missed this in the review where they were added, or have forgotten; why are there commented out field here and below, vs them just not being here?

(https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-abandonware)

stuartmorgan (changes requested):

packages/pigeon/platform_tests/shared_test_plugin_code/lib/test_types.dart
tarrinneal: Just things I missed when cleaning up the pr aka future work plans

packages/pigeon/platform_tests/shared_test_plugin_code/lib/test_types.dart
tarrinneal: bad merge :(

packages/pigeon/platform_tests/shared_test_plugin_code/lib/test_types.dart
```diff
@@ -54,12 +54,6 @@ void compareAllNullableTypesWithoutRecursion(
       true);
   expect(listEquals(allNullableTypesOne.enumList, allNullableTypesTwo.enumList),
       true);
-  // TODO(stuartmorgan): Enable this once the Dart types are fixed; see
-  // https://github.com/flutter/flutter/issues/116117
-  //for (int i = 0; i < allNullableTypesOne.listList!.length; i++) {
-  //  expect(listEquals(allNullableTypesOne.listList![i], allNullableTypesTwo.listList![i]),
-  //      true);
-  //}
```
stuartmorgan, line 62: \o/

stuartmorgan (approved):
LGTM

packages/pigeon/platform_tests/shared_test_plugin_code/lib/test_types.dart
tarrinneal: It's still not fixed, but the next pr will re-add a version of this. Then I'll add a more robust test in the final sequel.

stuartmorgan: 🤦🏻 Somehow I mis-parsed this as enabling the test rather than removing it.

> It's still not fixed, but the next pr will re-add a version of this.

👍🏻

PR #7108, opened 2024-07-12
[camera] add video stabilization

ruicraveiro:
Adds support for video stabilization to camera_platform_interface, camera_avfoundation, camera_android_camerax and camera packages.

The video stabilization modes are defined in the new VideoStabilizationMode enum defined in camera_platform_interface:
```dart
/// The possible video stabilization modes that can be capturing video.
enum VideoStabilizationMode {
  /// Video stabilization is disabled.
  off,

  /// Basic video stabilization is enabled.
  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_ON on Android
  /// and throws CameraException on iOS.
  on,

  /// Standard video stabilization is enabled.
  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION on Android
  /// (camera_android_camerax) and to AVCaptureVideoStabilizationModeStandard
  /// on iOS.
  standard,

  /// Cinematic video stabilization is enabled.
  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION on Android
  /// (camera_android_camerax) and to AVCaptureVideoStabilizationModeCinematic
  /// on iOS.
  cinematic,

  /// Extended cinematic video stabilization is enabled.
  /// Maps to AVCaptureVideoStabilizationModeCinematicExtended on iOS and
  /// throws CameraException on Android.
  cinematicExtended,
}
```
There is some subjectivity on the way with which I mapped the modes to both platforms, and here's a document that compares the several modes: https://docs.google.com/spreadsheets/d/1TLOLZHR5AcyPlr-y75aN-DbR0ssZLJjpV_OAJkRC1FI/edit?usp=sharing, which you can comment on.

*List which issues are fixed by this PR. You must list at least one issue.*
Partially implements https://github.com/flutter/flutter/issues/89525

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
[...]

PR #7548, opened 2024-08-30
Adds support for video stabilization to camera_platform_interface

ruicraveiro:
Adds support for video stabilization to camera_platform_interface.

This is the platform interface package PR derived from the combined PR https://github.com/flutter/packages/pull/7108.

The video stabilization modes are defined in the new VideoStabilizationMode enum defined in camera_platform_interface:
```dart
/// The possible video stabilization modes that can be capturing video.
enum VideoStabilizationMode {
  /// Video stabilization is disabled.
  off,

  /// Basic video stabilization is enabled.
  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_ON on Android
  /// and throws CameraException on iOS.
  on,

  /// Standard video stabilization is enabled.
  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION on Android
  /// (camera_android_camerax) and to AVCaptureVideoStabilizationModeStandard
  /// on iOS.
  standard,

  /// Cinematic video stabilization is enabled.
  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION on Android
  /// (camera_android_camerax) and to AVCaptureVideoStabilizationModeCinematic
  /// on iOS.
  cinematic,

  /// Extended cinematic video stabilization is enabled.
  /// Maps to AVCaptureVideoStabilizationModeCinematicExtended on iOS and
  /// throws CameraException on Android.
  cinematicExtended,
}
```

Partially implements https://github.com/flutter/flutter/issues/89525

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
+++ b/packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
@@ -273,10 +273,6 @@ abstract class CameraPlatform extends PlatformInterface {
   }
 
   /// Gets a list of video stabilization modes that are supported for the selected camera.
-  ///
-  /// Will return the list of supported video stabilization modes
-  /// on Android (when using camera_android_camerax package) and
-  /// on iOS. Will return an empty list on all other platforms.
   Future<Iterable<VideoStabilizationMode>> getVideoStabilizationSupportedModes(
       int cameraId) async {
     throw UnimplementedError(
@@ -284,10 +280,6 @@ abstract class CameraPlatform extends PlatformInterface {
   }
 
   /// Sets the video stabilization mode for the selected camera.
-  ///
-  /// On Android (when using camera_android_camerax) and on iOS
-  /// the supplied [mode] value should be a mode in the list returned
-  /// by [getVideoStabilizationSupportedModes].
   ///
   /// Throws a [CameraException] when a not supported video stabilization
   /// mode is supplied.
--- a/packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
+++ b/packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
@@ -272,6 +272,30 @@ abstract class CameraPlatform extends PlatformInterface {
     throw UnimplementedError('setZoomLevel() is not implemented.');
   }
 
+  /// Gets a list of video stabilization modes that are supported for the selected camera.
+  ///
+  /// Will return the list of supported video stabilization modes
+  /// on Android (when using camera_android_camerax package) and
+  /// on iOS. Will return an empty list on all other platforms.
+  Future<Iterable<VideoStabilizationMode>> getVideoStabilizationSupportedModes(
+      int cameraId) async {
+    throw UnimplementedError(
+        'getVideoStabilizationSupportedModes() is not implemented.');
+  }
+
+  /// Sets the video stabilization mode for the selected camera.
+  ///
+  /// On Android (when using camera_android_camerax) and on iOS
+  /// the supplied [mode] value should be a mode in the list returned
+  /// by [getVideoStabilizationSupportedModes].
+  ///
+  /// Throws a [CameraException] when a not supported video stabilization
+  /// mode is supplied.
+  Future<void> setVideoStabilizationMode(
+      int cameraId, VideoStabilizationMode mode) async {
+    throw UnimplementedError('setVideoStabilizationMode() is not implemented.');
+  }
+
   /// Pause the active preview on the current frame for the selected camera.
   Future<void> pausePreview(int cameraId) {
     throw UnimplementedError('pausePreview() is not implemented.');
--- a/packages/camera/camera_platform_interface/lib/src/types/video_stabilization_mode.dart
+++ b/packages/camera/camera_platform_interface/lib/src/types/video_stabilization_mode.dart
@@ -1,0 +1,41 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// The possible video stabilization modes that can be capturing video.
+enum VideoStabilizationMode {
+  /// Video stabilization is disabled.
+  off,
+
+  /// Basic video stabilization is enabled.
+  ///
+  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_ON on Android
+  /// and throws CameraException on iOS.
+  on,
+
+  /// Standard video stabilization is enabled.
+  ///
+  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION on Android
+  /// (camera_android_camerax) and to AVCaptureVideoStabilizationModeStandard
+  /// on iOS.
+  standard,
+
+  /// Cinematic video stabilization is enabled.
+  ///
+  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION on Android
+  /// (camera_android_camerax) and to AVCaptureVideoStabilizationModeCinematic
+  /// on iOS.
+  cinematic,
+
+  /// Extended cinematic video stabilization is enabled.
+  ///
+  /// Maps to AVCaptureVideoStabilizationModeCinematicExtended on iOS and
+  /// throws CameraException on Android.
+  cinematicExtended;
+
+  /// Returns the video stabilization mode for a given String.
+  factory VideoStabilizationMode.fromString(String str) {
+    return VideoStabilizationMode.values
+        .firstWhere((VideoStabilizationMode mode) => mode.name == str);
+  }
+}
--- a/packages/camera/camera_platform_interface/AUTHORS
+++ b/packages/camera/camera_platform_interface/AUTHORS
@@ -65,3 +65,4 @@ Anton Borries <mail@antonborri.es>
 Alex Li <google@alexv525.com>
 Rahul Raj <64.rahulraj@gmail.com>
 Mairramer <mairramer.dasilva28@hotmail.com>
+Rui Craveiro <ruicraveiro@squarealfa.com>
--- a/packages/camera/camera_platform_interface/CHANGELOG.md
+++ b/packages/camera/camera_platform_interface/CHANGELOG.md
@@ -1,5 +1,6 @@
-## NEXT
+## 2.9.0
 
+* Adds support for video stabilization.
 * Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
 
 ## 2.8.0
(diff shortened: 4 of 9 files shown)


packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
```diff
@@ -272,6 +272,30 @@ abstract class CameraPlatform extends PlatformInterface {
     throw UnimplementedError('setZoomLevel() is not implemented.');
   }
 
+  /// Gets a list of video stabilization modes that are supported for the selected camera.
+  ///
+  /// Will return the list of supported video stabilization modes
+  /// on Android (when using camera_android_camerax package) and
+  /// on iOS. Will return an empty list on all other platforms.
```
stuartmorgan, line 279: Platform interfaces should not make claims about the behavior of specific platforms.

packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
```diff
@@ -272,6 +272,30 @@ abstract class CameraPlatform extends PlatformInterface {
...
+  Future<Iterable<VideoStabilizationMode>> getVideoStabilizationSupportedModes(
+      int cameraId) async {
+    throw UnimplementedError(
+        'getVideoStabilizationSupportedModes() is not implemented.');
+  }
+
+  /// Sets the video stabilization mode for the selected camera.
+  ///
+  /// On Android (when using camera_android_camerax) and on iOS
```
stuartmorgan, line 288: Same.

ruicraveiro: For the previous comment, I simply removed the paragraph. On this one, however, I just removed:

```dart
  /// On Android (when using camera_android_camerax) and on iOS
  /// the supplied [mode] value should be a mode in the list returned
  /// by [getVideoStabilizationSupportedModes].
```

But now I am in doubt about this one:
```dart
  ///
  /// Throws a [CameraException] when a not supported video stabilization
  /// mode is supplied.
```
It is assuming the behaviour expected by the specific platforms, but it can't really guarantee that they will behave as expected. Should this comment stay or shall I remove it?

stuartmorgan: > But now I am in doubt about this one:
> 
> ```dart
>   ///
>   /// Throws a [CameraException] when a not supported video stabilization
>   /// mode is supplied.
> ```
> 
> It is assuming the behaviour expected by the specific platforms, but it can't really guarantee that they will behave as expected. Should this comment stay or shall I remove it?

There is a difference between an abstract interface's comment describing what an implementation of a method is required to do, and the comment making assertions about what the current implementations do.

The simple test in most cases is to ask: is this comment intended to be prescriptive about what a conforming implementation must do? If so, it makes sense; if not, it probably doesn't.

For instance:
> Will return an empty list on all other platforms.

Would it be a bug if someone implemented stabilization on Windows? Clearly not, which means this comment doesn't belong here.

> Throws a [CameraException] when a not supported video stabilization mode is supplied.

Would it be a bug if one implementation had divergent handling of invalid usage as compared to other platforms? Yes, because having predicable results to programming errors is a feature, so this does belong.

packages/camera/camera_platform_interface/lib/src/types/video_stabilization_mode.dart
```diff
@@ -0,0 +1,41 @@
...
+  ///
+  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION on Android
+  /// (camera_android_camerax) and to AVCaptureVideoStabilizationModeCinematic
+  /// on iOS.
+  cinematic,
+
+  /// Extended cinematic video stabilization is enabled.
+  ///
+  /// Maps to AVCaptureVideoStabilizationModeCinematicExtended on iOS and
+  /// throws CameraException on Android.
```
stuartmorgan, line 33: If next year, Android gains the ability to support extended cinematic stabilization, would adding support for it be a bug?

packages/camera/camera_platform_interface/lib/src/types/video_stabilization_mode.dart
```diff
@@ -0,0 +1,41 @@
...
+  standard,
+
+  /// Cinematic video stabilization is enabled.
+  ///
+  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION on Android
+  /// (camera_android_camerax) and to AVCaptureVideoStabilizationModeCinematic
+  /// on iOS.
+  cinematic,
+
+  /// Extended cinematic video stabilization is enabled.
```
stuartmorgan, line 30: This seems extremely closely tied to the specific iOS implementation. It would be much better to give a more generic description of what this is (which based on the iOS docs seems to be a better but slower/more expensive stabilization algorithm), so that it is clearer what it would mean for another platform to support this.

packages/camera/camera_platform_interface/lib/src/types/video_stabilization_mode.dart
```diff
@@ -0,0 +1,41 @@
...
+  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_ON on Android
+  /// and throws CameraException on iOS.
+  on,
+
+  /// Standard video stabilization is enabled.
+  ///
+  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION on Android
+  /// (camera_android_camerax) and to AVCaptureVideoStabilizationModeStandard
+  /// on iOS.
+  standard,
```
stuartmorgan, line 21: It seems confusing to me that `on` and `standard` are different things. If there's a generic `on` mode, isn't that by definition the standard mode?

packages/camera/camera_platform_interface/lib/src/types/video_stabilization_mode.dart
```diff
@@ -0,0 +1,41 @@
...
+
+/// The possible video stabilization modes that can be capturing video.
+enum VideoStabilizationMode {
+  /// Video stabilization is disabled.
+  off,
+
+  /// Basic video stabilization is enabled.
+  ///
+  /// Maps to CONTROL_VIDEO_STABILIZATION_MODE_ON on Android
+  /// and throws CameraException on iOS.
```
stuartmorgan, line 13: Related to my comment about `standard`, it seems very surprising that `on` would throw on iOS since clearly stabilization can be on.

Can we make a more generic set of options that can be more easily mapped to various implementations? Having semi-disjoint options across two platforms is a sign that this is not generic enough, and would thus get even more confusing if a third implementation were added later.

packages/camera/camera_platform_interface/lib/src/types/video_stabilization_mode.dart
ruicraveiro: OK, that answers my question, so I will remove that comment.

packages/camera/camera_platform_interface/lib/src/types/video_stabilization_mode.dart
ruicraveiro: I wonder if a first approach to solving the mapping issue doesn't start by doing something as simple as cleaning up the comments. For instance:

```dart
  /// Extended cinematic video stabilization is enabled.
  ///
  /// Maps to AVCaptureVideoStabilizationModeCinematicExtended on iOS and
  /// throws CameraException on Android.
  cinematicExtended;
```
Saying that it throws on Android is kind of redundant, because until the day it gets implemented on Android, it will never be included in the list returned by ```getVideoStabilizationSupportedModes()```. It's just that the comment is hinting the user on what the current implementation is doing. So, an easy thing to do would be to change it to:

```dart
  /// Extended cinematic video stabilization is enabled.
  ///
  /// Maps to AVCaptureVideoStabilizationModeCinematicExtended on iOS.
  cinematicExtended;
```

Next, does it make sense to keep the "Maps to AVCaptureVideoStabilizationModeCinematicExtended on iOS"? On one hand, it doesn't because it will be incomplete if other platforms get an implementation. It might mislead the user into thinking that only iOS still has an implementation even when it ceases to be exclusive to iOS. On the other hand, it is very convenient for the user to know how it maps to iOS just by looking at the enum (which is something that I do a lot with ResolutionPreset and to a lesser degree with ImageFormatGroup. So, I don't really know what the best approach would be.

stuartmorgan: > The 'Standard' mode maps to what I considered to be the standard video stabilization on each platform, so I mapped it to CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION and AVCaptureVideoStabilizationModeStandard.

Can you elaborate on why preview stabilization mode is what you would consider "standard" on Android, when from the enum naming it seems like the developers of the Android API would consider "on" to be the standard mode? Is it about matching iOS behaviors?

In practice, what is the behavioral difference between `CONTROL_VIDEO_STABILIZATION_MODE_ON ` and `AVCaptureVideoStabilizationModeStandard`, and between `CONTROL_VIDEO_STABILIZATION_MODE_PREVIEW_STABILIZATION` and `AVCaptureVideoStabilizationModeStandard`?

Part of the issue here I think is that the Android constant names communicate (albeit not very clearly) a behavioral difference, whereas the names you've chosen to map those two sound indistinguishable to me, so as a client I would have no idea what the difference between `on` and `standard` is. If the difference between them is *only* whether or not the preview is stabilized, then the names should reflect that, rather than hide it. (If that distinction doesn't exist on iOS it's okay to have a constant that doesn't exist on iOS, but it doesn't make sense for something called `on` not to exist on iOS.)

> So, yes, I am open to making changes, but it would be helpful if you could suggest the concrete mapping, including descriptions

I don't know what a better mapping is without more information; the spreadsheet contains the docs information, but that doesn't answer important questions about how they actually compare. E.g., I don't know from the spreadsheet what the preview behavior is on iOS.

What I do know is that I find the current names *actively confusing* as a naive client of the API, which is a red flag.

stuartmorgan: > Next, does it make sense to keep the "Maps to AVCaptureVideoStabilizationModeCinematicExtended on iOS"? On one hand, it doesn't because it will be incomplete if other platforms get an implementation. It might mislead the user into thinking that only iOS still has an implementation even when it ceases to be exclusive to iOS. On the other hand, it is very convenient for the user to know how it maps to iOS just by looking at the enum (which is something that I do a lot with ResolutionPreset and to a lesser degree with ImageFormatGroup. So, I don't really know what the best approach would be.

The approach we've taken in newer code is to frame the mapping as examples. It's still not ideal in that in theory the platform implementations could diverge over time because it's a different package, but we balance that against the value of people being able to see concretely what things will mean in practice. So it's fine to say something like:

> /// For example, this corresponds to:
> /// - AVFoo on iOS
> /// - CONTROL_VIDEO_BaR on Android

PR #7558, opened 2024-08-30
[google_maps_flutter_android] Convert `JointType` to an enum

yaakovschectman:
Replace the raw integers used in `PlatformPolyline` to an enum.

https://github.com/flutter/flutter/issues/153718#event-13934571301

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
@@ -1,4 +1,4 @@
-## NEXT
+## 2.14.5
 
 * Converts `JointType` to enum.
 
--- a/packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Converts `JointType` to enum.
+
 ## 2.14.4
 
 * Converts 'PlatformTileOverlay' to pigeon.
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
@@ -29,6 +29,7 @@ import com.google.android.gms.maps.model.CustomCap;
 import com.google.android.gms.maps.model.Dash;
 import com.google.android.gms.maps.model.Dot;
 import com.google.android.gms.maps.model.Gap;
+import com.google.android.gms.maps.model.JointType;
 import com.google.android.gms.maps.model.LatLng;
 import com.google.android.gms.maps.model.LatLngBounds;
 import com.google.android.gms.maps.model.PatternItem;
@@ -706,13 +707,13 @@ class Convert {
   static int jointTypeFromPigeon(Messages.PlatformJointType jointType) {
     switch (jointType) {
       case MITERED:
-        return 0;
+        return JointType.DEFAULT;
       case BEVEL:
-        return 1;
+        return JointType.BEVEL;
       case ROUND:
-        return 2;
+        return JointType.ROUND;
     }
-    return 0;
+    return JointType.DEFAULT;
   }
 
   static String interpretPolylineOptions(
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
@@ -712,7 +712,6 @@ class Convert {
       case ROUND:
         return 2;
     }
-    // These integer values come from a different package. This fallback ensures some value is returned in case a new enum value is added.
     return 0;
   }
 
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
@@ -703,6 +703,19 @@ class Convert {
     return polygon.getPolygonId();
   }
 
+  static int jointTypeFromPigeon(Messages.PlatformJointType jointType) {
+    switch (jointType) {
+      case MITERED:
+        return 0;
+      case BEVEL:
+        return 1;
+      case ROUND:
+        return 2;
+    }
+    // These integer values come from a different package. This fallback ensures some value is returned in case a new enum value is added.
+    return 0;
+  }
+
   static String interpretPolylineOptions(
       Messages.PlatformPolyline polyline,
       PolylineOptionsSink sink,
@@ -713,7 +726,7 @@ class Convert {
     sink.setEndCap(toCap(polyline.getEndCap(), assetManager, density));
     sink.setStartCap(toCap(polyline.getStartCap(), assetManager, density));
     sink.setGeodesic(polyline.getGeodesic());
-    sink.setJointType(polyline.getJointType().index);
+    sink.setJointType(jointTypeFromPigeon(polyline.getJointType()));
     sink.setVisible(polyline.getVisible());
     sink.setWidth(polyline.getWidth());
     sink.setZIndex(polyline.getZIndex());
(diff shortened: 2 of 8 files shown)


packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -713,7 +713,7 @@ static String interpretPolylineOptions(
     sink.setEndCap(toCap(polyline.getEndCap(), assetManager, density));
     sink.setStartCap(toCap(polyline.getStartCap(), assetManager, density));
     sink.setGeodesic(polyline.getGeodesic());
-    sink.setJointType(polyline.getJointType().intValue());
+    sink.setJointType(polyline.getJointType().index);
```
stuartmorgan, line 716: This isn't fixing the problem of implicitly relying on numeric values from two different sources lining up, it just changes exactly what the things that have to line up are; it would now silently break if the enum order were changed in the Pigeon file, or if Android changed the constants.

There should be a helper method that `switch`es over the enum and returns the SDK constants.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -1146,6 +1146,20 @@ PlatformZoomRange? _platformZoomRangeFromMinMaxZoomPreferenceJson(
...
+PlatformJointType platformJointTypeFromJointType(
+    JointType jointType) {
+  switch (jointType) {
+    case JointType.mitered:
+      return PlatformJointType.mitered;
+    case JointType.bevel:
+      return PlatformJointType.bevel;
+    case JointType.round:
+      return PlatformJointType.round;
+  }
```
stuartmorgan, line 1160: This needs a fallback to prevent undefined behavior if the platform interface gets a new value; see https://github.com/flutter/packages/blob/0e8c5da679c8c812ee6af457a7768a6e828e6308/packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart#L992-L998

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -181,6 +181,13 @@ class PlatformPolygon {
   final int zIndex;
 }
 
+/// Join types for polyline joints. The indices of this enum must match those used internally by the maps SDK.
```
stuartmorgan, line 184: These comments should be updated once my comment about this being problematic is addressed.

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -203,10 +210,8 @@ class PlatformPolyline {
   final int color;
   final bool geodesic;
 
-  /// The joint type as an integer. This must be a value corresponding to one of the values defined in the platform interface package's JointType enum. The integer values specified in this enum must match those used by the native SDK.
-  // TODO(schectman): Convert field to enum.
-  // https://github.com/flutter/flutter/issues/153718
-  final int jointType;
+  /// The joint type. The integer indices of this enum must match those used by the native SDK.
```
stuartmorgan, line 213: This is no longer correct.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -703,6 +703,19 @@ static String interpretPolygonOptions(Messages.PlatformPolygon polygon, PolygonO
...
+  static int jointTypeFromPigeon(Messages.PlatformJointType jointType) {
+    switch (jointType) {
+      case MITERED:
+        return 0;
+      case BEVEL:
+        return 1;
+      case ROUND:
+        return 2;
+    }
+    // These integer values come from a different package. This fallback ensures some value is returned in case a new enum value is added.
```
stuartmorgan, line 715: This is not correct at this layer; `PlatformJointType` is from this package. If there's a case defined in `messages.dart` but not handled above that's a programming error, not a plausible runtime error.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
stuartmorgan: This has not been addressed; if a value is added to `JointType` in the platform interface package, someone running the above version of this package's code with the new platform interface version would have undefined behavior here.

stuartmorgan (changes requested):

yaakovschectman:
@stuartmorgan I am unclear on what we should do about this. We could add a `default` clause to the switch statement, but this would produce a warning for the time being as it is already exhaustive.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
yaakovschectman: I see. I will remove the comment, but I believe we still need the return statement for Java compilation.

stuartmorgan:
Are you talking about the Dart side? I linked to an example of our standard pattern for this, which includes the relevant `ignore` directive.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -1146,6 +1146,22 @@ PlatformZoomRange? _platformZoomRangeFromMinMaxZoomPreferenceJson(
...
+PlatformJointType platformJointTypeFromJointType(JointType jointType) {
+  switch (jointType) {
+    case JointType.mitered:
+      return PlatformJointType.mitered;
+    case JointType.bevel:
+      return PlatformJointType.bevel;
+    case JointType.round:
+      return PlatformJointType.round;
+  }
+  // The JointType enum is defined in the platform interface package. A fallback return value here ensures the android package will not break if a new enum value is added.
```
stuartmorgan, line 1160: Please wrap comments to the same line length as the code.

Although I would prefer if this used the exact text that I linked to. We've used it in many other places, so it would be more consistent, and it provides more detail about the reason for the specific structure so that someone doesn't come along and think it would be cleaner to fold this into a `default` case.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -703,6 +703,18 @@ static String interpretPolygonOptions(Messages.PlatformPolygon polygon, PolygonO
     return polygon.getPolygonId();
   }
 
+  static int jointTypeFromPigeon(Messages.PlatformJointType jointType) {
+    switch (jointType) {
+      case MITERED:
+        return 0;
```
stuartmorgan, line 709: The Google Maps SDK provides named constants for these values.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## NEXT
```
stuartmorgan, line 1: This will need a version bump.

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
yaakovschectman: We will need a version bump before this lands. I was thinking it would make sense to do so when we are closer to landing due to this being a federated change that must be split into multiple steps.

stuartmorgan: > due to this being a federated change that must be split into multiple steps.

I don't follow. Are there other changes that aren't uploaded yet? I would not expect this to need changes to other packages.

yaakovschectman: I confused this with a different and rather similar PR. Never mind that.

stuartmorgan (approved):
LGTM

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -1146,6 +1146,26 @@ PlatformZoomRange? _platformZoomRangeFromMinMaxZoomPreferenceJson(
...
+    case JointType.bevel:
+      return PlatformJointType.bevel;
+    case JointType.round:
+      return PlatformJointType.round;
+  }
+  // The enum comes from a different package, which could get a new value at
+  // any time, so provide a fallback that ensures this won't break when used
+  // with a version that contains new values. This is deliberately outside
+  // the switch rather than a `default` so that the linter will flag the
+  // switch as needing an update.
```
reidbaker, line 1164: This comment is helpful.

reidbaker (approved):

PR #7586, opened 2024-09-05
[local_auth_darwin] - Converts implementation to Swift

Mairramer:
Part of https://github.com/flutter/flutter/issues/119015
Closes https://github.com/flutter/flutter/issues/119104

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Thanks for the contribution!

Could you split whatever substantive changes you've made to the native unit test file into a precursor PR? The diffs are enough that combining it with a rename means that we can't see what changes were made to the test file, and having both the test file and the implementation file showing as completely rewritten at the same time makes this much harder to review and higher risk.

Mairramer:
> 
> Could you split whatever substantive changes you've made to the native unit test file into a precursor PR? The diffs are enough that combining it with a rename means that we can't see what changes were made to the test file, and having both the test file and the implementation file showing as completely rewritten at the same time makes this much harder to review, and higher risk.

I can try to do that.

Mairramer:
@stuartmorgan I made every effort to minimize behavioral changes between the Objective-C and Swift implementations, anticipating the need for a precursor PR. Despite the numerous changes, which will require additional review time, I believe this approach will mitigate risk. I aimed to keep the tests and implementations closely aligned.

stuartmorgan:
> I made every effort to minimize behavioral changes between the Objective-C and Swift implementations

Could you elaborate on what you mean by "minimize"? I would not expect a language conversion to have *any* behavioral changes.

> Despite the numerous changes, which will require additional review time, I believe this approach will mitigate risk.

Changing tests and implementation at the same time, when the implementation changes are not supposed to change behavior, maximizes risk. And renaming a file at the same time as it is changed sufficient to break GitHub diff detection, such that it's impossible for a reviewer to see what is changed without an error-prone manual line-by-line analysis, wastes a lot of reviewer time in addition to maximizing risk there.

What is preventing a precursor change that prevents the situation where the diff shown in the review system for the 500-line test file is unusable?

Mairramer:
> Could you elaborate on what you mean by "minimize"? I would not expect a language conversion to have _any_ behavioral changes.

When I say "minimize", I mean taking steps to ensure that the translation from Objective-C to Swift maintains the same functional behavior, that is, that both implementations behave identically under the same conditions. This includes maintaining the same input/output behavior, error handling, performance characteristics, and any side effects that may exist in the system. The goal is to avoid introducing unintended behavioral differences that could affect functionality even though the code is now written in a different language.




> Changing tests and implementation at the same time, when the implementation changes are not supposed to change behavior, maximizes risk. And renaming a file at the same time as it is changed sufficient to break GitHub diff detection, such that it's impossible for a reviewer to see what is changed without an error-prone manual line-by-line analysis, wastes a lot of reviewer time in addition to maximizing risk there.

You are right that changing tests and implementation simultaneously increases risk. When tests are changed along with the implementation, it is harder to isolate the cause of a failure, since the problem may be in the test or the implementation. 

Do you have any suggestions for how to go about these changes? Just renaming won't be enough, I think.

stuartmorgan:
> Do you have any suggestions for how to go about these changes?

As I said before:

>> Could you split whatever substantive changes you've made to the native unit test file into a precursor PR?

If you make test changes separately from anything else, then they can be reviewed as diffs, and then the implementation rewrite won't have any test changes.

stuartmorgan:
> then they can be reviewed as diffs

That would also allow for dedicated discussion of why the tests appear to be almost completely rewritten from scratch even though they were already written in Swift and had a whole DI system in place. It's not at all clear to me the tests need such extensive changes.

Mairramer:
@stuartmorgan Ok, I'll create a precursor PR. It'll take some time, but I'll do it.

stuartmorgan:
> It'll take some time, but I'll do it.

Could you elaborate on what the complexity is? What is the reason there are non-trivial changes to the test file in the first place?

Mairramer:
@stuartmorgan I renamed the test file to what it was before.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -15,547 +16,509 @@ import XCTest
...
+  var plugin: LocalAuthPlugin!
+  var mockAuthContext: MockAuthContext!
+  var mockAlertController: MockAlertController!
+
+  override func setUp() {
+    super.setUp()
+    mockAuthContext = MockAuthContext()
+    mockAlertController = MockAlertController()
+    plugin = LocalAuthPlugin(authContext: mockAuthContext, alertController: mockAlertController)
+  }
```
stuartmorgan, line 109: Please revert these changes; we strongly prefer not having stateful tests, even at the expense of slightly more boilerplate in individual tests. https://abseil.io/tips/122 is a good resource on this topic (written for C++, but generall applicable).

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -15,547 +16,509 @@ import XCTest
...
-    }
-  #endif
-}
-
-final class StubAuthContext: NSObject, FLADAuthContext {
-  /// Whether calls to this stub are expected to be for biometric authentication.
-  ///
-  /// While this object could be set up to return different values for different policies, in
-  /// practice only one policy is needed by any given test, so this just allows asserting that the
-  /// code is calling with the intended policy.
```
stuartmorgan, line 105: Why were all of the comments removed from this class?

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -15,547 +16,509 @@ import XCTest
...
-        useErrorDialogs: false),
+      options: AuthOptions(biometricOnly: true, sticky: false, useErrorDialogs: false),
       strings: strings
-    ) { resultDetails, error in
+    ) { resultDetails in
       XCTAssertTrue(Thread.isMainThread)
-      // TODO(stuartmorgan): Fix this; this was the pre-Pigeon-migration
-      // behavior, so is preserved as part of the migration, but a failed
-      // authentication should return failure, not an error that results in a
-      // PlatformException.
```
stuartmorgan, line 231: Why was this TODO removed?

stuartmorgan (commented):
Thanks; undoing the rename makes it much easier to see what's changed. I've left some initial comments about places where the churn could be reduced, but I see that some is because of the result type change with the language change. I'll review again in more detail later this week now that I can see all the dfifs.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
Mairramer: I removed it while adjusting part of the return in the tests. I put it back on.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
Mairramer: I actually forgot to add it back when I switched to the protocol.

jmagman:
@LouiseHsu friendly ping to review.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,254 @@
...
+    resultHandler: @escaping (Result<AuthResultDetails, Error>) -> Void
+  ) {
+    self.options = options
+    self.strings = strings
+    self.resultHandler = resultHandler
+  }
+}
+
+public final class LocalAuthPlugin: NSObject, FlutterPlugin, LocalAuthApi {
+  private var authContext: AuthContext
```
stuartmorgan, line 30: Why have you switched to a singe persistent auth context? `AuthContext` is stateful, so re-using it from call to call could have unintended consequences. Things should only be persistent when there's a reason we want to preserve state.

stuartmorgan (commented):
I've done an initial pass over the implementation. There are a number of logic changes from the Obj-C version for reasons that aren't clear to me; I've left specific comments inline.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,254 @@
...
+  init(
+    authContext: AuthContext = DefaultAuthContext(),
+    alertController: AlertController = DefaultAlertController()
+  ) {
+    self.authContext = authContext
+    self.alertController = alertController
+
+    super.init()
+    #if os(iOS)
+      NotificationCenter.default.addObserver(
```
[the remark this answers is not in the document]
Mairramer: I changed it because it's something I see more commonly in Swift, as it's a common pattern for handling application lifecycle events. But I can revert to the previous implementation.

packages/local_auth/local_auth_darwin/lib/local_auth_darwin.dart
```diff
@@ -137,6 +137,7 @@ class LocalAuthDarwin extends LocalAuthPlatform {
     return AuthStrings(
       reason: localizedReason,
       lockOut: messages?.lockOut ?? macOSLockOut,
+      goToSettingsButton: '',
```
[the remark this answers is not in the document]
Mairramer: If I remember correctly, it was because of a null safety issue in pigeon the first time I tried to migrate. But I've already fixed it.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
Mairramer: You are right in saying that. I made this change with the aim of reuse, but I forgot that if any "contamination" occurs, this can lead to subsequent errors.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,254 @@
...
+
+    switch authError.code {
+    case LAError.passcodeNotSet.rawValue, LAError.biometryNotEnrolled.rawValue:
+      if options.useErrorDialogs {
+        alertController.showAlert(
+          message: strings.goToSettingsDescription ?? "",
+          dismissTitle: strings.cancelButton,
+          openSettingsTitle: strings.goToSettingsButton
+        ) { success in
+          let result: Result<AuthResultDetails, Error> =
```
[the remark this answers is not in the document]
Mairramer: The protocol for showing dialogs was changed to handle authentication results using Result<AuthResultDetails, Error> instead of a simple Bool completion. And also because I changed the encapsulation of the logic (Default AlertController for iOS and macOS), instead of using a Factory.

packages/local_auth/local_auth_darwin/example/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -218,7 +218,6 @@
 				331C80D1294CF70F00263BE5 /* Sources */,
 				331C80D2294CF70F00263BE5 /* Frameworks */,
 				331C80D3294CF70F00263BE5 /* Resources */,
-				4A22C8F8819194F5DFFCC074 /* [CP] Embed Pods Frameworks */,
```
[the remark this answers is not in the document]
Mairramer: I think so, I don't remember doing that.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
Mairramer: For `AuthContext`, I implemented a closure-based provider pattern that offers a different but effective approach compared to the protocol + implementation used for AlertController.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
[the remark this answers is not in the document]
Mairramer: it has the same behavior as before, but using NotificationCenter.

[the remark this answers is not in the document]
stuartmorgan: I understand that you intend it to have the same behavior. What I'm asking is why you are making a change to the logic of the plugin as part of converting the implementation to Swift. I don't see any reason the previous logic would not be trivial to convert to Swift, so I don't understand why this *unrelated* change is being made in a PR that is about changing the language, not the actual implementation.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan: > I implemented a closure-based provider pattern that offers a different but effective approach

Why have you changed the approach? Was there an issue using the previous approach, in order to minimize unrelated changes?

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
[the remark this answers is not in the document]
stuartmorgan: > And also because I changed the encapsulation of the logic

You've moved significant logic into a class that is replaced in tests, which means all of that logic is now untested, reducing test coverage relative to the previous approach. There is a reason we make the injectable test replacements absolutely minimal wrappers, rather than high-level abstractions.

stuartmorgan:
As a high-level comment on my specific comments above: as discussed earlier, the goal of a Swift conversion PR should be to convert the *existing* logic to Swift as directly as possible, modulo any language-specific idiomatic differences. A conversion is already a relatively high-risk change due to the magnitude of code being changed; intentionally mixing in refactors, subjective changes to implementation approach, etc. is highly undesirable in such a PR because it increases the risk even futher.

Any delta between the Obj-C logic and the Swift logic in this PR should be clearly and directly attributable to a language-specific reason.

If you have subjective disagreements about how the code is currently structured, please feel free to either make prequel PRs or follow-up PRs that make those changes, so they can be discussed, reviewed, and tested in isolation, rather than as part of a complete rewrite of the code.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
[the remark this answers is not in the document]
Mairramer: I'm just a bit confused about how to work with the alerts while maintaining what we had before. I’ll push some changes later today or tomorrow with these updates.

Mairramer:
@stuartmorgan I made some adjustments and I'm closer to the Objective-C implementation. There are still a few things left to be done, and if you think it's appropriate, I can mark this PR as a draft and continue working on the remaining details

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/AlertController.swift
```diff
@@ -0,0 +1,79 @@
...
+    // TODO(Mairramer): Investigate if this can be removed once the issue is fixed.
+    // Workaround for an issue with window hierarchy during the migration from Objective-C to Swift.
+    // The window hierarchy is not always properly maintained after the migration, which can lead to
+    // inconsistencies when presenting alerts.
+    // This approach ensures that the alert is presented only when the view hierarchy is fully initialized and the
+    // presenting view controller's window is available.
+    open override func present(
+      _ presentingViewController: UIViewController, animated: Bool, completion: (() -> Void)?
+    ) {
+      DispatchQueue.main.async {
```
stuartmorgan, line 38: If there is code running on an non-main thread that wants to show an alert, that should be handled at the call site, not abstracted away in an extension.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/AlertController.swift
```diff
@@ -0,0 +1,79 @@
...
+  #endif
+
+}
+
+#if os(macOS)
+  class DefaultAlertFactory: NSObject, AlertFactory {
+    // TODO(Mairramer): Check if is possible to remove the dispatch queue, and use the @MainActor annotation instead.
+    func createAlert() -> Alert {
+      var alert: Alert!
+      DispatchQueue.main.sync {
```
stuartmorgan, line 65: Same comment here. We certainly shouldn't be changing threading behavior in code that's not unit-testable. The Obj-C versions of the default implementations were, by explicit design, direct passthroughs to the underlying class, and that should be true for Swift as well.

stuartmorgan (changes requested):

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -416,146 +457,227 @@ class FLALocalAuthPluginTests: XCTestCase {
...
+
+    do {
+      let result = try plugin.deviceCanSupportBiometrics()
+      XCTAssertTrue(result)
+    } catch let thrownError as FlutterError {
+      error = thrownError
+    } catch {
+      XCTFail("Unexpected error thrown: \(error)")
+    }
     XCTAssertNil(error)
```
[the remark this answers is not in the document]
Mairramer: `XCTAssertThrowsError` is indeed the standard when you want to test if a function throws an error, but in this case, since we have a `Result`, it’s also common. It depends more on whether we want to capture that error.

Mairramer:
Sorry for the delay (1 month away from code).

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/AlertController.swift
```diff
@@ -0,0 +1,79 @@
...
+  }
+#endif
+
+#if os(macOS)
+  extension NSAlert: Alert {}
+#elseif os(iOS)
+  extension UIAlertController: AlertController {
+    // TODO(Mairramer): Investigate if this can be removed once the issue is fixed.
+    // Workaround for an issue with window hierarchy during the migration from Objective-C to Swift.
+    // The window hierarchy is not always properly maintained after the migration, which can lead to
```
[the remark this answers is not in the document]
Mairramer: For some reason, this is happening. Before I went on vacation, I was looking into this.

jmagman:

PR #7591, opened 2024-09-06
[RFC] [cross_file] New architecture.

ditman:
This PR attempts to solve the current issues of `XFile` by slightly re-architecting it.

In this PR:

* `XFile` is now an interface that does not have any code.
* Each underlying data structure contains a super simple implementation of the interface.
* `XFile` instances are now created from the platform-specific (`native` vs `web`) factories.

This does not remove `saveAs` from the `XFile` interface.

> [!NOTE]
> This is a breaking change because I've made the old `XFile` constructors deprecated (and _throwy_) but if we want to, we could possibly re-implement the "old" constructors with the new implementations and a few conditional imports?

Anyway, just a RFC!

## Pre-launch Checklist

- [ ] I signed the [CLA].


ditman:
/cc @stuartmorgan this is my idea for the "next" version of the x-files. PTAL and let me know what you think, or if you want this to go in another direction :)

The_Truth_Is_Out_There_tagline

packages/cross_file/lib/src/implementations/blob_x_file.dart
```diff
@@ -0,0 +1,109 @@
...
+  ///
+  /// See: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static
+  static Future<BlobXFile> fromObjectURL(
+    String objectUrl, {
+    String? mimeType,
+    String? displayName,
+    DateTime? lastModified,
+  }) async {
+    return BlobXFile(
+      await fetchBlob(objectUrl),
```
ditman, line 53: "ObjectURL" files could be made more specialized if we keep the objectURL as the information until the data is needed, and THEN we `fetchBlob`. This is a similar case as `.fromPath` from the native side.

packages/cross_file/lib/src/implementations/io_bytes_x_file.dart
```diff
@@ -0,0 +1,25 @@
...
+  BytesXFile(
+    super.bytes, {
+    super.mimeType,
+    super.displayName,
+    super.lastModified,
+  });
+
+  @override
+  Future<void> saveTo(String path) async {
+    final File fileToSave = File(path);
```
ditman, line 22: This is lifted from the old implementation but... should the `File` be created from `path + this.name`?

packages/cross_file/lib/src/implementations/web_bytes_x_file.dart
```diff
@@ -0,0 +1,27 @@
...
+    super.displayName,
+    super.lastModified,
+  });
+
+  /// Saves the data of this CrossFile at the location indicated by path.
+  /// For the web implementation, the path variable is ignored.
+  @override
+  Future<void> saveTo(String path) async {
+    // Convert [bytes] into a JS Blob, and save it in [path].
+    await downloadBlob(bytesToBlob(bytes, mimeType), name ?? '');
```
ditman, line 25: `name` is nullable, no need to `?? ''`

packages/cross_file/lib/src/implementations/blob_x_file.dart
ditman: Done.

packages/cross_file/lib/src/implementations/web_bytes_x_file.dart
ditman: Not nullable anymore.

packages/cross_file/lib/src/implementations/io_x_file.dart
```diff
@@ -0,0 +1,86 @@
...
+
+import 'dart:convert';
+import 'dart:io';
+import 'dart:typed_data';
+
+import '../x_file.dart';
+
+/// A CrossFile backed by a dart:io [File].
+///
+/// (Mobile-only).
```
navaronbracke, line 13: Don't forget desktop :)

```suggestion
/// This is only supported on non-web platforms.
```

packages/cross_file/lib/src/implementations/blob_x_file.dart
```diff
@@ -0,0 +1,150 @@
...
+  // The Blob backing the file.
+  @override
+  Future<Blob> get blob async => _blob;
+
+  /// Attempts to save the data of this [XFile], using the passed-in `blob`.
+  ///
+  /// The [path] variable is ignored.
+  @override
+  Future<void> saveTo(String path) async {
+    // Save a Blob to file...
```
navaronbracke, line 110: This comment seems a bit redundant.

packages/cross_file/README.md
```diff
@@ -27,16 +50,12 @@ You will find links to the API docs on the [pub page](https://pub.dev/packages/c
...
 ## Web Limitations
 
-`XFile` on the web platform is backed by [Blob](https://api.dart.dev/be/180361/dart-html/Blob-class.html)
+`XFile` on the web platform is backed by `Blob`
 objects and their URLs.
 
 It seems that Safari hangs when reading Blobs larger than 4GB (your app will stop
 without returning any data, or throwing an exception).
 
-This package will attempt to throw an `Exception` before a large file is accessed
```
navaronbracke, line 36: Just to be sure, was this comment removed, because we now refer to the browser's behavior? Or was this fixed in recent versions of Safari?

packages/cross_file/README.md
```diff
@@ -4,15 +4,38 @@ An abstraction to allow working with files across multiple platforms.
...
-using a path or byte array and use its methods and properties to
-access the file and its metadata.
+Many packages use `XFile` as their return type. In order for your
+application to consume those files, import 
+`package:cross_file/cross_file.dart`, and use its methods and properties
+to access the file data and metadata.
 
-Example:
+In order to instantiate a new `XFile`, import the correct factory class,
+either from `package:cross_file/native/factory.dart` (for native development) or
```
navaronbracke, line 13: Small drive-by comment:

Since the two implementations now have platform-specific imports (dart:io or package:web)
users who take the imports at face value might run into compilation issues, because they lack a conditional import?

Can we provide guidance on how to avoid this? With the legacy API we had the File or Blob types as internals and not on the public interface, which allowed for the plugin to provide a conditional import shim.

stuartmorgan:
Quick note to say I haven't forgotten about this, I just haven't had time to fully context switch to give this the attention it will need. I'll try to do it within the next week.

packages/cross_file/README.md
ditman: We don't explicitly throw the "4 Gigabyte" exception; the browser may fail in unexpected ways (depending on what we end up doing with the bytes of the file)

packages/cross_file/README.md
ditman: @navaronbracke thanks for the comment! This is a very interesting problem, for which I don't have a great answer. The TL;DR/Guidance is:

* Are you reading XFiles? Use the `XFile` interface only.
* Are you **creating** XFiles?
  * In mobile? Use the `native/factory.dart`
  * In web? Use the `web/factory.dart`

Each factory exposes constructors that are platform-specific, and we really don't have a cross-platform file constructor.

What problems do you think people are going to have? Is this a documentation problem, or a design issue?

navaronbracke: If people take the comment at face value and do `import "package:cross_file/web/factory.dart";` directly, they will run into errors like `library dart:js_interop is not available on this platform`. Since they don't directly import `dart:js_interop`, this might be confusing to new users.

I think this is merely a documentation issue. We should put emphasis on:
1) If this is in a platform-specific plugin implementation, you can probably import it directly (as the plugin system only uses the platform implementations of your plugin on the correct platforms)
2) If this is in a Flutter app, you'll have to use a conditional import on `if (dart.library.js_interop) ...` and you probably need an extra abstraction, since the web-only types (i.e. Blob) cannot be imported in native implementations (for the same reason why the plugin cannot provide the conditional import here)

ditman: OK, if it's a documentation issue, it's probably easier to fix (LOL famous last words :P)

Do you have any ideas with how this comment would be clearer? What would you like to read?

(I'll try to come up with something more specific as well!)

navaronbracke: Hmm, the main point is that the compiler itself would want you to use a conditional import.
So perhaps something like: 

"In order to instantiate a new `XFile`, import the correct factory class, through a conditional import.
Using a conditional import is required to use types that are only available on the current platform.

```
import 'package:cross_file/native/factory.dart'
 if (dart.library.js_interop) 'package:cross_file/web/factory.dart'
```
"

The wording is just a suggestion, so feel free to tweak it.

navaronbracke: Hmm, wait. Since the API for both factories is divergent, would a user that uses this conditional import know at compile time if they could access the `fromFile()` (both with a dart:io or package:web File) or the `fromBlob()/fromObjectUrl()` methods? I think that Dart will only provide the base type `XFileFactory`, but I _think_ you don't get a guarantee for the available methods.

Perhaps we can fix that by having the interface like:

- `XFileFactory` is the base type that is imported through a conditional import. It provides all the methods that do not include platform-specific types (so only fromPath/fromBytes/fromObjectUrl)
- `WebXFileFactory` and `NativeXFileFactory` implement that interface
- `WebXFileFactory` provides the `fromBlob/fromFile` implementations for package:web types
- `NativeXFileFactory` provides the `fromFile` implementation for dart:io
- unsupported methods on a platform just throw `UnsupportedError`
- we direct users to use `kIsWeb` and/or a type check on the factory

ditman: Thanks for the clarification @navaronbracke! I see what you mean! You want the `XFileFactory` to be cross-platform as well!

IMO the only cross-platform class from this package is **the `XFile` interface**! [^1]

If your cross-platform App needs to "create XFiles", you'll end up with some platform-aware code (because as you said, you can't import web stuff in native code), like this:

```dart
// Your app's main.dart
import 'src/get_file/get_file.dart'; // Platform-aware, see below

// ...

final XFile myFile = await getFile(/* config? data? */);
```

With a directory like this:

```
src/
  get_file/
    get_file.dart
    get_file_web.dart
    get_file_native.dart
```

Your `get_file.dart` could [conditionally export](https://dart.dev/guides/libraries/create-packages#conditionally-importing-and-exporting-library-files) the right file for the platform; like:

```dart
// get_file.dart

export "get_file_native.dart"
  if (dart.library.js_interop) "get_file_web.dart";
```

So each implementation file can use the platform-specific factories:

<table>
<tr>
  <th><tt>get_file_native.dart</tt></th>
  <th><tt>get_file_web.dart</tt></th>
</tr>
<tr>
<td>

```dart
import 'package:cross_file/native/factory.dart';

XFile getFile(/* config? data? */) async {
  // Do the right thing for native here, maybe
  // we write the data to a File in a temp dir
  // and then:
  return XFileFactory.fromFile(tmpFile);
}
```

</td>
<td>

```dart
import 'package:cross_file/web/factory.dart';

XFile getFile(/* config? data? */) async {
  // Do the right thing for the web here,
  // maybe we grab bytes, put them into a
  // Blob and do:
  return XFileFactory.fromBlob(dataBlob);
}
```

</td>
</tr>
</table>

Maybe the confusing part is calling the `XFileFactory` the same on both `native` and `web`? Should we just have `NativeXFileFactory` and `WebXFileFactory` to make the separation clearer, and signal that we never expect a shared interface across both classes?

---

> * It provides all the methods that do not include platform-specific types (so only fromPath/fromBytes/fromObjectUrl)

This is why I think having a base class for the `XFileFactory` is more trouble than what it's worth. Unless I'm missing something, for this approach to work, the `XFileFactory` needs to be an actual `instance` that must be replaced at run-time (similar to a federated plugin?).

Since we can't expose web-only types through this common API, users would still need to **conditionally import the native/web types to cast the instance** to the one that has the platform specific methods. And if you're conditionally importing, just **import the right `XFileFactory` static class?** 😛

[^1]: This is similar to the `Client` interface from `package:http` that has [multiple platform-specific implementations](https://pub.dev/packages/http#choosing-an-implementation) coming from different packages

stuartmorgan (commented):
> I'll try to do it within the next week.

😬 Time is an illusion, review time doubly so.

I like this a lot; if we don't want to jump to a breaking change (although I'm not 100% sure this isn't breaking in some obscure ways; we should probably do a quick check of GitHub to see if anyone is subclassing the old structure in any way that isn't compatible and if not call it non-breaking) this looks like a solid implementation.

I do wish we could do static extensions on `XFile` itself, but since we can't I do like your factory class approach more than what I had originally envisioned which was people interacting directly with subclass constructors. The way your version makes the subclass hidden implementation details that we can adjust over time is much better than that.

Thanks for tackling this, and sorry it took me so long to review!

packages/cross_file/lib/src/x_file.dart
```diff
@@ -2,6 +2,140 @@
...
+      name: name,
+      length: length,
+      lastModified: lastModified,
+      path: path,
+    );
+  }
+
+  /// Get the path of the picked file.
+  ///
+  /// **Not all `XFile` instances have a `path`.**
```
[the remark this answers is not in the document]
ditman: I think the changes in this package are sufficiently big to do a breaking change, with all the changes we might want.

I just did a non-breaking change "because I could", but as you said above, it might be "subtly breaking elsewhere".

How do you feel about making this a breaking change? There would be extra work in revisiting the consumers of this package and fixing those (which would also need a major version bump because they tend to re-export XFile as well :S)

I don't mind tackling that as well, LMK if you're okay with the change!

packages/cross_file/README.md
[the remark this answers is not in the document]
ditman: The problem with conditional exports/imports is that the analyzer freaks out when the APIs are different, which in this case they are. Or do you have something else in mind?

packages/cross_file/lib/src/x_file.dart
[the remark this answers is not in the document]
stuartmorgan: > There would be extra work in revisiting the consumers of this package and fixing those (which would also need a major version bump because they tend to re-export XFile as well :S)

Yes, working that through the ecosystem could be kind of rough. I think since your version doesn't make any real structural compromises to make a probably-not-breaking change, I want to dig into usage a bit. My gut is that nobody is doing complex subclassing that would be affected by it, but I want to try to validate that against at least the public code I can find, and especially some popular packages. If it look like it'll be safe (e.g., nobody is ever subclassing any of the classes), then we can do this version and make things easier on everyone.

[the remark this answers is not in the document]
ditman: There's some subclassing going on, most of it to add extra attributes/methods to the underlying XFile instance (they could wrap the underlying XFile and add stuff to it, or use extension methods I think?), one that could maybe migrated to one of the new web constructors, one used to store the file in "Hive", and one that is importing from src (this is illegal!)

* `extends XFile` ([search](https://github.com/search?q=%22extends+XFile%22+language%3ADart&type=code&l=Dart&p=1))
  * https://github.com/MixinNetwork/flutter-plugins/blob/b2beb15590f7b6c38e650ae501923b9bd3e7849c/packages/desktop_drop/lib/src/drop_item.dart#L5 (adds properties to XFile)
  * https://github.com/vidklopcic/flutter_persistent_socket/blob/1f7dfd216e9360c4d6bfb7aad8b29c0a897d6d22/lib/uploader/file_picker/file_picker_web.dart#L30 (`BlobXFile.fromFile`?)
  * https://github.com/slatei/asset_manager/blob/ba5b362b3c6ad26986b6448750d245d4823e3c17/lib/shared/hashed_file.dart#L5 (adds properties to XFile)
  * https://github.com/Wayaer/fl_assets_picker/blob/a66c525a9d9d937c1f31750b8b9f2d76513fa6ae/fl_image_picker/lib/src/extension.dart#L7 (adds properties to XFile)
  * https://github.com/diego78795/contact_app/blob/0d224aad5fb31353336569408133ef9951223003/lib/adapters/image_adapter.dart#L4 (extension method on XFile?)
  * https://github.com/correiarangel/camera_poc/blob/7d9e37549aa90001dfd97ed72bf65c37002c5dff/lib/core/shared/services/camera/library/z_xfile.dart#L3 (proxies XFile)
  * https://github.com/BreX900/cross_file_picker/blob/02dec6132bf5dacb020a6592d847d219d3aeded8/lib/src/x_file_stream.dart#L8 (imports from 'src')
  * https://github.com/monzir3bdo/task/blob/d2d4f906ed34bdaa6e92cc73a7b83512a1f66509/lib/features/products/data/models/photo_model.dart#L9 (could use the new interface)
* `implements XFile` ([search](https://github.com/search?q=%22implements+XFile%22+language%3ADart&type=code&l=Dart))
  * All seems to be test code.
* (extension ...) `on XFile` ([search](https://github.com/search?q=%22on+XFile%22+language%3ADart&type=code))
  * Everything should work (?)

---

I could create issues in their respective owner's repos to point them to this PR so they can see what's coming before we land it?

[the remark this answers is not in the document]
stuartmorgan: That looks like we would break existing packages, unfortunately. Warning people in advance isn't enough for that, we'd need to version accordingly.

So I think the options are:
1. Do a breaking change and deal with the resulting pain
2. Make the new interface a super-interface of `XFileBase`, and let individual packages adopt that incrementally (although that would still be a breaking change for the packages that export it in their API).

The main drawback to 2 is that we need to change the name of the class everyone uses, and we would be stuck with the legacy bits for however long we want to support both.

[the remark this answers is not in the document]
ditman: I think I prefer 1.

With the current implementation, the package is backwards-compatible for ourselves, so the transition for our packages should be relatively painless (and we could allow a version range).

(Unless we wan to do the nullability changes at the same time :))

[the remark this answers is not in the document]
stuartmorgan: If we're going to do a major version change, that's our chance to fix nullability; if we don't do it then we probably never will.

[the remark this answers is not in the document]
ditman: OK, I'll do all the incompatible changes in one go. Can you think of anything else? Maybe remove the `saveTo` method from the class, and make it a helper instead?

ditman:
I'm going to work on this on my next flutter/packages on-call rotation.

Piinks:
Hey @ditman! Greetings from stale PR triage. :)
Do you plan to pick this back up?

ditman:
@Piinks I do want to pick this up again, but I don't think I'll have time to do this before I/O :(

Coinners:
any updates?

Coinners:
sorry @ditman to bother you but do you plan on finishing this PR

Piinks:
Hey @ditman! Greetings from stale PR triage (again).
Should we continue to leave this open?

Piinks:
Given the many return trips to stale PR triage, the many merge conflicts, and the lack of progress for a year now, I am going to close this PR. Please feel free to reopen if needed @ditman. Thank you!

ditman:
Thanks @Piinks, no need to reopen, @bparrishMines is cooking (cooked already?) something better!

PR #7614, opened 2024-09-10
[pigeon] Validates support for nested collections with generic types.

tarrinneal:
Validates support for nested collections with generic types.

Really just adds tests to ensure behavior. 

small step toward https://github.com/flutter/flutter/issues/116117


stuartmorgan (approved):
LGTM

PR #7631, opened 2024-09-11
[google_maps_flutter_android] Convert `PlatformPolyline.pattern` to Pigeon

yaakovschectman:
Creates a structured `PlatformPatternItem` pigeon class to use in `PlatformPolyline`.

https://github.com/flutter/flutter/issues/154738

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -1166,6 +1165,29 @@ PlatformJointType platformJointTypeFromJointType(JointType jointType) {
...
+/// Converts a PatternItem to Pigeon's PlatformPatternItem for PlatformPolyline
+/// pattern member.
+@visibleForTesting
+PlatformPatternItem platformPatternItemFromPatternItem(PatternItem item) {
+  final List<Object> json = item.toJson() as List<Object>;
+  final String tag = json[0] as String;
+  PlatformPatternItemType type;
+  double? length;
+  switch (tag) {
+    case 'dot':
```
reidbaker, line 1177: Consider linking to the documentation for where these values are defined.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/messages.g.dart
```diff
@@ -1048,50 +1083,56 @@ class _PigeonCodec extends StandardMessageCodec {
     } else if (value is PlatformPolyline) {
       buffer.putUint8(138);
       writeValue(buffer, value.encode());
-    } else if (value is PlatformTile) {
+    } else if (value is PlatformPatternItem) {
       buffer.putUint8(139);
```
reidbaker, line 1087: I assume no but is there an issue with these ints changing values and users having miss matched versions?

yaakovschectman: Can you clarify what you mean by mismatched versions? Specifically, of what? These ints are only used within `google_maps_flutter_android`, at least as far as I can tell, which should mean that as long as they are consistent within each version, this should not pose an issue. Do you think I am mistaken about this, or misunderstanding your concern?

stuartmorgan: > These ints are only used within `google_maps_flutter_android`, at least as far as I can tell

That is correct, we never allow the two sides of Pigeon-generated code to span packages, because it would mean that any Pigeon changes or message change would require breaking changes, which would be a nightmare. We always use it internal to each implementation package.

(In theory someone could accidentally land a PR that only updated one language, but it would be a hard mistake to make, and a very hard mistake to get through automated test. We've discussed having a Pigeon "handshake" protocol to error out if the two sides don't match, but we haven't implemented it since it's low priority.)

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -870,6 +870,30 @@ private static List<List<LatLng>> toHoles(List<List<Messages.PlatformLatLng>> da
...
+        case GAP:
+          assert patternItem.getLength() != null;
+          pattern.add(new Gap(patternItem.getLength().floatValue()));
+          break;
+      }
+    }
+    return pattern;
+  }
+
   private static List<PatternItem> toPattern(Object o) {
```
stuartmorgan, line 897: This is dead code, and should be removed. Please always check for dead conversion code when doing Pigeon conversions.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/messages.g.dart
reidbaker: Thanks that helps me understand the code better.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -1166,6 +1165,34 @@ PlatformJointType platformJointTypeFromJointType(JointType jointType) {
...
+PlatformPatternItem platformPatternItemFromPatternItem(PatternItem item) {
+  final List<Object> json = item.toJson() as List<Object>;
+  final String tag = json[0] as String;
+  PlatformPatternItemType type;
+  double? length;
+
+  /// These string values identify the type of pattern. They are defined and
+  /// used in the PatternItem class's factory methods in
+  /// lib/src/types/pattern_item.dart, in the
+  /// google_maps_flutter_platform_interface package.
```
stuartmorgan, line 1180: This is basically the same problem as camera update; the real solution is to have structured information (either subclasses like camera update, or maybe just a type enum and an optional field for length since it seems plausible that those would be enough for the future).

You don't need to fix that here since this instance isn't as egregiously problematic as camera update was, but please file an issue to track it and reference it in a TODO here so it's clear to future us that this isn't desirable code.

stuartmorgan (changes requested):

stuartmorgan (approved):
LGTM with the re-addition of dead code removed.

PR #7648, opened 2024-09-14
[google_maps_flutter] Add ability to animate camera with duration

jokerttu:
Adds ability to configure camera animation duration on Android and iOS.

Resolves [#39810](https://github.com/flutter/flutter/issues/39810)
Resolves [#44284](https://github.com/flutter/flutter/issues/44284)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
+++ b/packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
@@ -13,6 +13,20 @@ import 'package:integration_test/integration_test.dart';
 
 import 'shared.dart';
 
+const double _kTestCameraZoomLevel = 10;
+const double _kTestZoomByAmount = 2;
+const LatLng _kTestMapCenter = LatLng(65, 25.5);
+const CameraPosition _kTestCameraPosition = CameraPosition(
+  target: _kTestMapCenter,
+  zoom: _kTestCameraZoomLevel,
+  bearing: 1.0,
+  tilt: 1.0,
+);
+final LatLngBounds _testCameraBounds = LatLngBounds(
+    northeast: const LatLng(50, -65), southwest: const LatLng(28.5, -123));
+final ValueVariant<CameraUpdateType> _cameraUpdateTypeVariants =
+    ValueVariant<CameraUpdateType>(CameraUpdateType.values.toSet());
+
 /// Integration Tests that use the [GoogleMapsInspectorPlatform].
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
@@ -612,6 +626,248 @@ void runTests() {
       expect(clusters.length, 0);
     }
   });
+
+  testWidgets(
+    'testAnimateCameraWithoutDuration',
+    (WidgetTester tester) async {
+      final Key key = GlobalKey();
+      final Completer<GoogleMapController> controllerCompleter =
+          Completer<GoogleMapController>();
+      final GoogleMapsInspectorPlatform inspector =
+          GoogleMapsInspectorPlatform.instance!;
+
+      /// Completer to track when the camera has come to rest.
+      Completer<void>? cameraIdleCompleter;
+
+      await tester.pumpWidget(Directionality(
+        textDirection: TextDirection.ltr,
+        child: GoogleMap(
+          key: key,
+          initialCameraPosition: kInitialCameraPosition,
+          onCameraIdle: () {
+            if (cameraIdleCompleter != null &&
+                !cameraIdleCompleter.isCompleted) {
+              cameraIdleCompleter.complete();
+            }
+          },
+          onMapCreated: (GoogleMapController controller) {
+            controllerCompleter.complete(controller);
+          },
+        ),
+      ));
+
+      final GoogleMapController controller = await controllerCompleter.future;
+
+      await tester.pumpAndSettle();
+      // TODO(cyanglaz): Remove this after we added `mapRendered` callback, and
+      // `mapControllerCompleter.complete(controller)` above should happen in
+      // `mapRendered`.
+      // https://github.com/flutter/flutter/issues/54758
+      await Future<void>.delayed(const Duration(seconds: 1));
+
+      // Create completer for camera idle event.
+      cameraIdleCompleter = Completer<void>();
+
+      final CameraUpdate cameraUpdate =
+          _getCameraUpdateForType(_cameraUpdateTypeVariants.currentValue!);
+      await controller.animateCamera(cameraUpdate);
+
+      // If platform supportes getting camera position, check that the camera
+      // has moved as expected.
+      CameraPosition? beforeFinishedPosition;
+      if (inspector.supportsGettingGameraPosition()) {
+        // Immediately after calling animateCamera, check that the camera hasn't
+        // reached its final position. This relies on the assumption that the
+        // camera move is animated and won't complete instantly.
+        beforeFinishedPosition =
+            await inspector.getCameraPosition(mapId: controller.mapId);
+
+        await _checkCameraUpdateByType(
+            _cameraUpdateTypeVariants.currentValue!,
+            beforeFinishedPosition,
+            null,
+            controller,
+            (Matcher matcher) => isNot(matcher));
+      }
+
+      // Wait for the animation to complete (onCameraIdle).
+      expect(cameraIdleCompleter.isCompleted, isFalse);
+      await cameraIdleCompleter.future;
+
+      // If platform supportes getting camera position, check that the camera
+      // has moved as expected.
+      if (inspector.supportsGettingGameraPosition()) {
+        // After onCameraIdle event, the camera should be at the final position.
+        final CameraPosition afterFinishedPosition =
+            await inspector.getCameraPosition(mapId: controller.mapId);
+        await _checkCameraUpdateByType(
+            _cameraUpdateTypeVariants.currentValue!,
+            afterFinishedPosition,
+            beforeFinishedPosition,
+            controller,
+            (Matcher matcher) => matcher);
+      }
+    },
+    variant: _cameraUpdateTypeVariants,
+    // TODO(stuartmorgan): Remove skip for Android platform once Maps API key is
+    // available for LUCI, https://github.com/flutter/flutter/issues/131071
+    skip: isAndroid,
+  );
+
+  /// Tests animating the camera with specified durations to verify timing
+  /// behavior.
+  ///
+  /// This test checks two scenarios: short and long animation durations.
+  /// It uses a midpoint duration to ensure the short animation completes in
+  /// less time and the long animation takes more time than that midpoint.
+  /// This ensures that the animation duration is respected by the platform and
+  /// that the default camera animation duration does not affect the test
+  /// results.
+  testWidgets(
+    'testAnimateCameraWithDuration',
+    (WidgetTester tester) async {
+      final Key key = GlobalKey();
+      final Completer<GoogleMapController> controllerCompleter =
+          Completer<GoogleMapController>();
+      final GoogleMapsInspectorPlatform inspector =
+          GoogleMapsInspectorPlatform.instance!;
+
+      /// Completer to track when the camera has come to rest.
+      Completer<void>? cameraIdleCompleter;
+
+      const int shortCameraAnimationDurationMS = 200;
+      const int longCameraAnimationDurationMS = 1000;
+
+      /// Calculate the midpoint duration of the animation test, which will
+      /// serve as a reference to verify that animations complete more quickly
+      /// with shorter durations and more slowly with longer durations.
+      const int animationDurationMiddlePoint =
+          (shortCameraAnimationDurationMS + longCameraAnimationDurationMS) ~/ 2;
+
+      // Stopwatch to measure the time taken for the animation to complete.
+      final Stopwatch stopwatch = Stopwatch();
+
+      await tester.pumpWidget(Directionality(
+        textDirection: TextDirection.ltr,
+        child: GoogleMap(
+          key: key,
+          initialCameraPosition: kInitialCameraPosition,
+          onCameraIdle: () {
+            if (cameraIdleCompleter != null &&
+                !cameraIdleCompleter.isCompleted) {
+              stopwatch.stop();
+              cameraIdleCompleter.complete();
+            }
+          },
+          onMapCreated: (GoogleMapController controller) {
+            controllerCompleter.complete(controller);
+          },
+        ),
+      ));
+
+      final GoogleMapController controller = await controllerCompleter.future;
+
+      await tester.pumpAndSettle();
+      // TODO(cyanglaz): Remove this after we added `mapRendered` callback, and
+      // `mapControllerCompleter.complete(controller)` above should happen in
+      // `mapRendered`.
+      // https://github.com/flutter/flutter/issues/54758
+      await Future<void>.delayed(const Duration(seconds: 1));
+
+      // Create completer for camera idle event.
+      cameraIdleCompleter = Completer<void>();
+
+      // Start stopwatch to check the time taken for the animation to complete.
+      // Stopwatch is stopped on camera idle callback.
+      stopwatch.reset();
+      stopwatch.start();
+
+      // First phase with shorter animation duration.
+      final CameraUpdate cameraUpdateShort =
+          _getCameraUpdateForType(_cameraUpdateTypeVariants.currentValue!);
+      await controller.animateCamera(
+        cameraUpdateShort,
+        duration: const Duration(milliseconds: shortCameraAnimationDurationMS),
+      );
+
+      // Wait for the animation to complete (onCameraIdle).
+      expect(cameraIdleCompleter.isCompleted, isFalse);
+      await cameraIdleCompleter.future;
+
+      // For short animation duration, check that the animation is completed
+      // faster than the midpoint benchmark.
+      expect(stopwatch.elapsedMilliseconds,
+          lessThan(animationDurationMiddlePoint));
+
+      // Reset camera to initial position before testing long duration.
+      await controller
+          .moveCamera(CameraUpdate.newCameraPosition(kInitialCameraPosition));
+      await tester.pumpAndSettle();
+
+      // Create completer for camera idle event.
+      cameraIdleCompleter = Completer<void>();
+
+      // Start stopwatch to check the time taken for the animation to complete.
+      // Stopwatch is stopped on camera idle callback.
+      stopwatch.reset();
+      stopwatch.start();
+
+      // Second phase with longer animation duration.
+      final CameraUpdate cameraUpdateLong =
+          _getCameraUpdateForType(_cameraUpdateTypeVariants.currentValue!);
+      await controller.animateCamera(
+        cameraUpdateLong,
+        duration: const Duration(milliseconds: longCameraAnimationDurationMS),
+      );
+
+      // If platform supportes getting camera position, check that the camera
+      // has moved as expected.
+      CameraPosition? beforeFinishedPosition;
+      if (inspector.supportsGettingGameraPosition()) {
+        // Immediately after calling animateCamera, check that the camera hasn't
+        // reached its final position. This relies on the assumption that the
+        // camera move is animated and won't complete instantly.
+        beforeFinishedPosition =
+            await inspector.getCameraPosition(mapId: controller.mapId);
+
+        await _checkCameraUpdateByType(
+            _cameraUpdateTypeVariants.currentValue!,
+            beforeFinishedPosition,
+            null,
+            controller,
+            (Matcher matcher) => isNot(matcher));
+      }
+
+      // Wait for the animation to complete (onCameraIdle).
+      expect(cameraIdleCompleter.isCompleted, isFalse);
+      await cameraIdleCompleter.future;
+
+      // For longer animation duration, check that the animation is completed
+      // slower than the midpoint benchmark.
+      expect(stopwatch.elapsedMilliseconds,
+          greaterThan(animationDurationMiddlePoint));
+
+      // If platform supportes getting camera position, check that the camera
+      // has moved as expected.
+      if (inspector.supportsGettingGameraPosition()) {
+        // Camera should be at the final position.
+        final CameraPosition afterFinishedPosition =
+            await inspector.getCameraPosition(mapId: controller.mapId);
+        await _checkCameraUpdateByType(
+            _cameraUpdateTypeVariants.currentValue!,
+            afterFinishedPosition,
+            beforeFinishedPosition,
+            controller,
+            (Matcher matcher) => matcher);
+      }
+    },
+    variant: _cameraUpdateTypeVariants,
+    // TODO(jokerttu): Remove skip once the web implementation is available,
+    // https://github.com/flutter/flutter/issues/159265
+    // TODO(stuartmorgan): Remove skip for Android platform once Maps API key is
+    // available for LUCI, https://github.com/flutter/flutter/issues/131071
+    skip: kIsWeb || isAndroid,
+  );
 }
 
 Marker _copyMarkerWithClusterManagerId(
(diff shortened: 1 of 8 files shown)


jokerttu:
Web support is not included, as there are still some areas open for discussion.
Test implementation here:
https://github.com/CodemateLtd/packages/compare/feature/animate-camera-duration...CodemateLtd:packages:feature/animate-camera-duration-web?expand=1

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
```diff
@@ -421,8 +421,20 @@ class MethodChannelGoogleMapsFlutter extends GoogleMapsFlutterPlatform {
     CameraUpdate cameraUpdate, {
     required int mapId,
   }) {
-    return channel(mapId).invokeMethod<void>('camera#animate', <String, Object>{
+    return channel(mapId).invokeMethod<void>('camera#animate',
+        <String, dynamic>{'cameraUpdate': cameraUpdate.toJson()});
```
stuartmorgan, line 425: Why was this changed to `dynamic`? We only use `dynamic` when we have an extremely compelling reason to do so as it bypasses a lot of static analysis.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
```diff
@@ -421,8 +421,20 @@ class MethodChannelGoogleMapsFlutter extends GoogleMapsFlutterPlatform {
...
+  @override
+  Future<void> animateCameraWithConfiguration(
+    CameraUpdate cameraUpdate, {
+    required CameraUpdateAnimationConfiguration configuration,
+    required int mapId,
+  }) {
+    return channel(mapId)
+        .invokeMethod<void>('camera#animate', <String, dynamic>{
       'cameraUpdate': cameraUpdate.toJson(),
+      'duration': configuration.duration?.inMilliseconds,
```
stuartmorgan, line 437: We don't generally want to send nulls in collections in method channels.

I would recommend just reverting the changes to this file; we don't use the method channel implementation, it's only still here because removing it requires a breaking change. There's no obligation to implement new features for it.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_flutter_platform.dart
```diff
@@ -194,6 +194,20 @@ abstract class GoogleMapsFlutterPlatform extends PlatformInterface {
...
   }
 
+  /// Starts an animated change of the map camera position using the provided
+  /// [CameraUpdateAnimationConfiguration].
+  ///
+  /// The returned [Future] completes after the change has been started on the
+  /// platform side.
+  Future<void> animateCameraWithConfiguration(
+    CameraUpdate cameraUpdate, {
+    required CameraUpdateAnimationConfiguration configuration,
```
stuartmorgan, line 204: This can be positional rather than named since it's a core parameter.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_flutter_platform.dart
```diff
@@ -194,6 +194,20 @@ abstract class GoogleMapsFlutterPlatform extends PlatformInterface {
...
+    required int mapId,
+  }) {
+    throw UnimplementedError(
+        'animateCameraWithConfiguration() has not been implemented.');
```
stuartmorgan, line 208: This should call `animateCamera`; see https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-platform-interface-method-parameters

stuartmorgan: And the forwarding should get a unit test in this package.

packages/google_maps_flutter/google_maps_flutter/lib/google_maps_flutter.dart
```diff
@@ -24,6 +24,7 @@ export 'package:google_maps_flutter_platform_interface/google_maps_flutter_platf
         CameraPositionCallback,
         CameraTargetBounds,
         CameraUpdate,
+        CameraUpdateAnimationConfiguration,
```
stuartmorgan, line 27: We generally don't want to expose these parameter objects at the app-facing package API level; they solve a problem (adding optional parameter being a breaking change for subclasses) that doesn't exist at this level, and add complexity for all clients (vs. just ourselves in implementation packages). The client-facing API should expose the duration directly, and that should be packed into the configuration object as an internal implementation detail.

packages/google_maps_flutter/google_maps_flutter/example/lib/animate_camera.dart
```diff
@@ -156,14 +180,33 @@ class AnimateCameraState extends State<AnimateCamera> {
...
+        Row(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
+            const Text(
+              'With configuration\n(10 second duration)',
+              textAlign: TextAlign.right,
+            ),
+            const SizedBox(width: 5),
+            Switch(
+              value: _cameraUpdateAnimationConfiguration != null,
```
stuartmorgan, line 201: These should all be "Duration" rather than "Configuration" once the other change is made.

packages/google_maps_flutter/google_maps_flutter/example/lib/animate_camera.dart
```diff
@@ -156,14 +180,33 @@ class AnimateCameraState extends State<AnimateCamera> {
...
               ],
             ),
           ],
-        )
+        ),
+        Row(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
+            const Text(
+              'With configuration\n(10 second duration)',
```
stuartmorgan, line 196: `'With 10 second duration'`

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
```diff
@@ -612,6 +627,186 @@ void runTests() {
...
+          _cameraUpdateTypeVariants.currentValue!,
+          afterFinishedPosition,
+          beforeFinishedPosition,
+          controller,
+          (Matcher matcher) => matcher);
+
+      await tester.pumpAndSettle();
+    },
+    variant: _cameraUpdateTypeVariants,
+    skip: kIsWeb,
```
stuartmorgan, line 808: This will need a tracking issue. Or is the plan to have web support ready before the app-facing PR lands?

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
```diff
@@ -612,6 +627,186 @@ void runTests() {
...
+          Completer<GoogleMapController>();
+      final GoogleMapsInspectorPlatform inspector =
+          GoogleMapsInspectorPlatform.instance!;
+
+      final StreamController<void> cameraFiredStream = StreamController<void>();
+      final StreamQueue<void> cameraFiredQueue =
+          StreamQueue<void>(cameraFiredStream.stream);
+
+      // We set the duration to 100ms and check that the animation is completed
+      // faster than the normal animation (~300ms on Android) to make sure the
```
stuartmorgan, line 718: This is run on other platforms as well, and it's not clear from the comment why this test is correct on those platforms.

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
```diff
@@ -612,6 +627,186 @@ void runTests() {
...
+      // configuration is really used.
+      const int cameraAnimationDurationMS = 100;
+
+      // The threshold is set to 200ms because the normal camera animation should
+      // take more than 200ms on android.
```
stuartmorgan, line 723: This is very likely to become a flaky test as written, since it assumes things about the timing of things beyond the animation itself (multiple async steps with cross-thread hops, potentially delays within the Google Maps SDK itself), and the tolerance here is relatively tight for environments like an emulator. It's also not ideal (per the previous comment) to have a test that would become subject to false negatives if a value we don't control in a different SDK changed.

What about having the test run once with an extremely short time, like 10ms, and once with a much longer time like a second, and make sure the difference between them is more than, say, half the larger value. That gives a fair amount of tolerance, but should still easily catch the value being ignored.

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
```diff
@@ -636,3 +831,94 @@ Marker _copyMarkerWithClusterManagerId(
     clusterManagerId: clusterManagerId,
   );
 }
+
+CameraUpdate _getCameraUpdateForType(CameraUpdateType type) {
+  switch (type) {
```
stuartmorgan, line 836: Consider using the newer `return switch (type) {...}` pattern syntax.

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
```diff
@@ -636,3 +831,94 @@ Marker _copyMarkerWithClusterManagerId(
...
+}
+
+Future<void> _checkCameraUpdateByType(
+  CameraUpdateType type,
+  CameraPosition currentPosition,
+  CameraPosition? oldPosition,
+  GoogleMapController controller,
+  Matcher Function(Matcher matcher) wrapMatcher,
+) async {
+  // The target might differ a bit from the expected target; this is why we use
```
stuartmorgan, line 865: "; this is why we" -> ", so"

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -103,6 +103,13 @@ class PlatformCameraUpdateZoomTo {
   final double zoom;
 }
 
+/// Pigeon representation of a CameraUpdateAnimationConfiguration.
+class PlatformCameraUpdateAnimationConfiguration {
```
stuartmorgan, line 107: Pigeon interfaces are internal APIs, and can be changed at will, so there is no need to use parameter objects in general; the method can just take a durationMilliseconds parameter directly (same on iOS).

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsTests.m
```diff
@@ -86,6 +87,122 @@ - (void)testHandleResultTileDownsamplesWideGamutImages {
...
+                             handler:^(NSError *_Nullable error) {
+                               if (error) {
+                                 XCTFail(@"Failed to receive camera idle event in time: %@", error);
+                               }
+                             }];
+
+  // Assert that the camera zoom is updated after the animation completes
+  XCTAssertEqual(controller.mapView.camera.zoom, 10.0,
+                 "Camera zoom should be updated to 10.0 after animation.");
+}
```
stuartmorgan, line 204: Actually running animations is pretty slow for unit tests; could we just dependency inject a simple wrapper around `CATransaction` and ensure that's being called as expected, rather than testing the animation itself (which is largely testing the underlying SDK rather than the plugin code)?

packages/google_maps_flutter/google_maps_flutter_ios/pigeons/messages.dart
```diff
@@ -51,6 +51,13 @@ class PlatformCameraUpdate {
   final Object json;
 }
 
+/// Pigeon representation of a CameraUpdateAnimationConfiguration.
+class PlatformCameraUpdateAnimationConfiguration {
```
stuartmorgan, line 55: Same note here as on Android.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
```diff
@@ -132,6 +132,15 @@ class GoogleMapsPlugin extends GoogleMapsFlutterPlatform {
     return moveCamera(cameraUpdate, mapId: mapId);
   }
 
+  @override
+  Future<void> animateCameraWithConfiguration(
+    CameraUpdate cameraUpdate, {
+    required CameraUpdateAnimationConfiguration configuration,
+    required int mapId,
+  }) {
+    return animateCamera(cameraUpdate, mapId: mapId);
```
stuartmorgan, line 141: This will need a TODO referencing an issue, as with the test disabling.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -937,12 +937,21 @@ public void moveCamera(@NonNull Messages.PlatformCameraUpdate cameraUpdate) {
...
     }
-    googleMap.animateCamera(Convert.cameraUpdateFromPigeon(cameraUpdate, density));
+    if (configuration != null && configuration.getDurationMilliseconds() != null) {
+      googleMap.animateCamera(
+          Convert.cameraUpdateFromPigeon(cameraUpdate, density),
+          configuration.getDurationMilliseconds().intValue(),
+          null);
+    } else {
+      googleMap.animateCamera(Convert.cameraUpdateFromPigeon(cameraUpdate, density));
+    }
```
stuartmorgan, line 954: There should be a native unit test validating the branching here (via a mock map).

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
jokerttu: Comment updated

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
jokerttu: I used suggested 10ms and 1000ms durations and tests seems to pass.
But the duration increase caused CI tests to timeout
@stuartmorgan, how to proceed with this?

stuartmorgan: Increasing a few integration test durations by a second shouldn't have caused timeouts; are you sure there wasn't some other change that's causing timeouts now?

jokerttu: @stuartmorgan 
Is there a way to access more detailed test output than this: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8729373696952662689/+/u/Run_package_tests/Firebase_Test_Lab/stdout
(I don't have access to the firebase console: https://console.firebase.google.com/project/flutter-infra-staging/testlab/histories/bh.8eb89ae93eb3eb36/matrices/6375674335592449274)

Tests run fine locally without timeouts.

Running tests for all CameraUpdateType variants (9 types) adds at least 12 seconds per package due to the testing with 200ms and 1000ms animation delays + map initialization. 

Should we consider testing just one type here since implementations between CameraUpdate types are identical on iOS and Android?
On web implementation testing each variation is more important because there is separate method for each camera update type.

stuartmorgan: Unfortunately the console isn't showing anything interesting that I can see. The way integration tests run in FTL we don't get incremental test output; so it's mostly device-level logs. For the rare cases like this where an issue happens in FTL but not locally, sometimes we have to do guess-and-check debugging unfortunately, by pushing alternate versions (I would suggest a new PR so you don't pollute this one with unwanted changes) with changes. E.g., try just a minimal new test, see if it works, then add more incrementally. It's not ideal, but the FTL-style run just doesn't always give us the info we need.

You could also try adding some native logging so that we can look for that in the device logs, to get a sense of what is happening on the native side. E.g., maybe log before each animated move request, with some details about the args to try to identify it, and also before sending each camera idle notification to Dart.

> Running tests for all CameraUpdateType variants (9 types) adds at least 12 seconds per package due to the testing with 200ms and 1000ms animation delays + map initialization.
> 
> Should we consider testing just one type here since implementations between CameraUpdate types are identical on iOS and Android?

We could, but for context the end-to-end integration test run for `google_maps_flutter_android` right now in CI is about 6 minutes, so 12 seconds of additional tests aren't really that big a deal.

jokerttu: Tested locally 
- with one CameraUpdate variant, all tests runned in average 48 seconds. (tested 3 times)
- with all 9 CameraUpdate variants, all tests runned in average 1 minute 22 seconds (tested 3 times)

I bet there is 1 minute timeout set for the tests.

Should we remove variants and test only with one CameraUpdate type?
@stuartmorgan ?

stuartmorgan: I was curious, so I just pulled this branch and tried it locally, and see it hang in my simulator. I'll debug a bit and see what I find.

stuartmorgan: What I see locally is that `await cameraFiredQueue.next;` never completes.

Are you testing with a maps key? We don't currently test with one in CI because in practice everything we were testing worked without it, but maybe this is something that doesn't work without a key?

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -1446,6 +1463,221 @@ void googleMapsTests() {
...
+      // Second phase with longer animation duration.
+      final CameraUpdate cameraUpdateLong =
+          _getCameraUpdateForType(_cameraUpdateTypeVariants.currentValue!);
+      await controller.animateCamera(
+        cameraUpdateLong,
+        duration: const Duration(milliseconds: longCameraAnimationDurationMS),
+      );
+
+      // Check that position is not updated immediately to the target.
+      final CameraPosition beforeFinishedPosition =
```
camsim99, line 1648: I have the same question here about flakiness as the test before. Is that concern why this is only checked for the longer animation and not the shorter?

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -1446,6 +1463,221 @@ void googleMapsTests() {
...
+          controller,
+          (Matcher matcher) => isNot(matcher));
+
+      // Wait for the camera idle callback to fire.
+      expect(cameraIdleFired, isFalse);
+      await cameraFiredQueue.next;
+      expect(cameraIdleFired, isTrue);
+
+      // For short animation duration, check that the animation is completed
+      // shower than the middle point of the animation test durations.
```
camsim99, line 1663: I think this was meant to be:
```suggestion
      // For longer animation duration, check that the animation is completed
      // slower than the middle point of the animation test durations.
```

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -1446,6 +1463,221 @@ void googleMapsTests() {
...
+      expect(stopwatch.elapsedMilliseconds,
+          greaterThan(animationDurationMiddlePoint));
+
+      // Check that position is animated after the animation is done.
```
camsim99, line 1667: Nit:
```suggestion
      // Check that position is updated after the animation is done.
```

camsim99 (commented):
The changes LGTM :) but left some questions/comments on the stability of the integration test.

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/maps_inspector.dart
jokerttu: @stuartmorgan 
`animateCamera` does not move the camera at all on Android if Google Maps API key is not given. 

Should we skip these tests on android for now?

stuartmorgan: In theory yes (https://github.com/flutter/flutter/issues/131071), but it will require some coordination with infra-level changes so it's not something I expect to happen in the very short term. We could try to prioritize it in Q1 of next year, potentially.

For now it's fine to skip this test on Android with a reference to that issue in a comment.

jokerttu: Skipping tests on Android
Tests works fine with api key, when enabled.

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -1446,6 +1463,221 @@ void googleMapsTests() {
...
+    if (cameraIdleFired) {
+      await cameraFiredQueue.next;
+      cameraIdleFired = false;
+    }
+
+    final CameraUpdate cameraUpdate =
+        _getCameraUpdateForType(_cameraUpdateTypeVariants.currentValue!);
+    await controller.animateCamera(cameraUpdate);
+
+    // Check that position is not updated immediately to the target.
```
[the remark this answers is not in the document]
jokerttu: Improved tests documentation to explain better why this is done.
And yes this could possible be flaky, but most likely not.
On android the default animation time is about 300ms. This should be long enought time to mitigate possible communication or SDK delays.
Reason why this is tested, is to mitigate out possiblity that camera is not animated, and is just moved instantly to new location. We could have also used stopwatch and wait cameraIdleEvent, and see if the event does not come too soon, but then we would need to hardcode limits.
It this test is removed, we can only test that camera is moved to right position, but cannot be aware if it is animated or not.

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -1446,6 +1463,221 @@ void googleMapsTests() {
...
+
+      // Wait for the camera idle callback to fire.
+      expect(cameraIdleFired, isFalse);
+      await cameraFiredQueue.next;
+      expect(cameraIdleFired, isTrue);
+
+      // For short animation duration, check that the animation is completed
+      // faster than the middle point of the animation test durations.
+      expect(stopwatch.elapsedMilliseconds,
+          lessThan(animationDurationMiddlePoint));
```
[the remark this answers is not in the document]
jokerttu: This solution is based on suggestion by @stuartmorgan https://github.com/flutter/packages/pull/7648#discussion_r1833390127
Testing with two times mitigate away the possiblity that animateCamera by default animates with same speed.
The duration won't be exactly same as detected by stopwatch, because internal delays on messaging and SDK functionality, so picking proper threshold for testing duration could not work.

Most likely having one test run with longer period, lets say 1 second, could work, but this assumes that default animation speed on each platform is different.

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
jokerttu: Short period is so short, that the might be possilbity that the latencies on communication and SDK might delay the camera position query so much that it is already animated to target location, causing flaky tests. 
For longer duration we could safely say that communication latencies are not causing 1 second delays.
Camera animation is not affected by networking either.

jokerttu:
Updated integration tests, by using Completer instead of StreamQueue to detect OnCameraIdle events.
This improved the test quality, and removed need to drain queue, which was possibly a bit flaky solution.

Integration tests are also now skipped on Android as they work only with Maps API Key which is not available on CI at the moment.
Tests passes on both projects for Android if enabled and Maps API Key is available.

camsim99 (commented):
Android changes LGTM! Thanks @jokerttu for the thorough explanations :)

devnoaman:
any update ?

jokerttu:
This is still awaiting iOS review from @cbracken (code owner) before I can proceed with splitting this into sub-PRs.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
```diff
@@ -362,6 +373,7 @@ class GoogleMapsPlugin extends GoogleMapsFlutterPlatform {
     GoogleMapsInspectorPlatform.instance = GoogleMapsInspectorWeb(
       (int mapId) => _map(mapId).configuration,
       (int mapId) => _map(mapId).clusterManagersController,
+      (int mapId) => _map(mapId),
```
ditman, line 376: Now that we're passing the whole map controller, the constructor for the `GoogleMapsInspectorWeb` could be simplified to just accept the controller, and have it internally extract `.configuration` and `.clusterManagersController`, as needed, I think.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
ditman: Unfortunately, I don't think Google Maps on the web supports *any* control of the duration of the animation... See docs for `moveCamera`, `panTo` and friends:

* https://developers.google.com/maps/documentation/javascript/reference/map#Map.moveCamera

So I think you can remove the TODO, the link and everything. Maybe we can log that the value will be silently ignored to the js debug console to let users know? Or at least document in the method of the plugin that the value will not be used on the web.

ditman (commented):
Web bits look OK to me, I had only a couple of minor remarks. Thanks!

reidbaker (commented):
Android code looks good. Deferring android final approval to @camsim99

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
jokerttu: Discussion about this on web-related issue here: https://github.com/flutter/flutter/issues/159265#issuecomment-2580116988

jokerttu: Web implementation is not necessary as platform interface now automatically defaults to 
`animateCamera`
https://github.com/flutter/packages/pull/7648/files#diff-82107f54c1af7174644137b415b811aa06290c54af28a3a81de0589ca9915e63R207
I will remove web-implementation from this PR as there is separate issue for web impementation: https://github.com/flutter/flutter/issues/159265

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/GoogleMapControllerTest.java
```diff
@@ -248,4 +255,47 @@ public void UpdateHeatmaps() {
...
+
+    try (MockedStatic<CameraUpdateFactory> mockedFactory = mockStatic(CameraUpdateFactory.class)) {
+      mockedFactory
+          .when(() -> CameraUpdateFactory.zoomBy(anyFloat()))
+          .thenReturn(mock(CameraUpdate.class));
+      googleMapController.animateCamera(cameraUpdate, durationMilliseconds);
+    }
+
+    verify(mockGoogleMap, times(1))
+        .animateCamera(any(CameraUpdate.class), eq(durationMilliseconds.intValue()), isNull());
```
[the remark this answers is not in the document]
jokerttu: animateCamera of googleMap instance is called like this:
```java
      googleMap.animateCamera(
          Convert.cameraUpdateFromPigeon(cameraUpdate, density),
          durationMilliseconds.intValue(),
          null);
```

This line only verifies that this method is called as supposed.
The last item is cancellable callback that is not currently supported:
https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap#public-final-void-animatecamera-cameraupdate-update,-int-durationms,-googlemap.cancelablecallback-callback

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_flutter_web.dart
ditman: I think this is fine, that way we don't accumulate too much technical debt from the web implementation, or have too many `skip: kIsWeb` tests in the app-facing package! Thanks!

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_inspector_platform.dart
```diff
@@ -137,4 +137,13 @@ abstract class GoogleMapsInspectorPlatform extends PlatformInterface {
       {required int mapId, required ClusterManagerId clusterManagerId}) {
     throw UnimplementedError('getClusters() has not been implemented.');
   }
+
+  /// Returns current camera position.
```
jokerttu, line 141: @ditman should I add following method here and use it in tests to see if getting camera position is supported on the platform.
Then I can drop the changes for web platform from this PR and everything on it's separate PR later after problems with it are solved?

```dart
  /// If the platform supports getting camera position
  bool supportsGettingGameraPosition() => false;
```

stuartmorgan: That sounds good; pretty much any time we can isolate platform-specific limitations behind a generic support query it makes the rest of the code more maintainable.

jokerttu: Done at https://github.com/flutter/packages/pull/7648/commits/f4a39ce081257bcbfe2d38e9ffa637023c2f318e

jokerttu:
As web platform changes are now dropped, this still needs review/approval from
Android: @reidbaker https://github.com/flutter/packages/blob/main/CODEOWNERS#L78
iOS: @cbracken https://github.com/flutter/packages/blob/main/CODEOWNERS#L95
App facing package: @stuartmorgan

camsim99 (approved):
Sorry I just commented before. Android LGTM! :)

stuartmorgan (commented):
I went ahead and did the iOS review, and it LGTM; you should be good to split out sub PRs now.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController_Test.h
```diff
@@ -4,9 +4,17 @@
 
 #import <Flutter/Flutter.h>
 #import <GoogleMaps/GoogleMaps.h>
+#import "FGMCATransactionWrapper.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface FGMMapCallHandler : NSObject <FGMMapsApi>
```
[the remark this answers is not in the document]
jokerttu: @stuartmorgan please check following commit if this addresses this issue.
https://github.com/flutter/packages/pull/7648/commits/d235debbf2bbdac667a3571e5db561f882771587

Commit exposes `FGMMapCallHandler` in `GoogleMapController.h` that allows using `(Test)` extension in `GoogleMapController_Test.h`.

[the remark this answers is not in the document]
stuartmorgan: Yes, thanks.

jokerttu:
There is now splitted PR for platform interface implementation at https://github.com/flutter/packages/pull/8596

stuartmorgan (approved):
LGTM!

PR #7652, opened 2024-09-16
Revert "[webview_flutter] Improve flaky scroll tests"

stuartmorgan:
Reverts flutter/packages#7621

The re-enabled tests are failing in post-submit.

Alternately, instead of reverting we could just re-add the skips with the link, if we still want to keep the updated version as a better foundation to work from.


bparrishMines (approved):
LGTM. I'm fine with just reverting.

tarrinneal:
Landing on red to fix the tree?

PR #7657, opened 2024-09-17
[video_player_avfoundation] fix playback speed resetting

misos1:
Calling play is the same as setting the rate to 1.0 (or to `defaultRate` depending on ios version, *documentation in this first link is clearly not updated because it does not mention `defaultRate`*):

https://developer.apple.com/documentation/avfoundation/avplayer/1386726-play?language=objc
https://developer.apple.com/documentation/avfoundation/avplayer/3929373-defaultrate?language=objc

*The second link contains a note about not starting playback by setting the rate to 1.0. I assume this is because of the introduction of `defaultRate` (which can be different than 1.0) and not because `play` may do something more than just setting `rate` as that wording is explicit about setting rate to 1.0, it says nothing about any other value.*

This is also why https://github.com/flutter/plugins/pull/4331 did not work well. It was setting `rate` to 1.0 (through `play`) then immediately to the value set by `setPlaybackSpeed`. One of them or both caused change to `playbackLikelyToKeepUp` which triggered `observeValueForKeyPath` with `playbackLikelyToKeepUpContext` which in turn called again `updatePlayingState` where was `rate` again changed first to 1.0 then to another value and so on. In this PR the rate is changed only once and then to the same value, seems when `rate` is assigned but not really changed it does not trigger anything.

In issues below `seekTo` can trigger `playbackLikelyToKeepUp` change which will call `updatePlayingState` and reset `rate` to 1.0 through `play`. When the speed is set in dart before initialization then the dart side will set it right after calling `play` but even some time after initialization there is still a flood of events calling `updatePlayingState` and it is likely that some of them will call it after `setPlaybackSpeed`. Actually even when `setPlaybackSpeed` was not called on dart side it (dart side) will always change speed after play to 1.0 so it ignores this new `defaultRate` feature.

- fixes https://github.com/flutter/flutter/issues/71264
- fixes https://github.com/flutter/flutter/issues/73643

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -397,7 +399,15 @@ - (void)updatePlayingState {
...
   if (_isPlaying) {
-    [_player play];
+    // Calling play is the same as setting the rate to 1.0 (or to defaultRate depending on ios
+    // version) so last set playback speed must be set here if any instead.
+    // https://github.com/flutter/flutter/issues/71264
+    // https://github.com/flutter/flutter/issues/73643
+    if (_playbackSpeed) {
+      [self updateRate];
+    } else {
+      [_player play];
```
stuartmorgan, line 409: Wouldn't the codepaths be simpler at this point if we just always used `rate` instead of `play`/`pause`?

misos1: That is to potentially support `defaultRate` when `_playbackSpeed` is unset, there can be new function like `setDefaultPlaybackSpeed` and support from dart side, or maybe speed set by `setPlaybackSpeed` can be relative to `defaultRate` but that would require checking version of ios.

stuartmorgan: Oh, I see. Sorry, I had forgotten about the part of the PR description explaining that the official docs for `play`/`pause` are actually wrong now.

stuartmorgan:
Sorry, this fell off my review queue. @hellohuanlin can you take a look as well?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -406,6 +416,29 @@ - (void)updatePlayingState {
...
+- (void)updateRate {
+  // See https://developer.apple.com/library/archive/qa/qa1772/_index.html for an explanation of
+  // these checks.
+  // If status is not AVPlayerItemStatusReadyToPlay then both canPlayFastForward
+  // and canPlaySlowForward are always false and it is unknown whether video can
+  // be played at these speeds, updatePlayingState will be called again when
+  // status changes to AVPlayerItemStatusReadyToPlay.
+  float speed = _playbackSpeed.floatValue;
+  if (speed > 2.0 && !_player.currentItem.canPlayFastForward) {
+    if (_player.currentItem.status != AVPlayerItemStatusReadyToPlay) {
```
hellohuanlin, line 428: what if we call `updateRate` with speed < 2.0? should we also check the status here? Should this status check be moved to top of the fucntion?

misos1: That check is to confirm whether `canPlay...` returned false because it cannot be played at speed>2 instead of because status was not ready to play. There is no need to check status for speeds for which is not needed `canPlay...`. But as I am thinking about it now maybe status should be fetched before `canPlay...` to avoid a false positive when status becomes ready right after `canPlay...`.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -406,6 +416,30 @@ - (void)updatePlayingState {
...
 
+- (void)updateRate {
+  // See https://developer.apple.com/library/archive/qa/qa1772/_index.html for an explanation of
+  // these checks.
+  // If status is not AVPlayerItemStatusReadyToPlay then both canPlayFastForward
+  // and canPlaySlowForward are always false and it is unknown whether video can
+  // be played at these speeds, updatePlayingState will be called again when
+  // status changes to AVPlayerItemStatusReadyToPlay.
+  float speed = _playbackSpeed.floatValue;
+  bool readyToPlay = _player.currentItem.status == AVPlayerItemStatusReadyToPlay;
```
stuartmorgan, line 427: `BOOL` since this is Obj-C code.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -82,6 +82,8 @@ @interface FVPVideoPlayer ()
 @property(nonatomic) CGAffineTransform preferredTransform;
 @property(nonatomic, readonly) BOOL disposed;
 @property(nonatomic, readonly) BOOL isPlaying;
+// Playback speed when video is playing.
+@property(nonatomic, readonly) NSNumber *playbackSpeed;
```
stuartmorgan, line 86: It would be more accurate to call this `targetPlaybackSpeed` (and document it as something like "The target playback speed requested by the plugin client."), since it's entirely possible for the playback speed to *not* be this value while playing.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -406,6 +416,30 @@ - (void)updatePlayingState {
   _displayLink.running = _isPlaying || self.waitingForFrame;
 }
 
+- (void)updateRate {
```
stuartmorgan, line 419: This needs a declaration comment, per the style guide.

stuartmorgan (changes requested):
The logic and tests all LGTM, just small comments on naming and style guide conformance.

cbracken:
@misos1 are you still working on this patch?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
misos1: Like for example? I cannot see any such comment above any other such function except maybe `disposeSansEventChannel`. What could be put here, something like "Sets the rate of the player."?

misos1:
> @misos1 are you still working on this patch?

Yes.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
stuartmorgan: > I cannot see any such comment above any other such function except maybe `disposeSansEventChannel`.

Unfortunately the reviews for a lot of our older plugin code didn't enforce [the style guide](https://google.github.io/styleguide/objcguide.html#declaration-comments). You're certainly welcome to fix other methods if you would like to, but since you didn't add those methods they weren't part of my review.

> What could be put here, something like "Sets the rate of the player."?

A meaningful description of what the method does. For instance: "Synchronizes the `player`'s playback rate with `playbackSpeed`, constrained by the playback rate capabilities of the player."

stuartmorgan (approved):
LGTM!

hellohuanlin (commented):
Sorry I realized that I wrote the comments a long time ago but didn't actually hit the submit button!

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -406,6 +416,30 @@ - (void)updatePlayingState {
...
+  // See https://developer.apple.com/library/archive/qa/qa1772/_index.html for an explanation of
+  // these checks.
+  // If status is not AVPlayerItemStatusReadyToPlay then both canPlayFastForward
+  // and canPlaySlowForward are always false and it is unknown whether video can
+  // be played at these speeds, updatePlayingState will be called again when
+  // status changes to AVPlayerItemStatusReadyToPlay.
+  float speed = _playbackSpeed.floatValue;
+  bool readyToPlay = _player.currentItem.status == AVPlayerItemStatusReadyToPlay;
+  if (speed > 2.0 && !_player.currentItem.canPlayFastForward) {
+    if (!readyToPlay) {
```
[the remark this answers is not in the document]
misos1: > What if readyToPlay is false, and we set the speed to a value between 1.0 and 2.0?

Everything is fine then, the rate can be set anytime.

> Why are we allowing setting the speed for this value but not setting the speed for clamped value?

To avoid setting it to a value which was not requested and seems each change to rate triggers re-buffering and `playbackLikelyToKeepUp` as is stated in description.

> I'm still confused why this check can't be moved to the top of the function?

I chose to do it strictly only when needed. So you probably argue to do it always due to simplicity maybe. But I believe setting up the rate before `ReadyToPlay` may avoid some re-buffering as mentioned above.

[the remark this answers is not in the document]
stuartmorgan: > Why are we allowing setting the speed for this value but not setting the speed for clamped value?

The problem isn't that we can't set the speed while buffering, it's that we don't know whether we need to clamp while buffering. But for 1.0-2.0 it's irrelevant because clamping wouldn't apply anyway.

(There is an unfortunate side-effect of this PR in that setting at unsupported value now silently clamps instead of throwing an exception, adding another potential way for native and Dart to get out of sync about state since Dart will think it's successfully set the out-of-range value, but we can address that edge case later with more event channel communication.)

hellohuanlin (approved):

stuartmorgan:
This will just need conflicts resolved against the code location refactoring, and then it can be landed!

misos1:
Trying to reopen...

hellohuanlin:
I can't re-open it either. You may have to create a new PR.

stuartmorgan:
This happened once before with another PR; I was never able to figure out what caused GitHub to put it into this state, or why :(

misos1:
Nice, after I pushed new changes it was possible to reopen.

Now I am planning to check that problem with `defaultRate` on ios 16 (as I now have enough info about it), so please wait.

misos1:
PR #8274 states that `rate` is set to `defaultRate` on seeking (https://github.com/flutter/packages/pull/8274#discussion_r1880779157) and references a link to documentation about `defaultRate`, but the provided link does not say anything like that and I cannot reproduce such behavior either (and it also sounds really unlikely). More likely is that what was observed by the author is what is explained in the description of this PR, seeking triggers re-buffering which triggers `observeValueForKeyPath` with `playbackLikelyToKeepUpContext` which calls `updatePlayingState` which calls `AVPlayer.play` which sets `rate` to `defaultRate` as is stated in the mentioned link.

sha3rawi33:
> PR #8274 states that `rate` is set to `defaultRate` on seeking ([#8274 (comment)](https://github.com/flutter/packages/pull/8274#discussion_r1880779157)) and references a link to documentation about `defaultRate`, but the provided link does not say anything like that and I cannot reproduce such behavior either (and it also sounds really unlikely). More likely is that what was observed by the author is what is explained in the description of this PR, seeking triggers re-buffering which triggers `observeValueForKeyPath` with `playbackLikelyToKeepUpContext` which calls `updatePlayingState` which calls `AVPlayer.play` which sets `rate` to `defaultRate` as is stated in the mentioned link.

Yeah, that's exactly what is causing the issue to arise; and setting `defaultRate` to the preferred playback rate prevents it from happening. I tested this on my apps on production and all went well. (Also native tests pass)
Also please note that it is mentioned in the link I provided in my PR. See the attached image.
it says: 
`
The next time you call the play method, the player restores the rate to the value of defaultRate.
`
Which happens on seeking or rebuffering causing the issue. 
image

misos1:
> Also please note that it is mentioned in the link I provided in my PR. See the attached image.
> it says:
> `The next time you call the play method, the player restores the rate to the value of defaultRate.`
> Which happens on seeking or rebuffering causing the issue.

Where does that documentation say that `play` is called (or `rate` is set to `defaultRate`) internally (meaning not through `FVPVideoPlayer`) on seeking or rebuffering? Where in that link is that mentioned?

stuartmorgan:
> Now I am planning to check that problem with `defaultRate` on ios 16 (as I now have more info about it), so please wait.

Since this PR improves over the current code, and is fully reviewed, I suggest that we land it as-is and then if there are remaining issues address them in a follow-up. @misos1 Any concerns with that approch?

stuartmorgan:
That failure was flake that I've seen outside of this PR, so I believe to be unrelated.

PR #7699, opened 2024-09-25
[google_maps_flutter_platform_interface] Convert `BitmapDescriptor` to typesafe subclasses

yaakovschectman:
`BitmapDescriptor` is currently just a wrapper around JSON, with the exception of two of its subclasses. This converts all cases to typesafe structures without breaking the current API.

Part of https://github.com/flutter/flutter/issues/155122

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
@@ -76,7 +76,9 @@ class BitmapDescriptor {
         if (jsonList.length == 3) {
           assert(jsonList[2] != null && jsonList[2] is String);
           assert((jsonList[2] as String).isNotEmpty);
+          return AssetBitmap(name: jsonList[1] as String, package: jsonList[2] as String);
         }
+        return AssetBitmap(name: jsonList[1] as String);
       case _fromAssetImage:
         assert(jsonList.length <= 4);
         assert(jsonList[1] != null && jsonList[1] is String);
@@ -329,6 +331,16 @@ class BytesBitmap extends BitmapDescriptor {
   Object toJson() {
     return <Object>[BitmapDescriptor._fromBytes, byteData, if (size != null && kIsWeb) <Object>[size!.width, size!.height]];
   }
+}
+
+class AssetBitmap extends BitmapDescriptor {
+  const AssetBitmap({required this.name, this.package}) : super._(const <Object>[]);
+
+  final String name;
+  final String? package;
+
+  @override
+  Object toJson() => <Object>[BitmapDescriptor._fromAsset, name, if (package != null) package!];
 }
 
 /// Represents a [BitmapDescriptor] base class for map bitmaps.
--- a/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
@@ -30,10 +30,10 @@ enum MapBitmapScaling {
 }
 
 MapBitmapScaling mapBitmapScalingFromString(String mode) => switch (mode) {
-    'auto' => MapBitmapScaling.auto,
-    'none' => MapBitmapScaling.none,
-    _ => throw ArgumentError('Unrecognized MapBitmapScaling $mode', 'mode'),
-  };
+      'auto' => MapBitmapScaling.auto,
+      'none' => MapBitmapScaling.none,
+      _ => throw ArgumentError('Unrecognized MapBitmapScaling $mode', 'mode'),
+    };
 
 // The default pixel ratio for custom bitmaps.
 const double _naturalPixelRatio = 1.0;
@@ -82,7 +82,8 @@ abstract class BitmapDescriptor {
         if (jsonList.length == 3) {
           assert(jsonList[2] != null && jsonList[2] is String);
           assert((jsonList[2] as String).isNotEmpty);
-          return AssetBitmap(name: jsonList[1] as String, package: jsonList[2] as String);
+          return AssetBitmap(
+              name: jsonList[1] as String, package: jsonList[2] as String);
         }
         return AssetBitmap(name: jsonList[1] as String);
       case _fromAssetImage:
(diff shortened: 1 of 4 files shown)


packages/google_maps_flutter/google_maps_flutter_platform_interface/test/types/bitmap_test.dart
```diff
@@ -21,10 +21,17 @@ void main() {
           BitmapDescriptor.fromJson(json);
 
       expect(descriptorFromJson, isNot(descriptor)); // New instance
-      expect(identical(descriptorFromJson.toJson(), json), isTrue); // Same JSON
+      expect(descriptorFromJson.toJson(), json);
```
yaakovschectman, line 24: Since the bitmaps are no longer just wrapping JSON, `toJson` will not return the same reference as was passed to `fromJson`, though they will still be equal by comparison.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
```diff
@@ -44,14 +51,14 @@ const double _naturalPixelRatio = 1.0;
 /// a default marker icon.
 /// Use the [BitmapDescriptor.defaultMarkerWithHue] to create a
 /// [BitmapDescriptor] for a default marker icon with a hue value.
-class BitmapDescriptor {
-  const BitmapDescriptor._(this._json);
+abstract class BitmapDescriptor {
```
reidbaker, line 54: I thought this was a breaking change but a spot check of the public usages in github show I am incorrect.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
```diff
@@ -44,14 +51,14 @@ const double _naturalPixelRatio = 1.0;
...
+  const BitmapDescriptor._();
 
   /// The inverse of .toJson.
   // TODO(stuartmorgan): Remove this in the next breaking change.
   @Deprecated('No longer supported')
-  BitmapDescriptor.fromJson(Object json) : _json = json {
-    assert(_json is List<dynamic>);
+  static BitmapDescriptor fromJson(Object json) {
```
reidbaker, line 60: Why not use the `: _json` syntax?

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
yaakovschectman: This PR removes the `_json` field from `BitmapDescriptor`, and since it also makes `BitmapDescriptor` an abstract class with a concrete subtype for each possible set of member data, `fromJson` needs to return an instance of one of those subtypes.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
yaakovschectman: The only ways to construct a `BitmapDescriptor` without constructing one of its subclasses was either through this private `_` factory, which is removed here, or the below `fromJson` public factory, which this PR converts to a static method, so unless we are expecting public usages of this package to add new subclasses of `BitmapDescriptor`, I believe that neither making it abstract nor removing the `_` factory should impact any public uses.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
```diff
@@ -29,6 +29,13 @@ enum MapBitmapScaling {
   none,
 }
 
+/// Convert a string from provided JSON to a MapBitmapScaling enum.
+MapBitmapScaling mapBitmapScalingFromString(String mode) => switch (mode) {
```
stuartmorgan, line 33: This should be private so we're not increasing the API surface of the package.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
```diff
@@ -306,10 +329,100 @@ class BitmapDescriptor {
...
-
   /// Convert the object to a Json format.
-  Object toJson() => _json;
+  Object toJson();
+}
+
+/// A BitmapDescriptor using the default marker.
+class DefaultMarker extends BitmapDescriptor {
+  /// Provide an optional [hue] for the default marker.
+  const DefaultMarker([this.hue]) : super._();
```
stuartmorgan, line 339: Please use named parameters so that we can change the parameter set in the future without breaking changes (or being stuck with a growing positional list, which doesn't scale).

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
yaakovschectman: Would you prefer we mark it `@visibleForTesting`, or remove the test https://github.com/flutter/packages/pull/7699/files#diff-a8066076d538f1e1e21b27bac466302cb9317b643500ce81b991102a94611070R685?

reidbaker: My preference would be visible for testing and keep the test.

yaakovschectman: That would be my thought as well, @stuartmorgan do you agree with this decision?

stuartmorgan: Yes, that's fine; from a semver standpoint we consider those equivalent.

stuartmorgan (commented):
LG once the `@visibleForTesting` annotation is added.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
```diff
@@ -5,6 +5,7 @@
 import 'dart:async' show Future;
 import 'dart:typed_data' show Uint8List;
 
+import 'package:flutter/cupertino.dart';
```
stuartmorgan, line 8: Looks like your IDE chose a weird auto-import here; this should be a new `show` below instead.

stuartmorgan: (Specifically, from `foundation.dart`)

stuartmorgan (approved):
LGTM

PR #7701, opened 2024-09-25
Update test to not depend on `Color.toString()`

chingjun:
In https://github.com/flutter/engine/pull/55231, the implementation of `Color.toString()` has changed. Update the test to not depend on the exact implementation.


gaaclarke (approved):
lgtm

PR #7703, opened 2024-09-25
[google_maps_flutter_platform_interface] Convert `PatternItem` and `Cap` to typesafe structures.

yaakovschectman:
Convert `PatternItem` and `Cap` from JSON wrappers to typesafe classes.

https://github.com/flutter/flutter/issues/155121

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
```diff
@@ -44,11 +71,34 @@ class Cap {
...
+/// CustomCap with a bitmap overlay centered at the start or
+/// end vertex of a [Polyline], orientated according to the direction of the line's
+/// first or last edge and scaled with respect to the line's stroke width.
+class CustomCap extends Cap {
+  /// [bitmapDescriptor] must not be null.
+  ///
+  /// [refWidth] is the reference stroke width (in pixels) - the stroke width for which
+  /// the cap bitmap at its native dimension is designed. Must be positive. Default value
+  /// is 10 pixels.
+  const CustomCap(this.bitmapDescriptor, [this.refWidth = 10])
```
stuartmorgan, line 92: Given that this is a public API we will ~never be able to change, let's make it a named parameter rather than an optional positional parameter, as named parameters can much more easily be added to later.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
```diff
@@ -6,27 +6,54 @@ import 'package:flutter/foundation.dart' show immutable;
 
 import 'types.dart';
 
+/// Enumeration of possible types of caps.
+enum CapType {
+  /// Cap that is squared off exactly at the start or end vertex of a [Polyline]
+  /// with solid stroke pattern, equivalent to having no additional cap beyond
+  /// the start or end vertex.
+  butt('buttCap'),
```
stuartmorgan, line 14: Since the only use case for the strings is `toJson`, and we want to discourage people relying on strings in the interface of this package, I would much prefer that these be completely internal to the implementation of `toJson` rather than public API, so we aren't effectively expanding our legacy API surface.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/pattern_item.dart
```diff
@@ -4,32 +4,50 @@
 
 import 'package:flutter/foundation.dart' show immutable;
 
+/// Enumeration of types of pattern items.
+enum PatternItemType {
+  /// A dot used in the stroke pattern for a [Polyline].
+  dot('dot'),
```
stuartmorgan, line 10: Same note here.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/pattern_item.dart
```diff
@@ -4,32 +4,50 @@
...
   /// [length] has to be non-negative.
   static PatternItem gap(double length) {
     assert(length >= 0.0);
-    return PatternItem._(<Object>['gap', length]);
+    return PatternItem._(PatternItemType.gap, length);
   }
 
-  final Object _json;
+  final PatternItemType _type;
+  final double? _length;
```
stuartmorgan, line 49: It seems like we'll need subclasses that expose `length` publicly, using a similar structure to your cap changes, otherwise the platform implementation code would still need to call and interpret `toJson` to get the required info.

packages/google_maps_flutter/google_maps_flutter_platform_interface/test/types/cap_test.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:google_maps_flutter_platform_interface/google_maps_flutter_platform_interface.dart';
+
+void main() {
```
stuartmorgan, line 8: Just to make sure we don't accidentally regress anything during this transition: have you double-checked that all of these tests pass with *and without* your code changes?

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter_platform_interface/test/types/cap_test.dart
yaakovschectman: I've confirmed this is true.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
```diff
@@ -6,27 +6,56 @@ import 'package:flutter/foundation.dart' show immutable;
...
+  /// first or last edge and scaled with respect to the line's stroke width.
+  custom,
+}
+
+String _capTypeToJson(CapType capType) => <CapType, String>{
+      CapType.butt: 'buttCap',
+      CapType.round: 'roundCap',
+      CapType.square: 'squareCap',
+      CapType.custom: 'customCap',
+    }[capType]!;
```
stuartmorgan, line 36: Please use a switch instead; a switch would fail `analyze` if a new enum value were added, while this would not, making this version harder to maintain.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/pattern_item.dart
```diff
@@ -4,32 +4,58 @@
...
+  /// A gap used in the stroke pattern for a [Polyline].
+  gap,
+}
+
+String _patternItemTypeToJson(PatternItemType itemType) =>
+    const <PatternItemType, String>{
+      PatternItemType.dot: 'dot',
+      PatternItemType.dash: 'dash',
+      PatternItemType.gap: 'gap',
+    }[itemType]!;
```
stuartmorgan, line 24: Same here.

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
yaakovschectman: I had not considered that, I will make the change and try to keep that in mind going forward. Thanks.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/pattern_item.dart
```diff
@@ -4,32 +4,57 @@
...
-    return PatternItem._(<Object>['gap', length]);
+    return PatternItem._(PatternItemType.gap, length);
   }
 
-  final Object _json;
+  /// The type of rendering used for an item in a pattern.
+  final PatternItemType patternItemType;
+
+  /// The length in pixels of a dash or gap.
+  final double? length;
```
stuartmorgan, line 53: Thinking about the future usage here again, don't we need different subclasses for whether or not there's a required length? With this version, consuming code is going to have to force-unwrap `length` for certain pattern types, which type safety was lost along the way (since length was a non-optional and non-nullable param when the pattern was created.)

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/pattern_item.dart
```diff
@@ -4,32 +4,73 @@
...
+
+  /// Converts this object to something serializable in JSON.
+  Object toJson() => <Object>[
+        _patternItemTypeToJson(patternItemType),
+      ];
+}
+
+/// A pattern item with a length, i.e. a dash or gap.
+@immutable
+class PatternItemWithLength extends PatternItem {
```
stuartmorgan, line 62: How about `VariableLengthPatternItem`.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
```diff
@@ -44,11 +73,36 @@ class Cap {
...
   }) {
     assert(refWidth > 0.0);
-    return Cap._(<Object>['customCap', bitmapDescriptor.toJson(), refWidth]);
+    return CustomCap(bitmapDescriptor, refWidth: refWidth);
   }
 
-  final Object _json;
+  /// The type of rendering used for the cap at a start or end vertex of a
+  /// [Polyline].
+  final CapType capType;
```
stuartmorgan, line 81: Sorry, I missed this before; this should just be called `type`. The fact that it's a cap type is inherent in the enclosing class.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/pattern_item.dart
```diff
@@ -4,32 +4,73 @@
...
   static PatternItem gap(double length) {
     assert(length >= 0.0);
-    return PatternItem._(<Object>['gap', length]);
+    return PatternItemWithLength._(
+        patternItemType: PatternItemType.gap, length: length);
   }
 
-  final Object _json;
+  /// The type of rendering used for an item in a pattern.
+  final PatternItemType patternItemType;
```
stuartmorgan, line 52: Same here; this should just be `type`.

stuartmorgan (changes requested):
A couple more minor notes, but mostly looks good.

If we were writing this from scratch I would say we should make an `abstract sealed` base class and then subclasses for each type, rather than using enums to distinguish, but technically adding `sealed` to the existing classes would be a breaking change (even though I *really* doubt it would be breaking), and I don't think the different for clients (being able to directly pattern-switch instead of doing an enum switch + casting, which is not as type-safe) is worth a breaking change to a platform interface package.

stuartmorgan (approved):
LGTM!

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
```diff
@@ -6,27 +6,56 @@ import 'package:flutter/foundation.dart' show immutable;
...
+  /// start or end vertex of a [Polyline] with solid stroke pattern.
+  square,
+
+  /// CustomCap with a bitmap overlay centered at the start or
+  /// end vertex of a [Polyline], orientated according to the direction of the line's
+  /// first or last edge and scaled with respect to the line's stroke width.
+  custom,
+}
+
+String _capTypeToJson(CapType capType) => switch (capType) {
```
reidbaker, line 31: If there is documentation for where these json values come from please link it here.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
```diff
@@ -44,11 +73,36 @@ class Cap {
...
+/// CustomCap with a bitmap overlay centered at the start or
+/// end vertex of a [Polyline], orientated according to the direction of the line's
+/// first or last edge and scaled with respect to the line's stroke width.
+class CustomCap extends Cap {
+  /// [bitmapDescriptor] must not be null.
+  ///
+  /// [refWidth] is the reference stroke width (in pixels) - the stroke width for which
+  /// the cap bitmap at its native dimension is designed. Must be positive. Default value
+  /// is 10 pixels.
+  const CustomCap(this.bitmapDescriptor, {this.refWidth = 10})
```
reidbaker, line 96: Is it documented if these are logical pixels or physical pixels? If so please specify.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/pattern_item.dart
```diff
@@ -4,32 +4,73 @@
...
+  dot,
+
+  /// A dash used in the stroke pattern for a [Polyline].
+  dash,
+
+  /// A gap used in the stroke pattern for a [Polyline].
+  gap,
+}
+
+String _patternItemTypeToJson(PatternItemType itemType) => switch (itemType) {
```
reidbaker, line 19: Same here if the values are documented

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
yaakovschectman: These are the old string values used before this PR's diff which are then used the platform packages' JSON-based approach. They do not correspond to any values outside this repo, e.g. the Maps SDK.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
yaakovschectman: A few redirections through the Maps SDK documentation describes these as "screen pixels." I will update the comment to include a link to the page.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
reidbaker: Perfect that means if we change them then nothing bad happens.

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/cap.dart
stuartmorgan: > Perfect that means if we change them then nothing bad happens.

No, that would be extremely bad. It would break all older versions of `google_maps_flutter`.

We're still cleaning up from tech debt mistakes in federating this plugin, where during federation magic string values that used to be internal to the monolithic plugin package were separated, with the Dart code producing them moved to `google_maps_flutter_platform_interface`, while the native code consuming them was left in the main package (and then eventually moved to platform implementation packages).

What *should* have happened was that we stopped using `toJson` from `platform_interface` as the serialization mechanism for the platform channels, in favor of package-internal versions. But since we didn't, these string values are effectively part of the public API surface of this package, consumed by our own implementations, with which we have to maintain compatibility.

reidbaker: Ah ok so these strings while not needing to mirror something external like the google maps sdk they do need to mirror something external to the package `google_maps_flutter_platform_interface` for cross version compatibility. 

If we changed the values then set a new minimum for each of the platform implementations would that prevent a breakage? I am not asking because I want to do that but because I want to understand the relationships. 

The reason why I was asking what they needed to align to was to add documentation to warn future maintainers what those values needed to align with.

stuartmorgan: > If we changed the values then set a new minimum for each of the platform implementations would that prevent a breakage?

No, because the versions of the platform implementations that call these `toJson` functions (and rely on the result matching what's in the native code) already exist, and people are using them. And those packages use standard `^2.x.y` constraints on `google_maps_flutter_platform_interface` and rely on the fact that doing so is safe because a change that breaks them wouldn't be made in a 2.x version.

Releasing a new `google_maps_flutter_android` would mean that people using the **new** `google_maps_flutter_android` would only get a compatible version of `google_maps_flutter_platform_interface`, but it would not stop people using an **old** version of `google_maps_flutter_android` from getting a new and incompatible version of `google_maps_flutter_platform_interface`.

> I am not asking because I want to do that but because I want to understand the relationships.

I think the easiest mental model is: because of the federation mistake, the exact structure and string values returned by these `toJson` functions are now, in practice, part of the official API surface of the method, and so changing any existing details of what it returns would be just like any incompatible change to the documented behavior of any public method: a breaking change requiring a major version bump.

From a docs standpoint, explicitly documenting what the return value structure is in the dart doc would accurately express the unfortunate state we are in now.

PR #7706, opened 2024-09-25
[draft] [go_router] wrap path_utils content in RoutePattern

cedvdb:
part of https://github.com/flutter/flutter/issues/155567

This is a draft PR to see if some part of go_router can be refactored to be more readable.
Currently it is hard to know if a parameter / variable is a pattern such as `/books/:bookId` or a path such as `/book/123` when reading the code of a given function. This PR is a stab on improving the readability on that front.

It wraps the content of path_utils in a `RoutePattern` class.

@chunhtai this does not compile but I'm interested in knowing whether this is a direction the team wants to go or not, I'll just close this if not.

I'll probably wait for my other PR's to go through before going any further anyway.

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/go_router/lib/src/configuration.dart
```diff
@@ -264,28 +265,26 @@ class RouteConfiguration {
...
+    final GoRoute route = _routesByName[name]!;
     assert(() {
-      // Check that all required params are present
-      final List<String> paramNames = <String>[];
-      patternToRegExp(path, paramNames);
-      for (final String paramName in paramNames) {
+      for (final String paramName in route.pattern.parameters) {
         assert(pathParameters.containsKey(paramName),
-            'missing param "$paramName" for $path');
+            'missing param "$paramName" for ${route.pattern}');
```
cedvdb, line 273: `RoutePattern` overrides toString

packages/go_router/lib/src/configuration.dart
```diff
@@ -503,12 +502,45 @@ class RouteConfiguration {
         .join(' => ');
   }
 
-  /// Get the location for the provided route.
-  ///
-  /// Builds the absolute path for the route, by concatenating the paths of the
-  /// route and all its ancestors.
-  String? locationForRoute(RouteBase route) =>
-      fullPathForRoute(route, '', _routingConfig.value.routes);
+  /// Concatenate a Route's pattern with all its ancestor patterns
```
cedvdb, line 505: (tree walk to find the path)

packages/go_router/lib/src/configuration.dart
```diff
@@ -503,12 +502,45 @@ class RouteConfiguration {
         .join(' => ');
   }
 
-  /// Get the location for the provided route.
-  ///
-  /// Builds the absolute path for the route, by concatenating the paths of the
-  /// route and all its ancestors.
-  String? locationForRoute(RouteBase route) =>
```
cedvdb, line 510: I believe this was supposed to be internal and ok to be renamed.

packages/go_router/lib/src/route.dart
```diff
@@ -432,16 +430,15 @@ class GoRoute extends RouteBase {
...
 
   /// Extract the path parameters from a match.
   Map<String, String> extractPathParams(RegExpMatch match) =>
-      extractPathParameters(pathParameters, match);
+      pattern.extractPathParameters(match);
 
   /// The path parameters in this route.
   @internal
-  final List<String> pathParameters = <String>[];
+  List<String> get pathParameters => pattern.parameters;
```
cedvdb, line 441: Not sure if the pattern should be accessed directly (`route.pattern.parameters`) or if the route should facade to the pattern.

packages/go_router/lib/src/configuration.dart
chunhtai: this may be a breaking change, this could be an easy migration if you write rules fix_data.yaml

Also in this library, `path` means `/book/:bookId`, the filled string like `book/123` is called `location`.

packages/go_router/lib/src/route.dart
chunhtai: if we expose the pattern, than they should probably be accessed directly to avoid duplication

chunhtai: do we want to expose pattern though?

chunhtai (commented):
wrapping path_utils does make it more readable. but not sure about changing the naming convention is worthing the breaking change.

cedvdb:
> wrapping path_utils does make it more readable. but not sure about changing the naming convention is worthing the breaking change.

I believe it's possible to make it only internal without breaking changes, that's the requirement I'm doing this under

chunhtai:
if we want to rename

path -> pattern
location -> path

I would like us to change the public facing API such as
GoRoute.path -> GoRoute.pattern etc.

but i am not sure if this worth it.

As for wrapping path_util, I think we can do it as less as breaking change as possible, and I am onboard with this.

packages/go_router/lib/src/route.dart
cedvdb: Currently no, the idea was to have `RoutePattern` used internally, (I have to make it internal in the future) and have the public representation still `String path`, ie: `GoRoute(path: '/some-path')`.  The `RoutePattern` is supposed to be used internally everywhere so we don't end up with `String path`, where we are unsure whether path is `/books/:bookId` or `/books/123` without looking at the wider context.

However I'm not a fan of this kind of constructor conversion, it's a bit harder to reason about.

chunhtai: in that case the RoutePattern should be kept in as private variable in GoRoute class.

cedvdb:
@chunhtai So the convention used is that `path` is a `/books/:bookId` and `location` is `/books/123` ? I had a feeling that was the case but wasn't sure. I don't think renaming is necessary.

chunhtai:
> @chunhtai So the convention used is that `path` is a `/books/:bookId` and `location` is `/books/123` ? I had a feeling that was the case but wasn't sure. I don't think renaming is necessary.

yes. that is the current naming convension.

chunhtai:
(triage) Hi @cedvdb are you still planning on coming back to this pr?

cedvdb:
@chunhtai I don't think I will, I do not have the bandwidth for a PR like this

chunhtai:
Thanks for reply, I will close this pr for now, feel free to reopen if  you have time in the future

PR #7709, opened 2024-09-26
[pigeon] update deprecated command in README of the example

EchoEllet:
*Replace a deprecated command in the instructions of using `pigeon`.*

Running `flutter pub run pigeon --input path/to/input.dart` will output the following at the start of the output:

```console
Deprecated. Use `dart run` instead.
```

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/pigeon/example/README.md
+++ b/packages/pigeon/example/README.md
@@ -40,7 +40,7 @@ needed for your project.
 Then make a simple call to run pigeon on the Dart file containing your definitions.
 
 ```sh
-flutter pub run pigeon --input path/to/input.dart
+dart run pigeon --input path/to/input.dart
 ```
 
 ## HostApi Example


tarrinneal:
Good catch! This will need an update to the changelog, I think a `NEXT` entry would be fine for this. @stuartmorgan is that the case for this? And tagged for second review once that is done.

EchoEllet:
Thank you for reviewing the change, IMO `CHANGELOG.md` is for the end user or the developer who will use the package, such minor changes is not in the list of interest for most users, and I think `CHANGELOG.md` should be different from Git log or the commit history. Such projects have many of those changes and having them all in a `CHANGELOG.md` will result in users will not read them even if it has critical changes, features, deprecation, or breaking changes. User expect that docs up to to date with related repositories. See [Keep a Changelog
](https://keepachangelog.com/en/1.1.0/) as an example format for `CHANGELOG.md`.

Though I can add it if needed.

tarrinneal:
> Thank you for reviewing the change, IMO `CHANGELOG.md` is for the end user or the developer who will use the package, such minor changes are not in the list of interest for most users, and I think `CHANGELOG.md` should be different from Git log or the commit history. Such projects have many of those changes and having them all in a `CHANGELOG.md` will result in users not reading them even if it has critical changes, features, deprecation, or breaking changes. User expect that docs up to to date with related repositories. See [Keep a Changelog ](https://keepachangelog.com/en/1.1.0/) as an example format for `CHANGELOG.md`.
> 
> Though I can add it if needed.

The issue is that README changes don't get published without a new version, I could wrap this change up in another pr I'm making to avoid the issue, or @stuartmorgan could grant an exception for this case.

EchoEllet:
if they can grant an exception I would wait, otherwise, If you prefer to have this change documented in `CHANGELOG.md`, it should be minimal to update `CHANGELOG.md`

stuartmorgan:
This README isn't published, because pub.dev prioritizes `example/lib/main.dart` (unfortunately, in this case, since that's empty) over `example/README.txt`, which is why it's not failing the version check. It's fine to CHANGELOG-exempt this; the only reason changes to unpublished example files aren't auto-exempted from the CHANGELOG check is because sometimes changes to example files are relevant to people understanding the package, such as adding a missing permission to the example app that a real app would need, so we add a human check.

stuartmorgan:
CHANGELOG override: minor update to instructions for local running.

stuartmorgan (approved):
LGTM

tarrinneal (approved):
Well, there ya go! Thanks for putting this together!

PR #7712, opened 2024-09-26
Add a `@SuppressWarnings` in advance 

matanlurey:
... to avoid a revert like https://github.com/flutter/engine/pull/55450.

/cc @jtmcdole another example of churn where I can't do the right thing without manual steps.


jonahwilliams (approved):
LGTM

PR #7722, opened 2024-09-27
Background/resuming `video_player` on Android sends one `initialized` event

matanlurey:
Closes https://github.com/flutter/flutter/issues/154602.

I also added small clarifications to the error in `video_player` (an assertion/error that are better than a `Completer.complete` default error) and updated the documentation in `video_player_platform` (it might have been obvious, but it definitely is now).


gmackall (approved):
I'm not super knowledgable about this plugin, but the change lgtm!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -14,11 +14,16 @@ final class ExoPlayerEventListener implements Player.Listener {
...
   private final VideoPlayerCallbacks events;
   private boolean isBuffering = false;
-  private boolean isInitialized = false;
+  private boolean isInitialized;
 
   ExoPlayerEventListener(ExoPlayer exoPlayer, VideoPlayerCallbacks events) {
+    this(exoPlayer, events, false);
+  }
+
+  ExoPlayerEventListener(ExoPlayer exoPlayer, VideoPlayerCallbacks events, boolean wasInitialized) {
```
tarrinneal, line 23: Maybe there is a design guide somewhere that disagrees, but doesn't `was` imply in the past and (usually) no longer? is `is` not more appropriate? Not a huge deal either way.

tarrinneal (approved):
looks good. One "nit" which I don't really care if you change, just curious what your thoughts about it are :)

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
matanlurey: Yeah I think this was a poor word choice on my part, I'd be happy to rename this `isInitialized` or even `initialized` for consistency. Let me do that before merging.

matanlurey: Done!

PR #7725, opened 2024-09-28
[pigeon] fix int bug

tarrinneal:
fixes https://github.com/flutter/flutter/issues/155512

### DIFF
--- a/packages/pigeon/CHANGELOG.md
+++ b/packages/pigeon/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 22.4.1
+
+* [dart] Fixes bug where special handling of ints is ignored if no custom types are used.
+
 ## 22.4.0
 
 * Adds support for non-nullable types in collections.
--- a/packages/pigeon/lib/dart_generator.dart
+++ b/packages/pigeon/lib/dart_generator.dart
@@ -339,44 +339,42 @@ class DartGenerator extends StructuredGenerator<DartOptions> {
     indent.write('class $_pigeonCodec extends StandardMessageCodec');
     indent.addScoped(' {', '}', () {
       indent.writeln('const $_pigeonCodec();');
-      if (enumeratedTypes.isNotEmpty) {
-        indent.writeln('@override');
-        indent.write('void writeValue(WriteBuffer buffer, Object? value) ');
-        indent.addScoped('{', '}', () {
-          indent.writeScoped('if (value is int) {', '}', () {
-            indent.writeln('buffer.putUint8(4);');
-            indent.writeln('buffer.putInt64(value);');
-          }, addTrailingNewline: false);
+      indent.writeln('@override');
+      indent.write('void writeValue(WriteBuffer buffer, Object? value) ');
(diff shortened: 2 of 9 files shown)


stuartmorgan (approved):
LGTM.

Did you audit the other generators to make sure none of them have this same bug in the other direction (for calling a `FlutterApi` with an int, and no custom types anywhere)?

tarrinneal:
> LGTM.
> 
> Did you audit the other generators to make sure none of them have this same bug in the other direction (for calling a `FlutterApi` with an int, and no custom types anywhere)?

Yeah, the change to handling ints is only a dart change so the logic doesn't exist in the other generators

EchoEllet:
It seem that [this part](https://github.com/flutter/packages/pull/7725/files#diff-0586de4d30521ddc85ce6b0c29ea52bbb41577fbd7f2cf4eb4d19e48d3f525cfR346-R347) causing Flutter analysis failure:

```dart
class _PigeonCodec extends StandardMessageCodec {
  const _PigeonCodec();
  @override
  void writeValue(WriteBuffer buffer, Object? value) {
    if (value is int) {
      buffer.putUint8(4);
      buffer.putInt64(value); // Unnecessary duplication of receiver. Try using a cascade to avoid the duplication
    } else {
      super.writeValue(buffer, value);
    }
  }
```

Should be updated to:

```dart
buffer..putUint8(4)
      ..putInt64(value);
```

This version was released as a minor version, pub always uses the latest available minor version and it causes [analysis failure](https://github.com/singerdmx/flutter-quill/actions/runs/11103808226/job/30846555011), maybe this should be changed to satisfy Dart analysis or ignore all lint rules in the top of the file, or maybe provide the option to?

Since at the moment, we can't directly depend on ` dart run pigeon --input ./pigeons/messages.dart`, instead we have to use our own command which runs `dart fix` for that generated file to fix them all.

stuartmorgan:
> Should be updated to

This PR has already landed, and cannot be updated. The place to report issues with code that has already landed is the issue tracker.

> it causes [analysis failure](https://github.com/singerdmx/flutter-quill/actions/runs/11103808226/job/30846555011) [...] maybe this should be changed to satisfy Dart analysis

That depends on your project's analysis settings; Dart has many optional analysis options, some of which are contradictory. We do not (and *cannot*, due to the contradictory settings) guarantee that all Pigeon-generated code will pass every possible set of analysis options.

>  or ignore all lint rules in the top of the file, or maybe provide the option to?

You can put arbitrary text in the `copyrightHeader` file, including ignore options.

EchoEllet:
> That depends on your project's analysis settings; Dart has many optional analysis options, some of which are contradictory. We do not (and _cannot_, due to the contradictory settings) guarantee that all Pigeon-generated code will pass every possible set of analysis options.

You're right, it seems that I have missed that the [cascade_invocations](https://dart.dev/tools/linter-rules/cascade_invocations) is not part of the recommended nor the core lints provided by `lints` and `flutter_lints`.

So an issue specific to a project and not part of this repo.

Thank you for clarifying.

PR #7632, opened 2024-09-11
Add support for passing query parameters into benchmark URIs

kenzieschmoll:
This will unblock https://github.com/flutter/devtools/pull/8320

PR #7743, opened 2024-09-30
Replace `initialPage` parameter with `benchmarkPath`

kenzieschmoll:
This will unblock https://github.com/flutter/devtools/pull/8320. This is a follow up to the initial attempt here: https://github.com/flutter/packages/pull/7632


eyebrowsoffire (approved):
LGTM!

PR #7759, opened 2024-10-01
Add flutter frame timings to benchmark metrics

kenzieschmoll:
Work towards https://github.com/flutter/flutter/issues/151551. I will follow on with a PR to add the percentile metrics (p50, p90, etc.).


packages/web_benchmarks/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: web_benchmarks
 description: A benchmark harness for performance-testing Flutter apps in Chrome.
 repository: https://github.com/flutter/packages/tree/main/packages/web_benchmarks
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+web_benchmarks%22
-version: 3.0.0
+version: 3.1.0-wip
```
yjbanov, line 5: Is `wip` a label that's understood by `pub`? Should we use something like `dev` instead, which I believe has special meaning to `pub`. Also, for a small package like this, maybe we don't need a label? Just make it 3.1.0, populate the changelog, and immediately after releasing 3.1.0 we bump it to 3.2.0. WDYT?

kenzieschmoll: wip is to indicate the package is still a work in progress and is not ready to publish. If the pubspec version is a clean version like '3.1.0' release github action will run and automatically publish the package.

packages/web_benchmarks/lib/src/metrics.dart
```diff
@@ -0,0 +1,45 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// The metrics measured for each benchmark.
```
yjbanov, line 5: Ooh, `enum` is so much better! Thanks for refactoring this.

Bonus points: change the style of how these metric names are described (including the top-level enum dartdoc). We claim that these are "metrics", but they are not. They are "labels" or "names" for the various metrics we collect.

packages/web_benchmarks/pubspec.yaml
yjbanov: Ah, I wasn't aware of how our github actions worked. Cool!

yjbanov (approved):

PR #7760, opened 2024-10-01
Add percentile computations to benchmark scores

kenzieschmoll:
This adds p50, p90, and p95 computations to each benchmark metric. This PR also refactors the benchmark computation definitions so they are defined by an enum.

Fixes https://github.com/flutter/flutter/issues/151551.


packages/web_benchmarks/lib/src/metrics.dart
```diff
@@ -50,6 +50,36 @@ enum BenchmarkMetric {
...
+
+  /// The name for the computed value tracking the average of outlier samples.
+  outlierAverage,
+
+  /// The name for the computed value tracking the outlier average divided by
+  /// the clean average.
+  outlierRatio,
+
+  /// The name for the computed value tracking the noise as a multiple of the
+  /// [average] value takes from clean samples.
```
yjbanov, line 67: nit: "taken"?

packages/web_benchmarks/lib/src/timeseries.dart
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  return computed;
+}
+
+// ignore: public_member_api_docs, this will be made private in a follow up PR
+extension PercentileExtension on BenchmarkMetricComputation {
+  // ignore: public_member_api_docs, this will be made private in a follow up PR
+  double? get toPercentile {
+    if (RegExp(r'^p\d+$').hasMatch(name)) {
```
yjbanov, line 52: Parsing enum names feels wrong somehow. Can we instead structure our enums in a way that doesn't require this? Here's one idea - use sealed classes with data fields instead of enums: https://gist.github.com/yjbanov/4639cffa4eec6db95a6fa0001c0cfe45

A simpler but less type-safe variant would be do attach `percentile` to the enum: https://gist.github.com/yjbanov/bfb52443c2f7db0121c55ecb4ed02f72

kenzieschmoll: used the sealed classes approach. Thanks.

packages/web_benchmarks/lib/src/recorder.dart
```diff
@@ -942,18 +965,25 @@ class Profile {
...
-      json['$key.outlierAverage'] = stats.outlierAverage;
-      json['$key.outlierRatio'] = stats.outlierRatio;
-      json['$key.noise'] = stats.noise;
+      json['$key.${BenchmarkMetricComputation.average.name}'] = stats.average;
+      json['$key.${BenchmarkMetricComputation.outlierAverage.name}'] =
+          stats.outlierAverage;
+      json['$key.${BenchmarkMetricComputation.outlierRatio.name}'] =
+          stats.outlierRatio;
+      json['$key.${BenchmarkMetricComputation.noise.name}'] = stats.noise;
+      for (final MapEntry<double, double> percentileEntry
```
yjbanov, line 982: Should we instead iterate over `PercentileMetricComputation.values` and extract computed data using `stats.percentiles[metric.percentile]`? Then we don't need to reconstruct the name of the metric from its `double` value. In fact the metric name can be anything human-readable, like "95th percentile". One issue with `double.toString` is that JS behavior can be different from Wasm/VM behavior for certain values.

Sketch:

```dart
for (final metric in PercentileMetricComputation.values) {
  jaons[metric.name] = stats.percentiles[metric.percentile]!;
}
```

yjbanov (approved):
lgtm

PR #7762, opened 2024-10-02
[various] Unpin some dev dependencies

stuartmorgan:
Several dev dependencies, most notably mockito, were only allow as exactly pinned versions, in order to avoid out-of-band breakage due to deprecations in minor version updates. Now that the policy for Flutter repos has changed to not include deprecation warnings in analysis, this is no longer an issue, so we can relax these dependencies to follow the more normal pattern of using ranges.

### DIFF
--- a/packages/camera/camera/pubspec.yaml
+++ b/packages/camera/camera/pubspec.yaml
@@ -32,7 +32,7 @@ dependencies:
 dev_dependencies:
   flutter_test:
     sdk: flutter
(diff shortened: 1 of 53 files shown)


stuartmorgan:
@tarrinneal or @bparrishMines please review; everyone else (auto-added by code owners) can ignore this.

tarrinneal (approved):

PR #7774, opened 2024-10-03
[url_launcher] Decode file URLs before passing it to ShellExecuteW

ftsui:
- ShellExecuteW does not handle file: urls that contain %-encoded UTF-8 strings correctly.  %-encoded ASCII
   strings are handled correctly, as are file "urls" that contain Unicode strings in its path component.
- This change perform URL decode on file: urls before passing to ShellExecuteW.

Fixes https://github.com/flutter/flutter/issues/156790

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/url_launcher/url_launcher_windows/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher_windows/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 3.1.3
 
 * Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
-* Fix handling of file: URLs that contain UTF-8 encoded paths.
+* Fixes handling of `file:` URLs that contain UTF-8 encoded paths.
 
 ## 3.1.2
 
--- a/packages/url_launcher/url_launcher_windows/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher_windows/CHANGELOG.md
@@ -1,6 +1,7 @@
-## NEXT
+## 3.1.3
 
 * Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
+* Fix handling of file: URLs that contain UTF-8 encoded paths.
 
 ## 3.1.2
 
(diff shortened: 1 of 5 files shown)


stuartmorgan:
Thanks for the contribution! Since the checklist isn't complete, I'm assuming this is a work in progress, and marking as a Draft accordingly. Once you've completed all of the steps, please mark it as ready for review.

It also appears that this PR assumes that `url_launcher` is only used for `file:` URLs. It's intended for use with any URL, and is very frequently used with schemes like `https` and `mailto`; a change that breaks those use cases would not be accepted.

stuartmorgan (approved):
LGTM, @cbracken for secondary review.

(Longer term we may need to revisit this logic; it's not at all clear that we are using `open` correctly current even though it mostly seems to work. The docs say "The item can be a file or folder", and even with this change we are passing a file *URL* rather than path, which may not be right. And it's not clear whether `open` is supposed to even work at all for other kinds of URLs, even though it apparently does.)

ftsui:
My guess would be that file: is being handled by ShellExecute as a file moniker (which happens to have a similar syntax, but without URL encoding).  That might also explain starting the image editor instead of a browser when passed an image file.  It would be a somewhat unfortunate change in behavior if this were to be changed to pass all URLs to the browser.

stuartmorgan:
> It would be a somewhat unfortunate change in behavior if this were to be changed to pass all URLs to the browser.

That's definitely an anti-goal. What we want is for the OS to determine the best handler for any given URL and pass it to that; that's explicitly what the underlying APIs used by all the other platform implementations of `url_launcher`. It seems like `open` is doing that in practice (at least mostly), but it doesn't seem like that's documented behavior so ideally we'd find something that explicitly is documented to do the thing we want (if there is such a thing on Windows).

cbracken (approved):

lgtm. There's a joke to be made somewhere about 管理ing all of my various 家s, but I'm not sure I'm clever enough to make it.

PR #7805, opened 2024-10-07
[google_maps_android_flutter] Convert `Config.sdk` to `minSdk` in Robolectric tests and lower to `LOLLIPOP`

yaakovschectman:
Tests in the title package that are run using the Robolectric test runner will now run on all SDKs level >= LOLLIPOP (i.e. 21). Raises max heap size to 1G.

Affects only tests, should be version-bump exempt.

Fixes https://github.com/flutter/flutter/issues/152931

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/build.gradle
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/build.gradle
@@ -66,6 +66,7 @@ android {
                outputs.upToDateWhen {false}
                showStandardStreams = true
             }
+            jvmArgs "-Xmx1g"
         }
     }
 }
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/build.gradle
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/build.gradle
@@ -65,6 +65,10 @@ android {
                outputs.upToDateWhen {false}
                showStandardStreams = true
             }
+            // The org.gradle.jvmargs property that may be set in gradle.properties does not impact
+            // the Java heap size when running the Android unit tests. The following property here
+            // sets the heap size to a size large enough to run the robolectric tests across
+            // multiple SDK levels.
             jvmArgs "-Xmx1g"
         }
     }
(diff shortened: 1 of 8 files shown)


yaakovschectman:
https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version states `PRs that only affect tests` are exempt from versioning changes.

packages/google_maps_flutter/google_maps_flutter_android/android/build.gradle
```diff
@@ -65,6 +65,7 @@ android {
                outputs.upToDateWhen {false}
                showStandardStreams = true
             }
+            jvmArgs "-Xmx1g"
```
reidbaker, line 68: Good find but instead of setting the jvm args in build.gradle set them in settings.gradle like we do in the rest of the places we bump memory. 

For this pr I think you need to create a gradle.properties and put this value there. 

Here is an example from where daco bumped the memory used in the templates. https://github.com/flutter/flutter/pull/156201/files

reidbaker: https://github.com/search?q=repo%3Aflutter%2Fpackages%20jvmArgs%20%22-Xmx1g%22&type=code
I see one example. Can you modify camera and this pr to use gradle.properties?

reidbaker: The reasoning is that keeping track of memory allocation is hard enough as a class of problem and having 2 different ways we do it makes it even harder. Especially because properties clober each other so any jvm args set in one place will be overridden in the second place.

yaakovschectman: I may be misunderstanding the actual use of this. Creating `gradle.properties` in the same path as this `build.gradle` file and populating it with `org.gradle.jvmargs=-Xmx1G` (and removing the added line from build.gradle) reintroduces the OOM that was the issue in the first place. And trying instead to set the property `android.testOptions.unitTests.all.jvmArgs` in gradle.properties also does not resolve the OOM error, either. It at least appears that whatever is set in gradle.properties is not impacting the android unit tests the way build.gradle does.

yaakovschectman: I've added a comment explaining this. Please see if it warrants an LGTM or further requests.

PR #7822, opened 2024-10-08
Applied Gradle Plugins Declaratively for `path_provider`

jesswrd:
Updated applying gradle plugins from usage of imperative apply to usage of declarative blocks {} apply. Intending on updating all android example apps under packages. Did one more as a proof of concept before doing more. 

More information on Flutter Gradle Plugin Apply [here](https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply)

Partially addresses [#152656](https://github.com/flutter/flutter/issues/152656)
## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Looks like this change triggers a new check on the existing `targetSdkVersion` value?

jesswrd:
> Looks like this change triggers a new check on the existing `targetSdkVersion` value?

Yeah, I think the current `targetSdkVersion` is outdated. I'm just going to update it to 35 (most recent one)

packages/path_provider/path_provider/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## NEXT
 
+* Updates example app Android compileSdkVersion to 35.
```
reidbaker, line 3: This changelog should also mention moving the example to include flutter as a declarative dependency.

reidbaker (approved):

packages/path_provider/path_provider/CHANGELOG.md
stuartmorgan: Why are we updating the CHANGELOG at all for this PR? Example-build-only changes are exempt; we generally only mention them if they are about something that is directly relevant to clients of the package (e.g., if a new version of Android requires a new permission to use the code in the package, we might mention adding that permission to the example).

reidbaker: Because we are changing the example app which is users facing and should be updated.

stuartmorgan: It's user-facing, but (other than the one file from the example directory shown on pub.dev) it's not directly pub-package-client-facing; if someone `flutter pub upgrade`s and gets a new version of `path_provider`, nothing in this PR is in the code they will import, and that's generally what we consider as the CHANGELOG audience.

We expect people running the example to check out the repo (in theory they could build and run it directly from their pub cache, but they really shouldn't), and the git history is the record of changes for someone interacting with the repo itself.

The things in the CHANGELOG change here are part of what our written policy explicitly exempts, and we also have not historically put this kind of thing in the CHANGELOG, so it seems odd to include it this time.

reidbaker: Ok, I misunderstood the policy. @jesswrd sorry for leading you astray but it will make your follow up prs much easier.

script/tool/lib/src/gradle_check_command.dart
```diff
@@ -216,6 +216,31 @@ buildscript {
...
+    // ...other plugins
+    id "com.google.cloud.artifactregistry.gradle-plugin" version "2.2.1"
+}
+  ''';
+
+  /// String printed as a valid example of settings.gradle repository
+  /// configuration without the artifact hub env variable.
+  /// GP stands for the gradle plugin method of flutter tooling inclusion.
+  @visibleForTesting
+  static String exampleSettingsWithoutArtifactHubStringGP = '''
```
stuartmorgan, line 236: Why is this defined here? It only seems to be used in tests.

script/tool/lib/src/gradle_check_command.dart
```diff
@@ -228,24 +253,31 @@ apply plugin: "com.google.cloud.artifactregistry.gradle-plugin"
...
+    }
+    if (!validArtifactConfiguration && !(artifactRegistryDefined && artifactRegistryPluginApplied)) {
       printError('Failed Artifact Hub validation. Include the following in '
           'example root settings.gradle:\n$exampleRootSettingsArtifactHubString');
     }
-
-    return documentationPresent &&
-        artifactRegistryDefined &&
-        artifactRegistryPluginApplied;
+    return validArtifactConfiguration;
```
stuartmorgan, line 280: If the only thing is missing is the documentation, before it would print an example of what needed to be included that included the documentation, whereas now it would fail without any actionable message.

stuartmorgan (changes requested):

script/tool/lib/src/gradle_check_command.dart
jesswrd: Moved it to the test file.

script/tool/lib/src/gradle_check_command.dart
jesswrd: I just tried out only removing only the documentation, and it does include an error message. I have a check for `documentationPresent` when `validArtifactConfiguration` is deinfed, and check for `!validArtifactConfiguration` when an error should be printed. Could you further explain why it would fail? I can add a test for this?

script/tool/lib/src/gradle_check_command.dart
```diff
@@ -228,24 +241,31 @@ apply plugin: "com.google.cloud.artifactregistry.gradle-plugin"
...
+    final bool validArtifactConfiguration = documentationPresent &&
+        ((artifactRegistryDefined && artifactRegistryPluginApplied) || artifactRegistryPluginAppliedGP);
 
-    if (!(documentationPresent &&
-        artifactRegistryDefined &&
-        artifactRegistryPluginApplied)) {
+    if (!validArtifactConfiguration && !artifactRegistryPluginAppliedGP) {
+      printError('Failed Artifact Hub validation. Include the following in '
+          'example root settings.gradle:\n$exampleSettingsArtifactHubStringGP');
+    }
```
reidbaker, line 263: Instead of these lines which I think are a response to [this thread](https://github.com/flutter/packages/pull/7822/files#r1819449426). 

add documentationPresent to the check for GP method of applying artifact registry.

reidbaker (changes requested):

script/tool/lib/src/gradle_check_command.dart
```diff
@@ -228,24 +239,44 @@ apply plugin: "com.google.cloud.artifactregistry.gradle-plugin"
...
+      printError('Failed Artifact Hub validation.');
+      if (!documentationPresent) {
+        printError('The link to the Artifact Hub documentation is missing. Include the following in '
+            'example root settings.gradle:\n// See $artifactHubDocumentationString for more info.');
+      }
+      if (artifactRegistryDefined || artifactRegistryPluginApplied || !declarativePluginBlockApplied) {
+        printError('Include the following in '
+            'example root settings.gradle:\n$exampleRootSettingsArtifactHubString');
+      }
+      else if (!declarativeArtifactRegistryApplied){
```
stuartmorgan, line 274: I think this file needs `dart format` run on it. (CI currently doesn't check formatting for script/tool, which is a bug.)

script/tool/test/gradle_check_command_test.dart
```diff
@@ -789,6 +869,72 @@ dependencies {
...
+    test('prints error for declarative method of applying gradle plugins', () async {
+      const String packageName = 'a_package';
+      final RepositoryPackage package =
+      createFakePackage('a_package', packagesDir);
+      writeFakePluginBuildGradle(package, includeLanguageVersion: true);
+      writeFakeManifest(package);
+      final RepositoryPackage example = package.getExamples().first;
+      writeFakeExampleBuildGradleGP(example,
+          pluginName: packageName,
+          // ignore: avoid_redundant_argument_values
```
stuartmorgan, line 882: If we always want to have the tests be explicit about all of their arguments, then let's make the test helper arguments be `required` instead of having to annotate all of the calls.

stuartmorgan (approved):
LGTM with a couple of small changes.

script/tool/lib/src/gradle_check_command.dart
jesswrd: done

script/tool/lib/src/gradle_check_command.dart
jesswrd: done

script/tool/test/gradle_check_command_test.dart
reidbaker: I dont see the version of the code where includeBuildArtifactHub and includeSettingsDocumentationArtifactHub are required without defaults  and the `// ignore: avoid_redundant_argument_values` are removed. Did you need to push a change?

jesswrd: I added `required` and removed the annotations for `writeFakeExampleBuildGradleGP` but not for `writeFakeExampleBuildGradles`. Let me update that

reidbaker (approved):

PR #7827, opened 2024-10-08
Disable `SurfaceProducer.Callback` when the surface is disposed manually.

matanlurey:
Closes https://github.com/flutter/flutter/issues/156158.

I want to address this better in https://github.com/flutter/flutter/issues/156434 but of course that will take another quarter!


camsim99 (approved):
Lgtm!

PR #7837, opened 2024-10-09
Remove additional (harmless but annoying) native stack traces.

matanlurey:
Closes https://github.com/flutter/flutter/issues/156451.

### DIFF
--- a/packages/video_player/video_player_android/CHANGELOG.md
+++ b/packages/video_player/video_player_android/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 2.7.12
+
+* Fixes a [bug](https://github.com/flutter/flutter/issues/156451) where
+  additional harmless but annoying warnings in the form of native stack traces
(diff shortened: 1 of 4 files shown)


camsim99 (approved):
This looks good to me! Just want to confirm that we don't need to care about calling `stop`? I see [this note](https://developer.android.com/reference/androidx/media3/common/Player#stop():~:text=This%20method%20must%20only%20be%20called%20if%20COMMAND_STOP%20is%20available.) about needing to include it in a particular case so just double checking!

matanlurey:
> I see [this note](https://developer.android.com/reference/androidx/media3/common/Player#stop():~:text=This%20method%20must%20only%20be%20called%20if%20COMMAND_STOP%20is%20available.) about needing to include it in a particular case so just double checking!

Good callout! I think it's more of a "can stop be called" versus stop _must_ be called, but thanks for checking!

PR #7846, opened 2024-10-10
[video_player_android] Correct rotation of videos recorded by the camera

camsim99:
Hopefully 🤞 corrects the calculation used to determine the rotation correction of the video being played. These are the different cases this addresses:

1. **Strictly below Android API 22:** Kept this calculation the same as before because I was not able to test it. For context, it uses an unapplied rotation degrees reported by the video's size to correct the rotation. Note that this is always 0 after Android API 22+ (see [its docs](https://github.com/google/ExoPlayer/blob/dd430f7053a1a3958deea3ead6a0565150c06bfc/library/common/src/main/java/com/google/android/exoplayer2/video/VideoSize.java#L65))
2. **Above Android API 22, strictly below Android API 29:** The `SurfaceTexture` Impeller backend is used. From my testing, I see the preview is correctly rotated and sized without any further intervention, so the correction is set 0.
3. **Android API 29+:** The `ImageReader` Impelled backend is used and a preview correction was noticed by the community (see https://github.com/flutter/flutter/issues/154696). To fix this, we now use the rotation correction reported by the video's format. We also now use this rotation to make the fix for a swapped = width and height when the correction is 90 or 270 degrees (indicating that the video is landscape) as the logic did before but instead with the unapplied rotation degrees (see case 1 for context).

Tested this on Android APIs 28, 30, 32, and 35.

A fix attempt for https://github.com/flutter/flutter/issues/154696.

## Pre-launch Checklist

- [x] I signed the [CLA].


camsim99:
Just tested this on an Android 8.1.0 phone and saw that it was the incorrect preview. In fact, it needed no correction. Assuming this actually makes to not using Impeller. Thinking these devices actually need no correction; only those not using Impeller.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -49,23 +53,59 @@ private void sendInitialized() {
...
-      if (rotationDegrees == 90 || rotationDegrees == 270) {
+      int reportedRotationCorrection = 0;
+
+      if (Build.VERSION.SDK_INT <= 21) {
+        // On API 21 and below, Exoplayer may not internally handle rotation correction
+        // and reports it through VideoSize.unappliedRotationDegrees. We may apply it to
+        // fix the case of upside-down playback.
+        reportedRotationCorrection = videoSize.unappliedRotationDegrees;
+        rotationCorrection = getRotationCorrectionFromUnappliedRotation(reportedRotationCorrection);
+      } else if (Build.VERSION.SDK_INT < 29) {
```
reidbaker, line 64: Should this use the new api matan added `handlesCropAndRotation` instead? https://github.com/flutter/engine/pull/55434/files

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -49,23 +53,59 @@ private void sendInitialized() {
...
+        // When the SurfaceTexture backend for Impeller is used, the preview should already
+        // be correctly rotated.
+        rotationCorrection = 0;
+      } else {
+        // Above API 21, Exoplayer handles the VideoSize.unappliedRotationDegrees
```
reidbaker, line 69: This comment is a bit confusing since the previous checks handle less than 29

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -49,23 +53,59 @@ private void sendInitialized() {
...
-      // upside-down playback for videos with rotationDegrees of 180 (other orientations work
-      // correctly without correction).
-      if (rotationDegrees == 180) {
-        rotationCorrection = rotationDegrees;
-      }
     }
     events.onInitialized(width, height, exoPlayer.getDuration(), rotationCorrection);
   }
 
+  private int getRotationCorrectionFromUnappliedRotation(int unappliedRotationDegrees) {
```
reidbaker, line 87: Consider using an enum instead of an int since our code does not actually handle in between values.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -49,23 +53,59 @@ private void sendInitialized() {
...
+    // work correctly without correction).
+    if (unappliedRotationDegrees == 180) {
+      rotationCorrection = unappliedRotationDegrees;
+    }
+
+    return rotationCorrection;
+  }
+
+  @OptIn(markerClass = androidx.media3.common.util.UnstableApi.class)
+  // The annotation is used to enable using a Format to determine the rotation
```
reidbaker, line 102: I think this comment should instead of explaining what an unstable api annotation does should instead explain why it is ok to use an unstable api and what the failure modes are that we can expect. (for example should we expect the signature to possibly change across minor versions).

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/ExoPlayerEventListenerTest.java
```diff
@@ -33,7 +35,7 @@
  * ({@link VideoPlayerCallbacks} and/or interface with the player instance as expected.
  */
 @RunWith(RobolectricTestRunner.class)
-public final class ExoPlayerEventListenerTests {
+public final class ExoPlayerEventListenerTest {
```
reidbaker, line 38: Did you mean to remove the "s" in tests?

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/ExoPlayerEventListenerTest.java
```diff
@@ -56,7 +59,25 @@ public void onPlaybackStateChangedReadySendInitialized() {
   }
 
   @Test
-  public void onPlaybackStateChangedReadyInPortraitMode90DegreesSwapWidthAndHeight() {
+  @Config(sdk = 29)
```
reidbaker, line 62: Consider minSdk instead of sdk here.

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/ExoPlayerEventListenerTest.java
```diff
@@ -56,7 +59,25 @@ public void onPlaybackStateChangedReadySendInitialized() {
...
+    when(mockExoPlayer.getVideoSize()).thenReturn(size);
+    when(mockExoPlayer.getDuration()).thenReturn(10L);
+    when(mockExoPlayer.getVideoFormat()).thenReturn(videoFormat);
+
+    eventListener.onPlaybackStateChanged(Player.STATE_READY);
+    verify(mockCallbacks).onInitialized(400, 800, 10L, rotationCorrection);
+  }
+
+  @Test
+  @Config(sdk = 21)
```
reidbaker, line 78: consider setting min/max sdk here and in other places where you use only sdk. If a wide range is not appropriate consider setting multiple sdk values to at least expand coverage. 

~Finally if this is not specific to api 21 but to the minimum flutter supported version then consider using  [FLUTTER_MIN](https://github.com/flutter/engine/pull/55002/files#diff-c4aa0dd8e137bba5e617ff29c44d35452536c5b83beb972936e153b2a91e7e29)~
Nvm that was in an engine pr.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
matanlurey: It's available in the _next_ stable unfortunately, i.e. the one to be published (but not yet).

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
matanlurey: +1 enum

matanlurey (approved):

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
reidbaker: @camsim99 consider a TODO here to use the new api.

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/ExoPlayerEventListenerTest.java
camsim99: Yes to make it consistent with the other tests in this plugin (and across plugins).

packages/video_player/video_player_android/android/build.gradle
```diff
@@ -57,6 +57,7 @@ android {
         unitTests.includeAndroidResources = true
         unitTests.returnDefaultValues = true
         unitTests.all {
+            jvmArgs "-Xmx1g"
```
reidbaker, line 60: Add a comment here similar to https://github.com/flutter/packages/pull/7805/files

```
            // The org.gradle.jvmargs property that may be set in gradle.properties does not impact
            // the Java heap size when running the Android unit tests. The following property here
            // sets the heap size to a size large enough to run the robolectric tests across
            // multiple SDK levels.
```

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -49,23 +75,62 @@ private void sendInitialized() {
...
+        // The video's Format also provides a rotation correction that may be used to
+        // correct the rotation, so we try to use that to correct the video rotation
+        // when the ImageReader backend for Impeller is used.
+        rotationCorrection = getRotationCorrectionFromFormat(exoPlayer);
+        reportedRotationCorrection = rotationCorrection;
+      }
+
+      // Switch the width/height if video was taken in portrait mode and a rotation
+      // correction was detected.
+      if (reportedRotationCorrection == 90 || reportedRotationCorrection == 270) {
```
reidbaker, line 103: Should this also use the enum?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -49,23 +75,62 @@ private void sendInitialized() {
...
-      // Switch the width/height if video was taken in portrait mode
-      if (rotationDegrees == 90 || rotationDegrees == 270) {
+      int reportedRotationCorrection = 0;
+
+      if (Build.VERSION.SDK_INT <= 21) {
+        // On API 21 and below, Exoplayer may not internally handle rotation correction
+        // and reports it through VideoSize.unappliedRotationDegrees. We may apply it to
+        // fix the case of upside-down playback.
+        reportedRotationCorrection = videoSize.unappliedRotationDegrees;
+        rotationCorrection = getRotationCorrectionFromUnappliedRotation(RotationDegrees.fromDegrees(reportedRotationCorrection));
```
reidbaker, line 85: I think this code need to catch illegal argument exception we are passing a value from an api we dont control and no correction is probably better than an app crash.

camsim99:
cc @matanlurey the video_player_android test `can pause` that we added is flaking pretty badly. I can file an issue but I think we should revisit it.

reidbaker (approved):

camsim99:
> cc @matanlurey the video_player_android test `can pause` that we added is flaking pretty badly. I can file an issue but I think we should revisit it.

Filed https://github.com/flutter/flutter/issues/157307

PR #7847, opened 2024-10-10
Convert use of `mini_player` in the `video_player_android` test to `AndroidVideoPlayer`.

matanlurey:
Closes https://github.com/flutter/flutter/issues/156424.

Also improved an error message that came up as a result.

### DIFF
--- a/packages/video_player/video_player_android/example/integration_test/video_player_test.dart
+++ b/packages/video_player/video_player_android/example/integration_test/video_player_test.dart
@@ -69,13 +69,14 @@ void main() {
     await player.dispose(textureId);
   });
 
-  testWidgets('can seek', (_) async {
+  testWidgets('can seek', (WidgetTester tester) async {
     final int textureId = (await player.create(DataSource(
       sourceType: DataSourceType.asset,
       asset: _videoAssetKey,
     )))!;
 
     await player.seekTo(textureId, const Duration(seconds: 3));
+    await tester.pumpAndSettle(_playDuration);
 
     expect(
       await player.getPosition(textureId),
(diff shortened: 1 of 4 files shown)


packages/video_player/video_player_android/example/integration_test/video_player_test.dart
```diff
@@ -38,132 +30,121 @@ String getUrlForAssetAsNetworkSource(String assetKey) {
...
 
-      expect(await controller.position, const Duration(seconds: 3));
-    });
+  testWidgets('can seek', (_) async {
+    final int textureId = (await player.create(DataSource(
+      sourceType: DataSourceType.asset,
+      asset: _videoAssetKey,
+    )))!;
 
-    testWidgets('can be paused', (WidgetTester tester) async {
```
camsim99, line 82: Should we add back a test for pausing like this one that was removed?

camsim99 (commented):
Looks good just a question about a removed test.

matanlurey:
Thanks @camsim99! PTAL again

packages/video_player/video_player_android/example/integration_test/video_player_test.dart
camsim99: Looks like there was an additional call to `await tester.pumpAndSettle(_playDuration)` after pause was called in the original test. Maybe that explains the slight difference in `player.getPosition(textureId)` and  `pausedDuration` causing the test failure?

camsim99 (approved):
LGTM but there's a test failure. Left a comment on maybe why

packages/video_player/video_player_android/example/integration_test/video_player_test.dart
matanlurey: Hmm, let's try it :)

PR #5815, opened 2024-01-06
[flutter_markdown] Allow for custom block element

dawidope:
Fixes problem with adding custom block syntax.

Issue: https://github.com/flutter/flutter/issues/135848

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

PR #7859, opened 2024-10-11
Allow custom blocks to be something other than `Column` or `SizedBox`

gspencergoog:
# Description

This adds support for allowing block tags recognized by the Markdown processor to insert something other than just a `Column` or a `SizedBox` (the defaults for blocks with children, and without). Without this ability, custom builders can't insert their own widgets to, say, make it be a colored container instead.

This addresses a customer request.

Fixes https://github.com/flutter/flutter/issues/135848

### DIFF
--- a/packages/flutter_markdown/lib/src/builder.dart
+++ b/packages/flutter_markdown/lib/src/builder.dart
@@ -33,6 +33,8 @@ final List<String> _kBlockTags = <String>[
 
 const List<String> _kListTags = <String>['ul', 'ol'];
 
+bool _isBlockTag(String? tag) => _kBlockTags.contains(tag);
+
 bool _isListTag(String tag) => _kListTags.contains(tag);
 
 class _BlockElement {
@@ -109,7 +111,6 @@ class MarkdownBuilder implements md.NodeVisitor {
     required this.builders,
     required this.paddingBuilders,
     required this.listItemCrossAxisAlignment,
-    required this.customBlockTags,
     this.fitContent = false,
     this.onSelectionChanged,
     this.onTapText,
@@ -155,9 +156,6 @@ class MarkdownBuilder implements md.NodeVisitor {
   /// does not allow for intrinsic height measurements.
   final MarkdownListItemCrossAxisAlignment listItemCrossAxisAlignment;
 
-  /// Collection of custom block tags to be used building block widgets.
-  final List<String> customBlockTags;
-
   /// Called when the user changes selection when [selectable] is set to true.
   final MarkdownOnSelectionChangedCallback? onSelectionChanged;
 
@@ -179,10 +177,6 @@ class MarkdownBuilder implements md.NodeVisitor {
   String? _currentBlockTag;
   String? _lastVisitedTag;
   bool _isInBlockquote = false;
-
-  bool _isBlockTag(String? tag) =>
-      _kBlockTags.contains(tag) ||
-      customBlockTags.contains(tag);
 
   /// Returns widgets that display the given Markdown nodes.
   ///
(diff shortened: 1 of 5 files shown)


packages/flutter_markdown/lib/src/builder.dart
```diff
@@ -156,6 +155,9 @@ class MarkdownBuilder implements md.NodeVisitor {
   /// does not allow for intrinsic height measurements.
   final MarkdownListItemCrossAxisAlignment listItemCrossAxisAlignment;
 
+  /// Collection of custom block tags to be used building block widgets.
+  final List<String>? customBlockTags;
```
stuartmorgan, line 159: Nullable collections are generally a red flag; what is the difference between empty and null? If there is one, it should be clearly documented; if there isn't it then it should be non-nullable with a default value of empty.

packages/flutter_markdown/lib/src/builder.dart
```diff
@@ -111,6 +109,7 @@ class MarkdownBuilder implements md.NodeVisitor {
     required this.builders,
     required this.paddingBuilders,
     required this.listItemCrossAxisAlignment,
+    required this.customBlockTags,
```
stuartmorgan, line 112: Adding a required parameter to a public class is a breaking change, and would need to be versioned as such.

Why does this need to be `required`?

packages/flutter_markdown/lib/src/builder.dart
```diff
@@ -178,6 +180,10 @@ class MarkdownBuilder implements md.NodeVisitor {
   String? _lastVisitedTag;
   bool _isInBlockquote = false;
 
+  bool _isBlockTag(String? tag) =>
+      _kBlockTags.contains(tag) ||
+      (customBlockTags ?? <String>[]).contains(tag);
```
stuartmorgan, line 185: This looks like the answer to the previous question is that it should be non-nullable.

packages/flutter_markdown/lib/src/builder.dart
stuartmorgan: Actually, given https://github.com/flutter/packages/pull/5815 why do we need this at all? Isn't this adding a new/different and potentially conflicting way to specific the same information that PR added a way to specify?

stuartmorgan (changes requested):
> This addresses a customer request.

This looks like it's https://github.com/flutter/flutter/issues/135848 ?

packages/flutter_markdown/lib/src/builder.dart
gspencergoog: Actually, I think you're right here.  I didn't realize that API would automatically add to the list. Let me ask the customer why they can't use the existing mechanism.

gspencergoog:
Okay, I've removed the new API surface, and just kept the fixed implementation of `builder.visitElementAfter` so that it can create custom blocks that are something other than a `Column` or a `SizedBox`.

An explanation of the remaining change in `visitElementAfter`:
Inside the `if(_isBlockTag(tag))` in `builder.visitElementAfter`, we are building a widget for the block element, which may be a container for its block children or as a standalone widget.

In the prior implementation, only a handful of elements get specialized block widgets (e.g., list, table, blockquote, pre, hr, see the conditional statements after the modified code), other block elements and custom blocks will be built as a `Column` or an empty `SizedBox`, which may not be ideal in many use cases. For example, there is no way to define a custom block that hosts all its children in a pink container. 

With the added code, we will visit the custom block element's `visitElementAfterWithContext` to build the block widget.

gspencergoog:
I think this does fix the last item in https://github.com/flutter/flutter/issues/135848, referenced in https://github.com/flutter/flutter/issues/135848#issuecomment-2199214436, so I marked it as fixing that bug.

packages/flutter_markdown/lib/src/builder.dart
```diff
@@ -385,16 +385,31 @@ class MarkdownBuilder implements md.NodeVisitor {
       final _BlockElement current = _blocks.removeLast();
       Widget child;
```
stuartmorgan, line 386: Let's move this to just above line 402 so it stays directly attached to the conditionals that initialize it.

packages/flutter_markdown/lib/src/builder.dart
```diff
@@ -385,16 +385,31 @@ class MarkdownBuilder implements md.NodeVisitor {
...
+          delegate.context,
+          element,
+          styleSheet.styles[tag],
+          _inlines.isNotEmpty ? _inlines.last.style : null,
         );
+        child = builderChild ?? defaultChild();
       } else {
-        child = const SizedBox();
+        child = defaultChild();
       }
```
stuartmorgan, line 413: Optional: I think this would be simpler as just:

```dart
final MarkdownElementBuilder? builder = builders[tag];
Widget child = builder?.visitElementAfterWithContext(...) ?? defaultChild();
```

That avoids the duplication of saying that `defaultChild()` is the fallback, and also avoids the deferred initialization of `child`. (The counter-argument is that it's a pretty long ternary statement once you fill in the [...]; I know others have a lower threshold for ternaries than I do, so if you think it's too much that's fine.)

stuartmorgan (approved):
LGTM with a couple of adjustments to the code (one optional).

packages/flutter_markdown/lib/src/builder.dart
gspencergoog: Yes, that's simpler, and I don't mind the ternary length because it eliminates the whole conditional block.

PR #7874, opened 2024-10-14
[camera_android] Convert Dart to native calls to use Pigeon

yaakovschectman:
Add structured types and convert implementations to use typesafe method calls for platform methods from Dart to native side.

Android analog of https://github.com/flutter/packages/pull/6601

## Pre-launch Checklist

- [ ] I signed the [CLA].


packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+            result.error(new Messages.FlutterError(errCode, errDesc, null));
+          }
+        });
+  }
+
+  @Override
+  public void initialize(
+      @NonNull Long cameraId,
+      @NonNull Messages.PlatformImageFormatGroup imageFormat,
+      @NonNull Messages.VoidResult result) {
```
stuartmorgan, line 159: All of the uses of `result` appear to be synchronous here, which means this method does not need to be `@async` in the Pigeon definition.

While the main improvement Pigeon provides is type safety, not having to track result objects when the implementation doesn't require it is another significant safety win because the compiler guarantees that there is always exactly one reply (as the method channel contract requires) vs. having to make sure there are never codepaths with 0 or >1 calls to result, so methods should only be declared `@sync` when they actually have to be.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+  }
+
+  @Override
+  public void takePicture(@NonNull Long cameraId, @NonNull Messages.Result<String> result) {
+    camera.takePicture(result);
+  }
+
+  @Override
+  public void prepareForVideoRecording(@NonNull Messages.VoidResult result) {
+    // This optimization is not required for Android.
```
stuartmorgan, line 183: No-ops should be no-op'd in Dart rather than native; there's a lot of overhead to doing it here.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    try {
+      camera.setExposurePoint(result, new Point(x, y));
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void getMinExposureOffset(
+      @NonNull Long cameraId, @NonNull Messages.Result<Double> result) {
```
stuartmorgan, line 268: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    try {
+      result.success(camera.getMinExposureOffset());
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void getMaxExposureOffset(
+      @NonNull Long cameraId, @NonNull Messages.Result<Double> result) {
```
stuartmorgan, line 279: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    try {
+      result.success(camera.getMaxExposureOffset());
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void getExposureOffsetStepSize(
+      @NonNull Long cameraId, @NonNull Messages.Result<Double> result) {
```
stuartmorgan, line 290: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    }
+    try {
+      camera.setFocusPoint(result, new Point(x, y));
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void startImageStream(@NonNull Messages.VoidResult result) {
```
stuartmorgan, line 348: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    try {
+      camera.startPreviewWithImageStream(imageStreamChannel);
+      result.success();
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void stopImageStream(@NonNull Messages.VoidResult result) {
```
stuartmorgan, line 359: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    try {
+      camera.startPreview();
+      result.success();
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void getMaxZoomLevel(@NonNull Long cameraId, @NonNull Messages.Result<Double> result) {
```
stuartmorgan, line 370: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    try {
+      float maxZoomLevel = camera.getMaxZoomLevel();
+      result.success((double) maxZoomLevel);
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void getMinZoomLevel(@NonNull Long cameraId, @NonNull Messages.Result<Double> result) {
```
stuartmorgan, line 383: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void lockCaptureOrientation(
+      @NonNull Long cameraId,
+      @NonNull Messages.PlatformDeviceOrientation platformOrientation,
+      @NonNull Messages.VoidResult result) {
```
stuartmorgan, line 412: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+      camera.lockCaptureOrientation(orientation);
+      result.success();
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void unlockCaptureOrientation(
+      @NonNull Long cameraId, @NonNull Messages.VoidResult result) {
```
stuartmorgan, line 427: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    try {
+      camera.unlockCaptureOrientation();
+      result.success();
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void pausePreview(@NonNull Long cameraId, @NonNull Messages.VoidResult result) {
```
stuartmorgan, line 438: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+    try {
+      camera.pausePreview();
+      result.success();
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void resumePreview(@NonNull Long cameraId, @NonNull Messages.VoidResult result) {
```
stuartmorgan, line 449: Doesn't need to be async.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,480 @@
...
+      CameraProperties cameraProperties =
+          new CameraPropertiesImpl(cameraName, CameraUtils.getCameraManager(activity));
+      camera.setDescriptionWhileRecording(result, cameraProperties);
+    } catch (Exception e) {
+      handleException(e, result);
+    }
+  }
+
+  @Override
+  public void dispose(@NonNull Long cameraId, @NonNull Messages.VoidResult result) {
```
stuartmorgan, line 473: Doesn't need to be async.

stuartmorgan (changes requested):
I haven't had time for a full review yet, but posting initial comments on the Pigeon API definition now since I had time to start looking this morning. Feel free to wait until I've had time to do a full review if you'd rather, but I thought I'd post this now in case you wanted to adjust the use of `async` in the meantime.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
yaakovschectman: I figured I may as well handle these as soon as I'm able. Let me know if the new state is any better.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraUtils.java
```diff
@@ -169,6 +121,28 @@ public static Messages.PlatformDeviceOrientation orientationToPigeon(
...
+      case PORTRAIT_UP:
+        return PlatformChannel.DeviceOrientation.PORTRAIT_UP;
+      case PORTRAIT_DOWN:
+        return PlatformChannel.DeviceOrientation.PORTRAIT_DOWN;
+      case LANDSCAPE_LEFT:
+        return PlatformChannel.DeviceOrientation.LANDSCAPE_LEFT;
+      case LANDSCAPE_RIGHT:
+        return PlatformChannel.DeviceOrientation.LANDSCAPE_RIGHT;
+    }
+    return PlatformChannel.DeviceOrientation.PORTRAIT_UP;
```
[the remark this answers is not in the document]
yaakovschectman: My intention was to return a default value here in case there is ever a new value added to this enum, as we do not define it, so that the plugin does not break.

packages/camera/camera_android/android/src/test/java/io/flutter/plugins/camera/CameraTest.java
```diff
@@ -549,47 +588,38 @@ public void setZoomLevel_shouldCallErrorOnResultOnCameraAccessException()
...
-    verify(mockResult, times(1)).error("setZoomLevelFailed", "Could not set zoom level.", null);
+    verify(mockResult, never()).success();
+    verify(mockResult, times(1))
+        .error(
+            argThat(
+                new FlutterErrorMatcher("setZoomLevelFailed", "Could not set zoom level.", null)));
   }
 
   @Test
   public void pauseVideoRecording_shouldSendNullResultWhenNotRecording() {
```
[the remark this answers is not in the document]
yaakovschectman: This (and other tests) previously verified that `Result.success` was called in their respective methods, and that `Result.error` was not. Now that those methods are synchronous, the equivalent behavior to calling `success` is to simply return without throwing an error, and the inverse for calling `error`, so my intention behind these tests as they are now is to ensure that they execute their respective methods without throwing any errors.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraUtils.java
```diff
@@ -186,6 +160,23 @@ public static Messages.PlatformFocusMode focusModeToPigeon(@NonNull FocusMode fo
     return Messages.PlatformFocusMode.AUTO;
   }
 
+  /**
+   * Converts a PlatformFocusMode from Pigeon to a FocusMode from the autofocus package.
+   *
+   * @param focusMode A PlatformFocusMode.
+   * @return The corresponding FocusMode.
+   */
+  @Nullable
```
stuartmorgan, line 169: NonNull

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraUtils.java
```diff
@@ -186,6 +160,23 @@ public static Messages.PlatformFocusMode focusModeToPigeon(@NonNull FocusMode fo
...
+  public static FocusMode focusModeFromPigeon(@NonNull Messages.PlatformFocusMode focusMode) {
+    switch (focusMode) {
+      case AUTO:
+        return FocusMode.auto;
+      case LOCKED:
+        return FocusMode.locked;
+    }
+    return null;
```
stuartmorgan, line 177: This can throw since it can't happen.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraUtils.java
```diff
@@ -203,4 +194,92 @@ public static Messages.PlatformExposureMode exposureModeToPigeon(
...
     return Messages.PlatformExposureMode.AUTO;
   }
+
+  /**
+   * Converts a PlatformExposureMode to ExposureMode from the exposurelock package.
+   *
+   * @param mode A PlatformExposureMode.
+   * @return The corresponding ExposureMode.
+   */
+  @Nullable
```
stuartmorgan, line 204: Same.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraUtils.java
```diff
@@ -203,4 +194,92 @@ public static Messages.PlatformExposureMode exposureModeToPigeon(
...
+    return ImageFormat.YUV_420_888;
+  }
+
+  /**
+   * Converts a PlatformFlashMode from Pigeon to a FlashMode from the flash package.
+   *
+   * @param mode A PlatformFlashMode.
+   * @return The corresponding FlashMode.
+   */
+  @Nullable
```
stuartmorgan, line 271: Same.

packages/camera/camera_android/android/src/test/java/io/flutter/plugins/camera/CameraTest.java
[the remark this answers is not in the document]
stuartmorgan: Maybe change the `shouldSendNullResult` to `shouldNotThrow` in the name to make the goal of the test clearer.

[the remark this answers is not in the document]
stuartmorgan: (Similarly with the other tests.)

packages/camera/camera_android/lib/src/android_camera.dart
```diff
@@ -356,221 +281,102 @@ class AndroidCamera extends CameraPlatform {
...
   Future<double> getMinExposureOffset(int cameraId) async {
-    final double? minExposureOffset = await _channel.invokeMethod<double>(
-      'getMinExposureOffset',
-      <String, dynamic>{'cameraId': cameraId},
-    );
-
-    return minExposureOffset!;
+    final double minExposureOffset =
+        await _hostApi.getMinExposureOffset(cameraId);
+    return minExposureOffset;
```
stuartmorgan, line 302: Nit: here and in the several methods below you can just `return _hostApi.[...]` directly instead of having the variable.

packages/camera/camera_android/lib/src/android_camera.dart
```diff
@@ -356,221 +281,102 @@ class AndroidCamera extends CameraPlatform {
...
   @override
   Future<double> getMaxZoomLevel(int cameraId) async {
-    final double? maxZoomLevel = await _channel.invokeMethod<double>(
-      'getMaxZoomLevel',
-      <String, dynamic>{'cameraId': cameraId},
-    );
-
-    return maxZoomLevel!;
+    final double maxZoomLevel = await _hostApi.getMaxZoomLevel(cameraId);
+    return maxZoomLevel;
```
stuartmorgan, line 342: These too.

packages/camera/camera_android/lib/src/utils.dart
```diff
@@ -29,37 +31,91 @@ DeviceOrientation deviceOrientationFromPlatform(
...
+PlatformDeviceOrientation deviceOrientationToPlatform(
+        DeviceOrientation orientation) =>
+    switch (orientation) {
+      DeviceOrientation.portraitUp => PlatformDeviceOrientation.portraitUp,
+      DeviceOrientation.portraitDown => PlatformDeviceOrientation.portraitDown,
+      DeviceOrientation.landscapeLeft =>
+        PlatformDeviceOrientation.landscapeLeft,
+      DeviceOrientation.landscapeRight =>
+        PlatformDeviceOrientation.landscapeRight,
+    };
```
stuartmorgan, line 44: This needs a post-switch handler in case the framework adds a new orientation.

packages/camera/camera_android/lib/src/utils.dart
```diff
@@ -29,37 +31,91 @@ DeviceOrientation deviceOrientationFromPlatform(
...
       PlatformExposureMode.auto => ExposureMode.auto,
       PlatformExposureMode.locked => ExposureMode.locked,
     };
 
+/// Converts a [ExposureMode] to [PlatformExposureMode].
+PlatformExposureMode exposureModeToPlatform(ExposureMode exposureMode) =>
+    switch (exposureMode) {
+      ExposureMode.auto => PlatformExposureMode.auto,
+      ExposureMode.locked => PlatformExposureMode.locked,
+    };
```
stuartmorgan, line 58: This needs a failsafe handler too since it's from the platform interface package.

packages/camera/camera_android/lib/src/utils.dart
```diff
@@ -29,37 +31,91 @@ DeviceOrientation deviceOrientationFromPlatform(
...
-  // switch as needing an update.
-  // ignore: dead_code
-  return 'portraitUp';
-}
+/// Converts a [FocusMode] to [PlatformFocusMode].
+PlatformFocusMode focusModeToPlatform(FocusMode focusMode) =>
+    switch (focusMode) {
+      FocusMode.auto => PlatformFocusMode.auto,
+      FocusMode.locked => PlatformFocusMode.locked,
+    };
```
stuartmorgan, line 72: And this.

packages/camera/camera_android/lib/src/utils.dart
```diff
@@ -29,37 +31,91 @@ DeviceOrientation deviceOrientationFromPlatform(
...
+PlatformResolutionPreset resolutionPresetToPlatform(ResolutionPreset? preset) =>
+    switch (preset) {
+      ResolutionPreset.low => PlatformResolutionPreset.low,
+      ResolutionPreset.medium => PlatformResolutionPreset.medium,
+      ResolutionPreset.high => PlatformResolutionPreset.high,
+      ResolutionPreset.veryHigh => PlatformResolutionPreset.veryHigh,
+      ResolutionPreset.ultraHigh => PlatformResolutionPreset.ultraHigh,
+      ResolutionPreset.max => PlatformResolutionPreset.max,
+      null => PlatformResolutionPreset.high,
+    };
```
stuartmorgan, line 84: And this.

packages/camera/camera_android/lib/src/utils.dart
```diff
@@ -29,37 +31,91 @@ DeviceOrientation deviceOrientationFromPlatform(
...
+/// [ImageFormatGroup.unknown] and [ImageFormatGroup.bgra8888] default to
+/// [PlatformImageFormatGroup.yuv420], which is the default on Android.
+PlatformImageFormatGroup imageFormatGroupToPlatform(ImageFormatGroup format) =>
+    switch (format) {
+      ImageFormatGroup.unknown => PlatformImageFormatGroup.yuv420,
+      ImageFormatGroup.yuv420 => PlatformImageFormatGroup.yuv420,
+      ImageFormatGroup.bgra8888 => PlatformImageFormatGroup.yuv420,
+      ImageFormatGroup.jpeg => PlatformImageFormatGroup.jpeg,
+      ImageFormatGroup.nv21 => PlatformImageFormatGroup.nv21,
+    };
```
stuartmorgan, line 107: And this.

packages/camera/camera_android/lib/src/utils.dart
```diff
@@ -29,37 +31,91 @@ DeviceOrientation deviceOrientationFromPlatform(
...
+
+/// Converts a [FlashMode] to [PlatformFlashMode].
+PlatformFlashMode flashModeToPlatform(FlashMode mode) => switch (mode) {
+      FlashMode.auto => PlatformFlashMode.auto,
+      FlashMode.off => PlatformFlashMode.off,
+      FlashMode.always => PlatformFlashMode.always,
+      FlashMode.torch => PlatformFlashMode.torch,
+    };
```
stuartmorgan, line 115: And this.

packages/camera/camera_android/test/android_camera_test.dart
```diff
@@ -696,221 +504,117 @@ void main() {
...
-          'reset': false
-        }),
-        isMethodCall('setExposurePoint', arguments: <String, Object?>{
-          'cameraId': cameraId,
-          'x': null,
-          'y': null,
-          'reset': true
-        }),
-      ]);
+      verify(mockCameraApi.setExposurePoint(cameraId, argThat(isNotNull)))
```
stuartmorgan, line 552: A lot of tests seem to have lost assertions about the arguments being correct. E.g., this should still be asserting that x and y are being passed the correct values (and x and y should be different so this would catch accidentally sending the wrong one for one of them; that's a latent issue with this particular test). Please audit to make sure that we're not losing coverage about the arguments being correct. In some cases we get that from mocks requiring specific arguments, but cases like here we need to capture the args and assert things about them.

stuartmorgan (changes requested):
Sorry for the delay in getting to the full review. Thanks for all the cleanup where async results were being passed down into `Camera.java` but didn't actually need to be to make even more methods sync!

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraUtils.java
yaakovschectman: I currently modified these methods to throw `IllegalStateException` to represent the unreachable code.

packages/camera/camera_android/test/android_camera_test.dart
yaakovschectman: I've modified a few tests to use `predicate` to require that the arguments' fields match. Since pigeons don't implement equality, we can't just use the object itself as a matcher.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,411 @@
...
+  // to be able to compile with <21 sdks for apps that want the camera and support earlier version.
+  @SuppressWarnings("ConstantConditions")
+  private <T> void handleException(Exception exception, Messages.Result<T> result) {
+    if (exception instanceof CameraAccessException) {
+      result.error(exception);
+      return;
+    }
+
+    // CameraAccessException can not be cast to a RuntimeException.
+    throw (RuntimeException) exception;
```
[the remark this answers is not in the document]
stuartmorgan: Sorry, I realized looking at this again today that I was wrong about this; this is—non-obviously—a regression in this PR. With method channels, on Android, the engine call to `onMethodCall` [is wrapped in an exception handler](https://github.com/flutter/engine/blob/31aaaaad868743b38ac3b7165f0d58eca94e521b/shell/platform/android/io/flutter/plugin/common/MethodChannel.java#L285-L291) (specifically a `RuntimeException` handler, thus the current structure of this code). So the old code *would* correctly report these exceptions back to Dart, but this version won't because Pigeon doesn't `try/catch` `async` methods.

So we can close https://github.com/flutter/flutter/issues/157021, and this code needs to be updated here. To be fully behavior-preserving, `handleException` should look like this:


```
// The code below exactly preserves the format of the native exceptions generated by pre-Pigeon
// code. Since `camera` currently leaks the raw platform exceptions to clients, there may be client
// code that relies on specific string values here, so these should not be changed. See
// https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#platform-exception-handling
// for longer-term solutions to this.
if (exception instanceof CameraAccessException) {
  result.error(new FlutterError("CameraAccess", exception.getMessage(), null));
  return;
}
result.error(new FlutterError("error", exception.getMessage(), null, Log.getStackTraceString(exception)));
```

Apologies for missing this before; I'm still more used to the Obj-C method channel behaviors that the Java method channels.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
```diff
@@ -0,0 +1,354 @@
...
+      throw new Messages.FlutterError("CameraAccessException", e.getMessage(), null);
+    }
+  }
+
+  @Override
+  public void resumePreview() {
+    try {
+      camera.resumePreview();
+    } catch (Exception e) {
+      throw new Messages.FlutterError("CameraAccessException", e.getMessage(), null);
```
stuartmorgan, line 334: I missed this, but here and below you are catching `Exception` but reporting it as a `CameraAccessException`. I don't think you need these two handlers at all since you aren't doing anything specific with the exceptions, and these are sync methods, so the generic Pigeon-level handler should be fine.

stuartmorgan: Same with `stopImageStream`.

stuartmorgan (changes requested):

bparrishMines (approved):
LGTM but @stuartmorgan can provide the final LGTM

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraApiImpl.java
yaakovschectman: `stopImageStream` calls `startPreview`, which can throw two types of non-runtime Exception, so it needs to be wrapped in a try/catch to compile. I'm not sure what you're referring to by "and below", but I made the change to this method and incorporated the feedback from your other comment.

stuartmorgan (approved):
LGTM!

PR #7879, opened 2024-10-15
[image_picker_web] Loosen mime dep to >=1.0.4 <3.0.0

cbracken:
Provides version interop for projects that depend on mime version 2.0.0 and later.

The breaking version change was due to a change in the return value of `extensionFromMime`, which is unused in `image_picker`.

See: https://github.com/dart-lang/tools/blob/main/pkgs/mime/CHANGELOG.md#200

Fixes: https://github.com/flutter/flutter/issues/156905

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] page, which explains my responsibilities.
- [X] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [X] I signed the [CLA] (or in my case, didn't, but work for Google).
- [X] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [X] I [linked to at least one issue that this PR fixes] in the description above.
- [X] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added an unnecessary checkbox since we clearly don't yet have enough.
- [X] I added new tests to check the change I am making, or this PR is [test-exempt].
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[...]

PR #7881, opened 2024-10-16
Removing unnecessary parenthesis

FMorschel:
This is simply removing unnecessary parenthesis from various places inside the packages. This change is because of a change to the unnecessary_parentesis lint that will trigger in these places. Here is the CL https://dart-review.googlesource.com/c/sdk/+/390161.

- https://github.com/dart-lang/linter/issues/4996

## Pre-launch Checklist

- [x] I signed the [CLA].

If anything else is needed please let me know. 

I'd like to ask for this PR to wait a bit until the bots are run again on that CL so that I can be sure nothing else will trigger, I will come back here and update this whenever everything is complete. Thanks!

### DIFF
--- a/packages/pigeon/CHANGELOG.md
+++ b/packages/pigeon/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 22.5.1
 
-* Removes unnecessary parenthesis.
+* Removes unnecessary parentheses.
 
 ## 22.5.0
 
@@ -221,8 +221,6 @@
     `myGeneratedClass.aBoolProperty = @NO` can silently set `aBoolProperty` to
     `YES`. Any data class or Flutter API interactions involving `bool`s should
     be carefully audited by hand when updating.
-
-
 
 ## 12.0.1
 
--- a/packages/pigeon/CHANGELOG.md
+++ b/packages/pigeon/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 22.5.1
+
+* Removes unnecessary parenthesis.
+
 ## 22.5.0
 
 * [swift] Adds implementation for `@ProxyApi`.
--- a/packages/pigeon/CHANGELOG.md
+++ b/packages/pigeon/CHANGELOG.md
@@ -1,7 +1,3 @@
-## 22.5.1
-
-* Removes unnecessary parentheses.
-
 ## 22.5.0
 
 * [swift] Adds implementation for `@ProxyApi`.
--- a/packages/camera/camera/example/lib/main.dart
+++ b/packages/camera/camera/example/lib/main.dart
@@ -534,7 +534,7 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
           onPressed: cameraController != null &&
                   cameraController.value.isInitialized &&
                   cameraController.value.isRecordingVideo
-              ? (cameraController.value.isRecordingPaused)
+              ? cameraController.value.isRecordingPaused
                   ? onResumeButtonPressed
                   : onPauseButtonPressed
               : null,
--- a/packages/camera/camera_android/example/lib/main.dart
+++ b/packages/camera/camera_android/example/lib/main.dart
@@ -540,7 +540,7 @@ class _CameraExampleHomeState extends State<CameraExampleHome>
           onPressed: cameraController != null &&
                   cameraController.value.isInitialized &&
                   cameraController.value.isRecordingVideo
-              ? (cameraController.value.isRecordingPaused)
+              ? cameraController.value.isRecordingPaused
                   ? onResumeButtonPressed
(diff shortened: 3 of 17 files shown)


cbracken (commented):
The change in its current form LGTM, but you'll need to update the changelog.

FMorschel:
> The change in its current form LGTM, but you'll need to update the changelog.

Do I add the change to the closest `CHANGELOG.md`? There are a lot of examples changed here and no example folder has a changelog.

Or do I only need to add it to the cases where the change is outside the example folder?

hellohuanlin (approved):

cbracken:
I've been working on Flutter nearly a decade and I'll be honest, until yesterday I didn't know either ([evidence](https://github.com/flutter/packages/pull/7879)).

The example changes won't need anything, but CI is likely complaining about the following:
* https://github.com/flutter/packages/pull/7881/files#diff-0586de4d30521ddc85ce6b0c29ea52bbb41577fbd7f2cf4eb4d19e48d3f525cf
* https://github.com/flutter/packages/pull/7881/files#diff-87d9ef34a84045acf8bf73fd8313f590ac151fdf4afc4695fa6908e7f35d41cd

So I suspect bumping the patch version here:
* https://github.com/flutter/packages/blob/main/packages/pigeon/pubspec.yaml

and adding a changelog entry here:
* https://github.com/flutter/packages/blob/main/packages/pigeon/CHANGELOG.md

should be sufficient.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 22.5.1
+
+* Removes unnecessary parenthesis.
```
cbracken, line 3: nit: "parentheses" plural, because English is terrible and weird.

cbracken:
Looks like this test needs fixing:
```
00:08 [32m+294[0m[31m -1[0m: test/version_test.dart: pigeon version matches pubspec [1m[31m[E][0m[0m                                                                                                                              
  Expected: '22.5.1'
    Actual: '22.5.0'
     Which: is different.
            Expected: 22.5.1
              Actual: 22.5.0
                           ^
             Differ at offset 5
  
  package:matcher              expect
  test/version_test.dart 17:5  main.<fn>
```

Looking at [this line in the pubspec.yaml](https://github.com/flutter/packages/blob/a35f02d79d0e1d8267f2a31e35124f1ce6bde998/packages/pigeon/pubspec.yaml#L5C50-L5C74), you'll want to update [this line in generator_tools.dart](https://github.com/flutter/packages/blob/a35f02d79d0e1d8267f2a31e35124f1ce6bde998/packages/pigeon/lib/generator_tools.dart#L17).

FMorschel:
Thanks. I had completely ignored that comment.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 22.5.1
```
tarrinneal, line 1: probably doesn't need to be a new version, could do NEXT instead. Not sure if it really matters though

tarrinneal (approved):

packages/pigeon/CHANGELOG.md
stuartmorgan: Since this is very clearly a no-op change other than analysis, it doesn't even need a NEXT entry; this falls under the "unless they only affect development of the package" exemption for CHANGELOG changes.

FMorschel: How do I proceed then? Only removing those mentions for the new version from the branch or do I need something else?

stuartmorgan: I'm not sure I understand the question. The changes in this PR don't require any version or CHANGELOG changes in `pigeon`.

FMorschel: My question was if I need to do anything else then, to signify that this falls under the exemption for CHANGELOG changes. From your answer, I guess I don't. If I do, please say so, since this is my first contribution here and I'm not aware. Fixed the version changes. Thanks a lot!

stuartmorgan:
CHANGELOG/version override: While this technically changes some production code, the change is a no-op other than to silence a warning. (Technically the example/lib/main.dart changes should have a version change under our policy since they are published to pub.dev, but the change is trivial enough that we can just pick that up on the next publish event.)

stuartmorgan (approved):

packages/pigeon/CHANGELOG.md
reidbaker: in this case I think no changelog or version change is the appropriate action

reidbaker: nvm i see this was merged.

PR #7882, opened 2024-10-16
[google_maps_flutter] Add Advanced Markers support

aednlaxer:
This PR adds Advanced Markers support to `google_maps_flutter` as discussed in [#155526](https://github.com/flutter/flutter/issues/155526). The document from the issue offers several options to implement Advanced Markers support, this PR uses option 1 (Advanced Marker Dart class is a subclass Marker class).

Summary of changes:
* Add map capability check
* Add `AdvancedMarker` class
* Add `MarkerCollisionBehavior` enum to control Advanced Marker's behavior when it collides with another marker
* Add `PinConfig` bitmap descriptor for customizing Advanced Marker's pin and icon
* Add `markerType` parameter to indicate that Advanced Markers should be used
* Rename `cloudMapId` to `mapId`

Notes:
* native implementations need to know what kind of marker should be used, this is needed to use correct marker options, cluster manager and marker controller. Indicating marker type is done by using a `markerType` option when creating a `GoogleMap` (could be `marker` or `advancedMarker`). Default option is `marker`
* `cloudMapId` is deprecated in favor of `mapId`. New name follows SDKs, documentation and Cloud Console naming.
* web implementation uses generics because `gmaps.Marker` and `gmaps.AdvancedMarkerElement` are not related to each other and should be handled differently
* legacy markers are deprecated but still supported in Maps Javascript API. `google_maps_flutter` still uses them by default in this PR because of backward-compatibility. [#130472](https://github.com/flutter/flutter/issues/130472) is related, package users will be able to use Advanced Markers to fix the deprecation warning
* Advanced Markers example on iOS  doesn't show advanced markers, seems to be [a known issue](https://issuetracker.google.com/issues/370536110)
* Using Flutter widget as Advanced Marker icon (so called View Marker) is not part of this PR, this was discussed in the document and later removed as we agreed that it should be a separate issue

Resolves [#155526](https://github.com/flutter/flutter/issues/155526)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter/example/android/app/build.gradle
```diff
@@ -30,7 +30,7 @@ android {
 
     defaultConfig {
         applicationId "io.flutter.plugins.googlemapsexample"
-        minSdkVersion 20
+        minSdkVersion flutter.minSdkVersion
```
reidbaker, line 33: Thank you for cleaning up our tech debt.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
```diff
@@ -89,11 +102,28 @@ void addClusterManagers(@NonNull List<Messages.PlatformClusterManager> clusterMa
...
+    clusterManagerIdToManager.put(clusterManagerId, clusterManager);
+  }
+
+  /**
+   * Initialize cluster renderer if it hasn't been initialized before. ClusterManager uses default
+   * renderer which should be changed to (legacy) renderer or advanced renderer
+   */
+  private void initializeRenderer(ClusterManager<MarkerBuilder> clusterManager) {
+    final ClusterRenderer<MarkerBuilder> renderer = clusterManager.getRenderer();
+    if (renderer.getClass() == MarkerClusterRenderer.class
```
reidbaker, line 115: Why do we need this reflection here?  Can we use non null instead as a signal here?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
```diff
@@ -225,6 +256,35 @@ protected void onClusterItemRendered(@NonNull T item, @NonNull Marker marker) {
     }
   }
 
+  /** AdvancedMarkerClusterRenderer is a ClusterRenderer that supports AdvancedMarkers */
+  @VisibleForTesting
+  static class AdvancedMarkerClusterRenderer<T extends MarkerBuilder>
```
reidbaker, line 261: Does this need to be a different class? If it does can it extend MarkerClusterRenderer? 

I worry about keeping the two in sync.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -980,6 +984,18 @@ public Boolean isInfoWindowShown(@NonNull String markerId) {
     return lastSetStyleSucceeded;
   }
 
+  @Override
+  public @NonNull Boolean isAdvancedMarkersAvailable() {
```
reidbaker, line 988: Can this be bool instead of Boolean if it is non null?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapFactory.java
```diff
@@ -47,11 +47,11 @@ public PlatformView create(@NonNull Context context, int id, @Nullable Object ar
     builder.setInitialHeatmaps(params.getInitialHeatmaps());
     builder.setInitialTileOverlays(params.getInitialTileOverlays());
 
-    final String cloudMapId = mapConfig.getCloudMapId();
-    if (cloudMapId != null) {
-      builder.setMapId(cloudMapId);
+    final String mapId = mapConfig.getMapId();
```
reidbaker, line 50: For my own education can you share why we would want to use mapId instead of cloudMapId and why the change is safe (wont cause a regression in existing apps)

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkerController.java
```diff
@@ -143,6 +143,9 @@ public void setZIndex(float zIndex) {
     marker.setZIndex(zIndex);
   }
 
+  @Override
+  public void setCollisionBehavior(int collisionBehavior) {}
```
reidbaker, line 147: Please add a javadoc that describes the valid collisionBehavior int's or at least where they are defined. 
Better yet convert int to something typesafe

reidbaker (commented):
High level android code review. I am concerned by having to runtime type check class. 

That said thank you for opening this pull request.

stuartmorgan (changes requested):
I've only looked at the app-facing and platform interface packages so far, but there are some fundamental structure questions at the interface layer.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
aednlaxer: I've removed this check - previously `initializeRenderer` (it was named differently) was called multiple times, now it's called once and there's no need to check if a renderer has been previously assigned

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
aednlaxer: Yes, it does. `android-maps-utils` have two different renderers - one for each marker type (legacy and advanced). This new class extends [DefaultAdvancedMarkersClusterRenderer](http://googlemaps.github.io/android-maps-utils/android-maps-utils%20/com.google.maps.android.clustering.view/-default-advanced-markers-cluster-renderer/index.html) which is meant to be used with `AdvancedMarker`s. As far as I remember, `android-maps-utils` recreates cluster markers under the hood so it has to know which marker type to use

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
aednlaxer: Looks like it can't be done like this. This code uses Pigeon-generated code and Dart's `bool isAdvancedMarkersAvailable()` is translated into Java's `@NonNull Boolean isAdvancedMarkersAvailable();`

packages/google_maps_flutter/google_maps_flutter/example/lib/main.dart
```diff
@@ -28,14 +31,20 @@ import 'scrolling_map.dart';
 import 'snapshot.dart';
 import 'tile_overlay.dart';
 
+/// Map Id is required for some examples to use advanced markers
+// ignore: unnecessary_nullable_for_final_variable_declarations, unreachable_from_main
+const String? mapId = null;
```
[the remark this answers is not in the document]
aednlaxer: Fixed

packages/google_maps_flutter/google_maps_flutter/example/lib/place_advanced_marker.dart
```diff
@@ -0,0 +1,114 @@
...
+    return BitmapDescriptor.pinConfig(
+      backgroundColor: isSelected ? Colors.blue : Colors.white,
+      borderColor: isSelected ? Colors.white : Colors.blue,
+      glyph: Glyph.color(isSelected ? Colors.white : Colors.blue),
+    );
+  }
+
+  /// Whether map supports advanced markers. Null indicates capability check
+  /// is in progress
+  bool? _isAdvancedMarkersAvailable;
```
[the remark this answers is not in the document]
aednlaxer: That's how it's named in Google Maps on the [web](https://developers.google.com/maps/documentation/javascript/advanced-markers/start#check_map_capabilities_optional) and [Android](https://developers.google.com/maps/documentation/navigation/android-sdk/reference/com/google/android/gms/maps/model/MapCapabilities#isAdvancedMarkersAvailable()). It's not necessary to have it named this way but I think it's good to have a consistent naming across platforms. What do you think?

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/advanced_marker.dart
```diff
@@ -0,0 +1,187 @@
...
+        position == other.position &&
+        rotation == other.rotation &&
+        visible == other.visible &&
+        zIndex == other.zIndex &&
+        clusterManagerId == other.clusterManagerId &&
+        collisionBehavior == other.collisionBehavior;
+  }
+
+  @override
+  int get hashCode => markerId.hashCode;
```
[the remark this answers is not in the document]
aednlaxer: This follows `Marker`'s implementation. I noticed other map objects having the same: Heatmap, Polygon, Polyline, Circle. Is it something that should be changed for all these classes? Probably a separate PR?

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/advanced_marker.dart
```diff
@@ -0,0 +1,187 @@
...
+    super.rotation,
+    super.visible,
+    super.zIndex,
+    super.clusterManagerId,
+    super.onTap,
+    super.onDrag,
+    super.onDragStart,
+    super.onDragEnd,
+    this.collisionBehavior = MarkerCollisionBehavior.required,
+  });
```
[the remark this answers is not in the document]
aednlaxer: I don't think it's needed here. Since `AdvancedMarker` is a `Marker`, it still works with other bitmap descriptors. I'd like to add an assertion about `Marker` not using `PinConfig` but didn't find the right place for it

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/glyph.dart
```diff
@@ -0,0 +1,50 @@
...
+
+  /// Color of the text
+  final Color? textColor;
+
+  /// Bitmap image to be displayed in center of the glyph
+  final BitmapDescriptor? bitmapDescriptor;
+
+  /// Color of the default glyph (circle)
+  final Color? color;
+}
```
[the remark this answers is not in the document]
aednlaxer: This is a really good suggestion, thanks. I've changed all three named constructors to be separate classes extending `BitmapDescriptor`. They're all now in `bitmap.dart`

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/glyph.dart
```diff
@@ -0,0 +1,50 @@
...
+
+import '../../google_maps_flutter_platform_interface.dart';
+
+/// Defines a glyph (the element at the center of an [AdvancedMarker] icon).
+/// Default glyph is a circle but can be configured to have a different color,
+/// some text or bitmap
+@immutable
+class Glyph {
+  /// Create a glyph with a circle of the specified [color]
+  factory Glyph.color(Color color) {
```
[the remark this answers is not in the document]
aednlaxer: Changed it to be `CircleGlyph` class that extends `BitmapDescriptor` (it's now in `bitmap.dart`)

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
```diff
@@ -954,3 +986,93 @@ class BytesMapBitmap extends MapBitmap {
...
+/// PinConfig(
+///   glyph: Glyph.bitmapDescriptor(
+///     BitmapDescriptor.asset(
+///       ImageConfiguration(size: Size(12, 12)),
+///       'assets/cat.png'
+///    )
+/// )
+/// ```
+///
+class PinConfig extends BitmapDescriptor {
```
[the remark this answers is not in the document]
aednlaxer: >What's the reason for making this a BitmapDescription, given that it's not a bitmap? It feels out of place here.

I tried making it consistent with [Android implementation](https://developers.google.com/android/reference/com/google/android/gms/maps/model/BitmapDescriptorFactory#fromPinConfig(com.google.android.gms.maps.model.PinConfig)):

```java
AdvancedMarkerOptions advancedMarkerOptions =
    new AdvancedMarkerOptions()
        .icon(BitmapDescriptorFactory.fromPinConfig(pinConfig))
        .position(MARKER_POSITION);
```

>And it seems very problematic that PinConfig is a BitmapDescriptor, which has a Glyph as a field, and Glyph can be constructed with Glyph.bitmap(BitmapDescriptor bitmapDescriptor), which could be passed a PinConfig, allowing for arbitrary nesting of BitmapDescriptors in ways that make no sense.

I've added an assert to `BitmapGlyph` about `bitmap` not being a `AdvancedMarkerGlyph` to prevent nesting of unnecessary pin configs.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/map_configuration.dart
```diff
@@ -112,6 +113,13 @@ class MapConfiguration {
...
   /// See https://developers.google.com/maps/documentation/get-map-id
   /// for more details.
+  final String? mapId;
+
+  /// Identifier that's associated with a specific cloud-based map style.
+  ///
+  /// See https://developers.google.com/maps/documentation/get-map-id
+  /// for more details.
+  @Deprecated('cloudMapId is deprecated. Use mapId instead')
   final String? cloudMapId;
```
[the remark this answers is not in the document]
aednlaxer: Yes, the intent is to rename `cloudMapId` to `mapId`, I've removed the field. Do you think it's okay to rename it and it should be done in this PR?

Why it should be renamed:
* `mapId` is what Google Maps platform implementations use
* "Map ID" is what Cloud Console uses on Map styles page
* `cloudMapId` feels a bit like it's used for cloud styling which is true but after introduction of Advanced markers map ID is required to make Advanced markers work even if map is not using a custom style.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapFactory.java
aednlaxer: I've added a comment about renaming it [here.](https://github.com/flutter/packages/pull/7882#discussion_r1893813468). It's safe because `cloudMapId` can still be used by existing clients, it's marked as deprecated and uses `mapId` under the hood

jmagman:
@aednlaxer it looks like you've pulled in far more changes than your original PR intended, can you fix that up and rebase on main, or whatever you would need to do?

aednlaxer:
> @aednlaxer it looks like you've pulled in far more changes than your original PR intended, can you fix that up and rebase on main, or whatever you would need to do?

Sorry for that, it should now contain new changes only

stuartmorgan:
Looks like some platform reviewers were dropped at some point on accident.

packages/google_maps_flutter/google_maps_flutter/example/lib/main.dart
```diff
@@ -28,14 +31,19 @@ import 'scrolling_map.dart';
 import 'snapshot.dart';
 import 'tile_overlay.dart';
 
+/// Map ID is required for some examples to use advanced markers.
+const String? _mapId = null;
```
stuartmorgan, line 35: I'm having trouble reconciling the comment and the line of code; if it's required for the examples, why is it null?

Is the idea that someone should put in their own map ID here and run the modified example? If so, the comment should be explicit about that.

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -319,8 +321,18 @@ class GoogleMap extends StatefulWidget {
...
+
+  /// Identifier that's associated with a specific cloud-based map style.
+  ///
+  /// See https://developers.google.com/maps/documentation/get-map-id
+  /// for more details.
+  @Deprecated('cloudMapId is deprecated. Use mapId instead')
   final String? cloudMapId;
 
+  /// Indicates whether map should use [AdvancedMarker]s or [Marker]s.
+  final MarkerType markerType;
```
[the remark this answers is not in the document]
stuartmorgan: This still needs to be updated.

stuartmorgan (changes requested):

rajubhusani:
I see Advanced merker icons are not rendering properly on the map.

packages/google_maps_flutter/google_maps_flutter/README.md
```diff
@@ -121,6 +121,21 @@ the `GoogleMap`'s `onMapCreated` callback.
 The `GoogleMap` widget should be used within a widget with a bounded size. Using it
 in an unbounded widget will cause the application to throw a Flutter exception.
 
+### Advanced Markers
+
+[Advanced Markers](https://developers.google.com/maps/documentation/javascript/advanced-markers/overview) are map markers that offer extra customization options. [Map Id](https://developers.google.com/maps/documentation/get-map-id) is required in order to use Advanced Markers:
```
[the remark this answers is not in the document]
aednlaxer: Thanks, fixed

packages/google_maps_flutter/google_maps_flutter/example/lib/main.dart
aednlaxer: Yes, the idea is that someone should put their own ID. I've updated the comment to

```dart
/// Place your map ID here. Map ID is required for pages that use advanced
/// markers.
```

Hope it makes it clear what needs to be done and why.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering.dart
```diff
@@ -121,15 +126,24 @@ class ClusterManagersController extends GeometryController {
...
+        markerClustererCluster.markers.map<MarkerId>((T marker) {
+      return getOnMarkerType(
+        marker: marker,
+        legacy: (gmaps.Marker marker) {
+          return MarkerId((marker.get('markerId')! as JSString).toDart);
+        },
+        advanced: (gmaps.AdvancedMarkerElement marker) {
+          return MarkerId((marker.getAttribute('id')! as JSString).toDart);
+        },
+      );
```
ditman, line 144: Having to use a `getOnMarkerType` here is unfortunate, but since this is the only case, you can just inline the `isA` check and return the correct type.

Ideally you could wrap this as a very specific `getMarkerId(JSObject marker)` method in `convert.dart` that encapsulates the exact logic that we need, instead of having a callback-based general method.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering.dart
```diff
@@ -109,7 +114,7 @@ class ClusterManagersController extends GeometryController {
   void _clusterClicked(
       ClusterManagerId clusterManagerId,
       gmaps.MapMouseEvent event,
-      MarkerClustererCluster markerClustererCluster,
+      MarkerClustererCluster<T> markerClustererCluster,
```
ditman, line 117: I'm fine with the clusterers being generics, because at the end of the day they're "a `List<T>` with a bunch of other stuff".

As long as we don't need too many calls to `doOnType` stuff...

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker_clustering_js_interop.dart
```diff
@@ -68,14 +68,14 @@ extension type MarkerClustererOptions._(JSObject _) implements JSObject {
 ///
 /// https://googlemaps.github.io/js-markerclusterer/classes/Cluster.html
 @JS('markerClusterer.Cluster')
-extension type MarkerClustererCluster._(JSObject _) implements JSObject {
+extension type MarkerClustererCluster<T>._(JSObject _) implements JSObject {
```
ditman, line 71: can you `T extends JSObject`? That way we get a minimum amount of correctness on this type. Not as good as a union type, but it's something.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/markers.dart
```diff
@@ -5,26 +5,26 @@
 part of '../google_maps_flutter_web.dart';
 
 /// This class manages a set of [MarkerController]s associated to a [GoogleMapController].
-class MarkersController extends GeometryController {
+class MarkersController<T, O> extends GeometryController {
```
ditman, line 8: I suspect we need a `class AdvancedMarkersController extends MarkersController` instead of generics here too.

ditman (dismissed):
Hey, thanks for the code! I took a quick look at the core plugin, the platform interface, but specially the web implementation.

Most of the stuff is stylistic choices, but I have a question:

* Why do we even need the `MarkerType` when creating a `Map`? It seems that it can be inferred the first time a `Marker` is used, and the availability of Advanced Markers?

And an objection to the `doXXXOnMarkerType` pattern, which IMO should be proper inheritance (like you did on `AdvancedMarker extends Marker`).

Thanks for the PR, this is a highly requested feature!

rajubhusani:
> @rajubhusani
> 
> > I see Advanced merker icons are not rendering properly on the map. Unhandled Exception: PlatformException(java.lang.IllegalArgumentException: Unable to interpret pin config as a valid image., IllegalArgumentException, Cause: java.lang.NullPointerException: null reference, Stacktrace: java.lang.IllegalArgumentException: Unable to interpret pin config as a valid image.
> 
> Hi, thanks for checking it out. Could please provide more details?
> 
> * Did you use one of the new examples from this PR? Which one?
> * Did you add your API key and map ID before running the app?
> * Did you build your own sample to test advanced markers? Can you share it or provide code that could help reproducing the issue?
> * What kind of image do you use? Would it be possible to test if this image works with legacy (non-advanced) markers?
> * Did you stop the Android app before running the code? Running app should be stopped and started again after checking out this PR's code. The reason is the platform-specific code that won't work using Flutter's hot-restart and should only work if native (Android) app is restarted.

@aednlaxer, Please find the responses,
I just cloned the code from your branch "CodemateLtd:feature/google-maps-advanced-markers-support". And directly ran the Android app from the google_maps_flutter_android/example project. I have not made any changes to that. I have used my own API key setup. All the options are working on the example except the Custom Marker & Cluster.

packages/google_maps_flutter/google_maps_flutter/example/lib/advanced_marker_icons.dart
```diff
@@ -0,0 +1,46 @@
...
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+import 'package:google_maps_flutter_platform_interface/google_maps_flutter_platform_interface.dart';
+
+import 'main.dart' as main;
+import 'marker_icons.dart';
+import 'page.dart';
+
+/// Page that demonstrates how to use custom [AdvanceMarker] icons
```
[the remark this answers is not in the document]
aednlaxer: I've changed this example to show default marker pin, custom color pin, bitmap glyph pin, pin with text glyph and a custom marker image

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/map_configuration.dart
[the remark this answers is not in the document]
aednlaxer: This should be fixed now

packages/google_maps_flutter/google_maps_flutter/example/lib/advanced_marker_icons.dart
```diff
@@ -0,0 +1,50 @@
...
+}
+
+class _AdvancedMarkerIconsBody extends MarkerIconsBody {
+  const _AdvancedMarkerIconsBody(this.mapId);
+
+  @override
+  final String? mapId;
+
+  @override
+  Marker createMarker(
```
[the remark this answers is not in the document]
aednlaxer: I've updated all advanced examples not to do that, should be simpler now. `Advanced marker icons` example now shows several pin config variations

packages/google_maps_flutter/google_maps_flutter/example/lib/place_advanced_marker.dart
```diff
@@ -0,0 +1,118 @@
...
+    // Check if map is capable of showing advanced markers.
+    if (controller != null) {
+      GoogleMapsFlutterPlatform.instance
+          .isAdvancedMarkersAvailable(mapId: controller!.mapId)
+          .then((bool result) {
+        setState(() {
+          _isAdvancedMarkersAvailable = result;
+        });
+      });
+    }
```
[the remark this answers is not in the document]
aednlaxer: I believe this won't work because of `controller!.mapId` (controller will be `null` in `initState`)

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
```diff
@@ -954,3 +982,173 @@ class BytesMapBitmap extends MapBitmap {
...
+          if (backgroundColor != null)
+            'backgroundColor': backgroundColor?.value,
+          if (borderColor != null) 'borderColor': borderColor?.value,
+          if (glyph != null) 'glyph': glyph?.toJson(),
+        }
+      ];
+}
+
+/// Defines a glyph (the element at the center of an [AdvancedMarker] icon).
+sealed class AdvancedMarkerGlyph extends BitmapDescriptor {
```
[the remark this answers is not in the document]
aednlaxer: Would it be okay not to do that in this PR to avoid unexpected behavior? It sounds quite similar to "view markers", I believe it was discussed in the [design doc](155526) and the decision was to implement view markers as a separate feature

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/marker.dart
```diff
@@ -117,6 +119,8 @@ class MarkerId extends MapsObjectId<Marker> {
 /// A marker icon is drawn oriented against the device's screen rather than
 /// the map's surface; that is, it will not necessarily change orientation
 /// due to map rotations, tilting, or zooming.
+///
+/// Deprecated on the web in favor of [AdvancedMarker].
 @immutable
 class Marker implements MapsObject<Marker> {
```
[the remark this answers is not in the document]
aednlaxer: It might cause unwanted changes when users will be forced to migrate from `Marker` to `AdvancedMarker`.  This is unwanted because legacy markers are only deprecated on the web, Android and iOS can still use legacy markers

packages/google_maps_flutter/google_maps_flutter_platform_interface/test/platform_interface/google_maps_flutter_platform_test.dart
```diff
@@ -121,6 +121,16 @@ void main() {
         expect(await platform.getStyleError(mapId: 0), null);
       },
     );
+
+    test(
+      'default implementation of isAdvancedMarkersAvailable throws Unimplemented Error',
```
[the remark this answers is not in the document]
aednlaxer: Yes, it was fixed earlier but not in this test 👍

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -271,13 +271,19 @@ gmaps.InfoWindowOptions? _infoWindowOptionsFromMarker(Marker marker) {
...
-// Attempts to extract a [gmaps.Size] from `iconConfig[sizeIndex]`.
+/// Attempts to extract a [gmaps.Size] from `iconConfig[sizeIndex]`.
 gmaps.Size? _gmSizeFromIconConfig(List<Object?> iconConfig, int sizeIndex) {
-  gmaps.Size? size;
+  final Size? size = _sizeFromIconConfig(iconConfig, sizeIndex);
+  return size != null ? gmaps.Size(size.width, size.height) : null;
+}
+
+/// Attempts to extract a [Size] from `iconConfig[sizeIndex]`.
+Size? _sizeFromIconConfig(List<Object?> iconConfig, int sizeIndex) {
```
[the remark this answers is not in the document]
aednlaxer: Great catch, fixed

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -446,21 +594,44 @@ Future<gmaps.Icon?> _gmIconFromBitmapDescriptor(
...
+      ..title = sanitizeHtml(marker.infoWindow.title ?? '')
+      ..zIndex = marker.zIndex
+      ..visible = marker.visible
+      ..opacity = marker.alpha
+      ..draggable = marker.draggable;
+    return options as O;
+  }
+
   // TODO(ditman): Compute anchor properly, otherwise infowindows attach to the wrong spot.
   // Flat and Rotation are not supported directly on the web.
```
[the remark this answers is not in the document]
aednlaxer: Good idea, this has been added 👍

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
```diff
@@ -37,11 +37,57 @@ class GoogleMapController {
...
+    if (markerTypes.isNotEmpty) {
+      assert(markerTypes.length == 1, 'All markers must be of the same type.');
+
+      switch (mapConfiguration.markerType) {
+        case null:
+        case MarkerType.marker:
+          assert(
+            markerTypes.first == Marker,
+            'All markers must be of type Marker because '
+            'mapConfiguration.markerType is MarkerType.legacy',
```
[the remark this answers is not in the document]
aednlaxer: Android implementation needs to know the marker type beforehand. It creates a cluster manager based on marker type. Often markers are not known when map widget is created (markers are added later) so it's not always possible to know what marker we're dealing with. I think it's good to let user decide what marker type they want to use. iOS and web most likely can use marker type to decide how to treat it but I think it's good to make it  all consistent.

>Who'd want the plugin to use deprecated markers on a Map after passing a mapId?

I might be wrong but some things that work with legacy markers don't work with advanced markers. For example, `BitmapDescriptor.defaultMarkerWithHue` doesn't work with Advanced markers. Whoever is using custom hues and styled map (requires map ID) may not want to use advanced markers

packages/google_maps_flutter/google_maps_flutter_web/lib/src/marker.dart
```diff
@@ -20,32 +24,30 @@ class MarkerController {
...
+      advanced: (gmaps.AdvancedMarkerElement marker) {
+        _initializeAdvancedMarkerElementListener(
+          marker: marker,
+          onDragStart: onDragStart,
+          onDrag: onDrag,
+          onDragEnd: onDragEnd,
+          onTap: onTap,
+        );
+      },
+    );
```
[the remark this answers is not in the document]
aednlaxer: I got rid of `_doOnMarkerType` and added classes for legacy and advanced markers

aednlaxer:
> @aednlaxer, Please find the responses, I just cloned the code from your branch "CodemateLtd:feature/google-maps-advanced-markers-support". And directly ran the Android app from the google_maps_flutter_android/example project. I have not made any changes to that. I have used my own API key setup. All the options are working on the example except the Custom Marker & Cluster.

@rajubhusani I'm still not able to reproduce it :( Can you debug it a little to find what the actual exception is? It seems to be happening in `Convert.java` file. If you remove the try-catch block in `getBitmapFromPinConfig()` function and run the example app again, the actual exception should be thrown and shown in the logs. Please note that there are quite many changes since your last comment

packages/google_maps_flutter/google_maps_flutter/example/lib/place_advanced_marker.dart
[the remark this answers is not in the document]
ditman: Ah yes, I didn't notice the `controller` call!

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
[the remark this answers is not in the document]
ditman: Yes, I think it's fine to leave this as is for now. I maybe wouldn't mark the class as `sealed` to leave an opportunity for others to create their own without forking the package, though!

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/marker.dart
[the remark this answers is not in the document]
ditman: Fair!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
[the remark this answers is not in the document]
ditman: Oh no, this looks like such a straightforward migration on the web. Thanks for the info about the other platforms!

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/bitmap.dart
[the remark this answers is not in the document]
stuartmorgan: I'd rather we wait to see if there's a use case that requires that; having new subclasses from outside the package is likely to end badly in general since the handlers for these are in native code.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/map_configuration.dart
```diff
@@ -112,13 +115,29 @@ class MapConfiguration {
...
   ///
   /// To clear a previously set style, set this to an empty string.
   final String? style;
 
+  /// The type of marker that the map should use.
+  ///
+  /// Advanced and legacy markers could be handled differently by platform
+  /// implementations. This property indicates which type of marker should be
+  /// used.
+  final MarkerType? markerType;
```
stuartmorgan, line 130: What does a null type indicate? If that's a distinct meaningful state, the doc comment needs to explain it. (If it's not, it shouldn't be nullable.)

stuartmorgan (commented):
Minor comments from re-reviewing the platform interface; still looking at the rest.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/markers.dart
```diff
@@ -52,73 +58,21 @@ class MarkersController<T, O> extends GeometryController {
...
+    final MarkerController<T, O> controller =
+        await createMarkerController(marker, currentMarker, gmInfoWindow);
     _markerIdToController[marker.markerId] = controller;
   }
 
+  /// Creates a [MarkerController] for a [Marker] object.
+  Future<MarkerController<T, O>> createMarkerController(
+    Marker marker,
+    T? currentMarker,
+    gmaps.InfoWindow? gmInfoWindow,
```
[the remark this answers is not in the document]
ditman: Also, suggestion, it seems that we need the `currentMarker` later on just to compute the options of the new `marker`. Would it make sense for the `createMarkerController` method to receive just the `marker`, the `markerOptions` (after computing them elsewhere) and the `gmInfoWindow`? Just a thought that might make this method signature a little bit more clear.

ditman (commented):
I just left a couple of minor comments for the web stuff, which in general looks good.

I'm trying to run the tests locally to verify that everything works as expected.

Please, note that there are some issues reported by CI that may need to be fixed before this lands.

Thanks for the contribution, web and the core plugin LGTM!

ditman:
PS, I think the web mocks need updating:

```
Launching integration_test/google_maps_controller_test.dart on Web Server in debug mode...
org-dartlang-app:/google_maps_controller_test.mocks.dart:282:16: Error: Type argument 'Object?' doesn't conform to
the bound 'JSObject' of the type variable 'T' on 'MarkersController'.
 - 'Object' is from 'dart:core'.
Try changing type arguments so that they conform to the bounds.
    implements _i2.MarkersController<Object?, Object> {
               ^
../lib/src/markers.dart:13:34: Context: This is the type variable whose bound isn't conformed to.
abstract class MarkersController<T extends JSObject, O>
                                 ^
```

And:

```
Launching integration_test/google_maps_plugin_test.dart on Web Server in debug mode...
org-dartlang-app:/google_maps_plugin_test.mocks.dart:113:21: Error: The return type of the method
'MockGoogleMapController.styles' is 'List<dynamic>', which does not match the return type, 'List<MapTypeStyle>',
of the overridden method, 'GoogleMapController.styles'.
 - 'List' is from 'dart:core'.
Change to a subtype of 'List<MapTypeStyle>'.
  List<dynamic> get styles =>
                    ^
../lib/src/google_maps_controller.dart:415:32: Context: This is the overridden method ('styles').
  List<gmaps.MapTypeStyle> get styles => _lastStyles;
                               ^
org-dartlang-app:/google_maps_plugin_test.mocks.dart:125:50: Error: The parameter 'markers' of the method
'MockGoogleMapController.debugSetOverrides' has type 'MarkersController<JSObject, Object>?', which does not match
the corresponding type, 'MarkersController<Object?, Object>?', in the overridden method,
'GoogleMapController.debugSetOverrides'.
 - 'MarkersController' is from 'package:google_maps_flutter_web/google_maps_flutter_web.dart'
 ('../lib/google_maps_flutter_web.dart').
 - 'Object' is from 'dart:core'.
Change to a supertype of 'MarkersController<Object?, Object>?', or, for a covariant parameter, a subtype.
    _i4.MarkersController<_i5.JSObject, Object>? markers,
                                                 ^
../lib/src/google_maps_controller.dart:191:8: Context: This is the overridden method ('debugSetOverrides').
  void debugSetOverrides({
       ^
```

If you're in linux, you can run the web integration tests with something like:

```
dart run /work/flutter/packages/script/tool/bin/flutter_plugin_tools.dart drive-examples --current-package --web --run-chromedriver
```

(Needs to have chrome and chromedriver in $PATH)

Or more directly, start chromedriver in port 4444, and run command: `flutter drive -d web-server --web-port=7357 --browser-name=chrome --driver test_driver/integration_test.dart --target integration_test/TEST FILE NAME.dart` inside the `example` directory.

ditman:
I tried to push a fix for the mocks, but:

```
$ git push
ERROR: Permission to CodemateLtd/packages.git denied to ditman.
fatal: Could not read from remote repository.
```

I'll try to send you a PR instead.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/advanced_marker.dart
```diff
@@ -88,29 +88,25 @@ class AdvancedMarker extends Marker {
...
-    addIfPresent('infoWindow', infoWindow.toJson());
-    addIfPresent('position', position.toJson());
-    addIfPresent('rotation', rotation);
-    addIfPresent('visible', visible);
-    addIfPresent('zIndex', zIndex);
-    addIfPresent('clusterManagerId', clusterManagerId?.value);
-    addIfPresent('collisionBehavior', collisionBehavior.index);
-    return json;
+    return <String, Object>{
+      'markerId': markerId.value,
```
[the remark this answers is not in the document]
aednlaxer: No because these fields are non-nullable. `clusterManagerIdValue` is nullable but it has a check already

packages/google_maps_flutter/google_maps_flutter_web/lib/src/markers.dart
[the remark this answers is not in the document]
aednlaxer: Great suggestion, thanks. I've implemented it and added docs

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/map_configuration.dart
aednlaxer: It only indicates that marker type wasn't set explicitly. It would be good to make it non-nullable but there could be complications with `diffFrom`, `applyDiff` and `isEmpty` (null is used to check if value is set or empty). How should it look there? If `isEmpty` ignores `markerType` value because it's always set it could break `isEmpty` check and result in missing or excessive updates (for example, controller is not updated with new map configuration if options are empty in `google_map.dart`)

stuartmorgan: Sorry about that, with just the local context of the change I forgot how this object works:

```dart
  /// Any options that aren't passed will be null, which allows this to serve
  /// as either a full configuration selection, or an update to an existing
  /// configuration where only non-null values are updated.
```

So for the unusual case of this object, nullable is indeed correct.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/markers.dart
```diff
@@ -61,15 +63,23 @@ abstract class MarkersController<T extends JSObject, O>
...
+  /// a [gmaps.Marker] or [gmaps.AdvancedMarkerElement] object. [markersOptions]
+  /// is either [gmaps.MarkerOptions] or [gmaps.AdvancedMarkerElementOptions].
+  ///
+  /// [gmInfoWindow] is marker's info window to show on tap.
   Future<MarkerController<T, O>> createMarkerController(
     Marker marker,
-    T? currentMarker,
+    O markerOptions,
     gmaps.InfoWindow? gmInfoWindow,
   );
```
ditman, line 84: Thanks for this change, I think this makes more sense now (to me, at least! :P)

aednlaxer: No problem, I think it's better this way, too 👍

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/ClusterManagersController.java
```diff
@@ -89,11 +102,22 @@ void addClusterManagers(@NonNull List<Messages.PlatformClusterManager> clusterMa
...
     ClusterManager<MarkerBuilder> clusterManager =
         new ClusterManager<MarkerBuilder>(context, googleMap, markerManager);
-    ClusterRenderer<MarkerBuilder> clusterRenderer =
-        new ClusterRenderer<MarkerBuilder>(context, googleMap, clusterManager, this);
+    initializeRenderer(clusterManager);
+    clusterManagerIdToManager.put(clusterManagerId, clusterManager);
+  }
+
+  /**
+   * Initializes cluster renderer if it hasn't been initialized before. ClusterManager uses default
```
stuartmorgan, line 110: This sounds like it would be a no-op if it had already been called, but that doesn't appear to be the case in the implementation.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -1230,7 +1280,8 @@ PlatformMapConfiguration _platformMapConfigurationFromOptionsJson(
     trafficEnabled: options['trafficEnabled'] as bool?,
     buildingsEnabled: options['buildingsEnabled'] as bool?,
     liteModeEnabled: options['liteModeEnabled'] as bool?,
-    cloudMapId: options['cloudMapId'] as String?,
+    markerType: _platformMarkerTypeFromIndex(options['markerType'] as int?),
```
stuartmorgan, line 1283: Do we actually need this helper method? I would not expect anything that supports marker types to call the deprecated method, so `markerType` should never be written.

Is there any code in any of our packages that would ever write this? If not, we can just hard-code it to `.marker`.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapController.m
```diff
@@ -726,6 +728,14 @@ - (nullable FlutterStandardTypedData *)takeSnapshotWithError:
   return imageData ? [FlutterStandardTypedData typedDataWithBytes:imageData] : nil;
 }
 
+- (nullable NSNumber *)isAdvancedMarkersAvailable:
+    (FlutterError *_Nullable __autoreleasing *_Nonnull)error {
+  NSUInteger advancedMarkerFlag =
+      self.controller.mapView.mapCapabilities & GMSMapCapabilityFlagsAdvancedMarkers;
+  BOOL boolValue = advancedMarkerFlag != 0;
+  return [NSNumber numberWithBool:boolValue];
```
stuartmorgan, line 736: Optional: `return [NSNumber numberWithBool:(advancedMarkerFlag != 0)];`. Or alternately, `return @(boolValue);`

(Don't do both though; Obj-C does weird things with box syntax and inline expressions that should evaluate to bools, which can cause type problems when run through the method channels.)

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -106,6 +106,32 @@ - (void)setZIndex:(int)zIndex {
...
+//                            screenScale:screenScale];
+//  [self setIcon:image];
+//  [self setFlat:platformMarker.flat];
+//  [self setConsumeTapEvents:platformMarker.consumeTapEvents];
+//  [self setPosition:FGMGetCoordinateForPigeonLatLng(platformMarker.position)];
+//  [self setRotation:platformMarker.rotation];
+//  [self setVisible:platformMarker.visible];
+//  [self setZIndex:platformMarker.zIndex];
+//  [self setCollisionBehavior:platformMarker.collisionBehavior];
+//}
```
stuartmorgan, line 126: Why is this commented out?

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.m
```diff
@@ -231,11 +258,58 @@ - (UIImage *)iconFromBitmap:(FGMPlatformBitmap *)platformBitmap
...
+      options.glyph = glyph;
+    }
+
+    image = [GMSPinImage pinImageWithOptions:options];
   }
 
   return image;
 }
 
+- (UIColor *)UIColorFromRGB:(NSInteger)rgbValue {
```
stuartmorgan, line 306: This already exists as `FGMGetColorForRGBA`.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
```diff
@@ -1085,7 +1142,8 @@ PlatformMapConfiguration _platformMapConfigurationFromOptionsJson(
     indoorViewEnabled: options['indoorEnabled'] as bool?,
     trafficEnabled: options['trafficEnabled'] as bool?,
     buildingsEnabled: options['buildingsEnabled'] as bool?,
-    cloudMapId: options['cloudMapId'] as String?,
+    markerType: _platformMarkerTypeFromIndex(options['markerType'] as int?),
```
stuartmorgan, line 1145: Same question here as on Android.

packages/google_maps_flutter/google_maps_flutter_ios/pigeons/messages.dart
```diff
@@ -546,6 +561,25 @@ enum PlatformMapBitmapScaling {
   none,
 }
 
+class PlatformBitmapPinConfig {
```
stuartmorgan, line 564: This should have a comment like the others indicating what other class it corresponds to.

stuartmorgan (commented):
I've re-reviewed the app-facing and platform interface packages, and skimmed Android and iOS (but those two should still get platform team reviews).

vashworth:
> Advanced Markers example on iOS doesn't show advanced markers, seems to be [a known issue](https://issuetracker.google.com/issues/370536110)

You can use Advanced Markers in example map. [You set mapId to `DEMO_MAP_ID`](https://developers.google.com/maps/documentation/ios-sdk/advanced-markers/add-marker#objective-c:~:text=Note%3A%20For%20testing%2C%20you%20can%20skip%20the%20step%20of%20creating%20and%20configuring%20a%20map%20ID%2C%20by%20using%20mapId%3A%20DEMO_MAP_ID%20in%20your%20app%20code.%20DEMO_MAP_ID%20is%20intended%20for%20testing%20purposes%20only.%20Don%27t%20use%20DEMO_MAP_ID%20in%20a%20production%20environment.)

packages/google_maps_flutter/google_maps_flutter_ios/example/ios15/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -311,6 +312,24 @@
...
+			);
+			name = "[CP] Embed Pods Frameworks";
+			outputPaths = (
+				"${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}/integration_test.framework",
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "\"${PODS_ROOT}/Target Support Files/Pods-Runner/Pods-Runner-frameworks.sh\"\n";
+			showEnvVarsInLog = 0;
+		};
```
vashworth, line 332: Please revert all changes to this file

packages/google_maps_flutter/google_maps_flutter_ios/pigeons/messages.dart
```diff
@@ -656,6 +691,10 @@ abstract class MapsApi {
 
   /// Takes a snapshot of the map and returns its image data.
   Uint8List? takeSnapshot();
+
+  /// Returns true if the map supports advanced markers
```
vashworth, line 695: Nit:
```suggestion
  /// Returns true if the map supports advanced markers.
```

packages/google_maps_flutter/google_maps_flutter_ios/pigeons/messages.dart
```diff
@@ -205,6 +206,13 @@ class PlatformMarker {
   final double zIndex;
   final String markerId;
   final String? clusterManagerId;
+  final MarkerCollisionBehavior? collisionBehavior;
+}
+
+enum MarkerCollisionBehavior {
+  required,
```
vashworth, line 213: `required` is a keyword: https://dart.dev/language/keywords

Perhaps `requiredDisplay`?

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/integration_test/google_maps_test.dart
```diff
@@ -1292,6 +1383,33 @@ void main() {
...
+
+    await tester.pumpWidget(
+      Directionality(
+        textDirection: ui.TextDirection.ltr,
+        child: ExampleGoogleMap(
+          initialCameraPosition: const CameraPosition(
+            target: LatLng(10.0, 20.0),
+          ),
+          markers: markers,
+          markerType: MarkerType.advancedMarker,
```
vashworth, line 1407: You should use a completer to wait for `onMapCreated`. See test `markerWithLegacyBytes` above for an example

vashworth (commented):
Overall, the iOS side LGTM. Left a few comments. 

Can you also add an example that sets the collision behavior?

packages/google_maps_flutter/google_maps_flutter_ios/example/shared/maps_example_dart/lib/place_marker.dart
```diff
@@ -300,6 +304,7 @@ class PlaceMarkerBodyState extends State<PlaceMarkerBody> {
         mainAxisAlignment: MainAxisAlignment.spaceEvenly,
         crossAxisAlignment: CrossAxisAlignment.stretch,
         children: <Widget>[
+          AdvancedMarkersCapabilityStatus(controller: controller),
```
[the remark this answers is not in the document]
aednlaxer: Great catch! Thanks, fixed

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
```diff
@@ -652,8 +652,9 @@ class MethodChannelGoogleMapsFlutter extends GoogleMapsFlutterPlatform {
       creationId,
       onPlatformViewCreated,
       widgetConfiguration: MapWidgetConfiguration(
-          initialCameraPosition: initialCameraPosition,
-          textDirection: textDirection),
+        initialCameraPosition: initialCameraPosition,
+        textDirection: textDirection,
```
stuartmorgan-g, line 656: Nit: this can be reverted.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/advanced_marker.dart
```diff
@@ -0,0 +1,171 @@
...
+    super.anchor,
+    super.consumeTapEvents,
+    super.draggable,
+    super.flat,
+    super.icon,
+    super.infoWindow,
+    super.position,
+    super.rotation,
+    super.visible,
+    super.zIndex,
```
stuartmorgan-g, line 31: I can't remember if this came up earlier in the review: Can we make this `int zIndex` and have it set the super `zIndex` to the int value in the constructor, so that we aren't replicating the API mistake (https://github.com/flutter/flutter/issues/155897) to this new class?

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/advanced_marker.dart
```diff
@@ -0,0 +1,171 @@
...
+    Offset? anchorParam,
+    bool? consumeTapEventsParam,
+    bool? draggableParam,
+    bool? flatParam,
+    BitmapDescriptor? iconParam,
+    InfoWindow? infoWindowParam,
+    LatLng? positionParam,
+    double? rotationParam,
+    bool? visibleParam,
+    double? zIndexParam,
```
stuartmorgan-g, line 57: And then the same change here.

stuartmorgan-g (commented):
App-facing and platform interface look good, so once the platform reviews are in this can be split apart.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/advanced_marker.dart
aednlaxer: It's possible to change this to `int zIndex` but there's a problem with changing double to int in `copyWith`. Updated `copyWith` is not a valid override of `Marker`'s `copyWith` where `zIndex` has to be double. 

One option is to keep `double? zIndex` in `copyWith` but it won't be consistent with `AdvancedMarker`'s constructor. 

Adding a new method like `copyObjectWith` doesn't sound right. Original `copyWith` will still be available and will return an instance of `Marker` when called on `AdvancedMarker`

stuartmorgan-g: Hm. I think the best option is probably to make it `int zIndex` in the constructor, and leave `copyWith` wrong, and just truncate in there. `copyWith` will be semi-fixed when we add the new int variant of zIndex to the base class (and thus to `copyWith`). Not ideal, but since we're inheriting it's probably the best option we have without a breaking change.

aednlaxer: I've made this change. There's a `.toInt()` call inside `copyWith` to convert from Marker's double to AdvancedMarker's int

aednlaxer:
@reidbaker @vashworth could you please give this another round of review? All comments should be resolved now

aednlaxer:
`Linux_web web_platform_tests_wasm_shard_3 master` is failing because [`innerHtml`](https://api.flutter.dev/flutter/dart-html/Element/innerHtml.html) returns a different object type on Flutter main:
* Flutter stable: `String?`
* Flutter main: `JSAny`

@ditman is this something that should be type-checked in tests or is there a better way to deal with it?

ditman:
@aednlaxer this has to do with (possible) differences in implementation of those types in JS vs WASM (the failing tests are WASM ones, it seems).

Have you tried compiling your demo app with `--wasm` and seeing if it actually fails, or is it a problem with the tests? If we're using `dart:html` in tests that's a problem-ish, we should move away from that and use `package:web` and `dart:js_interop` instead.

I can't take a look right now, but will try to figure out a patch later this week.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
```diff
@@ -1124,11 +1169,20 @@ PlatformMapConfiguration _platformMapConfigurationFromMapConfiguration(
...
     trafficEnabled: config.trafficEnabled,
     buildingsEnabled: config.buildingsEnabled,
-    cloudMapId: config.cloudMapId,
+    markerType: _platformMarkerTypeFromMarkerType(config.markerType),
+    mapId: config.mapId,
     style: config.style,
   );
 }
 
+PlatformMarkerType? _platformMarkerTypeFromMarkerType(MarkerType? markerType) {
```
vashworth, line 1178: Github won't let me add a comment on the line, but `collisionBehavior` needs to be added to `_platformMarkerFromMarker` function above

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -289,21 +289,32 @@ gmaps.Size? _gmSizeFromIconConfig(List<Object?> iconConfig, int sizeIndex) {
...
 }) {
-  final gmaps.Size gmapsSize = gmaps.Size(size.width, size.height);
-  icon.size = gmapsSize;
-  icon.scaledSize = gmapsSize;
+  icon.setAttribute(
+    'style',
+    <String>[
+      if (size != null) ...<String>[
+        'width: ${size.width}px;',
+        'height: ${size.height}px;',
```
ditman, line 308: I suspect the change in the test to ignore the decimals can be fixed here as well, by being more explicit with the conversion of `size.width`? Right now I think it's a `num`, so we could maybe call [`toStringAsFixed(1)`](https://api.flutter.dev/flutter/dart-core/num/toStringAsFixed.html), so this always renders the `num` with 1 decimal regardless of the compilation target?

aednlaxer: Great suggestion, thanks! That's a better solution than ignoring zeros 👍

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
aednlaxer: Great catch, thanks! I've added a test for `updateMarkers` that works with `AdvancedMarker`s and a check for `collisionBehavior` there. Do you think it's worth adding a test just for `_platformMarkerTypeFromMarkerType`?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/MarkerBuilder.java
```diff
@@ -115,6 +120,13 @@ public void setZIndex(float zIndex) {
     markerOptions.zIndex(zIndex);
   }
 
+  @Override
+  public void setCollisionBehavior(@AdvancedMarkerOptions.CollisionBehavior int collisionBehavior) {
```
camsim99, line 124: I think it may be nice to add `setCollisionBehavior` tests for the two implementation in this PR.

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -1949,6 +2040,33 @@ void googleMapsTests() {
     // https://github.com/flutter/flutter/issues/131071
     skip: true,
   );
+
+  testWidgets('markerWithPinConfig', (WidgetTester tester) async {
```
camsim99, line 2044: Is this testing anything? Can we remove it?

camsim99 (commented):
Overall, the Android implementation looks good to me! Left some questions regarding testing.

vashworth (approved):
iOS side LGTM

bridgein:
Is there any roadmap for the feature?

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_test.dart
```diff
@@ -1949,6 +2040,33 @@ void googleMapsTests() {
     // https://github.com/flutter/flutter/issues/131071
     skip: true,
   );
+
+  testWidgets('markerWithPinConfig', (WidgetTester tester) async {
```
aednlaxer: It doesn't explicitly check anything with `expect()`, it's similar to other marker tests (`markerWithLegacyBytes`, `markerWithAssetMapBitmapCreate` etc) which are "run it to make sure it doesn't throw an exception or fail in a different way" kind of tests

stuartmorgan-g:
@aednlaxer This has high-level LGs for all components now, so please feel free to split out the platform interface PR. I'll re-review that piece here in the meantime, but any comments can be addressed in the sub-PR.

stuartmorgan-g (commented):
I re-reviewed the platform interface portion, and everything looks good there :)

Piinks:
Is this change still on your radar? It looks like it has had quite the journey and a fair number of stamps, is it ready? :)

stuartmorgan-g:
This is waiting on the sub-PR at https://github.com/flutter/packages/pull/9773

stuartmorgan-g:
@aednlaxer It looks like only web was split out; is there a reason the Android and iOS implementations weren't included in that PR (or a parallel split-out PR)?

jokerttu:
> @aednlaxer It looks like only web was split out; is there a reason the Android and iOS implementations weren't included in that PR (or a parallel split-out PR)?


@stuartmorgan-g Discussed with @aednlaxer and I'll take over @aednlaxer 's work, and continue splitting the next PR's.
Already rebased to the main (not pushed yet), and currently fixing issues I found with iOS implementation before creating the PR's. Should be ready soon.

jokerttu:
Created split PR for Android. 
I will create split-PR for iOS implementation later as draft as this [known issue](https://issuetracker.google.com/issues/370536110) seems to be still visible.
I have a separate discussion open about this issue, and hopefully, I’ll get a resolution soon.

stuartmorgan-g:
@jokerttu Any update here? The referenced iOS issue appears to be have been closed for quite some time.

This intermediate state is (due to some unfortunate re-exports of platform-interface-package APIs in the app-facing package) causing some user confusion, so I would like to make sure we are on a clear path to getting the rest of this landed.

jokerttu:
> @jokerttu Any update here? The referenced iOS issue appears to be have been closed for quite some time.
> 
> This intermediate state is (due to some unfortunate re-exports of platform-interface-package APIs in the app-facing package) causing some user confusion, so I would like to make sure we are on a clear path to getting the rest of this landed.

@stuartmorgan-g 

The Android implementation is already split here: https://github.com/flutter/packages/pull/10381

### iOS: 
A PR for iOS has not been created yet due to an issue with the `GMSPinImage` not rendering on the map.
Does @caio1985 have an update on whether this issue is under investigation?

Would it be helpful if I created a draft PR for iOS now to make the issue easier to track?
 
I also created a native iOS demo app demonstrating this issue: [google-maps-advanced-markers-test-app](https://github.com/CodemateLtd/google-maps-advanced-markers-test-app)

Do you think it is worth moving forward with the app-facing package for Android and Web now, or should we wait for the iOS fix?  I worry that releasing partial support (where iOS throws `UnimplementedError`) will confuse developers even when well documented.

stuartmorgan-g:
> A PR for iOS has not been created yet due to an issue with the `GMSPinImage` not rendering on the map. Does @caio1985 have an update on whether this issue is under investigation?

I don't have any information on this, but I'm also confused because as I mentioned above, the issue you linked to and that you seem to be referring to here was closed as obsolete in April. I would not expect any active investigation to be happening on an issue that was closed 7 months ago.

> Would it be helpful if I created a draft PR for iOS now to make the issue easier to track?

A draft PR would be useful, but also a live upstream blocking issue would be helpful.

> Do you think it is worth moving forward with the app-facing package for Android and Web now, or should we wait for the iOS fix?

For now we should wait for iOS, but I also want to make sure there is some clear path forward for iOS and that it's being actively investigated, so we can make a more informed decision on that.

raphael-bmec-co:
> Do you think it is worth moving forward with the app-facing package for Android and Web now, or should we wait for the iOS fix? I worry that releasing partial support (where iOS throws UnimplementedError) will confuse developers even when well documented.

Please excuse my jumping in here - I have no idea what internal policy dictates. That said, the current production implementation has several unimplemented platform interfaces that fail silently. So while partial support is not ideal, my personal view is that it is better than the current status. Looking forward to having this landed. Thanks all.

cedvdb:
@jokerttu I was also confused, I believe you linked the [wrong issue](https://issuetracker.google.com/issues/370536110) above.

stuartmorgan-g:
From triage: This is still waiting on the sub PRs.

jokerttu:
While the platform-specific PRs are still pending, I’ve updated this package to the latest state to make cross-platform testing against the most recent platform implementations easier.
Web implementation does not have proper example app to test against, so manual testing with app-facing package is the only way to validate the behaviour.

jokerttu:
@cedvdb 
> @jokerttu I was also confused, I believe you linked the [wrong issue](https://issuetracker.google.com/issues/370536110) above.

The [issue](https://issuetracker.google.com/issues/370536110) is now re-opened.

stuartmorgan-g:
Thanks for the update!

stuartmorgan-g:
From triage: @jokerttu Have you been able to discuss with anyone on the SDK side about the outlook for the blocking issue?

caio1985:
> From triage: @jokerttu Have you been able to discuss with anyone on the SDK side about the outlook for the blocking issue?

Unfortunately we have an internal issue opened but it hasn't been picked up by anyone yet.

stuartmorgan-g:
If it's an SDK-level issue that affects everyone, not just the Flutter plugin, maybe we should consider just moving forward with landing the implementation and documenting the known issue for iOS. That way we won't continue to block availability of this feature for other platforms.

stuartmorgan-g:
> If it's an SDK-level issue that affects everyone, not just the Flutter plugin, maybe we should consider just moving forward with landing the implementation and documenting the known issue for iOS.

Capturing discussion elsewhere here: this is the direction we decided to go. The iOS implementation is now published, and I've forked this PR as https://github.com/flutter/packages/pull/11188 with minor updates so that we can get the feature out to everyone.

stuartmorgan-g:
@jokerttu Sorry, I didn't realize you were actively working on this PR. If you want to pull in my other changes from https://github.com/flutter/packages/pull/11188/commits we can approve and land this.

jokerttu:
> @jokerttu Sorry, I didn't realize you were actively working on this PR. If you want to pull in my other changes from https://github.com/flutter/packages/pull/11188/commits we can approve and land this.

@stuartmorgan-g 
I also missed your work... I cherry picked missing changes from your pr. Thank you!

stuartmorgan-g (approved):
LGTM!

PR #7892, opened 2024-10-18
[pigeon] adds event channel support for kotlin and swift

tarrinneal:
adds event channel support for kotlin and swift 
work towards https://github.com/flutter/flutter/issues/66711
adds sealed classes with extensions (empty base classes only)
fixes https://github.com/flutter/flutter/issues/155859 (Fix a small inconsistency with Pigeon docs)
adds some convenience methods to Root
fixes generation/format tests to include test pigeons
Makes swift codec class names upper camel case


packages/pigeon/platform_tests/shared_test_plugin_code/lib/integration_tests.dart
```diff
@@ -2840,6 +2840,57 @@ void runPigeonIntegrationTests(TargetGenerator targetGenerator) {
...
+        case EnumEvent():
+          expect(event.value, EventEnum.fortyTwo);
+          expect(count, 5);
+          count++;
+        case ClassEvent():
+          expect(event.value.aNullableInt, 0);
+          expect(count, 6);
+          count++;
+      }
+    });
```
stuartmorgan, line 2892: Don't we need an await on a completer that is completed in the last expected event handler, or something along those lines, so that this test doesn't just immediately finish without having run any of the expectations?

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/TestPlugin.kt
```diff
@@ -867,3 +872,57 @@ class TestPluginWithSuffix : HostSmallApi {
...
+  override fun onListen(p0: Any?, sink: PigeonEventSink<Long>) {
+    var count: Long = 0
+    val r: Runnable =
+        object : Runnable {
+          override fun run() {
+            handler.post {
+              sink.success(count)
+              count++
+            }
+            handler.postDelayed(this, 1000)
```
stuartmorgan, line 888: Is this for an automated test? If so, this is an extremely long delay. (If not, can we comment to say what it's for?)

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/TestPlugin.kt
```diff
@@ -867,3 +872,57 @@ class TestPluginWithSuffix : HostSmallApi {
...
+        object : Runnable {
+          override fun run() {
+            if (count >= eventList.size) {
+              sink.endOfStream()
+            } else {
+              handler.post {
+                sink.success(eventList[count])
+                count++
+              }
+              handler.postDelayed(this, 10)
```
stuartmorgan, line 922: Do we actually need a delay here? It would be nice not to add artificial delays to the tests.

packages/pigeon/platform_tests/test_plugin/ios/Classes/TestPlugin.swift
```diff
@@ -1214,14 +1217,63 @@ public class TestPluginWithSuffix: HostSmallApi {
...
 }
 
+class SendInts: StreamIntsStreamHandler {
+  var timerActive = false
+
+  override func onListen(withArguments arguments: Any?, sink: PigeonEventSink<Int64>) {
+    var count: Int64 = 0
+    if !timerActive {
+      timerActive = true
+      Timer.scheduledTimer(withTimeInterval: 0.01, repeats: true) { _ in
```
stuartmorgan, line 1227: These timers never end, so it'll spam the runloop with no-ops for as long as the process exists. I think you probably want `DispatchQueue.main.async` instead.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/TestPlugin.kt
stuartmorgan, line 888: Presumably you only want to do this if `count` is less than 5.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/TestPlugin.kt
```diff
@@ -867,3 +872,57 @@ class TestPluginWithSuffix : HostSmallApi {
...
+    val r: Runnable =
+        object : Runnable {
+          override fun run() {
+            handler.post {
+              sink.success(count)
+              count++
+            }
+            handler.postDelayed(this, 1000)
+            if (count >= 5) {
+              sink.endOfStream()
```
stuartmorgan, line 890: Doesn't this need to be done via `handler` so it's on the main thread?

packages/pigeon/platform_tests/test_plugin/macos/Classes/.gitignore
```diff
@@ -6,4 +6,5 @@
 # Keeping this makes it easier to review changes to ProxyApi generation.
 !ProxyApiTests.gen.swift
 # Contains the class declartions for testing ProxyApis.
-!ProxyApiTestClass.swift
\ No newline at end of file
+!ProxyApiTestClass.swift
+!EventChannelTests.gen.swift
```
stuartmorgan, line 10: Comment?

packages/pigeon/tool/shared/generation.dart
```diff
@@ -48,60 +48,79 @@ String _javaFilenameForName(String inputName) {
...
 
 Future<int> generateExamplePigeons() async {
-  return runPigeon(
+  int success = 0;
+  success = await runPigeon(
     input: './example/app/pigeons/messages.dart',
     basePath: './example/app',
     suppressVersion: true,
   );
+  success = await runPigeon(
```
stuartmorgan, line 57: You haven't checked the previous result; if the call above failed, this will continue, and then throw away that failure information.

packages/pigeon/tool/shared/generation.dart
```diff
@@ -48,60 +48,79 @@ String _javaFilenameForName(String inputName) {
...
+    'flutter_unittests':
+        <GeneratorLanguage>{}, // Only for Dart unit tests in shared_test_plugin_code
+    'message': <GeneratorLanguage>{},
+    'multiple_arity': <GeneratorLanguage>{},
+    'non_null_fields': <GeneratorLanguage>{},
+    'null_fields': <GeneratorLanguage>{},
+    'nullable_returns': <GeneratorLanguage>{},
+    'primitive': <GeneratorLanguage>{},
+    'proxy_api_tests': <GeneratorLanguage>{},
+  };
```
stuartmorgan, line 91: It would probably be easier to keep the set as-is, and then have a separate to-skip map with entries only where the skip set is non-empty, since we expect skipping to be exceptional rather than something most entries in the set will need.

script/tool/lib/src/podspec_check_command.dart
```diff
@@ -110,6 +110,11 @@ class PodspecCheckCommand extends PackageLoopingCommand {
   }
 
   Future<List<File>> _podspecsToLint(RepositoryPackage package) async {
+    // Since the pigeon podspecs require generated files that are not included in git,
+    // the podspec lint fails.
+    if (package.displayName == 'packages/pigeon') {
```
stuartmorgan, line 115: It's only the platform_tests, right? I would prefer to make this more targeted, filtering out any results that contain `packages/pigeon/platform_tests/` in the output, because the more specific we make anything that skips CI steps, the less chance of us accidentally having an unintended skip that we never notice (until it bites us).

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
```diff
@@ -37,16 +37,41 @@ private class PigeonFlutterApi {
...
   }
 }
 // #enddocregion swift-class-flutter
 
-@UIApplicationMain
+// #docregion swift-class-event
+class SendEvents: StreamEventsStreamHandler {
+  var timerActive = false
+
+  override func onListen(withArguments arguments: Any?, sink: PigeonEventSink<SealedBaseClass>) {
```
hellohuanlin, line 53: "Sealed" name coming from Kotlin? I've never seen this usage in Swift

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
```diff
@@ -0,0 +1,179 @@
...
+  return value is NSNull || value == nil
+}
+
+private func nilOrValue<T>(_ value: Any?) -> T? {
+  if value is NSNull { return nil }
+  return value as! T?
+}
+
+/// Generated class from Pigeon that represents data sent in messages.
+/// This class should not be extended by any user class outside of the generated file.
```
hellohuanlin, line 27: nit: protocol, not a class

packages/pigeon/example/app/lib/main.dart
```diff
@@ -114,6 +123,27 @@ class _MyHomePageState extends State<MyHomePage> {
...
+              stream: streamEvents(),
+              builder: (BuildContext context,
+                  AsyncSnapshot<SealedBaseClass> snapshot) {
+                if (snapshot.hasData) {
+                  String newString = '';
+                  switch (snapshot.data) {
+                    case null:
+                      newString = 'null, ';
+                    case IntEvent():
+                      newString = '${(snapshot.data! as IntEvent).data}, ';
```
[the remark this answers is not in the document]
tarrinneal: Still needs the casts :(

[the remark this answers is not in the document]
tarrinneal: and the force-unwrap

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/TestPlugin.kt
tarrinneal: That's just a leftover from some debugging I was doing. Should be much shorter

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
tarrinneal: It's coming from dart, it's semantically meaningless here, but it has to have the same name in all locations.

packages/pigeon/tool/shared/generation.dart
```diff
@@ -20,10 +20,20 @@ enum GeneratorLanguage {
   swift,
 }
 
+const Set<GeneratorLanguage> allButSwiftAndDart = <GeneratorLanguage>{
```
stuartmorgan, line 23: And Kotlin?

packages/pigeon/tool/shared/generation.dart
```diff
@@ -20,10 +20,20 @@ enum GeneratorLanguage {
...
+  GeneratorLanguage.objc,
+};
+
 // A map of pigeons/ files to the languages that they can't yet be generated
 // for due to limitations of that generator.
 const Map<String, Set<GeneratorLanguage>> _unsupportedFiles =
-    <String, Set<GeneratorLanguage>>{};
+    <String, Set<GeneratorLanguage>>{
+  'event_channel_tests': allButSwiftAndDart,
+  'proxy_api_tests': allButSwiftAndDart,
```
stuartmorgan, line 35: I'm not sure factoring out a shared constant is actually useful here, given that we could easily re-diverge them (e.g., by adding event channel support for Windows and Linux).

packages/pigeon/tool/shared/test_suites.dart
```diff
@@ -483,7 +483,7 @@ Future<int> _runCommandLineTests({bool ciMode = false}) async {
 
   int exitCode = 0;
   for (final List<String> arguments in testArguments) {
-    print('Testing dart $pigeonScript ${arguments.join(', ')}');
+    print('Testing Dart $pigeonScript ${arguments.join(', ')}');
```
stuartmorgan, line 486: Looks like you did a replace-all? This change is incorrect.

packages/pigeon/tool/test.dart
```diff
@@ -7,7 +7,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 /// Script for executing the Pigeon tests
 ///
-/// usage: dart run tool/test.dart
+/// usage: Dart run tool/test.dart
```
stuartmorgan, line 10: Same for both changes in this file.

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
```diff
@@ -37,16 +37,67 @@ private class PigeonFlutterApi {
...
+  }
+}
+// #enddocregion swift-class-event
+
+func sendEvents(_ eventListener: EventListener) {
+  var timerActive = false
+  var timer: Timer?
+  var count: Int64 = 0
+  if !timerActive {
+    timerActive = true
```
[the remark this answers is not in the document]
tarrinneal: This method activates 2 times on startup, this is to prevent that. I'm not certain why, maybe there's a better way, maybe it doesn't matter, maybe it just shouldn't be in the example.

packages/pigeon/example/README.md
```diff
@@ -351,6 +351,98 @@ pigeon_example_package_message_flutter_api_flutter_method(
...
+        yield '$intData, ';
+      case StringEvent():
+        final String stringData = event.data;
+        yield '$stringData, ';
+    }
+  }
+}
+```
+
+### Swift
```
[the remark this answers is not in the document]
tarrinneal: I did mean to add that, my bad.

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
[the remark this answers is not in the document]
stuartmorgan: How does a local variable change the behavior of multiple calls to the method?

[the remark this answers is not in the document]
tarrinneal: It was a problem before I refactored the code here, let me check again.

packages/pigeon/tool/shared/generation.dart
tarrinneal: Dart was supposed to be kotlin, lol

packages/pigeon/tool/shared/generation.dart
tarrinneal: Guess I expected it might be used more than this, but fair point.

packages/pigeon/platform_tests/shared_test_plugin_code/lib/example_app.dart
```diff
@@ -26,6 +26,7 @@ class ExampleApp extends StatefulWidget {
 class _ExampleAppState extends State<ExampleApp> {
   late final HostIntegrationCoreApi api;
   String status = 'Calling...';
+  String allData = '';
```
stuartmorgan, line 29: What is this for?

packages/pigeon/example/README.md
```diff
@@ -351,6 +351,93 @@ pigeon_example_package_message_flutter_api_flutter_method(
...
+
+This example gives a basic overview of how to use Pigeon to set up an Event Channel method.
+
+### Dart input
+
+<?code-excerpt "pigeons/event_channel_messages.dart (event-definitions)"?>
+```dart
+@EventChannelApi()
+abstract class EventApi {
+  SealedBaseClass streamEvents();
```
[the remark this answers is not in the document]
stuartmorgan: This isn't resolved. There shouldn't be a class called `SealedBaseClass` in the example, it should be called `PlatformEvent`, and the `@EventChannelApi` class should be given a different name.

stuartmorgan (commented):
WIP re-review; posting in case you want to update the class name while I'm continuing the review.

packages/pigeon/lib/generator_tools.dart
```diff
@@ -801,6 +809,24 @@ String toUpperCamelCase(String text) {
...
+  bool firstWord = true;
+  return text.split(separatorPattern).map((String word) {
+    if (word.isEmpty) {
+      return '';
+    }
+    if (firstWord) {
+      firstWord = false;
+      return word.substring(0, 1).toLowerCase() + word.substring(1);
+    }
+    return word.isEmpty
```
stuartmorgan, line 824: The `isEmpty` codepath is dead code now, due to the new early return up top. So this can just `return word.substring[...]`.

stuartmorgan (approved):
LGTM

PR #7935, opened 2024-10-26
Bump `path_provider_android` to `^2.2.5` due to Android v1 embedder.

matanlurey:
https://github.com/flutter/engine/pull/52022 removed the Android v1 embedder.

I'm seeing locally that older versions of `path_provider_android` fail due to missing symbols:
```txt
flutter run                   
Launching lib/main.dart on sdk gphone64 arm64 in debug mode...
You are applying Flutter's main Gradle plugin imperatively using the apply script method, which is deprecated and will be removed in a future release. Migrate to applying Gradle plugins with the declarative plugins block: https://flutter.dev/to/flutter-gradle-plugin-apply

Warning: This project is still reading the deprecated '.flutter-plugins. file.
In an upcoming stable release support for this file will be completely removed and your build will fail.
See https:/flutter.dev/to/flutter-plugins-configuration.
/Users/matanl/.pub-cache/hosted/pub.dev/path_provider_android-2.2.2/android/src/main/java/io/flutter/plugins/pathprovider/PathProviderPlugin.java:39: error: cannot find symbol
      @NonNull io.flutter.plugin.common.PluginRegistry.Registrar registrar) {
                                                      ^
  symbol:   class Registrar
  location: interface PluginRegistry
1 error

FAILURE: Build failed with an exception.
```

To be a good citizen, let's make sure users of `path_provider` get a good `path_provider_android`.

### DIFF
--- a/packages/path_provider/path_provider/CHANGELOG.md
+++ b/packages/path_provider/path_provider/CHANGELOG.md
@@ -1,6 +1,7 @@
-## NEXT
+## 2.1.5
 
 * Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
+* Updates minimum required `path_provider_android` to [2.2.5](https://pub.dev/packages/path_provider_android/changelog#225) as the v1 Android embedder is no longer available.
 
 ## 2.1.4
 
--- a/packages/path_provider/path_provider/pubspec.yaml
+++ b/packages/path_provider/path_provider/pubspec.yaml
@@ -5,7 +5,7 @@ issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+
 version: 2.1.5
 
 environment:
-  sdk: ^3.3.0
+  sdk: ^3.4.0
   flutter: ">=3.19.0"
 
 flutter:
--- a/packages/path_provider/path_provider/pubspec.yaml
+++ b/packages/path_provider/path_provider/pubspec.yaml
@@ -6,7 +6,7 @@ version: 2.1.5
 
 environment:
   sdk: ^3.4.0
-  flutter: ">=3.19.0"
+  flutter: ">=3.22.0"
 
 flutter:
   plugin:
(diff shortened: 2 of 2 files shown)


stuartmorgan (approved):
You'll need to update the min SDK version of `path_provider`, as [the min supported SDK can't be less that the min supported SDK resolvable by a required dependency]. LGTM with that change.

(We don't normally update min required versions just to "advertise" bug fixes in implementation dependencies, but we have done it for severe cases before, like when a new version of an implementation package is required for app store compliance, so it's reasonable to do here.)

vbuberen:
@stuartmorgan Just saw it and comparing with my closed one #7938 I am curios what is the difference? 
Just like you told in my PR using the latest Flutter with removed embedder v1 (when it gets released) is also a `developer choice` and not an actual requirement. So what is the different with the case in the issue I tried to fix where latest Android Studio and IntelliJ have JDK set to 21 by default which you called `developer choice`? 

https://github.com/flutter/packages/pull/7938#issuecomment-2440226906

It feels like there are double standards that somebody from the team is allowed to do some things, while another person not even when the issue had enough examples with other people experiencing the same problem.

stuartmorgan:
> @stuartmorgan Just saw it and comparing with my closed one #7938 I am curios what is the difference?
> [...]
> It feels like there are double standards that somebody from the team is allowed to do some things, while another person not

It appears that you have already (and incorrectly) decided what you believe the real answer is, which makes constructive conversation very difficult. If you would like to discuss your issue further in your issue, and you can do so within the bounds of our code of conduct, please feel free to post questions there.

vbuberen:
> It appears that you have already (and incorrectly) decided what you believe the real answer is, which makes constructive conversation very difficult.

You can hide this comment as well, but you didn't even bother providing an answer what is different between this PR and the one that you closed while they are clearly falling under the same case that the issue might happen due to developer's choice as you described in that other PR. 
I appreciate your opinion and explanation you provided in my issue regarding artificial and actual minimums. Moreover, I already accepted your decision regarding that PR, but after seeing this PR and the approving comment I have an impression which I shared. 

Considering that you already made a judgement about what I decided and what not, I would like to get an opinion regarding [that issue](https://github.com/flutter/packages/pull/7938) from some other maintainer, who is not biased the same way.

PR #6443, opened 2024-03-31
[camera_web] Support for camera stream on web

TecHaxter:
This PR aims to provide support for strartImageStream and stopImageStream on Web.

[#92460](https://github.com/flutter/flutter/issues/92460)

Based on [#6944](https://github.com/flutter/plugins/pull/6944) from the archived [plugins](https://github.com/flutter/plugins) repository

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

POV: my first PR on a public repo

PR #7950, opened 2024-10-28
[camera_web] Re: Support for camera stream on web

TecHaxter:
This PR aims to provide support for strartImageStream and stopImageStream on Web.

[#92460](https://github.com/flutter/flutter/issues/92460)

Based on [#6944](https://github.com/flutter/plugins/pull/6944) from the archived [plugins](https://github.com/flutter/plugins) repository

## Pre-launch Checklist

- [x] I signed the [CLA].

POV: my first PR on a public repo
Contains required commits from the PR https://github.com/flutter/packages/pull/6443, resolves unnecessary 202 commits


ditman:
I'll re-review this tomorrow!

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -638,4 +643,74 @@ class Camera {
...
+  }
+
+  /// Triggers animation frames in a loop as long as
+  /// [_cameraFrameStreamController.hasListener] and executes the callback
+  void _triggerAnimationFramesLoop(
+    VoidCallback action, {
+    required int fps,
+  }) {
+    Completer<void> completer = Completer<void>();
+    completer.complete();
```
ditman, line 677: Why do we need this? Remove?

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -638,4 +643,74 @@ class Camera {
...
+
+    void onAnimate(num _) {
+      if (!_cameraFrameStreamController.hasListener) {
+        return;
+      }
+
+      if (!completer.isCompleted) {
```
ditman, line 684: This is always false, isn't it? The `completer` is being `.complete()`d unconditionally in line 677?

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -638,4 +643,74 @@ class Camera {
...
+    VoidCallback action, {
+    required int fps,
+  }) {
+    Completer<void> completer = Completer<void>();
+    completer.complete();
+
+    void onAnimate(num _) {
+      if (!_cameraFrameStreamController.hasListener) {
+        return;
+      }
```
ditman, line 682: Do not control the stop of the animation like this. What you need to do is to keep the return value of `window.requestAnimationFrame` (which is an `int`), and then when the stream controller ceases to have any listeners, [call `window.cancelAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame) with the last ID returned by rAF.

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -363,4 +364,78 @@ class CameraService {
...
+      format: const CameraImageFormat(
+        ImageFormatGroup.jpeg,
+        raw: '',
+      ),
+      planes: <CameraImagePlane>[
+        CameraImagePlane(
+          bytes: byteBuffer.asUint8List(),
+          bytesPerRow: 0,
+        ),
+      ],
```
ditman, line 436: Is this `CameraImageData` correctly constructed?

First it says that this is a `jpeg` (which it isn't), and then puts it in the first plane, but we say it's "0 bytesPerRow".

My suggestions are:

* Make the format `ImageFormatGroup.unknown` (it's RGB with 4 bytes per-pixel, arranged in rows, but that's not an expected ImageFormatGroup for the plugin).
* set a `raw` value that [doesn't clash with anything possible with Android or iOS](https://github.com/flutter/packages/blob/e55be9676ec6d519598dc716c5ec8c6dcb0e7973/packages/camera/camera_platform_interface/lib/src/types/camera_image_data.dart#L57-L66) .
* The `bytesPerRow` of the browser's `getImageData` seems to be `width * 4`. [Docs](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Pixel_manipulation_with_canvas#the_imagedata_object).

ditman (dismissed):
Thanks for the contribution @TecHaxter! This is getting close. I left some comments that should improve the performance of the frame capture (maybe hit 60fps!) and the accuracy of the image data reported.

Have you used this code in any demo app compiled for the web?

packages/camera/camera_web/example/integration_test/camera_bitrate_test.dart
```diff
@@ -106,6 +106,10 @@ void main() {
       ).thenAnswer(
           (_) => Future<MediaStream>.value(canvasElement.captureStream()));
 
+      when(
+        () => cameraService.hasPropertyOffScreenCanvas(),
```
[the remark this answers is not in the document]
TecHaxter: The `cameraService` is a mock class object itself.
If we don't stub `hasPropertyOffScreenCanvas`, it throws an error - _"TypeError: null: type 'Null' is not a subtype of type 'bool'_

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -1704,5 +1709,56 @@ void main() {
...
+            const Size(10, 10),
+          );
+
+          final Camera camera = Camera(
+            textureId: textureId,
+            cameraService: cameraService,
+          )..videoElement = videoElement;
+
+          when(
+            () => cameraService.takeFrame(
```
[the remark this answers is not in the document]
TecHaxter: The cameraService is a mock class object itself.
If we don't stub hasPropertyOffScreenCanvas, it throws an error - "TypeError: null: type 'Null' is not a subtype of type 'bool'

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: I have changed the implementation of `_triggerAnimationFramesLoop` in upcoming commits.

packages/camera/camera_web/example/integration_test/camera_web_test.dart
```diff
@@ -100,6 +100,10 @@ void main() {
         (_) async => videoElement.captureStream(),
       );
 
+      when(
+        () => cameraService.hasPropertyOffScreenCanvas(),
+      ).thenAnswer((_) => true);
```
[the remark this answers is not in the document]
TecHaxter: The `cameraService` is a mock class object itself.
If we don't stub `hasPropertyOffScreenCanvas`, it throws an error - "TypeError: null: type 'Null' is not a subtype of type 'bool'

packages/camera/camera_web/lib/src/camera_service.dart
TecHaxter: I found `FLEX_RGBA_8888` in Android and `kCVPixelFormatType_32RGBA` in iOS which looks promising for compatibility with RGBA 4 bytes per-pixel arranged image coming from `ImageData` of Canvas (Although I haven't tested it yet)

Also found a comment in `\camera\camera_android_camerax\lib\src\android_camera_camerax.dart`
```dart
  /// [imageFormatGroup] is used to specify the image format used for image
  /// streaming, but CameraX currently only supports YUV_420_888 (supported by
  /// Flutter) and RGBA (not supported by Flutter). CameraX uses YUV_420_888
  /// by default, so [imageFormatGroup] is not used.
``` 

Seeing this, I went with the values you suggested

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -638,4 +643,74 @@ class Camera {
...
+
+      // Reset the completer and set up a delay
+      completer = Completer<void>();
+      Future<void>.delayed(
+        Duration(milliseconds: 1000 ~/ fps),
+      ).then((_) {
+        completer.complete();
+        // Schedule the next frame after the delay
+        window.requestAnimationFrame(onAnimate.toJS);
+      });
```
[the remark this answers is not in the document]
TecHaxter: In recent commits, I have made `onAnimate` function in `_triggerAnimationFramesLoop` similar to this [example ](https://stackoverflow.com/a/19772220). Although when I tested [Fiddle](https://jsfiddle.net/chicagogrooves/nRpVD/2/), it JS it was giving accurate FPS but when I tested it in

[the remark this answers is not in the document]
ditman: I'll take a look, I simplified your fiddle a little bit, check it out:

* https://jsfiddle.net/ditman/h50m4s6w/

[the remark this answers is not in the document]
TecHaxter: Thanks @ditman  for sharing [https://jsfiddle.net/ditman/h50m4s6w/](https://jsfiddle.net/ditman/h50m4s6w/)
I have added a new commit which implements the logic in `animate` function, It works like charm ✨✨
Getting the exact same FPS with this 🤩

packages/camera/camera_web/CHANGELOG.md
```diff
@@ -1,113 +1,111 @@
-## NEXT
+## 0.3.6
 
-* Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+- Supporting camera image stream on web.
+- Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
```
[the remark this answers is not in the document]
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/example/integration_test/camera_bitrate_test.dart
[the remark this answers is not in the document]
TecHaxter: Can we mark this resolved?

packages/camera/camera_web/example/integration_test/camera_service_test.dart
```diff
@@ -903,5 +904,69 @@ void main() {
...
+          final web.VideoElement videoElement =
+              getVideoElementWithBlankStream(size)
+                ..onLoadedMetadata.listen((_) {
+                  completer.complete();
+                })
+                ..load();
+          await completer.future;
+          final CameraImageData cameraImageData = cameraService.takeFrame(
+            videoElement,
+            canUseOffscreenCanvas: true,
```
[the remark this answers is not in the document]
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -61,6 +62,10 @@ void main() {
           cameraId: any(named: 'cameraId'),
         ),
       ).thenAnswer((_) => Future<MediaStream>.value(mediaStream));
+
+      when(
+        () => cameraService.hasPropertyOffScreenCanvas(),
+      ).thenAnswer((_) => true);
```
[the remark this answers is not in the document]
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/example/integration_test/camera_test.dart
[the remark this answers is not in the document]
TecHaxter: Can we mark this resolved?

packages/camera/camera_web/example/integration_test/camera_web_test.dart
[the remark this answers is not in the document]
TecHaxter: Can we mark this resolved?

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -200,6 +203,8 @@ class Camera {
         onEndedController.add(defaultVideoTrack);
       });
     }
+
+    canUseOffscreenCanvas = _cameraService.hasPropertyOffScreenCanvas();
```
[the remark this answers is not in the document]
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/lib/src/camera.dart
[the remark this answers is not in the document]
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -363,4 +364,78 @@ class CameraService {
...
+  ///Used in [takeFrame] if [canUseOffscreenCanvas] is false
+  web.CanvasElement? _canvasElement;
+
+  ///Used in [takeFrame] if [canUseOffscreenCanvas] is false
+  web.OffscreenCanvas? _offscreenCanvas;
+
+  ///Returns frame at a specific time using video element
+  CameraImageData takeFrame(
+    web.VideoElement videoElement, {
+    bool canUseOffscreenCanvas = false,
```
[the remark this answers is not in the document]
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -363,4 +364,78 @@ class CameraService {
...
+    if (canUseOffscreenCanvas) {
+      if (_offscreenCanvas == null ||
+          _offscreenCanvas!.width != width ||
+          _offscreenCanvas!.height != height) {
+        _offscreenCanvas = web.OffscreenCanvas(width, height);
+      }
+      final web.OffscreenCanvasRenderingContext2D context =
+          _offscreenCanvas!.getContext(
+        '2d',
+        <String, dynamic>{'willReadFrequently': true}.toJSBox,
```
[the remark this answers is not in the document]
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -6,7 +6,6 @@ import 'dart:async';
 import 'dart:js_interop';
 import 'dart:ui';
 import 'dart:ui_web' as ui_web;
-
```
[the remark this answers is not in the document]
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: This is fixed, can be marked as resolved?

packages/camera/camera_web/lib/src/camera_service.dart
TecHaxter: I have tested this in the example app of [google_mediapipe_face_detection](https://github.com/TecHaxter/google_mediapipe_face_detection) (code not pushed yet). It works like charm ✨✨

Can we mark this resolved?

TecHaxter:
Hi @ditman The PR is ready for another review, Thanks :)

stuartmorgan:
From triage: Ping @ditman on this re-review.

stuartmorgan:
Update from triage: we're working through trying to find a web reviewer with availability to look at this. Apologies for the delay.

Piinks:
@yjbanov can you suggest a reviewer for this change? It has been blocked on a web review.

stuartmorgan-g:
From triage: @mdebbar What's the status of the web review here?

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -363,4 +364,77 @@ class CameraService {
...
+      if (_offscreenCanvas == null ||
+          _offscreenCanvas!.width != width ||
+          _offscreenCanvas!.height != height) {
+        _offscreenCanvas = web.OffscreenCanvas(width, height);
+      }
+      final web.OffscreenCanvasRenderingContext2D context =
+          _offscreenCanvas!.getContext(
+        '2d',
+        <String, Object?>{'willReadFrequently': true}.jsify(),
+      )! as web.OffscreenCanvasRenderingContext2D;
```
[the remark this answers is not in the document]
mdebbar: @gemini-code-assist the `2d` context is supported on virtually all browsers: https://caniuse.com/mdn-api_htmlcanvaselement_getcontext_2d_context

packages/camera/camera_web/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 0.3.6
 
+* Supporting camera image stream on web.
```
mdebbar: To keep this consistent with the rest of the file:
```suggestion
* Adds support for camera image stream on web.
```

packages/camera/camera_web/example/integration_test/camera_service_test.dart
```diff
@@ -903,5 +904,103 @@ void main() {
...
+              window,
+              'OffscreenCanvas'.toJS,
+            ),
+          ).thenReturn(false);
+          final bool hasNotOffScreenCanvas =
+              cameraService.hasPropertyOffScreenCanvas();
+          expect(
+            hasNotOffScreenCanvas,
+            false,
+          );
```
mdebbar: Can use a loop to simplify:
```suggestion
          for (final supportsOffscreenCanvas in [true, false]) {
            when(
              () => jsUtil.hasProperty(
                window,
                'OffscreenCanvas'.toJS,
              ),
            ).thenReturn(value);
            final bool hasOffScreenCanvas =
                cameraService.hasPropertyOffScreenCanvas();
            expect(
              cameraService.hasPropertyOffScreenCanvas(),
              supportsOffscreenCanvas ? isTrue : isFalse,
            );
          }
```

packages/camera/camera_web/example/integration_test/camera_service_test.dart
```diff
@@ -903,5 +904,103 @@ void main() {
...
+        },
+      );
+      testWidgets(
+        'returns Camera Image of Size '
+        'when videoElement is of Size '
+        'when browser supports OffscreenCanvas',
+        (WidgetTester widgetTester) async {
```
mdebbar: This test and the one below are basically the same, the only difference is the boolean value of the `hasPropertyOffScreenCanvas` mock.

Combine both into a single test:
```dart
      testWidgets(
        'returns Camera Image of Size '
        'when videoElement is of Size '
        'regardless of OffscreenCanvas support',
        (WidgetTester widgetTester) async {
          for (final supportsOffscreenCanvas in [true, false]) {
            ...
          }
        }
```

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -1704,5 +1709,107 @@ void main() {
         });
       });
     });
+    group('cameraFrameStream', () {
+      testWidgets(
+        'CameraImageData bytes is a multiple of 4 '
+        'when browser supports OffscreenCanvas',
+        (WidgetTester tester) async {
```
mdebbar: Can be combined with the test below, similar to my previous comment.

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -159,6 +160,10 @@ class Camera {
   final StreamController<VideoRecordedEvent> videoRecorderController =
       StreamController<VideoRecordedEvent>.broadcast();
 
+  /// Used to check if allowed to paint canvas off screen
+  @visibleForTesting
+  bool canUseOffscreenCanvas = false;
```
mdebbar: Looks like this never changes, so let's make it final:
```suggestion
  final bool canUseOffscreenCanvas;
```

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -638,4 +643,70 @@ class Camera {
       ..height = '100%'
       ..objectFit = 'cover';
   }
+
+  final StreamController<CameraImageData> _cameraFrameStreamController =
+      StreamController<CameraImageData>.broadcast();
+
+  // TODO(replace): introduced fps in
+  /// [CameraImageStreamOptions]
```
mdebbar: Please follow the style guide for TODOs: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-checking-in-comments-that-ask-questions

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -638,4 +643,70 @@ class Camera {
...
+  }
+
+  /// Triggers animation frames in a loop at a specified FPS
+  /// as long as [animationFrameId] is not cancelled
+  void _triggerAnimationFramesLoop(
+    VoidCallback action, {
+    required int fps,
+  }) {
+    int? animationFrameId;
+    num then = 0, fpsInterval = 1000 / fps;
```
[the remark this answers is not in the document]
mdebbar: and let's give `then` a more descriptive name like `lastFrameTimestamp`.

mdebbar (commented):
This looks really close to being ready to land. I just some minor polishing comments.

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: I have updated the TODO comment. But haven't created an issue for that. Should I create one?

Update: I've created an issue and included that in the comment.

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: This is addressed now.

packages/camera/camera_web/example/integration_test/camera_test.dart
TecHaxter: I've taken care of this.

packages/camera/camera_web/example/integration_test/camera_service_test.dart
TecHaxter: This has been resolved now.

packages/camera/camera_web/example/integration_test/camera_service_test.dart
TecHaxter: Just so you know – this has been addressed now.

packages/camera/camera_web/CHANGELOG.md
TecHaxter: This is addressed.

packages/camera/camera_web/lib/src/camera.dart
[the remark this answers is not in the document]
TecHaxter: I don't think this TODO needs to be addressed in this PR, because PR checks would fail if it would contain changes in more than one package.

packages/camera/camera_web/example/integration_test/camera_test.dart
```diff
@@ -1704,5 +1709,107 @@ void main() {
...
+            cameraImageData,
+            equals(
+              isA<CameraImageData>().having(
+                (CameraImageData e) => e.planes.first.bytes.length % 4,
+                'bytes',
+                equals(0),
+              ),
+            ),
+          );
+        },
```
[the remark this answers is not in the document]
TecHaxter: This has been addressed.

packages/camera/camera_web/example/integration_test/camera_service_test.dart
```diff
@@ -903,5 +904,103 @@ void main() {
...
+              'OffscreenCanvas'.toJS,
+            ),
+          ).thenReturn(false);
+          final bool hasNotOffScreenCanvas =
+              cameraService.hasPropertyOffScreenCanvas();
+          expect(
+            hasNotOffScreenCanvas,
+            false,
+          );
+        },
```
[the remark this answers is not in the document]
TecHaxter: This is addressed now.

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -638,4 +643,70 @@ class Camera {
...
+
+    int? animate(num timestamp) {
+      // Schedule the next frame
+      animationFrameId = window.requestAnimationFrame(animate.toJS);
+      // Calculate the elapsed time since the last frame
+      final num elapsed = timestamp - then;
+
+      // if we're close to the next frame (by ~8ms), do it.
+      if (fpsInterval - elapsed <= 8) {
```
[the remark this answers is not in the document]
TecHaxter: This is addressed.

packages/camera/camera_web/lib/src/camera.dart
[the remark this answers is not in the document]
TecHaxter: This is addressed.

stuartmorgan-g:
From triage: @mdebbar This looks to be ready for re-review.

stuartmorgan-g:
From triage: @mdebbar, ping on this re-review.

stuartmorgan-g:
From triage: Still waiting on web team review.

packages/camera/camera_web/lib/src/camera_web.dart
```diff
@@ -629,6 +629,21 @@ class CameraPlugin extends CameraPlatform {
...
+    CameraImageStreamOptions? options,
+  }) {
+    try {
+      return getCamera(cameraId).cameraFrameStream(options: options);
+    } on web.DOMException catch (e) {
+      throw PlatformException(code: e.name, message: e.message);
+    } on CameraWebException catch (e) {
+      _addCameraErrorEvent(e);
+      throw PlatformException(code: e.code.toString(), message: e.description);
+    }
```
jiechic, line 618: you show add override function for onStreamedFrameAvailable

> /// Check whether this platform supports image streaming via [onStreamedFrameAvailable].
>  bool supportsImageStreaming() => false;  (change to true)

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -159,6 +160,13 @@ class Camera {
   final StreamController<VideoRecordedEvent> videoRecorderController =
       StreamController<VideoRecordedEvent>.broadcast();
 
+  /// Used to check if allowed to paint canvas off screen
+  @visibleForTesting
+  final bool canUseOffscreenCanvas;
+
+  /// The tolerance for the camera streaming frame time.
+  final int _frameTimeToleranceMs = 8;
```
mdebbar: Can this be a getter that changes dynamically based on the fps value?

E.g.

```suggestion
  final get int _frameTimeToleranceMs => 1000 / cameraStreamFPS ~/ 2;
```

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -359,4 +360,74 @@ class CameraService {
...
+    return jsUtil.hasProperty(window, 'OffscreenCanvas'.toJS);
+  }
+
+  ///Used in [takeFrame] if `OffscreenCanvas` is not supported
+  web.CanvasElement? _canvasElement;
+
+  ///Used in [takeFrame] if `OffscreenCanvas` is supported
+  web.OffscreenCanvas? _offscreenCanvas;
+
+  ///Returns frame at a specific time using video element
```
mdebbar:
```suggestion
  /// Returns frame at a specific time using video element
```

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -359,4 +360,74 @@ class CameraService {
...
+
+  ///Used to check if browser has OffscreenCanvas capability
+  bool hasPropertyOffScreenCanvas() {
+    return jsUtil.hasProperty(window, 'OffscreenCanvas'.toJS);
+  }
+
+  ///Used in [takeFrame] if `OffscreenCanvas` is not supported
+  web.CanvasElement? _canvasElement;
+
+  ///Used in [takeFrame] if `OffscreenCanvas` is supported
```
mdebbar:
```suggestion
  /// Used in [takeFrame] if `OffscreenCanvas` is supported
```

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -359,4 +360,74 @@ class CameraService {
         return DeviceOrientation.portraitUp;
     }
   }
+
+  ///Used to check if browser has OffscreenCanvas capability
+  bool hasPropertyOffScreenCanvas() {
+    return jsUtil.hasProperty(window, 'OffscreenCanvas'.toJS);
+  }
+
+  ///Used in [takeFrame] if `OffscreenCanvas` is not supported
```
mdebbar:
```suggestion
  /// Used in [takeFrame] if `OffscreenCanvas` is not supported
```

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -359,4 +360,74 @@ class CameraService {
         return DeviceOrientation.portraitUp;
     }
   }
+
+  ///Used to check if browser has OffscreenCanvas capability
```
mdebbar:
```suggestion
  /// Used to check if browser has OffscreenCanvas capability
```

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -359,4 +360,74 @@ class CameraService {
...
+    if (width == 0 || height == 0) {
+      throw Exception(
+        'Computed dimensions are zero: width=$width, height=$height',
+      );
+    }
+    late web.ImageData imageData;
+    if (hasPropertyOffScreenCanvas()) {
+      if (_offscreenCanvas == null ||
+          _offscreenCanvas!.width != width ||
+          _offscreenCanvas!.height != height) {
```
mdebbar: Can we resize the existing offscreen canvas instead of creating a new one?

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -359,4 +360,74 @@ class CameraService {
...
+                '2d',
+                <String, Object?>{'willReadFrequently': true}.jsify(),
+              )!
+              as web.OffscreenCanvasRenderingContext2D;
+      context.drawImage(videoElement, 0, 0);
+      imageData = context.getImageData(0, 0, width, height);
+    } else {
+      if (_canvasElement == null ||
+          _canvasElement!.width != width ||
+          _canvasElement!.height != height) {
```
mdebbar: Same here

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -359,4 +360,74 @@ class CameraService {
...
+  ///Returns frame at a specific time using video element
+  CameraImageData takeFrame(web.VideoElement videoElement) {
+    final int width = videoElement.videoWidth;
+    final int height = videoElement.videoHeight;
+    if (width == 0 || height == 0) {
+      throw Exception(
+        'Computed dimensions are zero: width=$width, height=$height',
+      );
+    }
+    late web.ImageData imageData;
```
mdebbar: I think this can be:

```suggestion
    final web.ImageData imageData;
```

mdebbar (commented):
Overall looks good! Sorry for the long delay.

I only have a few suggestions and nits. After that it should be good to go. Thank you for the contribution!

packages/camera/camera_web/lib/src/camera_web.dart
TecHaxter: Good catch. Added the support.

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: Sure. Done👍

packages/camera/camera_web/lib/src/camera_service.dart
TecHaxter: That would be a great optimization. Done👍
I have also optimized the offscreen canvas context call by preserving it from the first call. 
Please do take a final look at that. Many thanks.

packages/camera/camera_web/lib/src/camera_service.dart
TecHaxter: Sure. Done👍

TecHaxter:
### Demo
Dummy example app [source code](https://dartpad.dev/32bca9f9ffd3a7e0835c18c6b211c527)
<video src="https://github.com/user-attachments/assets/813aa90e-a1d5-457f-afd4-c12c22af85c5" width="300"/>

mdebbar:
Some lints have been reported here: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8685251527310718641/+/u/Run_package_tests/analyze/stdout

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -48,8 +48,11 @@ class Camera {
     required this.textureId,
     required CameraService cameraService,
     this.options = const CameraOptions(),
-    this.recorderOptions = const (audioBitrate: null, videoBitrate: null),
-  }) : _cameraService = cameraService;
+    ({int? audioBitrate, int? videoBitrate})? recorderOptions,
+  }) : recorderOptions =
+           recorderOptions ?? (audioBitrate: null, videoBitrate: null),
```
TecHaxter: **Why this change?**
- `mockito` walks the constructor and inspects parameter defaults.
- For a `const` record literal, it goes through `source_gen` constant revival (`reviveInstance`).
- That code path of the current version of the dependency does not handle record constants properly.
- As a result, it hits a null and **throws Null check operator used on a null value**.

TecHaxter:
Ready for review

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -642,4 +652,70 @@ class Camera {
...
   }
+
+  final StreamController<CameraImageData> _cameraFrameStreamController =
+      StreamController<CameraImageData>.broadcast();
+
+  // TODO(TecHaxter): Introduce FPS in CameraImageStreamOptions of
+  //                  package:camera_platform_interface.
+  //                  https://github.com/flutter/flutter/issues/176148
+  /// Used for running the camera frame stream at a specified FPS
+  final int cameraStreamFPS = 60;
```
ditman, line 663: Is this `@visibleForTesting`?

Also I think we should let the user know that this FPS is a maximum cap in the framerate, but not a guarantee. (Web devs should already know this, but maybe this is surprising for a mobile developer compiling for web)

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -642,4 +652,70 @@ class Camera {
...
+    num lastFrameTimestamp = 0;
+
+    int? animate(num timestamp) {
+      // Schedule the next frame
+      animationFrameId = window.requestAnimationFrame(animate.toJS);
+      // Calculate the elapsed time since the last frame
+      final num elapsed = timestamp - lastFrameTimestamp;
+
+      // If we're close to the next frame (~`_frameTimeToleranceMs`), do it.
+      if (fpsInterval - elapsed <= _frameTimeToleranceMs) {
```
ditman, line 695: Oh! So it is NOT a maximum cap on FPS. Does this mean that at 60fps we can trigger a new frame 8ms early?

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -361,4 +362,80 @@ class CameraService {
...
+      }
+      _offscreenCanvasContext ??=
+          _offscreenCanvas!.getContext(
+                '2d',
+                <String, Object?>{'willReadFrequently': true}.jsify(),
+              )!
+              as web.OffscreenCanvasRenderingContext2D;
+
+      _offscreenCanvasContext!.drawImage(videoElement, 0, 0);
+      imageData = _offscreenCanvasContext!.getImageData(0, 0, width, height);
```
ditman, line 406: Can you please split the behavior of taking an offscreen frame vs the fallback into two smaller methods? `_takeOffscreenCanvasFrame` and `_takeFallbackCanvasFrame` (both return `imageData`). That way the `takeFrame` method becomes smaller?

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -361,4 +362,80 @@ class CameraService {
...
+        0,
+        width.toDouble(),
+        height.toDouble(),
+      );
+      imageData = context.getImageData(0, 0, width, height);
+    }
+    final ByteBuffer byteBuffer = imageData.data.toDart.buffer;
+
+    return CameraImageData(
+      format: const CameraImageFormat(ImageFormatGroup.unknown, raw: 0),
```
ditman, line 430: Would it make sense to extend the `ImageFormatGroup` to better describe what the browser is capturing here?

(There's also [`settings` for `getImageData`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData#settings) that may be used here to capture frames in a "less unknown" format, maybe?)

ditman (commented):
This is very cool! I don't work on this nowadays, but here's some extra comments:

* Would [`requestVideoFrameCallback`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/requestVideoFrameCallback) be better than requestAnimationCallback?
* Would it be possible to use [`MediaStreamTrackProcessor`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrackProcessor) if available in `window`? There's an example here: [Video Frame Processing on the Web – WebAssembly, WebGPU, WebGL, WebCodecs, WebNN, and WebTransport](https://webrtchacks.com/video-frame-processing-on-the-web-webassembly-webgpu-webgl-webcodecs-webnn-and-webtransport)

Anyway, very cool. Thanks for reviving this!

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: Making it `@visibleForTesting` would be a good idea. Let me do that.

I have updated the comment to say it out loud more explicitly.

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: Yes, frames can trigger up to ~8ms early.
This means it's not really a maximum cap — it's more of a "try to hit 60fps with an 8ms tolerance window." You could actually see frame rates slightly exceed 60fps during burst scenarios.

packages/camera/camera_web/lib/src/camera_service.dart
TecHaxter: Valid point. I have addressed it.

packages/camera/camera_web/lib/src/camera_service.dart
TecHaxter: Supported [pixelFormat](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData#pixelformat) values for CanvasRenderingContext2D.[getImageData](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData) are:
1. `rgba-unorm8`, which corresponds to Flutter's `ui.PixelFormat.rgba8888`
2.` rgba-float16`

Neither is compatible with any current `ImageFormatGroup` value. For this PR, `ImageFormatGroup.unknown` is the appropriate value until `ImageFormatGroup.rgba8888` exists. `ImageFormatGroup.bgra8888` is also not suitable.

Here's the open issue for this https://github.com/flutter/flutter/issues/151193. I have added this in the TODO comment.

TecHaxter:
Thanks for the valuable comments Ditman and the suggestion about [MediaStreamTrackProcessor](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrackProcessor) :)
I have drafted some changes locally to use that method as primary streaming option, and `requestAnimationCallback` as a fallback. 
Please be patient for next couple of commits from my side very soon. Thanks again for your patience. 🙏

kevmoo:
@TecHaxter – we love thoughtful, polite contributors. Thanks for helping out!

packages/camera/camera_web/lib/src/camera_service.dart
ditman: There seems to be a fix incoming for #151193 (#11632) thanks for creating the issue!

ditman: (We can address this at any time later I think, so don't take my comment as a blocking one. Resolving!)

jiechic:
is any update on this PR？

stuartmorgan-g:
From triage: @mdebbar it looks from discussion above like this is ready for a hopefully final review round.

scognito:
Please release this awesome feature! At least, before GTA 6 :)

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -363,4 +370,274 @@ class CameraService {
...
+      );
+    }
+
+    final web.VideoFrame videoFrame = _wasmCompatible(
+      readResult.value,
+      isVideoFrameType,
+      cameraId: cameraId,
+    );
+    if (videoFrame.visibleRect == null) {
+      throw CameraWebException(
```
mdebbar, line 570: Let's call `videoFrame.close()` before throwing to avoid a memory leak.

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -587,4 +602,193 @@ class Camera {
       ..height = '100%'
       ..objectFit = 'cover';
   }
+
+  StreamController<CameraImageData>? _cameraFrameStreamController;
```
mdebbar, line 607: Should this controller be closed in the `dispose()` method?

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -363,4 +370,274 @@ class CameraService {
...
+  }) {
+    _canvasElement ??= web.CanvasElement()
+      ..height = height
+      ..width = width;
+    if (_canvasElement!.width != width || _canvasElement!.height != height) {
+      _canvasElement!
+        ..width = width
+        ..height = height;
+    }
+    final web.CanvasRenderingContext2D context = _canvasElement!.context2D;
```
mdebbar: Is there a reason we aren't using `'willReadFrequently': true` for this context?

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -587,4 +602,193 @@ class Camera {
...
+  /// Whether the camera frame stream is initialized.
+  bool _isCameraFrameStreamInitialized() =>
+      _cameraFrameStreamController != null && !_cameraFrameStreamController!.isClosed;
+
+  /// Whether the camera is currently streaming frames.
+  bool _isCameraFrameStreaming() =>
+      _isCameraFrameStreamInitialized() && _cameraFrameStreamController!.hasListener;
+
+  /// Used for safely emitting camera frame stream event
+  void _safelyEmitCamereFrameStreamEvent(CameraImageData data) {
```
mdebbar: typo: 
```suggestion
  void _safelyEmitCameraFrameStreamEvent(CameraImageData data) {
```

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -587,4 +602,193 @@ class Camera {
...
+    if (_isCameraFrameStreamInitialized()) {
+      _cameraFrameStreamController!.add(data);
+    }
+  }
+
+  /// Used for safely closing camera frame stream event
+  void _safelyCloseCamereFrameStream() {
```
mdebbar: typo:

```suggestion
  void _safelyCloseCameraFrameStream() {
```

packages/camera/camera_web/lib/src/camera.dart
```diff
@@ -587,4 +602,193 @@ class Camera {
...
+    if (_isCameraFrameStreamInitialized()) {
+      _cameraFrameStreamController!.close();
+    }
+  }
+
+  /// Used for safely emitting camera frame stream error
+  void _safelyEmitCamereFrameStreamError(Object error) {
```
mdebbar: typo:

```suggestion
  void _safelyEmitCameraFrameStreamError(Object error) {
```

packages/camera/camera_web/lib/src/camera_service.dart
```diff
@@ -363,4 +370,274 @@ class CameraService {
...
+    return CameraImageData(width: width, height: height, format: format, planes: [plane]);
+  }
+
+  /// Performs wasm compatibility checks on [JSAny] to return [T]
+  T _wasmCompatible<T extends JSAny>(
+    JSAny? raw,
+    bool Function(JSAny) checkType, {
+    int cameraId = 0,
+  }) {
+    // Use the callback to perform the compile-time bound check
```
mdebbar: Did you mean runtime instead of "compile-time"?

mdebbar (changes requested):

packages/camera/camera_web/lib/src/camera_service.dart
TecHaxter: Good point. Addressed it 👍

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: Good point. I overlooked this. Addressed it 👍

packages/camera/camera_web/lib/src/camera_service.dart
TecHaxter: Good point. I overlooked this. Addressed it 👍

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: Addressed it 👍

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: Addressed it 👍

packages/camera/camera_web/lib/src/camera.dart
TecHaxter: Addressed it 👍

packages/camera/camera_web/lib/src/camera_service.dart
TecHaxter: My mistake. Addressed it 👍

PR #7956, opened 2024-10-29
[various] Remove multiDexEnabled

stuartmorgan:
We had manually added `multiDexEnabled` to various plugin example apps over time to fix build failures, but it's not needed for plugins with a `minSdk` of 20+ ([reference](https://developer.android.com/build/multidex)), and Flutter (and thus all of our plugins) requires 21+, so the setting is just cruft.


reidbaker (approved):

gmackall (approved):

reidbaker:
Old man logic here. I remember when multidex started becoming an issue and I genuinely didn't think I would be doing android development still when we got to a point where it wasn't needed anymore. Woo

PR #7969, opened 2024-10-31
[camera_android] getAvailableCameras to handle non-integer camera IDs

QuiverYang:
### Background
Some Android devices, especially those with external or specialized cameras, may use non-integer camera IDs. The existing implementation attempts to parse cameraName as an integer, which causes issues for these devices.

### Changes
- Removed integer parsing of cameraName in the getAvailableCameras method.
- Directly used cameraName as a string for constructing camera descriptions.
- Enhanced compatibility with a broader range of Android devices, ensuring that both numeric and non-numeric camera IDs are supported.

### Testing
- Verified the changes on devices with both integer and non-integer camera IDs to ensure compatibility.

This PR aims to enhance the flexibility and compatibility of the camera plugin with various device configurations.

### DIFF
--- a/packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraUtils.java
+++ b/packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/CameraUtils.java
@@ -70,15 +70,6 @@ public final class CameraUtils {
     String[] cameraNames = cameraManager.getCameraIdList();
     List<Messages.PlatformCameraDescription> cameras = new ArrayList<>();
     for (String cameraName : cameraNames) {
-      int cameraId;
-      try {
-        cameraId = Integer.parseInt(cameraName, 10);
-      } catch (NumberFormatException e) {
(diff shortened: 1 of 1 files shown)


QuiverYang:
I just signed CLA!

camsim99 (commented):
The change sounds good to me! I think you'll need to update this test though:

https://github.com/flutter/packages/blob/da4c3eedb5e70d47452088d8cb30c02c058052a7/packages/camera/camera_android/android/src/test/java/io/flutter/plugins/camera/CameraUtilsTest.java#L31

stuartmorgan:
@QuiverYang Please make sure the email address associated with the commit is the same one you used when signing the CLA. If you've used the wrong address, you can amend the commit locally and then force-push.

stuartmorgan:
Unfortunately we’re unable to accept PRs without a signed CLA. If you sign the CLA in the future, please comment to let us know, and we can re-open this PR.

PR #7991, opened 2024-10-31
[vector_garphics] fix execution on the web with WebAssembly

kevmoo:
The `dart.library.html` variable is not available with wasm
Since the web version is (currently) a no-op, flip the condition to
check for dart:io instead

### DIFF
--- a/packages/vector_graphics/pubspec.yaml
+++ b/packages/vector_graphics/pubspec.yaml
@@ -4,7 +4,7 @@ repository: https://github.com/flutter/packages/tree/main/packages/vector_graphi
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+vector_graphics%22
 # See https://github.com/flutter/flutter/issues/157626 before publishing a new
 # version.
-version: 1.1.12
+version: 1.1.13
 
 environment:
   sdk: ^3.4.0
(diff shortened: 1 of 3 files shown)


jonahwilliams (approved):
LGTM

packages/vector_graphics/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ repository: https://github.com/flutter/packages/tree/main/packages/vector_graphi
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+vector_graphics%22
 # See https://github.com/flutter/flutter/issues/157626 before publishing a new
 # version.
```
stuartmorgan, line 6: This PR ignored the TODO here, so `vector_graphics` is now mis-aligned with the `vector_graphics_*` versions. @jonahwilliams We need a versioning plan here ASAP.

kevmoo: 💩 – sorry, folks

jonahwilliams: I thought these weren't being auto published yet?

kevmoo: I got confirmed so I flipped the bit and assumed the robots would "do the right thing" 🤷

stuartmorgan: > I thought these weren't being auto published yet?

I need to publish an initial version so that the pub.dev metadata (issue link, source link, etc.) would be correct. That literally landed yesterday at the end of my workday; I didn't expect someone to create a PR, ignore the comment on the version saying not to publish a new version yet, get it approved, and land it, all before I started my next day and could do a follow-up PR to mark them as unpublished.

PR #7995, opened 2024-11-01
[various] Minor cleanup in recently imported packages

stuartmorgan:
Fixes a few minor things from the recent import of dnfield's vector-graphics-related packages:
- Fixes a bad copy/paste in the `path_parsing` CHANGELOG so future versions of the CHANGELOG will be less confusing.
- Removes a coverage link from the `flutter_svg` README that is no longer correct.
- Disabled publishing of `vector_graphics*` until a new versioning plan is made.

Because these are extremely minor, I'm overriding our usual versioning policy; it's fine for the CHANGELOG and README fixes to go out the next time we happen to publish rather than immediately.


jonahwilliams (approved):
LGTM

PR #7997, opened 2024-11-01
[go_router] Update example app to Kotlin 1.9.0

stuartmorgan:
The tree is currently closed on an out-of-band failure in the build of this example app:

e: .../transformed/jetified-kotlin-stdlib-1.9.24.jar!/META-INF/kotlin-stdlib-jdk7.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
e: /b/s/w/ir/cache/gradle/caches/transforms-4/6cce2b947fc9e43cca6b44ff30180b61/transformed/jetified-kotlin-stdlib-1.9.24.jar!/META-INF/kotlin-stdlib-jdk8.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
e: .../transformed/jetified-kotlin-stdlib-1.9.24.jar!/META-INF/kotlin-stdlib.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
e: .../transformed/jetified-annotation-jvm-1.9.1.jar!/META-INF/annotation.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.

It appears that this is caused by
https://github.com/flutter/packages/pull/7986 as this example app does depend on `url_launcher`, and that PR therefore change a transitive dependency of this example app, and as Reid discovered that version bump [changed the Kotlin
version](https://android.googlesource.com/platform/frameworks/support/+log/2e6556166965445a8a129114765a3903c063735c..87b88ad088cc9b18d4ab75611fc8b74e8b01c24a/annotation)

There was a Flutter Fix that contained the correct instructions for resolving this, which is what this PR does.

PR #7088, opened 2024-07-09
[many] Remove dependency on `kotlin-bom`

gmackall:
~This dependency seems to no longer be necessary. I expected this would be because the the androidx upgrade that [landed recently](https://github.com/flutter/engine/pull/53592) fixed the problem ([see theory](https://github.com/flutter/packages/pull/7036#issuecomment-2218659328)), but it seems that the packages even build on stable successfully. Perhaps there have been updates to the underlying `androidx` libraries themselves that fix the conflict, and we updated the versions in plugins far enough? I'm unsure.~

Updated for 3.24 release.

In a sense, fixes https://github.com/flutter/flutter/issues/125062

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ ] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[...]

PR #7998, opened 2024-11-01
[various] Update example apps to Kotlin 1.9.0

stuartmorgan:
Follow-up to https://github.com/flutter/packages/pull/7997 to update all example apps to 1.9.0 to protect us from similar OOB failures as other transitive dependencies in the ecosystem update to newer versions of Kotlin.


gspencergoog (approved):

stuartmorgan:
@gmackall / @reidbaker My reasoning here is that now that we have a Flutter Fix telling people what to do in cases like https://github.com/flutter/packages/pull/7997, and that we deliberately removed the workaround that was intended to prevent it in https://github.com/flutter/packages/pull/7088, we consider that kind of failure something clients should fix in their app (per the Flutter Fix) rather than a bug in our plugins. If that's the case, it's better to preemptively update our example apps to avoid hitting it as an OOB failure.

If those assumptions aren't correct (and, e.g., we want to revert that `url_launcher_android` dependabot change because it pulled in Kotlin 1.9.0), let me know and we can instead revisit our testing approach instead.

tarrinneal (approved):

domesticmouse (approved):

gmackall (approved):
> @gmackall / @reidbaker My reasoning here is that now that we have a Flutter Fix telling people what to do in cases like https://github.com/flutter/packages/pull/7997, and that we deliberately removed the workaround that was intended to prevent it in https://github.com/flutter/packages/pull/7088, we consider that kind of failure something clients should fix in their app (per the Flutter Fix) rather than a bug in our plugins. If that's the case, it's better to preemptively update our example apps to avoid hitting it as an OOB failure.

I think it makes sense to have users update now that we have an error handler in place, lgtm.

PR #8012, opened 2024-11-03
[google_maps_flutter] Add a new zIndexInt param to marker and deprecate zIndex

Hari-07:
`zIndex` on marker is currently a double. This introduces inconsistency between its behaviours on android and ios since ios truncates it to an int. This PR aims to smooth out that difference by accepting an int param instead through the marker API in dart

Fixes https://github.com/flutter/flutter/issues/155897

## Pre-launch Checklist

- [x] I signed the [CLA].


Hari-07:
The checking for default value in native code feels a little weird, but since this is a parameter we've a default value for in the dart layer, not sure what other option is there.

The only other one I could think of was checking for 0 in dart itself and then passing down null, for `zIndex` if that was the case which would mean making it nullable . That might need a bunch of changes and wasn't sure if it's worth it for a deprecation

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -599,7 +599,14 @@ static void interpretMarkerOptions(
     sink.setPosition(toLatLng(marker.getPosition().toList()));
     sink.setRotation(marker.getRotation().floatValue());
     sink.setVisible(marker.getVisible());
-    sink.setZIndex(marker.getZIndex().floatValue());
+
+    int zIndexInt = marker.getZIndexInt().intValue();
+    float zIndex = marker.getZIndex().floatValue();
+    if (zIndexInt != 0) {
```
reidbaker, line 605: Can you add a comment here that the reason we check for zero is that zIndexInt is supposed to take priority over zindex? |

Please also add a test in https://github.com/flutter/packages/blob/2c8f22670812996a00e45e6304f79f1fd9b02fa1/packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ConvertTest.java to verify the behavior.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/marker.dart
```diff
@@ -150,7 +150,12 @@ class Marker implements MapsObject<Marker> {
     this.position = const LatLng(0.0, 0.0),
     this.rotation = 0.0,
     this.visible = true,
+    @Deprecated(
+      'Use zIndexInt instead. '
+      'Use of zIndex was deprecated as it gets truncated to int on iOS.',
```
reidbaker, line 155:
```suggestion
      'On iOS zIndex is truncated to an int, which can lead to incorrect/unstable ordering.',
```

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/marker.dart
```diff
@@ -219,8 +224,19 @@ class Marker implements MapsObject<Marker> {
   ///
   /// Overlays are drawn in order of z-index, so that lower values means drawn
   /// earlier, and thus appearing to be closer to the surface of the Earth.
+  @Deprecated(
+    'Use zIndexInt instead. '
+    'Use of zIndex was deprecated as it gets truncated to int on iOS.',
```
reidbaker, line 229:
```suggestion
      'On iOS zIndex is truncated to an int, which can lead to incorrect/unstable ordering.',
```

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/marker.dart
```diff
@@ -246,7 +262,12 @@ class Marker implements MapsObject<Marker> {
     LatLng? positionParam,
     double? rotationParam,
     bool? visibleParam,
+    @Deprecated(
+      'Use zIndexIntParam instead. '
+      'Use of zIndex was deprecated as it gets truncated to int on iOS.',
```
reidbaker, line 267:
```suggestion
      'On iOS zIndex is truncated to an int, which can lead to incorrect/unstable ordering.',
```

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_web/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## NEXT
+## 0.5.10+1
 
+* Deprecate zIndex parameter in Marker and replace with zIndexInt.
```
ditman, line 3: I'd expect to see a change here for the web?

* https://github.com/flutter/packages/blob/79f8b0b6655132ac8d11a50243afe85867a09405/packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart#L269

ditman (commented):
This is probably minor, but the web version is not handling the new `zIndexInt` value. We should address that, or create a follow-up ticket so somebody else can implement the desired behavior there.

packages/google_maps_flutter/google_maps_flutter_web/CHANGELOG.md
Hari-07: Thank you for the reminder, didn't think about web. I can just add that in right away

packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/marker_test.dart
```diff
@@ -205,5 +206,21 @@ void main() {
         }, throwsAssertionError);
       });
     });
+
+    testWidgets('interpretCorrectZIndexInConversion',
```
Hari-07, line 210: I didn't find any conversion tests in the web package, so Im not 100% sure if this is in the right place. Maybe it should be in a dedicated convert_test file

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -444,9 +444,10 @@ Future<gmaps.Icon?> _gmIconFromBitmapDescriptor(
   return icon;
 }
 
-// Computes the options for a new [gmaps.Marker] from an incoming set of options
-// [marker], and the existing marker registered with the map: [currentMarker].
-Future<gmaps.MarkerOptions> _markerOptionsFromMarker(
+/// Computes the options for a new [gmaps.Marker] from an incoming set of options
+/// [marker], and the existing marker registered with the map: [currentMarker].
+@visibleForTesting
```
Hari-07, line 449: This feels a little weird, and I think maybe all the methods here could be static methods of a class

ditman: Yeah, this file started up being small, then grew to epic proportions. It should eventually disappear, but there's still some "stringly typed" API coming from the core plugin (everything that is backed by a JSON object, for example).

Thanks for adding a convert_test!

Hari-07: >this file started up being small, then grew to epic proportions. It should eventually disappear
Haha, classic!

ditman: In order to not have to modify this method, I'd suggest you add the test to `markers_test.dart`, which is calling all the conversion methods when doing `controller.addMarkers`.

See this:

* https://github.com/flutter/packages/blob/c77ab99ab8385690f31e72b6a66f01f9c75dd60e/packages/google_maps_flutter/google_maps_flutter_web/example/integration_test/markers_test.dart#L57-L95

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/marker.dart
```diff
@@ -150,7 +150,12 @@ class Marker implements MapsObject<Marker> {
     this.position = const LatLng(0.0, 0.0),
     this.rotation = 0.0,
     this.visible = true,
+    @Deprecated(
+      'Use zIndexInt instead. '
+      'On iOS zIndex is truncated to an int, which can lead to incorrect/unstable ordering.',
+    )
     this.zIndex = 0.0,
+    this.zIndexInt = 0,
```
stuartmorgan, line 158: zIndexInt should not be a new field, as this creates the possibility of inconsistent values in the same marker. Instead, it can be:

```dart
  const Marker({
    ...
    double zIndex = 0.0,
    int zIndexInt = 0,
    ...
  }): assert(zIndex == 0.0 || zIndexInt == 0, 'Only one of zIndex and zIndexInt can be provided'), zIndex = zIndexInt == 0 ? zIndex : zIndexInt.toDouble();

  ...
  /// ...
  // TODO(stuartmorgan): Make this an int when removing the deprecated double zIndex parameter.
  @Deprecated(...)
  final double zIndex;

  /// ...
  int get zIndexInt => zIndex.round();
```

Then there is only a single value. iOS can read `zIndexInt` since it will be truncated anyway, and Android can read `zIndex` with a comment explaining that it's using a deprecated parameter for backward compatibility.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -198,6 +199,7 @@ class PlatformMarker {
   final double rotation;
   final bool visible;
   final double zIndex;
+  final int zIndexInt;
```
stuartmorgan, line 202: This should be removed; we don't need to send two values.

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -1503,6 +1503,7 @@ Marker _copyMarkerWithClusterManagerId(
     rotation: marker.rotation,
     visible: marker.visible,
     zIndex: marker.zIndex,
+    zIndexInt: marker.zIndexInt,
```
stuartmorgan, line 1506: This won't be needed after the suggested change in the platform interface. The `zIndex` will just need a comment saying that the deprecated parameter is being used to avoid losing precision.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -599,7 +599,15 @@ static void interpretMarkerOptions(
     sink.setPosition(toLatLng(marker.getPosition().toList()));
     sink.setRotation(marker.getRotation().floatValue());
     sink.setVisible(marker.getVisible());
-    sink.setZIndex(marker.getZIndex().floatValue());
+
+    // We compare with zero because zIndexInt if non zero should get higher priority than zIndex
+    int zIndexInt = marker.getZIndexInt().intValue();
+    float zIndex = marker.getZIndex().floatValue();
```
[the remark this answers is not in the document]
stuartmorgan: In fact, I believe once the other changes have been made, you can revert all of the Java changes in this PR.

stuartmorgan (changes requested):

stuartmorgan:
From triage: @Hari-07 Are you still planning on updating this based on the latest review?

stuartmorgan:
No, flutter/packages already uses `dart format`, so is unaffected.

Hari-07:
@stuartmorgan Sorry for further delays, got busy with the holiday season, will try and get to this at the earliest

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/marker.dart
Hari-07: >  zIndex = zIndexInt == 0 ? zIndex : zIndexInt.toDouble()

This would need making Marker non -const

stuartmorgan: Ah, so it would.

We'll need to go the slightly clunkier route of using a `num` then:

```dart
  const Marker({
    ...
    double zIndex = 0.0,
    int zIndexInt = 0,
    ...
  }): assert(zIndex == 0.0 || zIndexInt == 0, 'Only one of zIndex and zIndexInt can be provided'), _zIndexNum = zIndexInt == 0 ? zIndex : zIndexInt;

  final num _zIndexNum;

  double get zIndex => _zIndexNum.toDouble();

  int get zIndexInt => _zIndex.round();
```

stuartmorgan:
From triage: @Hari-07 was this ready for re-review? If so, it looks like it'll need a sync with `main` and then we can take a look again.

Piinks:
Greetings from stale PR triage, @Hari-07 do you have plans to return to this one?

Hari-07:
Hey @Piinks yeah sure, I had actually made the change he had suggested but forgot to push it up for some reason. Also looks like a sync with main is needed. Doing that now.

Missed the email about this reply

Hari-07:
@stuartmorgan-g this seems to finally be in a state that's good to go. It's been rebased to main and all CI is passing.

packages/google_maps_flutter/google_maps_flutter_platform_interface/test/types/marker_test.dart
stuartmorgan-g, line 1: This file needs to test all the codepaths through the new params: passing both asserting, passing each as non-zero.

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
stuartmorgan-g: It looks like this still needs to be done?

stuartmorgan-g (commented):
Minor comments, but feel free to split out the platform interface PR and address the tests there; there's nothing structural left that prevents proceeding with splitting the PR.

Hari-07:
I removed all overrides for the platform interface package from all other packages, the override that's remaining now is from the main package to android and ios. Does this mean android/ios needs to be merged in as a separate PR and then the main package needs to target the new published version?

Or is it fine to bump the main package's version requirement to a version that isn't published yet

stuartmorgan-g:
> Does this mean android/ios needs to be merged in as a separate PR and then the main package needs to target the new published version?

Yes.

> Or is it fine to bump the main package's version requirement to a version that isn't published yet

That would fail essentially everything in CI, since the `pub` resolution would fail, and thus be unlandable.

Hari-07:
Ok cool, have made a PR with everything except the main package. Once that's in can update the main package here to target the newly released packages and then this should be good to merge finally

Hari-07:
@stuartmorgan-g it might be a good idea to have some kind of automated action that uses `git checkout -- package/` to automatically create these partial PRs, that'd then allow you the reviewer to not have to check to make sure it has exactly the changes you expect, not more nor less, and easier for contributors too.

Hari-07:
Test failure
```
Running command: "flutter pub get" in /b/s/w/ir/x/w/packages/packages/google_maps_flutter/google_maps_flutter
Resolving dependencies...
Because google_maps_flutter depends on google_maps_flutter_ios ^2.15.4 which doesn't match any versions, version solving failed.


You can try the following suggestion to make the pubspec resolve:
* Consider downgrading your constraint on google_maps_flutter_ios: flutter pub add google_maps_flutter_ios:^2.15.3
Failed to update packages.
```

`2.15.4` is the new version that was just merged in a few minutes back, so maybe it just takes time before that's published and then the tests need to be rerun?

stuartmorgan-g:
Autopublish happens after all post-submit tests complete successfully, which isn't quite finished.

Hari-07:
@stuartmorgan-g Could you trigger a rerun of the tests here?

stuartmorgan-g (approved):
LGTM!

PR #8014, opened 2024-11-04
[flutter_svg] wasm compatibility

nank1ro:

## Pre-launch Checklist

- [x ] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I signed the [CLA].

### DIFF
--- a/packages/vector_graphics/CHANGELOG.md
+++ b/packages/vector_graphics/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.14
+
+* Make the package WASM compatible
+
 ## 1.1.13
 
 * Fix execution on the web with WebAssembly.
--- a/packages/vector_graphics/lib/src/debug.dart
+++ b/packages/vector_graphics/lib/src/debug.dart
@@ -2,4 +2,4 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-export '_debug_web.dart' if (dart.library.io) '_debug_io.dart';
+export '_debug_io.dart' if (dart.library.js_interop) '_debug_web.dart';
--- a/packages/vector_graphics/pubspec.yaml
+++ b/packages/vector_graphics/pubspec.yaml
@@ -5,7 +5,7 @@ issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+
 # See https://github.com/flutter/flutter/issues/157626 before publishing a new
 # version.
 publish_to: none
-version: 1.1.13
+version: 1.1.14
 
 environment:
   sdk: ^3.4.0
@@ -22,7 +22,7 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   # See https://github.com/flutter/flutter/issues/157626
-  vector_graphics_compiler: ">=1.1.11+1 <= 1.1.12"
+  vector_graphics_compiler: ">=1.1.11+1 <= 1.1.13"
 
 platforms:
   android:
--- a/packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
+++ b/packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
@@ -18,9 +18,9 @@ import 'src/svg/theme.dart';
 import 'src/vector_instructions.dart';
 
 export 'src/_initialize_path_ops_io.dart'
-    if (dart.library.html) 'src/_initialize_path_ops_web.dart';
+    if (dart.library.js_interop) 'src/_initialize_path_ops_web.dart';
 export 'src/_initialize_tessellator_io.dart'
-    if (dart.library.html) 'src/_initialize_tessellator_web.dart';
+    if (dart.library.js_interop) 'src/_initialize_tessellator_web.dart';
 export 'src/geometry/basic_types.dart';
 export 'src/geometry/matrix.dart';
 export 'src/geometry/path.dart';
--- a/packages/vector_graphics_compiler/pubspec.yaml
+++ b/packages/vector_graphics_compiler/pubspec.yaml
@@ -5,7 +5,7 @@ issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+
 # See https://github.com/flutter/flutter/issues/157626 before publishing a new
 # version.
 publish_to: none
-version: 1.1.12
+version: 1.1.13
 
 executables:
   vector_graphics_compiler:
@@ -36,7 +36,7 @@ dev_dependencies:
     sdk: flutter
   test: ^1.20.1
   # See https://github.com/flutter/flutter/issues/157626
-  vector_graphics: ">=1.1.11+1 <= 1.1.12"
+  vector_graphics: ">=1.1.11+1 <= 1.1.14"
   vector_math: ^2.1.2
 
 platforms:
--- a/third_party/packages/flutter_svg/lib/src/utilities/file.dart
+++ b/third_party/packages/flutter_svg/lib/src/utilities/file.dart
@@ -1,1 +1,1 @@
-export '_file_io.dart' if (dart.library.html) '_file_none.dart';
+export '_file_io.dart' if (dart.library.js_interop) '_file_none.dart';
--- a/third_party/packages/flutter_svg/pubspec.yaml
+++ b/third_party/packages/flutter_svg/pubspec.yaml
@@ -2,7 +2,7 @@ name: flutter_svg
 description: An SVG rendering and widget library for Flutter, which allows painting and displaying Scalable Vector Graphics 1.1 files.
 repository: https://github.com/flutter/packages/tree/main/third_party/packages/flutter_svg
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+flutter_svg%22
-version: 2.0.11
+version: 2.0.12
 
 environment:
   sdk: ^3.4.0
@@ -13,9 +13,9 @@ dependencies:
     sdk: flutter
   http: ^1.0.0
   # See https://github.com/flutter/flutter/issues/157626
-  vector_graphics: ">=1.1.11+1 <= 1.1.12"
+  vector_graphics: ">=1.1.11+1 <= 1.1.14"
   vector_graphics_codec: ">=1.1.11+1 <= 1.1.12"
-  vector_graphics_compiler: ">=1.1.11+1 <= 1.1.12"
+  vector_graphics_compiler: ">=1.1.11+1 <= 1.1.13"
 
 dev_dependencies:
   flutter_test:
--- a/packages/vector_graphics_compiler/CHANGELOG.md
+++ b/packages/vector_graphics_compiler/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.13
+
+* Make the package WASM compatible
+
 ## 1.1.12
 
 * Transfers the package source from https://github.com/dnfield/vector_graphics
--- a/third_party/packages/flutter_svg/CHANGELOG.md
+++ b/third_party/packages/flutter_svg/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.12
+
+* Make the package WASM compatible
+
 ## 2.0.11
 
 * Transfers the package source from https://github.com/dnfield/flutter_svg


packages/vector_graphics_compiler/pubspec.yaml
```diff
@@ -36,7 +36,7 @@ dev_dependencies:
     sdk: flutter
   test: ^1.20.1
   # See https://github.com/flutter/flutter/issues/157626
-  vector_graphics: ">=1.1.11+1 <= 1.1.12"
+  vector_graphics: ">=1.1.11+1 <= 1.1.14"
```
stuartmorgan, line 39: 1.1.14 doesn't exist yet, so we don't know if it'll be compatible under the versioning system `vector_graphics` has used previously.

packages/vector_graphics/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.1.14
+
+* Make the package WASM compatible
```
stuartmorgan, line 3: Please see the CHANGELOG style guide linked from the PR checklist; this needs to use the repo-standard verb form and punctuation.

packages/vector_graphics_compiler/pubspec.yaml
stuartmorgan, line 39: Isn't 11.1.13 the minimum required for wasm compatibility?

packages/vector_graphics/pubspec.yaml
```diff
@@ -22,7 +22,7 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   # See https://github.com/flutter/flutter/issues/157626
-  vector_graphics_compiler: ">=1.1.11+1 <= 1.1.12"
+  vector_graphics_compiler: ">=1.1.11+1 <= 1.1.13"
```
stuartmorgan, line 25: 1.1.13 doesn't exist; what's the purpose of this change?

third_party/packages/flutter_svg/pubspec.yaml
```diff
@@ -13,9 +13,9 @@ dependencies:
     sdk: flutter
   http: ^1.0.0
   # See https://github.com/flutter/flutter/issues/157626
-  vector_graphics: ">=1.1.11+1 <= 1.1.12"
+  vector_graphics: ">=1.1.11+1 <= 1.1.14"
   vector_graphics_codec: ">=1.1.11+1 <= 1.1.12"
-  vector_graphics_compiler: ">=1.1.11+1 <= 1.1.12"
+  vector_graphics_compiler: ">=1.1.11+1 <= 1.1.13"
```
stuartmorgan, line 18: Same question here as in the other package.

third_party/packages/flutter_svg/pubspec.yaml
```diff
@@ -13,9 +13,9 @@ dependencies:
     sdk: flutter
   http: ^1.0.0
   # See https://github.com/flutter/flutter/issues/157626
-  vector_graphics: ">=1.1.11+1 <= 1.1.12"
+  vector_graphics: ">=1.1.11+1 <= 1.1.14"
```
stuartmorgan, line 16: Same notes here as in the other package.

stuartmorgan (changes requested):
Thanks for the contribution!

stuartmorgan:
test-exempt: configuration change, rolling dependencies

packages/vector_graphics_compiler/pubspec.yaml
nank1ro: Not on pub.dev, but I just bumped it here https://github.com/flutter/packages/pull/8014/files#diff-ae3c1055bc547ec174edea702c972431d2d3d46fb9bd78a31378cacf0fd051eaR8

If I don't bump the constraints here, how is it supposed to work?

nank1ro: vector_graphics was already at 1.1.13, I bumped it to 1.1.14 but the v1.1.13 was not used here, no idea why

packages/vector_graphics/CHANGELOG.md
nank1ro: updated

packages/vector_graphics/pubspec.yaml
nank1ro: the same as https://github.com/flutter/packages/pull/8014#discussion_r1827783576

third_party/packages/flutter_svg/pubspec.yaml
nank1ro: the same as https://github.com/flutter/packages/pull/8014#discussion_r1827783576

third_party/packages/flutter_svg/pubspec.yaml
nank1ro: the same as https://github.com/flutter/packages/pull/8014#discussion_r1827783576

packages/vector_graphics/lib/src/debug.dart
```diff
@@ -2,4 +2,4 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-export '_debug_web.dart' if (dart.library.io) '_debug_io.dart';
```
stuartmorgan, line 5: This code was changed for wasm support in https://github.com/flutter/packages/pull/7991; why is it being re-changed here?

packages/vector_graphics_compiler/lib/vector_graphics_compiler.dart
```diff
@@ -18,9 +18,9 @@ import 'src/svg/theme.dart';
 import 'src/vector_instructions.dart';
 
 export 'src/_initialize_path_ops_io.dart'
-    if (dart.library.html) 'src/_initialize_path_ops_web.dart';
+    if (dart.library.js_interop) 'src/_initialize_path_ops_web.dart';
```
stuartmorgan, line 21: We should ideally be consistent about how we switch on web compilation. @kevmoo / @ditman is there a preference between this and `io` (as was done in https://github.com/flutter/packages/pull/7991)?

packages/vector_graphics_compiler/pubspec.yaml
stuartmorgan: You bumped the maximum, not the minimum. The CHANGELOG entry says this PR makes the package wasm compatible, which would generally involve also setting the minimum versions of any dependencies to wasm-compatible versions.

packages/vector_graphics/lib/src/debug.dart
nank1ro: Because I didn't now this one works too, I always used the inverse with `dart.library.js_interop` to be consistent

packages/vector_graphics_compiler/pubspec.yaml
nank1ro: You're right. Didn't notice someone already made vector_graphics wasm compatible. I just fixed the dev_dependency constraint to use the minimum the WASM version. Do you prefer another PR for a single line change (just dev_dependency)?

nank1ro: You're right, updated it right now

stuartmorgan: As long as the changes aren't interdependent, they can be in a single PR. If there are still any cases where you need to both change a package, and change the *minimum* dependency on that package in another package, that must be multiple PRs because it won't compile in CI.

If you're only changing a dev dependency in `v_g_c`, it should be fine to include here. But that would also mean you can't have (and shouldn't need) [this change](https://github.com/flutter/packages/pull/8014/files#diff-ae3c1055bc547ec174edea702c972431d2d3d46fb9bd78a31378cacf0fd051eaR25) in the PR.

nank1ro:
@stuartmorgan I opened [another PR](https://github.com/flutter/packages/pull/8021) for the package `vector_graphics_compiler`.
The other PR must be merged first and the package published on pub dev so this PR can pass all the tests.
I made all the changes requested, let me know if there is something else I can do.

stuartmorgan:
> The other PR must be merged first and the package published on pub dev so this PR can pass all the tests.

Why? Per [this comment](https://github.com/flutter/packages/pull/8014#discussion_r1828037241) I'm not clear why you are updating the minimum version in the dependency on `vector_graphics_compiler` when the only change you are making in `vector_graphics_compiler` is a dev_dependency change. By definition, a dev_dependency change doesn't impact transitive dependencies.

nank1ro:
> > The other PR must be merged first and the package published on pub dev so this PR can pass all the tests.
> 
> Why? Per [this comment](https://github.com/flutter/packages/pull/8014#discussion_r1828037241) I'm not clear why you are updating the minimum version in the dependencies on `vector_graphics_compiler` when the only change you are making in `vector_graphics_compiler` is a dev_dependency change. By definition, a dev_dependency change doesn't impact transitive dependencies.

Because the `vector_graphics_compiler` is in the dependencies of `flutter_svg`, [see here](https://github.com/flutter/packages/pull/8014/files#diff-b55c4ab4f1455da811c8cd990e7ed882c99fae60fea5561ff6f131040e76d4e7R18).
To make `flutter_svg` wasm compatibile, all dependencies must be wasm compatible

stuartmorgan:
Oh, I see now. I thought the `vector_graphics_compiler` changes were dev-only; I missed that the source changes are in non-test code.

nank1ro:
@stuartmorgan I fixed the conflics and merged main. If the tests pass I think it's ready to be merged

third_party/packages/flutter_svg/pubspec.yaml
```diff
@@ -12,9 +12,9 @@ dependencies:
   flutter:
     sdk: flutter
   http: ^1.0.0
-  vector_graphics: ^1.1.11+1
+  vector_graphics: ^1.1.13
   vector_graphics_codec: ^1.1.11+1
-  vector_graphics_compiler: ^1.1.11+1
+  vector_graphics_compiler: ^1.1.13
```
stuartmorgan, line 17: Based on the order things landed, this should be 1.1.14 (`vector_graphics` and `vector_graphics_compiler` ended up with opposite order of WASM support vs relaxed dependencies, it looks like.)

nank1ro: you're right, fixed

stuartmorgan (approved):
LGTM. Thanks for getting this submitted to the new repository so quickly, and for bearing with the churn while the dust was still settling on the initial import!

stuartmorgan:
@jonahwilliams / @domesticmouse / @ditman Could one of you do the secondary approval?

zanderso (approved):

third_party/packages/flutter_svg/lib/src/utilities/file.dart
```diff
@@ -1 +1 @@
-export '_file_io.dart' if (dart.library.html) '_file_none.dart';
+export '_file_io.dart' if (dart.library.js_interop) '_file_none.dart';
```
kevmoo, line 1: Sadly, this might still rank this package as NOT wasm-compatible in pub due to issues with configuration specific imports.

Better to do
```suggestion
export '_file_none.dart' if (dart.library.io) '_file_io.dart';
```

nank1ro: No I can confirm this works because I did it with the fork I published, flutter_svg_plus

kevmoo: I have no doubt that it works at runtime. I'm concerned that the wasm classification on `pub.dev` will still be broken without this change.

kevmoo: See https://github.com/dart-lang/pana/issues/1324

kevmoo: See https://pub.dev/api/packages/flutter_svg/metrics

Search for package:flutter_svg/src/utilities/file.dart

nank1ro: If you think the score for wasm on pub.dev is wrong, I assure you it is not.
See here https://pub.dev/api/packages/flutter_svg_plus/metrics and here https://pub.dev/api/packages/vector_graphics_compiler/metrics

stuartmorgan: > See [dart-lang/pana#1324](https://github.com/dart-lang/pana/issues/1324)

I'm confused; I don't see anything in that issue that says that a conditional import based on `js_interop` wouldn't work, and it's currently working in practice. Could you point to a specific discussion/comment that makes you think this won't work in the future?

stuartmorgan:
Given that the published version using `dart.library.html` definitely doesn't work, `dart.library.js_interop` is currently working with pana, and we have already published other libraries using it, I'm going to go ahead and land this. If we get confirmation later that this will stop working for some reason, we will already need to do a follow-up PR for other packages in the dependency chain anyway, so we can just update this at the same time.

PR #8016, opened 2024-11-04
Error handling enhancement

minhhung2556:
### Title
- Error handling when loading and parsing SVG files from anywhere.
- Expose errorBuilder to build a widget that will be shown if it fails to parse an SVG picture.

## To fix these related issues below
- https://github.com/dnfield/vector_graphics/issues/219
- https://github.com/dnfield/flutter_svg/issues/1100
- https://github.com/dnfield/flutter_svg/issues/1085
- https://github.com/dnfield/flutter_svg/issues/1080
- https://github.com/dnfield/flutter_svg/issues/1044
- https://github.com/dnfield/flutter_svg/issues/996
- https://github.com/dnfield/flutter_svg/issues/1007

## My related PRs at the old repositories
- https://github.com/dnfield/flutter_svg/pull/1104
- https://github.com/dnfield/vector_graphics/pull/258

### Changes Made
- catch error cases and return results to SVGPicture.
- expose errorWidgetBuilder.
- write more unit tests.
- update example.

### Testing
- Already tested by example and unit test in flutter_svg

### Impact
- This change should not affect other parts of the library.

### Actions needed
- upgrade version of vector_graphics dependency inside flutter_svg package after new version of vector_graphics is live.


stuartmorgan:
Thanks for the submission!

In the future, please do not delete the checklist that is in the PR template; it is there for a reason. This PR is missing required elements described in the checklist (I’ve restored it to the PR description), which need to be addressed before it moves forward with review.

I am marking the PR as a Draft. Please review the checklist, updating the PR as appropriate, and when that’s complete please feel free to mark the PR as ready for review.

minhhung2556:
@stuartmorgan thanks for your advice, I've just updated it, please help me check if it is ok now.

packages/vector_graphics/lib/src/vector_graphics.dart
```diff
@@ -674,7 +686,7 @@ class VectorGraphicUtilities {
   ///
   /// It is the caller's responsibility to handle disposing the picture when
   /// they are done with it.
-  Future<PictureInfo> loadPicture(
+  Future<PictureInfo?> loadPicture(
```
stuartmorgan, line 689: This looks like a breaking change, so will need to be evaluated in the context of https://github.com/flutter/flutter/issues/157626

If the decision there is (as I hope) to start using semver for `vector_graphics*`, this would need to be versioned accordingly, or would need a non-breaking alternative.

jonahwilliams: Rather than making it null, we can throw/catch an exception I think.

PR #8029, opened 2024-11-07
[google_sign_in/google_identity_services] Clear-up documentation of callbacks in various APIs and uses of those APIs

srujzs:
These APIs used to accept allowInterop'd callbacks. When the implementation was migrated to use dart:js_interop, they were wrapped with toJS. However, starting in 3.5, to be consistent with dart2wasm, double-wrapping a function is an error. The documentation should be clear that it accepts only Dart functions and uses should be cleaned up to not use allowInterop and type the callbacks' parameter and return value correctly.

### DIFF
--- a/packages/google_identity_services_web/CHANGELOG.md
+++ b/packages/google_identity_services_web/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.3.1+5
 
 * Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
-* Clean up documentation of callbacks in `CodeClientConfig`,
+* Cleans up documentation of callbacks in `CodeClientConfig`,
   `TokenClientConfig`, `onGoogleLibraryLoad`, and `revoke` to indicate they only
   accept Dart functions and not JS functions.
 
--- a/packages/google_identity_services_web/CHANGELOG.md
+++ b/packages/google_identity_services_web/CHANGELOG.md
@@ -1,13 +1,13 @@
 ## NEXT
 
 * Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+* Clean up documentation of callbacks in `CodeClientConfig`,
+  `TokenClientConfig`, `onGoogleLibraryLoad`, and `revoke` to indicate they only
+  accept Dart functions and not JS functions.
 
 ## 0.3.1+4
 
 * Restores support for Dart `^3.3.0`.
-* Clean up documentation of callbacks in `CodeClientConfig`,
-  `TokenClientConfig`, `onGoogleLibraryLoad`, and `revoke` to indicate they only
-  accept Dart functions and not JS functions.
 
 ## 0.3.1+3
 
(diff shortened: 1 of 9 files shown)


packages/google_identity_services_web/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## NEXT
```
ditman, line 1:
```suggestion
## 0.3.1+5
```

ditman (changes requested):
Please, also update the `pubspec.yaml` to update this package to `0.3.1+5`. That way it'll get published, and the changes here will be visible in the pub docs.

ditman:
Please, roll this to `0.3.1+5` (both in the pubspec.yaml and CHANGELOG.md)

ditman (approved):
Please, roll this to `0.3.1+5` (both in the pubspec.yaml and CHANGELOG.md) so this can be auto-published!

packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
```diff
@@ -188,9 +188,9 @@ class GisSdkClient {
   //
   // Token clients have an additional `error_callback` for miscellaneous
   // errors, like "popup couldn't open" or "popup closed by user".
-  void _onTokenError(Object? error) {
+  void _onTokenError(GoogleIdentityServicesError? error) {
```
ditman, line 191: Woops, this change also requires a changelog entry and pubspec version, sorry I didn't notice the separate package!

srujzs: No worries! I just wasn't sure if it makes sense to have a changelog here, but done!

packages/google_identity_services_web/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
-## NEXT
+## 0.3.1+5
 
 * Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+* Clean up documentation of callbacks in `CodeClientConfig`,
```
stuartmorgan, line 4: Cleans

The PR checklist linked to a CHANGELOG style guide for this repo's packages.

packages/google_sign_in/google_sign_in_web/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.12.4+3
+
+* Fix callback types for `TokenClientConfig`.
```
stuartmorgan, line 3: Fixes

stuartmorgan (changes requested):

packages/google_sign_in/google_sign_in_web/CHANGELOG.md
srujzs: Done.

stuartmorgan (approved):
LGTM

PR #7521, opened 2024-08-27
[many] Upgrade example apps to AGP 8.5.2 (if they were below 8.1.0)

gmackall:
For https://github.com/flutter/flutter/pull/153795 - specifically the bump to `compileSdk 35` requires AGP 8.1.0+ (this isn't documented anywhere, but folks on https://issuetracker.google.com/issues/342522139 are also experiencing it).

Upgrades to AGP 8.5.2 and Gradle 8.7, for example apps that were using AGP less than 8.1.0.

Unblocks framework->packages roller

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ ] I [linked to at least one issue that this PR fixes] in the description above.
- [ ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [ ] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version
[...]

PR #8036, opened 2024-11-08
Use dependabot multi-directory configuration for Android example gradle updates

jmagman:
[Multi-directory config](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#configuration-options-for-the-dependabotyml-file) allows duplicated dependabot configs can be removed in favor of one that specifies multiple directories.  Update the Android example apps so they can be updated all at once.

Also updated the dependabot CI command to look for either `directory` or `directories` key.

Part of https://github.com/flutter/flutter/issues/148098

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -55,6 +55,7 @@ updates:
       - /packages/path_provider/path_provider/example/android/app
       - /packages/path_provider/path_provider_android/example/android/app
       - /packages/platform/example/android/app
+      - /packages/pigeon/example/app/android/app
       - /packages/pigeon/platform_tests/test_plugin/example/android/app
       - /packages/pigeon/platform_tests/alternate_language_test_plugin/example/android/app
       - /packages/quick_actions/quick_actions_android/example/android/app
@@ -64,6 +65,7 @@ updates:
       - /packages/rfw/example/remote/android/app
       - /packages/shared_preferences/shared_preferences/example/android/app
       - /packages/shared_preferences/shared_preferences_android/example/android/app
+      - /packages/two_dimensional_scrollables/example/android/app
       - /packages/url_launcher/url_launcher_android/example/android/app
       - /packages/url_launcher/url_launcher/example/android/app
       - /packages/video_player/video_player/example/android/app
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -25,21 +25,87 @@ updates:
...
+      - /packages/url_launcher/url_launcher_android/example/android/app
+      - /packages/url_launcher/url_launcher/example/android/app
+      - /packages/video_player/video_player/example/android/app
+      - /packages/video_player/video_player_android/example/android/app
+      - /packages/webview_flutter/webview_flutter/example/android/app
+      - /packages/webview_flutter/webview_flutter_android/example/android/app
+      - /third_party/packages/flutter_svg/example/android/app
     commit-message:
-      prefix: "[animations]"
+      prefix: "[gradle]"
```
jmagman, line 76: @stuartmorgan @reidbaker This change would mean we lose the PR title prefix, which is unfortunate. This seems worth it, but maybe some tooling/bot is relying on this prefix matching a package name?

reidbaker (approved):

.github/dependabot.yml
stuartmorgan: > maybe some tooling/bot is relying on this prefix matching a package name?

Shouldn't be, the prefixing convention in the repo is just for our convenience when wading through PRs and issues.

stuartmorgan (approved):
LGTM if Reid is happy with it. I'm a bit skeptical that grouping all the non-example ones together will be easier in practice, since usually some plugins are fine and some need changes, but I know the current system has some issues too.

jmagman:
> LGTM if Reid is happy with it. I'm a bit skeptical that grouping all the non-example ones together will be easier in practice, since usually some plugins are fine and some need changes, but I know the current system has some issues too.

This PR doesn't implement `groups`, which would combine the different dependencies into the same PR (what https://github.com/flutter/packages/pull/8044 is doing).  

This PR is only to deduplicate the configs (nicer anyway) so I can start adding groups in one place instead of 40. I'll work with Reid to figure out which `groups` make sense.

stuartmorgan:
Thanks, I totally missed that distinction :)

(I suspect a group for all example apps would actually be pretty nice.)

jmagman:
> This PR is only to deduplicate the configs (nicer anyway)

I think this change will also mean for a given dependency bump there will be one PR that updates all the packages, instead of one PR per package.  I'm not 100% sure though, we can see what it does...

stuartmorgan:
> I think this change will also mean for a given dependency bump there will be one PR that updates all the packages

That's what I thought originally. If so, almost certainly good for examples, TBD for non-examples.

jmagman:
Wait, is dependabot working at all for examples? I know it's only supposed to work for major versions, but I see zero dependabot PRs bumping it.

Example commit history https://github.com/flutter/packages/commits/main/packages/google_maps_flutter/google_maps_flutter/example/android/app/build.gradle

Why would @gmackall have had to do part of this manually?
https://github.com/flutter/packages/pull/7521/files#diff-c40bf8281656da0025c18216f38b473c0f0c018ccce9b04c57cf25f8a391b91fL8

jmagman:
I'm going to swap this to just update the examples, though it seems like dependabot hasn't actually been needed to update the major versions (or possibly isn't working?).

Reid suggests we should do all of them and see what happens, so I'll follow up the rest in another PR that we can easily revert.

PR #8041, opened 2024-11-08
[flutter_svg] rely on File from package:file

alestiago:
## Description

The `flutter_svg` package relies on an internal `File` abstraction implementation. For Web, the abstracted `File` is quite simplistic and doesn't inherit from a common `File` abstraction, which can be problematic (the `.file` signature changes type depending on the platform).

The Dart team owns [`package:file`](https://pub.dev/packages/file) which already takes care of providing a generic file system abstraction for Dart, such `File` abstraction is more feature complete. 

## Pre-launch Checklist

- [X] I signed the [CLA].


third_party/packages/flutter_svg/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.15
+
+* Changes internal `File` implementation in favor of `package:file`'s `File` implementation.
```
stuartmorgan, line 3: It's not internal, it's part of the public interface of the package. Changing a type in the public interface is a breaking change that would require a major version bump.

stuartmorgan (changes requested):
> * [ ]  I [linked to at least one issue that this PR fixes](https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview) in the description above.

Please file an issue describing the issue you are trying to solve in detail; the PR description is vague about what the use case is exactly; it just says it "can be problematic".

(I'm aware of https://github.com/flutter/flutter/issues/157741#issuecomment-2442571735, but that particular case isn't worth a major version change and a disruption to clients.)

third_party/packages/flutter_svg/CHANGELOG.md
alestiago: Removed, "internal", good catch!

alestiago:
If I get some spare bandwidth, I'll try filing an issue with some reproductive steps showing an scenario where it can be problematic.

third_party/packages/flutter_svg/CHANGELOG.md
stuartmorgan: Removing the word internal doesn't address the primary problem, which is that this this is a *breaking change*, which is being versioned as if it weren't, and isn't prominently described as such following our CHANGELOG style guide.

jonahwilliams:
I don't think we should make this change. Adding more dependencies to flutter_svg is not worth the cost

stuartmorgan:
> I don't think we should make this change. Adding more dependencies to flutter_svg is not worth the cost

There's also the possibility we could use `dart:io`'s `File`, but per https://github.com/flutter/flutter/issues/157741#issuecomment-2465100542 I'm not sure we could do that without breaking wasm.

third_party/packages/flutter_svg/CHANGELOG.md
alestiago: Apologies, I was multitasking and didn't read through the entire comment. You are correct, since an `io.File` instance is not of type `file.File` it would be a breaking change.

I would also like to avoid a breaking change but would want to improve the current DX around the .file constructor for Web.

I'm going to close this PR for now and if I have any bandwidth I will open a related issue. 

Note that, in theory the `.file` constructor is syntactic sugar that the package offers, since the same can be achieved with other `ByteLoaders`; for example by reading the File content first (either bytes or String) and then using an `SvgBytesLoader` or `SvgStringLoader`. Thus, the package could remove the need of `File` entirely.

alestiago:
Closing, refer to https://github.com/flutter/packages/pull/8041#discussion_r1835143955.

PR #8043, opened 2024-11-08
[tools] Format Dart per-package

stuartmorgan:
Instead of running Dart formatting on the whole repo at once, run it per package, from the package's directory. This is slower, but necessary since the new formatter behaves differently depending on the package's min SDK version.


camsim99 (approved):
This LGTM! Thanks for jumping on this!

stuartmorgan:
Landing now to unblock the roller; none of the tests that are still running can be affected by these changes.

PR #8052, opened 2024-11-11
[tools] Run `pub get` before `format`

stuartmorgan:
The new Dart formatter needs to know the Dart language version of the code it is formatting, and it reads that from a file in `.dart_tool`, not `pubspec.yaml` directly. To avoid it failing to determine the version and assuming the latest (which will almost always be wrong in this repo):
- Adds a step to the `format` repo command to ensure that `pub get` appears to have been run, and runs it if not, and
- To avoid `pub get` running in `format` in CI, adds a deps-fetching step to the `repo_checks` task, as we have in other tasks that need to `pub get`.

This should unblock the roll.

### DIFF
--- a/script/tool/lib/src/format_command.dart
+++ b/script/tool/lib/src/format_command.dart
@@ -12,6 +12,7 @@ import 'package:meta/meta.dart';
 import 'common/core.dart';
 import 'common/output_utils.dart';
 import 'common/package_looping_command.dart';
+import 'common/pub_utils.dart';
 import 'common/repository_package.dart';
 
 /// In theory this should be 8191, but in practice that was still resulting in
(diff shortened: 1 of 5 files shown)


script/tool/lib/src/format_command.dart
```diff
@@ -120,6 +121,15 @@ class FormatCommand extends PackageLoopingCommand {
       relativeTo: package.directory,
     );
     if (getBoolArg(_dartArg)) {
+      // Ensure that .dart_tool exists, since without it `dart` doesn't know
+      // the lanugage version, so the formatter may give different output.
+      if (!package.directory.childDirectory('.dart_tool').existsSync()) {
```
ditman, line 126: Should we warn through the console in case this step needs to be performed?

That would let us know if something like `flutter clean` gets added by mistake?

ditman (approved):
LGTM!

script/tool/lib/src/format_command.dart
stuartmorgan: We pass stdout/stderr through in `runPubGet`, so it's super obvious in the console when this is happening.

stuartmorgan:
Since this only affects `repo_checks` and `format`, and those tasks/steps have completed and passed, I'm going to land this without waiting for (and de-flaking) the rest of the tests, to unblock the roller.

PR #8053, opened 2024-11-11
[google_identity_services_web] Migrate all types to JSObject

kevmoo:
And cleanup other legacy JS-interop
Also update tests so they pass with WebAssembly


packages/google_identity_services_web/example/integration_test/utils.dart
```diff
@@ -33,6 +49,12 @@ ExpectConfigValueFn createExpectConfigValue(JSObject config) {
 Matcher isAJs(String thing) => isA<JSAny?>()
     .having((JSAny? p0) => p0.typeofEquals(thing), 'typeof "$thing"', isTrue);
 
+final Matcher isJSArray = isA<JSAny?>()
+    .having((JSAny? p0) => _isArray(p0).toDart, 'Array.isArray', isTrue);
```
srujzs, line 53: It's up to you if you'd rather do an `isA<JSArray>` (`isA` being from `dart:js_interop` and not `package:test`) or what you have here. It probably shouldn't matter but `Array.isArray` is a little stricter.

srujzs (approved):
Thanks again for doing this Kevin!

kevmoo:
We have dart functions in these types...that probably should be JSFunction, right?

srujzs:
> We have dart functions in these types...that probably should be JSFunction, right?

Can you share an example of what you're referring to?

ditman:
Which reminds me that we need to run our web integration tests with `--wasm` enabled! I'll merge this ASAP.

ditman (approved):
LGTM!

packages/google_identity_services_web/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## 0.3.3
 
-* Move all the JavaScript types to extend `JSObject`.
+* Migrates all the JavaScript types to the latest syntax.
```
kevmoo, line 3: Yeah, but that doesn't tell the user WHAT changed. The fact that these types are now `JSObject` *is* the user-visible change.

ditman: Users never see these objects, they don't care about the underlying representation, but let me revert.

PR #8057, opened 2024-11-12
[interactive_media_ads] Adds support to define parameters that control the rendering of ads

bparrishMines:
Changes `PlatformAdsManager.init(AdsManagerInitParams)` to the correct shared `PlatformAdsManager.init(PlatformAdsRenderingSettings?)` feature.

**Android:** https://developers.google.com/interactive-media-ads/docs/sdks/android/client-side/api/reference/com/google/ads/interactivemedia/v3/api/BaseManager.html#init(com.google.ads.interactivemedia.v3.api.AdsRenderingSettings)
**iOS:** https://developers.google.com/interactive-media-ads/docs/sdks/ios/client-side/reference/Classes/IMAAdsManager#/c:objc(cs)IMAAdsManager(im)initializeWithAdsRenderingSettings:

The platform interface and implementation aren't exposed yet, so this change should not be breaking.

Fixes https://github.com/flutter/flutter/issues/152253

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/interactive_media_ads/lib/src/platform_interface/platform_ads_rendering_settings.dart
```diff
@@ -0,0 +1,134 @@
...
+  final int? bitrate;
+
+  /// If set, the SDK will instruct the player to load the creative in response
+  /// initializing the ads manager.
+  ///
+  /// This allows the player to preload the ad at any point before starting the
+  /// ads manager.
+  ///
+  /// If null, the platform will decide the default value.
+  final bool? enablePreloading;
```
bparrishMines, line 76: Note: This value defaults to `false` on Android and `true` on iOS. To prevent a breaking change, I made the value nullable to indicate the platform will handle the default. Same is true for the other fields.

packages/interactive_media_ads/lib/src/ios/ios_ads_manager.dart
```diff
@@ -30,8 +31,16 @@ class IOSAdsManager extends PlatformAdsManager {
...
+  Future<void> init([PlatformAdsRenderingSettings? settings]) {
+    if (settings == null) {
+      return _manager.initialize(null);
+    }
+
+    final IMAAdsRenderingSettings nativeSettings =
+        settings is IOSAdsRenderingSettings
+            ? settings.nativeSettings
+            : IOSAdsRenderingSettings(settings.params).nativeSettings;
+    return _manager.initialize(nativeSettings);
```
stuartmorgan, line 43: Same note here about not duplicating the `initialize` call.

packages/interactive_media_ads/lib/src/platform_interface/platform_ads_manager.dart
```diff
@@ -20,7 +18,7 @@ abstract class PlatformAdsManager {
   PlatformAdsManager();
 
   /// Initializes the ad experience using default rendering settings.
-  Future<void> init(AdsManagerInitParams params);
+  Future<void> init([PlatformAdsRenderingSettings? settings]);
```
stuartmorgan, line 21: This should be a named parameter unless we're very sure there will never be any other init params.

packages/interactive_media_ads/lib/src/platform_interface/platform_ads_rendering_settings.dart
```diff
@@ -0,0 +1,134 @@
...
+
+  /// Specifies a non-default amount of time to wait for media to load before
+  /// timing out, in milliseconds.
+  ///
+  /// This only applies to the IMA client-side SDK. Default time is 8000 ms.
+  final int? loadVideoTimeout;
```
stuartmorgan, line 82: `Duration`?

packages/interactive_media_ads/lib/src/platform_interface/platform_ads_rendering_settings.dart
```diff
@@ -0,0 +1,134 @@
...
+
+  /// The SDK will prioritize the media with MIME type on the list.
+  ///
+  /// If empty, the SDK will pick the media based on player capabilities. This
+  /// only refers to the mime types of videos to be selected for linear ads.
+  final List<String>? mimeTypes;
```
stuartmorgan, line 88: Non-nullable?

packages/interactive_media_ads/lib/src/platform_interface/platform_ads_rendering_settings.dart
```diff
@@ -0,0 +1,134 @@
...
+
+  /// For VMAP and ad rules playlists, only play ad breaks scheduled after this
+  /// time (in seconds).
+  ///
+  /// This setting is strictly after the specified time. For example, setting
+  /// `playAdsAfterTime` to 15 will ignore an ad break scheduled to play at 15s.
+  final double? playAdsAfterTime;
```
stuartmorgan, line 95: `Duration`?

packages/interactive_media_ads/lib/src/platform_interface/platform_ads_rendering_settings.dart
```diff
@@ -0,0 +1,134 @@
...
+
+  /// Sets the ad UI elements to be rendered by the IMA SDK.
+  ///
+  /// Some modifications to the uiElements list may have no effect for specific
+  /// ads.
+  final Set<UIElement>? uiElements;
```
stuartmorgan, line 101: Non-nullable?

packages/interactive_media_ads/lib/src/platform_interface/ui_element.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Describes an element of the ad UI, to be requested or rendered by the SDK.
+enum UIElement {
```
stuartmorgan, line 6: This seems like a very generic name that could end up require people to use prefixing; what about `AdUIElement`?

stuartmorgan (changes requested):

packages/interactive_media_ads/lib/src/ads_rendering_settings.dart
```diff
@@ -0,0 +1,80 @@
...
+  /// [AdsManager.start].
+  ///
+  /// If null, the platform will decide the default value.
+  bool? get enablePreloading => platform.params.enablePreloading;
+
+  /// Specifies a non-default amount of time to wait for media to load before
+  /// timing out, in milliseconds.
+  ///
+  /// This only applies to the IMA client-side SDK. Default time is 8000 ms.
+  int? get loadVideoTimeout => platform.params.loadVideoTimeout;
```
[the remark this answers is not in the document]
bparrishMines: Good suggestion!

packages/interactive_media_ads/lib/src/android/android_ads_rendering_settings.dart
```diff
@@ -0,0 +1,110 @@
...
+      playAdsAfterTime: params.playAdsAfterTime,
+      uiElements: params.uiElements,
+      enableCustomTabs: enableCustomTabs,
+    );
+  }
+
+  final InteractiveMediaAdsProxy _proxy;
+
+  /// Notifies the SDK whether to launch the click-through URL using Custom Tabs
+  /// feature.
```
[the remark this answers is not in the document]
bparrishMines: My thought process was that `null` essentially meant that the native value will not be set with any value. But since the documentation states that `false` is the default, it is probably safe to always set it and make the default on the Dart side `false`. 

I was actually considering what a better design for setting optional native parameters. Is it better to always set a value if the default is stated or provide an option to not set it (e.g. `null`).

packages/interactive_media_ads/lib/src/platform_interface/platform_ads_rendering_settings.dart
bparrishMines: Only the documentation for iOS actually states what happens if it is empty. Android doesn't actually mention it:

```
If specified, the SDK will prioritize the media with MIME type on the list. If not specified,
the SDK will pick the media based on player capabilities. This only refers to the mime
types of videos to be selected for linear ads.
```

This is related to having a "don't set" option for optional native parameters.

packages/interactive_media_ads/lib/src/platform_interface/platform_ads_rendering_settings.dart
bparrishMines: Same here. Not sure if there is a difference between setting to empty or not setting it all.

packages/interactive_media_ads/lib/src/platform_interface/platform_ads_rendering_settings.dart
stuartmorgan: That's an implementation detail though; if we want empty to mean "the SDK decides" as the docs currently say, then we should map an empty list to not setting on platforms that treat null as not setting.

If we actually want to treat `[]` and `null` differently, then we need to explain what the difference is in the comments, ideally in a cross-platform way.

stuartmorgan (approved):
LGTM

PR #8074, opened 2024-11-13
[tool] Improve Swift lint error message

stuartmorgan:
If the Swift linter finds issues, log a message that communicates that rather than sounding like the linter failed to run.


jmagman (approved):
Thanks!

PR #7849, opened 2024-10-11
[pigeon] Allows swift generator to skip error class generation

wamynobe:
Since the Kotlin generator allows skipping error class generation, it makes sense for the Swift generator to have the same option.

Related to:
#6183 
https://github.com/flutter/flutter/issues/142099

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version
[following repository CHANGELOG style]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style
[exempt from CHANGELOG changes]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog
[...]

PR #8087, opened 2024-11-14
[pigeon] Updates `README.md` with examples of using `@ProxyApi`

bparrishMines:
## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/example/README.md
```diff
@@ -284,7 +284,7 @@ private class PigeonFlutterApi {
     aString aStringArg: String?, completion: @escaping (Result<String, Error>) -> Void
   ) {
     flutterAPI.flutterMethod(aString: aStringArg) {
-      completion(.success($0))
+      completion(.success(try! $0.get()))
```
bparrishMines, line 287: Code didn't compile until I fixed this. I'm assuming our CI doesn't verify the example compiles. Is this unintentional?

I also had to fix the same line for Kotlin.

packages/pigeon/tool/shared/generation.dart
```diff
@@ -118,7 +118,7 @@ Future<int> generateTestPigeons(
           ? null
           : '$outputBase/ios/Classes/$pascalCaseName.gen.swift',
       swiftErrorClassName: swiftErrorClassName,
-      swiftIncludeErrorClass: input != 'core_tests',
+      swiftIncludeErrorClass: input != 'background_platform_channels',
```
bparrishMines, line 121: Fixes a test failure caused by https://github.com/flutter/packages/pull/7849

`core_tests` and `background_platform_channels` both generated the `PigeonError` class which causes conflicts. 

**tldr:**
This wasn't caught by CI since `core_tests` is a checked in and the order of the tests matters.

When I reran the generator and pushed my code, it removed the `PigeonError` from `core_tests`. Therefore the tests that run before the CI generates the unchecked in files would fail because `core_tests` was missing the reference to `PigeonError` that was in `background_platform_channels`. Therefore the fix was to switch the location the single `PigeonError` to reside in `core_tests` instead of `background_platform_channels`.

packages/pigeon/example/README.md
tarrinneal: Just ran into this in my pr as well. Not sure why the examples are getting compiled in ci though

packages/pigeon/tool/shared/generation.dart
tarrinneal: This is also addressed in my current pr. Your solution is smarter than mine :)

stuartmorgan:
We should hold off on this until I've had a chance to discuss with folks about overall interop plans.

packages/pigeon/example/README.md
stuartmorgan: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8731158450509496993/+/u/Run_package_tests/build_examples/stdout

```
============================================================
|| Running for packages/pigeon [@0:00]
============================================================

Building for: Android

Skipping Android for pigeon/example; not supported.
SKIPPING: No examples found supporting requested platform(s).
```

O.o

Let me look into what's going on here.

stuartmorgan: Ah, it's because the actual example is `example/app/`, not `example/`, but `example.pubspec.yaml` exists so the tooling sees that `example` is a package.

I'll play with fixing this in a separate PR. I think the pubspec may have only been there for the old excerpt system.

PR #8098, opened 2024-11-14
[pigeon] Adds Quick start guide to README

craiglabenz:
Adds a single "Quick start" guide to using Pigeon to the main README. Increases concept / explanation of Pigeon concepts alongside code samples.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+// To get started, see the Quickstart Guide in the [Example README](./example/README.md).
+
 Pigeon is a code generator tool to make communication between Flutter and the
 host platform type-safe, easier, and faster.
 
-Pigeon removes the necessity to manage strings across multiple platforms and languages.
-It also improves efficiency over common method channel patterns. Most importantly though,
-it removes the need to write custom platform channel code, since pigeon generates it for you.
+Pigeon works by reading special files which by convention are placed in a separate
+folder, `/pigeons`, alongside your `/lib` folder. You can put all of your definitions
```
stuartmorgan, line 10: This is a flutter/packages repo convention; there's no reason for it to be a convention anywhere else (or for there to even be a convention in general outside of our repo).

packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+in a single file (e.g., `/pigeons/messages.dart`), or you can divide your methods
+and classes into separate files as you wish. After reading these files, Pigeon
```
stuartmorgan, line 12: Using separate files has pitfalls and complexity; we shouldn't suggest it in the initial overview.

packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+generates:
+
+ * New Dart source files within your `/lib` folder at the specific paths you
+specify, and
+ * New native source files with matching code within your native projects at the
+ specific paths you specify
```
stuartmorgan, line 18: These seem very verbose; do we think people aren't familiar with the concept of a code generator? It seems like we could just say that it generates Dart and native code.

packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+
+Internally, the generated code uses `MethodChannel`s to communicate across Flutter's
+UI thread and the Platform thread where your host app initially launched. The value
+in Pigeon comes from automatically keeping this unpleasant boilerplate in sync and
+efficiently marshalling data between languages.
```
stuartmorgan, line 23: This is an internal implementation detail, not something people generally need to know.

packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+
+The generated code can either flow from Dart to native code, or native code back
+to Dart. Generated code on the receiving end uses interfaces or abstract classes,
+allowing you to provide implementations in concrete classes.
+
+Pigeon works in both complete Flutter apps and hybrid apps using the add-to-app paradigm.
+
+## Quickstart
```
stuartmorgan, line 31: There is a lot of overlap between this section and the existing `Usage` section; we shouldn't have both as separate sections.

packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+Begin by adding Pigeon to your Dart project's `pubspec.yaml` file:
+
+```sh
+$ flutter pub add pigeon --dev
+$ flutter pub get
+```
+
+### Setup
+
+To specify what code Pigeon should generate, create a `/pigeons` directory and add
```
stuartmorgan, line 44: I don't think we should be telling people they have to use this convention.

packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+a `.dart` configuration file. This guide will place all such configurations within
```
stuartmorgan, line 45: I would describe it as an interface definition file, not a configuration file.

packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+`/pigeons/messages.dart`, but you are free to choose any file name(s) you like.
+
+Begin by instantiating a `PigeonOptions` object, wrapped in the `@ConfigurePigeon`
```
stuartmorgan, line 48: This seems like a long-form tutorial rather than a quick-start.

Maybe a codelab would be a better place for some of this content?

packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+  Code code;
+  Map<String, String> data;
+}
+```
+
+No extra steps are necessary to register these classes - their inclusion
+in a `pigeons/*.dart` ensures Pigeon will generate matching Dart and native
+implementations.
+
+### Define which methods to expose
```
stuartmorgan, line 88: It's not clear to me how this step is different from the previous step.

packages/pigeon/example/README.md
```diff
@@ -1,48 +1,13 @@
...
 # Pigeon Examples
 
+For a quick start guide to Pigeon, see the [QuickStart guide](../README.md#Quickstart). For language-specific guides to instantiate code generated by Pigeon, continue with
+this guide.
+
 The examples here will cover basic usage. For a more thorough set of examples,
 check the [core_tests pigeon file](../pigeons/core_tests.dart) and 
 [platform test folder](../platform_tests/) ([shared_test_plugin_code](../platform_tests/shared_test_plugin_code/) and [alternate_language_test_plugin](../platform_tests/alternate_language_test_plugin/) especially).
 
-## Invocation
```
stuartmorgan, line 8: This removed sections seems like a much quicker quick-start than the new section that's called a quick start.

packages/pigeon/README.md
```diff
@@ -1,13 +1,354 @@
...
+
+The sample code in this Quickstart defines an example native API named `ExampleHostApi`.
+Pigeon will generate a complete Dart implementation and the equivalent of an abstract
+class in each native language (for example, a `protocol` in Swift and an `interface`
+in Kotlin).
+
+See the [language-specific guides](example/README.md#HostApi-Example) for help instantiating the native classes
+generated by Pigeon.
+
+In Dart, typical use within a `StatefulWidget` might look like this:
```
stuartmorgan, line 309: I don't think we have any code in our repo that looks like this, so I would not say this is typical use. I'm not sure why we need an example for this at all.

stuartmorgan-g (commented):
I started leaving specific comments, but the further I got into it the more I think we need to start from a discussion about what the intended audience is, and where the content should live. Overall, this seems much more like something I would expect to see in a codelab or on an updated version of https://docs.flutter.dev/platform-integration/platform-channels than in the README; our package READMEs are generally not structured as tutorials.

packages/pigeon/README.md
craiglabenz: [This code](https://github.com/flutter/packages/blob/main/packages/pigeon/example/app/lib/main.dart#L55-L59) (and on down) is what inspired it.

FWIW, this whole PR was inspired by me reading Pigeon's documentation, still being quite confused about the high level picture, and then having to piece it together from sample source code. My goal is to elevate the high level picture with snippets to the top-level README.

packages/pigeon/example/README.md
craiglabenz: It was shorter, that's true, but at least for me it was definitely not quicker. I also said this in an earlier comment, but in my first experience using Pigeon, I found myself searching through sample code to fill in large mental gaps in how everything was supposed to work.

packages/pigeon/README.md
craiglabenz: Good to know, thanks.

packages/pigeon/README.md
craiglabenz: FWIW, Our existing documentation does mention efficiency gains which personally made me wonder, "O RLY? Like what?"

packages/pigeon/README.md
craiglabenz: > I'm not sure why we need an example for this at all.

Pigeon is great, but I thinks fundamental API shape has a decent learning curve (writing a pseudo-class, generating an abstract class off that, defining a concrete class and registering it with the generated abstract class). It's all fairly unusual, so my general sense here is that we're currently short of context and perspective explainers around code samples.

packages/pigeon/README.md
craiglabenz: Is it fair to say that these files should be outside of the `/lib` folder?

tarrinneal: I think it should be outside the `/lib` folder, yes. It isn't code that is actively used by anything other than during pigeon generation.

packages/pigeon/README.md
stuartmorgan: > [This code](https://github.com/flutter/packages/blob/main/packages/pigeon/example/app/lib/main.dart#L55-L59) (and on down) is what inspired it.

That code is pretty artificial, due to the constraints of trying to fit an example of Pigeon usage into a single-file application.

> thinks fundamental API shape has a decent learning curve (writing a pseudo-class, generating an abstract class off that, defining a concrete class and registering it with the generated abstract class). It's all fairly unusual, so my general sense here is that we're currently short on context and perspective explainers around code samples.

Sure, but none of that is related to `StatefulWidget`. I would rather the readme focus on important concepts that are fundamental to using Pigeon, rather than spending space on the unrealistic example of embedding usage directly into a `StatefulWidget`.

stuartmorgan:
If we are going to overhaul the in-package docs, I think the first step is to decide what should be in `README.md` and what should be in `example/README.md`, because this version seems like it's adding a lot of overlap between the two.

packages/pigeon/README.md
stuartmorgan: These changes don't describe the efficiency gains, so I'm not sure how that related to adding this paragraph.

If we want to add a technical discussion somewhere of how Pigeon improves over the common pattern of using string-keyed dictionaries to send multiple parameters we could, but that would be completely orthogonal to making it easier to understand how to use Pigeon (and IMO wouldn't be a great use of README space).

packages/pigeon/README.md
stuartmorgan: > `Define your messages` intends to describe defining the data structures that will be passed back and forth; whereas `Define which methods to expose` is about the `HostApi()` and `FlutterApi()` decorators which expose a class with methods.

Many uses of Pigeon don't even require data structures; I don't see these are separate steps (and if they were, certainly wouldn't put defining data structures *before* figuring out what the methods are going to be that would need them).

packages/pigeon/README.md
```diff
@@ -3,11 +3,341 @@
...
+```
+
+No extra steps are necessary to register these classes - their inclusion
+in your definitions file ensures Pigeon will generate matching Dart and native
+implementations.
+
+### Define which methods to expose
+
+The point of Pigeon and the `MethodChannel`s it utilizes is to call native functions
+living on the Platform thread from Dart, or to call Dart functions living on the UI
```
stuartmorgan, line 82: This language seems very confusing to me; functions don't live on threads, calls are on threads. There is no inherent binding between a function and a thread.

stuartmorgan:
From triage: @tarrinneal will look over this again and see what makes sense in package docs vs a codelab.

tarrinneal:
Would it be a bad idea to have this as it's own .md file? We can keep the shorter README with a link up top suggesting a more in-depth getting started is [link to guide]?

tarrinneal:
If not in an alternate file, I think a codelab is probably a good idea. I'd prefer to keep the README concise if possible.

stuartmorgan:
From triage: @craiglabenz any concern with moving most of this content to a codelab?

stuartmorgan-g:
Converting to Draft while the question of having this be codelab content is evaluated.

stuartmorgan-g:
From triage: closing since the current consensus on the ecosystem team is that this level of details is better suited for a codelab.

PR #8102, opened 2024-11-15
[google_maps_flutter] Fix detection of WebAssembly compatibility

kevmoo:


packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
-## NEXT
+## 2.10.0
 
-* Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
+* Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+* Fix platform detection so this package correctly reports compatibility with
```
stuartmorgan, line 4: Fixes

stuartmorgan, line 4: A changelog for a package doesn't need to say that the changes are about "this package". How about just "Fixes detection of WebAssembly compatibility."?

packages/google_maps_flutter/google_maps_flutter/lib/src/platform.dart
stuartmorgan, line 1: This seems very unscalable; needing to do platform detection is not an uncommon problem. Has there been discussion of having a Dart-team-owned packages that does this abstraction internally rather than every package having to do it?

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -594,7 +594,7 @@ class _GoogleMapState extends State<GoogleMap> {
 
 /// Builds a [MapConfiguration] from the given [map].
 MapConfiguration _configurationFromMapWidget(GoogleMap map) {
-  assert(!map.liteModeEnabled || Platform.isAndroid);
+  assert(!map.liteModeEnabled || isAndroid);
```
stuartmorgan, line 597: If this is the only place we are checking platform, let's just drop the assert. It's fine to have it ignored silently instead of failing assertions, as it's documented that it's ignored on other platforms.

stuartmorgan (changes requested):

packages/google_maps_flutter/google_maps_flutter/lib/src/platform.dart
kevmoo: https://pub.dev/packages/platform ?

kevmoo: That needs to be fixed for wasm. But that should be easy!

kevmoo: I take it back. That package is not designed to work on the web at all..

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
kevmoo: done

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
kevmoo: done

kevmoo:
PTAL – also ponder test exemption

stuartmorgan (approved):
LGTM

stuartmorgan:
test-exempt: removing code

PR #8037, opened 2024-11-08
Applied Gradle Plugins Declaratively for Most Remaining Plugin Example Apps (Part 3)

jesswrd:
Updated applying gradle plugins from usage of imperative apply to usage of declarative blocks {} apply for 25 remaining plugin example apps. Intending on updating all android example apps under packages. 

Here are previous bulk migrations from imperative apply to declarative apply:
https://github.com/flutter/packages/pull/7968
https://github.com/flutter/packages/pull/8019

Previously migrated applying path_provider plugin example app from imperatively to declaratively in this PR [here](https://github.com/flutter/packages/pull/7822). Tests for changes in the linked PR above.

More information on Flutter Gradle Pluggin Apply [here](https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply)

All instances of `minSdkVersion` for example plugin apps use `flutter.minSdkVersion` instead of the hard-coded version https://github.com/flutter/packages/pull/8035.

Partially addresses [#152656](https://github.com/flutter/flutter/issues/152656)

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[...]

PR #8109, opened 2024-11-15
Temporarily Disabling Flaky Integration Tests in Plugin Example Apps

jesswrd:
Temporarily skipping flaky integration tests in plugin example apps for camera, camera_android, and camera_android_camerax to safely land https://github.com/flutter/packages/pull/8037.

The flakes are unrelated to the mentioned PR and can likley be traced back to the following issues:
https://github.com/flutter/flutter/issues/154682
https://github.com/flutter/flutter/issues/157556

Partially addresses https://github.com/flutter/flutter/issues/152656

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera/example/integration_test/camera_test.dart
+++ b/packages/camera/camera/example/integration_test/camera_test.dart
@@ -94,7 +94,7 @@ void main() {
         await controller.dispose();
       }
     }
-  });
+  }, skip: true);
 
   // This tests that the capture is no bigger than the preset, since we have
   // automatic code to fall back to smaller sizes when we need to. Returns
--- a/packages/camera/camera/example/integration_test/camera_test.dart
+++ b/packages/camera/camera/example/integration_test/camera_test.dart
@@ -94,6 +94,7 @@ void main() {
         await controller.dispose();
       }
     }
+    // TODO(camillesimon): Re-enable test when issue is fixed https://github.com/flutter/flutter/issues/154682.
   }, skip: true);
 
   // This tests that the capture is no bigger than the preset, since we have
--- a/packages/camera/camera_android/example/integration_test/camera_test.dart
+++ b/packages/camera/camera_android/example/integration_test/camera_test.dart
@@ -92,7 +92,7 @@ void main() {
         await controller.dispose();
       }
     }
-  });
+  }, skip: true);
 
   // This tests that the capture is no bigger than the preset, since we have
   // automatic code to fall back to smaller sizes when we need to. Returns
@@ -197,7 +197,7 @@ void main() {
     await videoController.dispose();
 
     expect(duration, lessThan(recordingTime - timePaused));
-  });
+  }, skip: true);
 
   testWidgets('Set description while recording', (WidgetTester tester) async {
     final List<CameraDescription> cameras =
@@ -234,7 +234,7 @@ void main() {
       // cameras switched
       expect(controller.description, cameras[1]);
     }
-  });
+  }, skip: true);
 
   testWidgets('Set description', (WidgetTester tester) async {
     final List<CameraDescription> cameras =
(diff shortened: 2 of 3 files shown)


matanlurey (changes requested):
Let's leave a TODO linking to an issue to renable these tests (could repurpose the ones you have already for Camille), i.e. at every line where you wrote `skip: true`. There should be at least one example in the repo if you want a reference.

This is important because otherwise we permanently skip test coverage which can have worse implications down the line.

Thanks for creating the PR!

jesswrd:
Thanks for the explanation! done.

matanlurey (approved):

camsim99 (approved):
LGTM! Thanks for this :)

stuartmorgan:
Are we sure these are flaky tests? The referenced PR is creating a situation that one of the referenced issues says will break the test, and the PR appears to be failing *every* Android camera package, even with retries. What's the evidence that this is pre-existing flake rather than the PR actually breaking the tests?

camsim99:
@stuartmorgan can confirm this test is broken by something in the plugin not flaky and should be disabled until I can investigate.

https://github.com/flutter/flutter/issues/154682 has evidence of it failing locally in the past.
b/376074365 has further evidence of this previously being broken.

I know these don't reference `camera` but if those run on Android expect them to fail too because I'm guessing that CameraX doesn't support lower resolutions anymore but need to look into it this week. I believe the test only appears to be flaky because it doesn't run all of the time.

If you need more proof, let me know and I can validate its broken-ness to help unblock https://github.com/flutter/packages/pull/8037!

camsim99:
Ah I just noticed https://github.com/flutter/packages/pull/8037#issuecomment-2480826738. It's possible that the bugs I've seen are related to bumping the target SDK version (though I don't think the b/376074365 is; it appears it's even more broken after the target SDK bump).

@stuartmorgan would a good compromise be not bumping the target SDK version until I fix the tests? @jesswrd not sure if your change https://github.com/flutter/packages/pull/8037 relies on that or not.

jesswrd:
> @stuartmorgan would a good compromise be not bumping the target SDK version until I fix the tests? @jesswrd not sure if your change #8037 relies on that or not.

@camsim99 I can't merge without bumping the target sdk version. Google Play requires a target sdk version of at least 33, so I upgraded the target sdk version for `camera`, `camera_android`, and `camera_android_camera_androidx`. I get an error if I don't. 

@stuartmorgan Given the above, I think I should disable the tests but let me know how you'd like me to proceed.

stuartmorgan:
> [flutter/flutter#154682](https://github.com/flutter/flutter/issues/154682) has evidence of it failing locally in the past.

Right, but it was failing in a PR that tried to bump the SDK version. And so the camera part was backed out of that PR. That's not flake, that's deterministic failure under the same conditions as the new PR.

It just seems like there's a mismatch here, where this PR describes the failures as "flakes [...] unrelated to the mentioned PR", but what I'm seeing is significant evidence that the failures are in fact directly caused by that PR, for reasons that have been previously established last time we tried to roll the target SDK.

> I believe the test only appears to be flaky because it doesn't run all of the time.

It runs any time anyone touches `camera` or and android sub-package, and also any time all tests run (such as every roll). If it were failing consistently on `main`, we would definitely know.

AFAICT the state is that it's fine on `main`, but attempting to roll the target SDK forward will break it. If we believe that the test is wrong and we want to disable it to roll the target SDK forward we can, but we should be clear that we are doing that—explicitly disabling a working test in order to land a PR that breaks that test—not disabling a flaky test.

camsim99:
> AFAICT the state is that it's fine on main, but attempting to roll the target SDK forward will break it. If we believe that the test is wrong and we want to disable it to roll the target SDK forward we can, but we should be clear that we are doing that—explicitly disabling a working test in order to land a PR that breaks that test—not disabling a flaky test. (And I would suggest we do the disabling in the PR that breaks the tests, in that case.)

This SGTM!

PR #8115, opened 2024-11-17
[local_auth] Allow empty localizedReason for Android

Luciferx86:
localizedReason should be optional for Android. There could be use case where we only want to show a single piece of text on the biometric prompt. Currently showing 2 pieces of texts is mandatory.

## Proposal

Only set the description parameter to the BiometricPrompt if a non empty String is passed in the call.

code:

AuthenticationHelper.java

use
```java
.setDescription(strings.getReason().isEmpty() ? null : strings.getReason())
```

instead of

```java
.setDescription(strings.getReason())
```

** screenshots attached in the original issue

Fixes https://github.com/flutter/flutter/issues/116885

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/local_auth/local_auth_android/CHANGELOG.md
+++ b/packages/local_auth/local_auth_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.47
+
+* Allowing empty localizedReason for Android
+
 ## 1.0.46
 
 * Updates Java compatibility version to 11.
--- a/packages/local_auth/local_auth_android/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
+++ b/packages/local_auth/local_auth_android/android/src/main/java/io/flutter/plugins/localauth/AuthenticationHelper.java
@@ -70,6 +70,10 @@ class AuthenticationHelper extends BiometricPrompt.AuthenticationCallback
 
     BiometricPrompt.PromptInfo.Builder promptBuilder =
         new BiometricPrompt.PromptInfo.Builder()
+            .setDescription(
+                strings.getReason().isEmpty()
+                    ? null
+                    : strings.getReason())
             .setDescription(strings.getReason())
             .setTitle(strings.getSignInTitle())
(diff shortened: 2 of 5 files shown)


stuartmorgan (changes requested):
Thanks for the contribution!

This PR makes the Android implementation inconsistent with the public documentation of `authenticate`, which says (at both the app-facing and platform interface level) that the parameter must not be empty. That would mean that only people who read the implementation source would know that they could do the opposite of what the docs say, which is not how we would want to land this.

A complete fix needs to update the docs on those methods, as well as provide clients a runtime way of knowing whether or not the current platform requires a non-empty reason, via a [support query](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#api-support-queries). So this needs a new platform interface method, called something like `requiresLocalizedReason()`, and for that to be wired up in the app-facing package as part of changing the documention for `authenticate`.

Luciferx86:
I understand that there needs to be a method like `requiresLocalizedReason()` for all the platforms, and for `android` in particular, this method will return false.

So to make this work, I would be required to update the documentation for `authenticate` method, as well as override this method in all platform implementations? 

Also implementation for authenticate then becomes something like?

```dart
if(requiresLocalizedReason()){
assert(localizedReason.isNotEmpty);
}
```

Is this implementation correct, or you had something else in mind? @stuartmorgan

stuartmorgan:
> as well as override this method in all platform implementations?

Since there needs to be a default implementation that returns `true` for backwards compatibility, it only actually needs to be overridden in the Android implementation.

> Also implementation for authenticate then becomes something like?
> 
> ```dart
> if(requiresLocalizedReason()){
> assert(localizedReason.isNotEmpty);
> }
> ```

The assertions are currently at the platform implementation package level; every platform implementation already knows whether it needs the assertion or not, without a new conditional.

stuartmorgan:
@Luciferx86 Are you still planning on updating this based on the review discussion above?

stuartmorgan:

PR #8184, opened 2024-11-26
[file_selector_android] Refactor interactions with `ContentProvider` provided filenames

gmackall:
https://developer.android.com/privacy-and-security/risks/untrustworthy-contentprovider-provided-filename#don%27t-trust-user-input

Adapted from this option
https://developer.android.com/privacy-and-security/risks/untrustworthy-contentprovider-provided-filename#sanitize-provided-filenames

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
> We could alternatively regress to not using a filename that is a function of the original file name

https://github.com/flutter/flutter/issues/64685 had a fair number of votes; I would strongly prefer not to regress it. If we want to go back to generated filenames, I think we would need to do that as a second, longer-term step with a new API that returns a data structure that contains the file as well as other metadata like the original name.

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -130,11 +130,14 @@ public static String getPathFromCopyOfFileFromUri(@NonNull Context context, @Non
         fileName = getBaseName(fileName) + extension;
       }
 
-      File file = new File(targetDirectory, fileName);
+      String filePath = new File(targetDirectory, fileName).getPath();
```
stuartmorgan, line 133: I'm pretty sure you want to sanitize before you pass it to `File`, since once you've attached the `targetDirectory` you are stripping off information we've added during sanitization.

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -206,4 +212,27 @@ private static String getBaseName(String fileName) {
...
+
+  protected static String sanitizeFilename(String displayName) {
+    if (displayName == null) {
+      return null;
+    }
+
+    String[] badCharacters = new String[] { "..", "/" };
+    String[] segments = displayName.split("/");
+    String fileName = segments[segments.length - 1];
+    for (String suspString : badCharacters) {
```
stuartmorgan, line 224: I'm confused as to why the docs do this step since it can't possibly contain any `/`s at this point, but 🤷🏻

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
gmackall: I believe the sanitization should have already happened here, as we get these values earlier:
```java
      String fileName = getFileName(context, uri);
      String extension = getFileExtension(context, uri);
```

And I added the sanitization to each of those methods. Does that make sense?

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
gmackall: Yeah... I'm inclined to leave it because its written by the privacy and security folks at Android but it's definitely odd...

reidbaker: I prefer to leave this code as it exists in the android documentation.

stuartmorgan: Yes, that's fine with me.

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
stuartmorgan: I see; sorry, I misread in the initial readthrough.

gmackall:
`image_picker` pr is also up at https://github.com/flutter/packages/pull/8188, but I'll wait to request review till this lands (the pr is extremely similar).

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -130,11 +135,13 @@ public static String getPathFromCopyOfFileFromUri(@NonNull Context context, @Non
         fileName = getBaseName(fileName) + extension;
       }
 
-      File file = new File(targetDirectory, fileName);
+      String filePath = new File(targetDirectory, fileName).getPath();
+      File outputFile = saferOpenFile(filePath, targetDirectory.getCanonicalPath());
```
reidbaker, line 139: This now throw an exception. Is there more work we need to do to turn that illegal argument exception into something a dart developer can handle?

packages/file_selector/file_selector_android/android/src/test/java/dev/flutter/packages/file_selector_android/FileUtilsTest.java
```diff
@@ -196,6 +197,17 @@ public void getPathFromCopyOfFileFromUri_returnsExpectedPathForUriWithUnknownTyp
...
 
+  @Test
+  public void getPathFromCopyOfFileFromUri_sanitizesPathIndirection() {
+    Uri uri = Uri.parse(MockMaliciousContentProvider.PNG_URI);
+    Robolectric.buildContentProvider(MockMaliciousContentProvider.class).create("dummy");
+    shadowContentResolver.registerInputStream(
+            uri, new ByteArrayInputStream("fileStream".getBytes(UTF_8)));
+    String path = FileUtils.getPathFromCopyOfFileFromUri(context, uri);
+    assertNotNull(path);
+    assertTrue(path.endsWith("_bar.png"));
```
reidbaker, line 208: Can this also assert that there are no `..` in the path?

packages/file_selector/file_selector_android/android/src/test/java/dev/flutter/packages/file_selector_android/FileUtilsTest.java
```diff
@@ -252,4 +264,51 @@ public int update(
       return 0;
     }
   }
+
+  // Mocks a malicious content provider attempting to use path indirection to modify files outside
+  // of the intended directory.
+  // See https://developer.android.com/privacy-and-security/risks/untrustworthy-contentprovider-provided-filename#don%27t-trust-user-input.
+  private static class MockMaliciousContentProvider extends ContentProvider {
```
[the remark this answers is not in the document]
reidbaker: I like this and I don't think it is too verbose given the value. If someone did then I would still prefer this pattern but with the content providers moved to another file.

reidbaker (approved):
approve % comments

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
```diff
@@ -206,4 +216,29 @@ private static String getBaseName(String fileName) {
...
+      fileName = fileName.replace(suspString, "_");
+    }
+    return fileName;
+  }
+
+  // From https://developer.android.com/privacy-and-security/risks/path-traversal#path-traversal-mitigations.
+  protected static @NonNull File saferOpenFile(@NonNull String path, @NonNull String expectedDir) throws IllegalArgumentException, IOException {
+    File f = new File(path);
+    String canonicalPath = f.getCanonicalPath();
+    if (!canonicalPath.startsWith(expectedDir)) {
```
oskar-zeinomahmalat-sonarsource, line 239: In theory, this check would allow a path traversal into a directory that has `expectedDir` as a prefix.
```java
String expectedDir = "/data/data/com.app/cache/123";
String path = "/data/data/com.app/cache/123/../12345/traversal.png";
String canonicalPath = "/data/data/com.app/cache/12345/traversal.png";
canonicalPath.startsWith(expectedDir); // true
```
However, because of the filename sanitization earlier, path traversal sequences are already prevented and this cannot occur. Additionally, the expected directory is an unknown UUID. So, this is not an issue here; I just wanted to mention it for next time. :)

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
reidbaker: When I was primarily an app/plugin dev we had a rule that plugins couldn't throw exceptions that are not handled in dart. Flutter/packages is not as strict but I would like us to investigate what it would take to let the host app know that this error happened. Keep in mind that the original report required a malicious content provider so there could be reasonable app responses to this that help the user not just prevent the bad behavior. Like letting an app developer know that someone is trying to use this attack to hurt their users. Or telling a user that the content provider they selected might be malicious.

gmackall: I've passed the Java exception through to dart and thrown there, and exported the type

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
gmackall: I think this would be an interesting comment to convey to the authors of that method/maintainers of the [page](https://developer.android.com/privacy-and-security/risks/path-traversal), I can try to bring it to their attention. Presumably this would be fixed by making the check instead be
```java
canonicalPath.startsWith(expectedDir + "/")
```
(after optionally stripping a `/` char at the very end so that we don't end up expecting `"some/path/"` + `"/"`)

reidbaker (approved):

packages/file_selector/file_selector_android/android/src/main/java/dev/flutter/packages/file_selector_android/FileUtils.java
oskar-zeinomahmalat-sonarsource: I found a specific GitHub CodeQL docs page about the Partial Path Traversal issue that explains a better mitigation using `Path` objects
https://codeql.github.com/codeql-query-help/java/java-partial-path-traversal-from-remote/

PR #8198, opened 2024-11-28
Why not include product name?

mhamdan91:
Could you guys add product name? 

This is a starter PR, needs a bit of work. lemme know if you want me to finish this off.


stuartmorgan:
> Could you guys add product name?

Please use the issue tracker for feature requests, rather than the PR system.

Closing as this appears to be a high-level proposal, rather than a viable PR, so would be more appropriate for the issue tracker. Once there's agreement on adding a feature, a (complete) PR would be appropriate at that time.

PR #8218, opened 2024-12-03
[in_app_purchase] Update Play Billing library to 7.1.1

mchudy:
Updates Play Billing Library to the latest version 7.1.1. Exposes new APIs as per [release notes](https://developer.android.com/google/play/billing/release-notes):
- Adds Dart representation of `ProductDetails.InstallmentPlanDetails`
- Adds Dart representation of `PendingPurchasesParams` and removes the deprecated `enablePendingPurchases` method on `BillingClientWrapper` (breaking change)
- Adds Dart representation of `Purchase.PendingPurchaseUpdate`
- Removes the deprecated `ProrationMode` as it has been removed from the native library (breaking change)

This PR introduces breaking changes in `in_app_purchase_android`, but does not introduce any breaking changes on the platform interface level.

Fixes https://github.com/flutter/flutter/issues/147394

## Pre-launch Checklist

- [X] I signed the [CLA].


stuartmorgan:
@reidbaker Ping on this review

reidbaker:
@gmackall I am not going to get to this. Deferring to you for review.

wph144:
The billingClient version upgrade process must proceed as quickly as possible. 
Older versions of billingClient cause a low probability of missing payments after a certain period of time. 
This may result in cases where the consumer makes a payment but does not receive the service for that payment, so the billingClient version upgrade must be done quickly. 

Currently, in_app_purchase 3.2.0 is using billingClient 6.2.0,
and billingClient 7.0.0 version was distributed in May 2024, 
so it has now taken more than 7 months. 
I hope this gets merged quickly.

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactory.java
```diff
@@ -24,5 +24,6 @@ interface BillingClientFactory {
   BillingClient createBillingClient(
       @NonNull Context context,
       @NonNull Messages.InAppPurchaseCallbackApi callbackApi,
-      PlatformBillingChoiceMode billingChoiceMode);
+      PlatformBillingChoiceMode billingChoiceMode,
+      Messages.PendingPurchasesParams pendingPurchasesParams);
```
gmackall, line 28: It looks like there is an existing convention of naming our versions of the `BillingClient` classes `Platform$ExsitingClassName` and then defining translation between our version and the original version in [`Translator.java`](https://github.com/flutter/packages/blob/a4e540f7160dc9ea82f9be7d6d42d94d4208a81c/packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/Translator.java).

Is there a reason you adopted this convention for `Messages.PlatformInstallmentPlanDetails` but not for `Messages.PendingPurchasesParams`? If not, can you adopt both of these conventions for this `Messages.PendingPurchasesParams` as well (i.e., rename to `PlatformPendingPurchasesParams`, and then contain all translation code in that one place and reference it in `BillingClientFactoryImpl` below?).

packages/in_app_purchase/in_app_purchase_android/android/src/test/java/io/flutter/plugins/inapppurchase/BillingClientFactoryImplTest.java
```diff
@@ -98,6 +98,17 @@ public void userChoiceBilling() {
     assertTrue(callbackCaptor.getValue().getProducts().isEmpty());
   }
 
+  @Test
+  public void pendingPurchasesForPrepaidPlans() {
+    // No logic to verify just ensure creation works.
```
gmackall, line 103: extreme nit
```suggestion
    // No logic to verify, just ensure creation works.
```

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/purchase_wrapper.dart
```diff
@@ -158,6 +161,58 @@ class PurchaseWrapper {
...
+  /// A PendingPurchaseUpdate is normally generated from a pending transaction
+  /// upgrading/downgrading an existing subscription.
+  /// Returns null if this purchase does not have a pending transaction.
+  final PendingPurchaseUpdateWrapper? pendingPurchaseUpdate;
+}
+
+@JsonSerializable()
+@immutable
+
+/// Represents a pending change/update to the existing purchase.
```
gmackall, line 176: nit: include docs links for classes that are direct wrappers
https://developer.android.com/reference/com/android/billingclient/api/Purchase.PendingPurchaseUpdate

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/subscription_offer_details_wrapper.dart
```diff
@@ -158,3 +165,61 @@ class PricingPhaseWrapper {
         recurrenceMode,
       );
 }
+
+/// Represents additional details of an installment subscription plan.
```
gmackall, line 169: nit: add docs link

gmackall (commented):
Mostly LGTM, added a couple comments. Sorry for the delay on review

packages/in_app_purchase/in_app_purchase/CHANGELOG.md
```diff
@@ -1,204 +1,206 @@
 ## NEXT
 
-* Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
+- Updates minimum supported SDK version to Flutter 3.19/Dart 3.3.
```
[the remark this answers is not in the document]
mchudy: Oh, sorry, yeah that was some autoformatting from VSCode

packages/in_app_purchase/in_app_purchase_android/android/src/main/java/io/flutter/plugins/inapppurchase/BillingClientFactory.java
mchudy: Yeah, no reason for that, I missed that for some reason. It's fixed now.

mchudy:
CI is complaining about versioning, but I believe we should first publish the new version of in_app_purchase_android and only once that lands on pub update the dependency in the base package? Or can this be handled inside a single PR?

stuartmorgan:
> CI is complaining about versioning, but I believe we should first publish the new version of in_app_purchase_android and only once that lands on pub update the dependency in the base package? Or can this be handled inside a single PR?

See https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins for the process of landing a multi-package PR with interdependencies.

packages/in_app_purchase/in_app_purchase_android/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.4.0
+
+* Updates Google Play Billing Library from 6.2.0 to 7.1.1.
+* **BREAKING CHANGES**:
+  * Removes the deprecated `ProrationMode` enum. `ReplacementMode` should be used instead.
+  * Removes the deprecated `BillingClientWrapper.enablePendingPurchases` method.
+  * Bumps `minSdk` to 21.
```
stuartmorgan, line 7: This would be a breaking change for `in_app_purchase` when rolling the dependency there; we strongly prefer not dropping OS support before the SDK itself.

In this case, Flutter 3.24+ only supports Android SDK 21+, so the way to make this change would be to update the plugin's min SDK version to 3.24, which makes this no longer a breaking change.

gmackall (approved):
LGTM, @stuartmorgan or @bparrishMines would either of you like to do the secondary review?

packages/in_app_purchase/in_app_purchase_android/lib/src/billing_client_wrappers/pending_purchases_params_wrapper.dart
```diff
@@ -0,0 +1,53 @@
...
+@JsonSerializable()
+@immutable
+class PendingPurchasesParamsWrapper {
+  /// Creates a [PendingPurchasesParamsWrapper].
+  const PendingPurchasesParamsWrapper({
+    required this.enablePrepaidPlans,
+  });
+
+  /// Factory for creating a [PendingPurchasesParamsWrapper] from a [Map].
+  @Deprecated('JSON serialization is not intended for public use, and will '
```
stuartmorgan, line 25: Why are we adding new code that is pre-deprecated?

The reason other classes had JSON serialization is that they were being serialized as JSON to send over the method channel. The reasons it's deprecated in all those classes is that we don't do that any more, because we use Pigeon. It was only left in place was to avoid a breaking change on the off-chance someone was using it for other reasons, since it was part of the public interface.

packages/in_app_purchase/in_app_purchase_android/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 0.4.0
+
+* Updates Google Play Billing Library from 6.2.0 to 7.1.1.
+* **BREAKING CHANGES**:
```
stuartmorgan, line 4: Since we are doing a breaking change, this seems like a good opportunity to rip out all the pre-Pigeon `JsonSerializable` code that was deprecated previously. @gmackall is there any reason not to include that here?

mchudy: Sounds like a good idea, I have one question though. E.g. `BillingResponseConverter` was used to map from raw ints (Android defined constants) to public-facing `BillingResponse` enum. Naive approach would be to introduce a new function in `pigeon_converters.dart`:

```dart
BillingResponse billingResponseFromPlatformResponseCode(int responseCode) {
  return switch (responseCode) {
    -3 => BillingResponse.serviceTimeout,
    -2 => BillingResponse.featureNotSupported,
    -1 => BillingResponse.serviceDisconnected,
    0 => BillingResponse.ok,
    1 => BillingResponse.userCanceled,
    2 => BillingResponse.serviceUnavailable,
    3 => BillingResponse.billingUnavailable,
    4 => BillingResponse.itemUnavailable,
    5 => BillingResponse.developerError,
    6 => BillingResponse.error,
    7 => BillingResponse.itemAlreadyOwned,
    8 => BillingResponse.itemNotOwned,
    12 => BillingResponse.networkError,
    _ => BillingResponse.error,
  };
}
```

That looks a bit sketchy to me though as we have to deal with those hardcoded constants in Dart. Should we perhaps change the definition of `PlatformBillingResult` in Pigeon contract and introduce a new `PlatformBillingResponse` enum as part of this refactoring? If the mapping is performed on Java side we can refer to constants defined in the library so the code is more maintainable. Correct me if I'm wrong please.

stuartmorgan: > Should we perhaps change the definition of `PlatformBillingResult` in Pigeon contract and introduce a new `PlatformBillingResponse` enum as part of this refactoring? If the mapping is performed on Java side we can refer to constants defined in the library so the code is more maintainable.

Yes, that's definitely what we want to do. The fact that we are still using `BillingResponseConverter().fromJson` in `pigeon_converters.dart` was a mistake on my part; I did an initial shallow conversion from manual method channels to Pigeon and then went back and converted objects more deeply to Pigeon structures in follow-ups, but missed this. We strongly prefer the clarity and safety of doing a Java constant -> Pigeon enum conversion in Java, then Pigeon enum -> Dart enum conversion in Dart, even though doing so is more verbose.

mchudy: I refactored this and introduced new enum in Pigeon contracts. I also noticed that the representation of https://developer.android.com/reference/com/android/billingclient/api/BillingClient.FeatureType was very outdated so I made it match the current API. As it removes two variants it's another breaking changes, but I think it would be good to include this with those other breaking changes.

reidbaker (approved):
Hold on merging for @stuartmorgan's approval. Specifically if all of the json removal work he wants removed has been done.

stuartmorgan (commented):
Awesome, thanks for all of this cleanup!

It looks like `bulid.yaml` can be removed, as well as the `json_annotation` and `json_serializable` dependencies in `pubspec.yaml`, but that's all I see that's left.

stuartmorgan (changes requested):
> Hold on merging for @stuartmorgan's approval.

This PR can't actually be landed as-is; it's saying that it bumps the `in_app_purchase_android` version in `in_app_purchase`, but doesn't (and can't in this PR).

I'm going to put a request-changes flag on the PR just to make sure nobody accidentally lands it as is. This is ready to be split into two PRs though (with the minor changes in my previous comment) for landing.

mchudy:
I removed the JSON dependencies and moved the `in_app_purchase` package part to https://github.com/flutter/packages/pull/8463

stuartmorgan (approved):
LGTM!

PR #8229, opened 2024-12-04
[shared_preferences] Tool for migrating from legacy shared_preferences to shared_preferences_async

tarrinneal:
fixes https://github.com/flutter/flutter/issues/150732
fixes https://github.com/flutter/flutter/issues/123153


tarrinneal:
@stuartmorgan this is ready for review now, though it will be waiting for https://github.com/flutter/packages/pull/8296 to land

packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_migration_tool_test.dart
```diff
@@ -0,0 +1,232 @@
...
+                Platform.isLinux ||
+                Platform.isMacOS ||
+                Platform.isIOS),
+      );
+    }
+
+    void runAllGroups(
+        {String? stringValue = testString, bool keysCollide = false}) {
+      setUp(() async {
+        await preferences.setBool(boolKey, testBool);
```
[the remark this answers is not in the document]
stuartmorgan: What do you think of [this version](https://github.com/stuartmorgan/packages/blob/migration-alt-tests/packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_migration_tool_test.dart)?

packages/shared_preferences/shared_preferences/lib/tools/legacy_to_async_migration_tool.dart
```diff
@@ -0,0 +1,67 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:shared_preferences_platform_interface/types.dart';
+
+import '../shared_preferences.dart';
+
+/// A tool to migrate from the legacy SharedPreferences system to
```
stuartmorgan, line 9: Similarly, I would say "Migrates preferences from [...]" so there's no potential confusion that this is a stand-alone tool that has to be packaged and run.

packages/shared_preferences/shared_preferences/lib/tools/legacy_to_async_migration_tool.dart
```diff
@@ -0,0 +1,67 @@
...
+/// [legacySharedPreferencesInstance] should be an instance of [SharedPreferences]
+/// that has been instantiated the same way it has been used throughout your app.
+/// If you have called [SharedPreferences.setPrefix] that must be done before using
+/// this tool.
+///
+/// [sharedPreferencesAsyncOptions] should be an instance of [SharedPreferencesOptions]
+/// that is set up the way you intend to use the new system going forward.
+/// This tool will allow for future use of [SharedPreferencesAsync] and [SharedPreferencesWithCache].
+///
+/// The [migrationCompletedKey] is a key that will be used to check if the migration
```
stuartmorgan, line 21: I think this should be explicit about the fact that the key is stored in the target preference system, not the legacy preference system.

packages/shared_preferences/shared_preferences/lib/tools/legacy_to_async_migration_tool.dart
```diff
@@ -0,0 +1,67 @@
...
+/// has run before, to avoid overwriting new data going forward. Make sure that
+/// there will not be any collisions with preferences you are or will be setting
+/// going forward, or there may be data loss.
+Future<void> migrateLegacySharedPreferencesToSharedPreferencesAsync(
+  SharedPreferences legacySharedPreferencesInstance,
+  SharedPreferencesOptions sharedPreferencesAsyncOptions,
+  String migrationCompletedKey, {
+  bool clearLegacyPreferences = false,
```
stuartmorgan, line 29: This appears to not do anything?

packages/shared_preferences/shared_preferences/lib/tools/legacy_to_async_migration_tool.dart
```diff
@@ -0,0 +1,67 @@
...
+      case const (String):
+        await sharedPreferencesAsyncInstance.setString(key, value! as String);
+      case const (List<String>):
+      case const (List<String?>):
+      case const (List<Object?>):
+      case const (List<dynamic>):
+        try {
+          await sharedPreferencesAsyncInstance.setStringList(
+              key, (value! as List<Object?>).cast<String>());
+        } catch (_) {} // Pass over Lists containing non-String values.
```
stuartmorgan, line 60: Do you want to catch *all* exceptions here, or only cast exceptions? The comment implies the former, but it does the latter which could silently swallow other problems.

packages/shared_preferences/shared_preferences/lib/tools/legacy_to_async_migration_tool.dart
```diff
@@ -0,0 +1,67 @@
...
+///
+/// [sharedPreferencesAsyncOptions] should be an instance of [SharedPreferencesOptions]
+/// that is set up the way you intend to use the new system going forward.
+/// This tool will allow for future use of [SharedPreferencesAsync] and [SharedPreferencesWithCache].
+///
+/// The [migrationCompletedKey] is a key that will be used to check if the migration
+/// has run before, to avoid overwriting new data going forward. Make sure that
+/// there will not be any collisions with preferences you are or will be setting
+/// going forward, or there may be data loss.
+Future<void> migrateLegacySharedPreferencesToSharedPreferencesAsync(
```
stuartmorgan, line 25: This should probably be called `migrateLegacySharedPreferencesToSharedPreferencesAsyncIfNecessary`. And the docs should be explicit that it's safe to call on every launch as it will do nothing if migration has already happened.

packages/shared_preferences/shared_preferences/lib/tools/legacy_to_async_migration_tool.dart
```diff
@@ -0,0 +1,67 @@
...
+  final SharedPreferencesAsync sharedPreferencesAsyncInstance =
+      SharedPreferencesAsync(options: sharedPreferencesAsyncOptions);
+
+  if (await sharedPreferencesAsyncInstance.containsKey(migrationCompletedKey)) {
+    return;
+  }
+
+  Set<String> keys = legacySharedPreferencesInstance.getKeys();
+  await legacySharedPreferencesInstance.reload();
+  keys = legacySharedPreferencesInstance.getKeys();
```
stuartmorgan, line 40: Why are the keys being set twice?

stuartmorgan (changes requested):
The README should be updated to explain this, in place of the discussion about manual migration.

packages/shared_preferences/shared_preferences/lib/tools/legacy_to_async_migration_tool.dart
tarrinneal: forgot about that. Is this a feature that's worth adding? I hadn't yet as it seemed like it could be problematic.

stuartmorgan: We could wait and see if anyone asks. On one hand, I would generally expect a migration to move data instead of duplicating it. On the other hand, it shouldn't be a lot of data.

My one real hesitation here is that if someone isn't using an allow list, on most platforms this would mean that an unprefixed new-style prefs object would be loading all the old prefs too.

packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_migration_tool_test.dart
```diff
@@ -0,0 +1,232 @@
...
+  const int testInt = 42;
+  const double testDouble = 3.14159;
+  const List<String> testList = <String>['foo', 'bar'];
+
+  group('shared_preferences', () {
+    late SharedPreferences preferences;
+    late SharedPreferencesOptions sharedPreferencesAsyncOptions;
+    const String migrationCompletedKey = 'migrationCompleted';
+
+    void runTests(
```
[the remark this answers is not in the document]
tarrinneal: Better to duplicate the tests then? I just wanted to ensure equivalent coverage without unnecessary duplication. I can see how it's a bit convoluted though.

packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_migration_tool_test.dart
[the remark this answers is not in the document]
tarrinneal: This makes sense, I guess I just don't like calling code before it's defined. This does seem clearer.

packages/shared_preferences/shared_preferences/example/integration_test/shared_preferences_migration_tool_test.dart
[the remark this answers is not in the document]
stuartmorgan: The version I proposed below makes the logic more directly traceable; it seemed like a reasonable compromise between trying to avoid spooky action at a distance (e.g., nested layers of `setUp` scattered through the file) without adding duplication.

packages/shared_preferences/shared_preferences/lib/tools/legacy_to_async_migration_tool.dart
tarrinneal: It also means that items not set by our plugin would get deleted. Potentially if not used carefully (or written carefully) it could write to the same file, then delete the prefs it just wrote.

stuartmorgan (changes requested):
`<seinfeldmeme>`You thumbs-up'd my comment about updating the README, but you didn't update the README`</seinfeldmeme>` 🙂

tarrinneal:
> `<seinfeldmeme>`You thumbs-up'd my comment about updating the README, but you didn't update the README`</seinfeldmeme>` 🙂

I use the resolve button to track work I have or haven't done yet, since that comment isn't resolvable, I missed it.

tarrinneal:
> `<seinfeldmeme>`You thumbs-up'd my comment about updating the README, but you didn't update the README`</seinfeldmeme>` 🙂

RESOLVED

packages/shared_preferences/shared_preferences/example/lib/main.dart
```diff
@@ -61,14 +65,25 @@ class SharedPreferencesDemoState extends State<SharedPreferencesDemo> {
     });
   }
 
+  Future<void> _migratePreferences() async {
+    // #docregion migrate
+    const SharedPreferencesOptions sharedPreferencesOptions =
+        SharedPreferencesOptions();
+    final SharedPreferences prefs = await SharedPreferences.getInstance();
+    await migrateLegacySharedPreferencesToSharedPreferencesAsyncIfNecessary(
+        prefs, sharedPreferencesOptions, 'migrationCompleted');
```
stuartmorgan, line 74: Since this is a README snippet, where there's no context about the arguments, let's make a
`const String migrationCompletedPrefKey = 'migrationCompleted'` as the previous line, and then use that variable here, so it's self-documenting in the README.

packages/shared_preferences/shared_preferences/lib/util/legacy_to_async_migration_util.dart
```diff
@@ -0,0 +1,69 @@
...
+import 'package:shared_preferences_platform_interface/types.dart';
+
+import '../shared_preferences.dart';
+
+/// Migrates preferences from the legacy SharedPreferences system to
+/// SharedPreferencesAsync.
+///
+/// This method can be run multiple times without worry of overwriting transferred data,
+/// as long as migrationCompletedKey is the same each time, and the value stored
+/// under migrationCompletedKey in the target preferences system is not modified.
```
stuartmorgan, line 14: migrationCompletedKey should have brackets in both instances.

stuartmorgan (changes requested):

stuartmorgan (approved):
LGTM once you re-run snippet generation.

tarrinneal:
> LGTM once you re-run snippet generation.

I'm waiting for the revert pr to land before I handle conflicts and merge this pr.

PR #4318, opened 2023-06-26
[video_player] added iOS exception on incorrect asset path

zopagaduanjr:
surfaced the incorrect asset path error to dart using try catch.

Fixes [#118660](https://github.com/flutter/flutter/issues/118660)

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#version-and-changelog-updates
[following repository CHANGELOG style]: https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests

PR #8237, opened 2024-12-06
[video_player_avfoundation] iOS platform view support

FirentisTFW:
This PR adds support for platform views on iOS as a way of displaying a video. When creating a video, it's now possible to choose between texture view approach (rendered using `Texture` widget on the Flutter side) and platform view approach (rendered on the native side, using `AVPlayerLayer`).

`FVPVideoPlayer` class now has nothing to do with texture. The texture-related code was moved from it to `FVPTextureBasedVideoPlayer` - a subclass of `FVPVideoPlayer` that adds texture functionality. In the plugin class (`createWithOptions` method) we create either the basic version (for platform view) or the texture subclass (in case of texture approach) based on the parameter passed in from Flutter side.

Platform view is only supported on iOS, no MacOS implementation is added in this PR. The functionality is not yet exposed in the app-facing package (only in example app) - it will be done later, once we add the Android implementation. The PR does not introduce breaking changes, I followed the rule "non-breaking changes, even at the expense of a less-clean API" (`buildViewWithOptions` and `createWithOptions` methods).

Up to this point, the variable naming relied heavily on the texture (we had a lot of `textureId` variables and properties). Since now you can use a platform view instead of a texture view, these variables and parameters were renamed to just `playerId`.

I left some comments in the PR to clarify/discuss some choices.

Resolves [#86613](https://github.com/flutter/flutter/issues/86613) (not fully though, as it is not yet available in the app-facing package).

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -31,6 +39,10 @@ - (FVPDisplayLink *)displayLinkWithRegistrar:(id<FlutterPluginRegistrar>)registr
 
 #pragma mark -
 
+/// The next non-texture player ID, initialized to a high number to avoid collisions with
+/// texture IDs (which are generated separately).
+static int64_t nextNonTexturePlayerId = 1000000;
```
FirentisTFW, line 44: I'm not sure if this is the best approach, but looks like texture ids start from 0 and then get incremented, so this starting point seems to be pretty safe. Alternatively, I assume we could generate random ids for non-texture players

FirentisTFW:
One PR check is failing, but this is due to changes in platform interface and local dependency overrides. The code can be reviewed already. Once this PR is reviewed, I'll create another one with platform interface changes only (according to `contributing/README.md`).

FirentisTFW:
I see two integration tests failing, but I assume this is due to tree-status check failure, since I can see the same tests failing there.

Additionally, I think `testSeekToWhilePausedStartsDisplayLinkTemporarily` (iOS native test) might be flaky. Sometimes it fails (like [here](https://ci.chromium.org/ui/p/flutter/builders/try/Mac_arm64%20macos_platform_tests%20master%20-%20packages/18813/overview)), but when I run the pipeline again (with no additional changes), it passes (or the other way around, first it passes, then it fails). Sometimes it also fails locally for me, but only once every 20-25 test runs (and I could not find a good way to reproduce it).

Anyway, I merged `main` branch, resolved conflicts, and I think the PR can be reviewed. These failing tests hopefully should not be there when the tree is fixed.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
tarrinneal: could use `INFINITY`

tarrinneal:
I re-ran the tests for you.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: And you'd suggest to decrement it? Or start with `-INFINITY` and then increment?

tarrinneal: didn't read enough to realize you were incrementing. If that is the case, bigint would make sense for starting. I don't know if we have a standard practice for this though, so maybe @stuartmorgan would have a better solution.

FirentisTFW:
Looks like one native test (`testAudioOnlyHLSControls` which was added 3 weeks ago) [was initially failing](https://ci.chromium.org/ui/p/flutter/builders/try/Mac_arm64%20macos_platform_tests%20master%20-%20packages/18926/overview) on MacOS, but it passes now. I could not reproduce the test failure locally, it was always passing, and could not see any issue in the code. So I added some logs to see why it was failing on the CI, but then it passed. I re-ran the tests, and everything passed again. So I removed the logs and force-pushed the original code. And it [passed again](https://ci.chromium.org/ui/p/flutter/builders/try/Mac_arm64%20macos_platform_tests%20master%20-%20packages/19048/overview). Since the issue is not happening anymore, I couldn’t reproduce it, and I already spent almost two days trying to find the cause of it, I’ll publish the PR. Maybe the issue was somehow related to the CI. If it comes back again after I push CR fixes, I’ll have another look at it. Anyway, the PR's ready for review now.

stuartmorgan:
That's one of our tests that loads an actual network resource, so it's potentially subject to network-related flake (either at the CI end or the server end).

stuartmorgan:
> Additionally, I think `testSeekToWhilePausedStartsDisplayLinkTemporarily` (iOS native test) might be flaky.

I've seen this flake at least once before as well.

packages/video_player/video_player/example/pubspec.yaml
```diff
@@ -35,3 +35,15 @@ flutter:
     - assets/bumble_bee_captions.srt
     - assets/bumble_bee_captions.vtt
     - assets/Audio.mp3
+
+# FOR TESTING AND INITIAL REVIEW ONLY. DO NOT MERGE.
```
stuartmorgan, line 39: You can go ahead and revert these tool-generated changes in the packages where nothing else is changing (`video_player`, `video_player_android`, `video_player_web`) now. The tool doesn't know what's going to be changed so it just overrides all references everywhere.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
stuartmorgan: Longer term we should fully decouple the ID vended out from texture IDs, and then collisions won't matter; for now to reduce scope, picking something suitably large like `INT_MAX` and decrementing is probably easiest.

packages/video_player/video_player_avfoundation/pigeons/messages.dart
```diff
@@ -17,13 +17,30 @@ import 'package:pigeon/pigeon.dart';
...
+}
+
 class CreationOptions {
   CreationOptions({required this.httpHeaders});
   String? asset;
   String? uri;
   String? packageName;
   String? formatHint;
   Map<String, String> httpHeaders;
+  PlatformVideoViewType? viewType;
```
stuartmorgan, line 43: In general, something should be nullable only if a clear meaning of `null` can be articulated that is distinct from other meanings. Looking at the native code, it appears that `null` is currently just a second, less intuitive way of writing `PlatformVideoViewType.textureView`, so unless I'm missing something this should be non-nullable at this layer.

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -163,6 +173,9 @@ class DataSource {
   /// The package that the asset was loaded from. Only set for
   /// [DataSourceType.asset] videos.
   final String? package;
+
+  /// The type of view to be used for displaying the video player.
+  final VideoViewType viewType;
```
stuartmorgan, line 178: Why is this part of `DataSource`? How it's displayed isn't related to the video source.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayerTextureApproach.h
```diff
@@ -0,0 +1,37 @@
...
+#import "FVPDisplayLink.h"
+#import "FVPFrameUpdater.h"
+#import "FVPVideoPlayer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// A subclass of FVPVideoPlayer that adds functionality related to texture-based view as a way of
+/// displaying the video in the app. It manages the CALayer associated with the Flutter view,
+/// updates frames, and handles display link callbacks.
+@interface FVPVideoPlayerTextureApproach : FVPVideoPlayer <FlutterTexture>
```
stuartmorgan, line 14: It's not obvious at call sites that this is a player; how about `FVPTextureBasedVideoPlayer`?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoView.h
```diff
@@ -0,0 +1,32 @@
...
+#endif
+
+#import <AVFoundation/AVFoundation.h>
+
+/// A class used to create a native video view that can be embedded in a Flutter app.
+/// This class wraps an AVPlayer instance and displays its video content.
+#if TARGET_OS_IOS
+@interface FVPNativeVideoView : NSObject <FlutterPlatformView>
+#else
+@interface FVPNativeVideoView : NSWindow
```
stuartmorgan, line 22: Why `NSWindow`? That's a very surprising class for a view to inherit from.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoView.h
```diff
@@ -0,0 +1,32 @@
...
+#endif
+/// Initializes a new instance of a native view.
+/// It creates a video view instance and sets the provided AVPlayer instance to it.
+- (instancetype)initWithPlayer:(AVPlayer *)player;
+
+#if TARGET_OS_IOS
+/// Returns the native UIView that displays the video content.
+- (UIView *)view;
```
stuartmorgan, line 30: Methods from implemented protocols are not generally redeclared in Obj-C.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -62,32 +81,44 @@ - (instancetype)initWithAVFactory:(id<FVPAVFactory>)avFactory
...
-  int64_t textureId = [self.registry registerTexture:player];
-  frameUpdater.textureId = textureId;
+  BOOL usesTextureApproach =
+      frameUpdater != nil && [player isKindOfClass:[FVPVideoPlayerTextureApproach class]];
+  int64_t playerId;
+  if (usesTextureApproach) {
+    playerId = [self.registry registerTexture:(FVPVideoPlayerTextureApproach *)player];
+    frameUpdater.textureId = playerId;
+  } else {
+    @synchronized(self) {
```
stuartmorgan, line 102: Why is this `@synchronized`? What other threads are using the ID?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation_ios/FVPNativeVideoView.m
```diff
@@ -0,0 +1,43 @@
...
+  return self;
+}
+
+- (FVPPlayerView *)view {
+  return _view;
+}
+
+- (void)dealloc {
+  [_view setPlayer:nil];
+  _view = nil;
```
stuartmorgan, line 40: Why is an ivar being nil`d in dealloc?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation_ios/FVPNativeVideoViewFactory.m
```diff
@@ -0,0 +1,36 @@
...
+#import "../video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoView.h"
+#import "../video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer.h"
+#import "../video_player_avfoundation/include/video_player_avfoundation/messages.g.h"
+
+@implementation FVPNativeVideoViewFactory {
+  NSObject<FlutterBinaryMessenger> *_messenger;
+  NSMutableDictionary<NSNumber *, FVPVideoPlayer *> *_playersById;
+}
+- (instancetype)initWithMessenger:(NSObject<FlutterBinaryMessenger> *)messenger
+                      playersById:(NSMutableDictionary<NSNumber *, FVPVideoPlayer *> *)playersById {
```
stuartmorgan, line 16: It would be cleaner to make this a block that returns a player for an ID, rather than passing in a mutable dictionary and creating the potential for two different classes to be modifying it.

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -139,14 +141,45 @@ class AVFoundationVideoPlayer extends VideoPlayerPlatform {
...
+  }
+
   @override
   Widget buildView(int textureId) {
-    return Texture(textureId: textureId);
+    return buildViewWithOptions(
+      VideoViewOptions(
+        playerId: textureId,
+        // Texture view was the only supported view type before
+        // buildViewWithOptions was introduced. We pass it here to maintain
```
stuartmorgan, line 155: I strongly prefer we not put "we" in code comments since it's often ambiguous who "we" is. In this case, I think the second sentence can just be removed entirely.

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
FirentisTFW: The explanation is in the PR description, I did it to avoid breaking changes.
> The PR does not introduce breaking changes, I followed the rule "non-breaking changes, even at the expense of a less-clean API" (buildViewWithOptions method, viewType parameter in DataSource).

Would you suggest a different approach here?

packages/video_player/video_player_avfoundation/pigeons/messages.dart
FirentisTFW: Yes, it should not be nullable, good catch

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayerTextureApproach.h
FirentisTFW: Good point, renamed to `FVPTextureBasedVideoPlayer`

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoView.h
FirentisTFW: Good catch, should be `NSView`

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoView.h
FirentisTFW: Thanks, I took it from [the documentation](https://docs.flutter.dev/platform-integration/ios/platform-views), removed now

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation_ios/FVPNativeVideoViewFactory.m
FirentisTFW: Good point, I modified the code like suggested

FirentisTFW:
> I've seen this flake at least once before as well.

I replaced a regular assertion with an expectation with a timeout, hope that is okay and fixes the problem.

https://github.com/flutter/packages/pull/8237/commits/9b7f9ad66ea3161970ee6db1c72e5ec4f8c6e70b

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: I used `INT_MAX`.

Should I add a todo for decoupling platform view IDs form texture based IDs? Who I should mention in the todo then? And will we need an issue for it?

FirentisTFW: Since you say it's a relatively minor thing, should it be done in this PR then? How do you see this decoupling?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: Thanks for the explanation, removed `@synchronized` then

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
FirentisTFW: Okay, I added `createWithOptions` then and left `DataSource` unchanged

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
stuartmorgan: I mean minor in the sense that I don't think it's particularly important, since I think this approach should work fine in practice even though the structure is a bit odd for maintainers to understand.

In code, I would expect that we would make the instance identifiers that are used at the platform interface API layer fully abstract, and track texture IDs separately internally for texture-based instances.

ditman:
I don't own these packages so removing myself, but this seems like a good entrypoint for an incoming web PR adding a "texture"-like rendering for the video player there. [PR](https://github.com/flutter/packages/pull/6286) (it got closed though, so I'm not sure if it'll ever get updated to leverage this).

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -476,3 +501,35 @@ class VideoPlayerWebOptionsControls {
...
+}
+
+/// [VideoCreationOptions] contains creation options for a video player.
+@immutable
+class VideoCreationOptions {
+  /// Constructs an instance of [VideoCreationOptions].
+  const VideoCreationOptions({
+    required this.dataSource,
+    required this.viewType,
```
FirentisTFW, line 527: @stuartmorgan Do you think we should put `VideoViewType.textureView` as a default here?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: Okay, I added a todo for that

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -31,6 +39,11 @@ - (FVPDisplayLink *)displayLinkWithRegistrar:(id<FlutterPluginRegistrar>)registr
 
 #pragma mark -
 
+// TODO(stuartmorgan): Decouple IDs for platform views and texture views.
+/// The next non-texture player ID, initialized to a high number to avoid collisions with
+/// texture IDs (which are generated separately).
+static int64_t nextNonTexturePlayerId = INT_MAX;
```
stuartmorgan, line 45: This can just be an ivar; we don't need IDs to be unique across engine instances.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -62,32 +84,42 @@ - (instancetype)initWithAVFactory:(id<FVPAVFactory>)avFactory
...
+  [self.playersById.allValues makeObjectsPerformSelector:@selector(disposeSansEventChannel)];
+  [self.playersById removeAllObjects];
   SetUpFVPAVFoundationVideoPlayerApi(registrar.messenger, nil);
 }
 
 - (int64_t)onPlayerSetup:(FVPVideoPlayer *)player frameUpdater:(FVPFrameUpdater *)frameUpdater {
-  int64_t textureId = [self.registry registerTexture:player];
-  frameUpdater.textureId = textureId;
+  BOOL textureBased =
+      frameUpdater != nil && [player isKindOfClass:[FVPTextureBasedVideoPlayer class]];
```
stuartmorgan, line 99: Why is `frameUpdater != nil` part of this check? If someone passed in a texture-based player without a frame updater (which would be a bug), it would still be texture-based.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation_ios/FVPNativeVideoViewFactory.m
```diff
@@ -0,0 +1,38 @@
...
+  NSNumber *playerId = @(args.playerId);
+  FVPVideoPlayer *player = _playerByIdProvider(playerId);
+  return [[FVPNativeVideoView alloc] initWithPlayer:player.player];
+}
+
+- (NSObject<FlutterMessageCodec> *)createArgsCodec {
+  return FVPGetMessagesCodec();
+}
+
+@end
```
stuartmorgan, line 38: Nit: please add the missing newline.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation_ios/FVPNativeVideoViewFactory.m
```diff
@@ -0,0 +1,38 @@
...
+
+#import "../video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoViewFactory.h"
+
+#import "../video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoView.h"
+#import "../video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer.h"
+#import "../video_player_avfoundation/include/video_player_avfoundation/messages.g.h"
+
+@implementation FVPNativeVideoViewFactory {
+  NSObject<FlutterBinaryMessenger> *_messenger;
+  FVPVideoPlayer * (^_playerByIdProvider)(NSNumber *);
```
stuartmorgan, line 13: The preferred repository style for new code is to use private properties and use `self.` access outside of init/dealloc.

stuartmorgan (changes requested):

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
```diff
@@ -0,0 +1,180 @@
...
+                  registrar:(NSObject<FlutterPluginRegistrar> *)registrar {
+  NSDictionary<NSString *, id> *options = nil;
+  if ([headers count] != 0) {
+    options = @{@"AVURLAssetHTTPHeaderFieldsKey" : headers};
+  }
+  AVURLAsset *urlAsset = [AVURLAsset URLAssetWithURL:url options:options];
+  AVPlayerItem *item = [AVPlayerItem playerItemWithAsset:urlAsset];
+  return [self initWithPlayerItem:item
+                     frameUpdater:(FVPFrameUpdater *)frameUpdater
+                      displayLink:(FVPDisplayLink *)displayLink
```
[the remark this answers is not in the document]
FirentisTFW: Looks like displayLink cast was actually there before. I removed them, thanks

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
```diff
@@ -0,0 +1,180 @@
...
+}
+
+- (void)updatePlayingState {
+  [super updatePlayingState];
+  // If the texture is still waiting for an expected frame, the display link needs to keep
+  // running until it arrives regardless of the play/pause state.
+  _displayLink.running = self.isPlaying || self.waitingForFrame;
+}
+
+- (void)seekTo:(int64_t)location completionHandler:(void (^)(BOOL))completionHandler {
```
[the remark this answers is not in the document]
FirentisTFW: Sure, I had a fixme comment there to clean it up, but then looks like I just removed it

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
```diff
@@ -0,0 +1,180 @@
...
+        }];
+}
+
+- (void)expectFrame {
+  self.waitingForFrame = YES;
+
+  _displayLink.running = YES;
+}
+
+- (CVPixelBufferRef)copyPixelBuffer {
```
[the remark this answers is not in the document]
FirentisTFW: Added pragma sections for `FlutterTexture`, `Overrides` and `Private methods`

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: Right, changed to ivar

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: You're right, removed this part

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -128,22 +87,12 @@ - (instancetype)initWithPlayerItem:(AVPlayerItem *)item
   _player = [avFactory playerWithPlayerItem:item];
   _player.actionAtItemEnd = AVPlayerActionAtItemEndNone;
 
-  // This is to fix 2 bugs: 1. blank video for encrypted video streams on iOS 16
-  // (https://github.com/flutter/flutter/issues/111457) and 2. swapped width and height for some
-  // video streams (not just iOS 16).  (https://github.com/flutter/flutter/issues/109116). An
-  // invisible AVPlayerLayer is used to overwrite the protection of pixel buffers in those streams
-  // for issue #1, and restore the correct width and height for issue #2.
-  _playerLayer = [AVPlayerLayer playerLayerWithPlayer:_player];
-  [self.flutterViewLayer addSublayer:_playerLayer];
```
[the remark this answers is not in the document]
FirentisTFW: Yes, I checked that on iOS 16 and it looks good. When I comment these two lines out, the aspect ration is broken again, so I assume the fix is still working, at least for the second issue. The first one (blank screen) I couldn't reproduce anymore (on iOS 16), I guess it's working fine, we also have a unit test for it.

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -48,11 +48,19 @@ abstract class VideoPlayerPlatform extends PlatformInterface {
...
-  /// Creates an instance of a video player and returns its textureId.
+  /// Creates an instance of a video player and returns its playerId.
+  @Deprecated('Use createWithOptions() instead.')
   Future<int?> create(DataSource dataSource) {
     throw UnimplementedError('create() has not been implemented.');
   }
 
+  /// Creates an instance of a video player based on creation options
+  /// and returns its playerId. If view type is [VideoViewType.textureView],
+  /// the playerId is also the texture id.
```
stuartmorgan, line 59: Looking at just the platform interface part again, I don't think we should encode this into the interface, as it will make it harder for the higher-level code to untangle the now-ideally-abstract player ID from the texture ID, which has a very specific meaning.

There's also a design issue here where both `createWithOptions` and `buildViewWithOptions` are being passed a view type, but once `createWithOptions` has been called, there is only one valid type that can be passed to `buildViewWithOptions` for that player ID, so the calling code is being forced to maintain view type state.

Couldn't `VideoViewOptions` only contain the player ID for now, and the state of how to map player ID to implementation details like the texture ID (if any) be maintained in the platform implementation, which is already stateful?

packages/video_player/video_player_avfoundation/test/avfoundation_video_player_test.dart
```diff
@@ -188,6 +188,126 @@ void main() {
...
+              asset: '/path/to/incorrect_asset',
+            ),
+            viewType: VideoViewType.textureView,
+          ),
+        );
+        fail('should throw PlatformException');
+      } catch (e) {
+        expect(e, isException);
+      }
+    });
```
FirentisTFW, line 223: I copied this test from the one above (for `create` method) which was added in [this PR](https://github.com/flutter/packages/pull/4318), but I don't think it checks anything.

I assume the call to `createWithOptions` (or `create` in the original test) should fail, then we should catch it, and verify the exception. But looks like what actually happens is that the `fail` call throws, we just catch it and verify that it is exception (which it is because it's a `TestFailure`).

Do we need this test? What value does it bring?

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
FirentisTFW: I did as suggested. Now the responsibility of maintaining view type state for a given player relies on the platform implementation.

The player ID as viewed by the app facing package should now more abstract, but currently we ended up in a very ugly state where most of the methods still use `textureId` for parameter names (which now should be renamed to `playerId`). I suppose that we could fix it all up when creating platform interface PR?

In case of texture view, the player ID will still be the same as texture ID, we would just not explicitly reveal that. If we wanted to change this and make the player ID fully decoupled from the texture ID, that would probably require some significant changes - either moving the player ID generation process to the Dart side or changing the pigeon interface so that it returns player ID as well as texture ID (which can be null).

packages/video_player/video_player_avfoundation/test/avfoundation_video_player_test.dart
stuartmorgan: > But looks like what actually happens is that the `fail` call throws, we just catch it and verify that it is exception (which it is because it's a `TestFailure`).

Sounds like the test is broken; the preferred way to write it would be with a `throwsA` matcher on the call, instead of a manual try/catch.

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
stuartmorgan: > In case of texture view, the player ID will still be the same as texture ID, we would just not explicitly reveal that.

That's fine for now as an implementation detail, I just don't want that to become something we are stuck with indefinitely because it's part of the API contract of the new methods.

stuartmorgan (commented):
I didn't fully re-review, but I looked at the platform interface part and skimmed how that affected the implementation. It looks like this approach does what we need, and will be clearly applicable to other platforms (and if ever necessary, added view types) so I think we're good to split that piece out to its own PR now, so that it can be landed and unblock Android work, while the iOS review proceeds.

packages/video_player/video_player_avfoundation/test/avfoundation_video_player_test.dart
FirentisTFW: Does this test make sense here in the first place? [The exception](https://github.com/flutter/packages/pull/4318/files#diff-85f606af958fc0025baafe608f631ef8622a8f35e583c1e4f3b5f99cfdc659de) it is supposed to check comes from the native side.

stuartmorgan: Ugh, I see; sorry, I hadn't looked at the initial PR. You are correct, this test makes no sense here. It should have been either a native unit test, or a Dart integration test, since a Dart unit test can't possibly exercise that code. Unfortunately that was missed in review.

FirentisTFW: I replaced it with a native unit test.

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -15,6 +15,11 @@ import 'messages.g.dart';
 class AVFoundationVideoPlayer extends VideoPlayerPlatform {
   final AVFoundationVideoPlayerApi _api = AVFoundationVideoPlayerApi();
 
+  /// A map that associates player ID with a view type.
+  /// This is used to determine which view type to use when building a view.
+  @visibleForTesting
+  final Map<int, VideoViewType> playerViewTypes = <int, VideoViewType>{};
```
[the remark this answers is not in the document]
FirentisTFW: Good point, I did as suggested.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
```diff
@@ -0,0 +1,178 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FVPTextureBasedVideoPlayer_Test.h"
+
+@interface FVPTextureBasedVideoPlayer ()
```
hellohuanlin, line 7: What's the corresponding type for platform view based implementation? is there a `FVPPlatformViewBasedVideoPlayer`?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -158,69 +193,84 @@ - (nullable NSNumber *)createWithOptions:(nonnull FVPCreationOptions *)options
...
-  FVPVideoPlayer *player = self.playersByTextureId[playerKey];
-  [self.registry unregisterTexture:textureId];
-  [self.playersByTextureId removeObjectForKey:playerKey];
+- (void)disposePlayer:(NSInteger)playerId error:(FlutterError **)error {
+  NSNumber *playerKey = @(playerId);
+  FVPVideoPlayer *player = self.playersById[playerKey];
+  if ([player isKindOfClass:[FVPTextureBasedVideoPlayer class]]) {
+    [self.registry unregisterTexture:playerId];
+  }
+  [self.playersById removeObjectForKey:playerKey];
```
hellohuanlin, line 239: Can we have 2 types of video players both implementing the same API and rely on polymorphism?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
FirentisTFW: Actually, there's no additional type for platform view. It uses regular `FVPVideoPlayer` class as it has all functionalities needed for a platform view to work. Would you see that differently?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: Hmm, isn't that what we currently have? Right now the structure is as follows:
- `FVPVideoPlayer` (contains the basic functionality and can be used directly for platform views),
- `FVPTextureBasedVideoPlayer` (extends `FVPVideoPlayer` adding texture-specific functionalities).
Would you change something here?

Or, if you're talking about this piece of code specifically, we could also call `unregisterTexture` in `[player dipose]`, then we'd need to pass `FlutterTextureRegistry` there (or an additional callback, like `onDisposed`). Shall I do that?

hellohuanlin: I was thinking about ways to remove this kind of logic which happens a lot in this file: 

```
func doSomething() {
  if texture based {
    do A
  } else {
    do B
  }
}
```

I was suggesting having 2 players, each encapsulating its own logic, so that we don't mix the 2 logic in the same file: 
```
class TextureBasedPlayer {
  func doSomething() {
    do A
  }
}
class PlatformBasedPlayer {
  func doSomething() {
    do B
  }
}
```

then in this function we can just do: 
```
func doSomething() {
  player.doSomething()
}
```

FirentisTFW: I understand, I think this is what we're currently doing though.

I general, that's right, we do have some checks in this file. However, most of them are in `createWithOptions` method, and these checks are just to properly initialize the player, I'm not sure if we can move them out of there (though probably the method itself can be refactored and/or split up, I'll have another look into it tomorrow). If we cross out those, we're left with two checks:
- Here in `disposePlayer`, which I think we can actually get rid of if we pass `onDisposed` callback when initializing the player.
- In `onPlayerSetup` where we obtain a player ID for the player that was just created

FirentisTFW: I cleaned up a little bit the code related to player initialization (`createWithOptions` method) and got rid of the type check in `dispose`

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -132,95 +163,130 @@ - (void)initialize:(FlutterError *__autoreleasing *)error {
...
+    }
+
+    return @([self onPlayerSetup:player]);
+  } @catch (NSException *exception) {
+    *error = [FlutterError errorWithCode:@"video_player" message:exception.reason details:nil];
+    return nil;
+  }
+}
+
+- (nullable FVPTextureBasedVideoPlayer *)createTextureBasedPlayerFromOptions:
```
hellohuanlin, line 194: nit: *With*Options

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -132,95 +163,130 @@ - (void)initialize:(FlutterError *__autoreleasing *)error {
...
   FVPFrameUpdater *frameUpdater = [[FVPFrameUpdater alloc] initWithRegistry:_registry];
   FVPDisplayLink *displayLink =
       [self.displayLinkFactory displayLinkWithRegistrar:_registrar
                                                callback:^() {
                                                  [frameUpdater displayLinkFired];
                                                }];
 
-  FVPVideoPlayer *player;
+  void (^onDisposed)(int64_t) = ^(int64_t textureId) {
+    [self.registry unregisterTexture:textureId];
```
hellohuanlin, line 204: weak self

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -132,95 +163,130 @@ - (void)initialize:(FlutterError *__autoreleasing *)error {
...
+                                               httpHeaders:options.httpHeaders
+                                                 avFactory:_avFactory
+                                                 registrar:self.registrar
+                                                onDisposed:onDisposed];
+  }
+
+  return nil;
+}
+
+- (nullable FVPVideoPlayer *)createRegularPlayerFromOptions:(nonnull FVPCreationOptions *)options {
```
hellohuanlin, line 228: what's a "regular" player

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
hellohuanlin: Are there any functionalities (properties or methods) unique to the platform view based implementation, that is not applicable to texture based implementation?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: I updated the method name to `createPlatformViewPlayerWithOptions` and left a comment that using `FVPVideoPlayer` is sufficient there

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: Right, fixed

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
FirentisTFW: Nope, platform view player uses a subset of texture player features. That's why I did it this way - `FVPVideoPlayer` as a base class that could be also used directly in case of platform views

FirentisTFW:
@stuartmorgan I [added a fallback](https://github.com/flutter/packages/pull/8237/commits/417e7716df27150aa5c3cc17af61215a341517ff) to always use texture view on MacOS, even if platform view is requested. I'm not sure how to best document it. For sure I'll put the information about it in app-facing package (when we pass the view type to controller), but I'm thinking if we should also describe it better here, in the platform implementation. What are your thoughts on that?

stuartmorgan:
> I'm thinking if we should also describe it better here, in the platform implementation. What are your thoughts on that?

We mostly don't expect clients to interact with the video player implementation packages directly, but it wouldn't hurt to add a `## Platform limitations` section to the README for this package that mentions the lack of support.

packages/video_player/video_player_avfoundation/README.md
```diff
@@ -13,3 +13,7 @@ should add it to your `pubspec.yaml` as usual.
 
 [1]: https://pub.dev/packages/video_player
 [2]: https://flutter.dev/to/endorsed-federated-plugin
+
+## Platform limitations
+
+On macOS, the plugin does not support platform views. Instead, a texture view is always used to display the video player, even if `VideoViewType.platformView` is specified as a parameter.
```
hellohuanlin, line 19: This should be documented in API as well (if it's not already so)

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
hellohuanlin: Gotcha. Can you add some comments in `FVPVideoPlayer` explaining this?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer.h
```diff
@@ -77,10 +65,34 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (void)seekTo:(int64_t)location completionHandler:(void (^_Nullable)(BOOL))completionHandler;
+@end
+
+@interface FVPVideoPlayer ()
+/// The AVPlayerItemVideoOutput associated with this video player.
+@property(nonatomic, readonly) AVPlayerItemVideoOutput *videoOutput;
+/// The plugin registrar, to obtain view information from.
+@property(nonatomic, readonly) NSObject<FlutterPluginRegistrar> *registrar;
+/// The CALayer associated with the Flutter view this plugin is associated with, if any.
+@property(nonatomic, readonly, nullable) CALayer *flutterViewLayer;
```
hellohuanlin, line 76: is this CALayer only for platform view based implementation, but not applicable to texture based implementation?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoView.h
```diff
@@ -0,0 +1,27 @@
...
+#else
+#import <Flutter/Flutter.h>
+#endif
+
+#import <AVFoundation/AVFoundation.h>
+
+/// A class used to create a native video view that can be embedded in a Flutter app.
+/// This class wraps an AVPlayer instance and displays its video content.
+#if TARGET_OS_IOS
+@interface FVPNativeVideoView : NSObject <FlutterPlatformView>
```
hellohuanlin, line 20: Could you point me to where we add the view to the view hierarchy?

FirentisTFW: [Here](https://github.com/flutter/packages/pull/8237/files#diff-0da911603637e04d371c207214dc218fb75a1a2d7520f2120ba5c1b3348b25dd) is the native platform view factory implementation where we create instances of `FVPNativeVideoView`, and [here](https://github.com/flutter/packages/pull/8237/files#diff-1a110fd39c95303a6bd97033203f739866aab30518994ac55bf134e7d9f60e03R189-R216) is where we create the view on the Flutter side.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer.h
FirentisTFW: It's actually only needed in texture based implementation ([this comment](https://github.com/flutter/packages/pull/8237/files#diff-79247173115dd0e1159d9abc3dfbbfb6666258e7c3d7a0814b40fe5d751614b5R80-R86) says it fixes two bugs there). It should not be in `FVPVideoPlayer` anymore, looks like it got there when I was resolving conflicts. Thanks, removed it now

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
FirentisTFW: Yes, updated the class doc

packages/video_player/video_player_avfoundation/README.md
FirentisTFW: I added a comment in the implementation [here](https://github.com/flutter/packages/pull/8237/commits/417e7716df27150aa5c3cc17af61215a341517ff#diff-1a110fd39c95303a6bd97033203f739866aab30518994ac55bf134e7d9f60e03R56-R59). Then based on the discussion with Stuart ([question](https://github.com/flutter/packages/pull/8237#issuecomment-2622020563) and [reply](https://github.com/flutter/packages/pull/8237#issuecomment-2622706965)) I mentioned it in the README. Would you suggest to add a dartdoc on `createWithOptions` method where this fallback is implemented?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer.h
```diff
@@ -77,10 +68,32 @@ NS_ASSUME_NONNULL_BEGIN
 /// Seeks to the specified location in the video and calls the completion handler when done, if one
 /// is supplied.
 - (void)seekTo:(int64_t)location completionHandler:(void (^_Nullable)(BOOL))completionHandler;
+@end
+
+@interface FVPVideoPlayer ()
```
stuartmorgan, line 73: Why is there a private category in the public header? If these should be public, they can just be part of the main interface.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer.h
```diff
@@ -77,10 +68,32 @@ NS_ASSUME_NONNULL_BEGIN
...
+/// link, AV factory, and registrar.
+- (instancetype)initWithPlayerItem:(AVPlayerItem *)item
+                         avFactory:(id<FVPAVFactory>)avFactory
+                         registrar:(NSObject<FlutterPluginRegistrar> *)registrar;
 
-/// Tells the player to run its frame updater until it receives a frame, regardless of the
-/// play/pause state.
-- (void)expectFrame;
+/// Updates the playing state of the video player.
+- (void)updatePlayingState;
```
stuartmorgan, line 96: Why does this need to be part of the public interface? It should be an implementation detail.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer.h
stuartmorgan: Although it's not clear to me why most of this would need to be public. Is this for subclassing? If so, that can be in an `..._Internal.h` file.

hellohuanlin (approved):
thanks for making the changes

packages/video_player/video_player_avfoundation/README.md
stuartmorgan: > since that's where developers use the api

~0 developers outside of the Flutter team will use `createWithOptions`. It's a platform interface API used by the app-facing package.

packages/video_player/video_player_avfoundation/pubspec.yaml
```diff
@@ -24,7 +24,7 @@ flutter:
 dependencies:
   flutter:
     sdk: flutter
-  video_player_platform_interface: ">=6.1.0 <7.0.0"
+  video_player_platform_interface: ">=6.3.0 <7.0.0"
```
stuartmorgan, line 27: This can just be `^6.3.0`.

packages/video_player/video_player_avfoundation/example/pubspec.yaml
```diff
@@ -16,7 +16,7 @@ dependencies:
     # The example app is bundled with the plugin so we use a path dependency on
     # the parent directory to use the current plugin's version.
     path: ../
-  video_player_platform_interface: ">=6.1.0 <7.0.0"
+  video_player_platform_interface: ">=6.3.0 <7.0.0"
```
stuartmorgan, line 19: `^6.3.0`

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
```diff
@@ -0,0 +1,205 @@
...
+                    registrar:(NSObject<FlutterPluginRegistrar> *)registrar
+                   onDisposed:(void (^)(int64_t))onDisposed {
+  NSString *path = [[NSBundle mainBundle] pathForResource:asset ofType:nil];
+#if TARGET_OS_OSX
+  // See https://github.com/flutter/flutter/issues/135302
+  // TODO(stuartmorgan): Remove this if the asset APIs are adjusted to work better for macOS.
+  if (!path) {
+    path = [NSURL URLWithString:asset relativeToURL:NSBundle.mainBundle.bundleURL].path;
+  }
+#endif
```
stuartmorgan, line 35: Asset key -> path logic should be extracted to a shared helper function, rather than duplicated.

tarrinneal:
I'll get to this review asap. Thanks for your patience.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoViewFactory.h
```diff
@@ -0,0 +1,24 @@
...
+
+#import "FVPVideoPlayer.h"
+
+/// A factory class responsible for creating native video views that can be embedded in a
+/// Flutter app.
+@interface FVPNativeVideoViewFactory : NSObject <FlutterPlatformViewFactory>
+/// Initializes a new instance of FVPNativeVideoViewFactory with the given messenger and
+/// a block that provides video players associated with their IDs.
+- (instancetype)initWithMessenger:(NSObject<FlutterBinaryMessenger> *)messenger
+               playerByIdProvider:(FVPVideoPlayer * (^)(NSNumber *))playerByIdProvider;
```
stuartmorgan, line 23: Nit: Id -> Identifier, per standard Obj-C naming.

stuartmorgan (changes requested):

packages/video_player/video_player_avfoundation/README.md
FirentisTFW: I'm leaving it as is then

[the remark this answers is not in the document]
FirentisTFW: Right, good point

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoView.h
```diff
@@ -0,0 +1,27 @@
...
+
+#import <TargetConditionals.h>
+
+#if TARGET_OS_OSX
+#import <FlutterMacOS/FlutterMacOS.h>
+#else
+#import <Flutter/Flutter.h>
+#endif
+
+#import <AVFoundation/AVFoundation.h>
```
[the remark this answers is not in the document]
FirentisTFW: Thanks, moved conditional imports to the end

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer.h
FirentisTFW: Yes, it's for subclassing. I wasn't aware of `..._Internal.h` pattern, I thought that having a class extension here is the common practice. I moved it to `FVPVideoPlayer_Internal.h` as suggested

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer.h
FirentisTFW: It was for subclassing, answered [here](https://github.com/flutter/packages/pull/8237#discussion_r1943269864).

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
FirentisTFW: Good catch, missed that. Extracted to a helper method

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoViewFactory.h
stuartmorgan: Ideally we would do it everywhere, but that's complicated by the fact that we don't have an annotation in Pigeon for renaming class members, params, etc., and `playerId` *is* idiomatic for Dart, so some amount of `playerId` will filter into the Obj-C code. I didn't audit for how much inconsistency we would get if we renamed all the Obj-C members and left just the Pigeon-generated code as using the non-idiomatic naming.

If the inconsistency is fairly limited that changing it everywhere would be preferred, as `Id` is very much not standard Obj-C (`Identifier` is dominant in Apple SDKs, occasionally they use `ID`). But if changing variables would add a lot of inconsistency I'd settle for just making method signatures use standard naming (thus flagging just this one).

> I had a quick look into other packages and looks like we use both `Id` and `Identifier` there.

The nature of this repo is unfortunately that because there's so much context switching between languages—even in a single PR—things that break language-specific naming idioms are frequently missed in review. I still try though 🙂

FirentisTFW: Okay, thanks for the detailed explanation. I'll see what I can do with renaming the params then. I'll let you know when I'm finished with this

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
```diff
@@ -0,0 +1,197 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "./include/video_player_avfoundation/FVPTextureBasedVideoPlayer_Test.h"
```
stuartmorgan, line 5: This should be:

```objc
#import "./include/video_player_avfoundation/FVPTextureBasedVideoPlayer.h"
#import "./include/video_player_avfoundation/FVPTextureBasedVideoPlayer_Test.h"
```

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -62,32 +81,44 @@ - (instancetype)initWithAVFactory:(id<FVPAVFactory>)avFactory
...
   SetUpFVPAVFoundationVideoPlayerApi(registrar.messenger, nil);
 }
 
-- (int64_t)onPlayerSetup:(FVPVideoPlayer *)player frameUpdater:(FVPFrameUpdater *)frameUpdater {
-  int64_t textureId = [self.registry registerTexture:player];
-  frameUpdater.textureId = textureId;
+- (int64_t)onPlayerSetup:(FVPVideoPlayer *)player {
+  BOOL textureBased = [player isKindOfClass:[FVPTextureBasedVideoPlayer class]];
+  int64_t playerId;
+  if (textureBased) {
```
stuartmorgan, line 100: Optional: an alternate way to structure this would be:

```objc
FVPTextureBasedVideoPlayer* textureBasedPlayer = [player isKindOfClass:[FVPTextureBasedVideoPlayer class]] ?
    (FVPTextureBasedVideoPlayer *)player : nil;

if (textureBasedPlayer) {
  playerId = [self.registry registerTexture:textureBasedPlayer];
  ...
}
...
[textureBasedPlayer expectFrame];
```

Then the only casting is happening in the same statement with the check that makes the cast safe, so there's less room for error in future refactoring.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -132,95 +163,133 @@ - (void)initialize:(FlutterError *__autoreleasing *)error {
...
                                    error:(FlutterError **)error {
+  BOOL textureBased = options.viewType == FVPPlatformVideoViewTypeTextureView;
+
+  @try {
+    FVPVideoPlayer *player;
+    if (textureBased) {
+      player = [self createTexturePlayerWithOptions:options];
+    } else {
+      player = [self createPlatformViewPlayerWithOptions:options];
+    }
```
stuartmorgan, line 180: Optional:

```objc
FVPVideoPlayer *player = textureBased
    ? [self createTexturePlayerWithOptions:options]
    : [self createPlatformViewPlayerWithOptions:options];
```

(This keeps the declaration and initialization in the same statement, making it clearer that it's always initialized.)

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -132,95 +163,133 @@ - (void)initialize:(FlutterError *__autoreleasing *)error {
...
+
+    return @([self onPlayerSetup:player]);
+  } @catch (NSException *exception) {
+    *error = [FlutterError errorWithCode:@"video_player" message:exception.reason details:nil];
+    return nil;
+  }
+}
+
+- (nullable FVPTextureBasedVideoPlayer *)createTexturePlayerWithOptions:
```
stuartmorgan, line 194: Nit: the naming of these two methods would be more idiomatic if you remove the `create`. Convenience constructors in Obj-C generally just start with the name of the thing they create.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -132,95 +163,133 @@ - (void)initialize:(FlutterError *__autoreleasing *)error {
...
 }
 
-- (void)disposePlayer:(NSInteger)textureId error:(FlutterError **)error {
-  NSNumber *playerKey = @(textureId);
-  FVPVideoPlayer *player = self.playersByTextureId[playerKey];
-  [self.registry unregisterTexture:textureId];
-  [self.playersByTextureId removeObjectForKey:playerKey];
-  if (!player.disposed) {
-    [player dispose];
+- (NSString *)prepareAssetPathFromCreationOptions:(nonnull FVPCreationOptions *)options {
```
stuartmorgan, line 247: Nit: `assetPathFromCreationOptions:`

(Similar reason here; since the method is intended to create and return something, the name should generally start by naming that thing, without a verb.)

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPNativeVideoViewFactory.h
```diff
@@ -0,0 +1,24 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Platform views are only supported on iOS as of now. Ifdef is used to avoid compilation errors.
+
+#import <TargetConditionals.h>
```
stuartmorgan, line 7: This file should import `Foundation` given the types it's using, at which point you won't need this.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/include/video_player_avfoundation/FVPVideoPlayer_Internal.h
```diff
@@ -0,0 +1,47 @@
...
+
+#if TARGET_OS_OSX
+#import <FlutterMacOS/FlutterMacOS.h>
+#else
+#import <Flutter/Flutter.h>
+#endif
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FVPVideoPlayer ()
```
stuartmorgan, line 17:
```objc
/// Interface intended for use by subclasses, but not other callers.
```

stuartmorgan (approved):
LGTM with minor comments (none that need another round of review, and feel free to ignore the optional ones if you disagree with them). For the `Id`/`Identifier` thing, you can use your judgement about where to stop renaming.

If there's anything you'd specifically like me to re-review let me know and I'll take a look, but otherwise feel free to land once the comments are addressed!

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
FirentisTFW: Thanks, I was wondering about that part

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
FirentisTFW: Thanks, wasn't aware of that. Fixed now!

FirentisTFW:
@stuartmorgan @hellohuanlin Thanks for the review, I addressed all of your comments.

For `Id`/`Identifier` - I renamed the `Id` to `Identifier` where possible. It worked great, now `Id` is mostly left in pigeon-generated files.
I also applied optional suggestions from the last review.

Thanks for the iteration here, I'm merging it now.

PR #8302, opened 2024-12-15
[pigeon] [swift] Fix `PigeonError` sendability conformance in Swift 6

bc-lee:
Ensure `PigeonError` conforms to `Sendable` for Swift 6.0 compatibility by updating the `details` property type from `Any?` to `Sendable?`.  

Fixes https://github.com/flutter/flutter/issues/160313

## Pre-launch Checklist

- [x] I signed the [CLA].


hellohuanlin:
This will break old sdk where Sendable is undefined. You may have to put it behind a flag.

bc-lee:
> This will break old sdk where Sendable is undefined. You may have to put it behind a flag.

Are there any minimum Xcode requirements for Flutter projects? Sendable was added in Xcode 13.0 with Swift 5.5.

bc-lee:
I found the answer.  
- Flutter 3.12 or later requires Xcode 14.0 or newer.  
  [Source](https://github.com/flutter/flutter/pull/125639)  
- Flutter 3.0 or later requires Xcode 13.0 or newer.  
  [Source](https://github.com/flutter/flutter/pull/97746)  

So, I think it is safe to use this in Pigeon, as Pigeon requires Dart 3.3 and Flutter 3.19.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## NEXT
```
tarrinneal, line 1: This should be updated to 22.7.1 along with the pubspec version and the generator_tools file.

I guess this is technically a breaking change, so 23.0.0 might be more correct.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
...
 
 * Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+* [swift] Make `PigeonError` Class Conform to `Sendable`
```
tarrinneal, line 4: ```[swift] Changes `PigeonError` class to conform to `Sendable`.```

tarrinneal (changes requested):

tarrinneal:
@stuartmorgan shouldn't this pr be failing the test exemption check?

hellohuanlin (approved):
lgtm after addressing the comments.

stuartmorgan (changes requested):
> @stuartmorgan shouldn't this pr be failing the test exemption check?

I think there's a false negative due to the changes to generated files in `platform_tests`; we should probably tune the Cocoon rules to ignore `.gen.` files in there.

Regardless, this does need testing. A Dart unit test should be fine for this for now (although long term we probably want to adjust the Swift integration tests to actually test a flow that ensures that everything is `Sendable`)

bc-lee:
> Regardless, this does need testing. A Dart unit test should be fine for this for now (although long term we probably want to adjust the Swift integration tests to actually test a flow that ensures that everything is `Sendable`)

I have added the Dart test introduced by this PR. I also agree that the long-term solution would involve updating the Swift integration tests—not just to check for `@Sendable`, but to cover all aspects related to Swift's strict concurrency model. However, I believe this would require additional setup.

tarrinneal:
@stuartmorgan you'll need to approve this one to land it.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 23.0.0
+
+* [swift] Changes `PigeonError` class to conform to `Sendable`.
```
stuartmorgan, line 3: Breaking changes need to follow the CHANGELOG style guide for breaking changes.

stuartmorgan (dismissed):

packages/pigeon/CHANGELOG.md
tarrinneal: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version-and-changelog-updates

hellohuanlin: I don't think change log is accurate, because `PigeonError` already conforms to `Error`, which is by design `Sendable`. This PR does not change that. 

How about "Fixes compile error related to `PigeonError`'s sendability conformance in Swift 6 language mode"?

hellohuanlin (changes requested):
Sorry for the block (I removed autosubmit tag), just saw the discussion on CHANGELOG and left a comment there

packages/pigeon/CHANGELOG.md
tarrinneal: If it was already Sendable, why would we need to change Any? to Sendable? I'm confused.

hellohuanlin: The `PigeonError` type itself was `Sendable`, but its `details` field was not `Sendable`, hence the compile error in Swift 6 (or compile warning in Swift 5's strict mode). 

That being said, this PR is still a breaking change tho, because the `init` signature is changed.

tarrinneal:
> Sorry for the block (I removed autosubmit tag), just saw the discussion on CHANGELOG and left a comment there: [#8302 (comment)](https://github.com/flutter/packages/pull/8302#discussion_r1937799207)

I fixed that, can you approve so we can land?

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 23.0.0
+
+* **Breaking Change** [swift] Fixes a compile error about PigeonError's sendability conformance in Swift 6.
```
hellohuanlin, line 3:
```suggestion
**Breaking Change** [swift] Fixes a compile error about PigeonError's sendability conformance in Swift 6. PigeonError's init now requires the "details" argument to be Sendable.
```

hellohuanlin (approved):

packages/pigeon/CHANGELOG.md
bc-lee: Sorry for the delay—I was on a national holiday.  

Yes, the above discussion is correct. `PigeonError` already conforms to `Error`, which also makes it `Sendable`. However, the type of its `details` field prevents it from conforming under Strict Concurrency, leading to a compilation error in Swift 6.

tarrinneal:
> autosubmit label was removed for flutter/packages/8302, because This PR has not met approval requirements for merging. Changes were requested by {stuartmorgan}, please make the needed changes and resubmit this PR. The PR author is not a member of flutter-hackers and needs 1 more review(s) in order to merge this PR.
> 
> * Merge guidelines: A PR needs at least one approved review if the author is already part of flutter-hackers or two member reviews if the author is not a flutter-hacker before re-applying the autosubmit label. **Reviewers**: If you left a comment approving, please use the "approve" review action instead.

@stuartmorgan not sure what the deal is here. I did "invalidate" your review since it had been addressed. Might be related.

stuartmorgan (approved):
Weird, seems like a Cocoon bug one way or another.

PR #8305, opened 2024-12-16
[vector_graphics] allow specifying target size when parsing svg

eEQK:
moved from https://github.com/dnfield/vector_graphics/pull/240

closes https://github.com/flutter/flutter/issues/158766
and technically also in old repo: https://github.com/dnfield/flutter_svg/issues/971


packages/vector_graphics/lib/src/listener.dart
```diff
@@ -477,6 +484,7 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
       _locale,
       _textDirection,
       _clipViewbox,
+      null,
```
jonahwilliams, line 487: ?

packages/vector_graphics/test/scale/scale_test.dart
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  testWidgets('resizes to the same size', (_) async {
+    final ui.Image image = await decode(
+      target: const Size(24, 24),
+    );
+
+    await expectLater(
+      image,
+      matchesGoldenFile('thumb_up_24px.png'),
```
jonahwilliams, line 65: @stuartmorgan is golden file usage common in this repo?

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -561,6 +569,14 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
 
   @override
   void onSize(double width, double height) {
+    final double targetWidth = _targetSize?.width ?? width;
+    final double targetHeight = _targetSize?.height ?? height;
+
+    final double sx = targetWidth / width;
+    final double sy = targetHeight / height;
+
+    _canvas.scale(min(sx, sy));
```
jonahwilliams, line 578: https://api.flutter.dev/flutter/dart-ui/Canvas/scale.html

You need to use separate x/y scaling otherwise this will won't work correctly for non-uniform scales

jonahwilliams (changes requested):

packages/vector_graphics/test/scale/scale_test.dart
stuartmorgan: We have a small number of packages that currently use goldens; I'm not sure if this is one offhand.

jonahwilliams:
Based on no reponse for 3 weeks closing PR

PR #8326, opened 2024-12-18
go_router_builder: latest dependencies

kevmoo:


chunhtai (approved):
LGTM

kevmoo:
@chunhtai @stuartmorgan – thoughts on `Linux analyze_legacy N-1` failing? I think it's testing a too-old version of Flutter. How can I tweak that?

stuartmorgan:
> thoughts on `Linux analyze_legacy N-1` failing? I think it's testing a too-old version of Flutter.

I don't understand what you mean by "too old". N-1 is the previous stable version. That's a version that exists, and that this package's pubspec claims it supports, but that fails to resolve with your changes:

```
Resolving dependencies...
Because analyzer >=6.9.0 depends on macros >=0.1.3-main.0 <0.1.4 which depends on _macros 0.3.3 from sdk, analyzer >=6.9.0 requires _macros 0.3.3 from sdk.
So, because no versions of _macros from sdk match 0.3.3 and go_router_builder depends on analyzer >=6.9.0 <8.0.0, version solving failed.
```

> How can I tweak that?

Tweak what? You can't change what the `N-1` test is testing because the whole point of that test is to find problems like this and prevent publishing them since they would break people.

You can either use min dependency versions that are compatible with the version of Flutter you are claiming support for, or you can change the version you are claiming support for to match the reality of what the dependencies you want will allow.

kevmoo:
@stuartmorgan – I think I tried to land this before the "current" SDK was bump up. Hopefully this fixes things. Sorry for the confusion.

packages/go_router_builder/tool/run_tests.dart
```diff
@@ -59,3 +63,20 @@ Future<void> main() async {
     }, timeout: const Timeout(Duration(seconds: 100)));
   }
 }
+
+Future<Version> _packageVersion() async {
+  final PackageConfig packageConfig =
+      await loadPackageConfigUri(Isolate.packageConfigSync!);
+
+  final String pkgPath = Platform.script.resolve('../').path;
```
stuartmorgan, line 71: I suspect the `/` here is what's breaking the test on Windows.

kevmoo: Good try. Let me hack.

kevmoo:
@stuartmorgan – sorry to bother you. Suggestions for dealing w/ unrelated flakes? 
We generally don't land even if we're very confident we didn't cause the failures

stuartmorgan:
Correct, we re-run the failing test until it passes. If it's some specific bad flake, we disable that test and file an issue.

PR #7941, opened 2024-10-28
[vector_graphics] Initial import

stuartmorgan:
Imports https://github.com/dnfield/vector_graphics into this repository, with history, and updates it to follow repository conventions:
- Adds missing licence headers.
- Removes analysis options and fixes resulting warnings.
- Updates min SDKs.
- Autoformats.
- Updates pubspecs to follow repo standard, including updating the metadata
  to point to the Flutter repositories.
- Allows `xml` as a pinned dependency, and relaxes the pinning requirement
  in the repo tools to allow an explicit, inclusive-endpoint range.
- Skips some unit tests on unsupported platforms.
- Adds repo metadata.
- Bumps versions and slightly relax version constraints to allow the new
  versions.

PR #8327, opened 2024-12-18
Drop vector bits from allowed_unpinned_deps.yaml

kevmoo:
Now that https://github.com/flutter/packages/pull/7941 has landed


stuartmorgan (approved):
LGTM

Thanks, I had forgotten about this. I thought I made a note to myself, but apparently not!

stuartmorgan:
FYI you are much better off re-running failed tests instead of merging in `main`. There's a certain amount of flake in both infra and in the integration tests (including the test frameworks themselves), and a PR like this that runs every test in every package is maximally subject to flake. If you merge, you are rolling the dice on every test again, including the ones that had passed before.

PR #8187, opened 2024-11-26
[shared_preferences] Limit Android decoding

stuartmorgan:
When decoding stored preferences for `List<String>` support, only allow lists and strings to avoid potentially instantiating other object types.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ ] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version
[following repository CHANGELOG style]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style
[exempt from CHANGELOG changes]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog
[test-exempt]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests

PR #8335, opened 2024-12-20
[shared_preferences] update List<String> encode/decode

tarrinneal:
Follow up to https://github.com/flutter/packages/pull/8187

Updates Android `List<String>` encoding/decoding to use JSON to avoid potential future breakages from restricted list types.


tarrinneal:
@stuartmorgan ready for review here as well

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
```diff
@@ -19,24 +19,24 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /** LegacySharedPreferencesPlugin */
 public class LegacySharedPreferencesPlugin implements FlutterPlugin, SharedPreferencesApi {
   private static final String TAG = "SharedPreferencesPlugin";
   private static final String SHARED_PREFERENCES_NAME = "FlutterSharedPreferences";
+  private static final String JSON_LIST_IDENTIFIER = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu!";
```
reidbaker, line 34: Does this need to be a very similar (and still random string) to `LIST_IDENTIFIER` if not then lets generate a new one to make it easier to tell them apart.

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
```diff
@@ -99,6 +92,7 @@ public void onDetachedFromEngine(@NonNull FlutterPlugin.FlutterPluginBinding bin
     return preferences.edit().remove(key).commit();
   }
 
+  // Deprecated, for testing purposes only.
```
reidbaker, line 95: Consider using `@Deprecated`

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
```diff
@@ -131,14 +125,13 @@ public void onDetachedFromEngine(@NonNull FlutterPlugin.FlutterPluginBinding bin
...
   // Optionally filtered also to only those items in the optional [allowList].
-  @SuppressWarnings("unchecked")
   private @NonNull Map<String, Object> getAllPrefs(
       @NonNull String prefix, @Nullable Set<String> allowList) throws RuntimeException {
     Map<String, ?> allPrefs = preferences.getAll();
     Map<String, Object> filteredPrefs = new HashMap<>();
     for (String key : allPrefs.keySet()) {
       if (key.startsWith(prefix) && (allowList == null || allowList.contains(key))) {
-        filteredPrefs.put(key, transformPref(key, allPrefs.get(key)));
+        filteredPrefs.put(key, transformPref(key, Objects.requireNonNull(allPrefs.get(key))));
```
reidbaker, line 134: What was requireNonNull needed here?  Also this code is getting a bit difficult to parse. (A loop with a conditional that sets some transformed filtered data) Consider breaking parts out into methods or adding comments.

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
```diff
@@ -148,32 +141,14 @@ public void onDetachedFromEngine(@NonNull FlutterPlugin.FlutterPluginBinding bin
   private Object transformPref(@NonNull String key, @NonNull Object value) {
     if (value instanceof String) {
       String stringValue = (String) value;
-      if (stringValue.startsWith(LIST_IDENTIFIER)) {
+      if (stringValue.startsWith(JSON_LIST_IDENTIFIER)) {
```
reidbaker, line 144: I think this is where having the keys be similar starts to be important. That said from reading the code the reader does not know the keys are overlapping and it would seem safe in code review to swap the order of these if statements. 

Consider making the keys independent so that you are not relying on the contents of the keys. If that is too much code change then consider making a "LIST_PREFIX" that can be checked so that it is at least clear in code what the relationship is. 

If you want further reading see 
https://www.16elt.com/2024/09/25/first-book-of-byte-sized-tech/

I think the first example explains the principal I am applying.

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
```diff
@@ -148,32 +141,14 @@ public void onDetachedFromEngine(@NonNull FlutterPlugin.FlutterPluginBinding bin
...
-      } else if (stringValue.startsWith(BIG_INTEGER_PREFIX)) {
-        // TODO (tarrinneal): Remove all BigInt code.
-        // https://github.com/flutter/flutter/issues/124420
-        String encoded = stringValue.substring(BIG_INTEGER_PREFIX.length());
-        return new BigInteger(encoded, Character.MAX_RADIX);
       } else if (stringValue.startsWith(DOUBLE_PREFIX)) {
         String doubleStr = stringValue.substring(DOUBLE_PREFIX.length());
         return Double.valueOf(doubleStr);
       }
-    } else if (value instanceof Set) {
```
reidbaker, line 162: Is this a breaking change or does the other json migration handle this now?

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/MessagesAsync.g.kt
```diff
@@ -105,7 +109,7 @@ interface SharedPreferencesAsyncApi {
   /** Gets individual int value stored with [key], if any. */
   fun getInt(key: String, options: SharedPreferencesPigeonOptions): Long?
   /** Gets individual List<String> value stored with [key], if any. */
-  fun getStringList(key: String, options: SharedPreferencesPigeonOptions): List<String>?
+  fun getStringList(key: String, options: SharedPreferencesPigeonOptions): Any?
```
reidbaker, line 112: Can we not keep the return a list of string? if not why not?

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.kt
```diff
@@ -31,6 +31,7 @@ import kotlinx.coroutines.runBlocking
 const val TAG = "SharedPreferencesPlugin"
 const val SHARED_PREFERENCES_NAME = "FlutterSharedPreferences"
 const val LIST_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu"
+const val JSON_LIST_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu!"
```
reidbaker, line 34: Does this need to be the same value as LegacySharedPreferencesPlugin? If so can you add  comment as to why and that they need to be identical? 

Same comments as above about using similar strings. I think this should be a different value entirely.

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.kt
```diff
@@ -189,9 +190,16 @@ class SharedPreferencesPlugin() : FlutterPlugin, SharedPreferencesAsyncApi {
   }
 
   /** Gets StringList at [key] from data store. */
-  override fun getStringList(key: String, options: SharedPreferencesPigeonOptions): List<String>? {
-    val value: List<*>? = transformPref(getString(key, options) as Any?, listEncoder) as List<*>?
-    return value?.filterIsInstance<String>()
+  override fun getStringList(key: String, options: SharedPreferencesPigeonOptions): Any? {
```
reidbaker, line 193: Can we force cast this to a List<String>?

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.kt
```diff
@@ -400,7 +411,9 @@ internal fun preferencesFilter(key: String, value: Any?, allowList: Set<String>?
 /** Transforms preferences that are stored as Strings back to original type. */
 internal fun transformPref(value: Any?, listEncoder: SharedPreferencesListEncoder): Any? {
   if (value is String) {
-    if (value.startsWith(LIST_PREFIX)) {
+    if (value.startsWith(JSON_LIST_PREFIX)) {
```
reidbaker, line 414: Same comment here about the hidden complexity of similar strings.

packages/shared_preferences/shared_preferences_android/example/integration_test/shared_preferences_test.dart
```diff
@@ -493,6 +492,37 @@ void main() {
         expect(values[key], null);
       }
     });
+
+    testWidgets(
+        'old list encoding with getPreferences can be re-added with new encoding without data loss',
```
reidbaker, line 497: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-newold-modifiers-in-code

 Consider instead "Non Json list encoding"

packages/shared_preferences/shared_preferences_android/example/integration_test/shared_preferences_test.dart
```diff
@@ -757,6 +787,69 @@ void main() {
         expect(await preferences.getDouble(doubleKey, options), testDouble);
         expect(await preferences.getStringList(listKey, options), testList);
       });
+
+      testWidgets(
+          'list encoding updates to new encoding process without data loss with $backend',
```
reidbaker, line 792: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-newold-modifiers-in-code

`list encoding updates to json encoding ...`

packages/shared_preferences/shared_preferences_android/example/integration_test/shared_preferences_test.dart
```diff
@@ -757,6 +787,69 @@ void main() {
...
+            listKey, testList, options);
+        final List<String>? oldSystemList =
+            await preferences.getStringList(listKey, options);
+        expect(oldSystemList, testList);
+        await preferences.setStringList(listKey, oldSystemList!, options);
+        expect(await preferences.getStringList(listKey, options), testList);
+      });
+
+      testWidgets(
+          'old list encoding still functions with getPreferences with $backend',
```
reidbaker, line 809: same

packages/shared_preferences/shared_preferences_android/example/integration_test/shared_preferences_test.dart
```diff
@@ -757,6 +787,69 @@ void main() {
...
+        ]);
+
+        final Map<String, Object> prefs = await preferences.getPreferences(
+            const GetPreferencesParameters(filter: PreferencesFilters()),
+            options);
+        expect(prefs[listKey], testList);
+      });
+
+      testWidgets(
+          'old list encoding with getPreferences can be re-added with new encoding without data loss with $backend',
```
reidbaker, line 831: same

packages/shared_preferences/shared_preferences_android/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 3.0.0
```
reidbaker, line 1: Are the breaking changes required? if not consider fixing/mitigating the security issue independent of the breaking changes. It makes it more likely for them to be adopted.

packages/shared_preferences/shared_preferences_android/lib/src/strings.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// String prefix for lists that are encoded with json in dart.
+const String jsonListPrefix = 'VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu!';
```
reidbaker, line 6: This is a 3rd place where these strings are found. If they are not being generated from a file someplace consider linking to each location from every other location in a comment.

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_android.dart
```diff
@@ -39,20 +41,27 @@ class SharedPreferencesAndroid extends SharedPreferencesStorePlatform {
...
+        return _api.setString(key, value);
       case 'Bool':
         return _api.setBool(key, value as bool);
       case 'Int':
         return _api.setInt(key, value as int);
       case 'Double':
         return _api.setDouble(key, value as double);
       case 'StringList':
+        return _api.setString(key, '$jsonListPrefix${jsonEncode(value)}');
+      case 'LegacyStringListForTesting':
```
reidbaker, line 60: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-newold-modifiers-in-code

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_android.dart
```diff
@@ -39,20 +41,27 @@ class SharedPreferencesAndroid extends SharedPreferencesStorePlatform {
   Future<bool> setValue(String valueType, String key, Object value) async {
     switch (valueType) {
       case 'String':
-        return _api.setString(key, value as String);
+        value as String;
+        if (value.startsWith(listPrefix) ||
```
reidbaker, line 45: Will this list grow? if so consider making a set of prefixes to check.

packages/shared_preferences/shared_preferences_android/CHANGELOG.md
stuartmorgan: Breaking changes to `shared_preferences` have the potential to be quite disruptive to the ecosystem; I would strongly prefer we keep the legacy type support around to making a breaking change.

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.kt
stuartmorgan: I think what we want is a structure like:

```kt
if (value.startsWith(LIST_PREFIX)) {
  // The newer JSON-encoded lists use an extended prefix to distinguish them.
  if (value.startsWith(JSON_LIST_PREFIX) {
    ...
  } else {
     ...
  }
} else if (value.startsWith(DOUBLE_PREFIX)) {
  ...
```

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_android.dart
stuartmorgan: We don't generally add types to `shared_preferences` since anything that all platforms don't support ends up with unpleasant complexity like the `List<String>` mess we are currently dealing with.

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_async_android.dart
```diff
@@ -126,6 +129,21 @@ base class SharedPreferencesAsyncAndroid
...
     SharedPreferencesOptions options,
+  ) async {
+    final SharedPreferencesPigeonOptions pigeonOptions =
+        _convertOptionsToPigeonOptions(options);
+    final SharedPreferencesAsyncApi api = _getApiForBackend(pigeonOptions);
+    final String stringValue = '$jsonListPrefix${jsonEncode(value)}';
+    return api.setString(key, stringValue, pigeonOptions);
+  }
+
+  /// Adds a `List<String>` to preferences using the no longer used system to test
```
stuartmorgan, line 140: Nit: "the no-longer-used system"

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
tarrinneal: It is meant to reduce potential breakage, but realistically it's not too likely to have conflict anyway. @stuartmorgan would you like to defend your idea here?

stuartmorgan: I defended it [here](https://github.com/flutter/packages/pull/8335#discussion_r1904440587) instead; I still would like to do this, and have the code be clearer that this is structurally true.

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
tarrinneal: It's a change suggested by a warning I was getting in Android Studio. Since the method being called requires nonnull arguments and getting a value from a map doesn't programmatically guarantee a value. In this context, there would always be a value though.

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
reidbaker: Ok I can understand that fear and why it might fail. Thank you for being open to structuring the code differently so that the overlap is easier to follow for future maintainers.

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
tarrinneal: support for Sets was removed a long time ago, this is removing legacy code for reading sets from preferences. It is a breaking change. As noted by the version bump

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/MessagesAsync.g.kt
tarrinneal: We cannot, this can now return either a list of strings, or an un-decoded string. The new way requires the string be encoded still, while the old way decoded on the platform, we cannot know which will happen in advance.

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.kt
tarrinneal: The legacy system and the new system can both read from the same files, so we need the same prefixes

packages/shared_preferences/shared_preferences_android/CHANGELOG.md
tarrinneal: Fair enough. I originally expected this change was going to end up being "technically" breaking due to the new prefix (before we discussed the prefix extension plan) so I wanted to wrap these long awaited breaking changes in as well. I'll revert them.

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
tarrinneal: > I defended it [here](https://github.com/flutter/packages/pull/8335#discussion_r1904440587) instead; I still would like to do this, and have the code be clearer that this is structurally true.

Sorry, I deleted my comment after I saw your other response

packages/shared_preferences/shared_preferences_android/CHANGELOG.md
tarrinneal: > Breaking changes to `shared_preferences` have the potential to be quite disruptive to the ecosystem; I would strongly prefer we keep the legacy type support around to making a breaking change.

Should I revert the error handling changes as well?

tarrinneal: > Should I revert the error handling changes as well?

I did this.

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/MessagesAsync.g.kt
stuartmorgan: I don't necessarily think this is a good idea, but to put it out there for consideration: we could have a `String? getJsonStringList()` and a `List<String>? getObjectEncodedStringList()`, and make the former return `''` as a sentinel value meaning "there's no JSON-encoded value, but there is an object-encoded value, so ask for that instead".

Pros:
- Strong typing.
- Separates new and legacy code more structurally.

Cons:
- Adds the complexity of understanding the sentinel value in the API flow. (We'd document it of course, but still, it's non-obvious.)
- Adds a bit of extra time to pre-conversion lookup. (I don't think this is a problem as it's only a one-time-per-legacy-value cost, but listing for completeness.)

I think I lean toward doing it for the type safety, but I'm on the fence.

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/SharedPreferencesPlugin.kt
```diff
@@ -189,9 +198,18 @@ class SharedPreferencesPlugin() : FlutterPlugin, SharedPreferencesAsyncApi {
...
 
   /** Gets StringList at [key] from data store. */
-  override fun getStringList(key: String, options: SharedPreferencesPigeonOptions): List<String>? {
-    val value: List<*>? = transformPref(getString(key, options) as Any?, listEncoder) as List<*>?
-    return value?.filterIsInstance<String>()
+  override fun getStringList(key: String, options: SharedPreferencesPigeonOptions): Any? {
+    val stringValue = getString(key, options)
+    stringValue?.let {
+      // The JSON-encoded lists use an extended prefix to distinguish them from
+      // lists that are encoded on the platform.
```
stuartmorgan, line 205: s/on the platform/using listEncoder/

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_async_android.dart
```diff
@@ -130,7 +133,22 @@ base class SharedPreferencesAsyncAndroid
...
     final SharedPreferencesAsyncApi api = _getApiForBackend(pigeonOptions);
-    return api.setStringList(key, value, pigeonOptions);
+    final String stringValue = '$jsonListPrefix${jsonEncode(value)}';
+    return api.setString(key, stringValue, pigeonOptions);
+  }
+
+  /// Adds a `List<String>` to preferences using platform encoding to test
+  /// moving from platform encoding to json encoding.
+  @visibleForTesting
+  Future<void> setStringListLegacyForTesting(
```
stuartmorgan, line 143: Generally the less test-only code in production code files the better; could we just expose `_getApiForBackend` for testing instead?

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_async_android.dart
```diff
@@ -189,10 +207,26 @@ base class SharedPreferencesAsyncAndroid
...
+        } catch (e) {
+          throw TypeError();
+        }
+      case const (List<Object?>):
+        return (dynamicStringList! as List<Object?>).cast<String>().toList();
+      case const (Null):
+        return null;
+      default:
+        throw TypeError();
+    }
```
stuartmorgan, line 229: You can use the new pattern system to avoid the need for force unwraps and casts here:

```dart
    switch (dynamicStringList) {
      case String s:
        final String jsonEncodedString =
            s.substring(jsonListPrefix.length);
        try {
          final List<String> decodedList =
              (jsonDecode(jsonEncodedString) as List<dynamic>).cast<String>();
          return decodedList;
        } catch (e) {
          throw TypeError();
        }
      case List<Object?> l:
        return l.cast<String>().toList();
      case null:
        return null;
      default:
        throw TypeError();
    }
```

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_async_android.dart
```diff
@@ -236,6 +270,16 @@ base class SharedPreferencesAsyncAndroid
       filter.allowList?.toList(),
       pigeonOptions,
     );
+    data.forEach((String? key, Object? value) {
+      if (value.runtimeType == String &&
+          (value! as String).startsWith(jsonListPrefix)) {
```
stuartmorgan, line 275: `if (value is String && value.startsWith(jsonListPrefix)`. (Using `is` will implicitly convert for the remainder of the block.)

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_async_android.dart
```diff
@@ -236,6 +270,16 @@ base class SharedPreferencesAsyncAndroid
...
+        data[key!] =
+            (jsonDecode((value as String).substring(jsonListPrefix.length))
```
stuartmorgan, line 277: You'll be able to replace `(value as String)` with just `value` here as well.

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_async_android.dart
```diff
@@ -236,6 +270,16 @@ base class SharedPreferencesAsyncAndroid
...
+                    as List<dynamic>)
+                .cast<String>()
+                .toList();
```
stuartmorgan, line 280: I don't think `toList()` should be necessary here.

packages/shared_preferences/shared_preferences_android/android/src/main/kotlin/io/flutter/plugins/sharedpreferences/MessagesAsync.g.kt
tarrinneal: There's a third less complex, somewhat stupid, but maybe better solution...

get String
decode
re-encode into json
send to dart
decode
profit

stuartmorgan: I thought of that, but then we have two different copies of the code to encode into JSON, in two different languages, which seemed potentially worse. I'm not sure how much Java code that is though; maybe it's so little it doesn't matter.

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_async_android.dart
tarrinneal: I'm not sure I understand this suggestion, are you saying I should add an option in that method that sets which version of setString is used?

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_async_android.dart
tarrinneal: I always forget about is

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
```diff
@@ -25,18 +25,21 @@
...
+import java.util.Objects;
 import java.util.Set;
 
 /** LegacySharedPreferencesPlugin */
 public class LegacySharedPreferencesPlugin implements FlutterPlugin, SharedPreferencesApi {
   private static final String TAG = "SharedPreferencesPlugin";
   private static final String SHARED_PREFERENCES_NAME = "FlutterSharedPreferences";
+  // All identifiers must match the SharedPreferencesPlugin.kt file, as well as the strings.dart file.
   private static final String LIST_IDENTIFIER = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu";
+  private static final String JSON_LIST_IDENTIFIER = LIST_IDENTIFIER + "!";
```
stuartmorgan, line 37: I think it's worth a comment that `!` is chosen as a character that can't be created by the base 64 encoding used with `LIST_IDENTIFIER`.

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
```diff
@@ -100,7 +103,15 @@ public void onDetachedFromEngine(@NonNull FlutterPlugin.FlutterPluginBinding bin
   }
 
   @Override
-  public @NonNull Boolean setStringList(@NonNull String key, @NonNull List<String> value)
+  public @NonNull Boolean setStringList(@NonNull String key, @NonNull String value)
```
stuartmorgan, line 106: Couldn't we eliminate this method from the Pigeon protocol and just change the `setString` check to `((value.startsWith(LIST_IDENTIFIER) && ! value.startsWith(JSON_LIST_IDENTIFIER) || ...`? Since the structure of the new design is that the Dart code is responsible for encoding it to just be a string, it seems odd to have a dedicated setter for that string.

packages/shared_preferences/shared_preferences_android/lib/src/shared_preferences_android.dart
```diff
@@ -47,7 +50,9 @@ class SharedPreferencesAndroid extends SharedPreferencesStorePlatform {
       case 'Double':
         return _api.setDouble(key, value as double);
       case 'StringList':
-        return _api.setStringList(key, value as List<String>);
+        return _api.setStringList(key, '$jsonListPrefix${jsonEncode(value)}');
+      case 'PlatformEncodedStringListForTesting':
```
stuartmorgan, line 54: This could also be eliminated by exposing a way for tests to get `_api`.

stuartmorgan (changes requested):

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
tarrinneal: we could, but then the method could be used to add a string with that json list identifier, which is the whole point of the check.

I could move the check for that one prefix to dart though, that is what you think is best.

stuartmorgan: Ah, right. I guess this is fine then.

Ideally what we would do is move all of the prefixing logic to Dart; the weirdness here is because we're now doing some special casing in native code, and some in Dart. But that's more refactoring than this PR needs.

packages/shared_preferences/shared_preferences_android/pigeons/messages.dart
```diff
@@ -37,7 +37,13 @@ abstract class SharedPreferencesApi {
 
   /// Adds property to shared preferences data set of type List<String>.
   @TaskQueue(type: TaskQueueType.serialBackgroundThread)
-  bool setStringList(String key, List<String> value);
+  bool setStringList(String key, String value);
```
stuartmorgan, line 40: Let's at least call this `setPreEncodedStringList` or something, so it's clear why it's called string list but takes a `String`.

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
tarrinneal: Yeah, I don't love it. Moving the checks without updating the error type seems silly.

stuartmorgan: I've done it in some other plugins, with a comment explaining why I'm throwing a weird `PlatformException` from Dart. It's a different kind of odd, but I don't think it's silly, since it allows us to make progress on moving logic to Dart, which in many cases makes our lives easier, without gating it behind breaking changes.

tarrinneal: I had considered it, I can add it to this pr if you want. I don't think it's a big deal

stuartmorgan (approved):
LGTM

packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java
stuartmorgan: Let's leave refactoring the existing logic to future work if we find that we are changing this enough to warrant the time.

tarrinneal: ok, for my understanding though. Is this error I wrote:
```
PlatformException(java.lang.RuntimeException: StorageError: This string cannot be stored as it
clashes with special identifier prefixes, RuntimeException, Cause: null, Stacktrace:
java.lang.RuntimeException: StorageError: This string cannot be stored as it clashes with special
identifier prefixes, null)
```
similar enough to 
```
PlatformException(java.lang.RuntimeException: StorageError: This string cannot be stored as it
clashes with special identifier prefixes, RuntimeException, Cause: null, Stacktrace:
java.lang.RuntimeException: StorageError: This string cannot be stored as it clashes with special
identifier prefixes
        at
io.flutter.plugins.sharedpreferences.LegacySharedPreferencesPlugin.setString(LegacySharedPreferencesPlugin.java:83)
        at
io.flutter.plugins.sharedpreferences.Messages$SharedPreferencesApi$-CC.lambda$setUp$2(Messages.java:204)
        at
io.flutter.plugins.sharedpreferences.Messages$SharedPreferencesApi$$ExternalSyntheticLambda2.onMessage(D8$$SyntheticClass:0)
        at
io.flutter.plugin.common.BasicMessageChannel$IncomingMessageHandler.onMessage(BasicMessageChannel.java:261)
        at io.flutter.embedding.engine.dart.DartMessenger.invokeHandler(DartMessenger.java:292)
        at
io.flutter.embedding.engine.dart.DartMessenger.lambda$dispatchMessageToQueue$0$io-flutter-embedding-engine-dart-DartMessenger(DartMessenger.java:319)
        at
io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)
        at io.flutter.embedding.engine.dart.DartMessenger$SerialTaskQueue.flush(DartMessenger.java:173)
        at
io.flutter.embedding.engine.dart.DartMessenger$SerialTaskQueue.lambda$dispatch$0$io-flutter-embedding-engine-dart-DartMessenger$SerialTaskQueue(DartMessenger.java:163)
        at
io.flutter.embedding.engine.dart.DartMessenger$SerialTaskQueue$$ExternalSyntheticLambda0.run(D8$$SyntheticClass:0)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
        at java.lang.Thread.run(Thread.java:1012)
, null)
```?

Comment (tarrinneal) [reply packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java]: I don't think I can replicate the entire stack trace reliably

Comment (stuartmorgan) [reply packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java]: Yes, if someone is relying on the format of the stack trace, they are going to have a bad time. The important things to preserve are:
- code, since that's the most likely thing for someone to match on, and
- message, which is debatable, and depends a lot on the code; some of our plugins use the same code for everything so people trying to distinguish have, for lack of a better option, done string matching on the message, so in those cases I try to preserve the message too.

Comment (tarrinneal) [reply packages/shared_preferences/shared_preferences_android/android/src/main/java/io/flutter/plugins/sharedpreferences/LegacySharedPreferencesPlugin.java]: Another important thing to note is that this error suffers from the pigeon error code swapping issue as well. (I couldn't find the filed issue to link)

**Comment by stuartmorgan** [comment]
It looks like this may have caused OOB failure in `shared_preferences` tests when it was published.

**Comment by camsim99** [comment]
This is causing `shared_preferences/shared_preferences/example/integration_test/shared_preferences_test.dart: shared_preferences_async Async throws TypeError when returned getStringList type is incorrect` to fail with what looks like an OOB issue and is blocking the tree currently:

```
The following TestFailure was thrown running a test (but after the test had completed):
Expected: throws <Instance of 'TypeError'>
  Actual: <Closure: () => Future<Null>>

#0      fail (package:matcher/src/expect/expect.dart:149:31)
#1      _expect.<anonymous closure> (package:matcher/src/expect/expect.dart:125:9)
<asynchronous suspension>
<asynchronous suspension>
(elided 6 frames from dart:async and package:stack_trace)
════════════════════════════════════════════════════════════════════════════════════════════════════

```

The plan is to revert.

PR #8341, opened 2024-12-23
[pigeon]: Support Swift async/await and Kotlin suspend methods

feduke-nukem:
- adds support of generating Swift `async/await` and Kotlin `suspend` methods in `HostApi`.
- adds `Async` annotation and `AsyncType.callback/AsyncType.await` to specify the style of generating async methods for Swift and Kotlin

Resolves [#123867](https://github.com/flutter/flutter/issues/123867), resolves [#147283](https://github.com/flutter/flutter/issues/147283)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/example/app/ios/Runner/AppDelegate.swift
```diff
@@ -25,6 +25,16 @@ private class PigeonApiImplementation: ExampleHostApi {
     }
     completion(.success(true))
   }
+
+  func sendMessageModernAsync(message: MessageData) async throws -> Bool {
```
hellohuanlin, line 29: you can just overload the function name `func sendMessage(...)`

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
```diff
@@ -25,6 +25,16 @@ private class PigeonApiImplementation: ExampleHostApi {
...
+    try? await Task.sleep(nanoseconds: 2_000_000_000)
```
hellohuanlin, line 30: is this used for unit test? we shouldn't wait for so long as it slows down the tests

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -222,6 +223,30 @@ class ExampleHostApiSetup {
...
+        "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.sendMessageModernAsync\(channelSuffix)",
+      binaryMessenger: binaryMessenger, codec: codec)
+    if let api = api {
+      sendMessageModernAsyncChannel.setMessageHandler { message, reply in
+        let args = message as! [Any?]
+        let messageArg = args[0] as! MessageData
+        Task {
+          do {
+            let result = try await api.sendMessageModernAsync(message: messageArg)
+            DispatchQueue.main.async {
```
hellohuanlin, line 237: you can do `Task { @MainActor }` here.

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
feduke-nukem: This is used in example app

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
feduke-nukem: I don't think that I understood.

It's generated Host API code. We could not overload functions in Dart even if we do in Swift

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -222,6 +223,30 @@ class ExampleHostApiSetup {
     } else {
       sendMessageChannel.setMessageHandler(nil)
     }
+    let sendMessageModernAsyncChannel = FlutterBasicMessageChannel(
```
hellohuanlin, line 226: can we just reuse the channel?

feduke-nukem: How?

hellohuanlin: can you point me to where this is used? im wondering if you can just use the original channel

feduke-nukem: What do you mean by "original channel"? Each method of Host API got generated `FlutterBasicMessageChannel`.

image

This is default Pigeon generated code flow. 

Maybe I've misunderstood you.

hellohuanlin: Maybe i misunderstood how pigeon is used. Just a strawman - can we share the same channel for both old and new ways of doing things?

packages/pigeon/example/app/ios/Runner/Messages.g.swift
feduke-nukem: > you can do `Task { @MainActor }` here.

Do you mean:

```swift
            sendMessageModernAsyncChannel.setMessageHandler { message, reply in
                let args = message as! [Any?]
                let messageArg = args[0] as! MessageData
                Task {
                    do { @MainActor
                        let result = try await api.sendMessageModernAsync(message: messageArg)
                        reply(wrapResult(result))

                    } catch {
                        reply(wrapError(error))
                    }
                }
            }
```

Will it lead to executing `api.sendMessageModernAsync` on main thread that could block it?

hellohuanlin: No, i meant you can replace `DispatchQueue.main.async {}` with `Task { @MainActor}`

packages/pigeon/example/app/ios/Runner/Messages.g.swift
feduke-nukem: The way we use channels will not differ at all. We just can not use two handlers for one channel at the same time.

packages/pigeon/example/app/ios/Runner/Messages.g.swift
feduke-nukem: > No, i meant you can replace `DispatchQueue.main.async {}` with `Task { @MainActor}`

What's the point/profit?

hellohuanlin: The purpose of swift concurrency is to replace GCD. So since you are using swift concurrency here, there's no reason to use GCD at all.

feduke-nukem: > GCD

Okay I get it.

I am not very familiar with swift concurrency actually. I've tried to do as you asked and this is not valid syntax:

image
 
Maybe I could use `MainActor.run` instead?

```swift
sendMessageModernAsyncChannel.setMessageHandler { message, reply in
    let args = message as! [Any?]
    let messageArg = args[0] as! MessageData
    Task {
        do {
            let result = try await api.sendMessageModernAsync(message: messageArg)

            MainActor.run(){
                reply(wrapResult(result))
            }

        } catch {
            MainActor.run(){
                reply(wrapError(error))
            }
        }
    }
}
```

hellohuanlin: In your first chunk of code, don't put `await` in front of `Task`. 

For this particular case, you can use `MainActor.run { ... }`, since `reply` doesn't require an async context.

feduke-nukem: @hellohuanlin done

feduke-nukem:
General question: should we allow to control generating `throws` in Swift async functions?

packages/pigeon/example/app/ios/Runner/Messages.g.swift
hellohuanlin: Can we have old api to be the wrapper of the new api?

feduke-nukem: Could you provide code sample?

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
feduke-nukem: Removed `Task.sleep`

feduke-nukem:
@hellohuanlin @LouiseHsu @tarrinneal CI check fails:

```
Changed packages: pigeon
[0:00] Running for packages/pigeon...
  No version change.
  Found NEXT; validating next version in the CHANGELOG.
No version change found, but the change to this package could not be verified to be exempt
from version changes according to repository policy.
If this is a false positive, please comment in the PR to explain why the PR
is exempt, and add (or ask your reviewer to add) the "override: no versioning needed" label.
```

Do I need to bump version in `CHANGELOG.md`? Currently I append related to pr changes below `NEXT` label.

packages/pigeon/example/app/ios/Runner/Messages.g.swift
hellohuanlin: nvm, i misunderstood what this code is doing.

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -222,6 +223,24 @@ class ExampleHostApiSetup {
...
+      name:
+        "dev.flutter.pigeon.pigeon_example_package.ExampleHostApi.sendMessageModernAsync\(channelSuffix)",
+      binaryMessenger: binaryMessenger, codec: codec)
+    if let api = api {
+      sendMessageModernAsyncChannel.setMessageHandler { message, reply in
+        let args = message as! [Any?]
+        let messageArg = args[0] as! MessageData
+        Task {
+          let result = await api.sendMessageModernAsync(message: messageArg)
+          await MainActor.run {
```
hellohuanlin, line 236: Why doesn't the old version switch thread? Is `setMessageHandler`'s callback guaranteed to be called on main thread?

feduke-nukem: The old version uses the callback approach that doesn't introduce asynchronous things like task so there was no need to switch to main to reply via channel

feduke-nukem: > Is `setMessageHandler`'s callback guaranteed to be called on main thread? 

I guess so as it is native platform channel way of communication

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
```diff
@@ -25,6 +25,10 @@ private class PigeonApiImplementation: ExampleHostApi {
     }
     completion(.success(true))
   }
+
+  func sendMessageModernAsync(message: MessageData) async -> Bool {
```
hellohuanlin, line 29: I just looked at the code again - it looks like the contract of pigeon is that the API (e.g. `sendMessage`) should be called on main thread. If that's the contract we want to preserve, we should mark the async API as main actor. 

```
@MainActor
func sendMessageModernAsync(message: MessageData) async -> Bool {}
```

packages/pigeon/example/app/ios/Runner/Messages.g.swift
hellohuanlin: After marking the func definition as `@MainActor` [discussed here](https://github.com/flutter/packages/pull/8341/files#r1898069753), here we can just do: 

```
Task { @MainActor in
  let result = await api.send...
  reply(wrapResult(result))
}
```

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
feduke-nukem: The only thing that should be called on the main thread is dispatching messages via channel

https://docs.flutter.dev/platform-integration/platform-channels#jumping-to-the-main-thread-in-ios

hellohuanlin: iiuc both way of communication needs to be on main thread. 

The link you add here is calling from platform to dart side. This is the opposite side communication, which needs to be on main as well.

feduke-nukem: Actually it is the same from the point of dispatching reply via channel, be it call from platform to dart or replying to call from dart

packages/pigeon/example/app/ios/Runner/Messages.g.swift
feduke-nukem: > After marking the func definition as `@MainActor` [discussed here](https://github.com/flutter/packages/pull/8341/files#r1898069753), here we can just do: 
> 
> ```
> Task { @MainActor in
>   let result = await api.send...
>   reply(wrapResult(result))
> }
> ```
> 

No problem it could be easily done.

But could you explain to me how it would work?

If we mark the whole function with @MainActor doesn't that mean that some expensive async/await function call could potentially be called at main?

hellohuanlin: > If we mark the whole function with @mainactor doesn't that mean that some expensive async/await function call could potentially be called at main?

The API contract is that this function will be called on main. The old API is also called on main and this does not change that. Marking it as `@MainActor` will allow compiler to enforce the implementor of this API to only use API that's safe to be used on main thread. That's one of the benefit of swift concurrency. 

An expensive aync/await call inside this function will be called on the thread specified by that call. For example, 
```
@MainActor
func sendMessageModernAsync() {
  let foo = await myActor.getExpensiveFoo()
}
```

Here `getExpensiveFoo` will be called under `myActor`'s async context, rather than that of the main actor.

feduke-nukem: Done

packages/pigeon/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 22.7.1
```
tarrinneal, line 1: You'll need to add this change to the pubspec.yaml file, and the generator_tools.dart file.

tarrinneal (changes requested):
I'll get to an in depth review as soon as I can. Just wanted to answer your question about failing checks.

Thank you for putting the effort into this pr.

packages/pigeon/CHANGELOG.md
feduke-nukem: Done

tarrinneal (changes requested):
A couple larger changes/ questions to answer for now. Again, when I have more time I will dive in further. Just wanted to give you a chance to respond/change things as requested.

packages/pigeon/lib/pigeon_lib.dart
```diff
@@ -41,6 +41,62 @@ class _Asynchronous {
...
+  /// ```
+  ///
+  /// If `false`:
+  ///
+  /// ```swift
+  /// func sendMessage(message: MessageData) async -> Bool
+  /// ```
+  ///
+  /// The default value is `true`.
+  final bool? isSwiftThrows;
```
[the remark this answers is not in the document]
feduke-nukem: Just an ability to choose. All in all, the protocol is generated, so it is nice to have a strict contract that will not "throw" any exception at all if it is desired.

packages/pigeon/lib/pigeon_lib.dart
```diff
@@ -41,6 +41,62 @@ class _Asynchronous {
   const _Asynchronous();
 }
 
+/// Metadata to annotate a Api method as asynchronous
+const Object async = _Asynchronous();
+
+/// {@macro pigeon_lib.modern_async}
+class ModernAsync {
```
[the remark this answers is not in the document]
feduke-nukem: I was thinking about that but the concern was specifying `isSwiftThrows` that you have already mentioned in discussion below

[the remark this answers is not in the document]
feduke-nukem: Have done some changes to combine async metas @tarrinneal

packages/pigeon/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
...
 
+* [swift, kotlin] Added support of modern asynchronous api for Swift (async) and Kotlin (suspend).
```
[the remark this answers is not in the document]
feduke-nukem: Done

packages/pigeon/lib/pigeon_lib.dart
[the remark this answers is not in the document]
hellohuanlin: or "structured concurrency" (apple's term)

tarrinneal (changes requested):
There's a section in the top level README titled 

### Synchronous and Asynchronous methods

That should be updated to include the new options.

This pr will also need integration tests. check out the `packages/pigeon/platform_tests` folder for current tests. and `packages/pigeon/pigeons/core_tests.dart` for integration tests pigeon definition file.

I think a single method that sends `AllTypes` and one that sends `AllNullableTypes` should be sufficient.

packages/pigeon/example/README.md
```diff
@@ -140,6 +153,10 @@ private class PigeonApiImplementation: ExampleHostApi {
     }
     completion(.success(true))
   }
+
+  func sendMessageModernAsync(message: MessageData) async -> Bool {
+    return true
+  }
```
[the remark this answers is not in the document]
feduke-nukem: Done

packages/pigeon/example/README.md
```diff
@@ -76,6 +76,9 @@ abstract class ExampleHostApi {
 
   @async
   bool sendMessage(MessageData message);
+
+  @Async(type: AsyncType.await(isSwiftThrows: false))
+  bool sendMessageModernAsync(MessageData message);
```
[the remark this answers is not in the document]
feduke-nukem: Done

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 22.8.0
+
+* [swift, kotlin] Added support of modern asynchronous api for Swift (async) and Kotlin (suspend).
```
hellohuanlin, line 3: i'd just call it "swift concurrency and kotlin coroutines"

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
hellohuanlin: it looks like this isn't addressed yet?

packages/pigeon/example/app/macos/Runner/MainFlutterWindow.swift
```diff
@@ -24,6 +24,12 @@ private class PigeonApiImplementation: ExampleHostApi {
     }
     completion(.success(true))
   }
+
+  /// Unlike implementations on other platforms, this function does not throw any exceptions
+  /// because the `@Async(type: AsyncType.await(isSwiftThrows: false))` annotation was specified.
```
hellohuanlin, line 29: is there an example where isSwiftThrows = true?

packages/pigeon/pigeons/core_tests.dart
```diff
@@ -723,13 +723,26 @@ abstract class HostIntegrationCoreApi {
...
+
   /// Returns the passed object, to test serialization and deserialization.
   @async
   @ObjCSelector('echoAsyncNullableAllNullableTypes:')
   @SwiftFunction('echoAsync(_:)')
   AllNullableTypes? echoAsyncNullableAllNullableTypes(
       AllNullableTypes? everything);
 
+  /// Returns the passed object, to test async serialization and deserialization using `await`-style.
+  @Async(type: AsyncType.await(isSwiftThrows: false))
```
tarrinneal, line 740: Lets add 2 more, both that have `isSwiftThrows` true, one of which throws

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
tarrinneal: Not all pigeon calls are on the main thread.

packages/pigeon/example/app/macos/Runner/MainFlutterWindow.swift
feduke-nukem: Not yet. Will add.

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
feduke-nukem:
```swift
Task { @MainActor in
  let result = await api.send...
  reply(wrapResult(result))
}
```

Isn't that enough within the generated code?

packages/pigeon/CHANGELOG.md
feduke-nukem: Changed to "[swift, kotlin] Added support of swift concurrency and kotlin coroutines."

packages/pigeon/pigeons/core_tests.dart
feduke-nukem: Added for `AllTypes` two methods.

But facing error at windows step:

```
C:\b\s\w\ir\x\w\packages\packages\pigeon\platform_tests\test_plugin\example\windows\flutter\ephemeral\.plugin_symlinks\test_plugin\windows\pigeon\core_tests.gen.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj [C:\b\s\w\ir\x\w\packages\packages\pigeon\platform_tests\test_plugin\example\build\windows\x64\plugins\test_plugin\test_plugin_plugin.vcxproj]
C:\b\s\w\ir\x\w\packages\packages\pigeon\platform_tests\test_plugin\example\windows\flutter\ephemeral\.plugin_symlinks\test_plugin\windows\pigeon\core_tests.gen.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj [C:\b\s\w\ir\x\w\packages\packages\pigeon\platform_tests\test_plugin\example\build\windows\x64\plugins\test_plugin\test_plugin_test.vcxproj]
```

packages/pigeon/example/app/macos/Runner/MainFlutterWindow.swift
feduke-nukem: Have added example

packages/pigeon/pigeons/core_tests.dart
tarrinneal: This may also just be a ci issue

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
hellohuanlin: > Not all pigeon calls are on the main thread. 

I didn't know this. If this is the case, you should add tests for both main and non-main usage. 

> According to [implementation](https://api.flutter.dev/ios-embedder/classflutter_1_1_platform_message_handler_ios.html#:~:text=72-,//%20Called%20from%20any%20thread.,-73%C2%A0%20%20%20%20%20%20%20%20%20%20if) code should be sufficient:
... 

This code calls `api.send...` on the main actor. However, `api.send...` itself is non-isolated, which means you will get warning by accessing main-actor only APIs (e.g. most UIKit APIs). 

> We only care about dispatching the result reply to the Flutter side on the main thread, don't we?

What I care more about is the API - if we are guaranteed to call this in main actor, the api should be marked as `@MainActor`. On the other hand, if we also call it in non-main, you need to add tests for both main and non-main usage.

tarrinneal: Our test coverage for `@TaskQueue` isn't very high, in fact there are no tests in the core tests integration tests. This is a prime example of something that could be breaking in that area.

packages/pigeon/pigeons/core_tests.dart
stuartmorgan: I can't repro locally, so it may be toolchain-version-specific. Adding that flag to platform_tests/windows/CMakeLists.txt to see if it make CI compilation happy is worth a try.

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
hellohuanlin: now is always a good time to add tests! especially since we are here :)

stuartmorgan: > > Not all pigeon calls are on the main thread.
> 
> I didn't know this.

Pigeon has the same threading semantics as the method channel APIs it wraps:
- Native handlers for Dart -> native calls (`HostApi` in Pigeon) are on the main thread by default, or some thread pool thread if the channel was configured to use task queues.
- The reply to an `@async` Dart -> native call (`FlutterApi') can (as of about a year ago, IIRC) be invoked on any native thread.
- Native -> Dart calls (`FlutterApi` in Pigeon) must still originate on the main thread.

feduke-nukem: Well.

Explain to me please where I should add tests and what kind of.

There are a lot of them and I am a bit lost already 😀

hellohuanlin: IC. since this callback is not guaranteed to be called on main, then it shouldn't be marked as `@MainActor`.

hellohuanlin: > Explain to me please where I should add tests and what kind of.

I'd like to see the case when the callback is called on background thread.

feduke-nukem: > IC. since this callback is not guaranteed to be called on main, then it shouldn't be marked as `@MainActor`. 

We have discussed it previously but I want to clarify it for myself.

Doesn't the:

`await MainActor.run(){ reply(wrapResult(result)) }`

Solve the problem?

Or the first version: `DispatchQueue.main.async...`?

hellohuanlin: My previous recommendation was based on the assumption that the callback is always on main. Since you are interested to learn why, I will explain it anyways even though the assumption is false. 

> Doesn't the: `await MainActor.run(){ reply(wrapResult(result)) }` Solve the problem?

No, because without the annotation, compiler doesn't know that this callback is on main. For example, let's say a plugin author writes this: 

```
func sendFoo() async {
  // Can't use UIKit API
  let view = UIView()
}
```
In the above code, even though `sendFoo()` is always called within main actor, the compiler doesn't know about it, so it will give you error in swift 6. They would have to switch to main by themselves. 

If you annotate `sendFoo()` with `@MainActor`, it tells the compiler that `sendFoo` should only be called on main actor (otherwise will error out). So that developers can simply write: 

```
@MainActor func sendFoo() async {
  // ok to call
  let view = UIView()
}
```

Again, the above reasoning is based on the assumption that the callback is always on main, which is false (thanks @tarrinneal and @stuartmorgan for pointing it out). 

The reason I want to see a non-main thread test case is that, I don't think it is possible to have a granular control of which queue to run in swift concurrency (expect for main vs non-main). Swift concurrency will only spin up the same number of queues as the number of CPU cores. I suspect the best you can do is to just distinguish between main vs non-main, by doing something like this: 

```
// 1
Task { // 2
  let result = await api.send...
  Task { @MainActor // 3
    reply(wrapResult(result))
  }
}
```

Notes for the above 1/2/3 comments: 
1. This is in method channel callback. It can be either main thread or a particular queue specified by `@TaskQueue` (say `Q1`)
2. `Task` creates an async context. Here we don't specify `@MainActor`, so if (1) is on main, (2) will be on main; if (1) is on `Q_1`, (2) will be on a "cooperative" queue managed by Swift (not `Q1`!). 
3. Another `Task` to switch back to `@MainActor`, so that `reply` is called on main.

stuartmorgan: > Again, the above reasoning is based on the assumption that the callback is always on main, which is false (thanks @tarrinneal and @stuartmorgan for pointing it out).

We do know, at generation time, whether any given method call was annotated as a task queue method, so if it's helpful to annotate Host API handlers that are *not* marked as task queue methods as `@MainActor` we could do that.

In practice, any plugin developer who wants to write code like your:
```swift
func sendFoo() async {
  // Can't use UIKit API
  let view = UIView()
}
```

is not going to have opted `sendFoo` into task queues, because doing so would be counterproductive.

feduke-nukem: All in all, do we have any final decision about what have to be done for this case? @stuartmorgan @hellohuanlin @tarrinneal 

I am feeling that the discussion still continues, doesn't it?

stuartmorgan: I think this probably needs a short design document to clearly establish what the goals are and how each method channel feature maps to the actor system. The comment threads is this PR seem like clear indications that there is not yet a clear consensus on what the PR is supposed to be doing, and a design doc would be a better format for resolving that.

feduke-nukem: The goal is to be able to generate methods that can use `async/await` for Swift and `suspend/coroutines` for Kotlin instead of callbacks using existing `async` or whatever other annotation.

That's it.

hellohuanlin: > We do know, at generation time, whether any given method call was annotated as a task queue method

Then we should annotate the api with @MainActor for main cases, and keep it non-isolated for non-main cases. 

> I think this probably needs a short design document

Agree +1

feduke-nukem: Could you please tell me where I could see examples of design docs related to Pigeon library?

hellohuanlin: It can be just a short one about this part we are discussing here.

hellohuanlin: If you like I can help write up a small doc since I'm familiar with this topic and what's the end state look like. But I won't have bandwidth actually implementing it.

feduke-nukem: > If you like I can help write up a small doc since I'm familiar with this topic and what's the end state look like. But I won't have bandwidth actually implementing it. 

That would be great

hellohuanlin: ok i should be able to find some time next week or the week after. in the mean time, i highly recommend looking into how to add tests for background thread usage. otherwise we won't feel confident enough for this change.

packages/pigeon/pigeons/core_tests.dart
feduke-nukem: Seems like adding into `platform_tests/test_plugin/windows/CMakeLists.txt`:

```cmake
if(MSVC)
  add_compile_options("/bigobj")
endif()
```

Helped

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
feduke-nukem: I’ve been exploring the `TaskQueue` annotation and noticed a difference in behavior: when the annotation is used, the `swift_generator` doesn’t add any `binaryMessenger.makeBackgroundTaskQueue`, whereas the `objc_generator` does.

I’m curious about the reasoning behind this difference. Could you help clarify @stuartmorgan @hellohuanlin @tarrinneal ?

Thank you!

packages/pigeon/platform_tests/test_plugin/windows/CMakeLists.txt
```diff
@@ -4,6 +4,12 @@
 # customers of the plugin.
 cmake_minimum_required(VERSION 3.14)
 
+# Due to "number of sections exceeded object file format limit: compile with /bigobj" error
+if(MSVC)
```
stuartmorgan, line 8: This doesn't need to be conditionalized; we don't support other compilers for Windows.

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
feduke-nukem: > I would expect that would be an easy fix.

Is this meant to be a separate pull request?

packages/pigeon/platform_tests/test_plugin/windows/CMakeLists.txt
feduke-nukem: Have removed the condition.

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
stuartmorgan: A prequel PR might be cleaner, but it's since it's closely related to this PR and should be small, including it in this PR would probably also be reasonable unless @tarrinneal would prefer it separate.

packages/pigeon/lib/kotlin_generator.dart
```diff
@@ -1300,9 +1316,21 @@ if (wrapped == null) {
...
+            if (asynchronousType.isAwait) {
+              indent.dec();
+              indent.writeln('}');
+            }
           }
         });
+        if (asynchronousType.isAwait) {
+          indent.dec();
+          indent.writeln('}');
+        }
```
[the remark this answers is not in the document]
tarrinneal: Just a not for myself, you don't need to change anything.

tarrinneal (changes requested):
Just realized I had unpublished review comments. Still not a complete review though.

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
feduke-nukem: I prefer separate PRs.

But what would the scenario be in such a case?

1. I create a new PR.
2. This PR waits for the new one.
3. As soon as the new PR is approved or merged, I merge the changes into this PR.

Or are there any other steps?

packages/pigeon/example/README.md
```diff
@@ -113,6 +123,26 @@ Future<bool> sendMessage(String messageText) {
...
+
+Future<bool> sendMessageModernAsyncAndThrow(String messageText) {
+  final MessageData message = MessageData(
+    code: Code.two,
+    data: <String, String>{'header': 'this is a header'},
+    description: 'uri text',
+  );
+
+  return _api.sendMessageModernAsyncThrows(message);
+}
```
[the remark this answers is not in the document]
feduke-nukem: Could you clarify how that should be done? The `tool/generate.dart` generates code for all supported languages and platforms.

packages/pigeon/example/README.md
```diff
@@ -140,6 +170,16 @@ private class PigeonApiImplementation: ExampleHostApi {
...
+
+  /// Unlike implementations on other platforms, this function does not throw any exceptions
+  /// because the `@Async(type: AsyncType.await(isSwiftThrows: false))` annotation was specified.
+  func sendMessageModernAsync(message: MessageData) async -> Bool {
+    return true
+  }
+
+  func sendMessageModernAsyncThrows(message: MessageData) async throws -> Bool {
+    throw PigeonError(code: "code", message: "message", details: "details")
+  }
```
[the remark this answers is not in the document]
feduke-nukem: That's a great idea!

What do you think about also simplifying the core_tests methods by reducing them to one?

```dart
  /// Returns the passed object, to test async serialization and deserialization using `await`-style
  /// and Swift can throw an exception.
  @Async(type: AsyncType.await(isSwiftThrows: true))
  @ObjCSelector('echoModernAsyncAllTypesAndNotThrow:')
  @SwiftFunction('echoModernAsyncAllTypesAndNotThrow(_:)')
  AllTypes echoModernAsyncAllTypesAndNotThrow(AllTypes everything);

  /// Returns the passed object, to test async serialization and deserialization using `await`-style
  /// and throws an exception.
  @Async(type: AsyncType.await(isSwiftThrows: true))
  @ObjCSelector('echoModernAsyncAllTypesAndThrow:')
  @SwiftFunction('echoModernAsyncAllTypesAndThrow(_:)')
  AllTypes echoModernAsyncAllTypesAndThrow(AllTypes everything);
  ```
 to a single `AllTypes echoModernAsyncAllTypesAndThrow(AllTypes everything);` 
 
This way, we could handle both cases in the tests with the condition you mentioned earlier. Would that work for you?

File: packages/pigeon/README.md [hunk 20]
```diff
@@ -42,7 +42,11 @@ pigeon methods can be written on the native side as synchronous methods,
 to make it simpler to always reply exactly once.
 
 If asynchronous methods are needed, the `@async` annotation can be used. This will require
-results or errors to be returned via a provided callback. [Example](./example/README.md#HostApi_Example).
+results or errors to be returned via a provided callback. [Example](./example/README.md#HostApi_Example). 
+
+It is possible to specify the type of approach to be applied, such as await-style using the @Async(type: AsyncType.await()) annotation, or callback-style using the @Async(type: AsyncType.callback()) annotation. Additionally, can be defined whether the Swift function will throw an exception by using the @Async(type: AsyncType.await(isSwiftThrows: false)) annotation. [Example](./example/README.md#HostApi_Example).
```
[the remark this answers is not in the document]
feduke-nukem: Done

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 22.8.0
+
+* [swift, kotlin] Added support of swift concurrency and kotlin coroutines.
```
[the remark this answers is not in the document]
feduke-nukem: Done

packages/pigeon/README.md
```diff
@@ -42,7 +42,11 @@ pigeon methods can be written on the native side as synchronous methods,
...
+
+Currently, `await`-style functionality is only available for Swift (async) and Kotlin (suspend).
```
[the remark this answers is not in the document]
feduke-nukem: Done

packages/pigeon/lib/ast.dart
```diff
@@ -23,6 +23,66 @@ enum ApiLocation {
...
+  bool get isNone => this is NoAsynchronous;
+}
+
+/// {@macro pigeon_lib.callback_async_type}
+class CallbackAsynchronous extends AsynchronousType {
+  /// Constructor for [CallbackAsynchronous].
+  const CallbackAsynchronous();
+}
+
+/// Represents an await-style asynchronous api will be used.
```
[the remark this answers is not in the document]
feduke-nukem: Done

packages/pigeon/example/app/ios/Runner/AppDelegate.swift
tarrinneal: That's it.

packages/pigeon/example/README.md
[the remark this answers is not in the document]
tarrinneal: no :( We need to handle both generated code paths, even though they are mostly redundant.

packages/pigeon/example/README.md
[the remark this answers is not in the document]
tarrinneal: I just mean they don't need to be included in the README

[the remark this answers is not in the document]
feduke-nukem: > I just mean they don't need to be included in the README

This README is generated with [script](https://github.com/flutter/packages/blob/main/script/tool/README.md#update-readmemd-from-example-sources) based on the code from example app.

I am not sure what should be done then.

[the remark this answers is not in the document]
feduke-nukem: Looks like that it should be isolated `HostApi` and excluded from being generated for unsupported platforms like `EventChannelTests` and `EventChannelApi` was done.

It will significantly reduce the amount of useless code (only Kotlin and Swift is supported)

packages/pigeon/example/app/windows/runner/flutter_window.cpp
```diff
@@ -41,6 +41,21 @@ class PigeonApiImplementation : public ExampleHostApi {
...
+    result(true);
+  }
+
+  void SendMessageModernAsyncThrows(
+      const MessageData& message,
+      std::function<void(ErrorOr<bool> reply)> result) {
+    result(FlutterError("code", "message", "details"));
+  }
 };
 // #enddocregion cpp-class
```
tarrinneal, line 60: This comment tells the script where to end the snippet. You can start and stop a snippet multiple times to add a `...` and skip lines you don't want included as well. So here I would end the snippet after `SendMessageModernAsync` then start it again after `SendMessageModernAsyncThrows` and leave the end comment where it is. 

Let me know if that's too confusing of an explanation.

packages/pigeon/example/README.md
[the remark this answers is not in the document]
tarrinneal: I left another comment with instructions, I don't think it's worth adding a new pigeon file for this. We can leave the code in the other language example files, it just needs to be removed from the README

packages/pigeon/example/app/windows/runner/flutter_window.cpp
feduke-nukem: > This comment tells the script where to end the snippet. You can start and stop a snippet multiple times to add a `...` and skip lines you don't want included as well. So here I would end the snippet after `SendMessageModernAsync` then start it again after `SendMessageModernAsyncThrows` and leave the end comment where it is.

Did I get it correct?

```dart
// #docregion host-definitions
@HostApi()
abstract class ExampleHostApi {
  String getHostLanguage();

  // These annotations create more idiomatic naming of methods in Objc and Swift.
  @ObjCSelector('addNumber:toNumber:')
  @SwiftFunction('add(_:to:)')
  int add(int a, int b);

  @async
  bool sendMessage(MessageData message);

  // #enddocregion host-definitions
  bool unwantedMethodInReadme();

  bool anotherUnwantedMethodInReadme();
  // #docregion host-definitions
}
// #enddocregion host-definitions
```

Will produce README:

```dart
@HostApi()
abstract class ExampleHostApi {
  String getHostLanguage();

  // These annotations create more idiomatic naming of methods in Objc and Swift.
  @ObjCSelector('addNumber:toNumber:')
  @SwiftFunction('add(_:to:)')
  int add(int a, int b);

  @async
  bool sendMessage(MessageData message);

  // ···
}
```

tarrinneal: yup

packages/pigeon/example/README.md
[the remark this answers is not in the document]
feduke-nukem: Have removed for Linux and Windows

packages/pigeon/example/README.md
```diff
@@ -231,10 +298,39 @@ static void handle_send_message(
...
+}
+
+static void handle_send_message_modern_async_throws(
+    PigeonExamplePackageMessageData* message,
+    PigeonExamplePackageExampleHostApiResponseHandle* response_handle,
+    gpointer user_data) {
+  g_autoptr(FlValue) details = fl_value_new_string("details");
+  pigeon_example_package_example_host_api_respond_error_send_message_modern_async_throws(
+      response_handle, "code", "message", details);
+}
```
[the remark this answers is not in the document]
feduke-nukem: done

packages/pigeon/example/README.md
```diff
@@ -231,10 +298,39 @@ static void handle_send_message(
...
+
 static PigeonExamplePackageExampleHostApiVTable example_host_api_vtable = {
     .get_host_language = handle_get_host_language,
     .add = handle_add,
-    .send_message = handle_send_message};
+    .send_message = handle_send_message,
+    .send_message_modern_async = handle_send_message_modern_async,
+    .send_message_modern_async_throws =
+        handle_send_message_modern_async_throws};
```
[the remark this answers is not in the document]
feduke-nukem: done

packages/pigeon/example/app/windows/runner/flutter_window.cpp
feduke-nukem: done

packages/pigeon/example/README.md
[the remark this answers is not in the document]
feduke-nukem: Done

tarrinneal:
You don't need to merge main unless there is a conflict that needs resolving. It actually makes it harder to review (and sends people an email every time).

feduke-nukem:
> You don't need to merge main unless there is a conflict that needs resolving. It actually makes it harder to review (and sends people an email every time).

Whoops, my bad.

Didn't mean to do it like that. It won't happen again.

tarrinneal:
It's not a big deal, just thought I'd let you know

PR #8346, opened 2024-12-26
[video_player]: reduce video player position update interval from 500ms to 100ms

abdelaziz-mahdy:
Description
reduce video player position update interval from 500ms to 100ms to fix the issue of fast captions not showing for their intended time

https://github.com/flutter/flutter/issues/160855
## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
Per discussion linked from https://github.com/flutter/flutter/issues/90080#issuecomment-1971463891 I'm skeptical that this is API we want to add. If subtitles don't work reliably at the current update frequency, and higher update frequencies don't have noticeable performance impact, then we should simply increase the frequency. If it does have performance issues, then we should figure out a different way of triggering subtitle changes. Either way, I would expect that we should be able to resolve the issue without new public API and without requiring extra configuration.

abdelaziz-mahdy:
I kept searching for the previous discussions but couldn't find them. Thank you for providing the link.

Would an update frequency of 100ms work? I think it would be very effective for subtitles and not frequent enough to cause a high load on devices.

if you agree, i can open a new pr with that change and add a test to make sure captions are included in the timer consideration

abdelaziz-mahdy:
note: this is how i override it from the app side, i didnt like the approach this is why i opened this pr
```
  void startPositionUpdaterForcer() {
    _positionUpdaterForcerTimer?.cancel();
    _positionUpdaterForcerTimer =
        Timer.periodic(Duration(milliseconds: 100), (timer) {
      _positionUpdaterForcer();
    });
  }

  void stopPositionUpdaterForcer() {
    _positionUpdaterForcerTimer?.cancel();
  }
  /// These parts exists in the video_player package but slightly edited to be called from outside of the class
  Future<void> _positionUpdaterForcer() async {
    Duration? position = await controller.position;
    if (position == null) {
      return;
    }
    VideoPlayerValue value = controller.value;
    if (position > value.duration) {
      position = value.duration;
    }
    value = value.copyWith(
      position: position,
      caption: await _getCaptionAt(position),
      isCompleted: position == value.duration,
    );
  }

  Future<Caption> _getCaptionAt(Duration position) async {
    /// the internal [ClosedCaptionFile]
    ClosedCaptionFile? closedCaptionFile;
    closedCaptionFile = await controller.closedCaptionFile;
    if (closedCaptionFile == null) {
      return Caption.none;
    }

    VideoPlayerValue value = controller.value;
    final Duration delayedPosition = position + value.captionOffset;
    // TODO(johnsonmh): This would be more efficient as a binary search.
    for (final Caption caption in (closedCaptionFile).captions) {
      if (caption.start <= delayedPosition && caption.end >= delayedPosition) {
        return caption;
      }
    }

    return Caption.none;
  }
```

abdelaziz-mahdy:
i updated the pr to be an update to value from 500 to 100 ms, and the test case. hope thats okay

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## NEXT
+## 2.9.3
 
 * Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+* Added `setVideoPositionUpdatesInterval` to allow setting position updates periodic intervals
```
stuartmorgan, line 4: Adds

https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style

stuartmorgan: This is also missing a period, also per the style guide.

stuartmorgan: And it needs to be rewritten to reflect what the change now does; this is describing the previous iteration of the PR.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -956,6 +956,41 @@ void main() {
       });
     });
   });
+  test('updates position', () async {
```
stuartmorgan, line 959: There should be a blank line between tests/groups.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -956,6 +956,41 @@ void main() {
...
+    final VideoPlayerController controller = VideoPlayerController.networkUrl(
+      _localhostUri,
+      videoPlayerOptions: VideoPlayerOptions(),
+    );
+
+    await controller.initialize();
+
+    // Set a custom interval
```
stuartmorgan, line 967: This no longer makes sense.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -956,6 +956,41 @@ void main() {
...
+      await Future<void>.delayed(customInterval);
+      fakeVideoPlayerPlatform._positions[controller.textureId] =
+          Duration(milliseconds: i * customInterval.inMilliseconds);
+    }
+
+    // Wait for at least 3 position updates
+    await intervalUpdateCompleter.future.timeout(const Duration(seconds: 5));
+
+    // Verify that the intervals between updates are approximately correct
+    expect(positions[1] - positions[0], greaterThanOrEqualTo(customInterval));
```
stuartmorgan, line 991: Wouldn't this test as written pass without the rest of the PR?

stuartmorgan (changes requested):

packages/video_player/video_player/CHANGELOG.md
abdelaziz-mahdy: thanks for the link, it should fit now

packages/video_player/video_player/test/video_player_test.dart
abdelaziz-mahdy: just tested using 500 ms for the updates and it failed, but i will add another test related to the captions since its the more related to this pr

abdelaziz-mahdy: I added the other test and I hope its shows the problem more, let me know if there anything else I can fix.

Thank you.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -957,6 +996,41 @@ void main() {
...
+    });
+    await controller.play();
+    for (int i = 0; i < 3; i++) {
+      await Future<void>.delayed(updatesInterval);
+      fakeVideoPlayerPlatform._positions[controller.textureId] =
+          Duration(milliseconds: i * updatesInterval.inMilliseconds);
+    }
+
+    // Wait for at least 3 position updates
+    await intervalUpdateCompleter.future.timeout(const Duration(seconds: 5));
```
stuartmorgan, line 1027: Tests should not include timeouts, per Flutter style. This should just await the future.

stuartmorgan (approved):
LGTM with one change. @tarrinneal for second review.

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## NEXT
+## 2.9.3
 
 * Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+* Reduces the update interval from 500ms to 100ms.
```
tarrinneal, line 4: I think this should be clearer about what update interval is changing

tarrinneal (approved):
lgtm, with some clarity in the changelog.

packages/video_player/video_player/CHANGELOG.md
abdelaziz-mahdy: thank you, I hope its more clear now.

stuartmorgan:
This is hitting the known camerax flake; it's unrelated to the PR.

PR #8347, opened 2024-12-26
[video_player] Optimize caption retrieval with binary search in VideoPlayerController

abdelaziz-mahdy:
Just found the todo in the code and it was from 3 years ago, so i wanted to implement it since it may help people using captions to have more optimized code

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player/CHANGELOG.md
+++ b/packages/video_player/video_player/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## NEXT
 
 * Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+* Use Binary search for finding the correct caption
 
 ## 2.9.2
 
--- a/packages/video_player/video_player/lib/video_player.dart
+++ b/packages/video_player/video_player/lib/video_player.dart
@@ -728,20 +728,36 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   ///
   /// If no [closedCaptionFile] was specified, this will always return an empty
   /// [Caption].
+
   Caption _getCaptionAt(Duration position) {
     if (_closedCaptionFile == null) {
       return Caption.none;
     }
 
     final Duration delayedPosition = position + value.captionOffset;
-    // TODO(johnsonmh): This would be more efficient as a binary search.
-    for (final Caption caption in _closedCaptionFile!.captions) {
-      if (caption.start <= delayedPosition && caption.end >= delayedPosition) {
-        return caption;
+
+    final List<Caption> captions = _closedCaptionFile!.captions;
+
+    int left = 0;
+    int right = captions.length - 1;
+
+    while (left <= right) {
+      final int mid = left + ((right - left) ~/ 2);
+      final Caption midCaption = captions[mid];
+
+      if (midCaption.start <= delayedPosition &&
+          midCaption.end >= delayedPosition) {
+        return midCaption; // Found the matching caption
+      } else if (midCaption.end < delayedPosition) {
+        // Move to the right half
+        left = mid + 1;
+      } else {
+        // Move to the left half
+        right = mid - 1;
       }
     }
 
-    return Caption.none;
+    return Caption.none; // No matching caption found
   }
 
   /// Returns the file containing closed captions for the video, if any.


packages/video_player/video_player/lib/video_player.dart
```diff
@@ -728,20 +728,36 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
-    for (final Caption caption in _closedCaptionFile!.captions) {
-      if (caption.start <= delayedPosition && caption.end >= delayedPosition) {
-        return caption;
+
+    final List<Caption> captions = _closedCaptionFile!.captions;
+
+    int left = 0;
+    int right = captions.length - 1;
+
+    while (left <= right) {
```
jmagman, line 744: https://github.com/flutter/plugins/pull/2487#discussion_r376564502

> Not a big deal either way, but for what it's worth I think you could trivially import this. https://api.flutter.dev/flutter/package-collection_collection/binarySearch.html

> Yes, it wouldn't be too hard, although we would have to either sort the captions at initialization, or make it clear that we assume a duration sorted list of captions (neither of which we currently do).

abdelaziz-mahdy: I implemented all the mentioned changes. However, to use the collection function, I had to convert to a list, so I cached it to avoid re-mapping on each check.

tarrinneal: I think what @jmagman is suggesting is that there is a package should be used to do the binary search.

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## NEXT
```
tarrinneal, line 1: This will need a version change to 2.9.3, as well as in the pubspec file

tarrinneal (changes requested):

tarrinneal:
@stuartmorgan does this change need a test? It shouldn't be changing any behaviors, I'm not sure how a test would work honestly.

stuartmorgan:
> @stuartmorgan does this change need a test? It shouldn't be changing any behaviors, I'm not sure how a test would work honestly.

The fact that it's a more efficient search doesn't inherently need a test; that's essentially just a refactoring. However, this seems like a good opportunity to make sure we have reasonably robust coverage of caption correctness. E.g., is there a unit test that would fail if the initial sort were commented out?

abdelaziz-mahdy:
> > @stuartmorgan does this change need a test? It shouldn't be changing any behaviors, I'm not sure how a test would work honestly.
> 
> The fact that it's a more efficient search doesn't inherently need a test; that's essentially just a refactoring. However, this seems like a good opportunity to make sure we have reasonably robust coverage of caption correctness. E.g., is there a unit test that would fail if the initial sort were commented out?

I can add that , so in general more caption edge cases

abdelaziz-mahdy:
tests implemented and improved the logic for the binary search since it was incorrect in some cases

for this test `test('captions are sorted correctly on initialization'`
this test can be ignored and if thats the case we can remove the new public member to the class (my main issue is that i was not able to expose a private member only for testing) if someone can point me to the right direction that will help a lot

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -378,6 +379,10 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
 
   Future<ClosedCaptionFile>? _closedCaptionFileFuture;
   ClosedCaptionFile? _closedCaptionFile;
+  List<Caption>? _sortedCaptions;
+
+  /// The sorted list of captions from the closed caption file.
+  List<Caption>? get sortedCaptions => _sortedCaptions;
```
stuartmorgan, line 385: Why is new public API needed? I don't see any implementation code using this.

abdelaziz-mahdy: To be able to access that list in testing, as I mentioned in the last comment I couldn't make it private and available for testing, 

If the test is not needed I can remove it with the test

abdelaziz-mahdy: i removed it and the test.

stuartmorgan: If something needs to be visible only for testing, it should be annotated `@visibleForTesting`.

It's not clear to me why even that would be necessary though; API to get the current caption is already public, and I would expect that it should be straightforward to construct a case where getting the caption for some time would be incorrect unless the captions are sorted internally.

abdelaziz-mahdy:
```
abstract class ClosedCaptionFile {
  /// The full list of captions from a given file.
  ///
  /// The [captions] will be in the order that they appear in the given file.
  List<Caption> get captions;
}

```

its only a getter so calling sort on it doesnt sort the original list, this is why i copied the list and sorted it and the public method was for the testing part only, if the test is not needed the public method is not needed

but `_sortedCaptions` is needed

stuartmorgan: > this is why i copied the list and sorted it
> [...]
> but `_sortedCaptions` is needed

I'm not sure what part of my comments led you to believe I don't understand `_sortedCaptions`. My comment is about the new public API.

> the public method was for the testing part only, if the test is not needed the public method is not needed

I still don't understand why you are presenting a binary choice between adding new public API and having testing. Whether or not the correct caption is fetched for a given time is already an observable property with existing public API. You should be able to construct a case where, with binary search, *the wrong caption is reported* without the internal sorting step, and thus validate that it is being handled correctly without directly inspecting internal state via new public accessors.

Testing internal implementation details is an anti-pattern, and should only be done when the important properties cannot be tested via the actual client APIs. I have not seen any argument for why that would not be the case here. What scenario are you unable to test without adding a new public API?

abdelaziz-mahdy: i didnt mean all the testing of the testing is ignored, just this test `captions are sorted correctly on initialization` since it requires the expose of the public interface so i can check if they are sorted correctly 

`List<Caption>? get sortedCaptions` without it i cant get the sorted list and without the sorted list i cant make that test

but i did add test cases for not sorted captions and tested retrieving them which works correctly and if the sorting is removed they wont pass

so i think those are enough, and no need to introduce the public interface since all of the cases are covered as you mentioned

and i did remove the public method.

stuartmorgan:
From triage: what's the status of this PR? Is it ready for re-review?

abdelaziz-mahdy:
> From triage: what's the status of this PR? Is it ready for re-review?

Yes it's ready for a re-review

packages/video_player/video_player/pubspec.yaml
```diff
@@ -25,11 +25,13 @@ dependencies:
   flutter:
     sdk: flutter
   html: ^0.15.0
+  collection: ^1.19.0
```
tarrinneal, line 28: You need to run the analyze tool as well to see what's failing - currently it's this line. These deps should be in alphabetical order.

tarrinneal (changes requested):
Changelog doesn't match our guidelines. And a tip with the failing tests.

Seems good to me besides.

packages/video_player/video_player/pubspec.yaml
tarrinneal: https://github.com/flutter/plugins/blob/master_archive/script/tool/README.md#commands

abdelaziz-mahdy: Thank you for the link, I sorted them.

packages/video_player/video_player/pubspec.yaml
```diff
@@ -30,6 +31,7 @@ dependencies:
   video_player_platform_interface: ^6.2.0
   video_player_web: ^2.1.0
 
+
```
tarrinneal, line 34: This line needs to be removed too

abdelaziz-mahdy: done.

tarrinneal:
Now I don't know why that test is failing. @stuartmorgan have you seen this type of failure with the analyze legacy tests before?

stuartmorgan:
> Now I don't know why that test is failing. @stuartmorgan have you seen this type of failure with the analyze legacy tests before?

Yes. The output explains the problem and what needs to be changed to fix it, so I'm not following what the issue is.

packages/video_player/video_player/pubspec.yaml
```diff
@@ -22,6 +22,7 @@ flutter:
         default_package: video_player_web
 
 dependencies:
+  collection: ^1.19.0
```
stuartmorgan, line 25: This constraint needs to be adjusted as described in the CI failure output.

tarrinneal: I didn't realize the fix was in the error. Sorry

tarrinneal: Needs to be version 1.18.0

abdelaziz-mahdy: Thank you very much! I didn’t find the error logs initially, but after you mentioned them, I looked further and found them.

tarrinneal:
Is this ready for re-review?

abdelaziz-mahdy:
> Is this ready for re-review?

Yes, all the checks and changes are fixed.

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## NEXT
+## 2.9.3
 
 * Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+* Optimizes caption retrieval with Binary search.
```
stuartmorgan, line 4: binary

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -763,6 +792,14 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     Future<ClosedCaptionFile>? closedCaptionFile,
   ) async {
     _closedCaptionFile = await closedCaptionFile;
+
+    _sortedCaptions = _closedCaptionFile?.captions;
+
+    /// Sort the captions by start time so that we can do a binary search.
```
stuartmorgan, line 798: "so that we can do" -> "to allow"

Using "we" in comments is often actively confusing for readers, because who "we" is isn't usually clear.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -763,6 +792,14 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+    _sortedCaptions?.sort((Caption a, Caption b) {
```
stuartmorgan, line 799: `_updateClosedCaptionWithFuture` is called frequently. Re-sorting every caption in the entire file every time this method is called is likely going to hurt overall performance relative to the state before this PR.

The sorted captions should be cached, and only re-generated when the file is changed.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -754,7 +776,32 @@ void main() {
...
+            VideoPlayerController.networkUrl(
+          _localhostUri,
+          closedCaptionFile: _loadClosedCaption(),
+        );
+
+        await controller.initialize();
+        final List<Caption> captions =
+            (await controller.closedCaptionFile)!.captions.toList();
+
+        // Check that captions are not in sorted order
```
stuartmorgan, line 790: This needs a period.

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -754,7 +776,32 @@ void main() {
     });
 
     group('caption', () {
-      test('works when seeking', () async {
+      test('makes sure the input captions are unsorted', () async {
```
stuartmorgan, line 779: Why is this a unit test? I.e., why is asserting that the visible caption list is not sorted useful? What user behavior would break if this test were failing?

stuartmorgan (changes requested):

packages/video_player/video_player/test/video_player_test.dart
abdelaziz-mahdy: the idea is this test makes sure that the input captions were unsorted, and the tests after it gets the right caption

I am just making sure that no one in the future resorting the input captions, and someone else removing the sorting from inside of the video player in that case the binary search will be broken for unsorted inputs.

please correct me if my thought process is wrong.

packages/video_player/video_player/lib/video_player.dart
abdelaziz-mahdy: I assumed _updateClosedCaptionWithFuture gets called only when file changes, and i cant find any other place it was called, but I made a method that that sorts the captions, and made calls to it when the file changes after `_updateClosedCaptionWithFuture`
I hope thats more efficient now.

stuartmorgan: > I assumed _updateClosedCaptionWithFuture gets called only when file changes, and i cant find any other place it was called

https://github.com/flutter/packages/blob/a7c1159ac3814bdd4669156bed3f7d5692cece4a/packages/video_player/video_player/lib/video_player.dart#L511C44-L511C68

abdelaziz-mahdy: noted, so my fix is not accurate, will think of other way and comment in this to let you know

abdelaziz-mahdy: I hope the added check is good enough, I couldnt think of a better way, since we need to make sure it runs on every new file, do you think I should move back the sorting with the new check in _updateClosedCaptionWithFuture or leaving them seperated is cleaner.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -378,6 +379,9 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
 
   Future<ClosedCaptionFile>? _closedCaptionFileFuture;
   ClosedCaptionFile? _closedCaptionFile;
+  ClosedCaptionFile? _previousClosedCaptionFile;
```
stuartmorgan, line 382: This is unnecessarily duplicating the memory requirements any time the file changes, and also does an extremely expensive compare for no reason every time. We know when the file is reset, without a compare, because there is an API to set it. We don't need extra state or checks to determine that.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -378,6 +379,9 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+  List<Caption>? _sortedCaptions;
```
stuartmorgan, line 383: Looking at the `ClosedCaptionFile` type to check the above, I realized it's storing the entire contents of the file in memory, which means that adding this variable is doubling the memory usage of the captioning system.

Do you have any benchmarks about lookup time in a profile build before and after this change with non-trivial files? Given that this is a memory/speed tradeoff, rather than simply an optimization, we need to understand the actual values involved here to know if this is a net positive.

stuartmorgan (changes requested):

stuartmorgan-g:
From triage: @abdelaziz-mahdy Are you still planning to update this PR based on the review comments?

abdelaziz-mahdy:
> From triage: @abdelaziz-mahdy Are you still planning to update this PR based on the review comments?

yes, I was busy, will update it and let you know

abdelaziz-mahdy:
i created a mockup benchmark test https://github.com/abdelaziz-mahdy/packages/commit/880c2d243f6d35f44ecc8eff87a06a650bdf9c8c#diff-2030dff5accbb0e6ccf7df8fbd5fc0374da840a562c9594478f997b164a5921e

the results are great even in the case of big files, if we need to eliminate the not sorted to avoid the duplicate, i think we can do that, what do you think? but that will require us to override it i think?
image

stuartmorgan-g:
> if we need to eliminate the not sorted to avoid the duplicate, i think we can do that, what do you think?

Unfortunately:
- the API of the caption file is explicitly that the captions are in file order, and
- the video player has public API to get back the caption file that was set.

I don't see a compelling reason for either of those things to be true, but I'd rather not make a breaking change just for this.

The speedup is definitely worthwhile given those benchmarks, as the lookup in a large file could be a non-trivial portion of a frame allowance. I think the best option here is to track fixing the API in the future (https://github.com/flutter/flutter/issues/168823), but for now duplicate the list for fast lookup.

(Per my earlier comment though, we don't need *two* extra copies of the file, or an expensive compare step to determine when to create a new sorted list. We just need to reset the cached sorted version in the setter.)

abdelaziz-mahdy:
I switched the logic to use `whenComplete` to only call the sort once on the future completion, I hope that matches what you meant, if not I currently don't know how to solve the event listener call to avoid resorting on every event fired,


the plan in my mind currently to add a flag so I know if the current file was sorted or not, but don't know if its good or not.

nateshmbhat:
this will be really cool to have in the subtitle selection @stuartmorgan-g

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -726,20 +729,46 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   ///
   /// If no [closedCaptionFile] was specified, this will always return an empty
   /// [Caption].
+
   Caption _getCaptionAt(Duration position) {
-    if (_closedCaptionFile == null) {
+    if (_closedCaptionFile == null || _sortedCaptions == null) {
```
stuartmorgan-g: Please add a `final List<Captions>? sortedCaptions = _sortedCaptions;` above this, and then use `sortedCaptions` everywhere, so that repeated force-unwraps aren't needed below.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -726,20 +729,46 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+      },
+    );
+
+    // -1 means not found by the binary search.
+    if (captionIndex == -1) {
+      return Caption.none;
+    }
+
+    final Caption caption = _sortedCaptions![captionIndex];
+    // check if it really fits within that caption's [start, end].
```
stuartmorgan-g: Under what circumstances will the `binarySearch` return a result that fails this check? If this is necessary, it seems like the `compare` function above is incorrect.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -760,8 +789,20 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   Future<void> _updateClosedCaptionWithFuture(
     Future<ClosedCaptionFile>? closedCaptionFile,
   ) async {
-    _closedCaptionFile = await closedCaptionFile;
-    value = value.copyWith(caption: _getCaptionAt(value.position));
+    if (closedCaptionFile != null) {
+      await closedCaptionFile.whenComplete(() async {
```
stuartmorgan-g: Why are you using `whenComplete` in an async function, rather than just `await`ing the future and then using the value?

stuartmorgan-g (changes requested):

packages/video_player/video_player/lib/video_player.dart
abdelaziz-mahdy: To avoid calling the sort Everytime the events get triggered.

stuartmorgan-g: I don't follow; every time you call `whenComplete`, the result will be this code running one time.

The way to avoid re-running sorting is to only do the sorting when the file actually changes. As I mentioned in a previous review, we know what that happens, because we controller the code of the setter (`setClosedCaptionFile`).

stuartmorgan-g:
@abdelaziz-mahdy Are you still planning on updating this PR to address the review comments above?

abdelaziz-mahdy:
> @abdelaziz-mahdy Are you still planning on updating this PR to address the review comments above?

Sadly I have been quite busy lately, and not easy to figure out, so I cant fix it yet

stuartmorgan-g:
Marking as a draft pending updates.

Piinks:

abdelaziz-mahdy:
Review comments addressed. Ready for re-review.

  Changes Made:

  1. Optimized caption lookup

  - Added local variable in _getCaptionAt to avoid repeated force-unwraps
  - Removed unnecessary validity check after binary search (compare function guarantees correctness)

  2. Fixed caching strategy

  - Refactored setClosedCaptionFile to always sort when explicitly setting a new file
  - Updated _updateClosedCaptionWithFuture to only sort on first initialization (_sortedCaptions == null)
  - Prevents expensive re-sorting on every event listener call

  3. Fixed test structure

  - Unnested incorrectly placed tests

  4. Added edge case tests

  Added 5 new tests covering binary search edge cases:
  - Exact caption start/end time boundaries
  - Gaps between captions
  - Single caption scenario
  - Overlapping captions

  The caching optimization ensures captions are only sorted when the file actually changes (initialization or explicit setClosedCaptionFile call), avoiding performance issues.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -758,20 +761,45 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
   /// [Caption].
+
   Caption _getCaptionAt(Duration position) {
-    if (_closedCaptionFile == null) {
+    if (_closedCaptionFile == null || _sortedCaptions == null) {
+      return Caption.none;
+    }
+
+    final List<Caption>? sortedCaptions = _sortedCaptions;
+    if (sortedCaptions == null) {
```
stuartmorgan-g: This is now being checked twice; you added a second copy of the check instead of just adding the local before the first copy.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -785,15 +813,41 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+      _sortedCaptions = List<Caption>.from(_closedCaptionFile!.captions)
+        ..sort((Caption a, Caption b) {
+          return a.start.compareTo(b.start);
+        });
+      value = value.copyWith(caption: _getCaptionAt(value.position));
+    } else {
+      _closedCaptionFile = null;
+      _sortedCaptions = null;
+      value = value.copyWith(caption: Caption.none);
+    }
```
stuartmorgan-g: Why have you duplicated all of `_updateClosedCaptionWithFuture` just to change one line?

All you need to do is unconditionally set `sortedCaptions` to null before calling the other method.

stuartmorgan-g (changes requested):

packages/video_player/video_player/lib/video_player.dart
abdelaziz-mahdy: thanks done.

packages/video_player/video_player/lib/video_player.dart
abdelaziz-mahdy: thanks done.

tarrinneal:
Sorry for the wait, this one slid through my process

PR #8342, opened 2024-12-23
[camera] Remove OCMock from tests

mchudy:
This PR removes dependency on OCMock in unit tests and serves as a first step for full Swift migration. https://github.com/flutter/flutter/issues/119109

- Introduces wrapper protocols around `AVFoundation` class with default implementations and mocks
- Replaces usages of `AVFoundation` classes with new protocols
- Refactors dependency injection with `FLTCamConfiguration` for easier usage in tests as new protocols increased the number of dependencies required to initialize a new `FLTCam` instance
- I tried to minimize changes to `FLTCam` to minimize the risk of introducing regressions. Splitting `FLTCam` into smaller classes will happen in further PRs.
- Relying on the new protocols in some cases is cumbersome because `AVFoundation` methods require pointers to raw objects thus I'm sometimes keeping the underlying instance exposed which is not very elegant, I'm open to any suggestions on to how we can clean this up.

I took some inspiration from how it was done in case of the `in_app_purchase` package https://github.com/flutter/flutter/issues/116383

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] page, which explains my responsibilities.
- [X] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [X] I signed the [CLA].
- [X] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [X] I [linked to at least one issue that this PR fixes] in the description above.
- [X] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [X] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making, or this PR is [test-exempt].
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[...]

PR #8351, opened 2024-12-27
[camera] Remove OCMock from CameraExposureTests and CameraFocusTests

mchudy:
Extracted from https://github.com/flutter/packages/pull/8342 to keep PRs smaller

- Introduces `FLTDeviceOrientationProviding` and `FLTCaptureDeviceControlling` protocols
- Removes `OCMock.h` reference from `CameraExposureTests.h` and `CameraFocusTests.h`

## Pre-launch Checklist

- [X] I signed the [CLA].


packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraExposureTests.m
```diff
@@ -5,51 +5,74 @@
...
-  _mockUIDevice = OCMPartialMock([UIDevice currentDevice]);
-}
+  _mockDevice = [[MockCaptureDeviceController alloc] init];
+  _mockDeviceOrientationProvider = [[MockDeviceOrientationProvider alloc] init];
 
-- (void)tearDown {
-  [_mockDevice stopMocking];
-  [_mockUIDevice stopMocking];
+  [_camera setValue:_mockDevice forKey:@"captureDevice"];
+  [_camera setValue:_mockDeviceOrientationProvider forKey:@"deviceOrientationProvider"];
```
hellohuanlin, line 26: can you inject these 2 via `FLTCam`'s constructor? The string keyPath can be error prone

hellohuanlin (commented):
Overall looks very good!

hellohuanlin:
CC @LouiseHsu who recently did a pretty big swift migration for in-app purchase plugin.

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraExposureTests.m
mchudy: Ah yes, absolutely, I think I did it in further commits in the initial PR, but I'll add DI for those here.

mchudy:
I'm not sure why the CI build fails. Locally it builds successfully. Do you have any tip maybe what can be the cause of this issue?
```
Failed to build iOS app
Lexical or Preprocessor Issue (Xcode): 'FLTCaptureDeviceControlling.h' file not found
/Volumes/Work/s/w/ir/x/w/packages/packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCamMediaSettingsAVWrapper.h:7:8
```

hellohuanlin:
> I'm not sure why the CI build fails. Locally it builds successfully. Do you have any tip maybe what can be the cause of this issue?
> 
> ```
> Failed to build iOS app
> Lexical or Preprocessor Issue (Xcode): 'FLTCaptureDeviceControlling.h' file not found
> /Volumes/Work/s/w/ir/x/w/packages/packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCamMediaSettingsAVWrapper.h:7:8
> ```

it's been a while since i touched this, but it sounds like either module map or umbrella header. I'd look into a similar header and see how it is setup.

hellohuanlin (approved):

hellohuanlin:
I don't find a 2nd reviewer from the CODEOWNER for this plugin. Maybe @LouiseHsu who recently did a similar swift migration?

mchudy:
I fixed the CI, kind of a workaround (for some reason I cannot get this to work with subfolders). I'm gonna try to fix that in further PRs, but I don't want to block this one.

stuartmorgan (changes requested):

packages/camera/camera_avfoundation/example/ios/RunnerTests/MockCaptureDeviceController.h
```diff
@@ -0,0 +1,78 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MockCaptureDeviceController : NSObject <FLTCaptureDeviceControlling>
+// Position/Orientation
+@property(nonatomic, assign) AVCaptureDevicePosition position;
+
+// Format/Configuration
+@property(nonatomic, strong) AVCaptureDeviceFormat *activeFormat;
+@property(nonatomic, strong) NSArray<AVCaptureDeviceFormat *> *formats;
+@property(nonatomic, copy) void (^setActiveFormatStub)(AVCaptureDeviceFormat *format);
```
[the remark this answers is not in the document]
stuartmorgan: Ah, I missed that in the doc. @hellohuanlin is there some source for that naming convention? The purpose of those methods does not at all match the use of "stub" I'm familiar with in tests.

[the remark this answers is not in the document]
hellohuanlin: No source - IIRC I wrote `XXStub` simply because it's to replace `OCMStub(...).andDo(...)`. I am fine with any convention we discuss here.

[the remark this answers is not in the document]
stuartmorgan: Hm, looking at the doc again, and the methods here, maybe there's just a mismatch in usage in this PR. In the doc it looks more like what I expect something called `stub` to be, which is replacing what the method does. In the usage here, it's basically just a notifier:

```objc
- (void)setFocusMode:(AVCaptureFocusMode)mode {
  _focusMode = mode;
  if (self.setFocusModeStub) {
    self.setFocusModeStub(mode);
  }
}
```

It seems like this is trying to both be a simple stub object (which takes state to return) and sort of a dynamic stub (which would use stub methods) at the same time, and sometimes the latter is really a stub and other times (like the above) it's not.

We should decide what pattern we want to use for read/write properties. If it's that the getter and setter have real implementations in the stub object, then any supplemental callback shouldn't be called "stub" IMO, because it has nothing to do with what the object returns. If it's that we have stub methods, then those should be the whole implementation, and the object should be stateless other than the stub methods.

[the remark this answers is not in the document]
mchudy: Oh, you're right, that's a mistake on my part. I also pictured those stub methods as having no default implementation so `_focusMode = mode` assignment shouldn't be there. I corrected this.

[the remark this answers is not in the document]
mchudy: @stuartmorgan can you take a look now?

packages/camera/camera_avfoundation/example/ios/RunnerTests/MockCaptureDeviceController.m
```diff
@@ -0,0 +1,119 @@
...
+}
+
+- (void)setExposureMode:(AVCaptureExposureMode)mode {
+  if (self.setExposureModeStub) {
+    self.setExposureModeStub(mode);
+  }
+}
+
+- (void)setExposurePointOfInterest:(CGPoint)point {
+  _exposurePointOfInterest = point;
```
stuartmorgan, line 58: This method has the same issue of having a stub setter, but also trying to be its own stub.

packages/camera/camera_avfoundation/example/ios/RunnerTests/MockCaptureDeviceController.m
```diff
@@ -0,0 +1,119 @@
...
+- (void)setVideoZoomFactor:(float)factor {
+  if (self.setVideoZoomFactorStub) {
+    self.setVideoZoomFactorStub(factor);
+  }
+}
+
+- (BOOL)lockForConfiguration:(NSError **)error {
+  if (self.lockForConfigurationStub) {
+    self.lockForConfigurationStub(error);
+    return !self.shouldFailConfiguration;
```
stuartmorgan, line 79: Same here; why isn't this whole method just

```objc
if (self.lockForConfigurationStub) {
  return self.lockForConfigurationStub(error);
}
return YES;
```

packages/camera/camera_avfoundation/example/ios/RunnerTests/MockCaptureDeviceController.m
```diff
@@ -0,0 +1,119 @@
...
+
+- (BOOL)isExposureModeSupported:(AVCaptureExposureMode)mode {
+  return self.exposureModeSupported;
+}
+
+- (AVCaptureInput *)createInput:(NSError *_Nullable *_Nullable)error {
+  if (self.createInputStub) {
+    self.createInputStub(error);
+  }
+  return self.inputToReturn;
```
stuartmorgan, line 116: Similar question here; why isn't the way of returning something to have `createInputStub` return it?

packages/camera/camera_avfoundation/example/ios/RunnerTests/MockCaptureDeviceController.m
mchudy: Yeah, you're right, I didn't look thoroughly. There's no reason whatsoever, I changed that.

stuartmorgan (approved):
LGTM, thanks!

PR #8363, opened 2025-01-02
[camera] Remove OCMock from AvailableCamerasTests, CameraOrientationTests and ThreadSafeEventChannelTests

mchudy:
Follow-up on https://github.com/flutter/packages/pull/8342. This should be merged after https://github.com/flutter/packages/pull/8351

- Removes OCMock dependency in `AvailableCamerasTests.m`, `CameraOrientationTests.m` and `ThreadSafeEventChannelTests.m`
- Introduces new protocol wrappers to enable mocking `FLTEventChannel` and `FLTCameraDeviceDiscovering`

## Pre-launch Checklist

- [X] I signed the [CLA].


packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCaptureDeviceControlling.m
```diff
@@ -0,0 +1,176 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "./include/camera_avfoundation/FLTCaptureDeviceControlling.h"
+
+@interface FLTDefaultCaptureDeviceController ()
+@property(nonatomic, strong) AVCaptureDevice *device;
```
hellohuanlin, line 8: Can we directly have `AVCaptureDevice` conform to `FLTDefaultCaptureDeviceControlling` protocol and save this boilerplate in this file?

mchudy: You're right, after playing around a bit I found that it is possible do that. I thought in ObjC that would be problematic and wanted to postpone it to the Swift migration (as Swift protocol extensions give much more flexibility). I had to change some other things like abstract `AVCaptureInput` creation to make this work, but it should be much cleaner now.

packages/camera/camera_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 0.9.17+8
```
hellohuanlin, line 1: I haven't seen 2 released in a single PR before. This is probably not a standard practice (@stuartmorgan will know)

May worth splitting into 2 PRs, which can make the review easier/faster.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCaptureDeviceControlling.h
```diff
@@ -0,0 +1,87 @@
...
+@protocol FLTCaptureInput <NSObject>
+@property(nonatomic, readonly) NSArray<AVCaptureInputPort *> *ports;
+@end
+
+@protocol FLTCaptureDeviceInputFactory <NSObject>
+- (nullable id<FLTCaptureInput>)deviceInputWithDevice:(id<FLTCaptureDeviceControlling>)device
+                                                error:(NSError **)error;
+@end
+
+@interface AVCaptureDevice (FLTCaptureDeviceControlling) <FLTCaptureDeviceControlling>
```
hellohuanlin, line 78: I'd just call the protocol `FLTCaptureDevice` since it's a noun (rather than a capability), [as discussed in the design doc](https://docs.google.com/document/d/1XsaulkJA6_ZSpM7chkQLhQY25sQqhEMiqHMYzw8H85o/edit?resourcekey=0-_cUjF1c0iBvRLKfV-3gK5A&tab=t.0#heading=h.a6hzk7tvmg6z)

packages/camera/camera_avfoundation/CHANGELOG.md
stuartmorgan: CI won't allow this; it can either be two PRs, or it can be one PR with a single version bump.

mchudy: So this PR was meant to be merged after https://github.com/flutter/packages/pull/8351 (it includes changes from that PR), that's why there are two new versions. So either we merge them one after the other or we close #8351 and only merge this one. Sorry for making it confusing.

hellohuanlin: There's a trick that you can set base branch your previous branch first, and after that gets landed, change the base branch back to main.

mchudy: Good idea, but I believe that I can only set the base branch as a branch that exists in the main repo, it doesn't list branches from my fork

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCaptureDeviceControlling.h
mchudy: Done

packages/camera/camera_avfoundation/CHANGELOG.md
hellohuanlin: oh i will review after the previous PR is in then. otherwise it's hard to tell which part is new change.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCaptureDevice.h
```diff
@@ -9,7 +9,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// A protocol which is a direct passthrough to AVCaptureDevice.
 /// It exists to allow replacing AVCaptureDevice in tests.
-@protocol FLTCaptureDeviceControlling <NSObject>
+@protocol FLTCaptureDevice <NSObject>
+
+// Device
+- (NSString *)uniqueID;
```
stuartmorgan, line 15: Is there a reason this isn't a read-only property rather than a method?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/CameraPlugin_Test.h
```diff
@@ -6,29 +6,41 @@
...
 @property(nonatomic, strong) dispatch_queue_t captureSessionQueue;
 
 /// An internal camera object that manages camera's state and performs camera operations.
-@property(nonatomic, strong) FLTCam *camera;
+@property(nonatomic, strong) FLTCam *_Nullable camera;
 
 /// Inject @p FlutterTextureRegistry and @p FlutterBinaryMessenger for unit testing.
-- (instancetype)initWithRegistry:(NSObject<FlutterTextureRegistry> *)registry
-                       messenger:(NSObject<FlutterBinaryMessenger> *)messenger;
+- (nonnull instancetype)initWithRegistry:(NSObject<FlutterTextureRegistry> *_Nullable)registry
```
stuartmorgan, line 28: nonnull shouldn't be needed here, due to the macro.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/CameraPlugin_Test.h
```diff
@@ -6,29 +6,41 @@
...
+                               messenger:(NSObject<FlutterBinaryMessenger> *_Nullable)messenger;
 
 /// Inject @p FlutterTextureRegistry, @p FlutterBinaryMessenger, and Pigeon callback handler for
 /// unit testing.
-- (instancetype)initWithRegistry:(NSObject<FlutterTextureRegistry> *)registry
-                       messenger:(NSObject<FlutterBinaryMessenger> *)messenger
-                       globalAPI:(FCPCameraGlobalEventApi *)globalAPI NS_DESIGNATED_INITIALIZER;
+- (nonnull instancetype)initWithRegistry:(NSObject<FlutterTextureRegistry> *_Nullable)registry
```
stuartmorgan, line 33: Same.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/CameraPlugin_Test.h
```diff
@@ -6,29 +6,41 @@
...
+                               globalAPI:(FCPCameraGlobalEventApi *_Nullable)globalAPI
+                        deviceDiscoverer:(id<FLTCameraDeviceDiscovering>)deviceDiscoverer
+                           deviceFactory:(CaptureNamedDeviceFactory)deviceFactory
+                   captureSessionFactory:(CaptureSessionFactory)captureSessionFactory
+               captureDeviceInputFactory:(id<FLTCaptureDeviceInputFactory>)captureDeviceInputFactory
+    NS_DESIGNATED_INITIALIZER;
 
 /// Hide the default public constructor.
-- (instancetype)init NS_UNAVAILABLE;
+- (nonnull instancetype)init NS_UNAVAILABLE;
```
stuartmorgan, line 43: Same.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/CameraPlugin_Test.h
```diff
@@ -40,7 +52,10 @@
 /// @param name the name of the camera.
 /// @param settings the creation settings.
 /// @param completion the callback to inform the Dart side of the plugin of creation.
-- (void)createCameraOnSessionQueueWithName:(NSString *)name
+- (void)createCameraOnSessionQueueWithName:(NSString *_Nullable)name
```
stuartmorgan, line 55: Is `name` really nullable?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/CameraPlugin_Test.h
```diff
@@ -6,29 +6,41 @@
...
+                               messenger:(NSObject<FlutterBinaryMessenger> *_Nullable)messenger;
 
 /// Inject @p FlutterTextureRegistry, @p FlutterBinaryMessenger, and Pigeon callback handler for
 /// unit testing.
-- (instancetype)initWithRegistry:(NSObject<FlutterTextureRegistry> *)registry
-                       messenger:(NSObject<FlutterBinaryMessenger> *)messenger
-                       globalAPI:(FCPCameraGlobalEventApi *)globalAPI NS_DESIGNATED_INITIALIZER;
+- (nonnull instancetype)initWithRegistry:(NSObject<FlutterTextureRegistry> *_Nullable)registry
+                               messenger:(NSObject<FlutterBinaryMessenger> *_Nullable)messenger
+                               globalAPI:(FCPCameraGlobalEventApi *_Nullable)globalAPI
```
stuartmorgan, line 35: Are these three properties really nullable? I would expect things not to work without them.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCameraDeviceDiscovering.m
```diff
@@ -0,0 +1,30 @@
...
+    discoverySessionWithDeviceTypes:(NSArray<AVCaptureDeviceType> *)deviceTypes
+                          mediaType:(AVMediaType)mediaType
+                           position:(AVCaptureDevicePosition)position {
+  AVCaptureDeviceDiscoverySession *discoverySession =
+      [AVCaptureDeviceDiscoverySession discoverySessionWithDeviceTypes:deviceTypes
+                                                             mediaType:mediaType
+                                                              position:position];
+
+  NSArray<AVCaptureDevice *> *devices = discoverySession.devices;
+  NSMutableArray<id<FLTCaptureDevice>> *deviceControllers = [NSMutableArray array];
```
stuartmorgan, line 21: Nit: `arrayWithCapacity:devices.count` to avoid unnecessary storage re-allocation.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCameraDeviceDiscovering.m
```diff
@@ -0,0 +1,30 @@
...
+
+  for (AVCaptureDevice *device in devices) {
+    [deviceControllers addObject:(id<FLTCaptureDevice>)device];
```
stuartmorgan, line 24: Is this duck-typing `device` even though device doesn't actually conform to the protocol, formally? That's a dangerous pattern because it discards compiler checks that things actually work; it would be much safer to manually write the (tedious but trivial) passthrough boilerplate of having a default implementation of the protocol that takes an AVCaptureDevice as its init param.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/CameraPlugin.m
```diff
@@ -117,14 +137,13 @@ - (void)availableCamerasWithCompletion:
...
-                               position:AVCaptureDevicePositionUnspecified];
-    NSArray<AVCaptureDevice *> *devices = discoverySession.devices;
+    NSArray<id<FLTCaptureDevice>> *devices =
+        [self.deviceDiscoverer discoverySessionWithDeviceTypes:discoveryDevices
+                                                     mediaType:AVMediaTypeVideo
+                                                      position:AVCaptureDevicePositionUnspecified];
     NSMutableArray<FCPPlatformCameraDescription *> *reply =
         [[NSMutableArray alloc] initWithCapacity:devices.count];
-    for (AVCaptureDevice *device in devices) {
+    for (id<FLTCaptureDevice> device in devices) {
```
stuartmorgan, line 146: Why is this cast necessary? This is how the array was typed above.

stuartmorgan (changes requested):

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/CameraPlugin_Test.h
mchudy: Yeah, you're right, they shouldn't be. That was a shortcut on my part that I forgot to remove later. It was easier to pass nil in tests, but those will need some mocking.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/CameraPlugin.m
mchudy: It's not really a cast, we have to provide the type in Obj-C for..in loops, so this is the exact type of the array item as the array is declared as `NSArray<id<FLTCaptureDevice>> *devices`. Previously that loop was operating on `AVCaptureDevice*`, but now that the session discovery is abstracted away this operates on the new protocol (it only accesses `uniqueID` and `position` properties inside).

stuartmorgan: > It's not really a cast, we have to provide the type in Obj-C for..in loops

Sorry, I have no idea what happened to my reading comprehension when I was reviewing that 🤦🏻

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/CameraPlugin_Test.h
mchudy: I removed the nullability, to fix warnings in tests I had to introduce new some mocks for Flutter-related protocols.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCameraDeviceDiscovering.m
mchudy: In this case the cast was not needed, I removed it. As for the main point, I also thought about this a lot, but I'm still not sure which approach is the best. I initially implemented all mocks with direct passthroughs, but they were also easy to misuse (in FLTCam many AVFoundation methods are called that require raw classes and not wrappers) and I got into runtime errors while running the example app (without any compiler errors being reported).

As changes to implementation in this PR are limited and there will be more of them in the upcoming PRs I think we can postpone this discussion until we have better examples and then decide on the approach we should take.

mchudy: > The updated version hasn't fundamentally changed anything, it's just highlighting that Obj-C apparently doesn't enforce types on collection generics.

You're right, sorry, I didn't look closely. It is unsafe, but with explicit wrapper it would also allow implicit cast I believe, which is quite unfortunate for Obj-C.

> Could you elaborate on this? 

One example from the [original PR](https://github.com/flutter/packages/pull/8342/files) that I have off the top of my head. This is an implementation of the method in `FLTDefaultCaptureSession` which is a wrapper around `AVCaptureSession`:
 
 ```objc
- (BOOL)canAddConnection:(id<FLTCaptureConnection>)connection {
  return [_captureSession canAddConnection:connection.connection];
}
```

You can see that inside I'm calling `canAddConnection` on the underlying `AVCaptureSession*` instance. However as that requires the AVFoundation class to be passed in as argument (it can call some selectors that are not defined on the protocol) I cannot pass `id<FLTCaptureConnection>` protocol directly. This would all be fine when we access the underlying AVFoundation class from connection as above, but the bad thing is that changing this implementation to:

 ```objc
- (BOOL)canAddConnection:(id<FLTCaptureConnection>)connection {
  return [_captureSession canAddConnection:connection];
}
```

unfortunately still compiles without any error and will crash at runtime, as AVCaptureSession will try to access methods that are not defined on the protocol (they could even be undocumented private framework methods).

(I don't remember if that crash would happen in this exact case, but I'm showing this as a simple example that Obj-C doesn't verify protocol instances passed with `id`. I remember for sure that I had runtime crashes with `createInput` type of methods).

If you have any suggestion how we can make that all safe at compile time I'm very happy to hear it and implement it.

> Could you elaborate on this? All it takes is someone adding anything to FLTCaptureDevice that does not exactly match AVCaptureDevice, and the result will be runtime crashes that the compiler cannot warn us about because its type checking has been bypassed.

That's absolutely true and I'm not principally against the passthrough pattern. I just wanted to highlight that the passthrough pattern won't easily save us from errors at runtime. Maybe there's a way to do this better, I'm not that familiar with all intricacies of Obj-C. The only good thing is that in Swift we'll be able to switch to protocol extensions which as far as I remember will give us both compile time safety and allow us to get rid of all the unnecessary boilerplate.

stuartmorgan: > I'm showing this as a simple example that Obj-C doesn't verify protocol instances passed with `id`.

That's really surprising. Do you get warnings if you use `NSObject<protocol>` instead?

mchudy: > That's really surprising. Do you get warnings if you use NSObject<protocol> instead?

Oh, you're right! This in fact enforces compile time type checking. If switching from `id<protocol>` to `NSObject<protocol>` gives us a compile time safety then sure, passthrough sounds a like a better choice overall.

packages/camera/camera_avfoundation/example/ios/RunnerTests/CameraCaptureSessionQueueRaceConditionTests.m
```diff
@@ -8,13 +8,33 @@
...
+#import "MockGlobalEventApi.h"
+
 @interface CameraCaptureSessionQueueRaceConditionTests : XCTestCase
 @end
 
 @implementation CameraCaptureSessionQueueRaceConditionTests
 
 - (void)testFixForCaptureSessionQueueNullPointerCrashDueToRaceCondition {
-  CameraPlugin *camera = [[CameraPlugin alloc] initWithRegistry:nil messenger:nil];
+  MockCaptureDevice *captureDevice = [[MockCaptureDevice alloc] init];
```
hellohuanlin, line 24: nit: maybe create a helper `createCameraPlugin` like in other tests, in case we want to add more test functions in this file.

packages/camera/camera_avfoundation/example/ios/RunnerTests/Mocks/MockCaptureDevice.m
```diff
@@ -101,11 +100,37 @@ - (BOOL)isExposureModeSupported:(AVCaptureExposureMode)mode {
   return self.exposureModeSupported;
 }
 
-- (AVCaptureInput *)createInput:(NSError *_Nullable *_Nullable)error {
-  if (self.createInputStub) {
-    return self.createInputStub(error);
+@synthesize device;
```
hellohuanlin, line 103: what is this for?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCaptureDevice.h
```diff
@@ -63,17 +71,41 @@ NS_ASSUME_NONNULL_BEGIN
...
+/// It exists to allow mocking instances of AVCaptureDeviceInput in tests.
+@protocol FLTCaptureDeviceInputFactory <NSObject>
+- (nullable NSObject<FLTCaptureInput> *)deviceInputWithDevice:(NSObject<FLTCaptureDevice> *)device
+                                                        error:(NSError **)error;
+@end
 
-/// Initializes the controller with the given device.
+/// A default implementation of `FLTCaptureDevice` which is a direct passthrough to the underlying
+/// `AVCaptureDevice`.
+@interface FLTDefaultCaptureDevice : NSObject <FLTCaptureDevice>
```
hellohuanlin, line 96: why do we need the wrapper here? Can we directly have `AVCaptureDevice` conform to `FLTCaptureDevice` protocol?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCaptureDevice.h
```diff
@@ -63,17 +71,41 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (instancetype)initWithDevice:(AVCaptureDevice *)device;
+@end
+
+/// A default implementation of `FLTCaptureInput` which is a direct passthrough to the underlying
+/// `AVCaptureInput`.
+@interface FLTDefaultCaptureInput : NSObject <FLTCaptureInput>
```
hellohuanlin, line 102: same here

hellohuanlin (commented):
Overall looks good. 

(Please make future PRs a bit smaller, as large PRs like this one are harder to review)

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/include/camera_avfoundation/FLTCaptureDevice.h
mchudy: Please refer to the above discussion with @stuartmorgan https://github.com/flutter/packages/pull/8363#discussion_r1934276991

hellohuanlin: am i reading it right that the reason here is that some AVFoundation APIs expect the actual `AVCaptureDevice` type to be passed in?

mchudy: That is related, but the biggest problem is that categories in Obj-C don't provide good type safety. In some cases we might be passing wrong type and the compiler won't warn us.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCameraDeviceDiscovering.m
```diff
@@ -0,0 +1,31 @@
...
+                           position:(AVCaptureDevicePosition)position {
+  AVCaptureDeviceDiscoverySession *discoverySession =
+      [AVCaptureDeviceDiscoverySession discoverySessionWithDeviceTypes:deviceTypes
+                                                             mediaType:mediaType
+                                                              position:position];
+
+  NSArray<AVCaptureDevice *> *devices = discoverySession.devices;
+  NSMutableArray<NSObject<FLTCaptureDevice> *> *deviceControllers =
+      [NSMutableArray arrayWithCapacity:devices.count];
+
```
stuartmorgan, line 23: Nit: Remove this blank line; the creation and population of this array are one conceptual code unit.

packages/camera/camera_avfoundation/example/ios/RunnerTests/Mocks/MockCaptureDevice.m
stuartmorgan: Conforming to a protocol that declares properties doesn't trigger auto-synthesis (unlike explicitly declaring a property in a class), so this is necessary to get what we would generally think of as default implementations of the properties that come from the protocols these mocks implement.

stuartmorgan (approved):
LGTM! Apologies that I didn't manage to finish this yesterday as I had planned.

mchudy:
I fixed all the nits.

stuartmorgan:
Changes LGTM; feel free to `autosubmit` :)

PR #8379, opened 2025-01-07
[video_player] Activate leak testing

ValentinVignal:
The package manipulates disposable objects. This PR activates leak testing to make sure disposable objects are correctly disposed.

It also fixes the memory leak warnings from the tests

See the documentation: https://github.com/dart-lang/leak_tracker/blob/main/doc%2Fleak_tracking%2FDETECT.md

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player/pubspec.yaml
+++ b/packages/video_player/video_player/pubspec.yaml
@@ -25,6 +25,7 @@ dependencies:
   flutter:
     sdk: flutter
   html: ^0.15.0
+  leak_tracker_flutter_testing: any
   video_player_android: ^2.3.5
   video_player_avfoundation: ^2.5.6
   video_player_platform_interface: ^6.2.0
--- a/packages/video_player/video_player/pubspec.yaml
+++ b/packages/video_player/video_player/pubspec.yaml
@@ -25,7 +25,6 @@ dependencies:
   flutter:
     sdk: flutter
   html: ^0.15.0
-  leak_tracker_flutter_testing: any
   video_player_android: ^2.3.5
   video_player_avfoundation: ^2.5.6
   video_player_platform_interface: ^6.2.0
@@ -34,6 +33,7 @@ dependencies:
 dev_dependencies:
   flutter_test:
     sdk: flutter
+  leak_tracker_flutter_testing: any
 
 topics:
   - video
--- a/packages/video_player/video_player/test/video_player_test.dart
+++ b/packages/video_player/video_player/test/video_player_test.dart
@@ -131,6 +131,7 @@ void main() {
 
   testWidgets('update texture', (WidgetTester tester) async {
     final FakeController controller = FakeController();
+    addTearDown(controller.dispose);
     await tester.pumpWidget(VideoPlayer(controller));
     expect(find.byType(Texture), findsNothing);
 
@@ -146,6 +147,7 @@ void main() {
 
   testWidgets('update controller', (WidgetTester tester) async {
     final FakeController controller1 = FakeController();
+    addTearDown(controller1.dispose);
     controller1.textureId = 101;
     await tester.pumpWidget(VideoPlayer(controller1));
     expect(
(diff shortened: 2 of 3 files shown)


ValentinVignal:
Since I modified only the tests, I think this PR is eligible for the tags `override: no versioning needed` and `override: no changelog needed`

ValentinVignal:
cc @polina-c

packages/video_player/video_player/test/video_player_test.dart
```diff
@@ -698,6 +723,7 @@ void main() {
         expect(controller.value.isPlaying, isTrue);
 
         await controller.pause();
+        await tester.runAsync(controller.dispose);
```
ValentinVignal, line 726: For those ones, I didn't manage to make the test pass without using `runAsync`.

If I use
```dart
addTearDown(() {
  controller.dispose();
});
```

or 

```dart
// end of test
unawaited(controller.dispose());
// a lot of 
tester.pump();
tester.pump();
```

it doesn't fix the memory leak.

If I use 

```dart
addTearDown(controller.dispose());
```

or

```dart
// end of test
await controller.dispose();
```

the test hangs and times out.


Do you have an idea on how to do it better?

polina-c: It looks smelly for me that `dispose` is async. But, it is test only class, so it is ok. 
And thus `runAsync` makes perfect sense.

ValentinVignal:
Any thoughts on this PR @tarrinneal ?

packages/video_player/video_player/pubspec.yaml
```diff
@@ -33,6 +33,7 @@ dependencies:
 dev_dependencies:
   flutter_test:
     sdk: flutter
+  leak_tracker_flutter_testing: any
```
tarrinneal, line 36: I'm not sure what our version policies are for dev_deps. @stuartmorgan is there guidance on this?

packages/video_player/video_player/test/video_player_test.dart
tarrinneal: Are these tests doing anything to test the real implementation of this class (VideoPlayerController)? In the actual usage of `dispose` there are asynchronous happenings. I'm not super familiar with how the leak testing works.

packages/video_player/video_player/pubspec.yaml
ValentinVignal: For more context, here is the documentation: https://prudential-ps.atlassian.net/browse/GTP-2844

> Put `any` instead of version, because the version is defined by your Flutter SDK.

packages/video_player/video_player/test/video_player_test.dart
ValentinVignal: From what I remember, they do. Those controller were created and I had issues making the tests wait for the dispose methods to finish

packages/video_player/video_player/pubspec.yaml
stuartmorgan: Yes, generally this would be a very bad idea, but it's specifically needed for the case of this package (IIUC because it's not in the SDK, so can't use `sdk:`, but is *pinned* by Flutter).

tarrinneal:
@stuartmorgan another question on this one. I think it should be exempt from changelog and versioning as it is only changing tests and dev-dependencies.

tarrinneal (approved):
lgtm sans response from Stuart about versioning.

stuartmorgan:
Version/changelog override: dev-only changes are exempt

stuartmorgan:
> @stuartmorgan another question on this one. I think it should be exempt from changelog and versioning as it is only changing tests and dev-dependencies.

That's correct; the version-check tooling just hasn't been taught to do analysis deeper than file path level, so it can't currently distinguish between pubspec.yaml chages that need publishing, and those that only affect `dev_dependencies`. I think I filed it at some point, but it hasn't been a priority since it's easy to override.

tarrinneal:
@stuartmorgan another one for your list

stuartmorgan (approved):
This shouldn't need a second approval, but LGTM

PR #8397, opened 2025-01-08
[two_dimensional_scrollables] prepare for `const AnimationStyle`

nate-thegrate:
unblocks https://github.com/flutter/flutter/pull/160564

Will go back through and implement `const` constructors in these spots once the framework PR lands.


Piinks (approved):
LGTM

PR #8111, opened 2024-11-16
[ci] Run web tests in wasm (unit + integration).

ditman:
Adds CI configuration to run web integration tests (in the master channel) compiled to Wasm.

It also removes the `build-examples` step from web integration tests, in some isolated testing:

| platform tests shard | With build-examples | Without build-examples |
|---|-----|-----|
| 1 | 30m | 21m |
| 2 | 13m | 11m |
| 3 | 17m | 10m |

## Issues

* Fixes https://github.com/flutter/flutter/issues/151664

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or this PR is [exempt from CHANGELOG changes].
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[exempt from version changes]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version
[...]

PR #8432, opened 2025-01-15
[google_maps_flutter] Support for Ground Overlay

jokerttu:
Adds Ground Overlay support for google_maps_flutter plugin.

Fixes https://github.com/flutter/flutter/issues/26479

[Android](https://developers.google.com/maps/documentation/android-sdk/groundoverlay):

[iOS](https://developers.google.com/maps/documentation/ios-sdk/overlays):

[Web](https://developers.google.com/maps/documentation/javascript/groundoverlays):

Documentation for groundOverlays parameter on app-facing plugin:
documentation

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
+++ b/packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
@@ -3,7 +3,6 @@
 // found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:io';
 import 'dart:ui' as ui;
 
 import 'package:flutter/foundation.dart';
@@ -22,7 +21,7 @@ void main() {
 }
 
 void runTests() {
-  const double floatTolerance = 1e-8;
+  const double floatTolerance = 1e-6;
 
   GoogleMapsFlutterPlatform.instance.enableDebugInspection();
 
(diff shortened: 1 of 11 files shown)


packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
```diff
@@ -22,7 +21,7 @@ void main() {
 }
 
 void runTests() {
-  const double floatTolerance = 1e-8;
+  const double floatTolerance = 1e-6;
```
jokerttu, line 24: Added bigger tolerance as on Android and iOS tests failed as diff was bigger than 1e-8

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
```diff
@@ -208,269 +207,574 @@ void runTests() {
...
+            onMapCreated: (GoogleMapController controller) {
+              mapIdCompleter.complete(controller.mapId);
+            },
+          ),
         ),
-        HeatmapGradientColor(
-          Color.fromARGB(255, 0, 0, 191),
-          0.6,
+      );
+      await tester.pumpAndSettle(const Duration(seconds: 3));
```
jokerttu, line 392: Without this line tests timeouted on Web platform, and we were not able to test ground overlay below.

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
```diff
@@ -208,269 +207,574 @@ void runTests() {
...
+            onMapCreated: (GoogleMapController controller) {
+              mapIdCompleter.complete(controller.mapId);
+            },
+          ),
         ),
-        HeatmapGradientColor(
-          Color.fromARGB(255, 255, 0, 0),
-          1,
+      );
+      await tester.pumpAndSettle(const Duration(seconds: 3));
```
jokerttu, line 450: Without this line heatmap tests timeouted on Web platform, and we were not able to test ground overlay below.

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
```diff
@@ -208,269 +207,574 @@ void runTests() {
...
-    List<WeightedLatLng> data2,
-  ) {
-    expect(data1.length, data2.length);
-    for (int i = 0; i < data1.length; i++) {
-      final WeightedLatLng wll1 = data1[i];
-      final WeightedLatLng wll2 = data2[i];
-      expect(wll1.weight, wll2.weight);
-      expect(wll1.point.latitude, moreOrLessEquals(wll2.point.latitude));
-      expect(wll1.point.longitude, moreOrLessEquals(wll2.point.longitude));
+  group('Heatmaps', () {
```
jokerttu, line 210: Added Heatmap tests to own grounp, like tile overlay tests and ground overlay tests.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_inspector_platform.dart
```diff
@@ -117,11 +117,7 @@ abstract class GoogleMapsInspectorPlatform extends PlatformInterface {
   }
 
   /// If the platform supports getting information about heatmaps.
-  bool supportsGettingHeatmapInfo() {
-    throw UnimplementedError(
-      'supportsGettingHeatmapInfo() has not been implemented.',
-    );
-  }
+  bool supportsGettingHeatmapInfo() => false;
```
jokerttu, line 120: This change is not ground overlay related, but as this method was not implemented for web platform, integration tests for web were failing because of this.
`supportsGettingHeatmapInfo() has not been implemented.`
We could have added implementation for this to web platform, but instead having this defaults to false, seems better solution.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_inspector_platform.dart
```diff
@@ -132,6 +128,20 @@ abstract class GoogleMapsInspectorPlatform extends PlatformInterface {
     throw UnimplementedError('getHeatmapInfo() has not been implemented.');
   }
 
+  /// If the platform supports getting information about ground overlays.
+  bool supportsGettingGroundOverlayInfo() => false;
```
jokerttu, line 132: This follows same pattern as heatmap implementation.
Having this method could be unnecessary, as all platforms support getting groung operlay info at the moment.
We can remove this method seen valid change.

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
ditman: Were they timing out in all tests, or wasm specifically? I know I had to revisit a bunch of plugins and add a few `pump`s and `pumpAndSettle` when enabling the wasm tests. See:

* https://github.com/flutter/packages/pull/8111

packages/google_maps_flutter/google_maps_flutter/example/integration_test/src/tiles_inspector.dart
ditman: (Same as above, resolving)

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_inspector_platform.dart
ditman: Sounds good to me, this is exactly what this "capability check" is supposed to do.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_inspector_platform.dart
ditman: I like that we're adding these capability checks to plugins, we should do this more.

stuartmorgan: Yep, doing this more is actually our written policy now :)

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/platform_interface/google_maps_inspector_platform.dart
stuartmorgan: Yes, this was a mistake; "supports"-style interface methods should default to false instead of throwing. Thanks for fixing it!

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
```diff
@@ -0,0 +1,155 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+@import google_maps_flutter_ios;
+@import google_maps_flutter_ios.Test;
+@import XCTest;
+@import GoogleMaps;
+
+#import <OCMock/OCMock.h>
+#import <google_maps_flutter_ios/FGMGroundOverlayController_Test.h>
```
stuartmorgan, line 11: Per style guide, this goes before all other imports, in its own section.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
```diff
@@ -0,0 +1,155 @@
...
+#import <google_maps_flutter_ios/messages.g.h>
+#import "PartiallyMockedMapView.h"
+
+@interface GoogleMapsGroundOverlayControllerTests : XCTestCase
+@end
+
+@implementation GoogleMapsGroundOverlayControllerTests
+
+/// Returns GoogleMapGroundOverlayController object instantiated with position and a mocked map
+/// instance
```
stuartmorgan, line 21: Nit: missing a period at the end of the sentence.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
```diff
@@ -0,0 +1,155 @@
...
+  PartiallyMockedMapView *mapView = [[PartiallyMockedMapView alloc] initWithOptions:mapViewOptions];
+
+  return [[FGMGroundOverlayController alloc] initWithGroundOverlay:groundOverlay
+                                                        identifier:@"id_1"
+                                                           mapView:mapView
+                                               isCreatedWithBounds:NO];
+}
+
+/// Returns GoogleMapGroundOverlayController object instantiated with bounds and a mocked map
+/// instance
```
stuartmorgan, line 49: Same.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
```diff
@@ -0,0 +1,155 @@
...
+@interface GoogleMapsGroundOverlayControllerTests : XCTestCase
+@end
+
+@implementation GoogleMapsGroundOverlayControllerTests
+
+/// Returns GoogleMapGroundOverlayController object instantiated with position and a mocked map
+/// instance
+///
+/// @return An object of FLTGoogleMapGroundOverlayController
+- (FGMGroundOverlayController *)groundOverlayControllerWithPositionWithMockedMap {
```
stuartmorgan, line 24: Do these need to be instance methods rather than class methods?

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.h
```diff
@@ -0,0 +1,57 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Flutter/Flutter.h>
+#import <GoogleMaps/GoogleMaps.h>
+
+#import "messages.g.h"
+
```
stuartmorgan, line 9: For new code, the repo preference is to use `NS_ASSUME_NONNULL_*` audited region macros around headers.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.h
```diff
@@ -0,0 +1,57 @@
...
+/// Controller of a single ground overlay  on the map.
+@interface FGMGroundOverlayController : NSObject
+
+/// The ground overlay this controller handles.
+@property(strong, nonatomic) GMSGroundOverlay *_Nonnull groundOverlay;
+
+/// Whether ground overlay is created with bounds or position.
+@property(nonatomic, assign) BOOL isCreatedWithBounds;
```
stuartmorgan, line 17: Properties should not be named "is*"; see https://google.github.io/styleguide/objcguide.html#objective-c-method-names

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.m
```diff
@@ -0,0 +1,240 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FGMGroundOverlayController.h"
+#import "FGMUtils.h"
```
stuartmorgan, line 6: Nit: please add a blank line above this; this corresponding header of an implementation file gets its own section.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.m
```diff
@@ -0,0 +1,240 @@
...
+  }
+}
+
+@end
+
+@interface FLTGroundOverlaysController ()
+
+/// A map from ground overlay id to the controller that manages it.
+@property(strong, nonatomic) NSMutableDictionary<NSString *, FGMGroundOverlayController *>
+    *groundOverlayIdentifierToController;
```
stuartmorgan, line 107: Nit: if you name it `groundOverlayControllerByIdentifier`, then code doing lookups into it will read more like a method, where the last thing names the argument (or in this case, the index).

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.m
```diff
@@ -0,0 +1,240 @@
...
+    if (!controller) {
+      continue;
+    }
+    [controller removeGroundOverlay];
+    [self.groundOverlayIdentifierToController removeObjectForKey:identifier];
+  }
+}
+
+- (void)didTapGroundOverlayWithIdentifier:(NSString *)identifier {
+  if (!identifier) {
```
stuartmorgan, line 202: Here and in the next method: the identifier params were annotated as non-nullable in the interface, so the implementation code shouldn't be handling nil.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMUtils.h
```diff
@@ -0,0 +1,18 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Flutter/Flutter.h>
+#import <GoogleMaps/GoogleMaps.h>
+#import "messages.g.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FGMUtils : NSObject
```
stuartmorgan, line 11: I strongly discourage "classes" that are just bags of class methods; Objective-C, unlike Java, has a perfectly good way of representing functions that aren't part of a class, which is functions :)

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMUtils.h
```diff
@@ -0,0 +1,18 @@
...
++ (UIImage *)iconFromBitmap:(FGMPlatformBitmap *)platformBitmap
```
stuartmorgan, line 12: These new APIs need declaration comments.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMUtils.h
stuartmorgan, line 1: Can this be given a more specific name? It's already the case that this plugin has at least one other file of utils (the conversion function file).

packages/google_maps_flutter/google_maps_flutter_ios/pigeons/messages.dart
```diff
@@ -357,6 +357,43 @@ class PlatformCameraTargetBounds {
...
+    required this.visible,
+    required this.clickable,
+    required this.zoomLevel,
+  });
+
+  final String groundOverlayId;
+  final PlatformBitmap image;
+  final PlatformLatLng? position;
+  final PlatformLatLngBounds? bounds;
+  final PlatformDoublePair? anchor;
```
stuartmorgan, line 388: Isn't this conceptually a point? We already have `PlatformPoint`, so shouldn't need a new `PlatformDoublePair` for this unless there's a conceptual difference in meaning here.

stuartmorgan (changes requested):
This looks really good! I did a pass over everything from web, and mostly just have minor comments and questions.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/ground_overlay.dart
```diff
@@ -0,0 +1,363 @@
...
+  /// To create ground overlay from assets, use [AssetMapBitmap],
+  /// [AssetMapBitmap.create] or [BitmapDescriptor.asset].
+  ///
+  /// To create ground overlay from raw PNG data use [BytesMapBitmap]
+  /// or [BitmapDescriptor.bytes].
+  ///
+  /// [MapBitmap.bitmapScaling] must be set to [MapBitmapScaling.none].
+  final MapBitmap image;
+
+  /// Geographical location to which the anchor will be fixed and the [width] of
```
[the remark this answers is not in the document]
jokerttu: All platforms, Android, iOS and Web differs a bit how GroundOverlays can be intitialized.
Using `bounds` work on each platform, but initializing with position (relative anchor position) is only supported by Android and iOS SDKs. We could also drop position support, and only support bounds if that seems better for unified api, but it would also limit a bit how devs can use the ground overlays when implementing only Android or iOS apps.

The consept of position/anchor, width/height/bounds is a bit hard to explain generally, without having platform specific documentation on platform_interface package.

The `position` of the is geo location of the `anchor` on the map. By default anchor is at position [0.5, 0.5] of the image. So if anchor is not changed and width of the image is given. half of the image is drawn to the west of the anchor position and half if the image to east (on android). If anchor is set to for example [0.0, 0.0] the whole image would be rendered `width`meters amount to the east of the `position`.

Should we add images to the docuementation like this: https://github.com/flutter/flutter/blob/63110570da9fff4a854414fe8910701a82aa3079/packages/flutter/lib/src/painting/box_fit.dart#L31

When initializing ground overlay with position developers need to give: 
-  On Android: `width` of the image in meters on map 
https://developers.google.com/maps/documentation/android-sdk/reference/com/google/android/libraries/maps/model/GroundOverlay
-  On iOS: `zoomLevel` 
https://developers.google.com/maps/documentation/ios-sdk/reference/objc/Classes/GMSGroundOverlay#+groundoverlaywithposition:icon:zoomlevel:

App-facing package has [support table](https://github.com/flutter/packages/pull/8432/files#diff-9bd2e43e18b051bf0a22c1fc3430348ca4b37755cdfd9b4dbb75a98ba156786bR231).

Current implementation also do not support changing the position/bounds/anchor of the groundoverlay after it's created. Not sure is there any use case really when this should be supported. Reason for this is that all platforms allow giving these values properly only on groundoverlay initialization, calculating anchor, position or bounds dynamically from other values could be not accurate. For example anchor can be only given on initialization, if that is changed later, the position would need to be calculated from bounds and updated. This could be implemented in future, if there is really need for it. Current workaround would be creating new groundoverlay instead of moving the existing.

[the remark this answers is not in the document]
jokerttu: @stuartmorgan updated documentation for the class and its parameters

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/ground_overlay.dart
```diff
@@ -0,0 +1,363 @@
...
+  /// the overlay (in meters). The [anchor] is, by default, 50% from the top of
+  /// the image and 50% from the left of the image. This can be changed by
+  /// with the [anchor] parameter.
+  /// You can optionally provide the [height] of the overlay (in meters).
```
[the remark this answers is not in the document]
jokerttu: Documentation update to clarify how overlays are scaled and placed to the map.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/ground_overlay.dart
```diff
@@ -0,0 +1,363 @@
...
+  final LatLng? position;
+
+  /// Width of the ground overlay (in meters).
+  final double? width;
+
+  /// Height of the ground overlay (in meters).
+  final double? height;
+
+  /// Bounds which will contain the image.
+  /// If [bounds] is specified, [position] must be null.
```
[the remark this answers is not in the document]
jokerttu: Documentation update to clarify how overlays are scaled and placed to the map.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/ground_overlay.dart
```diff
@@ -0,0 +1,363 @@
...
+  final LatLngBounds? bounds;
+
+  /// The image point is specified in normalized coordinates: An anchor of
+  /// (0.0, 0.0) means the top left corner of the image. An anchor
+  /// of (1.0, 1.0) means the bottom right corner of the image.
+  /// Defaults to (0.5, 0.5), which is the center of the image.
+  final Offset? anchor;
```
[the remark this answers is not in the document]
jokerttu: Documentation update to clarify how overlays are scaled and placed to the map.

packages/google_maps_flutter/google_maps_flutter_ios/pigeons/messages.dart
illuminati1911: Changed this to PlatformPoint.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMUtils.h
illuminati1911: I renamed it `FGMImageUtils`.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.m
illuminati1911: Good catch. Removed the null check.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.m
illuminati1911: Changed the name.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.m
illuminati1911: Done.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.h
illuminati1911: Changed it like this:
```objc
@property(nonatomic, assign, getter=isCreatedWithBounds) BOOL createdWithBounds;
```

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.h
illuminati1911: Added.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
illuminati1911: Actually no. Changed them to class methods.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
illuminati1911: Fixed.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
illuminati1911: Fixed.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
illuminati1911: Done.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMUtils.h
illuminati1911: My original idea was to use the class for namespacing, but I have now added the functions as global function with FGM prefix.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
stuartmorgan: It should be before *all* other imports; see https://google.github.io/styleguide/objcguide.html#order-of-includes

packages/google_maps_flutter/google_maps_flutter_ios/example/ios14/ios/RunnerTests/GoogleMapsGroundOverlayControllerTests.m
```diff
@@ -0,0 +1,156 @@
...
+// found in the LICENSE file.
+
+@import google_maps_flutter_ios;
+@import google_maps_flutter_ios.Test;
+@import XCTest;
+@import GoogleMaps;
+
+#import <google_maps_flutter_ios/FGMGroundOverlayController_Test.h>
+
+#import <OCMock/OCMock.h>
```
stuartmorgan, line 12: This should be grouped with the other framework imports. Grouping is by type of thing import, not by `@` vs `#`.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.h
```diff
@@ -0,0 +1,61 @@
...
+/// Removes ground overlays from the map.
+- (void)removeGroundOverlaysWithIdentifiers:(NSArray<NSString *> *)identifiers;
+
+/// Called when a ground overlay is tapped on the map.
+- (void)didTapGroundOverlayWithIdentifier:(NSString *)identifier;
+
+/// Returns true if a ground overlay with the given identifier exists on the map.
+- (bool)hasGroundOverlaysWithIdentifier:(NSString *)identifier;
+
+/// Returns FGMPlatformGroundOverlay for identifier.
```
stuartmorgan, line 57: "Returns the ground overlay with the given identifier."

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.h
```diff
@@ -0,0 +1,61 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import <Flutter/Flutter.h>
+#import <GoogleMaps/GoogleMaps.h>
```
stuartmorgan, line 6: There should be an import of `Foundation` here, as this header uses several Foundation types.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.m
```diff
@@ -0,0 +1,234 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FGMGroundOverlayController.h"
+
```
stuartmorgan, line 6: This file uses things from UIKit and CoreLocation, so needs to import them.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.m
```diff
@@ -0,0 +1,234 @@
...
+- (void)setIcon:(UIImage *)icon {
+  self.groundOverlay.icon = icon;
+}
+
+- (void)updateFromPlatformGroundOverlay:(FGMPlatformGroundOverlay *)groundOverlay
+                              registrar:(NSObject<FlutterPluginRegistrar> *)registrar
+                            screenScale:(CGFloat)screenScale {
+  [self setConsumeTapEvents:groundOverlay.clickable];
+  [self setVisible:groundOverlay.visible];
```
stuartmorgan, line 78: Based on investigation of similar methods in a recent PR (which hasn't landed), this needs to be last to avoid potential flickering (likely due to Google Maps using its own render thread), so should be at the end of the method, and have a comment saying that nothing should be added after it.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.m
stuartmorgan: It's still here in this method.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMImageUtils.m
```diff
@@ -0,0 +1,229 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FGMImageUtils.h"
+#import "Foundation/Foundation.h"
+
+static UIImage *scaleImage(UIImage *image, double scale);
+static UIImage *scaledImageWithScale(UIImage *image, CGFloat scale);
+static UIImage *scaledImageWithSize(UIImage *image, CGSize size);
+static UIImage *scaledImage(UIImage *image, NSNumber *width, NSNumber *height, CGFloat screenScale);
```
stuartmorgan, line 11: The declaration comments for all of these functions should be here, on the declarations, rather than on the definitions below.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMImageUtils.m
```diff
@@ -0,0 +1,229 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "FGMImageUtils.h"
+#import "Foundation/Foundation.h"
+
+static UIImage *scaleImage(UIImage *image, double scale);
```
stuartmorgan, line 8: This should be `scaled`, not `scale`, for consistency in naming.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FLTGoogleMapJSONConversions.m
```diff
@@ -113,6 +113,57 @@ GMSMapViewType FGMGetMapViewTypeForPigeonMapType(FGMPlatformMapType type) {
                      markerIds:markerIDs];
 }
 
+FGMPlatformGroundOverlay *FGMGetPigeonGroundOverlay(GMSGroundOverlay *groundOverlay,
+                                                    NSString *overlayId, BOOL isCreatedWithBounds,
+                                                    NSNumber *zoomLevel) {
+  /// Dummy image is used as image is required field of FGMPlatformGroundOverlay and converting
+  /// image back to bitmap image is not currently supported.
+  FGMPlatformBitmap *mockImage =
```
stuartmorgan, line 121: This should be called `placeholderImage` rather than `mockImage`, as "mock" generally refers to a specific test construct.

stuartmorgan (changes requested):
Just some scattered minor issues remaining. Feel free to split out the first sub-PR at this point!

jokerttu:
I've addressed the last remaining review comments and split the platform-specific implementations into a separate PR: https://github.com/flutter/packages/pull/8563.
Apologies for the delay!

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+   *     overlay options.
+   * @param sink the GroundOverlaySink where the options will be set.
+   * @param assetManager An instance of Android's AssetManager, which provides access to any raw
+   *     asset files stored in the application's assets directory.
+   * @param density the density of the display, used to calculate pixel dimensions.
+   * @param wrapper the BitmapDescriptorFactoryWrapper to create BitmapDescriptor.
+   * @return the identifier of the ground overlay.
+   * @throws IllegalArgumentException if required fields are missing or invalid.
+   */
+  static String interpretGroundOverlayOptions(
```
reidbaker, line 866: Please add non null or nullable annotations to the objects in this method.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+   * Set the options in the given ground overlay object to the given sink.
+   *
+   * @param groundOverlay the object expected to be a PlatformGroundOverlay containing the ground
+   *     overlay options.
+   * @param sink the GroundOverlaySink where the options will be set.
+   * @param assetManager An instance of Android's AssetManager, which provides access to any raw
+   *     asset files stored in the application's assets directory.
+   * @param density the density of the display, used to calculate pixel dimensions.
+   * @param wrapper the BitmapDescriptorFactoryWrapper to create BitmapDescriptor.
+   * @return the identifier of the ground overlay.
```
reidbaker, line 863: If you know can you add more about the mutability of the ground overlay identifier. For example how long it is a valid id.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+    if (groundOverlay.getAnchor() != null) {
+      sink.setAnchor(
+          groundOverlay.getAnchor().getX().floatValue(),
+          groundOverlay.getAnchor().getY().floatValue());
+    }
+    sink.setBearing(groundOverlay.getBearing().floatValue());
+    sink.setClickable(groundOverlay.getClickable());
+    sink.setImage(toBitmapDescriptor(groundOverlay.getImage(), assetManager, density, wrapper));
+    if (groundOverlay.getPosition() != null) {
+      assert groundOverlay.getWidth() != null;
```
reidbaker, line 884: Asserts I believe are compiled out so the else condition can still NPE here. Can you convert this assert to a check or handle null width?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+      if (groundOverlay.getHeight() != null) {
```
reidbaker, line 885: I believe these are the same except the width value. Can you minimize the scope by either making the check cover width only?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+                    .setImagePixelRatio(1.0)
+                    .setBitmapScaling(Messages.PlatformMapBitmapScaling.NONE)
+                    .build())
+            .build();
+
+    Messages.PlatformGroundOverlay.Builder builder =
+        new Messages.PlatformGroundOverlay.Builder()
+            .setGroundOverlayId(groundOverlayId)
+            .setImage(dummyImage)
+            .setWidth((double) groundOverlay.getWidth())
```
reidbaker, line 931: Why use doubles here but floats above for width?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+            .setHeight((double) groundOverlay.getHeight())
+            .setBearing((double) groundOverlay.getBearing())
+            .setTransparency((double) groundOverlay.getTransparency())
+            .setZIndex((long) groundOverlay.getZIndex())
+            .setVisible(groundOverlay.isVisible())
+            .setClickable(groundOverlay.isClickable());
+
+    if (isCreatedWithBounds) {
```
reidbaker, line 939: It is possible to determine this instead of having it be passed in? For example could a non null getBounds or getPosition indicate that one of those should be used? 

The code above in `interpretGroundOverlayOptions` seem to imply that position is more important that bounds. This code seem to imply that if bounds are used then position should not be.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GroundOverlaySink.java
```diff
@@ -0,0 +1,32 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.android.gms.maps.model.BitmapDescriptor;
+import com.google.android.gms.maps.model.LatLng;
+import com.google.android.gms.maps.model.LatLngBounds;
+
+/** Receiver of GroundOverlayOptions configuration. */
+interface GroundOverlaySink {
+  void setTransparency(float transparency);
```
reidbaker, line 15: Can you add javadocs to these methods for where someone can learn more about the inputs either here or in groundOverlayController

For example what are the min and max values for transparency. Can zindex be negative? what does setClickable do? etc.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GroundOverlaysController.java
```diff
@@ -0,0 +1,139 @@
...
+          density,
+          bitmapDescriptorFactoryWrapper);
+    }
+  }
+
+  private void removeGroundOverlay(@NonNull String groundOverlayId) {
+    GroundOverlayController groundOverlayController =
+        groundOverlayIdToController.get(groundOverlayId);
+    if (groundOverlayController != null) {
+      groundOverlayController.remove();
```
reidbaker, line 115: Does the order of these removals matter? if so please add a comment explaining the order.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GroundOverlaysController.java
```diff
@@ -0,0 +1,139 @@
...
+      groundOverlayIdToController.put(groundOverlayId, groundOverlayController);
+      googleMapsGroundOverlayIdToDartGroundOverlayId.put(groundOverlay.getId(), groundOverlayId);
+    }
+  }
+
+  private void changeGroundOverlay(@NonNull Messages.PlatformGroundOverlay platformGroundOverlay) {
+    String groundOverlayId = platformGroundOverlay.getGroundOverlayId();
+    GroundOverlayController groundOverlayController =
+        groundOverlayIdToController.get(groundOverlayId);
+    if (groundOverlayController != null) {
```
reidbaker, line 101: It is an error to call this with something that cant get a groundOverlayController? If so whos error is it (plugin or host app)?

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
jokerttu: Pigeon uses doubles and google maps uses floats, so conversion between these need to be done while converting values between PlatformGroundOverlay and GoogleMaps SDK GroundOverlay object

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
jokerttu: Google maps object internally have both, bounds and position when the object is created using either of them. When bounds are used, the position representes the achor point in LatLng format for the ground overlay. And if position (with width/height) the bounds are calculated automatically as well.
For integration testing purposes we need track which creation method was used, to be able to check that conversions are handled correctly on each platform.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GroundOverlaysController.java
jokerttu: Order does not matter here

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GroundOverlaysController.java
jokerttu: Api has updateGroundOverlays method takes GroundOverlayUpdates object as parameter:
https://github.com/flutter/packages/blob/ff7724c18a803542fc709f942c3bf1cecf7e4864/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/ground_overlay_updates.dart
This should be handled by the dart code so that changes only contains items that are available on map.
Not sure would it be good to throw error here if change is triggered for groundoverlay that is not added to the map beforehand. It is indeed misuse of the api, and this is tested in integration tests.

jokerttu:
@stuartmorgan This PR now contains only app-facing plugin changes.
Had to update minimum supported SDK versions to Flutter 3.27/Dart 3.6 to match the android package:
https://github.com/flutter/packages/blob/297d5a151fdbabc11a4e0c87d82dca637afa7ecd/packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml#L7-L9

stuartmorgan (approved):
LGTM!

PR #8466, opened 2025-01-21
[video_player_android] Platform view support

FirentisTFW:
This PR adds support for platform views on Android as a way of displaying a video. When creating a video, it's now possible to choose between texture view approach (rendered using `Texture` widget on the Flutter side) and platform view approach (rendered on the native side, using `SurfaceView`).

`VideoPlayer` class now has nothing to do with texture. The texture-related code was moved from it to `TextureVideoPlayer` - a subclass of `VideoPlayer` that adds texture functionality. There's also `PlatformViewVideoPlayer` - a subclass of `VideoPlayer` that adds platform view functionality. In the plugin class (`create` method) we create either the platform view version or the texture version based on the parameter (`viewType`) passed in from Flutter side.

The functionality is not yet exposed in the app-facing package (only in the example app) - it will be done in a separate PR. The PR does not introduce breaking changes.

Related issues:
- [#86613](https://github.com/flutter/flutter/issues/86613) - this issue requests platform view support for iOS. Here we also add it for the Android, so that it works on both mobile platforms.

## Pre-launch Checklist

- [x] I signed the [CLA].


jonahwilliams:
> to choose between texture view approach (rendered using Texture widget on the Flutter side) and platform view approach (rendered on the native side, using SurfaceView).

On Android all platform views are rendered via texture composition. So switching to the SurfaceView actually adds an extra compositing step compared to the TextureView, and won't be any more efficient (it will be less efficient).

stuartmorgan:
Supporting platform views isn't just about which is more efficient, they don't always have the same behaviors. E.g., the issues that reference this just above where some playback works in this mode but not with Texture. And on iOS there is no path to supporting DRM with Texture; I haven't verified recently for Android but I would imagine that it's true there as well (for the same reason).

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -9,31 +9,29 @@
...
 import androidx.media3.common.MediaItem;
 import androidx.media3.common.PlaybackParameters;
 import androidx.media3.exoplayer.ExoPlayer;
-import io.flutter.view.TextureRegistry;
 
-final class VideoPlayer implements TextureRegistry.SurfaceProducer.Callback {
+/**
+ * A class responsible for managing video playback using {@link ExoPlayer}. It provides methods to
+ * control playback, adjust volume, and handle seeking.
+ */
```
stuartmorgan, line 22: Thanks for fixing the missing declaration comment while refactoring :)

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -72,52 +68,37 @@ interface ExoPlayerProvider {
...
+        new ExoPlayerEventListener(
+            exoPlayer, videoPlayerEvents, getViewType(), wasPlayerInitialized()));
     setAudioAttributes(exoPlayer, options.mixWithOthers);
 
     return exoPlayer;
   }
 
+  protected boolean wasPlayerInitialized() {
+    // Can be overridden in subclasses.
+    return false;
```
stuartmorgan, line 95: This seems confusing; a platform-view-based player can definitely be initialized. The old naming was in the context of a local variable, where it was checking if it had been initialized previously to the create method in which the code was located, whereas as a method I would expect it to report current state whenever called.

How about `playerHasBeenSuspended`, with the comment here being `// The base implementation never suspends and re-creates the exoPlayer, but subclasses may."?

(It's a bit clunky since it embeds a certain knowledge of subclass behaviors in the superclass, but I think it's much clearer about what the return value means.)

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -161,12 +142,12 @@ long getPosition() {
     return exoPlayer.getCurrentPosition();
   }
 
+  @NonNull
+  ExoPlayer getExoPlayer() {
```
stuartmorgan, line 146: Why do we need a public getter, given that the ivar is now `protected`?

packages/video_player/video_player_android/example/pubspec.yaml
```diff
@@ -9,16 +9,19 @@ environment:
...
+    sdk: flutter
   video_player_android:
     # When depending on this package from a real application you should use:
     #   video_player_android: ^x.y.z
     # See https://dart.dev/tools/pub/dependencies#version-constraints
     # The example app is bundled with the plugin so we use a path dependency on
     # the parent directory to use the current plugin's version.
     path: ../
-  video_player_platform_interface: ">=6.1.0 <7.0.0"
+  video_player_platform_interface: ">=6.3.0 <7.0.0"
```
stuartmorgan, line 21: `^6.3.0`

stuartmorgan (commented):
This should definitely get an Android team review as well, but just minor notes from my review!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
FirentisTFW: It's used in `NativeVideoViewFactory` to obtain player's `ExoPlayer` for the platform view.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
FirentisTFW: I also added some additional info on when to use `VideoPlayer` and when - `TextureBasedVideoPlayer`.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
FirentisTFW: You're right, good point. I did as you suggested

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/TextureBasedVideoPlayer.java
```diff
@@ -16,7 +16,8 @@
 /**
  * A subclass of {@link VideoPlayer} that adds functionality related to texture-based view as a way
  * of displaying the video in the app. It manages the lifecycle of the texture and ensures that the
- * video is properly displayed on the texture.
+ * video is properly displayed on the texture. /// If you need to display a video using platform
```
stuartmorgan, line 19: The `///` shouldn't be here.

stuartmorgan (commented):
Changes all look good! I'll leave final approval to the Android team review.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/TextureBasedVideoPlayer.java
FirentisTFW: Of course, good catch

packages/video_player/video_player_android/example/android/app/src/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,20 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
```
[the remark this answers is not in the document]
FirentisTFW: I think so, it's used for native Android UI test with Espresso (though we currently `@Ignore` it anyway since it doesn't work on Firebase). I implemented it the same way it was done for other packages (e.g. `file_selector`).

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/NativeVideoView.java
```diff
@@ -0,0 +1,51 @@
...
+class NativeVideoView implements PlatformView {
+  @NonNull private final SurfaceView surfaceView;
+
+  /**
+   * Constructs a new NativeVideoView.
+   *
+   * @param context The context in which the view is running.
+   * @param exoPlayer The ExoPlayer instance used to play the video.
+   */
+  NativeVideoView(@NonNull Context context, @NonNull ExoPlayer exoPlayer) {
```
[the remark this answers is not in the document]
FirentisTFW: Added

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/NativeVideoViewFactory.java
```diff
@@ -0,0 +1,53 @@
...
+   * Creates a new instance of platform view.
+   *
+   * @param context The context in which the view is running.
+   * @param id The unique identifier for the view.
+   * @param args The arguments for creating the view.
+   * @return A new instance of NativeVideoView.
+   */
+  @NonNull
+  @Override
+  public PlatformView create(@NonNull Context context, int id, @Nullable Object args) {
```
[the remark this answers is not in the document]
FirentisTFW: I'm afraid it's not really possible to nicely test it the way you described. We don't really call this method explicitly from within the plugin, it gets called as a result of creating a platform view on the Flutter side. We also don't store a reference to the factory once we register it.

However, I added unit tests for `NativeVideoViewFactory` and `NativeVideoView` to test them more in isolation, please have a look.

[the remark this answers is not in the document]
FirentisTFW: The flow you described is also covered in Espresso unit test, but looks like we can't run it on Firebase Test Lab (seems to be a common issue across the packages).

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/TextureBasedVideoPlayer.java
matanlurey, line 1: Ditto above, let's put in `videoplayer/texture` and use consistent terminology (we don't say `Texture`_Based_ anywhere else, so let's just prefix this `Texture`.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -9,31 +9,32 @@
...
 import androidx.annotation.VisibleForTesting;
 import androidx.media3.common.AudioAttributes;
 import androidx.media3.common.C;
 import androidx.media3.common.MediaItem;
 import androidx.media3.common.PlaybackParameters;
 import androidx.media3.exoplayer.ExoPlayer;
-import io.flutter.view.TextureRegistry;
 
-final class VideoPlayer implements TextureRegistry.SurfaceProducer.Callback {
+/**
```
matanlurey, line 19: Thanks for adding docs!

A few notes:
- Keep the first line standalone: `* A class responsible for managing video playback using {@link ExoPlayer}.`
- In the second paragraph, keep most of the information.
- Avoid _you need to_, this is not a public API, there is never a "you".

I don't think we should have this class implicitly be platform views, and we should use the same strategy from the event listener - let's make this an abstract class with two concrete implementations instead (more on this below).

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -9,31 +9,32 @@
...
+ * A class responsible for managing video playback using {@link ExoPlayer}. It provides methods to
+ * control playback, adjust volume, and handle seeking. This class contains all functionalities
+ * needed to manage video playback in platform views and is typically used alongside {@link
+ * NativeVideoViewFactory}. If you need to display a video using a texture, use {@link
+ * TextureBasedVideoPlayer} instead.
+ */
+class VideoPlayer {
```
matanlurey, line 26: Make this an abstract class, and similar to above let's do:

```txt
videoplayer/
  texture/
    TextureVideoPlayer.java
  platformview/
    PlatformViewVideoPlayer.java
```

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -72,52 +71,37 @@ interface ExoPlayerProvider {
...
+    exoPlayer.addListener(
+        new ExoPlayerEventListener(
+            exoPlayer, videoPlayerEvents, getViewType(), playerHasBeenSuspended()));
     setAudioAttributes(exoPlayer, options.mixWithOthers);
 
     return exoPlayer;
   }
 
+  protected boolean playerHasBeenSuspended() {
+    // The base implementation never suspends and re-creates the exoPlayer, but subclasses may.
```
matanlurey, line 97: This comment is better suited as API documentation to the method itself, i.e. explain what it means and when it is true or false. Once you refactor this into a base class I'd like to review this again to make sure it makes sense.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -72,52 +71,37 @@ interface ExoPlayerProvider {
...
+    return false;
+  }
+
+  protected Messages.PlatformVideoViewType getViewType() {
```
matanlurey, line 101: This is a weird API, but I think it gets fixed by splitting the classes like discussed.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
```diff
@@ -23,6 +23,12 @@ public class VideoPlayerPlugin implements FlutterPlugin, AndroidVideoPlayerApi {
   private FlutterState flutterState;
   private final VideoPlayerOptions options = new VideoPlayerOptions();
 
+  /**
+   * The next non-texture player ID, initialized to a high number to avoid collisions with texture
+   * IDs (which are generated separately).
+   */
+  private Long nextNonTexturePlayerId = Long.MAX_VALUE;
```
matanlurey, line 30: I would prefer to see consistent terminology. Let's call this `nextPlatformViewPlayerId` like before.

I'm also curious, was this same strategy chosen for iOS?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
```diff
@@ -107,25 +114,46 @@ public void initialize() {
...
   @NonNull
-  private VideoPlayer getPlayer(long textureId) {
-    VideoPlayer player = videoPlayers.get(textureId);
+  private VideoPlayer getPlayer(long playerId) {
+    VideoPlayer player = videoPlayers.get(playerId);
 
     // Avoid a very ugly un-debuggable NPE that results in returning a null player.
     if (player == null) {
-      String message = "No player found with textureId <" + textureId + ">";
+      String message = "No player found with playerId <" + playerId + ">";
```
matanlurey, line 156: Since a goal here is to create a readable message, we should also include whether it is a texture or platform view.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
```diff
@@ -136,53 +164,53 @@ private VideoPlayer getPlayer(long textureId) {
   }
 
   @Override
-  public void dispose(@NonNull Long textureId) {
-    VideoPlayer player = getPlayer(textureId);
+  public void dispose(@NonNull Long playerId) {
```
matanlurey, line 167: Less important to resolve _in_ this PR, but if the base class's implementation is now `playerId` (since video players are not always textures anymore), can the base implementation also be changed? A new issue would be fine.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerProvider.java
matanlurey, line 1: Can this instead be a [static, nested] interface private to `PlatformViewVideoFactory` (or whatever we end up calling it) since it's not used anywhere else?

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/ExoPlayerEventListenerTest.java
matanlurey, line 1: Once we fork the classes, the tests should fork too, this is terribly hard to read and maintain.

That is, we'd have a `TextureExoPlayer...Test` and `PlatformViewExoPlayer...Test`.

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/TextureBasedVideoPlayerTest.java
matanlurey, line 1: Ditto naming conventions discussed elsewhere in this PR.

packages/video_player/video_player_android/example/lib/mini_controller.dart
```diff
@@ -83,6 +85,9 @@ class VideoPlayerValue {
   /// Indicates whether or not the video has been loaded and is ready to play.
   final bool isInitialized;
 
+  /// Degrees to rotate the video (clockwise) so it is displayed correctly.
+  final int rotationCorrection;
```
matanlurey, line 89: Can you clarify why this was added in this PR? Is this a general cleanup or now required?

If the former I'd prefer a separate PR.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -4,20 +4,26 @@
...
-
 /// An Android implementation of [VideoPlayerPlatform] that uses the
 /// Pigeon-generated [VideoPlayerApi].
 class AndroidVideoPlayer extends VideoPlayerPlatform {
   final AndroidVideoPlayerApi _api = AndroidVideoPlayerApi();
 
+  /// A map that associates player ID with a view state.
+  /// This is used to determine which view type to use when building a view.
+  @visibleForTesting
+  final Map<int, VideoPlayerViewState> playerViewStates =
```
matanlurey, line 24: I would prefer not to make new `@visibleForTesting` APIs, this tells me the end result is not ideally testable.

Instead, let's test the _behavior, that is:
- When we create with inputs X, we observe output Y

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -29,12 +35,27 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
...
+    await _api.dispose(playerId);
+    playerViewStates.remove(playerId);
+  }
+
+  @override
+  Future<int?> create(DataSource dataSource) {
+    return createWithOptions(
+      VideoCreationOptions(
+        dataSource: dataSource,
+        // Texture view was the only supported view type before
```
matanlurey, line 48: nit: Good comment. Could be a tad clearer:

```dart
// Compatibility; "create" is always a textureView" (createWithOptions allows selecting).
```

... or something like that,.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -145,17 +175,73 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
...
+      VideoPlayerTextureViewState(:final int textureId) =>
+        Texture(textureId: textureId),
+      VideoPlayerPlatformViewState() => _buildPlatformView(playerId),
+      null => throw Exception(
+          'Could not find corresponding view type for playerId: $playerId',
+        ),
+    };
+  }
+
+  Widget _buildPlatformView(int playerId) {
```
matanlurey, line 199: I'd prefer to see this as a standalone `StatelessWidget` than a nested `_buildX` function.

Maybe `lib/src/platform_view_player.dart` (or similar).

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -145,17 +175,73 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
...
+    const String viewType = 'plugins.flutter.dev/video_player_android';
+    final PlatformVideoViewCreationParams creationParams =
+        PlatformVideoViewCreationParams(playerId: playerId);
+
+    return Builder(
+      builder: (BuildContext context) => IgnorePointer(
+        // IgnorePointer so that GestureDetector can be used above the platform view.
```
matanlurey, line 206: Add this comment before the use of IgnorePointer instead of after.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -174,3 +260,46 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
...
+  VideoViewType viewType,
+) {
+  return switch (viewType) {
+    VideoViewType.textureView => PlatformVideoViewType.textureView,
+    VideoViewType.platformView => PlatformVideoViewType.platformView,
+  };
+}
+
+/// Base class representing the state of a video player view.
+@visibleForTesting
```
matanlurey, line 274: Similar to comments above, make this private, and don't make it `@visibleForTesting`.

Let's observe the output instead of querying the internal state for testing.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -174,3 +260,46 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
...
+/// Represents the state of a video player view that uses a texture.
+@visibleForTesting
+final class VideoPlayerTextureViewState extends VideoPlayerViewState {
+  /// Creates a new instance of [VideoPlayerTextureViewState].
+  const VideoPlayerTextureViewState({
+    required this.textureId,
+  });
+
+  /// The ID of the texture used by the video player.
+  final int textureId;
```
matanlurey, line 289: I don't really understand.

Can't this be simplified down to:

```dart
// class AndroidVideoPlayer
final Map<int, _VideoPlayerViewType> _playerIdToViewType

// Below:
enum _VideoPlayerViewType {
  texture,
  platformView;
}
```

Given the ID is always the key, it seems like we're storing extra information for no value (maybe I missed something?).

packages/video_player/video_player_android/lib/src/android_video_player.dart
matanlurey: Can we also clarify (re-state) that we don't expect collisions because platform view IDs start at MAX_INT, and  texture IDs start at 0??

packages/video_player/video_player_android/test/android_video_player_test.dart
```diff
@@ -186,25 +194,146 @@ void main() {
...
+            uri: 'someUri',
+          ),
+          viewType: VideoViewType.platformView,
+        ),
+      );
+      expect(log.log.last, 'create');
+      expect(log.passedCreateMessage?.viewType,
+          PlatformVideoViewType.platformView);
+      expect(playerId, 3);
+      expect(player.playerViewStates[3], const VideoPlayerPlatformViewState());
```
matanlurey, line 317: This is example, we should be testing the output of `<Widget>.build()`, not querying the internal state.

matanlurey (changes requested):
This review is predicated on this overall approach being something we want to do (I have not been involved with "let's use platform views to display instead of texture views", so I delegate that to others).

In general the changes LG.

My major concerns, and where the review is focused, is:
- Consistent terminology being used across this PR;
- Making it more clear what code (particularly on the Java side) is texture v platform views
- Testing the behavior of the code instead of the internal state

packages/video_player/video_player_android/lib/src/android_video_player.dart
stuartmorgan: > Given the ID is always the key, it seems like we're storing extra information for no value

This is a structure I requested. In the past, player ID and texture ID were always the same thing. In the new structure, having those be conflated in one mode makes the code much more confusing in practice because it requires passing something called `playerId` into APIs that expect a `textureId`, and so the knowledge that these are (in one mode) the same value has to be essentially "hard coded" in multiple parts of the code—including both Dart and Java.

I am happy to pay the cost of one int per player (which in practice is a small number) to have clearer code, and code that will not be fragile to later changes to how texture ID is managed.

stuartmorgan:
> This review is predicated on this overall approach being something we want to do (I have not been involved with "let's use platform views to display instead of texture views", so I delegate that to others).

It's not "instead", it's a second option. We do want to do offer both, because in practice there are different tradeoffs to texture vs platform view, and we already have examples of cases where this is working better.

packages/video_player/video_player_android/lib/src/android_video_player.dart
matanlurey: Thanks @stuartmorgan , that makes sense to me.

@FirentisTFW:

Given that, I guess my confusion is `nextNonTexturePlayerId` (above), why is it important to increment separate numbers for texture views versus platform views? It looks like we should be incrementing playerIds regardless of their type (texture or platform), with the added bit that texture views _also_ store their texture IDs.

matanlurey:
> It's not "instead", it's a second option

Right sorry, imprecise language - I just mean "I'm not reviewing the _validity_ of this change" 👍🏼

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/NativeVideoViewFactory.java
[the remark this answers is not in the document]
camsim99: Gotcha. Thanks for the explanation and additional tests! They lgtm :)

FirentisTFW:
@matanlurey Thanks for a thorough review, I'll try to address it fully tomorrow. One important thing to note here is that some things that are done in this PR are based on iOS implementation, so, for the sake of consistency, we might need to revisit iOS implementation and/or consider some trade-offs here. I'll leave a comment on that when I'm done with addressing the PR feedback

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
FirentisTFW: Could you clarify what exactly you had in mind here? What's the base implementation here and what changes would you see there?

packages/video_player/video_player_android/example/lib/mini_controller.dart
FirentisTFW: Yes, that's a general clean-up. I added this widget to the example app to apply the rotation correction on the video. It is used in the example app of app-facing package (`video_player`), but it wasn't added here. Because of that, it was harder to spot any issues with rotated videos (see [#159724](https://github.com/flutter/flutter/issues/159724)). I think it's a rather small change, don't you think it could stay in this PR?

matanlurey: Yeah, thanks for explaining, sounds good.

matanlurey:
> @matanlurey Thanks for a thorough review, I'll try to address it fully tomorrow. One important thing to note here is that some things that are done in this PR are based on iOS implementation, so, for the sake of consistency, we might need to revisit iOS implementation and/or consider some trade-offs here. I'll leave a comment on that when I'm done with addressing the PR feedback

Sounds good. I won't block based on decisions already made.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
FirentisTFW: We can't provide this information here - the purpose of this method is to provide player based on its ID. Since IDs are unique regardless of the view type, no view type information is needed for that. And if we can't find the player for a given ID, there's no view type information we could use here

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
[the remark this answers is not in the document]
FirentisTFW: Good point, I did as suggested

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -74,6 +82,36 @@ private void sendInitialized() {
...
+    if (viewType == Messages.PlatformVideoViewType.PLATFORM_VIEW) {
+      sendInitializedForPlatformViewBased();
+    } else {
+      sendInitializedForTextureBased();
+    }
+  }
+
+  @OptIn(markerClass = UnstableApi.class)
+  private void sendInitializedForPlatformViewBased() {
+    // We can't rely on VideoSize here, because at this point it is not available - the platform
```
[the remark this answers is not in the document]
FirentisTFW: Unfortunately, I don't think I can add much to the current comment. The `getVideoSize` method is not yet available, because the view was not created yet (it happens only after we send initialized event tom Flutter from this method).

Does it work the same way as `exoPlayer.getVideoSize()` - I can't be 100% sure here. I've tried it on multiple Android versions (from 7 to 15) and didn't observe any issues there. However, in code there's [a comment](https://github.com/flutter/packages/pull/7846/files#diff-78d4d55b23e0750747d4f6a68581b1c0ce7b6e6868e260b351c72e6007d5cafeR148-R150) initially introduced by that says it could differ across different API versions. Maybe @camsim99 (who wrote this comment) could shed some light on that and tell whether this is something we should be concerned about here?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -74,6 +82,36 @@ private void sendInitialized() {
...
+    // view was not created yet. We use the video format instead.
+    Format videoFormat = exoPlayer.getVideoFormat();
+    RotationDegrees rotationCorrection =
+        RotationDegrees.fromDegrees(Objects.requireNonNull(videoFormat).rotationDegrees);
+    int width = videoFormat.width;
+    int height = videoFormat.height;
+
+    if (rotationCorrection == RotationDegrees.ROTATE_90
```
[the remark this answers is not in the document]
FirentisTFW: Added

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/NativeVideoView.java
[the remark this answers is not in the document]
FirentisTFW: Changes pushed, I used `PlatformView` everywhere

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/NativeVideoView.java
```diff
@@ -0,0 +1,51 @@
...
+    surfaceView = new SurfaceView(context);
+    // The line below is needed to display the video correctly on older Android versions (blank
```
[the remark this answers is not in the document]
FirentisTFW: Added API guard and improved the comment. I also spot one issue on Android 9 (or at least it happened on my emulator, I could not verify it with a real device): first frame was not visible if the video was initially paused, so I had to handle that differently (using `setVideoSurface` instead of `setVideoSurfaceView`. Please have a look

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
FirentisTFW: Made it abstract and added two concrete implementations for different view types.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
FirentisTFW: Done, now `VideoPlayer` is not used directly anymore, there's `PlatformViewVideoPlayer`.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
FirentisTFW: After splitting `VideoPlayer` into two classes, we could actually get rid of it in a base class, now it's just in `TextureVideoPlayer`, where it belongs

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
FirentisTFW: Yes, got removed when splitting classes

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerProvider.java
FirentisTFW: I put it inside `PlatformViewVideoFactory`, I think it can't be `private` though since we use it to instantiate `PlatformViewVideoFactory` (I'm not sure if you referred to `private` in a meaning of access modifiers or placement in code).

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/ExoPlayerEventListenerTest.java
FirentisTFW: Yep, good idea with that, these tests now got split into three files.

packages/video_player/video_player_android/lib/src/android_video_player.dart
FirentisTFW: Right, I updated unit tests as suggested.

What would you like to restate here? I believe it should be rather clear that IDs are unique here since we store them as keys in a map. As of now, the Dart code has no knowledge of how the IDs are generated on the native side, we just treat them as unique identifiers of players, so I'm not sure if including such native implementation details has any benefits here.

packages/video_player/video_player_android/lib/src/android_video_player.dart
FirentisTFW: Updated comment as suggested.

packages/video_player/video_player_android/lib/src/android_video_player.dart
FirentisTFW: Extracted to a separate widget.

packages/video_player/video_player_android/test/android_video_player_test.dart
FirentisTFW: Changed to check for a widget type we get when building a view.

packages/video_player/video_player_android/lib/src/android_video_player.dart
FirentisTFW: I'd say we do that for now to reduce scope. If we used one counter for texture and platform views, we'd have to additionally store texture IDs in case of texture views. Then we'd need to pass them to Flutter which would require changes in the pigeon API.

There is [a plan](https://github.com/flutter/packages/pull/8237#discussion_r1904490223) to decouple the IDs eventually (I suspect that they might be generated on a Dart side then), there's a TODO for it in iOS code, now I also added a TODO to Android code (above `nextPlatformViewPlayerId`)

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -68,90 +60,16 @@ private void setBuffering(boolean buffering) {
     }
   }
 
-  @SuppressWarnings("SuspiciousNameCombination")
-  private void sendInitialized() {
+  private void sendInitializedPrivate() {
```
matanlurey, line 63: Just inline this method below, it doesn't add value.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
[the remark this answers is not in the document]
matanlurey: @camsim99 Please review before LGTM. Thanks!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
matanlurey: @FirentisTFW Can you clarify this again, given that playerId != textureId, why do we have the concept of `nonTexturePlayerId`? Couldn't/shouldn't a single counter be used for both types of players now?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
matanlurey: Ack. Sg.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
matanlurey: Nevermind, this isn't important. Thanks!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerProvider.java
matanlurey: Yeah this works. Thanks

packages/video_player/video_player_android/lib/src/android_video_player.dart
matanlurey: Ack, sounds good.

matanlurey (commented):
Mostly LG.

Holding off on approval so @camsim99 can also take a look. Thanks for addressing the comments!

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayerPlugin.java
FirentisTFW: It's because we don't have control of texture IDs - we obtain them from texture registry when registering a texture. They usually start at zero and then are incremented (by 1 or 2 - that's why we start at a high number for platform views). So we can't really try to manage them ourselves, alongside platform views, unless we add another layer on top of that - such a layer would hold abstract player IDs that we fully control, and the ones related to texture would also keep track of their texture IDs somehow.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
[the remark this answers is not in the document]
FirentisTFW: @camsim99 Could you have a look here now that other Matan's comments were addressed?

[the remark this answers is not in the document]
camsim99: IIRC we left that comment because it's an unstable API, but it's the only way we found that `Exoplayer` reported what we needed since `getVideoSize` is not used on API 21+. I see in my PR that I left a comment to revisit if we notice any differences, since it's hard to tell from the documentation, so I guess in the worst case this PR wouldn't break any cases that weren't already broken.

camsim99 (approved):
The other changes look reasonable to me, thanks!!

FirentisTFW:
@stuartmorgan @camsim99 @matanlurey Thanks for the review! I did some lasts manual tests on a couple of devices and everything seems to be working fine. I'll go ahead and merge this one.

PR #8501, opened 2025-01-24
[pigeon] Adds support of TaskQueue for Swift

feduke-nukem:
- Adds support of `TaskQueue` for Swift
- Adds related generator tests for Objective-c, Kotlin, Swift and Java

closes [111512](https://github.com/flutter/flutter/issues/111512)

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/lib/swift_generator.dart
```diff
@@ -851,6 +854,7 @@ if (wrapped == nil) {
             }) {
               return 'try instanceManager.removeAllObjects()';
             },
+            taskQueueType: TaskQueueType.serial,
```
hellohuanlin, line 857: is serial task queue the only supported type? do we support concurrent queue?

packages/pigeon/lib/swift_generator.dart
```diff
@@ -1531,8 +1536,26 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
 
     final String varChannelName = '${name}Channel';
     addDocumentationComments(indent, documentationComments, _docCommentSpec);
-    indent.writeln(
-        'let $varChannelName = FlutterBasicMessageChannel(name: "$channelName", binaryMessenger: binaryMessenger, codec: codec)');
+    String? taskQueue;
+    if (taskQueueType != TaskQueueType.serial) {
```
hellohuanlin, line 1540: im confused here - does `TaskQueueType.serial` mean it's not background queue? If it means it's a main queue, maybe we should rename to `serialMainQueue` instead.

packages/pigeon/platform_tests/test_plugin/ios/Classes/CoreTests.gen.swift
```diff
@@ -3059,10 +3276,149 @@ class HostIntegrationCoreApiSetup {
...
+    )
+    if let api = api {
+      echoAllTypesTaskQueueBackgroundChannel.setMessageHandler { message, reply in
+        let args = message as! [Any?]
+        let everythingArg = args[0] as! AllTypes
+        do {
+          let result = try api.echoAllTypesTaskQueueBackground(everythingArg)
+          reply(wrapResult(result))
+        } catch {
+          reply(wrapError(error))
```
hellohuanlin, line 3296: IIUC `reply` should still be called on main thread.

packages/pigeon/lib/swift_generator.dart
feduke-nukem: image

packages/pigeon/platform_tests/test_plugin/ios/Classes/CoreTests.gen.swift
feduke-nukem: Would it be sufficient to add?

```swift
DispatchQueue.main.async {
  reply(wrapResult(result))
}
```

like it is [mentioned in docs](https://docs.flutter.dev/platform-integration/platform-channels#jumping-to-the-main-thread-in-ios)

feduke-nukem: However, I wonder why it was not done for others generators in the past.

Outputs for methods marked with `TaskQueue`:

Objective-c:
image

Java:
image

Kotlin:
image

packages/pigeon/lib/swift_generator.dart
feduke-nukem: Doc states:

```dart
  /// Handlers are invoked serially on the default thread. This is the value if
  /// unspecified.
```

packages/pigeon/platform_tests/test_plugin/ios/Classes/CoreTests.gen.swift
stuartmorgan: > IIUC `reply` should still be called on main thread.

Please see the second bullet point [here](https://github.com/flutter/packages/pull/8341#discussion_r1919146820).

hellohuanlin: > > IIUC `reply` should still be called on main thread.
> 
> Please see the second bullet point [here](https://github.com/flutter/packages/pull/8341#discussion_r1919146820).

Sorry i misread your message.

packages/pigeon/lib/swift_generator.dart
stuartmorgan: > im confused here - does `TaskQueueType.serial` mean it's not background queue?

No, it means that all calls on that API object using the serial task queue will be executed in order, each one completing before the next starts.

As compared with, e.g., a default `NSOperationQueue`, which will dispatch jobs into a thread pool, and might run everything in parallel on N different threads.

packages/pigeon/platform_tests/test_plugin/ios/Classes/CoreTests.gen.swift
hellohuanlin: looking at it again, I must have misunderstood the 3rd bullet point as the `reply` call. Sorry for the confusion.

packages/pigeon/lib/swift_generator.dart
stuartmorgan: > do we support concurrent queue?

We have never had a use case in our own plugins for concurrent queues, AFAIK nobody else has ever requested it, and I would not expect concurrent task queues to be a common use case in platform channels, so currently we have had no reason to implement it.

packages/pigeon/lib/swift_generator.dart
hellohuanlin: > Handlers are invoked serially on the default thread.

Is the "default thread" referring to main thread?

stuartmorgan: No, it's an arbitrary background thread. That's why the API called by the generated code is `makeBackgroundTaskQueue`.

Calling things serially on the main thread is what platform channels do by default, and what they always did before the creation of the task queue API. The whole point of adding task queues to the channel APIs in the engine was to move them off of the main thread.

hellohuanlin: Please forgive my numerous questions, I think I may be missing something. 

What you just described sounds very like the `.serialBackgroundThread` case, rather than the `.serial` case. Also, the code here is: 

```
if NOT serial { // the only other case is serialBackgroundThread
  use makeBackgroundTaskQueue API
}
```

packages/pigeon/platform_tests/test_plugin/ios/Classes/CoreTests.gen.swift
feduke-nukem: > I must have misunderstood the 3rd bullet point as the `reply` call

Does it have any implications for the discussion in [this PR](https://github.com/flutter/packages/pull/8341)?

Apologies if I’m missing something.

packages/pigeon/lib/swift_generator.dart
stuartmorgan: > What you just described sounds very like the `.serialBackgroundThread` case, rather than the `.serial` case.

Sorry, I forgot that Pigeon for some reason has a task queue enum value that actually means "don't use task queues". I have no idea why it was written that way; it's incredibly confusing.

So yes, it's the main thread and everything I said was wrong.

> maybe we should rename to `serialMainQueue` instead

We should probably just remove it, or if we keep it call it the even simpler `none`, but either way that's out of scope here.

hellohuanlin: Gotcha. Thanks! 

@feduke-nukem while you are at it, maybe we can add an example usage in README? This should help clear up this kind of confusion for other people. What do you think? https://github.com/flutter/flutter/issues/162034

feduke-nukem: Sure @hellohuanlin

packages/pigeon/platform_tests/test_plugin/ios/Classes/CoreTests.gen.swift
hellohuanlin: Yes, in my [previously proposed code](https://github.com/flutter/packages/pull/8341#discussion_r1919328977) I proposed to use an inner `Task` to switch back to main before calling `reply`, because I mistakenly thought `reply` has to be called on `main`. 

Anyways I will put up a design doc which is easier for us to discuss.

PR #8510, opened 2025-01-27
Revert "[shared_preferences] update List<String> encode/decode"

camsim99:
Reverts flutter/packages#8335.

This is causing `shared_preferences/shared_preferences/example/integration_test/shared_preferences_test.dart: shared_preferences_async Async throws TypeError when returned getStringList type is incorrect` to fail with what looks like an OOB issue and is blocking the tree currently:

```
The following TestFailure was thrown running a test (but after the test had completed):
Expected: throws <Instance of 'TypeError'>
  Actual: <Closure: () => Future<Null>>

#0      fail (package:matcher/src/expect/expect.dart:149:31)
#1      _expect.<anonymous closure> (package:matcher/src/expect/expect.dart:125:9)
<asynchronous suspension>
<asynchronous suspension>
(elided 6 frames from dart:async and package:stack_trace)
════════════════════════════════════════════════════════════════════════════════════════════════════

```


packages/shared_preferences/shared_preferences_android/CHANGELOG.md
```diff
@@ -1,7 +1,3 @@
-## 2.4.3
```
reidbaker, line 1: I forget our rules here but I thought we couldnt roll back this version if it was published which the existence of an OOB change implies happened.

camsim99: @stuartmorgan please advise. Should we publish a new version?

stuartmorgan:
I just realized, I don't think we can revert this. It's been published and may have started converting real-world data to the new form. Updating to a reverted version would lose data.

stuartmorgan:
I'll look at the tests when I'm back at my desk to see if we can easily fix forward; hopefully the tests were just too specific to platform implementations. Otherwise I'll need to think about this more and discuss with @tarrinneal

PR #8512, opened 2025-01-27
[shared_preferences] Fix Android type mismatch regression

stuartmorgan:
`getStringList` should throw a `TypeError` if the stored value is of a different type, but the recent change to use JSON-encoded string lists regression that behavior if the stored type was a string, causing it to instead return null.

This restores the previous behavior by passing extra information from Kotlin to Dart when attempting to get an enecoded string list, so that if a non-encoded-list string is found, a TypeError can be created on the Dart side.

Since extra information is now being passed, the case of a legacy-encoded value is now communicated as well, so that we only have to request the legacy value if it's there, rather than always trying (which was not worth the complexity of adding extra data just for that initially, but now that we need extra data anyway, it's easy to distinguish that case).

Fixes OOB regression in `shared_preferences` tests that has closed the tree.

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan:
I did consider an alternate version where I just always returned the raw string value, then did the checks for null vs one of the prefixes vs a raw string all on the Dart side, but that meant always sending the entire string (even the potentially large base-64-legacy-encoded string) over the platform channel for basically no reason, so I felt like this would be cleaner.

(In a potential future world where we convert this to FFI, that won't be an issue and we could just have all the logic in Dart.)

packages/shared_preferences/shared_preferences_android/android/src/test/kotlin/io/flutter/plugins/sharedpreferences/SharedPreferencesTest.kt
```diff
@@ -217,7 +236,26 @@ internal class SharedPreferencesTest {
   fun testSetAndGetStringListWithSharedPreferences() {
     val plugin = pluginSetup(sharedPreferencesOptions)
     plugin.setEncodedStringList(listKey, testList, sharedPreferencesOptions)
-    Assert.assertEquals(plugin.getStringList(listKey, sharedPreferencesOptions), testList)
+    val result = plugin.getStringList(listKey, sharedPreferencesOptions)
+    Assert.assertEquals(result?.jsonEncodedValue, testList)
+  }
+
+  @Test
+  fun testSetAndGetStringListWithSharedPreferencesRedirectsForLegacy() {
```
reidbaker, line 244: non blocking nit: would prefer "...DeprecatedStringList" over "Legacy" but this is a test so the consequence is small.

packages/shared_preferences/shared_preferences_android/pigeons/messages_async.dart
```diff
@@ -24,6 +24,27 @@ class SharedPreferencesPigeonOptions {
...
+
+  /// Whether value using the legacy platform-side encoding was found.
+  ///
+  /// This value is only meaningful if [jsonEncodedValue] is null.
+  /// - If true, the value should be fetched with
+  ///   getPlatformEncodedStringList(...) instead.
+  /// - If false, an unexpected string (one without any encoding prefix) was
+  ///   found. This will happen if a client uses getStringList with a key that
+  ///   was used with setString.
+  bool foundPlatformEncodedValue;
```
reidbaker, line 45: This seems like the type of decision we will regret later when there is yet another way to store lists of strings. Consider an enum.

reidbaker (approved):

packages/shared_preferences/shared_preferences_android/pigeons/messages_async.dart
stuartmorgan: > This seems like the type of decision we will regret later when there is yet another way to store lists of strings.

I did consider an enum originally, but the two-value enum seemed like overkill once I started writing it. I'll go ahead and switch back, but this is an internal implementation detail, so if ($DIETY forbid) we add a third way to encode string lists we wouldn't regret the bool, we would just change it then.

(Also, ideally we'd refactor more logic into Dart before then so we wouldn't even need to.)

stuartmorgan: Fixed; it's now a three-state enum. One of the states is redundant with the string value being non-null, but that made the logic flow easier to understand than it was before.

packages/shared_preferences/shared_preferences_android/android/src/test/kotlin/io/flutter/plugins/sharedpreferences/SharedPreferencesTest.kt
stuartmorgan: Renamed to the more specific "PlatformEncoded".

camsim99:
This looks good to go, so let's land on red to fix the packages tree!

stuartmorgan:
The label the bot is looking for changed to `emergency` in https://github.com/flutter/cocoon/pull/4176, but this repo wasn't updated. I'll switch the labels and hopefully it will work (assuming that's not gated on flutter/flutter).

stuartmorgan:
Oops; I removed and re-added it because I wasn't sure the rename would get picked up, but it looks like maybe my timing on removing it was terrible.

stuartmorgan:
No, I think it's broken. Landing manually, and I'll file an issue.

PR #8494, opened 2025-01-24
[shared_preferences] Add shared preferences devtool

adsonpleal:
This PR adds the shared_preferences_tool package. This package uses the [devtools_extension](https://pub.dev/packages/devtools_extensions) tooling to create a tool for shared preferences. The idea of this PR came from @kenzieschmoll on this [issue](https://github.com/flutter/flutter/issues/145433). Initially I've published this tool as a [separate package](https://pub.dev/packages/shared_preferences_tools), but this PR aims to bring the functionality to the main shared_preferences package. Once this PR gets merged I'll archive the `shared_preferences_tools` package. 

https://github.com/flutter/packages/assets/11666470/fcf71145-c330-4397-b62e-c0c4c8bc9f01

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [X] I signed the [CLA].
- [X] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [X] I [linked to at least one issue that this PR fixes] in the description above.
- [X] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [X] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making, or this PR is [test-exempt].
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[...]

PR #8531, opened 2025-01-29
Re-land [shared_preferences] Add shared preferences devtool

stuartmorgan:
Re-re-lands https://github.com/flutter/packages/pull/8494 with a change to remove `--ignored` from the `git status` check in the repo tooling `publish` command, which is what caused the `release` failure in the last landing. `pub publish` has ignored files in `.gitignore` for years, so there's no longer any value in checking that they don't exist.

I've verified that `publish --dry-run` passes locally with this change. Server-only check failures could still potentially cause another revert, but those can't be found other than the hard way.


tarrinneal (commented):
here we go again

tarrinneal (approved):

stuartmorgan:
> here we go again

I don't know where we're going, but I sure know where we've been.

stuartmorgan:
Note to self: If we need to revert and re-land again (knock on wood), we should pull out the repo-wide changes (tooling and release action version) from the revert so that we don't have to run repo-wide tests on the next cycle.

tarrinneal:
It seems like that test is just hit it's time limit, nothing is hanging or failing.

PR #8532, opened 2025-01-30
[pigeon] reorg generator files

tarrinneal:
Just moves a few files into folders to clean things up a bit. Future organizing is likely, but I wanted to move the files with no changes first


tarrinneal:
And I do believe this is exempt from versioning

stuartmorgan:
This is actually a breaking change. The [Dart package convention](https://dart.dev/tools/pub/package-layout#implementation-files) is that anything in `lib/src/` is private to the package and should never be imported directly by clients, but anything else in `lib/` is fair game, so currently it would be fine, by language conventions, for someone to write code that imports these Pigeon files directly.

If we don't want the generator files to be directly imported (which I think we don't), we should do a breaking change to move a bunch of things into `src/`, and then in the future we can reorganize `src/` however we want.

tarrinneal:
I bumped the version and moved everything into src. I can wait to do this along with another breaking change if you'd like, but I think editing the files and moving them is a bit messy for reviews.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 23.0.0
+
+* Relocates `lib` files to `src` directory.
```
stuartmorgan, line 3: How about "Moves some files in `lib` that were not intended for direct client use to `lib/src`."

packages/pigeon/bin/pigeon.dart
```diff
@@ -4,7 +4,7 @@
 
 import 'dart:io' show exit;
 
-import 'package:pigeon/pigeon_cl.dart';
+import 'package:pigeon/src/pigeon_cl.dart';
```
stuartmorgan, line 7: Interesting, I guess `bin/` has special-case allowance in the analyzer to import from the same package's `src` even if it's using a `package:` import?

stuartmorgan (commented):
LGTM with changelog nit.

> I can wait to do this along with another breaking change if you'd like, but I think editing the files and moving them is a bit messy for reviews.

I expect ~0 people to actually be broken by this change, and we've always treated major version changes for `pigeon` as low-cost, so I don't think there's any real value in combining (and as you point out, not-trivial downsides for our ability to do code reviews).

packages/pigeon/bin/pigeon.dart
tarrinneal: I was surprised it was allowed as well, since the pigeon_cl.dart file is used outside of lib and tests, should I move it back out of src?

stuartmorgan: Nope, I found a reference later on the page I had linked to, it's explicitly allowed:

> You are free to import libraries that live in lib/src from within other Dart code in the same package (like other libraries in lib, **scripts in bin**, and tests) but you should never import from another package's lib/src directory.

And that's the preferred format for such an include:

> How you import libraries from within your own package depends on the locations of the libraries:
> 
> When [reaching inside or outside lib/](https://dart.dev/effective-dart/usage#dont-allow-an-import-path-to-reach-into-or-out-of-lib) (lint: [avoid_relative_lib_imports](https://dart.dev/tools/linter-rules/avoid_relative_lib_imports)), use package:.

stuartmorgan (approved):
I didn't push the button again. I need a Chrome extension for this.

tarrinneal:
We should land https://github.com/flutter/packages/pull/8302/ first to avoid a ton of churn on that pr

PR #8539, opened 2025-01-30
[shared_preferences_tool] Loosen vm_service constraint to allow for 15

parlough:
This is causing the `flutter-analyze-try` Dart try bot to [fail](https://ci.chromium.org/ui/p/dart/builders/try/flutter-analyze-try/26500/overview) due to [upgrading the version pinned by flutter](https://github.com/flutter/flutter/commit/3e3bb9fdd586e71bc7fb797ee7158b53a54cd634).

### DIFF
--- a/packages/shared_preferences/shared_preferences_tool/pubspec.yaml
+++ b/packages/shared_preferences/shared_preferences_tool/pubspec.yaml
@@ -2,7 +2,7 @@ name: shared_preferences_tool
 description: "DevTools extension for package:shared_preferences. Manage SharedPreferences efficiently. Edit, search, and view keys."
 publish_to: 'none'
 
-version: 1.0.0+2
+version: 1.0.1
 
 environment:
   sdk: '>=3.6.0 <4.0.0'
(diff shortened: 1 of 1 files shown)


packages/shared_preferences/shared_preferences_tool/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: shared_preferences_tool
 description: "DevTools extension for package:shared_preferences. Manage SharedPreferences efficiently. Edit, search, and view keys."
 publish_to: 'none'
 
-version: 1.0.0+1
+version: 1.0.0+2
```
stuartmorgan, line 5: This should be 1.0.1; we don't use `+` versions beyond `0.x`. (This was apparently not caught in the initial review, but we should fix it now.)

stuartmorgan (dismissed):
Have you manually verified that this actually works? Because master here is still using the older version, [presubmit isn't providing any meaningful testing of this change](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8724281455701567089/+/u/Run_package_tests/analyze/stdout); if 15.x breaks this package, CI won't tell us until it's too late.

PR #8540, opened 2025-01-30
[shared_preferences] Fix a late initialized error with the example app

kenzieschmoll:
The `_counter` future was throwing a late initialization error because it was accessed from the `FutureBuilder` in `build()` before it was initialized. This adds a completer `_preferencesReady` that we wait for before trying to build the `FutureBuilder` for `_counter`.


kenzieschmoll:
I wasn't sure if I needed to update the pubspec version because I added a changelog entry. Happy to revert the pubspec change if need be, just let me know.

tarrinneal:
I'm not 100% sure, but I think this does need a version change. You should check out the changelog style guide. https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style

packages/shared_preferences/shared_preferences/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.5.2
+
+* Fixed a bug in the example app.
```
tarrinneal, line 3: `Fixes`

packages/shared_preferences/shared_preferences/example/test/example_app_test.dart
```diff
@@ -0,0 +1,25 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:shared_preferences_example/main.dart';
+import 'package:shared_preferences_platform_interface/shared_preferences_async_platform_interface.dart';
+
+import '../..//test/shared_preferences_async_test.dart';
```
tarrinneal, line 9: Not sure if the `//` is intentional, but I don't think it's necessary

tarrinneal (approved):
Just this tense change in the changelog and then it's good to land

stuartmorgan:
From triage: Is fixing the failing unit test on your radar?

stuartmorgan:
From triage: @kenzieschmoll Ping on the status of this PR?

packages/shared_preferences/shared_preferences/example/test/example_app_test.dart
```diff
@@ -23,3 +24,100 @@ void main() {
     });
   });
 }
+
+// Note: this code is duplicated in
+// shared_preferences/test/shared_preferences_async_test.dart. Since we cannot
+// import the relative path ../../test/shared_preferences_async_test.dart on the
+// web platform, we had to copy it here for use in this test library.
```
tarrinneal, line 31: Is there a way to put this code somewhere that can be imported in both locations that isn't exported to users?

kenzieschmoll: Short of creating a shared package for testing utilities, I'm not sure how else to share this code.

I'm not entirely sure why the relative path can be imported successfully from a test for native platforms but not for the web platform. CC @jakemac53 do you know if this behavior is expected from the test package for this case? 

Another option is to add `TestOn('vm')` here so that this test does not run for the web platform.

jakemac53: This is expected yes, the way apps are hosted on the web we have no way to access URIs outside of the package.

The "root" of the server is either the test directory or the package directory (can't remember which), so you can only access things from the current package and then package: URIs (via synthetic /packages/ paths).

jakemac53: These sorts of relative paths also really shouldn't be used anyways, they can easily be broken (that file could have dependencies you don't have for instance).

The solution is a shared testing package (no need to publish it).

kenzieschmoll: @tarrinneal are you okay landing this with a todo to create a shared testing package or do we need to do that as part of this PR?

4rthurmonteiro:
this pr closes https://github.com/flutter/flutter/issues/164949 ?

PR #8563, opened 2025-02-04
[google_maps_flutter] Ground overlay support - platform impls

jokerttu:
This PR contains platform implementations for the ground overlays support (#8432).

Linked issue: https://github.com/flutter/flutter/issues/26479

## Pre-launch Checklist

- [x] I signed the [CLA].


ditman:
Looking at web stuff!

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+   *     overlay options.
+   * @param sink the GroundOverlaySink where the options will be set.
+   * @param assetManager An instance of Android's AssetManager, which provides access to any raw
+   *     asset files stored in the application's assets directory.
+   * @param density the density of the display, used to calculate pixel dimensions.
+   * @param wrapper the BitmapDescriptorFactoryWrapper to create BitmapDescriptor.
+   * @return the identifier of the ground overlay.
+   * @throws IllegalArgumentException if required fields are missing or invalid.
+   */
+  static String interpretGroundOverlayOptions(
```
stuartmorgan, line 866: The return and the parameters should get nullability annotations.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+
+  /**
+   * Builds a PlatformDoublePair representing the anchor point for a GroundOverlay.
+   *
+   * @param groundOverlay the GroundOverlay object.
+   * @return the PlatformDoublePair representing the anchor point.
+   */
+  @VisibleForTesting
+  public static @NonNull Messages.PlatformDoublePair buildGroundOverlayAnchorForPigeon(
+      GroundOverlay groundOverlay) {
```
stuartmorgan, line 957: `@NonNull`

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/GroundOverlaysControllerTest.java
```diff
@@ -0,0 +1,150 @@
...
+                .build()));
+    Mockito.verify(groundOverlay, times(1)).setTransparency(newTransparency);
+
+    controller.removeGroundOverlays(Collections.singletonList(googleGroundOverlayId));
+
+    Mockito.verify(groundOverlay, times(1)).remove();
+  }
+
+  // Helper method to generate 1x1 pixel base64 encoded png test image
+  private String generateBase64Image() {
```
stuartmorgan, line 131: This can be pulled out of the other file into a test helper file, rather than duplicating the code.

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.h
```diff
@@ -0,0 +1,64 @@
...
+@property(strong, nonatomic) GMSGroundOverlay *groundOverlay;
+
+/// Whether ground overlay is created with bounds or position.
+@property(nonatomic, assign, getter=isCreatedWithBounds) BOOL createdWithBounds;
+
+/// Zoom level when ground overlay is initialized with position.
+@property(nonatomic, strong, nullable) NSNumber *zoomLevel;
+
+/// Initializes an instance of this class with a GMSGroundOverlay, a map view, and identifier.
+- (instancetype _Nullable)initWithGroundOverlay:(GMSGroundOverlay *)groundOverlay
```
stuartmorgan, line 28: Does `nullable instancetype` not work? (I can never remember the exact set of cases where `nullable` doesn't work).

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_map_inspector_ios.dart
```diff
@@ -103,6 +104,60 @@ class GoogleMapsInspectorIOS extends GoogleMapsInspectorPlatform {
...
+      return null;
+    }
+
+    // Create dummy image to represent the image of the ground overlay.
+    final BytesMapBitmap dummyImage = BytesMapBitmap(
+      Uint8List.fromList(<int>[0]),
+      bitmapScaling: MapBitmapScaling.none,
+    );
+
+    if (groundOverlayInfo.position != null) {
```
stuartmorgan, line 127: Nit: If you make locals for `groundOverlay.position` and `groundOverlay.bounds` just before the if block, and use the locals everywhere after that (including the conditions), you won't need most of the force-unwraps below.

stuartmorgan (commented):
Android and iOS LGTM; I did a full review of both. Please wait for @ditman's review covering web before landing though.

stuartmorgan:
(@ash-google Feel free to review as well, I just had already looked over the mobile platforms in the main PR so I went ahead and did a review here to cover those.)

packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/FGMGroundOverlayController.h
jokerttu: removed nullable annotation, as the function never returns null instance

jokerttu: I removed the nullable annotation from these instance initialisations as they do not explicitly return nil.
There isn't any handling for null instances on code either.
I would assume that `self = [super init];` returns instance, and if not, something is really broken.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+    if (groundOverlay.getAnchor() != null) {
+      sink.setAnchor(
+          groundOverlay.getAnchor().getX().floatValue(),
+          groundOverlay.getAnchor().getY().floatValue());
+    }
+    sink.setBearing(groundOverlay.getBearing().floatValue());
+    sink.setClickable(groundOverlay.getClickable());
+    sink.setImage(toBitmapDescriptor(groundOverlay.getImage(), assetManager, density, wrapper));
+    if (groundOverlay.getPosition() != null) {
+      assert groundOverlay.getWidth() != null;
```
ash-google, line 884: This should have an assertion message indicating why it is required or at least log an error saying that width is required.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,139 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+    // Calculate normalized latitude.
+    double height = bounds.northeast.latitude - bounds.southwest.latitude;
+    double normalizedLatitude = 1.0 - ((position.latitude - bounds.southwest.latitude) / height);
+
+    // Calculate normalized longitude.
+    // For longitude, if the bounds cross the antimeridian (west > east),
+    // adjust the width accordingly.
+    double west = bounds.southwest.longitude;
+    double east = bounds.northeast.longitude;
+    double width = (west <= east) ? (east - west) : (360.0 - (west - east));
```
ash-google, line 973: Can we extract these constants into local constants with good naming so it's easier to understand why we need these magic numbers?

ash-google (approved):

ash-google:
Did a second look over Android and LGTM. Let's wait for @ditman for web.

jokerttu:
There is some crash issues on Android integration tests: [Linux_android android_device_tests_shard_1 master](https://github.com/flutter/packages/pull/8563/checks?check_run_id=36954136393)
Haven't been able to re-produce the crash issue (with same sdk version as tests are runned). Not yet sure if these crashes are related to ground overlay implementation or not.

stuartmorgan:
> There is some crash issues on Android integration tests: [Linux_android android_device_tests_shard_1 master](https://github.com/flutter/packages/pull/8563/checks?check_run_id=36954136393)

It's https://github.com/flutter/flutter/issues/159731. We're seeing flare-ups of this on other PRs that touch maps, so I don't think it's related. We've seen spikes in this crash in our CI before, potentially due to server-side changes in Google Maps.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
jokerttu: @ash-google is this what you were looking for? https://github.com/flutter/packages/pull/8563/commits/cdf4d25920131a6bce4ed5f21f2c1b3883c20dfe

matijakevic:
I tested this implementation a bit on Android. Unfortunately, I ran into an issue which occurs when ground overlay images change quickly (<250ms interval). This results in a quick but noticeable black flicker.

The images are stored in-memory as `Uint8List` and loaded through `BytesMapBitmap`.

The `GroundOverlay` has the same unchanging `groundOverlayId` all the time, only the image is changing.

I am not sure what causes this, maybe the thread that updates the ground overlays doesn't manage to decode bitmaps from raw data quickly enough?

I attached a video demonstrating this on Xiaomi 12T. Maybe I could try creating minimal reproducible example if needed.


https://github.com/user-attachments/assets/d048c9be-7aa0-4971-b0c4-8a6138257c67

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,137 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+    if (groundOverlay.getAnchor() != null) {
+      sink.setAnchor(
+          groundOverlay.getAnchor().getX().floatValue(),
+          groundOverlay.getAnchor().getY().floatValue());
+    }
+    sink.setBearing(groundOverlay.getBearing().floatValue());
+    sink.setClickable(groundOverlay.getClickable());
+    sink.setImage(toBitmapDescriptor(groundOverlay.getImage(), assetManager, density, wrapper));
+    if (groundOverlay.getPosition() != null) {
+      assert groundOverlay.getWidth() != null
```
reidbaker, line 884: By default asserts are stripped in java. If you want to check this value then you will need to throw an exception or log or do some other form of error handling.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -849,6 +850,137 @@ static Tile tileFromPigeon(Messages.PlatformTile tile) {
...
+   * @param isCreatedWithBounds indicates if the GroundOverlay was created with bounds.
+   * @return the converted PlatformGroundOverlay object.
+   */
+  static @NonNull Messages.PlatformGroundOverlay groundOverlayToPigeon(
+      @NonNull GroundOverlay groundOverlay,
+      @NonNull String groundOverlayId,
+      boolean isCreatedWithBounds) {
+
+    // Dummy image is used as image is required field of PlatformGroundOverlay and converting image
+    // back to image descriptor is not currently supported.
```
reidbaker, line 910: Can you add a link to what layer of the stack does not support this functionality? Also why is it safe to use a dummy image?

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ConvertTest.java
```diff
@@ -665,33 +660,124 @@ public void ConvertInterpretHeatmapOptionsReturnsCorrectData() {
...
+    when(mockGroundOverlay.getBearing()).thenReturn(50f);
+    when(mockGroundOverlay.getTransparency()).thenReturn(0.6f);
+    when(mockGroundOverlay.getZIndex()).thenReturn(7f);
+    when(mockGroundOverlay.isVisible()).thenReturn(true);
+    when(mockGroundOverlay.isClickable()).thenReturn(false);
+
+    Messages.PlatformGroundOverlay result =
+        Convert.groundOverlayToPigeon(mockGroundOverlay, "overlay_1", false);
+
+    Assert.assertEquals("overlay_1", result.getGroundOverlayId());
```
reidbaker, line 715: Consider extracting a helper method that takes a actual ground overlay object and expected values and handles the assertions for equals. It would make new tests easier to write/review.

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -21,6 +20,7 @@ const double _kInitialZoomLevel = 5;
 const CameraPosition _kInitialCameraPosition =
     CameraPosition(target: _kInitialMapCenter, zoom: _kInitialZoomLevel);
 const String _kCloudMapId = '000000000000000'; // Dummy map ID.
+const double _floatTolerance = 1e-8;
```
reidbaker, line 23: Can you add a comment for why this value was chosen?

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
```diff
@@ -995,7 +995,7 @@ void googleMapsTests() {
   },
       // TODO(cyanglaz): un-skip the test when we can test this on CI with API key enabled.
       // https://github.com/flutter/flutter/issues/57057
-      skip: Platform.isAndroid);
+      skip: true);
```
reidbaker, line 998: Did you mean to skip this for all platforms? if so should it be deleted instead?

packages/google_maps_flutter/google_maps_flutter_android/example/lib/ground_overlay.dart
```diff
@@ -0,0 +1,325 @@
...
+
+  void _onGroundOverlayTapped() {
+    _changePosition();
+  }
+
+  void _setBearing() {
+    assert(_groundOverlay != null);
+    setState(() {
+      _groundOverlay = _groundOverlay!.copyWith(
+          bearingParam: _groundOverlay!.bearing >= 350
```
reidbaker, line 135: Can you add a comment explaining where these magic numbers come from? 
I believe this is just bumping the bearing 10 degrees  and making sure we dont go over 360 but  that is after reading the whole pr.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
jokerttu: This is used only by inpector api, for testing purposes. App facing widget stores groundoverlays on dart implementation, and there is no real reason to fetch the data back.
Image is mandatory field on PlatformGroundOverlay (and it should be kept non-nullable), therefore image must be set for the object.
The image is description either contains set of bytes, or path to asset. This info is converted to format google maps uses (BitmapDescription), and the original data is not stored on native code.
So converting read and possibly rescaled asset bitmap back orginal instruction with asset url is not currently possible without storing the original PlatformImageDescriptor to memory; which is not optimal.
Therefore using dummy image here should be fine?

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
jokerttu: No, this was committed by accident. Not related to this PR.

stuartmorgan: This is a legacy of this file starting its life as a copy of the integration tests in `google_maps_flutter`. In the context of `google_maps_flutter_android`, `Platform.isAndroid` is just a confusing way of writing `true`, so this change was good.

> if so should it be deleted instead?

I would prefer we keep it, because of the TODO above. We are starting to see features that do not work without an API key, so we will need to fix the CI limitation at some point. (I was working on it with infra folks a while ago, but that got derailed.)

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
jokerttu: Done

packages/google_maps_flutter/google_maps_flutter_android/example/lib/ground_overlay.dart
jokerttu: Is this ok?

packages/google_maps_flutter/google_maps_flutter_android/example/integration_test/google_maps_tests.dart
jokerttu: This is true; but already reverted this change before I saw @stuartmorgan:n comment about this, as this was chore change, not related to ground overlay at all. I can commit this back, if it is ok to have unrelated changes in the PR as well.

stuartmorgan: It's fine either way; if you've already reverted it don't worry about adding it back unless you feel like it.

reidbaker (approved):

jokerttu:
I'll address review comments made to main PR by @reidbaker later this week: https://github.com/flutter/packages/pull/8432#pullrequestreview-2638465717
This is still waiting review from @ditman as well.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
reidbaker: Thanks this helps answer my question. Consider putting a shorter version of that in the code. 

Maybe something like "Image is mandatory field on PlatformGroundOverlay (and it should be kept non-nullable), therefore image must be set for the object. The image is description either contains set of bytes, or path to asset. This info is converted to format google maps uses (BitmapDescription), and the original data is not stored on native code."

stuartmorgan (commented):
I took a look at web since I'm not sure when @ditman will have time. Mostly it seems straightforward, but I have a couple of questions.

packages/google_maps_flutter/google_maps_flutter_android/example/lib/ground_overlay.dart
jokerttu: https://github.com/flutter/packages/pull/8563/files#diff-db70730aaecef994ff31ac3529c7fc25e0a68e0c17c0d2ccba0ae429efc3a2c4R134-R135

stuartmorgan (approved):
Web LGTM

jokerttu:
@reidbaker are you going to check the changed made addressing these (on this PR and also mirrored to original app facing PR) https://github.com/flutter/packages/pull/8432#pullrequestreview-2638465717
or is the earlier approval still valid?

reidbaker:
Thanks for asking, @ash2moon's approval for android is good enough for me. Thank you for this change.

PR #8567, opened 2025-02-04
[pigeon] Add errors for ProxyAPI callback methods and null instances when reading in a ProxyApiBaseCodec

bparrishMines:
* Adds an error message when a ProxyAPI callback method that returns a non-null value is nullable. Returning a null value causes error on the platform side. This ensures the Dart side prevents this from happening.

* Adds an error message in the `ProxyApiBaseCodec` when an instance could not be retrieved when reading a value.

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/platform_tests/shared_test_plugin_code/lib/src/generated/proxy_api_tests.gen.dart
```diff
@@ -1489,123 +1469,6 @@ class ProxyApiTestClass extends ProxyApiSuperClass
         _PigeonInternalProxyApiBaseCodec(
             pigeon_instanceManager ?? PigeonInstanceManager.instance);
     final BinaryMessenger? binaryMessenger = pigeon_binaryMessenger;
-    {
```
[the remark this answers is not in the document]
bparrishMines: This removes the `pigeon_newInstance` callback which handles creating a call from the platform side to create a new Dart instance. However the class has a required callback method, so all new instances must be created starting from Dart. Its the language equivalent of an `abstract class`.

The generated Dart class has something like this:

```dart
class TestClass {
  TestClass({required this.aBool, required this.myCallback});

  final bool aBool;
  final void Function() myCallback;
}
```

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTestApiImpls.kt
```diff
@@ -105,78 +105,6 @@ class ProxyApiTestClassApi(override val pigeonRegistrar: ProxyApiRegistrar) :
     return ProxyApiSuperClass()
   }
 
-  override fun aBool(pigeon_instance: ProxyApiTestClass): Boolean {
```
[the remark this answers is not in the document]
bparrishMines: For the same reason as above, all the platform API field methods like `boolean aBool()` were removed since they were only used in the `pigeonNewInstance()` method that instantiates a new instance from the platform side.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
```diff
@@ -3095,64 +3060,12 @@ abstract class PigeonApiProxyApiTestClass(
       Result.success(Unit)
```
stuartmorgan, line 3060: I just noticed this being different from the other paths; is this a latent bug? Shouldn't it be `callback(Result(success(Unit))`?

stuartmorgan: Looking at the whole method: does this really need to take a callback at all? It seems like it could just directly return `Result<Unit>`, which would have prevented both of these bugs.

packages/pigeon/platform_tests/test_plugin/ios/Classes/ProxyApiTests.gen.swift
```diff
@@ -2853,64 +2820,13 @@ final class PigeonApiProxyApiTestClass: PigeonApiProtocolProxyApiTestClass {
       completion(.success(Void()))
       return
     }
```
stuartmorgan, line 2822: Same structural question for this whole method (I can't comment on the method declaration since it's outside of the change context): could this have a simple return instead of a completion, reducing the complexity?

packages/pigeon/platform_tests/test_plugin/ios/Classes/ProxyApiTests.gen.swift
```diff
@@ -2853,64 +2820,13 @@ final class PigeonApiProxyApiTestClass: PigeonApiProtocolProxyApiTestClass {
       completion(.success(Void()))
```
hellohuanlin, line 2820: minor nit: `Void()` can be just `()`

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
bparrishMines: It mainly takes a callback because the code that generates the `MethodCall` is reused from the `FlutterApi` generation. I think removing the callback would require me to write separate code for the `MethodCall`.

packages/pigeon/platform_tests/test_plugin/ios/Classes/ProxyApiTests.gen.swift
bparrishMines: I changed the code to use a `if...else if...else`, so it no longer needs a return.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
stuartmorgan: It's using `_writeFlutterMethod`, which is a thin wrapper around two functions, the declaration part of which [is already parameterized as to whether or not it's async](https://github.com/flutter/packages/blob/2c3c9476996af6c3700167cab615972978829047/packages/pigeon/lib/src/swift/swift_generator.dart#L1446), so unless there are details I'm missing it should be as easy as just re-exposing `isAsynchronous` (as a default-true argument) in `_writeFlutterMethod`, and setting it to false for this call.

packages/pigeon/platform_tests/test_plugin/ios/Classes/ProxyApiTests.gen.swift
stuartmorgan: The issue isn't the return, it's that (as the Kotlin bug demonstrated) having a function use an async callback is inherently more error-prone, because the compiler will enforce that you return exactly once, but callbacks don't. Since we don't need any async behavior here, the method would be safer if it used a return value instead of a callback.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
bparrishMines: [_writeProxyApiNewInstanceMethod](https://github.com/flutter/packages/blob/2c3c9476996af6c3700167cab615972978829047/packages/pigeon/lib/src/swift/swift_generator.dart#L2270) only uses [_writeFlutterMethodMessageCall](https://github.com/flutter/packages/blob/2c3c9476996af6c3700167cab615972978829047/packages/pigeon/lib/src/swift/swift_generator.dart#L1462) and not `_writeFlutterMethod`. And `_writeFlutterMethodMessageCall` always assumes that the Flutter method uses a `completion`. All methods in `FlutterMethods` are always async on the platform side.

packages/pigeon/platform_tests/test_plugin/ios/Classes/ProxyApiTests.gen.swift
bparrishMines: Yes, but technically the method is making a async call when it makes the `MethodCall` to create the Dart instance. However, the case that the `MethodCall` is replaced by the error, the method probably could just return a `Result`. Were you suggesting that the method should return a `Result` when its not actually making a method call?

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
stuartmorgan: I was looking at the Kotlin generator (I assumed the Swift generator had the same structure), where `_writeProxyApiNewInstanceMethod` is almost entirely a call to `_writeFlutterMethod`, so maybe this is easy for Kotlin but not Swift.

We can leave this for now though since the PR is still an improvement, even if it leaves the potential for a new version of this bug in the future.

packages/pigeon/platform_tests/test_plugin/ios/Classes/ProxyApiTests.gen.swift
stuartmorgan: I'm not following where there's an async step in this method, looking at the code. The structure of `pigeonNewInstance` is:

```swift
func pigeonNewInstance(
    pigeonInstance: ProxyApiTestClass,
    completion: @escaping (Result<Void, ProxyApiTestsError>) -> Void
  ) {
    if pigeonRegistrar.ignoreCallsToDart {
      completion(...)
    } else if pigeonRegistrar.instanceManager.containsInstance(pigeonInstance as AnyObject) {
      completion(...)
    } else {
      completion(...)
    }
  }
```

In every branch, it's just synchronously calling completion. I don't see any reason (other than the logistics of the generator logic helper function discussed in the other thread) it couldn't be

```swift
pigeonNewInstance(pigeonInstance: ProxyApiTestClass) -> Result<Void, ProxyApiTestsError>
```

with every `completion(...)` replaced with a `return ...`.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
```diff
@@ -3089,70 +3054,17 @@ abstract class PigeonApiProxyApiTestClass(
       callback(
           Result.failure(
               ProxyApiTestsError("ignore-calls-error", "Calls to Dart are being ignored.", "")))
+    } else if (pigeonRegistrar.instanceManager.containsInstance(pigeon_instanceArg)) {
+      callback(Result.success(Unit))
       return
```
stuartmorgan, line 3059: This return shouldn't be generated any more; with the change to the if/else if/else structure it's not needed, and it's inconsistent with the other branches.

packages/pigeon/lib/src/kotlin/kotlin_generator.dart
```diff
@@ -1906,64 +1914,70 @@ if (wrapped == null) {
...
         );
         indent.writeScoped(
-          'if (pigeonRegistrar.instanceManager.containsInstance(${classMemberNamePrefix}instanceArg)) {',
-          '}',
+          'else if (pigeonRegistrar.instanceManager.containsInstance(${classMemberNamePrefix}instanceArg)) {',
+          '} ',
           () {
-            indent.writeln('Result.success(Unit)');
+            indent.writeln('callback(Result.success(Unit))');
             indent.writeln('return');
```
stuartmorgan, line 1933: I think this is corresponding generator line?

stuartmorgan (approved):
LGTM with a nit about a stray `return`.

packages/pigeon/platform_tests/test_plugin/ios/Classes/ProxyApiTests.gen.swift
bparrishMines: For most classes, the final completion is done inside of a `MethodCall`: https://github.com/flutter/packages/blob/2c3c9476996af6c3700167cab615972978829047/packages/pigeon/platform_tests/test_plugin/ios/Classes/ProxyApiTests.gen.swift#L3885

```swift
  } else {
      let pigeonIdentifierArg = pigeonRegistrar.instanceManager.addHostCreatedInstance(
        pigeonInstance as AnyObject)
      let binaryMessenger = pigeonRegistrar.binaryMessenger
      let codec = pigeonRegistrar.codec
      let channelName: String =
        "dev.flutter.pigeon.pigeon_integration_tests.ProxyApiSuperClass.pigeon_newInstance"
      let channel = FlutterBasicMessageChannel(
        name: channelName, binaryMessenger: binaryMessenger, codec: codec)
      channel.sendMessage([pigeonIdentifierArg] as [Any?]) { response in
        guard let listResponse = response as? [Any?] else {
          completion(.failure(createConnectionError(withChannelName: channelName)))
          return
        }
        if listResponse.count > 1 {
          let code: String = listResponse[0] as! String
          let message: String? = nilOrValue(listResponse[1])
          let details: String? = nilOrValue(listResponse[2])
          completion(.failure(ProxyApiTestsError(code: code, message: message, details: details)))
        } else {
          completion(.success(()))
        }
      }
    }
```

Its based on whether the Dart class has required callback methods to be instantiated.

stuartmorgan: Oh, I see now. Sorry about that; I missed that there were two different forms of it.

packages/pigeon/platform_tests/test_plugin/android/src/main/kotlin/com/example/test_plugin/ProxyApiTests.gen.kt
bparrishMines: Whoops I didn't realize this comment was made on a different file than the one below. But it looks like the [_writeFlutterMethod](https://github.com/flutter/packages/blob/main/packages/pigeon/lib/src/kotlin/kotlin_generator.dart#L1329) in the Kotlin generator contains `onWriteBody`, which gives the method more flexibility than the swift one. I think the main problem is that the method parameters for the method signature is different than the parameters passed in the `MethodCall`. Basically [_writeProxyApiNewInstanceMethod](https://github.com/flutter/packages/blob/2c3c9476996af6c3700167cab615972978829047/packages/pigeon/lib/src/swift/swift_generator.dart#L2270) does

```dart
void method(MyClass newInstance) {
  // Check if a new dart instance needs be created and return if not.

  // Retrieve all the field values needed by the Dart instance. This part is not included in a typical Flutter method.

  // use _writeFlutterMethodMessageCall with all the field values from above
}
```

There's probably a better way to do this and make it clearer that this is happening.

packages/pigeon/lib/src/kotlin/kotlin_generator.dart
```diff
@@ -1906,64 +1914,69 @@ if (wrapped == null) {
...
+                      $errorClassName("ignore-calls-error", "Calls to Dart are being ignored.", "")))''',
             );
           },
+          addTrailingNewline: false,
         );
         indent.writeScoped(
-          'if (pigeonRegistrar.instanceManager.containsInstance(${classMemberNamePrefix}instanceArg)) {',
-          '}',
+          'else if (pigeonRegistrar.instanceManager.containsInstance(${classMemberNamePrefix}instanceArg)) {',
+          '} ',
```
tarrinneal, line 1930: Pretty minor nit, but I think it's better to put the spaces before the `else` rather than after the `}` since the scoped statement doesn't need it, but the `else` statement does.

packages/pigeon/lib/src/swift/swift_generator.dart
```diff
@@ -2310,71 +2313,78 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
     indent.writeScoped('$methodSignature {', '}', () {
       indent.writeScoped(
         'if pigeonRegistrar.ignoreCallsToDart {',
-        '}',
+        '} ',
```
tarrinneal, line 2316: here too

packages/pigeon/lib/src/swift/swift_generator.dart
```diff
@@ -2310,71 +2313,78 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
...
           );
         },
+        addTrailingNewline: false,
       );
 
       indent.writeScoped(
-        'if pigeonRegistrar.instanceManager.containsInstance(pigeonInstance as AnyObject) {',
-        '}',
+        'else if pigeonRegistrar.instanceManager.containsInstance(pigeonInstance as AnyObject) {',
+        '} ',
```
tarrinneal, line 2332: here

tarrinneal (approved):
tiny nit, optional really

PR #8582, opened 2025-02-07
[url_launcher] When not fully loaded, clicking close causes the callback to not be triggered correctly.

StanleyCocos:
When SFSafariViewController is loading a webpage, if the close button is clicked during the loading process, it can prevent the callback from being triggered correctly.

- Fixes https://github.com/flutter/flutter/issues/161790
- Fixes https://github.com/flutter/flutter/issues/47271

## Pre-launch Checklist

- [x] I signed the [CLA].


StanleyCocos:
The missing test is because I need to first launch and quickly close it, then compare whether it can throw an exception as before. However, it seems that await or then can be tricky. I would greatly appreciate any guidance here.

stuartmorgan:
This should be testable with native unit tests that directly drive calls to the `URLLaunchSession`.

packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/URLLaunchSession.swift
```diff
@@ -12,6 +12,7 @@ final class URLLaunchSession: NSObject, SFSafariViewControllerDelegate {
 
   private let completion: OpenInSafariCompletionHandler
   private let url: URL
+  private var isLoadCompleted: Bool?
```
stuartmorgan, line 15: A nullable boolean should only be used when there are three conceptually different states. The code in this PR only actually recognizes two.

StanleyCocos:
> This should be testable with native unit tests that directly drive calls to the `URLLaunchSession`.

I’m really sorry, could you please provide more specific guidance? Thank you very much!


I intended to use the test case below, but since they are all simulated data, I can’t replicate the real fast closing behavior, which is causing me a lot of trouble.

```dart
test('launch and then close', () async {
    when(api.openUrlInSafariViewController(_webUrl)).thenAnswer((_) async {
      await Future<void>.delayed(const Duration(seconds: 2));
      return InAppLoadResult.success;
    });
    final UrlLauncherIOS launcher = UrlLauncherIOS(api: api);
    bool callback = false;
    Future<void> launchUrl() async {
      try {
        await launcher.launchUrl(_webUrl,
            const LaunchOptions(mode: PreferredLaunchMode.inAppWebView));
      } catch (_) {}
      // It won’t reach this point before the fix.
      callback = true;
    }

    launchUrl();
    await api.closeSafariViewController();
    expect(callback, true);
  });
```

stuartmorgan:
(Marking as Draft pending the test, so it doesn't show up in review queues.)

StanleyCocos:
Oh, my God. I forgot. Thanks for the reminder. I’ve added the test. If you have time, you can continue review

packages/url_launcher/url_launcher_ios/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## NEXT
+## 6.3.3
 
 * Updates minimum supported SDK version to Flutter 3.22/Dart 3.4.
+* Fixes When not fully loaded, clicking close causes the callback to not be triggered correctly.
```
loic-sharma, line 4:
```suggestion
* Ensures completion callback is invoked if the user dismisses the Safari view before the initial URL load completes.
```

packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/URLLaunchSession.swift
```diff
@@ -46,12 +47,16 @@ final class URLLaunchSession: NSObject, SFSafariViewControllerDelegate {
...
     }
+    isLoadCompleted = true
   }
 
   /// Called when the user finishes using the Safari view controller.
   ///
   /// - Parameter controller: The Safari view controller.
   func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
+    if !isLoadCompleted {
+      completion(.success(.failedToLoad))
```
loic-sharma, line 58: Is `.failedToLoad` the correct result here? Or we should add a new `InAppLoadResult.dismissed` result?

If this is the correct result, should we update this comment to indicate this case can happen if the user dismisses the Safari view before the initial load completes?

https://github.com/flutter/packages/blob/fd53793899d567856b7f20c7a196787c3bca1278/packages/url_launcher/url_launcher_ios/pigeons/messages.dart#L30-L31

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
```diff
@@ -131,6 +131,24 @@ final class URLLauncherTests: XCTestCase {
...
+      switch result {
+      case .success(let details):
+        XCTAssertEqual(details, .failedToLoad)
+      case .failure(let error):
+        XCTFail("Unexpected error: \(error)")
+      }
+      expectation.fulfill()
+    }
+    plugin.closeSafariViewController()
+    wait(for: [expectation], timeout: 30)
```
loic-sharma, line 149: Is this `timeout: 30` necessary or is `timeout: 1` would be sufficient?

From my understanding `plugin.closeSafariViewController()` calls the completion callback, which in turn calls `expectation.fulfill()`. The expectation should be already be fulfilled at this point.

I apologize if this is a silly question, I'm not very familiar with XCTest.

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
```diff
@@ -131,6 +131,24 @@ final class URLLauncherTests: XCTestCase {
     XCTAssertEqual(launcher.passedOptions?[.universalLinksOnly] as? Bool, true)
   }
 
+  func testLaunchSafariViewControllerWithClose() {
+    let launcher = FakeLauncher()
+    let plugin = createPlugin(launcher: launcher)
+
+    let expectation = XCTestExpectation(description: "completion called")
+    plugin.openUrlInSafariViewController(url: "https://flutter.dev") { result in
```
loic-sharma, line 139: I worry that this test would flake if `https://flutter.dev` loads before the test calls `closeSafariViewController`. Is there some URL we could test that is known to hang forever? Or am I overthinking this? cc @stuartmorgan

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
stuartmorgan: Our standard for expectation timeouts in the repo is 30 (older code still uses lower timeouts). Flutter style is [not to have timeouts at all](https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#never-check-if-a-port-is-available-before-using-it-never-add-timeouts-and-other-race-conditions) but that's awkward with XCTest APIs, so we just use a value that is absurdly long for unit tests.

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
stuartmorgan: I would not expect it to be possible for that to happen. Unless `SFSafariViewController`'s `didCompleteInitialLoad` callback happens on a non-main thread (which I wouldn't expect since it passed the view controller as a parameter), it shouldn't be possible for the callback to trigger until the runloop runs again, which I would think could not happen before the `wait` call.

There are some assumptions in there, so I wouldn't be confident in saying that it's impossible, but I'd be fine with waiting to see if it does actually happen before we add a bunch of extra logic (since the reliable way to avoid that would be dependency injection of a fake `SFSafariViewController` that we know won't load anything) to avoid I case that I strongly suspect is in fact impossible.

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
StanleyCocos: Under normal circumstances, wait(for: [expectation], timeout: 30) won’t actually wait, because the closure has already been triggered before it. The 30-second timeout is set to prevent accidentally removing this PR functionality. If this PR functionality is deleted, setting the wait time to 1 might not make sense, as the callback may never be reached for verification.

packages/url_launcher/url_launcher_ios/CHANGELOG.md
StanleyCocos: It has been adjusted.

packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/URLLaunchSession.swift
StanleyCocos: > If this is the correct result, should we update this comment to indicate this case can happen if the user dismisses the Safari view before the initial load completes?

Do you think it’s possible to modify the comment like this:

// The URL did not load successfully (or close the SFSafariViewController earlier)

packages/url_launcher/url_launcher_ios/example/ios/RunnerTests/URLLauncherTests.swift
loic-sharma: Gotcha, thanks for the background. Closing!

StanleyCocos:
@loic-sharma
Hi, could you take another look? 
Thank you!

packages/url_launcher/url_launcher_ios/pigeons/messages.dart
```diff
@@ -27,7 +27,7 @@ enum InAppLoadResult {
   /// The URL was successfully loaded.
   success,
 
-  /// The URL did not load successfully.
+  /// The URL did not load successfully (or close the SFSafariViewController earlier).
```
loic-sharma, line 30: What do you think of something like this?

```suggestion
  /// The URL did not load successfully (or the user closed the view controller before it loaded).
```

loic-sharma (approved):
This looks good to me, thanks for the contribution!

packages/url_launcher/url_launcher_ios/pigeons/messages.dart
StanleyCocos: I have made the changes. Thank you for your suggestion.

StanleyCocos:
@stuartmorgan  
Could you take some time to review this request again?
Thank you!

StanleyCocos:
cc @stuartmorgan  
Sorry to bother you, but could you take a look when you have time? This request seems to have been pending for quite a while.

packages/url_launcher/url_launcher_ios/ios/url_launcher_ios/Sources/url_launcher_ios/URLLaunchSession.swift
stuartmorgan: Returning this will [throw an exception saying there was an error loading the page](https://github.com/flutter/packages/blob/1b53da5bba9a7793eb72b96901338bedc4086885/packages/url_launcher/url_launcher_ios/lib/url_launcher_ios.dart#L167-L170), which seems misleading. I think a new Pigeon return type that Dart interprets to return `false` would be better.

StanleyCocos: I created a new dismissed state to represent the current status.

StanleyCocos: cc @stuartmorgan   
Could you take another look?

jmagman:
@stuartmorgan-g could you take a look at this and re-review? Thank you!

stuartmorgan-g (approved):
LGTM, sorry for the delay

PR #8588, opened 2025-02-08
[url_launcher]: (Web) When click on `Link` widget without calling `followLink` function still takes browser to that link

aarajput:
### Bug

If you wrap a ElevatedButton with Link widget on web, and don't pass followLink function to ElevatedButton, when you click on that ElevatedButton button, it still takes user to that link.

### Sample Code
```dart
import 'package:url_launcher/link.dart' as ul;

class GoogleButton extends StatelessWidget {
  const GoogleButton();

  @override
  Widget build(final BuildContext context) {
    return ul.Link(
      uri: Uri.parse('https://google.com'),
      target: ul.LinkTarget.blank,
      builder: (context, followLink) => ElevatedButton(
        onPressed: null,
        child: Text('Open Google'),
      ),
    );
  }
}
```

The issue was happening because `Link` widget is creating below element on web:
```html
<a rel="noreferrer noopener" aria-hidden="true" tabindex="-1" href="https://google.com" target="_blank" style="opacity: 0; display: block; width: 100%; height: 100%; cursor: unset;"></a>
```
Even when i am not passing `followLink` to `onPressed`, it still opens google in new tab. 

### Fix

We need to pass `onclick="event.preventDefault();"` to `a` html element so that it won't open link on it own.

This is how it looks after the fix:
```html
<a rel="noreferrer noopener" aria-hidden="true" tabindex="-1" href="https://google.com" target="_blank" style="opacity: 0; display: block; width: 100%; height: 100%; cursor: unset;" onclick="event.preventDefault();"></a>
```

* Fix Issue: [[url_launcher] [Web] Clicking on Link without calling followLink function still takes user to that link](https://github.com/flutter/flutter/issues/162927)

## Pre-launch Checklist

- [x] I signed the [CLA].
- [x] this PR is [test-exempt].

### DIFF
--- a/packages/url_launcher/url_launcher_web/CHANGELOG.md
+++ b/packages/url_launcher/url_launcher_web/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.4.1
+
+* Fixes when click on `Link` widget without calling `followLink` function still takes browser to that link.
+
 ## 2.4.0
 
 * Enhances handling of out-of-order events.
--- a/packages/url_launcher/url_launcher_web/lib/src/link.dart
+++ b/packages/url_launcher/url_launcher_web/lib/src/link.dart
@@ -499,6 +499,7 @@ class LinkViewController extends PlatformViewController {
 
     _element.setAttribute('aria-hidden', 'true');
     _element.setAttribute('tabIndex', '-1');
+    _element.setAttribute('onclick', 'event.preventDefault();');
 
     final Map<String, dynamic> args = <String, dynamic>{
       'id': viewId,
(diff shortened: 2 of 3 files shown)


stuartmorgan:
> We need to pass `onclick="event.preventDefault();"` to `a` html element so that it won't open link on it own.

This would break a lot of usage of `Link`. One of the critical use cases `Link` addresses is that browsers often restrict opening windows outside of the context of a synchronous UI interaction, and calling `openUrl` from a Flutter event isn't synchronous relative to the browser's interaction model. Having an actual link avoids that issue.

Given that, this is not a change we would take. The next step here is to understand the use case in the issue, so that we can see if there is a way to address that use case without breaking common `Link` usage.

aarajput:
Ok, what about i create a property like 'preventDefault', and if its true then i add event. preventDefault() in `a` tags?

ditman:
I think @mdebbar fixed this one recently:

* See: https://github.com/flutter/packages/pull/8675

PR #8627, opened 2025-02-13
[pigeon] Update task queue handling

stuartmorgan:
Updates the task queue implementation across all supporting generators to use one task queue for an entire host API, rather than one per method, so that multiple methods using task queues will have ordering guarantees amongst them. Also changes the Obj-C implementation to no-op on macOS instead of failing to compile, so that shared iOS/macOS code can use task queues on iOS without breaking macOS. (We could deliberately add a compile error with a clear error message in a macOS ifdef if we get reports of user confusion in the future, but my expectation is that no-op is the best option for most people.)

Related changes:
- Adds integration testing that task queues work, instead of just that the generator step succeeds.
  - For platforms that don't support task queues, they are still included in the integration test to ensure that 
  - I removed the now-vestigial background_platform_channels.dart.
- Adds Swift support, since it was easier to add than to skip it in the tests.

Fixes https://github.com/flutter/flutter/issues/162624
Fixes https://github.com/flutter/flutter/issues/111512

## Pre-launch Checklist

- [x] I signed the [CLA].


packages/pigeon/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 24.2.0
```
tarrinneal, line 1: This could be a breaking change couldn't it? I know that the old behavior was essentially just "wrong", but this could "break" what used to be unintentionally correct

packages/pigeon/lib/src/kotlin/kotlin_generator.dart
```diff
@@ -654,6 +654,14 @@ if (wrapped == null) {
         indent.addScoped('{', '}', () {
           indent.writeln(
               r'val separatedMessageChannelSuffix = if (messageChannelSuffix.isNotEmpty()) ".$messageChannelSuffix" else ""');
+          String? serialBackgroundQueue;
+          if (api.methods.any((Method m) =>
+              m.taskQueueType == TaskQueueType.serialBackgroundThread)) {
+            serialBackgroundQueue = 'taskQueue';
+            serialBackgroundQueue = 'taskQueue';
```
tarrinneal, line 661: I think one time is probably good enough

packages/pigeon/lib/src/swift/swift_generator.dart
```diff
@@ -9,6 +9,7 @@ import '../ast.dart';
 import '../functional.dart';
 import '../generator.dart';
 import '../generator_tools.dart';
+import '../pigeon_lib.dart';
```
tarrinneal, line 12: Is this necessary?

packages/pigeon/platform_tests/shared_test_plugin_code/lib/integration_tests.dart
```diff
@@ -2871,6 +2871,28 @@ void runPigeonIntegrationTests(TargetGenerator targetGenerator) {
...
+    // https://github.com/flutter/flutter/issues/93945
+    // Rather than skip the test, this changes the expectation, so that there
+    // is test coverage of the code path, even though the actual backgrounding
+    // doesn't happen. This is especially important for macOS, which may need to
+    // share generated code with iOS, falling back to the main thread since
+    // background is not supported.
+    final bool taskQueuesSupported =
+        defaultTargetPlatform == TargetPlatform.android ||
+            defaultTargetPlatform == TargetPlatform.iOS;
+    expect(await api.taskQueueIsBackgroundThread(), taskQueuesSupported);
```
tarrinneal, line 2893: This is pretty funny, makes sense though

tarrinneal (approved):
Looks good. I know we discussed the breaking change comment before, but I wanted to check again. A couple nits

packages/pigeon/CHANGELOG.md
stuartmorgan: The old behavior had undefined ordering, and one of the possible orderings was the one that this will guarantee. So any code that reliably behaved correctly still will, because this is one of the orderings it must have handled.

packages/pigeon/lib/src/swift/swift_generator.dart
stuartmorgan: This is for `TaskQueueType`. Per offline discussion, I moved that type to a new file instead.

packages/pigeon/lib/src/kotlin/kotlin_generator.dart
stuartmorgan: Fixed.

PR #8629, opened 2025-02-13
[camera_android_camerax] Fix 90°-off preview rotation

camsim99:
Fixes 90°-off preview rotation without locked capture orientation. Part of https://github.com/flutter/flutter/issues/154241.

The major changes in this PR:

1. Ensures the preview widget is rebuilt when a new device orientation is detected by changing the `buildPreview` to return a `StatefulWidget` instead of a `Texture` that wouldn't pick up changes in device orientation.
2. Uses [`handlesCropAndRotation`](https://api.flutter.dev/javadoc/io/flutter/view/TextureRegistry.SurfaceProducer.html#handlesCropAndRotation()) to detect if the preview is already correctly rotated or not. This API was added exactly for that purpose.
3. Corrects the preview when needed using https://developer.android.com/media/camera/camera2/camera-preview#orientation_calculation (also subtracts rotation applied in `CameraPreview` widget).

See https://github.com/flutter/flutter/issues/154241#issuecomment-2655098620 for slightly more context if desired.

## Pre-launch Checklist

- [x] I signed the [CLA].


cnkygmr:
When exactly will this update end? I will update it but I can't.

camsim99:
> @camsim99 Do these fixes also address [flutter/flutter#156974](https://github.com/flutter/flutter/issues/156974)?

This wouldn't fix that issue, but I'll follow up there.

packages/camera/camera_android_camerax/test/android_camera_camerax_test.dart
```diff
@@ -1311,21 +1314,65 @@ void main() {
...
 
+    camera.handlesCropAndRotation = true;
     final Widget widget = camera.buildPreview(cameraId);
 
-    expect(widget is Texture, isTrue);
+    expect(widget, isA<Texture>());
     expect((widget as Texture).textureId, cameraId);
   });
 
+  test(
```
matanlurey, line 1333: This is a good start, but I'd split this into separate tests.

This is even complicated enough where I'd support a `preview_rotation_test.dart`.

Matrix of options I'd spect:
1. When `handlesCropAndRotation = true, the preview is <not rotated or w/e it is>`
  - Sub test: Verify rotation the stream still works as expected
2. When `handlesCropAndRotation = false`
  - Sub test: Based on the initial orientation; 1 test for each orientation
  - Sub test: Preview changes as orientation changes (1 test is fine IMO)
  - Sub test: How `sensorOrientationDegrees` plays into things
  - Sub test: How `facingSign` plays into things

Basically, we'd want to know that all of these parameters are being used to render the camera preview as expected. It should be possible to read this test suite and understand the _contract_ of the camera preview.

packages/camera/camera_android_camerax/test/rotated_preview_test.dart
matanlurey, line 1: I don't find this particularly useful, IMO. I would treat this widget as an implementation detail; we could _delete it_ and the test could still pass.

The same concepts you used though could be copied into _a_ `rotated_preview_test` similar to what I discussed in the previous file; let's test the user-visible API, not the private API.

(The only reason this widget isn't `_Private` is for readability/organization)

packages/camera/camera_android_camerax/test/rotated_preview_test.dart
```diff
@@ -0,0 +1,150 @@
...
+  double calculateExpectedPreviewRotation(double sensorOrientationDegrees,
+      double deviceOrientationDegrees, int facingSign) {
+    return ((sensorOrientationDegrees -
+                deviceOrientationDegrees * facingSign +
+                360) %
+            360) -
+        deviceOrientationDegrees;
+  }
+
+  void findAndTestRotatedPreviewHasExpectedRotation(
```
matanlurey, line 32: I already left comments about changing this test, but I wanted to comment on this.

This is a duplication of the implementation code, so we now we have the possibility of getting it wrong, twice, but all the tests passing.

Instead, like I mentioned in the previous file, I'd want to see something like:

"When the camera preview does not handle crop and rotation, and sensor is X, and device is Y, the preview is rotated 90 degrees clockwise"

In fact, the test case hopefully would look like that:

```dart
const _$90DegreesClockwise = 1;

RotatedBox widget = ...
expect(
  widget.quarterTurns,
  _$90DegreesClockwise,
  reason: 'Expected 1 quarter turn, which is 90 degrees clockwise',
);
```

It will help a lot with making the error messages (on CI and locally) readable as well.

matanlurey (changes requested):
From an implementation POV, LGTM, great work.

From a testing POV, we can rely a bit less on mocks and testing internal state, and more on testing the exported state of the code itself; this will make the system less fragile (we're doing _closer_ to e2e tests in a unit test environment) and more easily understood.

There are external guides talking about API testing, but I'll message one in Flutter Android chat that is internal I like a lot.

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/PreviewHostApiImpl.java
```diff
@@ -179,6 +179,15 @@ public void setTargetRotation(@NonNull Long identifier, @NonNull Long rotation)
     preview.setTargetRotation(rotation.intValue());
   }
 
+  @NonNull
+  public Boolean surfaceProducerHandlesCropAndRotation() {
+    if (flutterSurfaceProducer != null) {
+      return flutterSurfaceProducer.handlesCropAndRotation();
+    }
+    throw new IllegalStateException(
```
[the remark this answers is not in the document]
camsim99: Added it to the other method that is at risk of this error.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -239,11 +240,16 @@ class AndroidCameraCameraX extends CameraPlatform {
...
   @visibleForTesting
-  late int sensorOrientation;
+  late double sensorOrientationDegrees;
 
-  /// Subscription for listening to changes in device orientation.
-  StreamSubscription<DeviceOrientationChangedEvent>?
-      _subscriptionForDeviceOrientationChanges;
+  /// Whether or not the Android surface producer automatically handles
+  /// correcting the rotation of camera previews for the device this plugin runs on.
+  @visibleForTesting
```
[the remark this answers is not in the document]
camsim99: Left these as visible so that we can set them directly in the `preview_rotation_test.dart`. These values being fetched as expected is tested in `android_camera_camerax_test.dart`.

packages/camera/camera_android_camerax/test/android_camera_camerax_test.dart
camsim99: Added `preview_rotation_test.dart` with the intention of covering all of these cases.

> Sub test: Verify rotation the stream still works as expected

Not 100% sure what you meant but I:
1. Already have a test in `android_camera_camerax_test..dart` that tests the `onDeviceOrientationChanged()`.
2. Ensured that when `handlesCropAndRotation` is `false`, no `RotatedBox` is used in `preview_rotation_test.dart`.
3. Added to `camera/camera`'s `CameraPreview` test to ensure that its rotation behaves as expected. cc @bparrishMines because this is why your review was requested :)

3 made me realize a case that this PR will not cover; the preview rotation logic written here does not account for locked capture orientations. If it is locked, `CameraPreview` is rotated according to that instead of the device orientation. I'll file an issue and follow this PR up with a fix for that.

@matanlurey let me know if these cases don't cover what you were talking about!

camsim99: > 3 made me realize a case that this PR will not cover; the preview rotation logic written here does not account for locked capture orientations. If it is locked, CameraPreview is rotated according to that instead of the device orientation. I'll file an issue and follow this PR up with a fix for that.

Filed https://github.com/flutter/flutter/issues/163857

packages/camera/camera/test/camera_preview_test.dart
```diff
@@ -139,6 +139,16 @@ class FakeController extends ValueNotifier<CameraValue>
 
 void main() {
   group('RotatedBox (Android only)', () {
+    int getClockwiseQuarterTurnsFromDeviceOrientation(
```
matanlurey, line 142: nit: So this is a personal preference thing, and does not have to be changed.

I generally dislike doing computations or lookups in a test, particularly in a for-loop, because it becomes very difficult to understand the specific test cases. Consider the following (simulated) test case instead:

```dart
test('when the device is portaitUp, the preview is not rotated', () async {
  setOrientation(DeviceOrientation.portraitUp);
  await pumpChangesOrWhatever();
  expect(widget.quarterTurns, 0);
});

test('when the device is landscapeRight, the preview is rotated accordingly', () async {
  setOrientation(DeviceOrientation.landscapeRight);
  await pumpChangesOrWhatever();
  expect(widget.quarterTurns, 1, reason: '90 degrees clockwise');
});
```

Notice how the complexity (of setting up the widget) is still hidden and we don't have (much) duplicate code, but it's much easier to read "oh, we have 4 tests, and here is how the tests work".

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/PreviewHostApiImpl.java
```diff
@@ -179,6 +179,15 @@ public void setTargetRotation(@NonNull Long identifier, @NonNull Long rotation)
     preview.setTargetRotation(rotation.intValue());
   }
 
+  @NonNull
+  public Boolean surfaceProducerHandlesCropAndRotation() {
```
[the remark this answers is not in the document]
matanlurey: Ah great, thank you!

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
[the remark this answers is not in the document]
matanlurey: > Left these as visible so that we can set them directly in the preview_rotation_test.dart.

Is it possible to write that test _without_ setting these variables directly? By setting them directly, we are bypassing the logic that is initially used to set them (in `createCameraWithSettings`) which if I remember correctly _was_ a significant source of difficulty.

packages/camera/camera_android_camerax/lib/src/rotated_preview.dart
```diff
@@ -0,0 +1,117 @@
...
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+
+/// Widget that rotates the camera preview to be upright according to the
+/// current device orientation.
+final class RotatedPreview extends StatefulWidget {
```
matanlurey, line 12: nit: We could use `@internal` from `package:meta` to help others understand "this is an implementation detail".

As a bonus we'll get a lint warning if we ever accidentally expose this in the public API.

https://pub.dev/documentation/meta/latest/meta/internal-constant.html

packages/camera/camera_android_camerax/test/preview_rotation_test.dart
matanlurey, line 1: These tests look really good.

My only other comment (left in a previous file) is it would be better to go through the `createCameraWithSettings` branch instead of carving out the `@visibleForTesting` members.

If that is really difficult (for mocking or testability reasons), and you want help, let's pair (it's possible I'll agree and give up this point, but I'd want to give it a shot).

matanlurey (commented):
Mostly LGTM. I'd like to see us try not to use `@visibleForTesting` at all to manipulate the internal state for tests, but it's possible that is not doable or requires a lot more work than I expect. PTAL.

packages/camera/camera_android_camerax/lib/src/rotated_preview.dart
```diff
@@ -0,0 +1,117 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+
+/// Widget that rotates the camera preview to be upright according to the
+/// current device orientation.
```
reidbaker, line 11: Nit: Is this actually based the device orientation or the "app" orientation. 
I think it is based on the apps orientation. This is especially important as newer versions of android allow forcing the orientation of apps independent of the sensor.

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -239,11 +240,16 @@ class AndroidCameraCameraX extends CameraPlatform {
 
   /// The camera sensor orientation.
   @visibleForTesting
-  late int sensorOrientation;
+  late double sensorOrientationDegrees;
```
reidbaker, line 243: Can you expand on the dart doc here that sensor orientation degrees can change when the camera being used is changed but is independent of device orientation or app orientation. (assuming that is true, I think so based on reading this pr. )

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/PreviewHostApiImpl.java
```diff
@@ -155,7 +155,10 @@ String getProvideSurfaceErrorDescription(int resultCode) {
   public void releaseFlutterSurfaceTexture() {
     if (flutterSurfaceProducer != null) {
       flutterSurfaceProducer.release();
+      return;
     }
+    throw new IllegalStateException(
```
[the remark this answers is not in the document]
camsim99: We throw exceptions like this throughout the `camera_android_camerax` implementation to signal that there may be an error in the plugin implementation to plugin authors, so this is not meant to be caught by users.

If you have greater concerns about this pattern, I suggest we file an issue and address this for the plugin as a whole in a follow up PR.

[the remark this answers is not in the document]
reidbaker: OK, Non blocking question. Do we catch this (or other) exception(s) or is it thrown to users code?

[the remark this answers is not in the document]
camsim99: We don't catch it. They're really just supposed to help debug.

packages/camera/camera_android_camerax/test/preview_rotation_test.dart
camsim99: Done. Also, did a minor refactor to remove a duplicate call to get the sensor orientation. Also also, I'll note it requires a call to `availableCameras` and `createCameraWithSettings`, so both all of the tests go through that path. PTAL

matanlurey (approved):
LGTM! Thank you!

PR #8658, opened 2025-02-19
[camera_android]fixed some android devices when you calling real quick to stop camera…

OnClickListener2048:
fixed some android devices when you calling real quick to stop camera recording, It will throw a RuntimeException instead IllegalStateException
also it fixed https://github.com/flutter/flutter/issues/163279

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_android/CHANGELOG.md
+++ b/packages/camera/camera_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.10+1
+
+* Fixes Android Platform `mediaRecorder.stop` throws exceptiq did not get caught
+
 ## 0.10.10
 
 * Adds API support query for image streaming.
--- a/packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
+++ b/packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
@@ -107,8 +107,10 @@ class Camera
   private CameraProperties cameraProperties;
   private final CameraFeatureFactory cameraFeatureFactory;
   private final Activity activity;
+
   /** A {@link CameraCaptureSession.CaptureCallback} that handles events related to JPEG capture. */
   private final CameraCaptureCallback cameraCaptureCallback;
+
   /** A {@link Handler} for running tasks in the background. */
   Handler backgroundHandler;
 
@@ -119,12 +121,15 @@ class Camera
   CameraCaptureSession captureSession;
   @VisibleForTesting ImageReader pictureImageReader;
   ImageStreamReader imageStreamReader;
+
   /** {@link CaptureRequest.Builder} for the camera preview */
   CaptureRequest.Builder previewRequestBuilder;
 
   @VisibleForTesting MediaRecorder mediaRecorder;
+
   /** True when recording video. */
   boolean recordingVideo;
+
   /** True when the preview is paused. */
   @VisibleForTesting boolean pausedPreview;
 
@@ -132,6 +137,7 @@ class Camera
 
   /** Holds the current capture timeouts */
   private CaptureTimeoutsWrapper captureTimeouts;
+
   /** Holds the last known capture properties */
   private CameraCaptureProperties captureProps;
 
@@ -294,8 +300,10 @@ class Camera
 
     MediaRecorderBuilder mediaRecorderBuilder;
 
-    // TODO(camsim99): Revert changes that allow legacy code to be used when recordingProfile is null
-    // once this has largely been fixed on the Android side. https://github.com/flutter/flutter/issues/119668
+    // TODO(camsim99): Revert changes that allow legacy code to be used when recordingProfile is
+    // null
+    // once this has largely been fixed on the Android side.
+    // https://github.com/flutter/flutter/issues/119668
     if (SdkCapabilityChecker.supportsEncoderProfiles() && getRecordingProfile() != null) {
       mediaRecorderBuilder =
           new MediaRecorderBuilder(
@@ -367,7 +375,8 @@ class Camera
             cameraDevice = new DefaultCameraDeviceWrapper(device);
             try {
               startPreview();
-              if (!recordingVideo) { // only send initialization if we werent already recording and switching cameras
+              if (!recordingVideo) { // only send initialization if we werent already recording and
+                                     // switching cameras
                 dartMessenger.sendCameraInitializedEvent(
                     resolutionFeature.getPreviewSize().getWidth(),
                     resolutionFeature.getPreviewSize().getHeight(),
@@ -859,11 +868,16 @@ class Camera
       // Ignore exceptions and try to continue (changes are camera session already aborted capture).
     }
 
+    boolean haveCaptureError = false;
+
     try {
       mediaRecorder.stop();
+    } catch (IllegalStateException e) {
+      // Ignore exceptions and try to continue (changes are camera session already aborted capture).
     } catch (RuntimeException e) {
-      // Ignore exceptions and try to continue (a RuntimeException is intentionally thrown to the application
-      // , if no valid audio/ video data has been received when stop() is called).
+      // We marked that mediaRecorder.stop() is failed when no valid audio/video data has been
+      // received when stop() is called, so we should reset mediaRecorder and start preview.
+      haveCaptureError = true;
     }
 
     mediaRecorder.reset();
@@ -874,6 +888,11 @@ class Camera
     }
     String path = captureFile.getAbsolutePath();
     captureFile = null;
+
+    // if there was an RuntimeException during stopping the video recording, return null to let the user know.
+    if (haveCaptureError) {
+      return null;
+    }
     return path;
   }
 
@@ -1158,7 +1177,8 @@ class Camera
   }
 
   public void startPreview() throws CameraAccessException, InterruptedException {
-    // If recording is already in progress, the camera is being flipped, so send it through the VideoRenderer to keep the correct orientation.
+    // If recording is already in progress, the camera is being flipped, so send it through the
+    // VideoRenderer to keep the correct orientation.
     if (recordingVideo) {
       startPreviewWithVideoRendererStream();
     } else {
@@ -1362,7 +1382,8 @@ class Camera
           "setDescriptionWhileRecordingFailed", "Device was not recording", null);
     }
 
-    // See VideoRenderer.java; support for this EGL extension is required to switch camera while recording.
+    // See VideoRenderer.java; support for this EGL extension is required to switch camera while
+    // recording.
     if (!SdkCapabilityChecker.supportsEglRecordableAndroid()) {
       throw new Messages.FlutterError(
           "setDescriptionWhileRecordingFailed",
--- a/packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
+++ b/packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
@@ -855,10 +855,17 @@ class Camera
     try {
       closeRenderer();
       captureSession.abortCaptures();
-      mediaRecorder.stop();
-    } catch (CameraAccessException | RuntimeException e) {
+    } catch (CameraAccessException e) {
       // Ignore exceptions and try to continue (changes are camera session already aborted capture).
     }
+
+    try {
+      mediaRecorder.stop();
+    } catch (RuntimeException e) {
+      // Ignore exceptions and try to continue (a RuntimeException is intentionally thrown to the application
+      // , if no valid audio/ video data has been received when stop() is called).
+    }
+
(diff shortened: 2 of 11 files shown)


OnClickListener2048:
@stuartmorgan

OnClickListener2048:
I don't know why the [Linux repo_checks](https://github.com/flutter/packages/pull/8658/checks?check_run_id=37439163385) always fail, I didn't change any of that code

stuartmorgan:
> I don't know why the [Linux repo_checks](https://github.com/flutter/packages/pull/8658/checks?check_run_id=37439163385) always fail, I didn't change any of that code

You can click through to the [failure details](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8722546276683368929/+/u/Run_package_tests/CHANGELOG_and_version_validation/stdout) to see why a CI task has failed.

> * [x]  I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning), or this PR is [exempt from version changes](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version).
> * [x]  I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style), or this PR is [exempt from CHANGELOG changes](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog).
> ...
> * [x]  I added new tests to check the change I am making, or this PR is [test-exempt](https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests).

Please don't check boxes in the checklist without doing those steps. The purpose of the checklist is to guide contributors through the required parts of our PR process, not just to have a list of checked boxes. Please mark this as ready for review once the steps are completed.

packages/camera/camera_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.10+1
+
+* Fixes Android Platform `mediaRecorder.stop` throws exceptiq did not get caught
```
camsim99, line 3: Rewording suggestion:
```suggestion
* Fixes Android Platform `mediaRecorder.stop` uncaught exception.
```

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -856,7 +856,7 @@ public String stopVideoRecording() {
       closeRenderer();
       captureSession.abortCaptures();
       mediaRecorder.stop();
```
camsim99, line 858: It doesn't look like `MediaRecorder.stop` throws a `RuntimeException`: https://developer.android.com/reference/android/media/MediaRecorder#stop(). So, it must be another call. I think we should identify which call it is and use a separate try/catch to catch it since this catch as-is is ambiguous.

camsim99 (changes requested):

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
OnClickListener2048: Thank you for your review!

According to those existing issues: 
https://github.com/flutter/flutter/issues/121459
https://github.com/flutter/flutter/issues/79306
https://github.com/flutter/flutter/issues/93686

It looks like somehow it does throws a `RuntimeException` when you calling `MediaRecorder.stop` and it does on a unit test I quote https://github.com/flutter/flutter/issues/93686#issuecomment-1852889118

with my physical phone, brand: OPPO Reno 9 PHM110,  I can reproduce this error:`Failed to handle method call (Ask Gemini)
                                         java.lang.RuntimeException: stop failed.
                                         	at android.media.MediaRecorder.stop(Native Method)
                                         	at io.flutter.plugins.camera.Camera.stopVideoRecording(Camera.java:810)
                                         	at io.flutter.plugins.camera.MethodCallHandlerImpl.onMethodCall(MethodCallHandlerImpl.java:129)
                                         	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:258)
                                         	at io.flutter.embedding.engine.dart.DartMessenger.invokeHandler(DartMessenger.java:295)
                                         	at io.flutter.embedding.engine.dart.DartMessenger.lambda$dispatchMessageToQueue$0$io-flutter-embedding-engine-dart-DartMessenger(DartMessenger.java:322)
                                         	at io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0.run(Unknown Source:12)
                                         	at android.os.Handler.handleCallback(Handler.java:958)
                                         	at android.os.Handler.dispatchMessage(Handler.java:99)
                                         	at android.os.Looper.loopOnce(Looper.java:257)
                                         	at android.os.Looper.loop(Looper.java:368)
                                         	at android.app.ActivityThread.main(ActivityThread.java:8839)
                                         	at java.lang.reflect.Method.invoke(Native Method)
                                         	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:572)
                                         	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1049)`
 
So I believe that this PR is a solution for this issue

camsim99: Ah I see. I apologize; I missed the note in the docs about RuntimeException being thrown.

According to that note, sounds like we need to ensure proper cleanup is done if a RuntimeException is thrown, since the video would not be valid. I'm concerned about burying the exception versus, for example, forwarding a `FlutterError` to the developer so they can handle anything that relies on a valid video.

Let me know what you think, and @stuartmorgan if you have thoughts on this let me know.

stuartmorgan: Yes, this should be reported back to the Dart code somehow so that the Dart code can throw a `CameraException`.

If this is the only error case for stopping, I would suggest changing the Pigeon `stopVideoRecording()` method to return a nullable string and document that a null return means that it failed to produce video data, then have the Dart code convert that to a `CameraException`. (If there are other failures, we could use a return class instead, containing a `String?` for the file path, and an enum and/or error message for different errors.)

OnClickListener2048: > Yes, this should be reported back to the Dart code somehow so that the Dart code can throw a `CameraException`.
> 
> If this is the only error case for stopping, I would suggest changing the Pigeon `stopVideoRecording()` method to return a nullable string and document that a null return means that it failed to produce video data, then have the Dart code convert that to a `CameraException`. (If there are other failures, we could use a return class instead, containing a `String?` for the file path, and an enum and/or error message for different errors.)

Hi Cam & stuart , Thanks for your enlightenment.
  I agree that we need to put the error to the dart code side and let the developer deal with it.

 `mediaRecorder.stop()` would throws two type of Exception:
 `IllegalStateException` if it is called before start()
  `RuntimeException` if no valid audio/ video data has been received when stop() is called. This happens if stop() is called immediately after start()
  the former code only catch and ignored the `IllegalStateException`, do you think we need to wrap these errors into a class and throw it up to dart, or we still ignore the `IllegalStateException` and return a nullable string as we handle the `RuntimeException`

stuartmorgan: Known cases of exceptions [should not be surfaced as raw `PlatformException`s](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#platform-exception-handling), so both should be caught, and translated in Dart (`StateError`, in the case of the `IllegalStateException` you described).

OnClickListener2048: > Known cases of exceptions [should not be surfaced as raw `PlatformException`s](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#platform-exception-handling), so both should be caught, and translated in Dart (`StateError`, in the case of the `IllegalStateException` you described).

As far as I know, In Dart `stopVideoRecording()` the former code does not throw any error , So I think what we are doing would make a breaking change for developers

stuartmorgan: Sorry, I missed the part of your comment where you said that `IllegalStateException` is currently being caught and silently ignored.

So for the purposes of this PR, that behavior should stay as-is, and only the runtime exception should get new handling.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -855,10 +864,22 @@ public String stopVideoRecording() {
...
+
+    boolean haveCaptureError = false;
+
+    try {
       mediaRecorder.stop();
-    } catch (CameraAccessException | IllegalStateException e) {
+    } catch (IllegalStateException e) {
       // Ignore exceptions and try to continue (changes are camera session already aborted capture).
+    } catch (RuntimeException e) {
+      // We marked that mediaRecorder.stop() is failed when no valid audio/video data has been
```
stuartmorgan-g, line 878: Please don't use "we" in comments; it's not clear to me who "we" is here.

What is this comment referring to exactly? What does "marked [...] is failed" mean, and where does that happen?

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -855,10 +864,22 @@ public String stopVideoRecording() {
...
+      // received when stop() is called, so we should reset mediaRecorder and start preview.
```
stuartmorgan-g, line 879: The things that this comment says should be done aren't done in the `haveCaptureError` block, so it's not clear to me what this is describing.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -867,6 +888,12 @@ public String stopVideoRecording() {
     }
     String path = captureFile.getAbsolutePath();
     captureFile = null;
+
+    // if there was an RuntimeException during stopping the video recording, return null to let the
```
stuartmorgan-g, line 892: This comment doesn't say anything that the code it is commenting doesn't trivially indicate (other than claiming that it's because of a RuntimeException, which could easily become incorrect over time), so it can just be removed.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Messages.java
```diff
@@ -942,6 +942,7 @@ public interface Result<T> {
     /** Failure case callback method for handling errors. */
     void error(@NonNull Throwable error);
   }
+
```
stuartmorgan-g, line 945: Are you using a Java formatter other than the one the repo tool's `format` runs? If the Pigeon version didn't change, I don't see why there would be newlines added in this file.

packages/camera/camera_android/lib/src/android_camera.dart
```diff
@@ -216,7 +216,11 @@ class AndroidCamera extends CameraPlatform {
 
   @override
   Future<XFile> stopVideoRecording(int cameraId) async {
-    final String path = await _hostApi.stopVideoRecording();
+    final String? path = await _hostApi.stopVideoRecording();
+    if (path == null) {
+      throw StateError(
```
stuartmorgan-g, line 221: Is it possible for a plugin client to prevent this case from happening? If not, `StateError` isn't a useful thing to throw; per docs "An Error object represents a program failure that the programmer should have avoided." If so, the error should explain what the developer did incorrectly.

If this is unforeseeable, then it should throw some kind of Exception, not an Error, so that it can be caught and handled at runtime.

packages/camera/camera_android/pigeons/messages.dart
```diff
@@ -153,8 +153,8 @@ abstract class CameraApi {
   void startVideoRecording(bool enableStream);
 
   /// Ends video recording on the camera with the given ID and returns the path
-  /// to the resulting file.
-  String stopVideoRecording();
+  /// to the resulting file. return null if no valid audio/video data was generated.
```
stuartmorgan-g, line 156: "... and returns the path to the resulting file, or null if no valid ..."

stuartmorgan-g (changes requested):

jesswrd:
from android triage: @OnClickListener2048 Are you still working on this?

OnClickListener2048:
> from android triage: @OnClickListener2048 Are you still working on this?

Yes, I've been a little busy those days, but I'll fix these asap

reidbaker:

PR #8675, opened 2025-02-21
[url_launcher][web] Prevent browser from navigating when followLink isn't called

mdebbar:
When a DOM click event is received, we need to make a decision before the end of the event loop whether to allow the browser to navigate or not. At the end of the event loop, the browser will perform the navigation if we don't prevent it.

The problem occurs when the `followLink` signal arrives AFTER the event loop (this can happen in semantics mode when the web engine uses a debouncer to queue events and send them to the framework after some delay). This leads to a situation where we can't make a definitive decision by the end of the event loop, so this PR does the following:

1. [best case] If the `followLink` signal is received before the end of the event loop, we let the browser do the navigation for a full web link experience.

2. [meh case] If no `followLink` signal is received before the end of the event loop, we PREVENT the browser from navigating. But we keep waiting for the `followLink` signal. If the signal arrives a bit later, we fallback to a programmatic navigation through the `launchUrl` API.

Fixes https://github.com/flutter/flutter/issues/162927
Fixes https://github.com/flutter/flutter/issues/162408

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/url_launcher/url_launcher_web/example/integration_test/link_widget_test.dart
+++ b/packages/url_launcher/url_launcher_web/example/integration_test/link_widget_test.dart
@@ -445,25 +445,23 @@ void main() {
       await followLinkCallback!();
       await Future<void>.delayed(const Duration(seconds: 1));
       final html.Event event1 = _simulateClick(anchor);
+      await Future<void>.delayed(const Duration(seconds: 1));
 
       // The link shouldn't have been triggered.
       expect(pushedRouteNames, isEmpty);
       expect(testPlugin.launches, isEmpty);
-      expect(event1.defaultPrevented, isFalse);
-
-      await Future<void>.delayed(const Duration(seconds: 1));
+      expect(event1.defaultPrevented, isTrue);
 
       // Signals with large delay (in reverse order).
       final html.Event event2 = _simulateClick(anchor);
       await Future<void>.delayed(const Duration(seconds: 1));
       await followLinkCallback!();
+      await Future<void>.delayed(const Duration(seconds: 1));
 
       // The link shouldn't have been triggered.
       expect(pushedRouteNames, isEmpty);
       expect(testPlugin.launches, isEmpty);
-      expect(event2.defaultPrevented, isFalse);
-
-      await Future<void>.delayed(const Duration(seconds: 1));
+      expect(event2.defaultPrevented, isTrue);
 
       // A small delay is okay.
       await followLinkCallback!();
@@ -473,6 +471,7 @@ void main() {
       // The link should've been triggered now.
       expect(pushedRouteNames, <String>['/foobar']);
       expect(testPlugin.launches, isEmpty);
+      // (default is prevented here because the link is internal)
       expect(event3.defaultPrevented, isTrue);
     });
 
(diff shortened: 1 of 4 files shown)


packages/url_launcher/url_launcher_web/example/integration_test/link_widget_test.dart
```diff
@@ -972,6 +971,109 @@ void main() {
...
+
+      final html.Element semanticsHost =
+          html.document.createElement('flt-semantics-host');
+      html.document.body!.append(semanticsHost);
+      final html.Element semanticsAnchor = html.document.createElement('a')
+        ..setAttribute('id', 'flt-semantic-node-99')
+        ..setAttribute('flt-semantics-identifier', 'test-link-71')
+        ..setAttribute('href', uri.toString())
+        ..textContent = 'My Text Link';
+      semanticsHost.append(semanticsAnchor);
```
ditman, line 1059: Is this needed because we can't turn on semantics in the test, so the engine creates all the semantics nodes for the Link?

packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -254,7 +264,26 @@ class LinkTriggerSignals {
...
+      // navigate programmatically using `launchUrl`.
+      scheduleMicrotask(() {
+        if (!_hasDomEvent) {
+          // By the time the microtask runs, the link may have already been triggered. In that case,
+          // we want to do nothing here.
+          return;
+        }
+        canBrowserNavigate = false;
+        mouseEvent.preventDefault();
+      });
```
ditman, line 285: Instead of calling `_scheduleReset()`, should we just call `reset` directly at the beginning of this microtask?

In this case we really don't have to wait a "long" timeout to reset the state of the plugin, do we? Is it because of how `followLink` might happen at a random time?

ditman (approved):
Ship it! This code is getting super complicated to just follow a Link on the web, but I guess there are considerations from other platforms that I'm not taking into consideration :)

packages/url_launcher/url_launcher_web/example/integration_test/link_widget_test.dart
mdebbar: Semantics mode doesn't work end-to-end in tests, so I have to mimic what the engine would've created.

packages/url_launcher/url_launcher_web/lib/src/link.dart
mdebbar: Yeah, we have to wait for `followLink`, which may (or may not) come later. So the timer is really acting like a deadline for waiting for `followLink`. If `followLink` doesn't show up by that time, we assume it's never coming.

In the normal case, `followLink` should arrive before the end of the microtask and everything works beautifully. But in semantics mode, we have a thing called "click debouncer" that delays pointer events. That's what causes the framework's `onTap` callback to be fired late.

mdebbar:
> Ship it! This code is getting super complicated to just follow a Link on the web, but I guess there are considerations from other platforms that I'm not taking into consideration :)

Agreed! To explain the reason behind the complication, there are 2:

1. We want to comply with the framework's hit testing behavior. E.g. a circular button may create a rectangular `<a>` tag. So a simple click listener on the `<a>` tag will trigger the link even when the click didn't land on the actual circular button. This is why we need to wait for 2 signals: DOM event + followLink (the latter is called when the app/framework register a hit test).

2. Semantics mode adds another layer of complexity. One example is the click debouncer affecting the order and timing of events.

PR #8685, opened 2025-02-24
[video_player] Fix layout issue caused by `Transform.rotate` not affecting space calculation.

fa0311:
This pull request resolves [#159724](https://github.com/flutter/flutter/issues/159724).
The issue occurs because `Transform.rotate` is used when `rotation` is not 0. Since `Transform` does not affect space calculations, this leads to the issues.

This is related to [#7846](https://github.com/flutter/packages/pull/7846). 

## Pre-launch Checklist

- [x] I signed the [CLA].


fa0311:
@stuartmorgan 

I have added tests to verify the fix:

- Updated existing tests to check for RotatedBox instead of Transform.rotate.
- Added a new test to ensure RotatedBox is not used when rotation is not a multiple of 90.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -888,12 +887,15 @@ class _VideoPlayerWithRotation extends StatelessWidget {
...
 
   @override
-  Widget build(BuildContext context) => rotation == 0
-      ? child
-      : Transform.rotate(
-          angle: rotation * math.pi / 180,
-          child: child,
-        );
+  Widget build(BuildContext context) {
+    if (rotation % 90 != 0 || rotation == 0) {
```
stuartmorgan, line 891: Why is silently ignoring any rotation value that's not a multiple of 90 correct? If there are actually cases where this happens, wouldn't silently ignoring them be even more broken than the current behavior?

fa0311: @stuartmorgan  
The `rotation` value in this context can only be 0, 90, 180, or 270 degrees, as most video formats (such as MP4 and MOV) store rotation metadata in **90-degree increments** within the `tkhd` (Track Header) box.  

For example, in the current Flutter ExoPlayer backend, any value outside this range is ignored:  
https://github.com/flutter/packages/blob/2d3b24ebd9e8d72399d1176089065440e18cc033/packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/texture/TextureExoPlayerEventListener.java#L67-L74  

Since `RotatedBox` only supports quarter-turn rotations, this shouldn't be an issue. However, if an unexpected rotation value appears, should we fall back to `Transform.rotate` instead of ignoring it?

stuartmorgan: If the expectation is that it's impossible for this Dart code to receive values that aren't multiples of 90 degrees, that should be expressed via an assert, not a runtime conditional.

fa0311: I've added an assert to ensure that only multiples of 90 degrees are allowed.

fa0311: I've updated the conditional statement to reflect the assert change

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 2.9.4
 
 * Reduces the position update interval from 500ms to 100ms.
+* Fix layout issue caused by Transform.rotate not affecting space calculation.
```
stuartmorgan, line 4: Fixes

See the CHANGELOG style guide linked from the PR checklist.

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 2.9.4
```
stuartmorgan, line 1: This needs a new version bump; the previous one was lost in a merge.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -887,12 +886,18 @@ class _VideoPlayerWithRotation extends StatelessWidget {
...
-      : Transform.rotate(
-          angle: rotation * math.pi / 180,
-          child: child,
-        );
+  Widget build(BuildContext context) {
+    if (rotation == 0) {
+      return child;
+    }
+    if (rotation % 90 != 0) {
+      throw ArgumentError('Rotation must be a multiple of 90');
```
stuartmorgan, line 894: An argument error should be thrown at the point where it is passed as an argument, which is the constructor on line 884.

stuartmorgan (changes requested):

packages/video_player/video_player/CHANGELOG.md
fa0311: The issue has been fixed. A new version bump has been added.

packages/video_player/video_player/lib/video_player.dart
fa0311: The issue has been fixed.

packages/video_player/video_player/CHANGELOG.md
stuartmorgan: The package's actual version also needs to be bumped.

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8719859349019297185/+/u/Run_package_tests/CHANGELOG_and_version_validation/stdout

stuartmorgan (approved):
LGTM once the version is fixed.

tarrinneal (approved):
lgtm after pubspec version bump.

fa0311:
I have updated the version.

PR #8693, opened 2025-02-24
Manual roll Flutter from 911aa7547ed7 to 043b71954ce7

stuartmorgan:
https://github.com/flutter/packages/pull/8692 with fixes for indirect effects of `--explicit-package-dependencies`


.ci/flutter_master.version
reidbaker, line 1: Where can I read more about how this file works in the package repo

packages/flutter_plugin_android_lifecycle/example/android/app/src/main/java/io/flutter/plugins/flutter_plugin_android_lifecycle_example/MainActivity.java
```diff
@@ -18,9 +17,10 @@ public class MainActivity extends FlutterActivity {
   private static final String TAG = "MainActivity";
 
   @Override
-  public void configureFlutterEngine(FlutterEngine flutterEngine) {
+  public void configureFlutterEngine(@NonNull FlutterEngine flutterEngine) {
+    super.configureFlutterEngine(flutterEngine);
     flutterEngine.getPlugins().add(new TestPlugin());
-    flutterEngine.getPlugins().add(new IntegrationTestPlugin());
```
reidbaker, line 23: Why do we not need the integration test plugin? was it dead code?

packages/interactive_media_ads/android/build.gradle
```diff
@@ -54,7 +54,7 @@ android {
         testImplementation 'org.jetbrains.kotlin:kotlin-test'
         testImplementation "org.mockito.kotlin:mockito-kotlin:5.4.0"
         testImplementation 'org.mockito:mockito-inline:5.1.0'
-        testImplementation 'androidx.test:core:1.3.0'
```
reidbaker, line 57: FYI I filed https://github.com/flutter/flutter/issues/164025 to track making these updates easier.

reidbaker (approved):

tarrinneal (approved):

packages/flutter_plugin_android_lifecycle/example/android/app/src/main/java/io/flutter/plugins/flutter_plugin_android_lifecycle_example/MainActivity.java
stuartmorgan: This is a speculative change, but the fact that this wasn't calling `super` seems suspicious. Normally plugins are registered by the generated registrant.

This wasn't compiling because it's unconditionally using the integration test plugin, but that plugin is a dev dependency. If this doesn't work, I'll need to make this example app have a non-dev dependency on it instead, but that feels wrong.

.ci/flutter_master.version
stuartmorgan: I don't know that it's specifically documented anywhere, but its used in ci.yaml ([example](https://github.com/flutter/packages/blob/c44c2282aad10b3f30af83d2a87d4a3023c13882/.ci.yaml#L122)), and it's rolled by the auto-roller. The generic packages recipe uses that ci.yaml file to control the Flutter version that is checked out.

matanlurey:
> Hm, it looks like integration tests are failing pretty much across the board. @matanlurey Did you test with your changes to see if the instructions [here](https://docs.flutter.dev/testing/integration-tests#add-the-integration_test-dependency) and [here](https://docs.flutter.dev/testing/integration-tests#test-in-firebase-test-lab-android) are still correct? I'm wondering if having a dev dependency on `integration_test` is no longer sufficient for FTL tests.

Neither of those two steps should have been impacted. The scenario I could imagine breaking is if the result of `flutter build apk` (which defaults to `--release`) is what was being deployed.

To test that theory you could move the dependency from `dev_dependencies` to `dependencies`. 

Otherwise I would slap the config opt-out and assign me a P1, and I'll look at it tomorrow morning, and rollback the flag if there isn't a revert. Bonus points if there is a local repro.

stuartmorgan:
> The scenario I could imagine breaking is if the result of `flutter build apk` (which defaults to `--release`) is what was being deployed.

That's definitely not the case; this is what we are deploying:

https://github.com/flutter/packages/blob/c44c2282aad10b3f30af83d2a87d4a3023c13882/script/tool/lib/src/firebase_test_lab_command.dart#L285-L288

> To test that theory you could move the dependency from `dev_dependencies` to `dependencies`.

Running now with that change for the `camera*` packages to see what happens.

stuartmorgan:
That worked; the camera packages where I moved the `integration_test` dependency from `dev_dependencies` to `dependencies` are no longer failing.

matanlurey:
I checked out this PR, and ran:

```sh
$ cd packages/google_sign_in/google_sign_in_android/example
$ flutter test integration_test
00:00 +0: loading /Users/matanl/Developer/packages/packages/google_sign_in/google_sign_in_android/example/integration_test/google_sign_in_test.dart                                                                     R00:54 +0: loading /Users/matanl/Developer/packages/packages/google_sign_in/google_sign_in_android/example/integration_test/google_sign_in_test.dart                                                                54.4s
✓ Built build/app/outputs/flutter-apk/app-debug.apk
00:55 +0: loading /Users/matanl/Developer/packages/packages/google_sign_in/google_sign_in_android/example/integration_test/google_sign_in_test.dart                                                                     Installing build/app/outputs/flutter-apk/app-debug.apk...          534ms
00:58 +2: All tests passed!
```
I have to now assume compilation for FTL does something I don't understand. Looking into this further.

stuartmorgan:
> I checked out this PR, and ran:
> 
> ```shell
> $ cd packages/google_sign_in/google_sign_in_android/example
> $ flutter test integration_test
> ```

Yes, this path is working in CI. We have tests that do `flutter test` on emulator, and other tests that do all the FTL machinery to test on physical devices, and only the latter are failing.

matanlurey:
Cool, let me play around. This works, let me try running it on a device or emulator:
```sh
$ cd packages/google_sign_in/google_sign_in_android/example/android
$  ./gradlew app:assembleAndroidTest
$ ./gradlew app:assembleDebug -Ptarget=integration_test/google_sign_in_test.dart
```
Following this path:
- https://firebase.google.com/docs/test-lab/android/instrumentation-test >
- https://developer.android.com/studio/test/index.html >
- https://developer.android.com/studio/test/command-line

I just trying doing:
```sh
$ ./gradlew connectedDebugAndroidTest
```

The test launches on my emulator, but hangs and does not complete. Maybe it's the same problem?

matanlurey:
Looking at [`espresso/example`'s FTL logs](https://firebase.corp.google.com/project/flutter-infra-staging/testlab/histories/bh.368058afb9fac8d2/matrices/6684088934273081669/executions/bs.fb69f7885bd22e3e/logs):

```txt
02-24 17:28:22.859 20296 20296 I flutter : Warning: integration_test plugin was not detected.
02-24 17:28:22.859 20296 20296 I flutter : 
02-24 17:28:22.859 20296 20296 I flutter : If you're running the tests with `flutter drive`, please make sure your tests
02-24 17:28:22.859 20296 20296 I flutter : are in the `integration_test/` directory of your package and use
02-24 17:28:22.859 20296 20296 I flutter : `flutter test $path_to_test` to run it instead.
02-24 17:28:22.859 20296 20296 I flutter : 
02-24 17:28:22.859 20296 20296 I flutter : If you're running the tests with Android instrumentation or XCTest, this means
02-24 17:28:22.859 20296 20296 I flutter : that you are not capturing test results properly! See the following link for
02-24 17:28:22.859 20296 20296 I flutter : how to set up the integration_test plugin:
02-24 17:28:22.859 20296 20296 I flutter : 
02-24 17:28:22.859 20296 20296 I flutter : https://docs.flutter.dev/testing/integration-tests
02-24 17:28:22.859 20296 20296 I flutter : 
02-24 17:28:22.861 20296 20296 I flutter : 00:00 +2: All tests passed!
```

matanlurey:
OK, I think I found the problem, our documentation needs to be updated:

https://docs.flutter.dev/testing/integration-tests#upload-an-android-apk

... and here:
https://github.com/flutter/packages/blob/c44c2282aad10b3f30af83d2a87d4a3023c13882/script/tool/lib/src/firebase_test_lab_command.dart#L251-L259

This is implicitly passing `--release`, which in turn means it is stripping integration test.

@stuartmorgan Can you try changing this to pass `--debug` (or `--profile` if perf concerns)?

stuartmorgan:
> @stuartmorgan Can you try changing this to pass `--debug` (or `--profile` if perf concerns)?

Running that now. Where exactly will the difference show up? I wasn't sure what part of the project to look at to see the difference in what these commands output.

FYI (@reidbaker FYI as well since he implemented the Android version of `--config-only`), in generally `flutter`'s structure has been that once the project files have been generated by building once, developers are free to change their build mode later via native tooling. E.g., on iOS you can `flutter build` once, then open Xcode, and make Debug and Release builds without re-running `flutter build`. It sounds like that use case has a new failure mode? (In the past there were cases where that didn't work, but they were more niche than having a dev dependency on a plugin.)

reidbaker:
Fwiw my naive answer is that config only should not care what release mode you use. 

The actual build cares because it is the same as changing the running gradle tasks.

stuartmorgan:
It looks like re-running `flutter build --debug` did in fact work. I filed https://github.com/flutter/flutter/issues/164133 to discuss the fact that this is an unfortunate regression, which ideally we would fix.

stuartmorgan:
CHANGELOG override: minor change to an internal implementation detail of an example does not need a client-facing log entry.

PR #8708, opened 2025-02-25
[camera] Disable flaky tests

stuartmorgan:
Skips the `create does not call create on the Java side` tests due to excessive flake, while the issue is investigated.

https://github.com/flutter/flutter/issues/164132


camsim99 (approved):
SGTM

bparrishMines (approved):
LGTM

PR #8709, opened 2025-02-25
[pigeon] adds Internal options

tarrinneal:
Separates the user facing options from the internal use of options to avoid exposing settings we don't need the users to access. fixes https://github.com/flutter/flutter/issues/161634

Also adds option to prevent merging of pigeon definition file options. fixes https://github.com/flutter/flutter/issues/163159

Also removes injection of overflow members for testing from public facing api surface.


packages/pigeon/lib/src/cpp/cpp_generator.dart
```diff
@@ -93,15 +95,61 @@ class CppOptions {
...
+    required this.cppHeaderOut,
+    required this.cppSourceOut,
+    Iterable<String>? copyrightHeader,
+  })  : headerIncludePath =
+            options.headerIncludePath ?? path.basename(cppHeaderOut),
+        namespace = options.namespace,
+        copyrightHeader = options.copyrightHeader ?? copyrightHeader,
+        headerOutPath = options.headerOutPath;
+
+  /// The path to the header that will get placed in the source filed (example:
```
stuartmorgan, line 124: Typo: file

packages/pigeon/lib/src/objc/objc_generator.dart
```diff
@@ -91,15 +93,62 @@ class ObjcOptions {
...
+    Iterable<String>? copyrightHeader,
+  })  : headerIncludePath =
+            options.headerIncludePath ?? path.basename(objcHeaderOut),
+        prefix = options.prefix,
+        copyrightHeader = options.copyrightHeader ?? copyrightHeader,
+        fileSpecificClassNameComponent =
+            options.fileSpecificClassNameComponent ??
+                fileSpecificClassNameComponent;
+
+  /// The path to the header that will get placed in the source filed (example:
```
stuartmorgan, line 123: Typo: file

packages/pigeon/lib/src/pigeon_lib.dart
```diff
@@ -274,7 +275,6 @@ class PigeonOptions {
     this.gobjectOptions,
     this.dartOptions,
     this.copyrightHeader,
-    this.oneLanguage,
```
stuartmorgan, line 277: Doesn't this mean the PR needs to be a major version change, not a bugfix change? Unless I'm missing something, this whole file is exported, and the class is public.

packages/pigeon/lib/src/pigeon_lib.dart
```diff
@@ -449,13 +444,158 @@ class PigeonOptions {
...
-        'or add --dartPackageName={name_of_package} to your command line pigeon call.',
+        'or add --package_name={name_of_package} to your command line pigeon call.',
       );
     }
     return name;
   }
 }
 
+/// Options used when running the code generator.
+class InternalPigeonOptions {
```
stuartmorgan, line 455: Does this need to be in an exported file, or could it be moved to another file that's not fully exported for legacy reasons?

stuartmorgan: Hm, I guess it's used in public API below.

Although if we're doing a breaking change for library clients anyway, maybe we can rethink what's exported and what's not?

stuartmorgan (approved):
LGTM with minor comments.

melvinotieno:
Kindly requesting what the status of this PR is?

PR #2832, opened 2022-11-21
[ci] Continue .cirrus.yml alignment with flutter/plugins

stuartmorgan:
Adds various tasks/checks that are currently in flutter/plugins but not flutter/packages to .cirrus.yaml. Most of these are currently harmless no-ops since they skip any packages without parts relevant to the check. Two substantive changes:
- Enables Gradle dependabot to manage all of the example apps, as we have in flutter/plugins; this will help avoid the gradle versions in examples getting more and more outdated over time.
  - As part of this, removed the `android/` directory from the pigeon Obj-C e2e test directory, which was presumably just created by default and left there. (I expect to remove the whole e2e directory once the new Pigeon integration tests are in place, but this is the minimal change for now.)
- Moves Linux platform tests to the light workload group, which is where they are in flutter/plugins (since they don't actually need heavy workload machines).

Part of https://github.com/flutter/flutter/issues/113764

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [ ] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[...]

PR #8710, opened 2025-02-25
[various] Bump androidx.test:core to 1.4.0

stuartmorgan:
This is pulled out of https://github.com/flutter/packages/pull/8693 to land on its own while the other issues with that PR are investigated, since it's a useful change regardless.

This updates all packages in the repo to have `androidx.test:core` 1.4.0 if it's not already higher. This fixes a latent problem where older versions of `androidx.test.core` aren't compatible with targeting the current SDK, because it has activities that are not annotated as exported. This issue is currently being masked by the dependency on `integration_test`, which forced a higher version of that package, but `integration_test` will stop exporting that dependency onto the app as of recent master, dropping all the packages here back to a too-old `androidx.test:core`.

### DIFF
--- a/packages/quick_actions/quick_actions_android/example/android/app/build.gradle
+++ b/packages/quick_actions/quick_actions_android/example/android/app/build.gradle
@@ -22,7 +22,7 @@ if (flutterVersionName == null) {
     flutterVersionName = '1.0'
 }
 
-def androidXTestVersion = '1.2.0'
+def androidXTestVersion = '1.4.0'
 
 android {
     namespace 'io.flutter.plugins.quickactionsexample'
@@ -54,7 +54,6 @@ flutter {
 
 dependencies {
     testImplementation 'junit:junit:4.13.2'
-    androidTestImplementation 'androidx.test:runner:1.2.0'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
     api "androidx.test:core:$androidXTestVersion"
 
--- a/packages/file_selector/file_selector_android/android/build.gradle
+++ b/packages/file_selector/file_selector_android/android/build.gradle
@@ -38,7 +38,7 @@ android {
         implementation 'androidx.annotation:annotation:1.9.1'
         testImplementation 'junit:junit:4.13.2'
         testImplementation 'org.mockito:mockito-inline:5.1.0'
-        testImplementation 'androidx.test:core:1.3.0'
+        testImplementation 'androidx.test:core:1.4.0'
         testImplementation "org.robolectric:robolectric:4.12.1"
     }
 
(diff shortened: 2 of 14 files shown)


packages/quick_actions/quick_actions_android/example/android/app/build.gradle
```diff
@@ -54,7 +54,6 @@ flutter {
 
 dependencies {
     testImplementation 'junit:junit:4.13.2'
-    androidTestImplementation 'androidx.test:runner:1.2.0'
```
stuartmorgan, line 57: Deleted because there's a duplicate dependency 4 lines below, but using `androidXTestVersion`.

stuartmorgan:
test-exempt: rolling dependencies

reidbaker (approved):
FWIW 1.4 is from 2021 the latest stable is 1.6.1
https://developer.android.com/jetpack/androidx/releases/test#version_140_3

stuartmorgan:
> FWIW 1.4 is from 2021 the latest stable is 1.6.1 

Yes, I picked something conservative since it was part of the roller fix because I didn't want to get into conflict hell or cascading-changes hell.

Separate from this, maybe we should consider turning dependabot back on for this package? We are explicitly ignoring it right now, IIRC because the Android team felt that it wasn't valuable enough for the PR churn it caused. This latent issue the flip side of that decision. (/cc @jmagman who I know has been looking at our dependabot configs).

matanlurey (approved):

gmackall (approved):

tarrinneal (approved):

jmagman:
> Separate from this, maybe we should consider turning dependabot back on for this package?

It's only only for major versions, I can swap it to minor as well.
https://github.com/flutter/packages/blob/c44c2282aad10b3f30af83d2a87d4a3023c13882/.github/dependabot.yml#L129-L130

jmagman:
> IIRC because the Android team felt that it wasn't valuable enough for the PR churn it caused.

And it wasn't that intentional, I made all the test dependencies ignore minor and patch. Probably wiser to ignore just patches and then lock minor down on a case-by-case basis.

jmagman:
> And it wasn't that intentional, I made all the test dependencies ignore minor and patch.

Correction: it wasn't that intentional by me.
https://github.com/flutter/packages/pull/2832/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R120-R121

jmagman:
> And it wasn't that intentional, I made all the test dependencies ignore minor and patch. Probably wiser to ignore just patches and then lock minor down on a case-by-case basis.

https://github.com/flutter/packages/pull/8712

PR #8715, opened 2025-02-26
[ci] Lengthen custom tests timeout

stuartmorgan:
This test has been at the edge of its time limit for a while, and is now consistently timing out. Longer term we should see if we can shorten the Pigeon test time (it accounts for about 20 minutes), but for now, increase the timeout so that it doesn't fail.

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -341,7 +341,7 @@ targets:
 
   - name: Linux_android custom_package_tests master
     recipe: packages/packages
-    timeout: 30
+    timeout: 60
     dimensions:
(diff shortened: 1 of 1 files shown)


tarrinneal (approved):
Do you know why the pigeon tests are running so long?

stuartmorgan:
> Do you know why the pigeon tests are running so long?

Nope, that's why I did the other PR. I'm guessing that adding the overflow version of the tests a while ago put us close to the limit though, and something else made things slightly longer (maybe the web_benchmark tests).

PR #8716, opened 2025-02-26
[pigeon] Timestamp test steps in CI

stuartmorgan:
Pigeon's tests are taking a substantial amount of CI time, so it would be useful to be able to audit how long each step takes so we have more information if we need to shorten or shard these tests.


stuartmorgan:
test-exempt: is a test

tarrinneal (approved):
I guess just a lot of formatting and running tests multiple times

PR #8730, opened 2025-02-27
[tool] Move changed file detection to base command class

stuartmorgan:
Consolidates the code to find all changed file paths into the `PackageLoopingCommand` class that is the base of almost all of the repo tooling commands. This in a preparatory PR for a future change to allow each command to define a list of files or file patterns that definitively *don't* affect that test, so that CI can be smarter about what tests to run (e.g., not running expensive integration tests for README changes).

A side effect of this change is that tests of almost all commands now need a mock `GitDir` instance. This would add a lot of copy/pasted boilerplate to the test setup, and there is already too much of that, so instead this refactors common test setup:
- Creating a memory file system
- Populating it with a packages directory
- Creating a RecordingProcessRunner to mock out process calls
- Creating a mock GitDir that forwards to a RecordingProcessRunner

into a helper method (using records and destructuring to easily return multiple values). While some tests don't need all of these steps, those that don't can easily ignore parts of it, and it will make it much easier to update tests in the future if they need them, and it makes the setup much more consistent which makes it easier to reason about test setup in general.

Prep for https://github.com/flutter/flutter/issues/136394

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/common/git_version_finder.dart
+++ b/script/tool/lib/src/common/git_version_finder.dart
@@ -28,15 +28,6 @@ class GitVersionFinder {
   /// The base branche used to find a merge point if baseSha is not provided.
   final String _baseBranch;
 
-  static bool _isPubspec(String file) {
-    return file.trim().endsWith('pubspec.yaml');
-  }
-
-  /// Get a list of all the pubspec.yaml file that is changed.
-  Future<List<String>> getChangedPubSpecs() async {
-    return (await getChangedFiles()).where(_isPubspec).toList();
-  }
-
   /// Get a list of all the changed files.
   Future<List<String>> getChangedFiles(
       {bool includeUncommitted = false}) async {
--- a/script/tool/test/common/git_version_finder_test.dart
+++ b/script/tool/test/common/git_version_finder_test.dart
@@ -56,18 +56,6 @@ file2/file2.cc
     expect(changedFiles, equals(<String>['file1/file1.cc', 'file2/file2.cc']));
   });
 
-  test('get correct pubspec change based on git diff', () async {
-    gitDiffResponse = '''
-file1/pubspec.yaml
-file2/file2.cc
-''';
-    final GitVersionFinder finder =
-        GitVersionFinder(gitDir, baseSha: 'some base sha');
-    final List<String> changedFiles = await finder.getChangedPubSpecs();
-
-    expect(changedFiles, equals(<String>['file1/pubspec.yaml']));
-  });
-
   test('use correct base sha if not specified', () async {
     mergeBaseResponse = 'shaqwiueroaaidf12312jnadf123nd';
     gitDiffResponse = '''
--- a/script/tool/test/common/package_command_test.dart
+++ b/script/tool/test/common/package_command_test.dart
@@ -4,53 +4,46 @@
 
 import 'package:args/command_runner.dart';
 import 'package:file/file.dart';
-import 'package:file/memory.dart';
 import 'package:flutter_plugin_tools/src/common/core.dart';
 import 'package:flutter_plugin_tools/src/common/package_command.dart';
 import 'package:git/git.dart';
 import 'package:mockito/annotations.dart';
-import 'package:mockito/mockito.dart';
 import 'package:test/test.dart';
 
 import '../mocks.dart';
 import '../util.dart';
-import 'package_command_test.mocks.dart';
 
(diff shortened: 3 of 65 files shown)


script/tool/lib/src/common/git_version_finder.dart
```diff
@@ -28,15 +28,6 @@ class GitVersionFinder {
...
   final String _baseBranch;
 
-  static bool _isPubspec(String file) {
-    return file.trim().endsWith('pubspec.yaml');
-  }
-
-  /// Get a list of all the pubspec.yaml file that is changed.
-  Future<List<String>> getChangedPubSpecs() async {
-    return (await getChangedFiles()).where(_isPubspec).toList();
-  }
```
stuartmorgan, line 38: This code was only used in one place (and isn't really related to `git` so it's not clear why it was in this class), so I inlined it where it was used so that I could use the cached changed files list instead of recomputing it here.

script/tool/test/common/git_version_finder_test.dart
```diff
@@ -56,18 +56,6 @@ file2/file2.cc
...
-    gitDiffResponse = '''
-file1/pubspec.yaml
-file2/file2.cc
-''';
-    final GitVersionFinder finder =
-        GitVersionFinder(gitDir, baseSha: 'some base sha');
-    final List<String> changedFiles = await finder.getChangedPubSpecs();
-
-    expect(changedFiles, equals(<String>['file1/pubspec.yaml']));
-  });
```
stuartmorgan, line 69: This was the test for the now-inlined code. The existing command-level tests cover the functionality in that context.

script/tool/test/common/package_command_test.dart
```diff
@@ -4,53 +4,46 @@
...
   late RecordingProcessRunner processRunner;
+  late RecordingProcessRunner gitProcessRunner;
   late SamplePackageCommand command;
   late CommandRunner<void> runner;
-  late FileSystem fileSystem;
   late MockPlatform mockPlatform;
   late Directory packagesDir;
   late Directory thirdPartyPackagesDir;
 
+  SamplePackageCommand configureCommand({bool includeSubpackages = false}) {
```
stuartmorgan, line 27: I extracted a helper like this in a couple of tests. Some of our unit tests have the unfortunate pattern of setting up a usual command in `setUp`, but then in a few specific tests creating a custom command instance that overrides more things, ignoring the one in `setUp`, and using the local one.

I didn't want to try to fix all of those in this PR since this is already quite large, but when I hit those cases I did pull out a helper so that changes like "add `gitDir: gitDir` because now the command needs that mocked out" can at least be made in only one place in the file instead of having to look at why a subset of a command's tests are failing and then updating them with exactly the same change.

script/tool/test/util.dart
```diff
@@ -31,19 +32,11 @@ const String _defaultFlutterConstraint = '>=2.5.0';
 String getFlutterCommand(Platform platform) =>
     platform.isWindows ? 'flutter.bat' : 'flutter';
 
-/// Creates a packages directory in the given location.
-///
-/// If [parentDir] is set the packages directory will be created there,
-/// otherwise [fileSystem] must be provided and it will be created an arbitrary
-/// location in that filesystem.
-Directory createPackagesDirectory(
-    {Directory? parentDir, FileSystem? fileSystem}) {
```
stuartmorgan, line 40: We no longer needed the `parentDir` version, so I was able to simplify this method quite a bit.

stuartmorgan (commented):
You'll want to read the `test/util.dart` and `test/mocks.dart` changes before all the rest of the test changes since those are the foundation all the test changes are built on.

script/tool/test/util.dart
```diff
@@ -494,3 +487,45 @@ class ProcessCall {
...
+/// The returned GitDir instance forwards to a mock process runner, as described
+/// in [createForwardingMockGitDir]. This process runner is separate, so that
+/// tests can easily treat most git commands called as internal implementation
+/// details, and assert on the exact list of non-git commands that are run.
+({
+  Directory packagesDir,
+  RecordingProcessRunner processRunner,
+  RecordingProcessRunner gitProcessRunner,
+  GitDir gitDir,
+}) configureBaseCommandMocks({
```
tarrinneal, line 504: nice

tarrinneal (approved):
Much cleaner than before. I like it.

PR #8732, opened 2025-02-27
[camera] Skip another flaky Android test

stuartmorgan:
This test is showing the same flake as the recently-skipped "detached create does not call create on the Java side" tests.

See https://github.com/flutter/flutter/issues/164132

### DIFF
--- a/packages/camera/camera_android_camerax/test/recorder_test.dart
+++ b/packages/camera/camera_android_camerax/test/recorder_test.dart
@@ -71,7 +71,7 @@ void main() {
 
       verify(mockApi.create(argThat(isA<int>()), aspectRatio, bitRate,
           qualitySelectorIdentifier));
-    });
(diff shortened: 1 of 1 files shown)


stuartmorgan:
Hm, I just had to re-run a *different* test that flaked with the same `PlatformException(channel-error, Unable to establish connection on channel., null, null)`. Maybe this is a structural problem with the version of the proxy code it's currently using?

bparrishMines (approved):
LGTM

The method call fails asynchronously, so its probably possible that it depends on if the machine finishes all the tests before the failure happens.

stuartmorgan:
> The method call fails asynchronously, so its probably possible that it depends on if the machine finishes all the tests before the failure happens.

I may need to disable all of these tests then, because now the next test is failing all the time (maybe because it's a shorter test run)

stuartmorgan:
Closing as disabling individual tests doesn't work here.

PR #8734, opened 2025-02-27
[various] Updates packages to use flutter.compileSdkVersion

reidbaker:
- **Updates compileSdk 34 to flutter.compileSdkVersion.**

Related to flutter/flutter/issues/149836
Want to wait a couple of days after 
https://github.com/flutter/packages/pull/8700 lands before landing. 

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_android_camerax/pubspec.yaml
+++ b/packages/camera/camera_android_camerax/pubspec.yaml
@@ -2,7 +2,7 @@ name: camera_android_camerax
 description: Android implementation of the camera plugin using the CameraX library.
 repository: https://github.com/flutter/packages/tree/main/packages/camera/camera_android_camerax
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+camera%22
-version: 0.6.14
+version: 0.6.14+1
 
 environment:
   sdk: ^3.6.0
--- a/packages/camera/camera_android/CHANGELOG.md
+++ b/packages/camera/camera_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.10+1
+
+* Updates compileSdk 34 to flutter.compileSdkVersion.
+
 ## 0.10.10
 
 * Adds API support query for image streaming.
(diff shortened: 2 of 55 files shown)


packages/camera/camera_android_camerax/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: camera_android_camerax
 description: Android implementation of the camera plugin using the CameraX library.
 repository: https://github.com/flutter/packages/tree/main/packages/camera/camera_android_camerax
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+camera%22
-version: 0.6.14
+version: 0.6.14+1
```
reidbaker, line 5: You might wonder why this plugin didnt need to update to 3.27. It is because it was already updated.

reidbaker:
Ignoring conflicts in changelog and pubspec until after approval since they will keep happening.

reidbaker:
I think the failure that i am seeing is https://github.com/flutter/flutter/issues/164132

ryanheise:
Maybe the plugin template should also implement this change too since it currently hardcodes `compileSdk` to 35 instead of `flutter.compileSdkVersion`.

Also, just a question about what you've done here. I can see in the example's `android/app/build.gradle` file that there is this explicit flutter block:

```gradle
flutter {
    source '../..'
}
```

But in the plugin itself, the `android/build.gradle` file contains no such explicit flutter block, so I'm curious how the reference to `flutter.compileSdkVersion` is resolved. Is the idea that it is inherited from whatever app is being built? When I open the plugin in Android Studio, it has a problem resolving the reference to `flutter.compileSdkVersion`, yet apps will build fine from the command line.

Finally, I question whether this change might be considered as a breaking change, since other plugins or apps that depend on such a plugin that use `compileSdk` 34 might now stop during the build because the plugin requires `compileSdk` 35. Do we ignore that sort of thing when considering semantic versioning of plugins?

reidbaker:
"I question whether this change might be considered as a breaking change, since other plugins or apps that depend on such a plugin that use compileSdk 34 might now stop during the build because the plugin requires compileSdk 35. " 

This change also set the minimum flutter verison to `3.27.0` that version of flutter is expecting to compiled with sdk 35. If you were running a version of flutter that expected sdk 34 then you would not get this change. In my mind that is why this is not a breaking change.

PR #8761, opened 2025-02-28
[tool] Validate gradle compileSdk

stuartmorgan:
Checks that plugins use `compileSdk` rather than the deprecated `compileSdkVersion`, and that if they set it to
`flutter.compileSdkVersion` they require at least Flutter 3.27.

Would have prevented https://github.com/flutter/flutter/issues/164362

### DIFF
--- a/script/tool/lib/src/gradle_check_command.dart
+++ b/script/tool/lib/src/gradle_check_command.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'package:collection/collection.dart';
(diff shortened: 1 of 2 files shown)


stuartmorgan:
> Would have prevented [flutter/flutter#164362](https://github.com/flutter/flutter/issues/164362)

In addition to the unit tests, I ran this locally against the tree at the commit before https://github.com/flutter/packages/commit/bf5e9e7efb50e086c86ec17c9c06bbbc34e68d3e and verified that it did indeed fail on that plugin.

reidbaker (approved):

PR #8777, opened 2025-03-04
Update PR template for clarity

stuartmorgan:
Attempt to make the PR checklist more clear about expectations:
- Retitle the checklist to make it explicit that it's for review, not landing.
- Make it explicit that the version, CHANGELOG, and test steps require some action, rather than just personally believing that a PR should be exempt.


stuartmorgan:
I don't expect this to prevent PR submissions that don't follow the process of course, but hopefully it will cut down on legitimate confusion about what some of the steps mean, and reduce the number of comments we need to leave.

tarrinneal (approved):

PR #8787, opened 2025-03-04
[webview_flutter_wkwebview] Extended Web View API on iOS to add flexibility when working with local HTML content

agavrilko:
## Overview

The `readAccessURLProvider` property has been added to the `WebKitWebViewControllerCreationParams` class. This function allows customization of the path to associated resources when loading a local HTML page using `loadFile` on iOS.

### What Is the Issue

Fixes https://github.com/flutter/flutter/issues/136479

The native `WKWebView` takes two arguments when loading local HTML pages:

1. **The file URL** – The local HTML file to be loaded.
2. **The read access URL** – A file or directory that `WKWebView` can access (e.g., .css, .js files).

Currently, the Flutter implementation always sets the parent folder of the specified HTML file as the read access directory. This behavior is not configurable, which limits how local web content can be structured.

For example, the following structure does not work because the .css and .js files are not in the parent directory of the HTML file:
```
/app_resources/
│── styles/
│    ├── main.css
│
│── scripts/
│    ├── app.js
│
│── pages/
     ├── index.html  (Loaded file)
```
With the existing behavior, `WKWebView` cannot `access styles/main.css` and `scripts/app.js` because the parent folder of `index.html` (`/pages/`) does not contain them.

### How This Resolves the Issue

The `readAccessURLProvider` property has been added to the `WebKitWebViewControllerCreationParams` class. This property accepts a function that takes the path of the HTML file being loaded and returns the path to the associated resources.

Screenshot at Mar 04 14-49-46

Each time `loadFile` is called, the controller invokes this function and passes its result as the second argument to the underlying `WKWebView` implementation.

By default, the function returns the parent directory of the specified HTML file, preserving the existing behavior when no custom provider is set.

### Impact on End Users

1. Developers can now explicitly specify the directory that `WKWebView` should allow access to when loading a local HTML page, providing greater flexibility in organizing assets.
2. The existing API signature remains unchanged, ensuring that the update does not require any modifications to existing codebases unless the new functionality is utilized.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/webview_flutter/webview_flutter_android/CHANGELOG.md
+++ b/packages/webview_flutter/webview_flutter_android/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.9.0
+
+* Adds support for `PlatformWebViewController.loadFileWithParams`.
+* Introduces `AndroidLoadFileParams`, a platform-specific extension of `LoadFileParams` for Android that adds support for `headers`.
+
 ## 4.8.2
 
 * Bumps gradle from 8.9.0 to 8.11.1.
--- a/packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
+++ b/packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
@@ -18,6 +18,41 @@ import 'android_webkit_constants.dart';
 import 'platform_views_service_proxy.dart';
 import 'weak_reference_utils.dart';
 
+/// Object specifying parameters for loading a local file in a
+/// [AndroidWebViewController].
+@immutable
+base class AndroidLoadFileParams extends LoadFileParams {
+  /// Constructs a [AndroidLoadFileParams], the subclass of a [LoadFileParams].
+  AndroidLoadFileParams({
+    required String absoluteFilePath,
+    this.headers = const <String, String>{},
+  }) : super(
+          absoluteFilePath: absoluteFilePath.startsWith('file://')
+              ? absoluteFilePath
+              : Uri.file(absoluteFilePath).toString(),
+        );
+
+  /// Constructs a [AndroidLoadFileParams] using a [LoadFileParams].
+  factory AndroidLoadFileParams.fromLoadFileParams(
+    LoadFileParams params, {
+    Map<String, String> headers = const <String, String>{},
+  }) {
+    return AndroidLoadFileParams(
+      absoluteFilePath: params.absoluteFilePath,
+      headers: headers,
+    );
+  }
+
+  /// Additional HTTP headers to be included when loading the local file.
+  ///
+  /// If not provided at initialization time, doesn't add any additional headers.
+  ///
+  /// On Android, WebView supports adding headers when loading local or remote
+  /// content. This can be useful for scenarios like authentication,
+  /// content-type overrides, or custom request context.
+  final Map<String, String> headers;
+}
+
 /// Object specifying creation parameters for creating a [AndroidWebViewController].
 ///
 /// When adding additional fields make sure they can be null or have a default
@@ -386,12 +421,29 @@ class AndroidWebViewController extends PlatformWebViewController {
   Future<void> loadFile(
     String absoluteFilePath,
   ) {
-    final String url = absoluteFilePath.startsWith('file://')
-        ? absoluteFilePath
-        : Uri.file(absoluteFilePath).toString();
+    return loadFileWithParams(
+      AndroidLoadFileParams(
+        absoluteFilePath: absoluteFilePath,
+      ),
+    );
+  }
 
-    _webView.settings.setAllowFileAccess(true);
-    return _webView.loadUrl(url, <String, String>{});
+  @override
+  Future<void> loadFileWithParams(
+    LoadFileParams params,
+  ) async {
+    switch (params) {
+      case final AndroidLoadFileParams params:
+        await Future.wait(<Future<void>>[
+          _webView.settings.setAllowFileAccess(true),
+          _webView.loadUrl(params.absoluteFilePath, params.headers),
+        ]);
+
+      default:
+        await loadFileWithParams(
+          AndroidLoadFileParams.fromLoadFileParams(params),
+        );
+    }
   }
 
   @override
--- a/packages/webview_flutter/webview_flutter_android/test/android_webview_controller_test.dart
+++ b/packages/webview_flutter/webview_flutter_android/test/android_webview_controller_test.dart
@@ -274,7 +274,7 @@ void main() {
           ));
     }
 
-    test('loadFile without file prefix', () async {
+    test('Initializing WebView settings on controller creation', () async {
       final MockWebView mockWebView = MockWebView();
       final MockWebSettings mockWebSettings = MockWebSettings();
       createControllerWithMocks(
@@ -292,56 +292,209 @@ void main() {
       verify(mockWebSettings.setUseWideViewPort(false)).called(1);
     });
 
-    test('loadFile without file prefix', () async {
-      final MockWebView mockWebView = MockWebView();
-      final MockWebSettings mockWebSettings = MockWebSettings();
-      final AndroidWebViewController controller = createControllerWithMocks(
-        mockWebView: mockWebView,
-        mockSettings: mockWebSettings,
-      );
+    group('loadFile', () {
+      test('Without file prefix', () async {
+        final MockWebView mockWebView = MockWebView();
+        final MockWebSettings mockWebSettings = MockWebSettings();
+        final AndroidWebViewController controller = createControllerWithMocks(
+          mockWebView: mockWebView,
+          mockSettings: mockWebSettings,
+        );
 
-      await controller.loadFile('/path/to/file.html');
+        await controller.loadFile('/path/to/file.html');
 
-      verify(mockWebSettings.setAllowFileAccess(true)).called(1);
-      verify(mockWebView.loadUrl(
-        'file:///path/to/file.html',
-        <String, String>{},
-      )).called(1);
+        verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+        verify(mockWebView.loadUrl(
+          'file:///path/to/file.html',
+          <String, String>{},
+        )).called(1);
+      });
+
+      test('Without file prefix and characters to be escaped', () async {
+        final MockWebView mockWebView = MockWebView();
+        final MockWebSettings mockWebSettings = MockWebSettings();
+        final AndroidWebViewController controller = createControllerWithMocks(
+          mockWebView: mockWebView,
+          mockSettings: mockWebSettings,
+        );
+
+        await controller.loadFile('/path/to/?_<_>_.html');
+
+        verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+        verify(mockWebView.loadUrl(
+          'file:///path/to/%3F_%3C_%3E_.html',
+          <String, String>{},
+        )).called(1);
+      });
+
+      test('With file prefix', () async {
+        final MockWebView mockWebView = MockWebView();
+        final MockWebSettings mockWebSettings = MockWebSettings();
+        final AndroidWebViewController controller = createControllerWithMocks(
+          mockWebView: mockWebView,
+        );
+
+        when(mockWebView.settings).thenReturn(mockWebSettings);
+
+        await controller.loadFile('file:///path/to/file.html');
+
+        verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+        verify(mockWebView.loadUrl(
+          'file:///path/to/file.html',
+          <String, String>{},
+        )).called(1);
+      });
     });
 
-    test('loadFile without file prefix and characters to be escaped', () async {
-      final MockWebView mockWebView = MockWebView();
-      final MockWebSettings mockWebSettings = MockWebSettings();
-      final AndroidWebViewController controller = createControllerWithMocks(
-        mockWebView: mockWebView,
-        mockSettings: mockWebSettings,
-      );
+    group('loadFileWithParams', () {
+      group('Using LoadFileParams model', () {
+        test('Without file prefix', () async {
+          final MockWebView mockWebView = MockWebView();
+          final MockWebSettings mockWebSettings = MockWebSettings();
+          final AndroidWebViewController controller = createControllerWithMocks(
+            mockWebView: mockWebView,
+            mockSettings: mockWebSettings,
+          );
 
-      await controller.loadFile('/path/to/?_<_>_.html');
+          await controller.loadFileWithParams(
+            const LoadFileParams(absoluteFilePath: '/path/to/file.html'),
+          );
 
-      verify(mockWebSettings.setAllowFileAccess(true)).called(1);
-      verify(mockWebView.loadUrl(
-        'file:///path/to/%3F_%3C_%3E_.html',
-        <String, String>{},
-      )).called(1);
-    });
+          verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+          verify(mockWebView.loadUrl(
+            'file:///path/to/file.html',
+            <String, String>{},
+          )).called(1);
+        });
 
-    test('loadFile with file prefix', () async {
-      final MockWebView mockWebView = MockWebView();
-      final MockWebSettings mockWebSettings = MockWebSettings();
-      final AndroidWebViewController controller = createControllerWithMocks(
-        mockWebView: mockWebView,
-      );
+        test('Without file prefix and characters to be escaped', () async {
+          final MockWebView mockWebView = MockWebView();
+          final MockWebSettings mockWebSettings = MockWebSettings();
+          final AndroidWebViewController controller = createControllerWithMocks(
+            mockWebView: mockWebView,
+            mockSettings: mockWebSettings,
+          );
 
-      when(mockWebView.settings).thenReturn(mockWebSettings);
+          await controller.loadFileWithParams(
+            const LoadFileParams(absoluteFilePath: '/path/to/?_<_>_.html'),
+          );
 
-      await controller.loadFile('file:///path/to/file.html');
+          verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+          verify(mockWebView.loadUrl(
+            'file:///path/to/%3F_%3C_%3E_.html',
+            <String, String>{},
+          )).called(1);
+        });
 
-      verify(mockWebSettings.setAllowFileAccess(true)).called(1);
-      verify(mockWebView.loadUrl(
-        'file:///path/to/file.html',
-        <String, String>{},
-      )).called(1);
+        test('With file prefix', () async {
+          final MockWebView mockWebView = MockWebView();
+          final MockWebSettings mockWebSettings = MockWebSettings();
+          final AndroidWebViewController controller = createControllerWithMocks(
+            mockWebView: mockWebView,
+            mockSettings: mockWebSettings,
+          );
+
+          await controller.loadFileWithParams(
+            const LoadFileParams(absoluteFilePath: 'file:///path/to/file.html'),
+          );
+
+          verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+          verify(mockWebView.loadUrl(
+            'file:///path/to/file.html',
+            <String, String>{},
+          )).called(1);
+        });
+      });
+
+      group('Using WebKitLoadFileParams model', () {
+        test('Without file prefix', () async {
+          final MockWebView mockWebView = MockWebView();
+          final MockWebSettings mockWebSettings = MockWebSettings();
+          final AndroidWebViewController controller = createControllerWithMocks(
+            mockWebView: mockWebView,
+            mockSettings: mockWebSettings,
+          );
+
+          await controller.loadFileWithParams(
+            AndroidLoadFileParams(absoluteFilePath: '/path/to/file.html'),
+          );
+
+          verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+          verify(mockWebView.loadUrl(
+            'file:///path/to/file.html',
+            <String, String>{},
+          )).called(1);
+        });
+
+        test('Without file prefix and characters to be escaped', () async {
+          final MockWebView mockWebView = MockWebView();
+          final MockWebSettings mockWebSettings = MockWebSettings();
+          final AndroidWebViewController controller = createControllerWithMocks(
+            mockWebView: mockWebView,
+            mockSettings: mockWebSettings,
+          );
+
+          await controller.loadFileWithParams(
+            AndroidLoadFileParams(absoluteFilePath: '/path/to/?_<_>_.html'),
+          );
+
+          verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+          verify(mockWebView.loadUrl(
+            'file:///path/to/%3F_%3C_%3E_.html',
+            <String, String>{},
+          )).called(1);
+        });
+
+        test('With file prefix', () async {
+          final MockWebView mockWebView = MockWebView();
+          final MockWebSettings mockWebSettings = MockWebSettings();
+          final AndroidWebViewController controller = createControllerWithMocks(
+            mockWebView: mockWebView,
+            mockSettings: mockWebSettings,
+          );
+
+          await controller.loadFileWithParams(
+            AndroidLoadFileParams(
+                absoluteFilePath: 'file:///path/to/file.html'),
+          );
+
+          verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+          verify(mockWebView.loadUrl(
+            'file:///path/to/file.html',
+            <String, String>{},
+          )).called(1);
+        });
+
+        test('With additional headers', () async {
+          final MockWebView mockWebView = MockWebView();
+          final MockWebSettings mockWebSettings = MockWebSettings();
+          final AndroidWebViewController controller = createControllerWithMocks(
+            mockWebView: mockWebView,
+            mockSettings: mockWebSettings,
+          );
+
+          await controller.loadFileWithParams(
+            AndroidLoadFileParams(
+              absoluteFilePath: 'file:///path/to/file.html',
+              headers: const <String, String>{
+                'Authorization': 'Bearer test_token',
+                'Cache-Control': 'no-cache',
+                'X-Custom-Header': 'test-value',
+              },
+            ),
+          );
+
+          verify(mockWebSettings.setAllowFileAccess(true)).called(1);
+          verify(mockWebView.loadUrl(
+            'file:///path/to/file.html',
+            const <String, String>{
+              'Authorization': 'Bearer test_token',
+              'Cache-Control': 'no-cache',
+              'X-Custom-Header': 'test-value',
+            },
+          )).called(1);
+        });
+      });
     });
 
     test('loadFlutterAsset when asset does not exist', () async {
--- a/packages/webview_flutter/webview_flutter_wkwebview/CHANGELOG.md
+++ b/packages/webview_flutter/webview_flutter_wkwebview/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.23.0
+
+* Adds support for `PlatformWebViewController.loadFileWithParams`.
+* Introduces `WebKitLoadFileParams`, a platform-specific extension of `LoadFileParams` for iOS and macOS that adds support for `readAccessPath`.
+
 ## 3.22.1
 
 * Changes the handling of a Flutter method failure from throwing an assertion error to logging the
--- a/packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
+++ b/packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
@@ -38,6 +38,39 @@ enum PlaybackMediaTypes {
   }
 }
 
+/// Object specifying parameters for loading a local file in a
+/// [WebKitWebViewController].
+@immutable
+base class WebKitLoadFileParams extends LoadFileParams {
+  /// Constructs a [WebKitLoadFileParams], the subclass of a [LoadFileParams].
+  WebKitLoadFileParams({
+    required super.absoluteFilePath,
+    String? readAccessPath,
+  })  : readAccessPath = readAccessPath ?? path.dirname(absoluteFilePath),
+        super();
+
+  /// Constructs a [WebKitLoadFileParams] using a [LoadFileParams].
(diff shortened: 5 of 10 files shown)


bparrishMines:
@agavrilko To follow the pattern of other methods, we should introduce a new platform interface method to `webview_flutter_platform_interface`. Something like `PlatformWebViewController.loadFileWithParams(LoadFileParams)`. `LoadFileParams` would contain a `url` parameter. Then the `webview_flutter_wkwebview` package can create its own implementation of `LoadFileParams` with the `readAccessUrl` parameter because it is unique to iOS/macOS.

bparrishMines (changes requested):
@agavrilko To follow the pattern of other methods, we should introduce a new platform interface method to `webview_flutter_platform_interface`. Something like `PlatformWebViewController.loadFileWithParams(LoadFileParams)`. `LoadFileParams` would contain a `url` parameter. Then the `webview_flutter_wkwebview` package can create its own implementation of `LoadFileParams` with the `readAccessUrl` parameter because it is unique to iOS/macOS.

agavrilko:
Hello @bparrishMines,
Thank you for taking the time to review this MR!
I would like to ensure that I understand your feedback before proceeding.

I initially aimed to introduce this minor update without making any API changes, focusing solely on exposing the functionality that was already present but not accessible to users.
However, based on your comment, it seems you are suggesting a broader update that would introduce new APIs (e.g., `LoadFileParams` and a new/updated platform interface method).

Based on my understanding, your suggestion would require:
1. Introducing `LoadFileParams` at the platform interface level.
2. Extending the `PlatformWebViewController` to accept `LoadFileParams` instead of a plain string.
3. Adapting `webview_flutter_wkwebview` with a `WebKitLoadFileParams` subclass.

webview_design_update

While this design would make the API more consistent and extensible, it would also introduce a breaking change for current users, requiring them to update their code.
That is why I am in doubt.

And if my understanding is correct, should I just update an existing `loadFile` method or introduce a new one?

bparrishMines:
@agavrilko I think you should create a new platform interface method named something like: `PlatformWebViewController.loadFileWithParams(LoadFileParams)`. This is consistent with updates to other platform interfaces where we realized in hindsight the method should have taken a class so that we can add parameters later.

Everything else you stated seems correct.

Additionally, after this feature lands, `PlatformWebViewController.loadFile` can be deprecated and it can just call `loadFileWithParams`.

agavrilko:
Hello @bparrishMines

I'm seeing that the pipeline fails due to a version resolution issue:
> Because webview_flutter_example depends on webview_flutter_platform_interface ^2.12.0 which doesn't match any versions, version solving failed.

Could you please advise on the proper workflow for updating versions across the packages?
Also, when you have a moment, could you review the changes to ensure everything aligns with your expectations?

Thank you!

bparrishMines:
@agavrilko Yes, you can follow https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins for the process.

agavrilko:
Hello @bparrishMines,

I need further assistance with this PR.
I expected to see `...you can't publish a package that has dependency overrides...`, as stated [here](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins).

But it looks like changing `_platform_interface` and implementations in the same PR is not allowed:
> Changed packages: webview_flutter/webview_flutter,webview_flutter_android,webview_flutter_platform_interface,webview_flutter_wkwebview
> [0:00] Running for packages/webview_flutter/webview_flutter...
> Dart changes are not allowed to other packages in webview_flutter in the same PR as changes to public Dart code in webview_flutter_platform_interface, as this can cause accidental breaking changes to be missed by automated checks. Please split the changes to these two packages into separate PRs.

Should this change be split into 2 different PRs?
If yes, please specify what parts should go to each.

packages/webview_flutter/webview_flutter/lib/src/webview_controller.dart
```diff
@@ -131,10 +131,25 @@ class WebViewController {
...
+  /// Loads a local HTML file using the provided [params].
+  ///
+  /// The [params] object should contain the absolute path to the
+  /// file as it is stored on the device. For example:
+  /// `/Users/username/Documents/www/index.html`.
+  /// In addition, it may include platform-specific fields,
+  /// such as headers (on Android) or resource paths (on iOS/macOS).
+  ///
+  /// Throws a `PlatformException` if the [absoluteFilePath] does not exist.
+  Future<void> loadFileWithParams(LoadFileParams params) {
```
bparrishMines: This method shouldn't be necessary. A user can now use the platform specific `loadFileWithParams` to add platform specific parameters.

packages/webview_flutter/webview_flutter/lib/src/webview_controller.dart
```diff
@@ -131,10 +131,25 @@ class WebViewController {
   /// `/Users/username/Documents/www/index.html`.
   ///
   /// Throws a `PlatformException` if the [absoluteFilePath] does not exist.
+  @Deprecated('Use loadFileWithParams(LocalFileParams params) instead. '
```
bparrishMines: This shouldn't need deprecation. This can just change the `platform` call from `platform.loadFile` to `platform.loadFileWithParams`.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/load_file_params.dart
```diff
@@ -0,0 +1,40 @@
...
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+
+/// Object specifying parameters for loading a local HTML file into a web view.
+///
+/// Platform-specific implementations can add additional fields by extending
+/// this class.
+///
+/// {@tool sample}
```
bparrishMines:
```suggestion
```

This doesn't actually do anything right now. We just haven't removed them from other classes yet.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/load_file_params.dart
```diff
@@ -0,0 +1,40 @@
...
+///   const WebKitLoadFileParams({
+///     required super.absoluteFilePath,
+///     required String readAccessPath,
+///   });
+///
+///   /// The directory to which the WebView is granted read access.
+///   final String readAccessPath;
+/// }
+/// ```
+/// {@end-tool}
```
bparrishMines:
```suggestion
```

Same here

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/load_file_params.dart
```diff
@@ -0,0 +1,40 @@
...
+
+import 'package:flutter/material.dart';
+
+/// Object specifying parameters for loading a local HTML file into a web view.
+///
+/// Platform-specific implementations can add additional fields by extending
+/// this class.
+///
+/// {@tool sample}
+/// This example demonstrates how to extend [LocalFileParams] to provide
```
bparrishMines:
```suggestion
/// This example demonstrates how to extend [LoadFileParams] to provide
```

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/load_file_params.dart
```diff
@@ -0,0 +1,40 @@
...
+/// additional platform-specific parameters.
+///
+/// When extending [LocalFileParams], additional parameters should always accept
```
bparrishMines:
```suggestion
/// When extending [LoadFileParams], additional parameters should always accept
```

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
```diff
@@ -404,11 +438,18 @@ class WebKitWebViewController extends PlatformWebViewController {
   }
 
   @override
-  Future<void> loadFile(String absoluteFilePath) {
```
bparrishMines, line 407: To prevent a breaking change. The `loadFile` method should still be implemented and it can just call `loadFileWithParams`.

packages/webview_flutter/webview_flutter_wkwebview/test/webkit_webview_controller_test.dart
```diff
@@ -330,15 +330,38 @@ void main() {
       });
     });
 
-    test('loadFile', () async {
```
bparrishMines: We should keep the `loadFile` test and create a new test for `loadFileWithParams`.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
```diff
@@ -38,6 +38,40 @@ enum PlaybackMediaTypes {
   }
 }
 
+/// Object specifying parameters for loading a local file in a
+/// [WebKitWebViewController].
+@immutable
+class WebKitLoadFileParams extends LoadFileParams {
+  /// Constructs a [WebKitLoadFileParams], the subclass of a [LoadFileParams].
+  ///
+  /// The optional [readAccessPath] defines the directory that the WebView is
```
bparrishMines: I think this information should be moved to the documentation for the field.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
```diff
@@ -18,6 +18,48 @@ import 'android_webkit_constants.dart';
 import 'platform_views_service_proxy.dart';
 import 'weak_reference_utils.dart';
 
+/// Object specifying parameters for loading a local file in a
+/// [AndroidWebViewController].
+@immutable
+class AndroidLoadFileParams extends LoadFileParams {
+  /// Constructs a [AndroidLoadFileParams], the subclass of a [LoadFileParams].
+  ///
+  /// Optionally, [headers] map may be included, which contains key-value pairs
```
bparrishMines: I think this information should be moved to the field.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
```diff
@@ -18,6 +18,48 @@ import 'android_webkit_constants.dart';
...
+  /// that will be passed as additional HTTP headers when loading the file.
+  AndroidLoadFileParams({
+    required String absoluteFilePath,
+    this.headers = _defaultHeaders,
```
bparrishMines: This should work right?

```suggestion
    this.headers = const <String, String>{},
```

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
```diff
@@ -383,15 +425,19 @@ class AndroidWebViewController extends PlatformWebViewController {
       _webView.pigeon_instanceManager.getIdentifier(_webView)!;
 
   @override
-  Future<void> loadFile(
```
bparrishMines, line 386: As stated in the iOS implementation as well, we keep `loadFile` and it should just call `loadFileWithParams`.

packages/webview_flutter/webview_flutter_android/test/android_webview_controller_test.dart
```diff
@@ -274,7 +274,7 @@ void main() {
           ));
     }
 
-    test('loadFile without file prefix', () async {
+    test('Initializing WebView settings on controller creation', () async {
```
bparrishMines, line 277: Since the `loadFile` method will be kept in `AndroidWebViewController`, one of the tests for `loadFile` should be kept.

bparrishMines (commented):
@agavrilko Yes, this PR won't be published as is. The first step is to get this PR approved first, then individual PRs for each package will be created. This allows us to review all the changes together before we start landing seaprate PRs.

packages/webview_flutter/webview_flutter/lib/src/webview_controller.dart
agavrilko: I am not sure I am property understand this comment.
Do you suggest to remove this method from the `WebViewController` class?
The loading of the file with custom parameters would look like the following:

>  switch (controller.platform) {
>    case final WebKitWebViewController platform:
>      platform.loadFileWithParams(...)
>
>    default:
>      ...
>  }

Just want to ensure I understand it right.

packages/webview_flutter/webview_flutter/lib/src/webview_controller.dart
agavrilko: Updated.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/load_file_params.dart
agavrilko: Removed.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/load_file_params.dart
agavrilko: Removed.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/load_file_params.dart
agavrilko: Fixed.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/load_file_params.dart
agavrilko: Fixed.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
agavrilko: Updated.

packages/webview_flutter/webview_flutter_wkwebview/test/webkit_webview_controller_test.dart
agavrilko: 'loadFile' test restored.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
agavrilko: Moved.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
agavrilko: Moved.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
agavrilko: Updated.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
agavrilko: Updated.

packages/webview_flutter/webview_flutter_android/test/android_webview_controller_test.dart
agavrilko: 'loadFile' tests restored under 'loadFile' group.
NOTE: the test on 277 line was not related to the 'loadFile' method, this change just fixes the label.

packages/webview_flutter/webview_flutter/lib/src/webview_controller.dart
bparrishMines: Correct. `WebViewController.loadFileWithParams` should be removed and users could use the code snippet you provided to access the platform-specific params. The goal of the app-facing package is to provide a simple and clean API while the platform interface provides a more robust and extendable API.

bparrishMines:
@agavrilko Is this waiting for another review? `WebViewController.loadFileWithParams` is still included in `webview_flutter`.

packages/webview_flutter/webview_flutter/lib/src/webview_controller.dart
agavrilko: The method was removed.
All changes to the `WebViewController` were reverted.

agavrilko:
> @agavrilko Is this waiting for another review? `WebViewController.loadFileWithParams` is still included in `webview_flutter`.

Hello @bparrishMines, my apologies, I did not see the notification about your reply.
The method was removed as requested.

I have rebased the branch onto the latest `main` and cleaned the rest of the code.
Let me know what I should do next.
Thank you for your assistance.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/platform_webview_controller.dart
```diff
@@ -61,6 +61,24 @@ abstract class PlatformWebViewController extends PlatformInterface {
...
+  /// Loads a local HTML file using the provided [params].
+  ///
+  /// The [params.absoluteFilePath] should contain the absolute path to the file
+  /// on the device.
+  /// For example: `/Users/username/Documents/www/index.html`.
+  ///
+  /// Platform-specific implementations may extend [LocalFileParams] to support
+  /// additional parameters, such as iOS/macOS-specific read access options.
+  ///
+  /// Throws an [ArgumentError] if the file specified in
```
bparrishMines: nit: I don't think this is true anymore for this method. I see that this was copied from `loadFile`, but this sentence can be removed for this method.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
```diff
@@ -18,6 +18,40 @@ import 'android_webkit_constants.dart';
...
+    LoadFileParams params, {
+    Map<String, String> headers = const <String, String>{},
+  }) {
+    return AndroidLoadFileParams(
+      absoluteFilePath: params.absoluteFilePath,
+      headers: headers,
+    );
+  }
+
+  /// Additional HTTP headers to be included when loading the local file.
```
bparrishMines, line 46: nit: Separate first sentence.
```suggestion
  /// Additional HTTP headers to be included when loading the local file.
  ///
```

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
```diff
@@ -386,12 +420,30 @@ class AndroidWebViewController extends PlatformWebViewController {
...
+    );
+  }
+
+  @override
+  Future<void> loadFileWithParams(
+    LoadFileParams params,
+  ) async {
+    switch (params) {
+      case final AndroidLoadFileParams params:
+        await _webView.settings.setAllowFileAccess(true);
```
bparrishMines: nit: Using `Future.wait` here is a little faster. Method calls are called sequentially and are handled synchronously.
```suggestion
        await Future.wait(<Future<void>>[
          _webView.settings.setAllowFileAccess(true),
          _webView.loadUrl(params.absoluteFilePath, params.headers),
        ]);
```

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
```diff
@@ -38,6 +38,38 @@ enum PlaybackMediaTypes {
...
+    LoadFileParams params, {
+    String? readAccessPath,
+  }) {
+    return WebKitLoadFileParams(
+      absoluteFilePath: params.absoluteFilePath,
+      readAccessPath: readAccessPath,
+    );
+  }
+
+  /// The directory to which the WebView is granted read access.
```
bparrishMines, line 63: nit: Separate first sentence.
```suggestion
  /// The directory to which the WebView is granted read access.
  ///
```

packages/webview_flutter/webview_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.8.0
+
+* Introduces `AndroidLoadFileParams`, a platform-specific extension of `LoadFileParams` for Android that adds support for `headers`.
```
bparrishMines, line 4: I would also add something like this 
```suggestion
* Adds support for `PlatformWebViewController.loadFileWithParams`.
* Introduces `AndroidLoadFileParams`, a platform-specific extension of `LoadFileParams` for Android that adds support for `headers`.
```

packages/webview_flutter/webview_flutter_wkwebview/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## NEXT
+## 3.23.0
 
 * Updates minimum supported SDK version to Flutter 3.27/Dart 3.6.
+* Introduces `WebKitLoadFileParams`, a platform-specific extension of `LoadFileParams` for iOS and macOS that adds support for `readAccessPath`.
```
bparrishMines, line 4: Same comment from Android.
```suggestion
* Adds support for `PlatformWebViewController.loadFileWithParams`.
* Introduces `WebKitLoadFileParams`, a platform-specific extension of `LoadFileParams` for iOS and macOS that adds support for `readAccessPath`.
```

packages/webview_flutter/webview_flutter_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.14.0
+
+* Adds `PlatformWebViewController.loadFileWithParams(LoadFileParams)` to support loading local HTML files with platform-specific parameters.
+* Deprecates `PlatformWebViewController.loadFile(String)` in favour of `PlatformWebViewController.loadFileWithParams(LoadFileParams)`.
```
bparrishMines: This can be removed for now since we won't deprecate it until after the platform implementations have landed.

stuartmorgan-g: We can do deprecations in the initial pass now. That used to be impossible due to linting, but we no longer have deprecation lints enabled in the repo.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/platform_webview_controller.dart
agavrilko: Removed from `loadFileWithParams` only.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
agavrilko: Updated.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
agavrilko: Updated.
I'm not familiar with the internal implementation, but these operations seem closely related. Could this introduce any potential race conditions? Setting the access flag after the page starts loading doesn't seem reliable to me.

packages/webview_flutter/webview_flutter_wkwebview/lib/src/webkit_webview_controller.dart
agavrilko: Updated.

packages/webview_flutter/webview_flutter_android/CHANGELOG.md
agavrilko: Added.

packages/webview_flutter/webview_flutter_wkwebview/CHANGELOG.md
agavrilko: Added.

packages/webview_flutter/webview_flutter_platform_interface/CHANGELOG.md
agavrilko: Removed from the change log, since we removed it from the code.

packages/webview_flutter/webview_flutter/CHANGELOG.md
stuartmorgan-g, line 1: Everything in this package can be reverted; there are no changes at this layer.

stuartmorgan-g (changes requested):
@bparrishMines Longer term we probably want to consider plumbing a platform-subclassable load options class to the main API; having to call the platform interface's load method instead of just passing a platform subclass to an app-facing API method is somewhat awkward, and we generally don't want clients to be in the habit of directly using the platform interface itself.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/platform_webview_controller.dart
```diff
@@ -61,6 +61,21 @@ abstract class PlatformWebViewController extends PlatformInterface {
...
+  /// The [params.absoluteFilePath] should contain the absolute path to the file
+  /// on the device.
+  /// For example: `/Users/username/Documents/www/index.html`.
+  ///
+  /// Platform-specific implementations may extend [LocalFileParams] to support
+  /// additional parameters, such as iOS/macOS-specific read access options.
+  Future<void> loadFileWithParams(
+    LoadFileParams params,
+  ) {
+    throw UnimplementedError(
```
stuartmorgan-g: This should call the old method, not throw. See step 1 of https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-platform-interface-method-parameters

stuartmorgan-g: And that pass-through should be unit-tested in this package.

packages/webview_flutter/webview_flutter/CHANGELOG.md
agavrilko: Changes on `webview_flutter` were reverted.

packages/webview_flutter/webview_flutter_platform_interface/lib/src/platform_webview_controller.dart
agavrilko: `loadFileWithParams` implementation was updated. Unit test added.

stuartmorgan-g (approved):
Looks good to go forward with splitting out the platform interface PR.

agavrilko:
Hello @bparrishMines, @stuartmorgan-g

Here is the PR with only platform interface changes.
[https://github.com/flutter/packages/pull/9697](https://github.com/flutter/packages/pull/9697)

Let me know if I applied the version bump correctly.

agavrilko:
Hello @bparrishMines, @stuartmorgan-g 
I updated PR after the platform interface was merged.

All tests and formatting passes just fine locally,
However there is an issue on pipeline, which I am afraid I do not understand:
image

> error - lib/src/webkit_webview_controller.dart:44:41 - Classes can only extend other classes. Try specifying a different superclass, or removing the extends clause. - extends_non_class
>   error - lib/src/webkit_webview_controller.dart:47:20 - No associated named super constructor parameter. Try changing the name to the name of an existing named super constructor parameter, or creating such named parameter. - super_formal_parameter_without_associated_named
>   error - lib/src/webkit_webview_controller.dart:49:57 - The argument type 'dynamic' can't be assigned to the parameter type 'String'.  - argument_type_not_assignable
>   error - lib/src/webkit_webview_controller.dart:54:5 - Undefined class 'LoadFileParams'. Try changing the name to the name of an existing class, or creating a class with the name 'LoadFileParams'. - undefined_class
>   error - lib/src/webkit_webview_controller.dart:448:5 - Undefined class 'LoadFileParams'. Try changing the name to the name of an existing class, or creating a class with the name 'LoadFileParams'. - undefined_class
>   error - lib/src/webkit_webview_controller.dart:453:18 - The getter 'absoluteFilePath' isn't defined for the type 'WebKitLoadFileParams'. Try importing the library that defines 'absoluteFilePath', correcting the name to the name of an existing getter, or defining a getter or field named 'absoluteFilePath'. - undefined_getter
>   error - test/webkit_webview_controller_test.dart:353:17 - The name 'LoadFileParams' isn't a class. Try correcting the name to match an existing class. - creation_with_non_type
> warning - lib/src/webkit_webview_controller.dart:447:16 - The method doesn't override an inherited method. Try updating this class to match the superclass, or removing the override annotation. - override_on_non_overriding_member
> warning - lib/src/webkit_webview_controller.dart:457:7 - Dead code. Try removing the code, or fixing the code before it so that it can be reached. - dead_code
> warning - lib/src/webkit_webview_controller.dart:458:9 - Dead code. Try removing the code, or fixing the code before it so that it can be reached. - dead_code
>    info - lib/src/webkit_webview_controller.dart:447:16 - Missing documentation for a public member. Try adding documentation for the member. - public_member_api_docs
> 
> 11 issues found.

Looks like the latest version is not being pulled:
> webview_flutter_platform_interface 2.13.0 (2.14.0 available)

But why this happens I can't tell.
Could you guide please?

stuartmorgan-g:
> Looks like the latest version is not being pulled:
> 
> > webview_flutter_platform_interface 2.13.0 (2.14.0 available)
> 
> But why this happens I can't tell. Could you guide please?

Because you have code that requires 2.14.0, but (incorrectly) haven't indicated in pubspec.yaml that the package requires 2.14.0, so the `downgrade` tests that validate that the minimum package versions are correctly expressing the requirements—to prevent clients from having exactly this same compilation failure—is failing.

agavrilko:
Thank you @stuartmorgan-g for explaining what `downgrade` tests are.

In this case, since main webview_backage depends on platform interface and depends on wkwebview & andorid, that depend on platform interface as well:

image

I believe it will be right to create another PR to update `webview_flutter_android` and `webview_flutter_wkwebview`.
And then have just `webview_flutter` updated in this one.

Is my understanding correct?

stuartmorgan-g:
> I believe it will be right to create another PR to update `webview_flutter_android` and `webview_flutter_wkwebview`.
> And then have just `webview_flutter` updated in this one.

I don't understand what you are proposing here, since there are no `webview_flutter` changes as part of this PR.

The solution to the CI failures is to update the packages in this PR to properly express the minimum versions of the dependencies they require.

agavrilko:
I have update the references in android and wkwebview packages.

packages/webview_flutter/webview_flutter_android/lib/src/android_webview_controller.dart
```diff
@@ -18,6 +18,41 @@ import 'android_webkit_constants.dart';
...
+  }) {
+    return AndroidLoadFileParams(
+      absoluteFilePath: params.absoluteFilePath,
+      headers: headers,
+    );
+  }
+
+  /// Additional HTTP headers to be included when loading the local file.
+  ///
+  /// If not provided at initialization time, doesn't add any addition headers.
```
bparrishMines: nit:

```suggestion
  /// If not provided at initialization time, doesn't add any additional headers.
```

agavrilko: Updated.

PR #8788, opened 2025-03-04
[pigeon] Dart Class Equality

tarrinneal:
Adds Equality and Hash methods to custom classes in Dart.
I am still working on other languages, but this can be a standalone pr while I complete the others.

fixes https://github.com/flutter/flutter/issues/118087


packages/pigeon/example/app/lib/src/event_channel_messages.g.dart
```diff
@@ -11,6 +11,65 @@ import 'dart:typed_data' show Float64List, Int32List, Int64List, Uint8List;
...
+  for (int i = 0; i < list1.length; i++) {
+    if (list1[i] is List) {
+      elementsMatch =
+          _listEquals(list1[i] as List<Object?>?, list2[i] as List<Object?>?);
+    } else if (list1[i] is Map) {
+      elementsMatch = _mapEquals(list1[i] as Map<Object?, Object?>?,
+          list2[i] as Map<Object?, Object?>?);
+    } else {
+      elementsMatch = list1[i] == list2[i];
+    }
```
stuartmorgan, line 34: The logic here is the same as below, so you can extract a helper. And you can also simplify the helper with patterns:

```dart
bool _deepEquals(Object? a, Object? b) {
  return switch ((a, b)) {
    (List a, List b) => _listEquals(a, b),
    (Map a, Map b) => _mapEquals(a, b),
    _ => a == b,
  };
}
```

Then everything here after the initial checks becomes:
```dart
  for (int i = 0; i < a.length; i++) {
    if (!_deepEquals(a[i], b[i])) {
      return false;
    }
  }
  return true;
```

packages/pigeon/example/app/lib/src/event_channel_messages.g.dart
```diff
@@ -11,6 +11,65 @@ import 'dart:typed_data' show Float64List, Int32List, Int64List, Uint8List;
...
+    } else if (map1[key] is Map) {
+      elementsMatch = _mapEquals(map1[key] as Map<Object?, Object?>?,
+          map2[key] as Map<Object?, Object?>?);
+    } else {
+      elementsMatch = map1[key] == map2[key];
+    }
+    if (!elementsMatch) {
+      return false;
+    }
+  }
```
stuartmorgan, line 69: And this one becomes even simpler, with this whole loop becoming:

```dart
  return !a.keys.any((Object? key) => !b.containsKey(key) || !_deepEquals(a[key], b[key]));
```

packages/pigeon/example/app/lib/src/event_channel_messages.g.dart
```diff
@@ -41,18 +120,38 @@ class StringEvent extends PlatformEvent {
 
   String data;
 
-  Object encode() {
+  List<Object?> toList() {
```
stuartmorgan, line 123: Let's make this private.

packages/pigeon/lib/src/dart/dart_generator.dart
```diff
@@ -288,6 +303,49 @@ class DartGenerator extends StructuredGenerator<DartOptions> {
...
+            classDefinition.fields
+                .map((NamedType field) => field.type.baseName == 'List' ||
+                        field.type.baseName == 'Float64List' ||
+                        field.type.baseName == 'Int32List' ||
+                        field.type.baseName == 'Int64List' ||
+                        field.type.baseName == 'Uint8List'
+                    ? '_listEquals(${field.name}, other.${field.name})'
+                    : field.type.baseName == 'Map'
+                        ? '_mapEquals(${field.name}, other.${field.name})'
+                        : '${field.name} == other.${field.name}')
```
stuartmorgan, line 336: You would also then have the option of just unconditionally using `_deepEquals` here. There is the argument that doing the type check here at generation time makes the runtime check slightly more efficient, but absent an actual use case where we someone cares about micro-optimizing equality checks (our own use cases are for tests, where it definitely doesn't matter) I would lean toward making the generator code simpler, but either way is fine with me.

stuartmorgan (changes requested):

packages/pigeon/lib/src/dart/dart_generator.dart
tarrinneal: Makes sense to me, I just don't like losing the type safety. It will obviously be added in the method itself though, so it's not a big deal.

tarrinneal:
I didn't quite do what you requested, but I did something.

packages/pigeon/example/app/lib/src/messages.g.dart
```diff
@@ -29,6 +29,23 @@ List<Object?> wrapResponse(
   return <Object?>[error.code, error.message, error.details];
 }
 
+bool _deepEquals(Object? a, Object? b) {
+  if (a is List && b is List || a is Map && b is Map) {
+    if (a is List && b is List) {
+      a = a.asMap();
+      b = b.asMap();
```
stuartmorgan, line 36: Why? This is constructing an entirely new collection that's larger than the list, just to test equality—and then doing a bunch of extra compares as well.

packages/pigeon/example/app/lib/src/messages.g.dart
```diff
@@ -29,6 +29,23 @@ List<Object?> wrapResponse(
...
+    }
+    final Map<Object?, Object?> a1 = a! as Map<Object?, Object?>;
+    final Map<Object?, Object?> b1 = b! as Map<Object?, Object?>;
+
+    final List<Object?> keys = a1.keys.toList();
```
stuartmorgan, line 41: The `toList` here is also constructing an entirely new data structure, when all you need is the `Iterable` that `keys` returns.

stuartmorgan (changes requested):

packages/pigeon/example/app/lib/src/messages.g.dart
tarrinneal: honestly, mostly just to see if there was a way to merge the logic as much as possible. I will break it apart again for efficiency though :)

packages/pigeon/example/app/lib/src/messages.g.dart
```diff
@@ -29,6 +29,20 @@ List<Object?> wrapResponse(
   return <Object?>[error.code, error.message, error.details];
 }
 
+bool _deepEquals(Object? a, Object? b) {
+  if (a is List && b is List) {
+    return !a.indexed
+        .any(((int, dynamic) item) => !_deepEquals(item.$2, b[item.$1]));
```
hellohuanlin, line 35: (1) nit: maybe more readable with `every`?
(2) what if `b` has more items than `a`?

packages/pigeon/example/app/lib/src/messages.g.dart
```diff
@@ -29,6 +29,20 @@ List<Object?> wrapResponse(
...
+  }
+  if (a is Map && b is Map) {
+    final Iterable<Object?> keys = (a as Map<Object?, Object?>).keys;
+    return !keys.any((Object? key) =>
+        !(b as Map<Object?, Object?>).containsKey(key) ||
+        !_deepEquals(a[key], b[key]));
```
hellohuanlin, line 41: same here.

packages/pigeon/example/app/lib/src/messages.g.dart
tarrinneal: I spent some time looking for an 'all' method, but every is what I was hoping for, thanks for mentioning that.

packages/pigeon/example/app/lib/src/messages.g.dart
```diff
@@ -29,6 +29,22 @@ List<Object?> wrapResponse(
   return <Object?>[error.code, error.message, error.details];
 }
 
+bool _deepEquals(Object? a, Object? b) {
+  if (a is List && b is List) {
```
stuartmorgan, line 33: Not a fan of the record+pattern approach, eh? ;)

packages/pigeon/example/app/lib/src/messages.g.dart
```diff
@@ -29,6 +29,22 @@ List<Object?> wrapResponse(
...
+    return a.length == b.length &&
+        a.indexed
```
stuartmorgan, line 35: Nice, I forgot `indexed` was a thing.

packages/pigeon/platform_tests/shared_test_plugin_code/lib/integration_tests.dart
```diff
@@ -57,7 +57,7 @@ void runPigeonIntegrationTests(TargetGenerator targetGenerator) {
       final HostIntegrationCoreApi api = HostIntegrationCoreApi();
 
       final AllTypes echoObject = await api.echoAllTypes(genericAllTypes);
-      compareAllTypes(echoObject, genericAllTypes);
+      expect(echoObject == genericAllTypes, true);
```
stuartmorgan, line 60: Can't these all be `expect(foo, bar)` instead of `expect(foo == bar, true)`? The former should give better failure messages.

stuartmorgan (approved):
LGTM

packages/pigeon/platform_tests/shared_test_plugin_code/lib/integration_tests.dart
tarrinneal: As long as expect uses the built in equals method, yeah.

tarrinneal: Just realized this is in the integration tests changes, not the tests I wrote for this change. Will update.

PR #8803, opened 2025-03-05
manual roll with video player test disabled

tarrinneal:
This manual roll should fix the broken rollers and green the tree.

issue filed https://github.com/flutter/flutter/issues/164651 to track this tests failure.


stuartmorgan (approved):
LGTM

tarrinneal:
The test I disabled is still failing somehow...

stuartmorgan:
Try adding a no-op test in the file? It may be that a file with no tests that actually run is considered an error.

PR #8887, opened 2025-03-17
[pigeon] kotlin equality methods

tarrinneal:
Adds `equals` and `hash` methods to Kotlin classes.

More classes may be added to this pr if reviews are late enough.

work towards https://github.com/flutter/flutter/issues/118087


tarrinneal:
@stuartmorgan I added a minor change to naming of internal classes to avoid lint issues as well. In the process I noticed that I hadn't extended the base class for the language internal classes. I am not sure why it was working before at all. Might be worth another quick review.

packages/pigeon/lib/src/generator_tools.dart
```diff
@@ -861,3 +861,11 @@ String makeClearChannelName(String dartPackageName) {
     dartPackageName: dartPackageName,
   );
 }
+
+/// Whether the type is a collection.
+bool isCollectionType(TypeDeclaration type) {
+  return !type.isClass &&
+      !type.isEnum &&
+      !type.isProxyApi &&
+      (type.baseName.contains('List') || type.baseName.startsWith('Map'));
```
stuartmorgan, line 870: `endsWith('List')` for slightly more efficiency?

PR #8897, opened 2025-03-25
[all_packages] Enforce property assignment for compile sdk over method assignment

reidbaker:
No issue I just noticed that we were almost consistent in our approach to compileSdk. 

- [x] I signed the [CLA].


script/tool/lib/src/gradle_check_command.dart
```diff
@@ -421,12 +421,20 @@ for more details.''';
...
     if (compileSdkLine == null) {
-      printError('${indentation}No compileSdk or compileSdkVersion found.');
+       // Equals regex not found check for method pattern.
+      final RegExp compileSpacePattern = RegExp(r'^\s*compileSdk');
+      final String? methodAssignmentLine = gradleLines
+        .firstWhereOrNull((String line) => compileSpacePattern.hasMatch(line));
+      if (methodAssignmentLine == null) {
+        printError('${indentation}No compileSdk or compileSdkVersion found.');
+      } else {
+        printError('${indentation}No compileSdk = found. Please use property assignment.');
```
stuartmorgan, line 436: How about `No "compileSdk =" found` to make this easier to parse if you don't already know what the issue is?

stuartmorgan (commented):
Tool changes LGTM, but I'm still seeing ~40 violations of this locally.

jesswrd (approved):

tarrinneal (approved):

reidbaker:
dart ./script/tool/bin/flutter_plugin_tools.dart gradle-check now passes locally. 
 Command used to find all the build.gradle files in violation `find . -type f -name "build.gradle" | xargs grep -e "compileSdk flut" | cut -d : -f1 | xargs code`

.ci/legacy_project/all_packages/android/app/build.gradle
```diff
@@ -24,7 +24,7 @@ if (flutterVersionName == null) {
 
 android {
     namespace = "com.example.all_packages"
-    compileSdkVersion flutter.compileSdkVersion
```
stuartmorgan, line 27: Do we actually want to change this one? If we somehow broke it (maybe that's not even possible, but then... gradle), we would want to know presumably. And this is what the template generated at the time.

packages/flutter_plugin_android_lifecycle/example/android/app/build.gradle
```diff
@@ -24,13 +24,13 @@ if (flutterVersionName == null) {
...
-    compileSdk flutter.compileSdkVersion
+    compileSdk = flutter.compileSdkVersion
 
 
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
         applicationId "io.flutter.plugins.flutter_plugin_android_lifecycle_example"
-        minSdkVersion flutter.minSdkVersion
+        minSdkVersion = flutter.minSdkVersion
         targetSdkVersion flutter.targetSdkVersion
```
stuartmorgan, line 34: Why `minSdkVersion` but not `targetSdkVersion`?

Is the plan just to wait and fix the assignment type when standardizing on `targetSdk`?

script/tool/lib/src/gradle_check_command.dart
```diff
@@ -421,12 +421,21 @@ for more details.''';
 
   bool _validateCompileSdkUsage(
```
stuartmorgan, line 422: Do we want to enforce the `ndkVersion` style too?

stuartmorgan (approved):
LGTM. All comments optional.

packages/flutter_plugin_android_lifecycle/example/android/app/build.gradle
reidbaker: accidental add, removing. I do want to move us to use equals and minSdk and targetSdk but only with the enforcement to keep it that way.

script/tool/lib/src/gradle_check_command.dart
reidbaker: Eventually, I reverted the min change but most of the ndk versions were adjacent to the compileSdk so I went ahead and did them. I am willing to revert the ndk changes if you want to enforce it with tooling.

stuartmorgan: If we care about it I think we should eventually enforce it, but I don't see any reason to block incidental updates while touching surrounding code on enforcement.

.ci/legacy_project/all_packages/android/app/build.gradle
stuartmorgan: > That said what would break is not our code but agp code

If that happened we would want to put in a Flutter Fix or an auto-migrator though, presumably? Getting a warning about needing to do that from our CI could be nice.

> I decided removing the last instance of compileSdkVersion was a good idea. If you feel strongly then I can revert this.

I would lean toward leaving it the way it was since in general the idea of `legacy_project` is to make absolutely minimal changes to it, but I'm fine with you making a call on it either way.

> I do not think gradle check runs against the legacy project so it wont cause a breakage.

Right, the check runs on individual packages, and won't include this file.

reidbaker: reverted.

chunhtai (approved):
LGTM

reidbaker:
Failures appear unrelated. Hopefully an updated branch will fix the issue.

reidbaker:
Blocked on https://github.com/flutter/flutter/issues/165664

ditman:
Clicked "Update branch" because web tests seem to be fixed in master after @stuartmorgan's [fix](https://github.com/flutter/packages/pull/8909).

stuartmorgan:
> autosubmit label was removed for flutter/packages/8897, because - The status or check suite [Mac_arm64 ios_platform_tests_shard_2 master](https://cr-buildbucket.appspot.com/build/8719234070045012897) has failed. Please fix the issues identified (or deflake) before re-applying this label.

I'm seeing this same issue on a roller PR. I'm wondering if it's OOB failure from a maps SDK or server change.

ditman:
Looking at [the build history for the suite](https://ci.chromium.org/ui/p/flutter/builders/try/Mac_arm64%20ios_platform_tests_shard_2%20master?limit=200), it seems this was the first build broken like this was on: 2025-03-25 @ 09:55:

* https://ci.chromium.org/ui/p/flutter/builders/try/Mac_arm64%20ios_platform_tests_shard_2%20master/20650/overview

Associated to this commit:

* https://github.com/flutter/packages/commit/347d39a7bc6e3fd3391d98e48dec296a3f986382 (the branch name according to github was "xcode_16_test" 🔮)

stuartmorgan-g:
Once https://github.com/flutter/packages/pull/8967 lands, merging in `main` should unblock this.

PR #8909, opened 2025-03-26
[various] Disable sandbox in Chrome dart tests

stuartmorgan:
`flutter test` automatically disables sandbox in headless mode, but `dart test` does not, and the Linux CI bots no longer support the sandbox, so this turns it off explicitly for packages that don't rely on Flutter (and thus use `dart test`).


stuartmorgan:
test-exempt: is test config

tarrinneal (approved):

ditman:
Oh jeez, I've just seen this issue+pr. Thanks for fixing this!

PR #8920, opened 2025-03-27
Use a more deterministic way of waiting for ad widgets to be ready.

eyebrowsoffire:
These unit tests were failing when a skwasm change caused a subtle timing difference. See https://github.com/flutter/flutter/issues/165347

### DIFF
--- a/packages/google_adsense/example/integration_test/experimental_ad_unit_widget_test.dart
+++ b/packages/google_adsense/example/integration_test/experimental_ad_unit_widget_test.dart
@@ -19,6 +19,17 @@ import 'js_interop_mocks/adsense_test_js_interop.dart';
 const String testClient = 'test_client';
 const String testSlot = 'test_slot';
 
+class CallbackTracker {
+  void Function() createCallback() {
+    final int callbackIndex = _callbackStates.length;
+    _callbackStates.add(false);
+    return () => _callbackStates[callbackIndex] = true;
+  }
+
+  bool get allCalled => _callbackStates.every((bool state) => state);
+  final List<bool> _callbackStates = <bool>[];
+}
+
 void main() async {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
@@ -47,15 +58,17 @@ void main() async {
 
       await adSense.initialize(testClient);
 
+      final CallbackTracker tracker = CallbackTracker();
       final Widget adUnitWidget = AdUnitWidget(
         configuration: AdUnitConfiguration.displayAdUnit(
           adSlot: testSlot,
           adFormat: AdFormat.AUTO, // Important!
         ),
         adClient: adSense.adClient,
+        onInjected: tracker.createCallback(),
       );
 
-      await pumpAdWidget(adUnitWidget, tester);
+      await pumpAdWidget(adUnitWidget, tester, tracker);
 
       // Then
       // Widget level
@@ -81,11 +94,13 @@ void main() async {
 
       await adSense.initialize(testClient);
 
+      final CallbackTracker tracker = CallbackTracker();
       final Widget adUnitWidget = AdUnitWidget(
         configuration: AdUnitConfiguration.displayAdUnit(
           adSlot: testSlot,
         ),
         adClient: adSense.adClient,
+        onInjected: tracker.createCallback(),
       );
 
       final Widget constrainedAd = Container(
@@ -93,7 +108,7 @@ void main() async {
         child: adUnitWidget,
       );
 
-      await pumpAdWidget(constrainedAd, tester);
+      await pumpAdWidget(constrainedAd, tester, tracker);
 
       // Then
       // Widget level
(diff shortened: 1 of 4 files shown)


packages/google_adsense/example/integration_test/experimental_ad_unit_widget_test.dart
```diff
@@ -19,6 +19,17 @@ import 'js_interop_mocks/adsense_test_js_interop.dart';
...
 const String testSlot = 'test_slot';
 
+class CallbackTracker {
+  void Function() createCallback() {
+    final int callbackIndex = _callbackStates.length;
+    _callbackStates.add(false);
+    return () => _callbackStates[callbackIndex] = true;
+  }
+
+  bool get allCalled => _callbackStates.every((bool state) => state);
```
ditman, line 29: What about something like:

`Future<bool> allCalled => Future.wait(_callbackFutures).timeout(const Duration(seconds: 10));`

The implementation of createCallback would insert a `Future<bool>` (or `Future<void>`, probably doesn't matter) into the `_callbackStates`, and the function would just `.complete` the completer for that future.

With the timeout, we can ensure the "busy loop" pumping frames doesn't keep pumping frames until the test times out (which can take several minutes to do so), and tests should fail kinda fast. (I think that if left unhandled, the `timeout` just throws a TimeoutException.)

packages/google_adsense/example/integration_test/experimental_ad_unit_widget_test.dart
```diff
@@ -203,10 +225,10 @@ Future<void> pumpAdWidget(Widget adUnit, WidgetTester tester) async {
     ),
   );
 
-  // This extra pump is needed for the platform view to actually render in the DOM.
-  await tester.pump();
-  // One more for skwasm.
-  await tester.pump();
+  while (!tracker.allCalled) {
```
ditman, line 228: If `allCalled` becomes a `Future`, this loop may be reimplemented with `Future.doWhile` or a `tracker.allCalled.then((_) { allCalled = true });` to stop the while loop.

packages/google_adsense/lib/src/adsense/ad_unit_widget.dart
```diff
@@ -58,13 +61,14 @@ class _AdUnitWidgetWebState extends State<AdUnitWidget>
   @override
   bool get wantKeepAlive => true;
 
-  static final web.ResizeObserver _adSenseResizeObserver = web.ResizeObserver(
+  web.ResizeObserver get _adSenseResizeObserver => web.ResizeObserver(
```
ditman, line 64: This may now be creating multiple ResizeObservers on the DOM, whereas we only used to have a single one that was reused for all DOM elements.

I don't think this is a big deal performance-wise, but this could be refined a little bit so we keep the single static resize observer, but then expose an instance API to attach listeners to it like:

`void observe(Element, onResize)`, 

instead of directly using the `static _adSenseResizeObserver` directly.

This is not a blocking issue (but maybe a TODO for the future?)

packages/google_adsense/example/integration_test/experimental_ad_unit_widget_test.dart
```diff
@@ -19,6 +19,17 @@ import 'js_interop_mocks/adsense_test_js_interop.dart';
...
 
+class CallbackTracker {
+  void Function() createCallback() {
+    final int callbackIndex = _callbackStates.length;
+    _callbackStates.add(false);
+    return () => _callbackStates[callbackIndex] = true;
+  }
+
+  bool get allCalled => _callbackStates.every((bool state) => state);
+  final List<bool> _callbackStates = <bool>[];
```
ditman, line 30: (Note to self: This does not need a "reset" method or similar, because each test creates its own instance of the `CallbackTracker`)

ditman (approved):
This LGTM! I'd make the `allCalled` a `Future` so we can race it against a timeout, and not leave the test hanging "forever" in a busy while loop; we can also spit out a better error message when the timeout fails, rather than the generic "the test timed out" failure, which is quite obscure in the CI logs.

ditman:
Ah, just noticed the `Linux repo_checks`. Since this is changing the implementation of the package (making the ResizeObserver non static), this probably needs a patch version roll at least, so the pubspec.yaml version need to be updated (to `0.1.1+1`, for example), and a small CHANGELOG entry added.

packages/google_adsense/example/integration_test/experimental_ad_unit_widget_test.dart
```diff
@@ -203,10 +226,14 @@ Future<void> pumpAdWidget(Widget adUnit, WidgetTester tester) async {
...
   );
 
-  // This extra pump is needed for the platform view to actually render in the DOM.
-  await tester.pump();
-  // One more for skwasm.
-  await tester.pump();
+  final Stopwatch timer = Stopwatch()..start();
+  while (!tracker.allCalled) {
+    if (timer.elapsedMilliseconds > 1000) {
+      fail('timeout while waiting for ad widget to be injected');
```
ditman, line 232: Sweet!

PR #8967, opened 2025-03-31
[google_maps_flutter] Skip more hanging iOS tests

stuartmorgan-g:
Now `testAnimateCameraWithDuration` is failing as well, as I had expected previously. It's not clear why it didn't fail in the PR where I disabled the `WithoutDuration` variant but is failing now.

Part of https://github.com/flutter/flutter/issues/166139


vashworth (approved):
LGTM

PR #8971, opened 2025-04-01
[pigeon] swift equality methods

tarrinneal:
adds == and hash methods to classes


packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
```diff
@@ -23,6 +23,67 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
...
+    return true
+
+  case (nil, _), (_, nil):
+    return false
+
+  case is (Void, Void):
+    return true
+
+  case let (cleanLhsHashable, cleanRhsHashable) as (AnyHashable, AnyHashable):
+    return cleanLhsHashable == cleanRhsHashable
```
hellohuanlin, line 40: Hashable doesnt mean equal

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
```diff
@@ -23,6 +23,67 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
...
+  case let (cleanLhsDictionary, cleanRhsDictionary) as ([AnyHashable: Any?], [AnyHashable: Any?]):
+    guard cleanLhsDictionary.count == cleanRhsDictionary.count else { return false }
+    for (key, cleanLhsValue) in cleanLhsDictionary {
+      guard let cleanRhsValue = cleanRhsDictionary[key] else { return false }
+      if !deepEqualsEventChannelMessages(cleanLhsValue, cleanRhsValue) {
+        return false
+      }
+    }
+    return true
+
```
hellohuanlin, line 60: can you also check the case when both conform to `Equatable`, then use `==` to compare them? (e.g. String is Equatable)

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
```diff
@@ -23,6 +23,67 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
...
+  default:
+    return String(describing: cleanLhs) == String(describing: cleanRhs)
```
hellohuanlin, line 62: where does the input come from? one could overwrite a type's `description` to return arbitrary strings.

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
tarrinneal: hashable inherits from equitable, there is no AnyEquitable.

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
tarrinneal: String is Hashable, therefore already covered. Any type that is supported by pigeon is covered by hashable except lists and maps. And certain classes in proxy apis, but this isn't meant to solve for non-Hashable classes, as there is no real way to compare if they aren't Equitable.

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
tarrinneal: yeah, I couldn't find a better way to manage this besides just returning false. This seemed potentially better? This isn't (theoretically) hittable anyway since pigeon types are limited

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
hellohuanlin: sorry i misread. i thought you were comparing 2 hash values.

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
hellohuanlin: Sorry a String isn't a good example. Worth putting a comment there to explain a bit, since this is missing Non Hashable but Equatable case.

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
hellohuanlin: when will we hit the default case? we might as well just return `false` if we already covered all known cases that return `true`. Also worth adding a comment.

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
```diff
@@ -23,6 +23,67 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
...
+      if !deepEqualsEventChannelMessages(element, cleanRhsArray[index]) {
+        return false
+      }
+    }
+    return true
+
+  case let (cleanLhsDictionary, cleanRhsDictionary) as ([AnyHashable: Any?], [AnyHashable: Any?]):
+    guard cleanLhsDictionary.count == cleanRhsDictionary.count else { return false }
+    for (key, cleanLhsValue) in cleanLhsDictionary {
+      guard let cleanRhsValue = cleanRhsDictionary[key] else { return false }
```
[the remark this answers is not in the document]
tarrinneal: we don't support nil keys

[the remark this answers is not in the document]
stuartmorgan-g: Not nil keys, nil values. `guard let cleanRhsValue = cleanRhsDictionary[key] else { return false }` will unconditionally return false for `"foo": nil` even if both dictionaries have it.

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
```diff
@@ -23,6 +23,67 @@ private func nilOrValue<T>(_ value: Any?) -> T? {
...
+}
+
+func deepHashEventChannelMessages(value: Any?, hasher: inout Hasher) {
+  if let valueList = value as? [AnyHashable] {
+    for item in valueList { deepHashEventChannelMessages(value: item, hasher: &hasher) }
+    return
+  }
+
+  if let valueDict = value as? [AnyHashable: AnyHashable] {
+    for key in valueDict.keys {
```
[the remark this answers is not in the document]
tarrinneal: This is already in this shape, not sure why it's not showing in your review

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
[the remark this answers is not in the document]
tarrinneal: ah, misread, my bad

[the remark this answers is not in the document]
tarrinneal: There is no solution to this problem that doesn't include creating a list of keys

stuartmorgan-g (approved):
LGTM

PR #4281, opened 2023-06-23
[google_maps_flutter] Fix for memory leak impacting all platforms due to subscriptions not getting cleaned up

wolfie82:
*Prior to this change, GoogleMaps would create a condition where its state wasn't cleaned up.  Any widget that was leveraging GoogleMap wasn't cleaned up if it relied on a subscription, such as onCameraIdle, onCameraMoved, etc.  This was creating crashes for our team specifically.*

*List which issues are fixed by this PR. You must list at least one issue.*

[#129089](https://github.com/flutter/flutter/issues/129089)
[#115283](https://github.com/flutter/flutter/issues/115283)
[#92788](https://github.com/flutter/flutter/issues/92788)

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [X] I signed the [CLA].
- [X] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [X] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [X] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[...]

PR #8972, opened 2025-04-01
[google_maps_flutter] Fix for memory leak impacting all platforms due to subscriptions not getting cleaned up 

gentlemanxzh:
This PR is #4281 follow-up,Fix map memory leak

[#129089](https://github.com/flutter/flutter/issues/129089)
Fixes [#115283](https://github.com/flutter/flutter/issues/115283)
[#92788](https://github.com/flutter/flutter/issues/92788)

## Pre-Review Checklist

- [x] I signed the [CLA].


gentlemanxzh:
Will this PR still be reviewed? @stuartmorgan-g

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.12.2
+
+* Fixes memory leak by implementing mechanism for disposing stream subscriptions in `GoogleMapController`.
```
stuartmorgan-g, line 3: Just this sentence is enough for the changelog; anyone interested in the level of detail in the following sentences can always look into the PR and/or issue.

stuartmorgan-g, line 3: Also, "by implementing mechanism for disposing" -> "by disposing".

packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
```diff
@@ -18,6 +18,14 @@ class GoogleMapController {
   /// The mapId for this controller
   final int mapId;
 
+  /// List of active stream subscriptions for map events.
+  ///
+  /// This list keeps track of all event subscriptions created for the map,
+  /// including camera movements, marker interactions, and other map events.
+  /// These subscriptions are properly disposed when the controller is disposed.
```
stuartmorgan-g, line 25: "are properly disposed" -> "should be disposed"

Nothing about the existence of the field guarantees that the disposal actually happens, so the comment on the field shouldn't assert it.

packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
```diff
@@ -38,53 +46,108 @@ class GoogleMapController {
...
+          ),
+    );
+  }
+
+  /// Adds a stream subscription to the list of active subscriptions.
+  ///
+  /// This method is used to track and manage all stream subscriptions
+  /// created for map events. The subscriptions are stored in
+  /// [_streamSubscriptionList] and will be properly disposed when the
+  /// controller is disposed.
```
stuartmorgan-g, line 135: This second sentence should be removed, as it is asserting things outside of the control of this method.

In general, it is an anti-pattern for a comment in one part of the code to make assertions about *other code*, because the chances of someone changing code updating all assertions about that code anywhere else in the codebase are extremely low, making them highly prone to becoming outdated and misleading.

packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
```diff
@@ -18,6 +18,14 @@ class GoogleMapController {
   /// The mapId for this controller
   final int mapId;
 
+  /// List of active stream subscriptions for map events.
+  ///
+  /// This list keeps track of all event subscriptions created for the map,
+  /// including camera movements, marker interactions, and other map events.
+  /// These subscriptions are properly disposed when the controller is disposed.
+  final List<StreamSubscription<dynamic>> _streamSubscriptionList =
```
stuartmorgan-g, line 26: Generally variable names shouldn't include their types unless they communicate some important information. This can just be called `_streamSubscriptions`.

packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
```diff
@@ -38,53 +46,108 @@ class GoogleMapController {
...
+  }
+
+  /// Adds a stream subscription to the list of active subscriptions.
+  ///
+  /// This method is used to track and manage all stream subscriptions
+  /// created for map events. The subscriptions are stored in
+  /// [_streamSubscriptionList] and will be properly disposed when the
+  /// controller is disposed.
+  void _addSubscription(StreamSubscription<dynamic> subscription) {
+    _streamSubscriptionList.add(subscription);
```
stuartmorgan-g, line 137: What's the purpose of having a wrapper method? `_addSubscription` doesn't seem any more readable than `_streamSubscriptions.add`.

packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
```diff
@@ -38,53 +46,108 @@ class GoogleMapController {
...
+  }
+
+  /// Adds a stream subscription to the tracked list for testing purposes.
+  ///
+  /// This method allows tests to add custom stream subscriptions to the
```
stuartmorgan-g, line 142: Adding `visibleForTesting` methods should only be done if there is no way to test the functionality using public APIs, and it generally indicates that some part of the actual production flow is not being tested.

In this case, this method is being used to bypass testing of all of the changes in `_connectStreams`. The goal of this PR is not for there to be a method that tracks subscriptions for disposal, the goal is for all the subscriptions to be cleaned up. If all of the changes to `_connectStreams` were reverted (or more subtly, any one of them were missing), the tests would still pass even though the issues this PR is intended to fix would not actually be fixed.

Given that the unit tests control the `GoogleMapsFlutterPlatform.instance` that vends all of the subscriptions, it should be possible to test—using only public API—that the platform instance's streams are listened to after init, but none of them are after dispose, thus fully verifying the desired behavior.

stuartmorgan-g (changes requested):
Thanks for the contribution!

packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
gentlemanxzh: You make a good point. Based on your suggestion, I was wondering if we could test this by checking the return value of `GoogleMapsFlutterPlatform.instance.mapEventStreamController.hasListener`. When the map is initialized, the return value would be `true`, indicating that the platform instance's stream is being listened to after initialization. When the map is disposed of, the return value would be `false`, indicating that the stream is no longer being listened to.

packages/google_maps_flutter/google_maps_flutter/test/controller_test.dart
```diff
@@ -0,0 +1,82 @@
...
+      (WidgetTester tester) async {
+    final FakeGoogleMapsFlutterPlatform platform =
+        FakeGoogleMapsFlutterPlatform();
+
+    GoogleMapsFlutterPlatform.instance = platform;
+
+    final FakePlatformViewsController fakePlatformViewsController =
+        FakePlatformViewsController();
+
+    tester.binding.defaultBinaryMessenger.setMockMethodCallHandler(
```
stuartmorgan-g, line 26: What issue was this code added to solve? Tests at this layer should not be relying on the internal implementation details of a (legacy) implementation from another package. Since you're setting the fake instance above it's not clear to me what code could be triggering calls to this handler.

packages/google_maps_flutter/google_maps_flutter/test/controller_test.dart
```diff
@@ -0,0 +1,82 @@
...
+      SystemChannels.platform_views,
+      fakePlatformViewsController.fakePlatformViewsMethodHandler,
+    );
+
+    final ValueNotifier<GoogleMapController?> controllerNotifier =
```
stuartmorgan-g, line 31: Why is this a value notifier rather than completer? Nothing registers any listeners so it's not clear to me what this is doing.

packages/google_maps_flutter/google_maps_flutter/test/controller_test.dart
```diff
@@ -0,0 +1,82 @@
...
+
+      await tester.binding.runAsync(() async {
+        await tester.pump();
+      });
+
+      expect(platform.mapEventStreamController.hasListener, false);
+
+      controllerNotifier.dispose();
+    } else {
+      fail('GoogleMapController not created');
```
stuartmorgan-g, line 68: This will fail racily, because nothing is waiting `onMapCompleter`. Using a completer with an await will fix that, and avoid the need to even have this code.

packages/google_maps_flutter/google_maps_flutter/test/controller_test.dart
```diff
@@ -0,0 +1,82 @@
...
+    ));
+
+    await tester.pump();
+
+    final GoogleMapController? controller = controllerNotifier.value;
+
+    if (controller != null) {
+      expect(platform.mapEventStreamController.hasListener, true);
+
+      await tester.pumpWidget(const Directionality(
```
stuartmorgan-g, line 55: An explicit comment that this is removing the map from the widget tree would help make it clear what is happening here; it's easy to miss while reading that this isn't the usual test boilerplate of putting the widget under test in the tree (and is in fact the opposite).

stuartmorgan-g (changes requested):

stuartmorgan-g (approved):
LGTM

@bparrishMines Could you do the secondary review here?

bparrishMines (approved):
LGTM

PR #4587, opened 2023-07-29
[camera_avfoundation] ignore audio samples until first video sample

misos1:
Samples from ios video and audio input should be regarded as arriving from independent streams because in fact they are going from independent sources. Video is going from instance `AVCaptureVideoDataOutput` while audio from `AVCaptureAudioDataOutput`. Therefore it is possible that one of them will be initialised soon or later than the other and the first few samples will be only audio samples so the video file will start without image data and will be black until the first video sample arrives. This patch addresses this by ignoring audio samples until the first video sample arrives (alternatively it could wait until both video and audio samples are actively arriving).

Fix https://github.com/flutter/flutter/issues/57831

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran the auto-formatter. (Unlike the flutter/flutter repo, the flutter/packages repo does use `dart format`.)
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is [exempt from version changes].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style].
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning
[...]

PR #8982, opened 2025-04-02
[camera_avfoundation] handle interruptions and use single offset

misos1:
Handle interruptions of capture sessions by doing "internal pause". Another possibility would be to remove audio input from the session but this would be too specific for interruptions caused by an incoming call or alarm. Or to stop video recording as the native ios camera app does but this would require some additional event.

Handle asynchronous error notifications posted about capture sessions. This can happen for example when there is an incoming phone call and the camera starts recording with audio. For this specific case would be enough to just call `startRunning` on capture sessions but that is too specific and would require ability to restart capture session from dart side as seems there is no good way how to test when such phone call or other such event ends, and trying to restart it for example every second inside `FLTCam` does not look like good idea. Instead the camera controller enters an error state and can be recovered by disposing and recreating which should be more generic.

Use single offset for both video and audio as separate offsets can cause unsynchronized video with audio. Seems audio timestamps tend to be ignored except for the first sample so prefer them for calculating offset to avoid desynchronization. Avoid errors when trying to append samples with timestamp less or equal to the previous sample. Fix a bug when after adjusting offset also last sample time needs to be updated to avoid adding past offsets multiple times.

- fixes https://github.com/flutter/flutter/issues/149978
- fixes https://github.com/flutter/flutter/issues/151253

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_avfoundation/CHANGELOG.md
+++ b/packages/camera/camera_avfoundation/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.9.18+14
+
+* Handle video and audio interruptions and errors.
+* Use a single time offset for both video and audio.
+
 ## 0.9.18+1
 
 * Refactors implementations to reduce usage of OCMock in internal testing.
--- a/packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
+++ b/packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
@@ -56,8 +56,6 @@ static FlutterError *FlutterErrorFromNSError(NSError *error) {
 @property(readonly, nonatomic) FCPPlatformMediaSettings *mediaSettings;
 @property(readonly, nonatomic) FLTCamMediaSettingsAVWrapper *mediaSettingsAVWrapper;
 @property(nonatomic) FLTImageStreamHandler *imageStreamHandler;
-@property(readonly, nonatomic) AVCaptureSession *videoCaptureSession;
-@property(readonly, nonatomic) AVCaptureSession *audioCaptureSession;
 
 @property(readonly, nonatomic) AVCaptureInput *captureVideoInput;
 /// Tracks the latest pixel buffer sent from AVFoundation's sample buffer delegate callback.
@@ -74,8 +72,7 @@ static FlutterError *FlutterErrorFromNSError(NSError *error) {
 @property(assign, nonatomic) BOOL isFirstVideoSample;
 @property(assign, nonatomic) BOOL isRecording;
 @property(assign, nonatomic) BOOL isRecordingPaused;
-@property(assign, nonatomic) BOOL videoIsDisconnected;
-@property(assign, nonatomic) BOOL audioIsDisconnected;
+@property(assign, nonatomic) BOOL isRecordingDisconnected;
 @property(assign, nonatomic) BOOL isAudioSetup;
 
 /// Number of frames currently pending processing.
@@ -85,10 +82,10 @@ static FlutterError *FlutterErrorFromNSError(NSError *error) {
 @property(assign, nonatomic) int maxStreamingPendingFramesCount;
 
 @property(assign, nonatomic) UIDeviceOrientation lockedCaptureOrientation;
-@property(assign, nonatomic) CMTime lastVideoSampleTime;
-@property(assign, nonatomic) CMTime lastAudioSampleTime;
-@property(assign, nonatomic) CMTime videoTimeOffset;
-@property(assign, nonatomic) CMTime audioTimeOffset;
+@property(assign, nonatomic) CMTime recordingTimeOffset;
+@property(assign, nonatomic) CMTime lastSampleEndTime;
+@property(nonatomic) AVCaptureOutput *outputForOffsetAdjusting;
+@property(assign, nonatomic) CMTime lastAppendedVideoSampleTime;
 @property(nonatomic) CMMotionManager *motionManager;
 @property AVAssetWriterInputPixelBufferAdaptor *videoAdaptor;
 /// All FLTCam's state access and capture session related operations should be on run on this queue.
@@ -313,7 +310,29 @@ static void selectBestFormatForRequestedFrameRate(
 
   [self updateOrientation];
 
+  // Handle video and audio interruptions and errors.
+  // https://github.com/flutter/flutter/issues/151253
+  for (AVCaptureSession *session in @[_videoCaptureSession, _audioCaptureSession]) {
+    [NSNotificationCenter.defaultCenter addObserver:self
+                                           selector:@selector(captureSessionWasInterrupted:)
+                                               name:AVCaptureSessionWasInterruptedNotification
+                                             object:session];
+
+    [NSNotificationCenter.defaultCenter addObserver:self
+                                           selector:@selector(captureSessionRuntimeError:)
+                                               name:AVCaptureSessionRuntimeErrorNotification
+                                             object:session];
+  }
+
   return self;
+}
+
+- (void)captureSessionWasInterrupted:(NSNotification *)notification {
+  _isRecordingDisconnected = YES;
+}
+
+- (void)captureSessionRuntimeError:(NSNotification *)notification {
+  [self reportErrorMessage:[NSString stringWithFormat:@"%@", notification.userInfo[AVCaptureSessionErrorKey]]];
 }
 
 - (AVCaptureConnection *)createConnection:(NSError **)error {
@@ -717,78 +736,71 @@ static void selectBestFormatForRequestedFrameRate(
       });
     }
   }
-  if (_isRecording && !_isRecordingPaused) {
+  if (_isRecording && !_isRecordingPaused && _videoCaptureSession.running && _audioCaptureSession.running) {
     if (_videoWriter.status == AVAssetWriterStatusFailed) {
       [self reportErrorMessage:[NSString stringWithFormat:@"%@", _videoWriter.error]];
       return;
     }
 
-    // ignore audio samples until the first video sample arrives to avoid black frames
-    // https://github.com/flutter/flutter/issues/57831
-    if (_isFirstVideoSample && output != _captureVideoOutput) {
-      return;
+    // do not append sample buffer when readyForMoreMediaData is NO to avoid crash
+    // https://github.com/flutter/flutter/issues/132073
+    if (output == _captureVideoOutput) {
+      if (!_videoWriterInput.readyForMoreMediaData) {
+        return;
+      }
+    } else {
+      // ignore audio samples until the first video sample arrives to avoid black frames
+      // https://github.com/flutter/flutter/issues/57831
+      if (_isFirstVideoSample || !_audioWriterInput.readyForMoreMediaData) {
+        return;
+      }
+      _outputForOffsetAdjusting = output;
     }
 
-    CMTime currentSampleTime = CMSampleBufferGetPresentationTimeStamp(sampleBuffer);
+    CMTime sampleTime = CMSampleBufferGetPresentationTimeStamp(sampleBuffer);
 
     if (_isFirstVideoSample) {
-      [_videoWriter startSessionAtSourceTime:currentSampleTime];
+      [_videoWriter startSessionAtSourceTime:sampleTime];
       // fix sample times not being numeric when pause/resume happens before first sample buffer
       // arrives
       // https://github.com/flutter/flutter/issues/132014
-      _lastVideoSampleTime = currentSampleTime;
-      _lastAudioSampleTime = currentSampleTime;
+      _isRecordingDisconnected = NO;
       _isFirstVideoSample = NO;
     }
 
-    if (output == _captureVideoOutput) {
-      if (_videoIsDisconnected) {
-        _videoIsDisconnected = NO;
+    CMTime currentSampleEndTime = sampleTime;
+    CMTime dur = CMSampleBufferGetDuration(sampleBuffer);
+    if (CMTIME_IS_NUMERIC(dur)) {
+      currentSampleEndTime = CMTimeAdd(currentSampleEndTime, dur);
+    }
 
-        if (_videoTimeOffset.value == 0) {
-          _videoTimeOffset = CMTimeSubtract(currentSampleTime, _lastVideoSampleTime);
-        } else {
-          CMTime offset = CMTimeSubtract(currentSampleTime, _lastVideoSampleTime);
-          _videoTimeOffset = CMTimeAdd(_videoTimeOffset, offset);
-        }
-
-        return;
+    // Use a single time offset for both video and audio.
+    // https://github.com/flutter/flutter/issues/149978
+    if (_isRecordingDisconnected) {
+      if (output == _outputForOffsetAdjusting) {
+        CMTime offset = CMTimeSubtract(currentSampleEndTime, _lastSampleEndTime);
+        _recordingTimeOffset = CMTimeAdd(_recordingTimeOffset, offset);
+        _lastSampleEndTime = currentSampleEndTime;
+        _isRecordingDisconnected = NO;
       }
+      return;
+    }
 
-      _lastVideoSampleTime = currentSampleTime;
+    if (output == _outputForOffsetAdjusting) {
+      _lastSampleEndTime = currentSampleEndTime;
+    }
 
+    if (output == _captureVideoOutput) {
       CVPixelBufferRef nextBuffer = CMSampleBufferGetImageBuffer(sampleBuffer);
-      CMTime nextSampleTime = CMTimeSubtract(_lastVideoSampleTime, _videoTimeOffset);
-      // do not append sample buffer when readyForMoreMediaData is NO to avoid crash
-      // https://github.com/flutter/flutter/issues/132073
-      if (_videoWriterInput.readyForMoreMediaData) {
+      CMTime nextSampleTime = CMTimeSubtract(sampleTime, _recordingTimeOffset);
+      if (CMTIME_COMPARE_INLINE(nextSampleTime, >, _lastAppendedVideoSampleTime)) {
         [_videoAdaptor appendPixelBuffer:nextBuffer withPresentationTime:nextSampleTime];
+        _lastAppendedVideoSampleTime = nextSampleTime;
       }
     } else {
-      CMTime dur = CMSampleBufferGetDuration(sampleBuffer);
-
-      if (dur.value > 0) {
-        currentSampleTime = CMTimeAdd(currentSampleTime, dur);
-      }
-
-      if (_audioIsDisconnected) {
-        _audioIsDisconnected = NO;
-
-        if (_audioTimeOffset.value == 0) {
-          _audioTimeOffset = CMTimeSubtract(currentSampleTime, _lastAudioSampleTime);
-        } else {
-          CMTime offset = CMTimeSubtract(currentSampleTime, _lastAudioSampleTime);
-          _audioTimeOffset = CMTimeAdd(_audioTimeOffset, offset);
-        }
-
-        return;
-      }
-
-      _lastAudioSampleTime = currentSampleTime;
-
-      if (_audioTimeOffset.value != 0) {
+      if (_recordingTimeOffset.value != 0) {
         CMSampleBufferRef adjustedSampleBuffer =
-            [self copySampleBufferWithAdjustedTime:sampleBuffer by:_audioTimeOffset];
+            [self copySampleBufferWithAdjustedTime:sampleBuffer by:_recordingTimeOffset];
         [self newAudioSample:adjustedSampleBuffer];
         CFRelease(adjustedSampleBuffer);
       } else {
@@ -820,10 +832,8 @@ static void selectBestFormatForRequestedFrameRate(
     }
     return;
   }
-  if (_videoWriterInput.readyForMoreMediaData) {
-    if (![_videoWriterInput appendSampleBuffer:sampleBuffer]) {
-      [self reportErrorMessage:@"Unable to write to video input"];
-    }
+  if (![_videoWriterInput appendSampleBuffer:sampleBuffer]) {
+    [self reportErrorMessage:@"Unable to write to video input"];
   }
 }
 
@@ -834,10 +844,8 @@ static void selectBestFormatForRequestedFrameRate(
     }
     return;
   }
-  if (_audioWriterInput.readyForMoreMediaData) {
-    if (![_audioWriterInput appendSampleBuffer:sampleBuffer]) {
-      [self reportErrorMessage:@"Unable to write to audio input"];
-    }
+  if (![_audioWriterInput appendSampleBuffer:sampleBuffer]) {
+    [self reportErrorMessage:@"Unable to write to audio input"];
   }
 }
 
@@ -862,6 +870,7 @@ static void selectBestFormatForRequestedFrameRate(
     CFRelease(_latestPixelBuffer);
   }
   [_motionManager stopAccelerometerUpdates];
+  [NSNotificationCenter.defaultCenter removeObserver:self];
 }
 
 - (CVPixelBufferRef)copyPixelBuffer {
@@ -907,10 +916,10 @@ static void selectBestFormatForRequestedFrameRate(
     _isFirstVideoSample = YES;
     _isRecording = YES;
     _isRecordingPaused = NO;
-    _videoTimeOffset = CMTimeMake(0, 1);
-    _audioTimeOffset = CMTimeMake(0, 1);
-    _videoIsDisconnected = NO;
-    _audioIsDisconnected = NO;
+    _isRecordingDisconnected = NO;
+    _recordingTimeOffset = CMTimeMake(0, 1);
+    _outputForOffsetAdjusting = _captureVideoOutput;
+    _lastAppendedVideoSampleTime = kCMTimeNegativeInfinity;
     completion(nil);
   } else {
     completion([FlutterError errorWithCode:@"Error"
@@ -950,8 +959,7 @@ static void selectBestFormatForRequestedFrameRate(
 
 - (void)pauseVideoRecording {
   _isRecordingPaused = YES;
-  _videoIsDisconnected = YES;
-  _audioIsDisconnected = YES;
+  _isRecordingDisconnected = YES;
 }
 
 - (void)resumeVideoRecording {
--- a/packages/camera/camera/lib/src/camera_controller.dart
+++ b/packages/camera/camera/lib/src/camera_controller.dart
@@ -174,7 +174,7 @@ class CameraValue {
   }) {
     return CameraValue(
       isInitialized: isInitialized ?? this.isInitialized,
-      errorDescription: errorDescription,
+      errorDescription: errorDescription ?? this.errorDescription,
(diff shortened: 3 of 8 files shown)


packages/camera/camera_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.9.18+14
+
+* Handle video and audio interruptions and errors.
+* Use a single time offset for both video and audio.
```
hellohuanlin, line 4: ... to fix a bug where audio and video are out of sync after interrupted by phone cals.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -278,9 +275,33 @@ - (instancetype)initWithConfiguration:(nonnull FLTCamConfiguration *)configurati
...
+  // https://github.com/flutter/flutter/issues/151253
+  for (NSObject<FLTCaptureSession> *session in @[ _videoCaptureSession, _audioCaptureSession ]) {
+    [NSNotificationCenter.defaultCenter addObserver:self
+                                           selector:@selector(captureSessionWasInterrupted:)
+                                               name:AVCaptureSessionWasInterruptedNotification
+                                             object:session.captureSession];
+
+    [NSNotificationCenter.defaultCenter addObserver:self
+                                           selector:@selector(captureSessionRuntimeError:)
+                                               name:AVCaptureSessionRuntimeErrorNotification
```
hellohuanlin, line 288: when will run time error happen?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -788,10 +803,8 @@ - (void)newVideoSample:(CMSampleBufferRef)sampleBuffer {
     }
     return;
   }
-  if (_videoWriterInput.readyForMoreMediaData) {
```
hellohuanlin, line 791: this is moved to top right?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -88,10 +85,10 @@ @interface FLTCam () <AVCaptureVideoDataOutputSampleBufferDelegate,
...
 
 @property(assign, nonatomic) UIDeviceOrientation lockedCaptureOrientation;
-@property(assign, nonatomic) CMTime lastVideoSampleTime;
-@property(assign, nonatomic) CMTime lastAudioSampleTime;
-@property(assign, nonatomic) CMTime videoTimeOffset;
-@property(assign, nonatomic) CMTime audioTimeOffset;
+@property(assign, nonatomic) CMTime recordingTimeOffset;
+@property(assign, nonatomic) CMTime lastSampleEndTime;
+@property(nonatomic) AVCaptureOutput *outputForOffsetAdjusting;
+@property(assign, nonatomic) CMTime lastAppendedVideoSampleTime;
```
hellohuanlin, line 91: is `lastAppendedVideoSampleTime` the same as `lastSampleEndTime`? 

Can you add some doc on these 4 properties?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -88,10 +85,10 @@ @interface FLTCam () <AVCaptureVideoDataOutputSampleBufferDelegate,
 @property(assign, nonatomic) int maxStreamingPendingFramesCount;
 
 @property(assign, nonatomic) UIDeviceOrientation lockedCaptureOrientation;
-@property(assign, nonatomic) CMTime lastVideoSampleTime;
-@property(assign, nonatomic) CMTime lastAudioSampleTime;
-@property(assign, nonatomic) CMTime videoTimeOffset;
-@property(assign, nonatomic) CMTime audioTimeOffset;
+@property(assign, nonatomic) CMTime recordingTimeOffset;
+@property(assign, nonatomic) CMTime lastSampleEndTime;
```
hellohuanlin, line 89: am i reading it right that 

(1) `recordingTimeOffset` covers `videoTimeOffset` and `audioTimeOffset` and 

(2) `lastSampleEndTime` covers `lastVideoSampleTime` and `lastAudioSampleTime`?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
```diff
@@ -88,10 +85,10 @@ @interface FLTCam () <AVCaptureVideoDataOutputSampleBufferDelegate,
...
+@property(nonatomic) AVCaptureOutput *outputForOffsetAdjusting;
```
hellohuanlin, line 90: we didn't have this output. can this be just a local variable?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
hellohuanlin, line 91: is `lastAppendedVideoSampleTime` the same as `lastSampleEndTime`? 

Can you add some doc on these 4 properties?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: As is in the description "for example when there is an incoming phone call and the camera starts recording with audio". When `startVideoRecording` is called with enabled audio while there is an incoming call (does not need to be answered in my case, ringing is enough).

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: Yes, `lastAudioSampleTime` was actually the end time of a sample so I renamed it accordingly. And `lastVideoSampleTime` is practically end time too, according to logic for offset calculation both before and after this PR, it just happens that video samples tend to not have duration which can be interpreted as zero duration in this case so end time is same as start time. For that logic to work right this needs to be end time of sample and not start time.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: `lastAppendedVideoSampleTime` is related to this text from the description "Avoid errors when trying to append samples with timestamp less or equal to the previous sample.". `appendPixelBuffer` throws an error if a sample with specified sample time (`PresentationTime`) is appended after another sample with higher or equal time was already appended before (and maybe this should be end time rather than start time too).

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
misos1: That output is either audio or video output depending on whether audio or video output should be used for offset adjusting.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/FLTCam.m
hellohuanlin: can you put some comments there?

stuartmorgan-g:
From triage: what's the status of this PR? It's not clear to me if it's waiting for updates, or waiting for re-review.

misos1:
It is waiting for updates.

misos1:
Test `testImageStreamEventFormat` is failing locally with `XCTAssertEqual(camera.isStreamingImages, true)` `XCTAssertEqual failed: ("false") is not equal to ("true")`. Seems `waitForQueueRoundTrip` is "waiting" for the wrong queue as `isStreamingImages` is changed on `captureSessionQueue`. But even with the right queue it cannot wait on something which was not already dispatched at the time `waitForQueueRoundTrip` was called, and the block where that variable changes may not be. There should be a better way to wait for some variable to change in xc test.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation_objc/include/camera_avfoundation/FLTCam.h
```diff
@@ -42,10 +42,14 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property(strong, nonatomic, nullable) NSObject<FLTAssetWriter> *videoWriter;
-@property(assign, nonatomic) BOOL videoIsDisconnected;
-@property(assign, nonatomic) BOOL audioIsDisconnected;
-@property(assign, nonatomic) CMTime videoTimeOffset;
-@property(assign, nonatomic) CMTime audioTimeOffset;
+/// Whether the recording is disconnected.
+@property(assign, nonatomic) BOOL isRecordingDisconnected;
+/// Represents sum of all recording pauses/interruptions.
+@property(assign, nonatomic) CMTime recordingTimeOffset;
+/// Output to use for adjusting of recording offset.
```
hellohuanlin: can you add comment on what "adjusting of recording offset" means and why we need to adjust the offset?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -288,75 +288,74 @@ final class DefaultCamera: FLTCam, Camera {
...
+        if !(videoWriterInput?.readyForMoreMediaData ?? false) {
+          return
+        }
+      } else {
+        // ignore audio samples until the first video sample arrives to avoid black frames
+        // https://github.com/flutter/flutter/issues/57831
+        if isFirstVideoSample || !(audioWriterInput?.readyForMoreMediaData ?? false) {
+          return
+        }
+        outputForOffsetAdjusting = output
```
hellohuanlin, line 1231: can you add comment on why using the audio output as the `outputForOffsetAdjusting`

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -405,10 +404,8 @@ final class DefaultCamera: FLTCam, Camera {
       }
       return
     }
-    if audioWriterInput?.readyForMoreMediaData ?? false {
-      if !(audioWriterInput?.append(sampleBuffer) ?? false) {
-        reportErrorMessage("Unable to write to audio input")
-      }
+    if !(audioWriterInput?.append(sampleBuffer) ?? false) {
+      reportErrorMessage("Unable to write to audio input")
```
hellohuanlin, line 1403: i may need some help understanding this change. Can you walk me thought this change?

misos1: That check was moved elsewhere (line 308).

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation_objc/include/camera_avfoundation/FLTCam.h
misos1: It should have been "adjusting of recording time offset" -> `recordingTimeOffset` which as mentioned is sum of all pauses in recording so adjusting it of course means that if there is another pause then add duration of that pause into that offset. So after that offset is applied then we get continuous recording.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
misos1: This is because seems that audio samples (except the first) cannot be really moved in time meaning that `audioWriterInput.append` accepts sample with presentation time but seems ignores that time (except for the first) and just puts next sample at time of end of previous sample so they are always continuous. I suspect that is because these audio samples are stored in a video file in a single big chunk with just single presentation time. So if audio times cannot really be adjusted then video times need to be flexed according to audio times. I cannot be really sure that this is how it works on every platform and device so in case individual audio samples can be adjusted somewhere then it should be irrelevant whether audio or video is used for such adjusting. Thus audio output is preferred for adjusting.

stuartmorgan-g:
From triage: Ping @hellohuanlin on this review.

hellohuanlin:
@misos1 I took another look at the PR, and still feel hard to fully digest the change. It sounds like you are trying to do multiple things at once: 

1. handling phone call interrupts where we can pause the recording
2. handling system errors that's not recoverable
3. fixing audio/video sync

I'm wondering if we can split it into separate PRs so it's easier to review.

stuartmorgan-g:
From triage: @misos1 Are you planning on updating this PR based on the feedback above?

misos1:
@hellohuanlin Points 1. and 2. are just different instantiations of the same problem. It has just different manifestations. In one case it sends interruption and in another case it sends an error. Point 3. is an essential part of the fix. Without it would be useless to disconnect as it would not work like pause for both audio and video so audio would be shifted.

okorohelijah:
@hellohuanlin review from ios-triage

hellohuanlin:
@misos1 To make review easier, could you update your PR description? I have some trouble following it. Please also explain what the bug is. 

> Handle interruptions of capture sessions by doing "internal pause". 

What is "internal pause"

> Another possibility would be to remove audio input from the session but this would be too specific for interruptions caused by an incoming call or alarm. 

Why do we need to remove audio input? When to do it? why is it too specific? 

> Or to stop video recording as the native ios camera app does but this would require some additional event.

what additional event? 

> Handle asynchronous error notifications posted about capture sessions. This can happen for example when there is an incoming phone call and the camera starts recording with audio. 

start recording before or after incoming phone call? 

> For this specific case would be enough to just call startRunning on capture sessions but that is too specific and would require ability to restart capture session from dart side as seems there is no good way how to test when such phone call or other such event ends, and trying to restart it for example every second inside FLTCam does not look like good idea. 

Could you split this long sentence so it's easier to understand? 
startRunning on audio or video? Why requires ability to restart form dart side? 

> Instead the camera controller enters an error state and can be recovered by disposing and recreating which should be more generic.

Is this what you plan to do or what the current state is? 

> Use single offset for both video and audio as separate offsets can cause unsynchronized video with audio. 

Is this what you plan to do or what the current state is? 

> Seems audio timestamps tend to be ignored except for the first sample so prefer them for calculating offset to avoid desynchronization. 

Who ignores the audio timestamps? flutter or apple? What does "them" refer to? the "audio samples"? or the "first sample"? 

> Avoid errors when trying to append samples with timestamp less or equal to the previous sample. 

How to avoid? 

> Fix a bug when after adjusting offset also last sample time needs to be updated to avoid adding past offsets multiple times.

What bug is fixed?

LouiseHsu:
@misos1 Do you think you will be able to make the requested changes? :)

misos1:
@hellohuanlin

> > Handle interruptions of capture sessions by doing "internal pause".
> 
> What is "internal pause"

That meant to not process any video and audio samples like when the camera controller paused.

> > Another possibility would be to remove audio input from the session but this would be too specific for interruptions caused by an incoming call or alarm.
> 
> Why do we need to remove audio input? When to do it? why is it too specific?

Removing audio input during a call would enable the camera to continue without audio. But this is specific to calls and alarms. For example what if interruption is due to something that wants exclusive access to the device camera? I did not see any simple way to detect whether it will be enough to remove audio input or not (ios just interrupts all your sessions if I remember correctly). Doing "pause" by ignoring both audio and video until interruption ends is a more generic solution.

> > Or to stop video recording as the native ios camera app does but this would require some additional event.
> 
> what additional event?

Event sent to dart side to inform it that video recording was stopped. So the dart side can handle that and somehow retrieve the resulting prematurely finished video file.

> > For this specific case would be enough to just call startRunning on capture sessions but that is too specific and would require ability to restart capture session from dart side as seems there is no good way how to test when such phone call or other such event ends, and trying to restart it for example every second inside FLTCam does not look like good idea.
> 
> startRunning on audio or video?

Probably startRunning on both video and audio as I mentioned if I remember correctly it interrupts all sessions.

> Why requires ability to restart form dart side?

Because I did not find any sensible way to detect when it is "safe" to start sessions again. As I wrote it could try to do it every second until it succeeds but that looks like a bad solution. Instead maybe it is better to let users handle asynchronous errors from the dart side who can for example recreate the camera controller after clicking on a button or whatever looks like the best solution for a specific app.

> > Instead the camera controller enters an error state and can be recovered by disposing and recreating which should be more generic.
> 
> Is this what you plan to do or what the current state is?
> > Use single offset for both video and audio as separate offsets can cause unsynchronized video with audio.
> 
> Is this what you plan to do or what the current state is?

Current state both.

The rest is in updated description (I hope).

packages/camera/camera_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.9.20+2
+
+* Handles video and audio interruptions and errors.
```
hellohuanlin: can you be more explicit about how it's handled (pause video/audio when interrupted)?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation_objc/FLTCam.m
```diff
@@ -444,6 +470,7 @@ - (BOOL)setCaptureSessionPreset:(FCPPlatformResolutionPreset)resolutionPreset
 
 - (void)dealloc {
   [_motionManager stopAccelerometerUpdates];
+  [NSNotificationCenter.defaultCenter removeObserver:self];
```
hellohuanlin: nit: this can be omitted iirc

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -288,75 +288,74 @@ final class DefaultCamera: FLTCam, Camera {
...
-      if isFirstVideoSample && output != captureVideoOutput.avOutput {
-        return
+      // do not append sample buffer when readyForMoreMediaData is NO to avoid crash
+      // https://github.com/flutter/flutter/issues/132073
+      if output == captureVideoOutput.avOutput {
+        if !(videoWriterInput?.readyForMoreMediaData ?? false) {
+          return
+        }
+      } else {
+        // ignore audio samples until the first video sample arrives to avoid black frames
```
hellohuanlin, line 1226: nit: can you add a comment saying that this is because video frames will be paused when interrupted (e.g. phone call)

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -288,75 +288,74 @@ final class DefaultCamera: FLTCam, Camera {
...
-      if output == captureVideoOutput.avOutput {
-        if videoIsDisconnected {
-          videoIsDisconnected = false
-
-          videoTimeOffset =
-            videoTimeOffset.value == 0
-            ? CMTimeSubtract(currentSampleTime, lastVideoSampleTime)
-            : CMTimeAdd(videoTimeOffset, CMTimeSubtract(currentSampleTime, lastVideoSampleTime))
+      var currentSampleEndTime = sampleTime
+      let dur = CMSampleBufferGetDuration(sampleBuffer)
```
hellohuanlin: nit: `frameDuration`

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -288,75 +288,74 @@ final class DefaultCamera: FLTCam, Camera {
...
+      if CMTIME_IS_NUMERIC(dur) {
+        currentSampleEndTime = CMTimeAdd(currentSampleEndTime, dur)
+      }
 
-          return
+      // Use a single time offset for both video and audio.
```
hellohuanlin: nit: Use shared offset to avoid out of sync video and audio.

hellohuanlin:
@misos1 are you still working on updating this PR? Thanks

misos1:
@hellohuanlin yes

hellohuanlin:
FYI this likely conflict with the ongoing swift migration, and you may have to migrate this PR too.

hellohuanlin:
@misos1 from triage: thank you soooooooo much for working on this. we are making this a draft. Please re-open when it's ready for another review.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
misos1: This part is not about interruptions. This is from #4587.

misos1:
> we are making this a draft. Please re-open when it's ready for another review.

Funny, I have been resolving conflicts of this for almost a year now.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -233,6 +237,33 @@ final class DefaultCamera: NSObject, Camera {
...
+        selector: #selector(captureSessionWasInterrupted),
+        name: AVCaptureSession.wasInterruptedNotification,
+        object: session)
+
+      NotificationCenter.default.addObserver(
+        self,
+        selector: #selector(captureSessionRuntimeError),
+        name: AVCaptureSession.runtimeErrorNotification,
+        object: session)
+    }
```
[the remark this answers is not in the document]
misos1: This changed from ios 9, no?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation_objc/FLTCam.m
misos1: https://developer.apple.com/documentation/foundation/notificationcenter/removeobserver(_:name:object:)

"If your app targets iOS 9.0 and later or macOS 10.11 and later, and you used [addObserver(_:selector:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/addobserver(_:selector:name:object:)) to create your observer, you do not need to unregister the observer. If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it."

But maybe it would be better to add it back because what if it takes a very long time until something is posted again? Although I think I read somewhere that this is done either as a weak reference or it is detected differently that object was deleted and in case of weak reference this should not matter.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -233,6 +237,33 @@ final class DefaultCamera: NSObject, Camera {
...
+  }
+
+  @objc private func captureSessionWasInterrupted(notification: NSNotification) {
+    isRecordingDisconnected = true
+  }
+
+  @objc private func captureSessionRuntimeError(notification: NSNotification) {
+    reportErrorMessage(
+      "\(String(describing: notification.userInfo?[AVCaptureSessionErrorKey] as? Error))")
```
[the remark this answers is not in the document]
misos1: @hellohuanlin I saw that "Optional(...)" string but I do not care a lot. To me a lot of error messages here look not very user-friendly. To me these errors from flutter or flutter plugins are more like debug messages for me. I would rather display something else to the actual user in a lot of cases. So should I proceed?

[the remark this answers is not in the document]
hellohuanlin: Gemini probably assumed this message is displayed to real users. But we actual send it to developers, so unlocalized message should be fine.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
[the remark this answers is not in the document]
stuartmorgan-g: The behavior changed in iOS 9, yes, but also everything said here was wrong anyway.

> This will cause a memory leak

It won't, and never did. The behavior pre-9 was a crash, not a leak.

> because NotificationCenter holds a strong reference to the observer (self).

It doesn't, and never did.

> Please make sure to remove the observers when the camera is closed, for example in the `deinit` or `close` method

If the retain cycle it is describing were real, adding cleanup in `deinit` would do nothing, because by definition a retain cycle leak means that `deinit` won't be called.

stuartmorgan-g:
@LouiseHsu Is this on your radar for secondary review?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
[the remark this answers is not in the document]
misos1: https://developer.apple.com/documentation/foundation/notificationcenter/removeobserver(_:name:object:)

"If your app targets iOS 9.0 and later or macOS 10.11 and later, and you used [addObserver(_:selector:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/addobserver(_:selector:name:object:)) to create your observer, you do not need to unregister the observer. If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it."

But maybe it would be better to add it back anyway because what if it takes a very long time until something is posted again? Although I think I read somewhere that this is done either as a weak reference or it is detected differently that object was deleted and in case of weak reference this should not matter.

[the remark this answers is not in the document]
misos1: Oh never mind. As I think about it now there is no way that the object would be not deallocated until something is posted again. I think what they mean by "the system cleans up" is just removing the invalid weak reference or something. Otherwise removing the observer from deinit would not work, as you mentioned.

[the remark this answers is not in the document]
stuartmorgan-g: > think what they mean by "the system cleans up" is just removing the invalid weak reference or something.

Yes, I've always read that as being weirdly specific documentation of the internal pointer management implementation.

stuartmorgan-g:
It looks like this has the necessary approvals; let's get it landed before anything else can conflict with it :)

PR #8988, opened 2025-04-03
[camera_avfoundation] Implementation swift migration

RobertOdrowaz:
Migrates camera implementation as part of https://github.com/flutter/flutter/issues/119109

~This PR includes the `CameraPlugin` class migration. Some additional changes regarding imports were necessary to keep Swift Package Manager compatibility. SPM does not support mixed languages in a single target. I assume there might be a better solution so I'm open to suggestions~

After splitting only a new target for SPM is created in this PR and the objc implementation is moved to a dedicated directory. 

During the plugin's migration from ObjC to Swift, there will be a transitional period during which the implementation will be mixed, with some parts already migrated to Swift and some still in ObjC. This is somewhat problematic with SwiftPM. SwiftPM does support packages with mixed language implementation but different languages have to be in separate targets. Every target defines a separate module which means that in some cases both have to be imported. However, this only applies to SwiftPM with Cocoapods we can mix Swift and ObjC code in the same pod and have a single module. Simmingly the best solution is to use conditional imports for the additional module when SwiftPM during the migration.
```swift
@testable import camera_avfoundation

#if canImport(camera_avfoundation_objc)
  @testable import camera_avfoundation_objc
#endif
```

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/camera/camera_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.9.18+14
+
+* Migrates the CameraPlugin class to Swift.
+* Splits the implementation into Objective-C and Swift targets for the Swift Package Manager.
```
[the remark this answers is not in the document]
RobertOdrowaz: Done although I couldn't make a completely empty target because with no source compilation fails on missing `camera_avfoundation.o` file

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation_objc/QueueUtils.m
```diff
@@ -13,3 +13,7 @@ void FLTEnsureToRunOnMainQueue(dispatch_block_t block) {
     block();
   }
 }
+
+void FLTDispatchQueueSetSpecific(dispatch_queue_t queue, const void *key) {
+  dispatch_queue_set_specific(queue, key, (void *)key, NULL);
```
hellohuanlin, line 18: revert this?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation_objc/include/camera_avfoundation/QueueUtils.h
```diff
@@ -16,4 +16,13 @@ extern const char* FLTCaptureSessionQueueSpecific;
 /// @param block the block to be run on the main queue.
 extern void FLTEnsureToRunOnMainQueue(dispatch_block_t block);
 
+/// Calls `dispatch_queue_set_specific` with a key that is used to identify the
+/// queue. This method is needed for compatibility of Swift tests with
+/// Objective-C code. In Swift, the API for setting key-value pairs on a queue
+/// is different, so Swift tests need to call this method to set the key-value
+/// pair on the queue in a way that's compatible with the existing Objective-C
+/// code.
+extern void FLTDispatchQueueSetSpecific(dispatch_queue_t queue,
```
hellohuanlin, line 25: revert?

packages/camera/camera_avfoundation/example/ios/RunnerTests/AvailableCamerasTests.swift
```diff
@@ -7,6 +7,10 @@ import XCTest
 
 @testable import camera_avfoundation
 
+#if canImport(camera_avfoundation_objc)
```
hellohuanlin, line 10: why is this check required and when will it be false?

RobertOdrowaz: It will be false when cocoapods are used. From the pods perspective, there is still a single `camera_avfoundation` module containing both swift and objc code

RobertOdrowaz: Of course, this is temporary once everything is migrated to Swift we will have a single SwiftPM target and those conditional imports will no longer be needed

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation_objc/QueueUtils.m
RobertOdrowaz: I removed it from this PR although I don't think it was critical

hellohuanlin: Do we still need at least 1 swift file for the swift target?

RobertOdrowaz: I don't see the relation between this change and the Swift placeholder file. Before this was just moved from `QueueTestUtils` to `QueueUtils` because will be needed outside of tests (during the transition when `CameraPlugin` is in Swift and `FLTCam` is still in ObjC)

packages/camera/camera_avfoundation/ios/camera_avfoundation.podspec
```diff
@@ -13,9 +13,13 @@ A Flutter plugin to use the camera from your Flutter app.
   s.author           = { 'Flutter Dev Team' => 'flutter-dev@googlegroups.com' }
   s.source           = { :http => 'https://github.com/flutter/packages/tree/main/packages/camera_avfoundation' }
   s.documentation_url = 'https://pub.dev/packages/camera_avfoundation'
-  s.source_files = 'camera_avfoundation/Sources/camera_avfoundation/**/*.{h,m}'
-  s.public_header_files = 'camera_avfoundation/Sources/camera_avfoundation/include/**/*.h'
-  s.module_map = 'camera_avfoundation/Sources/camera_avfoundation/include/CameraPlugin.modulemap'
+  s.source_files = 'camera_avfoundation/Sources/camera_avfoundation*/**/*.{h,m,swift}'
```
hellohuanlin, line 16: can you add a comment here to explain why we need to handle 2 folders (the `camera_avfoundation*` part).

packages/camera/camera_avfoundation/ios/camera_avfoundation.podspec
```diff
@@ -13,9 +13,13 @@ A Flutter plugin to use the camera from your Flutter app.
   s.author           = { 'Flutter Dev Team' => 'flutter-dev@googlegroups.com' }
   s.source           = { :http => 'https://github.com/flutter/packages/tree/main/packages/camera_avfoundation' }
   s.documentation_url = 'https://pub.dev/packages/camera_avfoundation'
-  s.source_files = 'camera_avfoundation/Sources/camera_avfoundation/**/*.{h,m}'
-  s.public_header_files = 'camera_avfoundation/Sources/camera_avfoundation/include/**/*.h'
-  s.module_map = 'camera_avfoundation/Sources/camera_avfoundation/include/CameraPlugin.modulemap'
```
hellohuanlin, line 18: it's been a while since i touch this. Why we are removing the module map?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Package.swift
```diff
@@ -12,20 +12,29 @@ let package = Package(
...
+      dependencies: ["camera_avfoundation_objc"],
+      path: "Sources/camera_avfoundation",
+      resources: [
+        .process("Resources")
+      ]
+    ),
+    .target(
+      name: "camera_avfoundation_objc",
       dependencies: [],
-      exclude: ["include/camera_avfoundation-umbrella.h", "include/CameraPlugin.modulemap"],
```
hellohuanlin, line 22: same question to the umbrella header.

packages/camera/camera_avfoundation/ios/camera_avfoundation.podspec
RobertOdrowaz: It's not supported with Swift static libraries.
```
[!] Using Swift static libraries with custom module maps is currently not supported. Please build `camera_avfoundation` as a framework or remove the custom module map.
```

It's still the same issue you've encountered here https://github.com/flutter/plugins/pull/6369

packages/camera/camera_avfoundation/ios/camera_avfoundation/Package.swift
RobertOdrowaz: Correct me if I'm missing something but I don't see much value in having a custom umbrella header without a custom module map. Without the module map cocoapods generate the public umbrella header that includes everything either way

RobertOdrowaz:
> I'm confused, the ObjC target doesn't define a "swift" module right? isn't it still a clang module?
Yes, there won't be a swift module for the ObjC target. I've corrected the description

packages/camera/camera_avfoundation/ios/camera_avfoundation.podspec
hellohuanlin: Oh i completely forgot about this! thanks for the reference.

hellohuanlin (approved):

PR #8998, opened 2025-04-04
[camera_android] prevent startImageStream OOM error when main thread paused (flutter#166533)

kwikwag:
*I have yet not written a test for this.* I am unfamiliar with the testing system, but I am in the process of trying to write one. I thought it'd be good to put this out there first, especially if there are comments on the solution itself. Once I write a test I will edit the pull request. Meanwhile, perhaps someone else more familiar with the system might contribute a test, and that would probably be better than the test I'll come up with.

When streaming images using CameraController.startImageStream(), images are being post()-ed to the main looper even if it's halted. This is common when debugging. This quickly results in an OOM and the app crashes abruptly.

The fix is done by counting the number of images that have been posted to the main thread but have not yet arrived. If too many images are in transit (I arbitrarily chose 3 as the maximum allowable amount) subsequent frames are dropped, until the main thread receives the pending images.

A log message is emitted for each dropped frame.

Fixes flutter/flutter#166533

That issue also provides a demo program.

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/media/ImageStreamReader.java
+++ b/packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/media/ImageStreamReader.java
@@ -9,6 +9,7 @@ import android.media.Image;
 import android.media.ImageReader;
 import android.os.Handler;
 import android.os.Looper;
+import android.util.Log;
 import android.view.Surface;
 import androidx.annotation.NonNull;
 import androidx.annotation.VisibleForTesting;
@@ -22,6 +23,7 @@ import java.util.Map;
 
 // Wraps an ImageReader to allow for testing of the image handler.
 public class ImageStreamReader {
+  private static final String TAG = "ImageStreamReader";
 
   /**
    * The image format we are going to send back to dart. Usually it's the same as streamImageFormat
@@ -32,6 +34,13 @@ public class ImageStreamReader {
 
   private final ImageReader imageReader;
   private final ImageStreamReaderUtils imageStreamReaderUtils;
+
+  /**
+   * This solves a memory issue, when the main thread hangs (e.g. when pausing the debugger) while
+   * many frames are being sent using android.os.Handler.post(). This causes an OOM error rather
+   * quickly. So to avoid this we apply some back-pressure.
+   */
+  private static int numImagesInTransit = 0;
 
   /**
    * Creates a new instance of the {@link ImageStreamReader}.
@@ -96,6 +105,13 @@ public class ImageStreamReader {
       @NonNull CameraCaptureProperties captureProps,
       @NonNull EventChannel.EventSink imageStreamSink) {
     try {
+      // The limit was chosen so it would not drop frames for reasonable lags of the main thread.
+      if (numImagesInTransit > 2) {
+        Log.d(TAG, "Dropping frame due to images pending on main thread.");
+        image.close();
+        return;
+      }
+
       Map<String, Object> imageBuffer = new HashMap<>();
 
       // Get plane data ready
@@ -115,11 +131,19 @@ public class ImageStreamReader {
           "sensorSensitivity", sensorSensitivity == null ? null : (double) sensorSensitivity);
 
       final Handler handler = new Handler(Looper.getMainLooper());
-      handler.post(() -> imageStreamSink.success(imageBuffer));
+      ++numImagesInTransit;
+      boolean postResult =
+          handler.post(
+              () -> {
+                imageStreamSink.success(imageBuffer);
+                --numImagesInTransit;
+              });
+
       image.close();
 
     } catch (IllegalStateException e) {
-      // Handle "buffer is inaccessible" errors that can happen on some devices from ImageStreamReaderUtils.yuv420ThreePlanesToNV21()
+      // Handle "buffer is inaccessible" errors that can happen on some devices from
+      // ImageStreamReaderUtils.yuv420ThreePlanesToNV21()
       final Handler handler = new Handler(Looper.getMainLooper());
       handler.post(
           () ->
--- a/packages/camera/camera_android/CHANGELOG.md
+++ b/packages/camera/camera_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.10+3
+
+* Fix flutter#166533 - prevent startImageStream OOM error when main thread paused.
+
 ## 0.10.10+2
 
 * Don't set the FPS range unless video recording. It can cause dark image previews on some devices becuse the auto exposure algorithm is more constrained after fixing the min/max FPS range to the same value. This change has the side effect that providing the `fps` parameter will not affect the camera preview when not video recording. And if you need a lower frame rate in your image streaming handler, you can skip frames by checking the time it passed since the last frame.
--- a/packages/camera/camera_android/pubspec.yaml
+++ b/packages/camera/camera_android/pubspec.yaml
@@ -3,7 +3,7 @@ description: Android implementation of the camera plugin.
 repository: https://github.com/flutter/packages/tree/main/packages/camera/camera_android
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+camera%22
 
-version: 0.10.10+2
+version: 0.10.10+3
 
 environment:
   sdk: ^3.6.0


reidbaker:
Sorry @camsim99 is out of town. Adding team android to give a first pass review.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/media/ImageStreamReader.java
```diff
@@ -114,13 +136,23 @@ public void onImageAvailable(
...
     } catch (IllegalStateException e) {
-      // Handle "buffer is inaccessible" errors that can happen on some devices from ImageStreamReaderUtils.yuv420ThreePlanesToNV21()
-      final Handler handler = new Handler(Looper.getMainLooper());
+      // Handle "buffer is inaccessible" errors that can happen on some devices from
+      // ImageStreamReaderUtils.yuv420ThreePlanesToNV21()
+      final Handler handler =
+          this.handler != null ? this.handler : new Handler(Looper.getMainLooper());
       handler.post(
           () ->
               imageStreamSink.error(
```
reidbaker, line 158: Shouldnt this also have `--numImagesInTransit` otherwise if we had more than 3 images error then we would no longer update any frames.

reidbaker (commented):
Thank you this is a tricky problem and clearly impacts developers. 

I am a bit worried that we are dropping frames instead of creating a frame queue but I understand that is probably a more difficult pr to author. 

My largest worry is that this will cause an issue in non debug code. What do you think about renaming `numImagesInTransit` to `consecutiveImagesInTranset` and when any frame is delivered setting the value to zero. That would solve the debugger paused execution issue by dropping frames but remove an off by one set of conditions from causing the camera frames to stop. 

This will still need @camsim99 to approve before we merge it.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/media/ImageStreamReader.java
kwikwag: @reidbaker addressing the issues one-by-one

1. As opposed to how I initially phrased the issue, this will not only happen when the debugger is paused, but any time the main loop is lagging. I see no benefit to the 'consecutive' method:

  * I'm not sure what you mean by an "off-by-one set of conditions." If the main thread is lagging by one frame, it will do so regardless of this fix. Dropping frames will only happen when at least two frames have already been sent to the handler but have not yet been handled. Being less familiar with Flutter/Android workings, I'm not sure if the `EventSink.success()` call can raise an exception. If it can, it would perhaps be wish to wrap the call with a `try..finally`. However, I would not move the decrement before the `success()` call, as decrement should happen when the image memory can readily be released, otherwise a queue build-up might still occur.
  * Resetting the number of frames whenever any frame reaches the main thread can have an adverse affect, too. Imagine a scenario where a frame is generated every 10ms, but handled every 20ms on the main thread itself. You will start accumulating an infinite amount of frames pending on `post()` and eventually crash the app with an out-of-memory error. Ensuring that `post()` arrives and completes is exactly the queue capability you are talking about, as far as I understand.

2. Regarding not updating frames anymore - the only scenario I see that you will no longer update frames in, is when the handler stops handling the `Runnable`s passed to `post()`. If that is the case, it means the main thread is paused or hanging, in which case it should be OK to not update the frames until it releases and the `Runnable` gets invoked. I see no risk of deadlock, as AFAIK there is no way for the main thread to wait on whatever calls `onImageAvailable()`.

 How long will it be before @camsim99 comes back?

kwikwag:
I can also suggest the following alternative implementations:

  1. Make `imageBuffer` an instance variable, and keeping only one `post()` invocation alive (e.g. by testing first with `Handler.hasCallbacks()`. This will eliminate the queue count and will only provide the latest frame to the handler.
  2. Using `Handler.sendMessage()` and `Handler.removeMessages()` to prevent the queue from building up.

kwikwag:
Yet another suggestion: keep image data as a soft/weak reference and skip frames where memory is evicted by the runtime. Only keep a hard reference to the most recent frame. Then we don't need this special case that counts the messages pending on the looper. I am not sure whether this will in fact eliminate the out of memory error, but I can check if this is a requirement.

packages/camera/camera_android/android/src/test/java/io/flutter/plugins/camera/media/ImageStreamReaderUtilsTest.java
```diff
@@ -25,52 +22,10 @@ public void setUp() {
...
-    when(mockImage.getPlanes()).thenReturn(planes);
-
-    return mockImage;
-  }
-
   /** Ensure that passing in an image with padding returns one without padding */
   @Test
   public void yuv420ThreePlanesToNV21_trimsPaddingWhenPresent() {
-    Image mockImage = getImage(160, 120, 16);
+    Image mockImage = ImageStreamReaderTest.getImage(160, 120, 16, ImageFormat.YUV_420_888);
```
camsim99, line 28: Nit: seems non-intuitive for test files to call between each other. Can you create an `ImageStreamReaderTestUtils.java` or something like that and put it there?

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/media/ImageStreamReader.java
camsim99: My two cents:

- @reidbaker can you clarify the off-by-one issue? I'm also a bit confused there.
- Based on my current understanding of things (off-by-one issue aside), I agree with @kwikwag that we should not reset `numImagesInTransit` back to 0 when a frame is delivered. My understanding is that if we do, then we would no longer only be waiting on 3 images to accumulate to start closing images, but instead, would be waiting on 3 then 6 then potentially 9 and so on and so forth.
- I agree with @reidbaker that in this case the comment is linked to (the case where an `IllegalStateException` is thrown streaming an image), we should also decrement `numImagesInTransit` because we have essentially "handled" that image.

kwikwag:
@camsim99 Thank you for the review and the comments.

I added fixes re the two issues mentioned. Note that the naming of the added file `ImageStreamReaderTestUtils` might be a bit confusing as the same folder holds `ImageStreamReaderUtilsTest`...

Rather than decreasing the `numImagesInTransit` in two places, I just moved the exception handling to surround only relevant code. Hope that's good.

Finally, I also added an [extra branch with the `WeakReference` solution](https://github.com/kwikwag/flutter_packages/tree/fix/issue_166533_alt2). It seems to work too, and doesn't require a counter. However since the unit test requires to check what happens when a `WeakReference` is released, the `Runnable` code there is a bit more complex. If you think that's better, I can create a pull request for that instead.

camsim99:
@kwikwag Thanks for the updates! The changes look reasonable to me 👍

Concerning the `WeakReference` solutions, what is the main difference there? I know it doesn't use the counter, but it's still present in that code, so I'm wondering how images would be closed differnetly?

kwikwag:
@camsim99 I'm not sure focusing on the `image.close()` call is a beneficial way to look at it - that happens in the thread that has `onAvailable` invoked. 

In the `WeakReference` solution, the idea is that the thread keeps on `post()`ing messages, and delegate memory management considerations to the runtime. If the messages reach the other side while the runtime has still kept the frames, that reference is made hard and passed downstream. It does mean that there is some extra processing done when when the main thread is blocked (copying planes from the `image` object) as opposed to the explicit back-pressure mechanism in this pull request. It also means messages are continued to be sent to the main thread, which could theoretically cause an OOM, but maybe only after a very very very long time.

The only main pro on using the `WeakReference` solution is that it might be applicable in other places, where messages sent to the main looper are heavy or frequent. However, there aren't really other places where messages are likely to be more frequency than 1 per second. The only two places I found that might be affected are under `webview_flutter_android` where a `JavaScriptChannel class sends `postMessage` messages, and in `WebViewProxyApi` there are `onScrollChanged` messages that perhaps might be sent a lot. I thought about doing a test for webview, but it's a bit beyond the scope of this pull request.

If the solution in this pull request is acceptable I would like to submit it. I can submit the `WeakReference` solution at a later point if this turns out to be something of a more general consequence.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/media/ImageStreamReader.java
reidbaker: My worry is that over time either changes to the behavior of android or the camera code we relyon will cause us to not decrement at the same rate we increment and cause the camera to break for users. 

The weak reference suggestion would mitigate that I think.

camsim99: @kwikwag Thank you for the clear explanation on the weak reference solution! Considering the concern of this PR changing behavior, I am for pursuing that solution instead of this one. What do you think?

kwikwag:
Hey @camsim99 any updates?

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/media/ImageStreamReader.java
camsim99: @kwikwag We're leaning towards the weak reference solution based on this discussion. What do you think?

kwikwag:
Hey @camsim99 @reidbaker please let me know if any other fix / rewrite /alternative dolution is needed. I'd hate see this problem disregarded and this work go to waste, as the camera_android plugin is still relevant to many developers.

camsim99:
Hi @kwikwag I think you missed our messages in https://github.com/flutter/packages/pull/8998#discussion_r2039692991 -- we are leaning towards proceeding with the weak reference solution!

stuartmorgan-g:
From triage: @kwikwag is the alternate approach discussed above something you are planning on pursuing here?

kwikwag:
Her @stuartmorgan-g I already have a weak reference solution ready. @camsim99  I indeed missed the comments on the thread. I'll submit it as a separate pull request.

stuartmorgan-g:
@kwikwag Should this be closed now that the other PR landed, or are there parts of this you still wanted to pursue?

kwikwag:
@stuartmorgan-g In testing in my own app *in an emulator*, I've noticed that this solution was actually more stable than the WeakReference solution. I can't swear to it, but leaving memory management entirely to the runtime might not be as wholesome as one may hope. So I'm leaving it up to you.

reidbaker:
After talking to @camsim99 we think that relying on the languages reference counting and weak references will be easier to maintain. Additionally unwinding the other pr and landing this one is a factor against landing this implementation. 

Thank you so much for being willing to author 2 prs to solve this issue we are grateful for your time and effort. If someone later finds that this solution is indeed more stable and can show it with a unit test or code sample and is willing to revert the other implementation we will choose to take the more stable options. However for now we think the right solution landed!

Closing this pr :)

PR #9025, opened 2025-04-08
[video_player_web] Fix blank first frame on iOS.

ksokolovskyi:
Fixes [First Frame of Video Once VideoPlayer Controller doesn't show in iOS Safari/Chrome](https://github.com/flutter/flutter/issues/139107)

This PR fixes the blank first frame issue by adding a `_videoElement.load()` call at the end of the `initialize` function. This call forces the browser to load media in preparation for playback (https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/load).

Additionally, this PR removes `loadedmetadata` event listener, as the `canplay` event is now reliably emitted across all platforms. Keeping `loadedmetadata` listener could result in marking the video controller as initialized before the first frame is actually rendered. This causes the loader to disappear too early, leaving a blank space where the video should be until the browser displays the first frame. Relying on `canplay` ensures the first frame is ready before we proceed.

> [!NOTE]
> It turns out that this issue is iOS-specific and was reported not only in Flutter, but in some other packages and tools:
> 1) [video-react/First frame not shown as the poster on iOS Safari](https://github.com/video-react/video-react/issues/328)
> 2) [WordPress/gutenberg/Video Block: First frame not shown as the poster on iOS Safari — Fix: Use Fragment URL](https://github.com/WordPress/gutenberg/issues/51995)

> [!TIP]
> As a temporary workaround until this PR is merged, users can append the `#t=0.001` fragment to the asset or URL `src` to force the browser to load the first frame (e.g., `assets/butterfly.mp4#t=0.001`, `https://flutter.github.io/assets-for-api-docs/assets/videos/butterfly.mp4#t=0.001`).

The following repository was used for issue reproduction and testing: https://github.com/ksokolovskyi/video_player_blank_frame_bug

| Before | After |
| :---: | :---: |
| https://video-player-blank-frame-bug.web.app | https://video-player-blank-frame-fix.web.app |
| <video src="https://github.com/user-attachments/assets/4077fcc4-10cf-4dbd-8c59-2fd824b4e4c4" /> | <video src="https://github.com/user-attachments/assets/2951af36-1569-4a5c-90dd-1d52a5d1e7f2" /> |

## Pre-Review Checklist

- [X] I signed the [CLA].


packages/video_player/video_player_web/lib/src/video_player.dart
```diff
@@ -131,6 +129,10 @@ class VideoPlayer {
     if (src != null) {
       _videoElement.src = src;
     }
+
+    // Explicitly triggers media loading in preparation for playback. Needed on
+    // iOS to ensure the first frame becomes visible before playback begins.
+    _videoElement.load();
```
ditman, line 135: Does this trigger `oncanplay` so we don't need to listen to `onloadedmetadata`?

packages/video_player/video_player_web/example/integration_test/video_player_test.dart
```diff
@@ -27,6 +29,18 @@ void main() {
         ..playsInline = false;
     });
 
+    testWidgets('initialize() calls load', (WidgetTester _) async {
+      bool loadCalled = false;
+
+      video['load'] = () {
```
ditman, line 35: Can this be `video.addEventListener('load'...`, instead of writing to the video element directly?

packages/video_player/video_player_web/example/integration_test/video_player_test.dart
```diff
@@ -222,8 +241,7 @@ void main() {
         expect(events[0].eventType, VideoEventType.initialized);
       });
 
-      // Issue: https://github.com/flutter/flutter/issues/137023
```
ditman, line 225: Are videos still working on iOS17? :)

packages/video_player/video_player_web/lib/src/video_player.dart
ditman: Yes, as stated on the video description.

ditman (commented):
Thanks for the fix! Very small comment about attaching the `load` event in tests.

packages/video_player/video_player_web/example/integration_test/video_player_test.dart
ksokolovskyi: Thanks for the suggestion!

I looked into using `video.addEventListener('load', ...)`, but the load event doesn't fire for `<video>` elements ([MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/load_event)).

Since there's no corresponding event for the `load()` method, the only reliable way I found to verify that it's called is by overriding `video['load']`, as done here.

Let me know if you'd prefer a different approach. I'm open to ideas!

packages/video_player/video_player_web/example/integration_test/video_player_test.dart
ksokolovskyi: I've just verified this by opening https://video-player-blank-frame-fix.web.app on the iOS 17 simulator (as I don't have a physical device with iOS 17 installed), and videos work well.

https://github.com/user-attachments/assets/f0333e04-04b6-4332-86a4-054b77f7059f

ditman (approved):
LGTM! Let's go!

PR #8932, opened 2025-03-29
[vector_graphics] Add auto RenderingStrategy for better performance

gaodi-sec:
### Background
Currently, `vector_graphics` requires manually selecting either raster or picture rendering mode. However, raster mode has limited flexibility, as the generated cache size may not always be appropriate for the current drawing conditions, leading to inefficient rendering. On the other hand, picture mode can introduce additional RenderPass execution overhead.

### Changes
This PR adds an auto mode to vector_graphics, which:

1. Determines the actual rendering resolution using the canvas transformation matrix, ensuring a pixel-perfect raster cache without aliasing.

2. Includes color and colorFilter effects in the raster cache. Caches are stored globally, preventing redundant creation when identical icons are used.

3. Switches to picture mode if the calculated rendering resolution exceeds 2048x2048, avoiding excessive memory usage and maintaining performance.

4. Delays raster cache creation by two frames, using picture mode for drawing in the meantime. If properties change before the cache is created, the delay is extended by another frame. This prevents excessive cache regeneration during scaling animations.

5. Staggers cache creation across frames when multiple caches need to be generated, ensuring that only one cache is created per frame. For example, if three icons require caching, they will be generated after 2, 3, and 4 frames, respectively. This minimizes frame-time spikes and prevents UI stuttering.

### Issue Fixed
This change addresses [#166184](https://github.com/flutter/flutter/issues/166184), allowing `flutter_svg` to dynamically adjust its rendering strategy for better performance and reduced manual configuration.

Why doesn't Skia have this performance issue?

Because in Skia mode, there is component-level raster caching that automatically stores the result. However, according to [#131206](https://github.com/flutter/flutter/issues/131206), flutter previously decided not to port the raster cache to Impeller.

### Test
When a page contains 20 SVGs, its rendering trace for a single frame is as follows:
image

Each small SVG requires a short period of time to render(two RenderPass):
image

Use this auto Strategy, the rendering trace can be this:
image
Each frame saves a significant amount of time spent on executing RenderPass.

Because we have a limit of generating one raster cache per frame, the frame time gradually decreases as the page with 20 SVGs loads, as shown below:
image

The raster cache will be generated with a delay of two frames after the component is drawn. It is created in the raster thread, running concurrently with UI operations before the actual rendering occurs:
image

Once this PR is merged, the changes to integrate this feature into `flutter_svg` will be proposed.

## Pre-Review Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
[...]

PR #9082, opened 2025-04-15
[vector_graphics] Fix raster cache size calculation and incorporate ColorFilter/Opacity effects into cache

gaodi-sec:
This PR fixes the bugs mentioned in [#8932](https://github.com/flutter/packages/pull/8932).

1. The raster cache size generated by `vector_graphics` takes `BoxFit` into account but does not consider the impact of the `Transform` widget. In practice, the raster cache size can be derived directly from the canvas's scale, which naturally reflects all visual transformations—including BoxFit, Transform, and any other factors affecting the final display size of the component. This PR replaces the previous size calculation with one based on the canvas scale.

2. Currently, the raster cache does not include `ColorFilter` or `Opacity,` meaning that rendering overhead remains high when either is present, due to the need for additional offscreen rendering passes. However, in real-world scenarios—especially when using SVG assets—it's very common to apply `ColorFilter` for tinting. This PR incorporates `ColorFilter` and `Opacity` effects into the raster cache, reducing the overall rendering cost.

Of course, this PR also aims to address issue [#166184](https://github.com/flutter/flutter/issues/166184).

Regarding point 2, before applying the fix, the trace output was as follows:
image

At that time, two SVGs were causing two vkQueueSubmit calls, resulting in two offscreen renderings, which took approximately 2ms.

After the fix, the trace output is as follows:

image

The vkQueueSubmit calls are no longer present, and the processing time has decreased to 0.6ms.

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/vector_graphics/CHANGELOG.md
+++ b/packages/vector_graphics/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.19
+
+* Fix incorrect scaling under Transform in raster mode and reduce extra rendering cost from ColorFilter/Opacity in raster mode.
+
 ## 1.1.18
 
 * Allow transition between placeholder and loaded image to have an animation.
--- a/packages/vector_graphics/lib/src/render_vector_graphic.dart
+++ b/packages/vector_graphics/lib/src/render_vector_graphic.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:math' as math;
 import 'dart:ui' as ui;
 
 import 'package:flutter/animation.dart';
@@ -15,7 +16,7 @@ import 'listener.dart';
 @immutable
 class RasterKey {
   /// Create a new [RasterKey].
-  const RasterKey(this.assetKey, this.width, this.height);
+  const RasterKey(this.assetKey, this.width, this.height, this.paint);
 
   /// An object that is used to identify the raster data this key will store.
   ///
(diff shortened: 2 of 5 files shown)


jonahwilliams (commented):
I don't think you can just use the canvas scale factor, as a frame may include many different canvases.

jonahwilliams:
That doesn't really follow. This change is not safe. You cannot assume there is no widget higher in the tree applying a scale factor.

jonahwilliams:
Closing as feedback was not addressed.

gaodi-sec:
Sorry for the late reply. I'm trying to understand the phrase “no widget higher in the tree applying a scale factor”: 
even if a higher-level widget applies a scale factor, this factor will eventually be applied to the canvas. **As long as the canvas is actually rendering to the screen, the resolution calculated using the canvas's scale will reflect the real rendering resolution.**

e.g.：
widget1 ：
&nbsp;&nbsp;canvas.store();
&nbsp;&nbsp;canvas.scale(0.2,0.2);
&nbsp;&nbsp;draw child widget2;
&nbsp;&nbsp;&nbsp;&nbsp;**here the render size is the picture size * 0.2**;
&nbsp;&nbsp;&nbsp;&nbsp;canvas.drawPicture();
&nbsp;&nbsp;canvas.restore;

So there's only one case where this wouldn't hold true—when the canvas used for drawing doesn't render directly to the screen but comes from a recording canvas instead. In that case, the actual resolution wouldn't be available.

However, I haven't found any case where a widget is drawn by another widget using a recording canvas. From what I see, only internal drawing within a widget might use such caching via recording.

Given this, can we assume there's no issue with resolution calculation in normal cases?

gaodi-sec:
The changes that move the color and colorFilter effects into the raster cache have been separated into [PR #9398](https://github.com/flutter/packages/pull/9398).

PR #9112, opened 2025-04-18
[various] Scrubs pre-SDK-21 Android code

stuartmorgan-g:
- Removes any annotations or runtime checks related to Android API 20 or 21, as 21 is the minimum API version supported by the versions of Flutter these plugins support.
- Removes comment/README references to limitations for <21.
- Updates READMEs for app-facing packages to reflect the current support levels.

Fixes https://github.com/flutter/flutter/issues/157106

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ConvertTest.java
```diff
@@ -49,10 +48,8 @@
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 import org.robolectric.RobolectricTestRunner;
-import org.robolectric.annotation.Config;
 
 @RunWith(RobolectricTestRunner.class)
-@Config(minSdk = Build.VERSION_CODES.LOLLIPOP)
```
reidbaker, line 55: I think the @config annotation should stay and you should use the new minimum. 

minSdk= is supposed to run the tests for every version of android after that version.

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/GoogleMapControllerTest.java
```diff
@@ -41,10 +40,8 @@
 import org.mockito.MockitoAnnotations;
 import org.robolectric.Robolectric;
 import org.robolectric.RobolectricTestRunner;
-import org.robolectric.annotation.Config;
 
 @RunWith(RobolectricTestRunner.class)
-@Config(minSdk = Build.VERSION_CODES.LOLLIPOP)
```
reidbaker, line 47: same here.

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/GoogleMapInitializerTest.java
```diff
@@ -23,10 +22,8 @@
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 import org.robolectric.RobolectricTestRunner;
-import org.robolectric.annotation.Config;
 
 @RunWith(RobolectricTestRunner.class)
-@Config(minSdk = Build.VERSION_CODES.LOLLIPOP)
```
reidbaker, line 29: and here.  Not marking any more but if we agree on @config staying then update all instances.

reidbaker (approved):
approve % comment about @config with minsdk set.

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ConvertTest.java
stuartmorgan-g: Maybe I misunderstood `@Config`. Does it restrict the APIs that this test *can* be run under, or does it *cause* the test to be run with multiple API levels? I thought it was the former.

If it's the latter, I'm skeptical we want this. We don't do this pretty much anywhere else in the repo, and I can't think of any reason we would want to multi-run things for the maps SDK in particular, since the SDK isn't system-API-level based.

reidbaker: My understanding is it is the latter. Here is some documentation that I think supports that conclusion. https://robolectric.org/configuring/#configure-sdk-level. If you wanted to run on a specific set then you would use sdk = <SOME SET OF API LEVELS>. 

Looking through ConvertTest I agree but there are other places in packages where it does make sense to run with a minSdk or other broader android versions set. Primarily to make sure branching logic based on android api gets exercised. 

Update: a spot check of all of these google maps test entries shows minimal value in multi api level tests because they do not branch.

stuartmorgan-g: I'll go ahead and remove them then; the fact that we only do this for these tests, and these tests don't seem like they benefit, makes me think that adding it was a result of a misunderstanding like mine.

tarrinneal (approved):

PR #9151, opened 2025-04-25
[webview_flutter_android] feat: Expose setUseWideViewPort on Android (#106999)

Gustl22:
The option to set whether the webview should use the wide viewport should be available.
Also the default is changed to false, because this is the expected behavior on Android devices:
https://developer.android.com/develop/ui/views/layout/webapps/targeting
On iOS also the default behavior is to NOT use wide viewport, so this is aligned.

> Note: When your page is rendered in a WebView, it doesn't use wide viewport mode by default. You can enable wide viewport mode with setUseWideViewPort().

This also partically solves flutter/flutter#106999, as then the correct height is returned from `document.documentElement.scrollHeight;` property.

## Pre-launch Checklist

- [x] I signed the [CLA].


stuartmorgan-g:
> Also the default is changed to false, because this is the expected behavior on Android devices

This may break some people; see https://github.com/flutter/flutter/issues/93125 for example.

Gustl22:
Thank you for the intels.

> This may break some people; see https://github.com/flutter/flutter/issues/93125 for example.

That's true, what exactly should be the next steps? I also tried to reproduce your linked issue, but couldn't as the linked was reworked and doesn't show this problem any more.

1. Use `true` as default for `setUseWideViewPort`
  - would ensure that nothing breaks due to keeping the default
  - may leads to annoyance as package users expect the same behavior on Android and iOS
2. Use `false` as default for `setUseWideViewPort`
  - Making a breaking change and write a migration guide
  - probably very few care as the behavior was different on Android and iOS and noone noticed
  - Nontheless could lead to unexpected outcome for certain products, when using flutter specifically on only one platform
  - WebViews are hard to test so the behavior should be aligned in both platforms to not challenge more differences across both platforms
  - Raise a major version for webview package
  - Aligns with Androids default and iOS implementation


<table>
<tr>
<th>Webkit / iOS</th>
<th>Android setUseWideViewPort(true)</th>
<th>Android setUseWideViewPort(false)</th>
</tr>
<tr>
<td width="33%"></td>
<td width="33%"></td>
<td width="33%"></td>
</tr>
</table>

stuartmorgan-g:
> * probably very few care as the behavior was different on Android and iOS and noone noticed
> * WebViews are hard to test so the behavior should be aligned in both platforms to not challenge more differences across both platforms

I think these are good arguments for switching the default back. And clients will have the escape hatch of the new setting if they really need the other behavior. Thanks for the analysis!

Since the initial change wasn't a major version change, and as you point out this is likely not going to affect most clients, we don't need to do a major version change to switch it back IMO.

Gustl22:
Thank you for the fast reply. 

Should I leave the viewport meta example in the example project? Or is it also desired to be adapted in the `webview_flutter_android` example project? Do example project updates need version changes?

packages/webview_flutter/webview_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 4.5.0
+
+* Adds support to set using wide view port.
+  Change default to `false` to align with native WebViews.
+  See `AndroidWebViewController.setUseWideViewPort`.
```
bparrishMines, line 5: Change -> Changes
nit: Start a new bullet point 
```suggestion
* Changes default to `false` to align with native WebViews. See `AndroidWebViewController.setUseWideViewPort`.
```

bparrishMines (changes requested):
@Gustl22 The example apps are mostly used for maintainers to test things manually. Since this flag is unique to Android, I think you should move the changes you made  in `webview_flutter` example to `webview_flutter_android` example.

bparrishMines (approved):
LGTM 

@stuartmorgan-g for secondary review

stuartmorgan-g (approved):
LGTM

PR #9175, opened 2025-04-29
[in_app_purchase] Make Storekit 2 the default

LouiseHsu:
making sure ci is happy


LongCatIsLooong:
It looks like there's no way for the developer to opt-out of StoreKit 2 with this change (unless the user is on a lower version of iOS/macOS and `enableStoreKit2` is called)? If that's intended does that mean we should get rid of the `enableStoreKit2` API and instead does the version check internally?

LouiseHsu:
> It looks like there's no way for the developer to opt-out of StoreKit 2 with this change (unless the user is on a lower version of iOS/macOS and `enableStoreKit2` is called)? If that's intended does that mean we should get rid of the `enableStoreKit2` API and instead does the version check internally?

sorry this shouldve been a draft, i just wanted to make sure ci is happy with making storekit2 the default. But yes, enableStoreKit2 only enables if the device is over ios 15. They shouldnt be able to opt out because storekit 1 is deprecated.

PR #9198, opened 2025-05-04
[video_player_android] isBuffering flag always return true after calling seekTo method - fixed

raju-muliyashiya:
fixes  [Issue #165149](https://github.com/flutter/flutter/issues/165149) 

## Pre-Review Checklist

- [X] I signed the [CLA].

### Root Cause

The issue stems from the `onPlaybackStateChanged` callback. After calling `seekTo`, the player transitions to `Player.STATE_BUFFERING`, setting the buffering flag to `true`. Shortly after, it transitions to `Player.STATE_READY`. However, if the player was already initialized, the method returned early due to:

```java
if (isInitialized) {
  return;
}
```

prevented `setBuffering(false)` from being called in the shared logic below the switch


camsim99 (commented):
Overall this change seems reasonable, but I want to ensure that this fixes the root cause of the issue.

Can you update the description to include a justification? Then, I think we should try re-enabling the integration test mentioned in https://github.com/flutter/flutter/issues/165149#issuecomment-2734518026: https://github.com/flutter/packages/blob/e064b04006573ffe53ccac3405231131e1c1bbd6/packages/video_player/video_player/example/integration_test/controller_swap_test.dart#L17.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -70,11 +70,10 @@ public void onPlaybackStateChanged(final int playbackState) {
         events.onBufferingUpdate(exoPlayer.getBufferedPosition());
         break;
       case Player.STATE_READY:
-        if (isInitialized) {
-          return;
+        if (!isInitialized) {
+          isInitialized = true;
+          sendInitialized();
```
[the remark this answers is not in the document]
raju8000: Sure, will add test case.

stuartmorgan-g:
> Overall this change seems reasonable, but I want to ensure that this fixes the root cause of the issue.

The timing certainly checks out; the early return was added in https://github.com/flutter/packages/pull/8466 on March 4th, and the issue was first reported on March 13th.

packages/video_player/video_player_android/CHANGELOG.md
```diff
@@ -110,7 +115,9 @@
 
 ## 2.7.0
 
-* Re-adds [support for Impeller](https://docs.flutter.dev/release/breaking-changes/android-surface-plugins).
+*
+
+Re-adds [support for Impeller](https://docs.flutter.dev/release/breaking-changes/android-surface-plugins).
```
camsim99, line 120: Please undo this change:
```suggestion
* Re-adds [support for Impeller](https://docs.flutter.dev/release/breaking-changes/android-surface-plugins).
```

packages/video_player/video_player_android/CHANGELOG.md
```diff
@@ -284,3 +296,4 @@
 ## 2.2.17
 
 * Splits from `video_player` as a federated implementation.
+  `
```
camsim99, line 299: Please undo this change too.

Also nit: the rest of the changes seem like auto-formatter changes, that are not needed.

camsim99 (approved):
LGTM but please fix the CHANGELOG :)

raju-muliyashiya:
@stuartmorgan-g pending review

PR #9168, opened 2025-04-28
[all_packages] Bump `compileSdk` version to 35

camsim99:
Bumps `all_packages` app to build with `compileSdk` version 35. Then,

- Updates the Gradle and AGP versions to support this new Android SDK version
- Refactors the legacy all_packages CI build target such that the regular `all_packages` CI build targets (master, stable) will build both the modern `all_packages` app and the legacy version with the same JDK 17 version, since the legacy version now also requires it (example successful builds: [master](https://ci.chromium.org/ui/p/flutter/builders/try/Linux_android%20android_build_all_packages%20master/17102/overview), [stable](https://ci.chromium.org/ui/p/flutter/builders/try/Linux_android%20android_build_all_packages%20stable/17086/overview))
- Updates legacy `all_packages` app `README.md` to more accurately describe how the app was transformed since its creation

Prompted by https://github.com/flutter/packages/pull/8414.

## Pre-Review Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter].
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ ] I [linked to at least one issue that this PR fixes] in the description above.
- [ ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or I have commented below to indicate which [version change exemption] this PR falls under[^1].
- [ ] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or I have commented below to indicate which [CHANGELOG exemption] this PR falls under[^1].
- [ ] I updated/added any relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or I have commented below to indicate which [test exemption] this PR falls under[^1].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[^1]: Regular contributors who have demonstrated familiarity with the repository guidelines only need to comment if the PR is not auto-exempted by repo tooling.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[...]

PR #9267, opened 2025-05-16
[google_sign_in] Redesign API for current identity SDKs

stuartmorgan-g:
This is a full overhaul of the `google_sign_in` API, with breaking changes for all component packages—including the platform interface. The usual model of adding the new approach while keeping the old one is not viable here, as the underlying SDKs have changed significantly since the original API was designed. Web already had some only-partially-compatible shims for this reason, and Android would have had to do something similar; see https://github.com/flutter/flutter/issues/119300 and https://github.com/flutter/flutter/issues/154205, and [the design doc](https://flutter.dev/go/google-sign-in-authn-authz-updates) for more background.

- Fixes https://github.com/flutter/flutter/issues/119300
- Fixes https://github.com/flutter/flutter/issues/154205
- Fixes https://github.com/flutter/flutter/issues/139406
- Fixes https://github.com/flutter/flutter/issues/150365
- Fixes https://github.com/flutter/flutter/issues/137727
- Fixes https://github.com/flutter/flutter/issues/124206
- Fixes https://github.com/flutter/flutter/issues/117794
- Fixes https://github.com/flutter/flutter/issues/107532
- Fixes https://github.com/flutter/flutter/issues/85439
- Fixes https://github.com/flutter/flutter/issues/74308
- Fixes https://github.com/flutter/flutter/issues/71607
- Fixes https://github.com/flutter/flutter/issues/70427
- Fixes https://github.com/flutter/flutter/issues/161890
- Fixes https://github.com/flutter/flutter/issues/157639
- Fixes https://github.com/flutter/flutter/issues/67308
- Fixes https://github.com/flutter/flutter/issues/36673
- Fixes https://github.com/flutter/flutter/issues/32441 (there may be more to do here over time, since we may find exceptions that are not caught, but we now have a structured system to convert errors to as we find specific unhandled cases)

## Pre-Review Checklist

- [x] I signed the [CLA].


ycv005:
Waiting for this.

stuartmorgan-g:
@ycv005 Please stop spamming PRs and issues with [unhelpful comments](https://github.com/flutter/flutter/blob/master/docs/contributing/issue_hygiene/README.md#do-not-add-me-too-or-same-or-is-there-an-update-comments-to-issues-or-prs).

stuartmorgan-g:
Looks like this will need either https://github.com/flutter/packages/pull/9168 or a localized SDK 35 change.

packages/google_sign_in/google_sign_in_web/README.md
```diff
@@ -190,35 +49,43 @@ You can tell `flutter run` to listen for requests in a specific host and port wi
...
+* `401`: Missing or invalid access token.
+* `403`: Expired access token.
 
-## Issues and feedback
+See the "Integration considerations > [UX separation for authentication and authorization](https://developers.google.com/identity/gsi/web/guides/integrate#ux_separation_for_authentication_and_authorization)
+guide" in the official GIS SDK documentation for more information about this.
 
-Please file [issues](https://github.com/flutter/flutter/issues/new)
-to send feedback or report a bug.
+### Migration from versions before 0.12
```
ditman, line 87: Is this paragraph duplicated on purpose? IMO this could be a `##` after the Issues and feedback section, and have it there only.

packages/google_sign_in/google_sign_in_web/example/integration_test/google_sign_in_web_test.dart
```diff
@@ -173,139 +135,182 @@ void main() {
...
 
-      testWidgets('returns the signed-in user', (_) async {
-        final GoogleSignInUserData someUser = extractUserData(person)!;
-
-        mockito
-            .when(mockGis.signIn())
-            .thenAnswer((_) => Future<GoogleSignInUserData>.value(someUser));
+    group('clientAuthorizationTokensForScopes', () {
+      const String someAccessToken = '50m3_4cc35_70k3n';
+      const List<String> scopes = <String>['scope1', 'scope2'];
```
ditman, line 164: Would it make sense to retain the `'asserts no scopes have any spaces'` functionality and test here?

packages/google_sign_in/google_sign_in_web/example/integration_test/people_test.dart
```diff
@@ -52,7 +52,6 @@ void main() {
       expect(user.id, expectedPersonId);
       expect(user.displayName, expectedPersonName);
       expect(user.photoUrl, expectedPersonPhoto);
-      expect(user.idToken, isNull);
```
ditman, line 55: I thought we no longer did the People API fallback on the web? Maybe this whole test can be removed as well?

packages/google_sign_in/google_sign_in_web/lib/src/gis_client.dart
```diff
@@ -12,7 +12,6 @@ import 'package:web/web.dart' as web;
 
 import 'button_configuration.dart'
     show GSIButtonConfiguration, convertButtonConfiguration;
-import 'people.dart' as people;
```
ditman, line 15: I think this `people.dart` file is now dead code, you can maybe delete it and its tests?

ditman (commented):
I don't think I have anything blocking. I took a look at `google_sign_in`, `google_sign_in_platform_interface` and `google_sign_in_web`.

packages/google_sign_in/google_sign_in/README.md
```diff
@@ -6,182 +6,145 @@ A Flutter plugin for [Google Sign In](https://developers.google.com/identity/).
...
 
-The `requestScopes` returns a `boolean` value that is `true` if the user has
-granted all the requested scopes or `false` otherwise.
-
-Once your app determines that the current user `isAuthorized` to access the
-services for which you need `scopes`, it can proceed normally.
-
 ### Authorization expiration
 
 In the web, **the `accessToken` is no longer refreshed**. It expires after 3600
```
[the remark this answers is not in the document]
stuartmorgan-g: [Looks that way](https://cloud.google.com/docs/authentication/token-types#id-lifetime). I can adjust the README accordingly, but maybe you just didn't mention it because normally (IIUC) clients don't need to keep using ID tokens they way they do access tokens.

packages/google_sign_in/google_sign_in/example/lib/main.dart
```diff
@@ -50,51 +52,84 @@ class _SignInDemoState extends State<SignInDemo> {
...
+    // #docregion CheckAuthorization
+    GoogleSignInAccount? user;
+    // #enddocregion CheckAuthorization
+    String error = '';
+    switch (event) {
+      case GoogleSignInAuthenticationEventSignIn():
+        user = event.user;
+      case GoogleSignInAuthenticationEventSignOut():
+        user = null;
+      case GoogleSignInAuthenticationEventException():
```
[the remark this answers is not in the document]
stuartmorgan-g: I think I just got carried away with structured error returns on iOS and Android. I'll revisit this for both layers of streams.

packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
```diff
@@ -27,134 +24,129 @@ abstract class GoogleSignInPlatform extends PlatformInterface {
 
   static final Object _token = Object();
```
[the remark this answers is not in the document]
stuartmorgan-g: I did seriously consider doing this, since it can only be done during a breaking change to the platform interface.

I may still tackle it here for that reason, but the downside is that I would have to replace the Mockito-generated mocks with manual mocks, and rewrite the tests accordingly.

packages/google_sign_in/google_sign_in_web/README.md
stuartmorgan-g: They aren't quite duplicated; authn and authz have different migration docs. See lines 44/45 in the pre-PR version of the file.

packages/google_sign_in/google_sign_in_web/example/integration_test/google_sign_in_web_test.dart
stuartmorgan-g: I was confused by the fact that the web implementation was asserting this in the first place. Is there some web-specific extra requirement on scopes? Is having spaces that shouldn't be there different from just typoing the scope?

packages/google_sign_in/google_sign_in_web/example/integration_test/people_test.dart
stuartmorgan-g: I think I did indeed just forget to delete these files.

packages/google_sign_in/google_sign_in/example/lib/main.dart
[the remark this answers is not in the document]
ditman: IMO I would only change the user-facing stream. In the implementation of the plugin we're already looking at the continuous stream of events coming from the platform code, so maybe the platform -> plugin stream can continue be a single one, with error events ("it never crashes") but the plugin -> app stream can emit errors normally? But yeah, if there's a benefit to making the platform -> plugin also emit errors, I won't oppose it changing :)

packages/google_sign_in/google_sign_in/pubspec.yaml
```diff
@@ -35,6 +35,7 @@ dev_dependencies:
     sdk: flutter
   http: ">=0.13.0 <2.0.0"
   mockito: ^5.4.4
+  plugin_platform_interface: ^2.1.8
```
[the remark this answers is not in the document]
ditman: Ahhh ok! I like that the platform_interface vends the mocking interface as well. Also just noticed that this is in `dev_dependencies`. Thanks for the clarification!

packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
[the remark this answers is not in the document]
ditman: > the downside is that I would have to replace the Mockito-generated mocks with manual mocks, and rewrite the tests accordingly.

Ah yes...

Why is this change considered breaking, though? Is it just in case people are using "implements" for their own federated versions of the plugin, or mocks? If you're using the class "as intended" (with extends), your code shouldn't need to change by us moving it to `base` class. Would it?

One upside of the extra work, should you want to tackle it, is that the mocks that you create here can be distributed as a `testing` library from the package, so those users who were creating their own mocks can use officially maintained mock classes that we vend. Similar to: https://pub.dev/documentation/http/1.4.0/testing

packages/google_sign_in/google_sign_in_web/README.md
ditman: I totally didn't notice the difference in the migration link. Too subtle after reading this far :P

packages/google_sign_in/google_sign_in_web/example/integration_test/google_sign_in_web_test.dart
ditman: This code comes from the first version of the plugin, and this is why I added this, when asked about "why not put this in the app-facing package":

https://github.com/flutter/plugins/pull/2280#discussion_r347682416

> [Spaces are] a problem here because we use the space as a "join" character to pass the scopes to JS. In the mobile versions, the List is preserved all the way through.

I just checked the API docs again, here: https://developers.google.com/identity/protocols/oauth2/javascript-implicit-flow#redirecting and `scope` is still "A space-delimited list of scopes that identify the resources that your application could access on the user's behalf."

IMO we should still assert (and test the assert) that scopes don't contain spaces, but it doesn't look like a super big source of problems.

stuartmorgan-g: Interesting; I'll put it back with a comment explaining it. I wonder if the mobile SDKs fail under the hood, or if they escape the spaces.

packages/google_sign_in/google_sign_in_android/README.md
```diff
@@ -13,3 +13,28 @@ should add it to your `pubspec.yaml` as usual.
 
 [1]: https://pub.dev/packages/google_sign_in
 [2]: https://flutter.dev/to/endorsed-federated-plugin
+
+## Integration
```
camsim99, line 17: Can we consider linking to the `credentials` docs on this? https://developer.android.com/identity/sign-in/credential-manager-siwg#set-google for the non-firebase option? If it's more confusing than helpful because some steps don't apply, maybe not though.

camsim99 (commented):
Reviewed the Android implementation!

packages/google_sign_in/google_sign_in_android/README.md
stuartmorgan-g: By the time I was updating the READMEs I was so used to skipping down to the code on that page I forgot it had setup instructions! I've adjusted to link to that as the non-Firebase option, and removed some of the details since they are covered more thoroughly there.

packages/google_sign_in/google_sign_in_android/lib/google_sign_in_android.dart
```diff
@@ -14,111 +15,250 @@ class GoogleSignInAndroid extends GoogleSignInPlatform {
...
+        : ServerAuthorizationTokenData(serverAuthCode: serverAuthCode);
   }
 
-  @override
-  Future<bool> requestScopes(List<String> scopes) {
-    return _api.requestScopes(scopes);
+  Future<PlatformGoogleIdTokenCredential?> _authenticate({
+    required bool filterToAuthorized,
+    required bool autoSelectEnabled,
+    required bool useButtonFlow,
```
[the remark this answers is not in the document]
camsim99: Ok cool this is much clearer, thank you!

packages/google_sign_in/google_sign_in_android/README.md
camsim99: Hahaha totally understandable! LGTM!

stuartmorgan-g:
@LongCatIsLooong / @cbracken Ping on the iOS portion of this review.

stuartmorgan-g:
@bparrishMines Could you review the platform interface and app-facing package changes?

packages/google_sign_in/google_sign_in/README.md
```diff
@@ -6,182 +6,146 @@ A Flutter plugin for [Google Sign In](https://developers.google.com/identity/).
...
 
-_(Only implemented in the web platform, from version 6.1.0 of this package)_
+[Full list of available scopes](https://developers.google.com/identity/protocols/googlescopes).
 
 ### Requesting more scopes when needed
 
 If an app determines that the user hasn't granted the scopes it requires, it
-should initiate an Authorization request. (Remember that in the web platform,
-this request **must be initiated from an user interaction**, like a button press).
+should initiate an Authorization request. On some platforms, such as web,
```
LongCatIsLooong: nit: this sounds like there are platforms other than web where user interactions are required for Authorization request. Is that the case? How do I know which of the platforms my app targets have such restriction?

packages/google_sign_in/google_sign_in/example/lib/main.dart
```diff
@@ -50,58 +51,107 @@ class _SignInDemoState extends State<SignInDemo> {
...
-// #docregion CanAccessScopes
-      // In mobile, being authenticated means being authorized...
-      bool isAuthorized = account != null;
-      // However, on web...
-      if (kIsWeb && account != null) {
-        isAuthorized = await _googleSignIn.canAccessScopes(scopes);
-      }
-// #enddocregion CanAccessScopes
+    // #docregion Setup
+    final GoogleSignIn signIn = GoogleSignIn.instance;
```
LongCatIsLooong, line 62: Hmm this reminds me of https://github.com/flutter/flutter/issues/168100, if the user connects / disconnects a mouse / keyboard on Android the activity will restart and that would result in the element tree being re-inflated anew. 

I guess that does not break the new flow (from reading the docs `initialize` has to be called exactly once in a program) since the new activity will spin up a new dart vm? Still it would be an annoyance that every time I unplug my keyboard I have to go over the sign-in process again.

packages/google_sign_in/google_sign_in/example/lib/main.dart
```diff
@@ -50,58 +51,107 @@ class _SignInDemoState extends State<SignInDemo> {
...
 
-      // Now that we know that the user can access the required scopes, the app
-      // can call the REST API.
-      if (isAuthorized) {
-        unawaited(_handleGetContact(account!));
-      }
+  Future<void> _handleAuthenticationEvent(
+      GoogleSignInAuthenticationEvent event) async {
+    // #docregion CheckAuthorization
+    GoogleSignInAccount? user;
```
LongCatIsLooong: uber nit, consider making `user` and `authorization` final if possible?

packages/google_sign_in/google_sign_in/lib/google_sign_in.dart
```diff
@@ -155,321 +95,426 @@ class GoogleSignInAccount implements GoogleIdentity {
...
+        return true;
+    }
   }
 
-  /// Starts the interactive sign-in process.
+  /// Whether or not the current platform supports the [authenticate] method.
   ///
-  /// Returned Future resolves to an instance of [GoogleSignInAccount] for a
-  /// successful sign in or `null` in case sign in process was aborted.
+  /// If this returns false, [authenticate] will throw an UnsupportedError if
```
LongCatIsLooong: nit: missing backticks / square brackets around UnsupportedError?

packages/google_sign_in/google_sign_in/lib/google_sign_in.dart
```diff
@@ -155,321 +95,426 @@ class GoogleSignInAccount implements GoogleIdentity {
...
+  /// called. See the platform-specific documentation for the package to
+  /// determine how authentication his handled. For instance, the platform may
```
LongCatIsLooong: his -> is

packages/google_sign_in/google_sign_in/lib/src/event_types.dart
```diff
@@ -0,0 +1,36 @@
...
+import 'package:flutter/foundation.dart';
+
+import '../google_sign_in.dart';
+
+export 'package:google_sign_in_platform_interface/google_sign_in_platform_interface.dart'
+    show GoogleSignInException;
+
+/// A base class for authentication event streams.
+@immutable
+sealed class GoogleSignInAuthenticationEvent {
```
LongCatIsLooong, line 14: Are these classes needed? It seems the sealed type is a glorified `GoogleSignInAccount?`.

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Sources/google_sign_in_ios/FLTGoogleSignInPlugin.m
```diff
@@ -23,35 +23,83 @@
...
 
 @interface FLTGoogleSignInPlugin ()
 
 // The contents of GoogleService-Info.plist, if it exists.
-@property(strong, nullable) NSDictionary<NSString *, id> *googleServiceProperties;
+@property(nonatomic, nullable) NSDictionary<NSString *, id> *googleServiceProperties;
 
 // The plugin registrar, for querying views.
-@property(strong, nonnull) id<FlutterPluginRegistrar> registrar;
+@property(nonatomic) id<FlutterPluginRegistrar> registrar;
```
LongCatIsLooong: is this still `nonnull`?

packages/google_sign_in/google_sign_in/example/lib/main.dart
LongCatIsLooong: Or when the activity restarts, the lightweight authentication path will be taken and the flow will typically be invisible to user / require no additional user interactions?

EDIT: Just read the design doc and it mentioned "Fully silent sign in is no longer available"

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Sources/google_sign_in_ios/FLTGoogleSignInPlugin.m
```diff
@@ -260,15 +293,13 @@ - (void)signInWithHint:(nullable NSString *)hint
 
 // Wraps the iOS and macOS scope addition methods.
 - (void)addScopes:(NSArray<NSString *> *)scopes
-       completion:(nullable void (^)(GIDSignInResult *_Nullable signInResult,
-                                     NSError *_Nullable error))completion {
-  GIDGoogleUser *currentUser = self.signIn.currentUser;
+    forGoogleSignInUser:(GIDGoogleUser *)user
```
LongCatIsLooong: I assume the `user` argument is not nullable, otherwise it's failing silently?

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Sources/google_sign_in_ios/FLTGoogleSignInPlugin.m
```diff
@@ -101,150 +149,135 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
...
+                                              NSError *_Nullable error) {
+    [weakSelf handleAuthResultWithUser:refreshedUser
+                        serverAuthCode:nil
+                                 error:error
+                            completion:completion];
   }];
 }
 
+- (void)addScopes:(nonnull NSArray<NSString *> *)scopes
+          forUser:(NSString *)userId
```
LongCatIsLooong: can userId be null?

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Sources/google_sign_in_ios/FLTGoogleSignInPlugin.m
```diff
@@ -101,150 +149,135 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
...
+                                           completion:completion];
+                 }];
+  } @catch (NSException *e) {
+    completion(nil, [FlutterError errorWithCode:@"request_scopes" message:e.reason details:e.name]);
+  }
+}
+
 - (void)signOutWithError:(FlutterError *_Nullable *_Nonnull)error {
   [self.signIn signOut];
+  [self.usersByIdentifier removeAllObjects];
```
LongCatIsLooong: The `signOut` method seems to only sign out the current user: https://developers.google.com/identity/sign-in/ios/reference/Classes/GIDSignIn

Why do we have to remove everything in the dictionary? If `refreshedAuthorizationTokensForUser` gets called for a different user (is that possible`?) the app would get an error it seems?

packages/google_sign_in/google_sign_in/README.md
LongCatIsLooong: `supportsAuthenticate`? But that's for authentication I assume?

packages/google_sign_in/google_sign_in/example/lib/main.dart
stuartmorgan-g: Made `user` final. Making `authorization` final would require adding an `else { authorization = null; }` or restructuring to use a ternary in the assignment or using patterns, all of which would make the example more complicated for the README snippet.

packages/google_sign_in/google_sign_in/lib/google_sign_in.dart
stuartmorgan-g: Fixed.

packages/google_sign_in/google_sign_in/lib/google_sign_in.dart
stuartmorgan-g: Fixed.

packages/google_sign_in/google_sign_in/lib/src/event_types.dart
stuartmorgan-g: Originally there were three subclasses, but exceptions moved into `onError`; now we could make them `GoogleSignInAccount?`.

It would make things less self-documenting though. It's non-obvious that getting `null` as an `authenticationEvent` means "the user signed out". Is that worth the slightly easier usage?

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Sources/google_sign_in_ios/FLTGoogleSignInPlugin.m
stuartmorgan-g: Yes, I must have deleted that by mistake when removing `strong`. Re-added.

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Sources/google_sign_in_ios/FLTGoogleSignInPlugin.m
stuartmorgan-g: Yes. I've added some more nullability annotations for the private methods; in past projects we didn't annotate implementation files in general, but there's no reason not to here, and since Obj-C nullability is basically just documentation anyway we may as well express it (as we generally do with our newer Java plugin code).

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Sources/google_sign_in_ios/FLTGoogleSignInPlugin.m
stuartmorgan-g: Nope, added annotation.

packages/google_sign_in/google_sign_in_ios/darwin/google_sign_in_ios/Sources/google_sign_in_ios/FLTGoogleSignInPlugin.m
stuartmorgan-g: > The `signOut` method seems to only sign out the current user

That's true, but unlike the other platforms, Google Sign In for iOS bakes in the idea of a single current user in various places.

> Why do we have to remove everything in the dictionary? If `refreshedAuthorizationTokensForUser` gets called for a different user (is that possible`?) the app would get an error it seems?

I was going to say it would anyway, but it looks like [valid tokens would still be returned](https://github.com/google/GoogleSignIn-iOS/blob/594f8b79b3380cef6c075b81e09f6e611685512a/GoogleSignIn/Sources/GIDGoogleUser.m#L121-L127). Adding scopes, on the other hand [will just fail](https://github.com/google/GoogleSignIn-iOS/blob/594f8b79b3380cef6c075b81e09f6e611685512a/GoogleSignIn/Sources/GIDGoogleUser.m#L203-L213). The underlying SDK API is a little strange in that it will potentially vend multiple user objects, but some operations will fail on all but one (and as a result, our API has the same behavior on iOS).

But it looks like we'll actually get better behavior by never clearing `usersByIdentifier`, so calls will continue to use the underlying SDK user objects if they have ever been available, and we can leave it to the SDK to work or not depending on its implementation details. User objects should be pretty lightweight, and in practice I expect it will be rare to make more than one anyway.

I've replaced this with a comment explaining why we're not removing anything from the dictionary.

packages/google_sign_in/google_sign_in/README.md
stuartmorgan-g: In practice, it's just web (and that may well be true forever). I changed this because we are generally moving toward trying to be as platform agnostic as possible in the app-facing package, because the idea of a federated plugin is that we don't know what other platforms might be supported. You make a great point though; making the README more general without providing a corresponding API just means it's vague and less actionable.

I like the idea of using support queries for this. I plumbed through an `authorizationRequiresUserInteraction()` for this purpose, and updated the README and API docs accordingly.

packages/google_sign_in/google_sign_in/example/lib/main.dart
stuartmorgan-g: > Still it would be an annoyance that every time I unplug my keyboard I have to go over the sign-in process again.

That's up to whether the app developer chooses to include auth state in their state restoration.

> Or when the activity restarts, the lightweight authentication path will be taken

That's up to the app developer too.

packages/google_sign_in/google_sign_in/README.md
LongCatIsLooong: Sounds good (I don't see a new commit tho).

packages/google_sign_in/google_sign_in/example/lib/main.dart
LongCatIsLooong: >  include auth state in their state restoration

Oh I thought the developer would have to restore GoogleSignIn, and state restoration only supports value types. If the dart part of GoogleSignIn doesn't keep any secret state itself then that makes sense.

packages/google_sign_in/google_sign_in/example/lib/main.dart
LongCatIsLooong: Does this work?

```dart
final authorization = await user?.authorizationClient.authorizationForScopes(scopes);
```

packages/google_sign_in/google_sign_in/lib/src/event_types.dart
LongCatIsLooong: I was thinking maybe we can avoid introducing new classes by using typedef or extension types. I played with that idea but found the resulting code harder to read (the user code would still be the same though). So never mind.

LongCatIsLooong (commented):
The changes SGTM but I don't see a new commit?

stuartmorgan-g:
> The changes SGTM but I don't see a new commit?

Oops, I didn't notice that the push failed (I forgot I'd done a merge from main via the GitHub UI).

stuartmorgan-g:
> Does this work?
> 
> ```dart
> final authorization = await user?.authorizationClient.authorizationForScopes(scopes);
> ```

🤦🏻  Yes, that's much simpler.

(I can't reply inline to that comment, I think because I force-pushed so the commit it was on no longer exists.)

packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
```diff
@@ -27,134 +24,142 @@ abstract class GoogleSignInPlatform extends PlatformInterface {
...
-      signInOption: params.signInOption,
-      hostedDomain: params.hostedDomain,
-      clientId: params.clientId,
-    );
-  }
+  /// The default is true, but platforms that cannot support [authenticate] can
+  /// override this to return false, throw [UnsupportedError] from
+  /// [authenticate], and provide a different, platform-specific authentication
+  /// flow.
+  bool supportsAuthenticate() => true;
```
bparrishMines: Why does this default to `true` instead of `false` like we typically do? Also should this not return `Future<bool>`?

packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
```diff
@@ -27,134 +24,142 @@ abstract class GoogleSignInPlatform extends PlatformInterface {
...
+
+  /// Whether or not authorization calls that could show UI must be called from
+  /// a user interaction, such as a button press, on the current platform.
+  ///
+  /// The default is true, but platforms that can fail to show UI without an
```
bparrishMines: I think this is meant to say false:
```suggestion
  /// The default is false, but platforms that can fail to show UI without an
```

packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
```diff
@@ -27,134 +24,142 @@ abstract class GoogleSignInPlatform extends PlatformInterface {
...
-      {required String email, bool? shouldRecoverAuth}) async {
-    throw UnimplementedError('getTokens() has not been implemented.');
-  }
+  /// Signs out previously signed in accounts.
+  Future<void> signOut(SignOutParams params);
 
-  /// Signs out the current account from the application.
-  Future<void> signOut() async {
-    throw UnimplementedError('signOut() has not been implemented.');
+  /// Revokes all of the scopes that all signed in users granted, and then them
```
bparrishMines: This was missing a word
```suggestion
  /// Revokes all of the scopes that all signed in users granted, and then signs them
```

packages/google_sign_in/google_sign_in/lib/google_sign_in.dart
```diff
@@ -70,61 +49,22 @@ class GoogleSignInAccount implements GoogleIdentity {
...
-  Future<Map<String, String>> get authHeaders async {
-    final String? token = (await authentication).accessToken;
-    return <String, String>{
-      'Authorization': 'Bearer $token',
-      // TODO(kevmoo): Use the correct value once it's available from authentication
-      // See https://github.com/flutter/flutter/issues/80905
-      'X-Goog-AuthUser': '0',
-    };
+  /// This returns the authentication information that was returned at the time
+  /// of the initial authentication. Clients are strongly encouraged to use this
```
bparrishMines: nit: Empty space before 2nd sentence

packages/google_sign_in/google_sign_in/lib/google_sign_in.dart
```diff
@@ -155,321 +95,389 @@ class GoogleSignInAccount implements GoogleIdentity {
...
+/// platform and the current application state. Examples include:
+/// - If there is an active authentication session in the application already,
+///   the authorization tokens may be associated for that user.
+/// - If no user has been authenticated, this may trigger a combined
+///   authentication+authorization flow. In that case, whether
+///   [GoogleSignIn]'s authenticationEvents stream will be informed of the
+///   authentication depends on the platform implementation. You should not
+///   assume the user information or authenication tokens will be available in
+///   this case.
+class GoogleSignInAuthorizationClient {
```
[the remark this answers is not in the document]
bparrishMines: I've started using `@internal` from the `meta` package to help with this. But I'm still on the fence whether this is a good pattern or not.

LongCatIsLooong (approved):

packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
stuartmorgan-g: I defaulted it to true since that was the more common case, but since it's a completely new platform interface there's no good reason not to just make it abstract like the rest.

The `bool` vs `Future<bool>` I'm torn on. For both of these support queries, I was having a hard time imagining a case where we would need to determine it dynamically within a platform implementation; `supportsAuthenticate` seems like it will be a basic property of the platform SDK we are using, and `authorizationRequiresUserInteraction` is more fundamental to the platform itself. Normally I say we should err on the side of using `Future` regardless, in case the assumptions are wrong on some future version or platform, but here the tricky part is that the way I expect clients to use `supportsAuthenticate` is in building their UI, and needing an async value in `build` makes things much more complicated, so I'm reluctant to add async to this that we may well never need.

packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
stuartmorgan-g: I made this abstract as well, so I rewrote this paragraph

packages/google_sign_in/google_sign_in_platform_interface/lib/google_sign_in_platform_interface.dart
stuartmorgan-g: Fixed!

packages/google_sign_in/google_sign_in/lib/google_sign_in.dart
stuartmorgan-g: Added.

bparrishMines:
LGTM. Github keeps giving me an error code when I attempt to leave a review (either with a comment or approval).

PR #9268, opened 2025-05-16
Ignore .cxx/ directories

stuartmorgan-g:
Android builds now create .cxx directories in some cases; this has been added to the app template's ignore file, so should be ignored here as well. As is generally the case for this repo, we ignore it at the root level instead of updating every package.


stuartmorgan-g:

gmackall (approved):

PR #9302, opened 2025-05-21
[video_player] Skip some integration tests on web

stuartmorgan-g:
Skips some tests on web; see https://github.com/flutter/flutter/issues/169219


stuartmorgan-g:
This passed once, so it'll at least help. I'm running everything again now to see if it hits other tests on the second try.

tarrinneal (approved):

PR #9306, opened 2025-05-22
[pigeon] Use a const for custom type ids for gobject generated files (#156100)

JesseRiemens:
Adds a custom type identifier to generated gobject headers for the user. Calling fl_value_new_custom_object is now possible with that constant.

This fixes https://github.com/flutter/flutter/issues/156100

## Pre-Review Checklist

- [X] I signed the [CLA].

### DIFF
--- a/packages/pigeon/CHANGELOG.md
+++ b/packages/pigeon/CHANGELOG.md
@@ -1,5 +1,6 @@
-## NEXT
+## 25.4.0
 
+* [gobject] Adds type id constants in header files so that they can be used by the user.
 * Updates minimum supported SDK version to Flutter 3.27/Dart 3.6.
 
 ## 25.3.2
--- a/packages/pigeon/lib/src/gobject/gobject_generator.dart
+++ b/packages/pigeon/lib/src/gobject/gobject_generator.dart
@@ -338,6 +338,28 @@ class GObjectHeaderGenerator
     indent.newln();
     _writeDeclareFinalType(indent, module, _codecBaseName,
         parentClassName: _standardCodecName);
+
+    final Iterable<EnumeratedType> customTypes =
+        getEnumeratedTypes(root, excludeSealedClasses: true);
+
+    if (customTypes.isNotEmpty) {
+      indent.newln();
+      addDocumentationComments(
+          indent,
+          <String>[
+            'Custom type ID constants:',
+            '',
+            'Constants used to identify custom types in the codec.',
+            'They are used in the codec to encode and decode custom types.',
+            'They may be used in custom object creation functions to identify the type.',
+          ],
+          _docCommentSpec);
+    }
+
+    for (final EnumeratedType customType in customTypes) {
+      final String customTypeId = _getCustomTypeId(module, customType);
+      indent.writeln('extern const int $customTypeId;');
+    }
   }
 
   @override
@@ -1019,10 +1041,18 @@ class GObjectSourceGenerator
     _writeDefineType(indent, module, _codecBaseName,
         parentType: 'fl_standard_message_codec_get_type()');
 
+    indent.newln();
+    for (final EnumeratedType customType in customTypes) {
+      final String customTypeId = _getCustomTypeId(module, customType);
+      indent.writeln('const int $customTypeId = ${customType.enumeration};');
+    }
+
     for (final EnumeratedType customType in customTypes) {
       final String customTypeName = _getClassName(module, customType.name);
       final String snakeCustomTypeName =
           _snakeCaseFromCamelCase(customTypeName);
+      final String customTypeId = _getCustomTypeId(module, customType);
+
       indent.newln();
       final String valueType = customType.type == CustomTypes.customClass
           ? '$customTypeName*'
@@ -1030,7 +1060,7 @@ class GObjectSourceGenerator
       indent.writeScoped(
           'static gboolean ${codecMethodPrefix}_write_$snakeCustomTypeName($_standardCodecName* codec, GByteArray* buffer, $valueType value, GError** error) {',
           '}', () {
-        indent.writeln('uint8_t type = ${customType.enumeration};');
+        indent.writeln('uint8_t type = $customTypeId;');
         indent.writeln('g_byte_array_append(buffer, &type, sizeof(uint8_t));');
         if (customType.type == CustomTypes.customClass) {
           indent.writeln(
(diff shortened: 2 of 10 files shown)


packages/pigeon/platform_tests/test_plugin/linux/pigeon/core_tests.gen.h
```diff
@@ -1433,6 +1433,23 @@ G_DECLARE_FINAL_TYPE(CoreTestsPigeonTestMessageCodec,
                      CORE_TESTS_PIGEON_TEST, MESSAGE_CODEC,
                      FlStandardMessageCodec)
 
+const uint8_t core_tests_pigeon_test_an_enum_type_id = 129;
```
stuartmorgan-g, line 1436: I haven't done a full review, but from just skimming the generated code changes:
- Why are these `uint8_t` when the ID is an `int` in the `FlValue` headers?
- Constants should be declared (as `extern`) in a header, and defined in the implementation file, rather than have copies inlined into every file including this header.

JesseRiemens:
The FLValue headers in some places use `int`, and in some places use `uint8_t`:
```c
// fl_value.h:303
FlValue* fl_value_new_custom(int type, // uses int
                             gconstpointer value,
                             GDestroyNotify destroy_notify);

// fl_standard_message_codec.h:148
gboolean fl_standard_message_codec_write_value(FlStandardMessageCodec* codec,
                                               GByteArray* buffer,  // uses GByteArray*, which is used (subclassed) in generated files:
                                               FlValue* value,
                                               GError** error);
// core_tests.gen.cc:2422
static gboolean
core_tests_pigeon_test_message_codec_write_core_tests_pigeon_test_an_enum(
    FlStandardMessageCodec* codec, GByteArray* buffer, FlValue* value,
    GError** error) {
  uint8_t type = core_tests_pigeon_test_an_enum_type_id; // appends uint8_t to buffer
  g_byte_array_append(buffer, &type, sizeof(uint8_t));
  return fl_standard_message_codec_write_value(codec, buffer, value, error);
}
```
I think this might be related to [#152916](https://github.com/flutter/flutter/issues/152916). Maybe it is best to leave it as is, and for this PR use `int` for the types.

stuartmorgan-g:
> The FLValue headers in some places use `int`, and in some places use `uint8_t`:

Where is the header using `uint8_t` for the type? I'm seeing in your snippet:
- A header declaration that uses `int` for the type.
- A header declaration that doesn't have the type at all.
- A file that's not an engine header.

JesseRiemens:
My mistake, it's not the FlValue header that uses uint8_t, but only the generated code using the engine header. 
I did not fully understand how it worked, but now I see that the generated dart code at the other side also uses `uint8`:
```dart
          indent.writeln('buffer.putUint8(4);');
```
[dart_generator.dart:432](https://github.com/flutter/packages/blob/5a7d40f877274916bdd81ba18a6b508e2d266aa3/packages/pigeon/lib/src/dart/dart_generator.dart#L432)

That makes more sense to me now. I guess that means we can keep it like it is.

JesseRiemens:
@stuartmorgan-g  
Hey, I don't want to be pushy, but could you please review this PR?

packages/pigeon/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## NEXT
+## 25.4.0
 
-* Updates minimum supported SDK version to Flutter 3.27/Dart 3.6.
+* [gobject] Adds type id constants in header files so that they can be used by the user.
+* Updates minimum supported SDK version to Flutter 3.24/Dart 3.5.
```
tarrinneal: Why did the min sdk version change?

tarrinneal (commented):
Pr looks good, just one question and I can lgtm

packages/pigeon/CHANGELOG.md
JesseRiemens: Probably an old version of the tree, I rebased and removed that change.

packages/pigeon/lib/src/gobject/gobject_generator.dart
```diff
@@ -338,6 +338,15 @@ class GObjectHeaderGenerator
     indent.newln();
     _writeDeclareFinalType(indent, module, _codecBaseName,
         parentClassName: _standardCodecName);
+
+    final Iterable<EnumeratedType> customTypes =
+        getEnumeratedTypes(root, excludeSealedClasses: true);
+
+    for (final EnumeratedType customType in customTypes) {
+      final String customTypeId = _getCustomTypeId(module, customType);
+      indent.newln();
```
stuartmorgan-g: Nit: Let's remove this; it will make more sense as a block of constants.

packages/pigeon/lib/src/gobject/gobject_generator.dart
```diff
@@ -338,6 +338,15 @@ class GObjectHeaderGenerator
     indent.newln();
     _writeDeclareFinalType(indent, module, _codecBaseName,
         parentClassName: _standardCodecName);
+
+    final Iterable<EnumeratedType> customTypes =
+        getEnumeratedTypes(root, excludeSealedClasses: true);
+
```
stuartmorgan-g, line 344: We should output an explanatory comment (if customTypes is not empty) about what the constants are (one comment for the whole block, not one per type).

packages/pigeon/lib/src/gobject/gobject_generator.dart
```diff
@@ -1023,14 +1032,19 @@ class GObjectSourceGenerator
       final String customTypeName = _getClassName(module, customType.name);
       final String snakeCustomTypeName =
           _snakeCaseFromCamelCase(customTypeName);
+      final String customTypeId = _getCustomTypeId(module, customType);
+
+      indent.newln();
+      indent.writeln('const int $customTypeId = ${customType.enumeration};');
```
stuartmorgan-g: These should be output all at once in a block rather than interleaved with all of the other code. It's more important for the output file to be easier to understand than to save a couple of lines of code in the generator.

stuartmorgan-g (changes requested):

packages/pigeon/lib/src/gobject/gobject_generator.dart
JesseRiemens: I put a comment, but it's not exactly in gdoc style as there's no direct symbol to refer to. (as the comment concerns a block of consts.) 

If you don't like it, please give me a suggestion and I'll put it in. 

```c
/**
 * Custom type ID constants:
 *
 * Constants used to identify custom types in the codec.
 * They are used in the codec to encode and decode custom types.
 * They may be used in custom object creation functions to identify the type.
 */
```

JesseRiemens:
If you'd like to compare, [this commit](https://github.com/flutter/packages/commit/75a101f72801660d911d62036e7eceea5e54bb32) contains the changes since the last review

stuartmorgan-g (approved):
LGTM

stuartmorgan-g:
@tarrinneal This just needs your final re-review.

JesseRiemens:
@tarrinneal Can you please autosubmit / merge?

PR #9332, opened 2025-05-28
[go_router] Fix replace routes hanging the originating completer

nouvist:
The behavior of `replace` and `pushReplacement` involves popping the last route and ignoring the completer, leaving the last route's future unresolved. This PR introduces new behavior for both methods to inherit the last route's completer if the last route is an `ImperativeRouteMatch`.

Expected issue(s) should be fixed by this:
- https://github.com/flutter/flutter/issues/141251

### Previous behavior:
https://github.com/user-attachments/assets/add06187-9bee-4347-8eac-e129c7a53457

### This PR behavior:
https://github.com/user-attachments/assets/98c20fe3-5fea-4db2-ae95-747950650172

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/go_router/lib/src/information_provider.dart
+++ b/packages/go_router/lib/src/information_provider.dart
@@ -36,6 +36,13 @@ enum NavigatingType {
   restore,
 }
 
+/// Maps the replacement route result to the correct generic type.
+typedef MapRouteResultCallback<T extends Object?> = T Function(Object? result);
+
+/// Pipe completer to the last route match completer.
+typedef PipeRouteCompleterCallback = Completer<Next?>
+    Function<Next extends Object?>(Completer<Next?> next);
+
 /// The data class to be stored in [RouteInformation.state] to be used by
 /// [GoRouteInformationParser].
 ///
@@ -49,9 +56,16 @@ class RouteInformationState<T> {
     this.completer,
     this.baseRouteMatchList,
     required this.type,
+    MapRouteResultCallback<T?>? mapReplacementResult,
   })  : assert((type == NavigatingType.go || type == NavigatingType.restore) ==
             (completer == null)),
-        assert((type != NavigatingType.go) == (baseRouteMatchList != null));
+        assert((type != NavigatingType.go) == (baseRouteMatchList != null)),
+        mapReplacementResult =
+            mapReplacementResult ?? _defaultMapReplacementResult;
+
+  static T? _defaultMapReplacementResult<T>(Object? result) {
+    return null;
+  }
 
   /// The extra object used when navigating with [GoRouter].
   final Object? extra;
@@ -63,6 +77,10 @@ class RouteInformationState<T> {
   /// [NavigatingType.restore].
   final Completer<T?>? completer;
 
+  /// Maps the replacement result to the appropriate type, to resolve the
+  /// [completer].
+  final MapRouteResultCallback<T?> mapReplacementResult;
+
   /// The base route match list to push on top to.
   ///
   /// This is only null if [type] is [NavigatingType.go].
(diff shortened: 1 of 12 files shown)


ikurek:
@nouvist LGTM, thank you for submitting the fix, I was facing the same issue

packages/go_router/lib/src/parser.dart
```diff
@@ -168,6 +168,18 @@ class GoRouteInformationParser extends RouteInformationParser<RouteMatchList> {
     return redirectedFuture;
   }
 
+  /// Ensures the replacement routes can resolve the originating route completer.
+  Completer<T?>? _createInheritedCompleter<T>(
```
chunhtai: What if the type T mismatched for the new route and the old route. Instead of inherited the completer, I think it should resolve the previous route's future because it is replaced.

nouvist: Should it resolve when `replace` is invoked, or resolve when `pop` is invoked, similar to my PR but returning null instead? Alternatively, what if I propose that `replace` could have an optional callback that map the replace route result to the correct `T`?

```dart
Future<T?> replace<T>(
  String location, {
  Object? extra,
  T? Function(Object?)? mapResult, // default map to null if not defined
});
```

I think I prefer this behavior because the waiting route can wait until the route screen appears again. If it resolves when `replace`, the waiting route might operate while the screen is not yet displayed.

nouvist: Yeah, that's the problem, but I still think resolve the completer when the replacement pops up is my preferred, as the waiting route can hold until its screen reappears.

I propose, what if the `push` (etc) had a `mapReplacementResult`? But it may need to be on `RouteInformationState` and `ImperativeRouteMatch`, which I'm not sure if there will be breaking changes or not.

packages/go_router/example/devtools_options.yaml
```diff
@@ -0,0 +1,3 @@
+description: This file stores settings for Dart & Flutter DevTools.
```
chunhtai, line 1: this file seems not needed?

packages/go_router/lib/src/information_provider.dart
```diff
@@ -196,8 +232,12 @@ class GoRouteInformationProvider extends RouteInformationProvider
 
   /// Removes the top-most route match from `base` and pushes the `location` as a
   /// new route on top.
-  Future<T?> pushReplacement<T>(String location,
-      {required RouteMatchList base, Object? extra}) {
+  Future<T?> pushReplacement<T>(
+    String location, {
+    required RouteMatchList base,
+    Object? extra,
+    MapRouteResultCallback<T?>? mapReplacementResult,
```
chunhtai, line 239: This feels weird, If we want to separate two use cases.
1. resolve the result immediately in pushReplacement
2. replace the future in pushReplacement.

I think we should make two different API, maybe a pushReplacement for (1) and a pushReplaceResultFuture for (2), and in (2) we should force the type to be the same as previously pushed route.

packages/go_router/example/devtools_options.yaml
nouvist: didn't realize it added, I will remove

packages/go_router/lib/src/information_provider.dart
nouvist: case (2) that match its originating type is guaranteed. this implementation makes the caller of the originating route is the one who's responsible for replacing future. for example, A is the one that responsible for replacing C result to match with B result type when it's pushing B. that's why `mapReplacementResult` is required for each pushed new route, because the assumption is any route can be replaced.
```
Route A -- (pushing) --> Route B -- (replaced by) --> Route C
```

---

I do consider to make the `replace` and `pushReplacement` itself that responsible for replacing future. but I struggle to make it match the originating route type. I think it's similar to `pop` where we have no idea what type it should be.

I didn't consider case (1) before, it is indeed necessary to implement. but then, when routes can be resolve immediately when replaced, it needs value to return, and then again, we have no idea what type it should be, like pop is.

justinmc:
@nouvist Do you  still plan to return to this PR? FYI there are merge conflicts now.

justinmc:
@nouvist I'm going to close this PR as abandoned, but let me know if you end up getting time to return to it and want to reopen.

PR #9361, opened 2025-06-03
Removes unnecessary boolean operations

FMorschel:
I'm fixing https://github.com/dart-lang/sdk/issues/60614, which adds some other cases to `no_literal_bool_comparisons`, and these were triggered here.

I'm unsure of the intended behaviour of the `packages\url_launcher\url_launcher\lib\src\legacy_api.dart` part, but I can fix it if the result wasn't intended this way (the operators' priority may be less known here).

About the other cases with `|| true`, I think these might be a pattern from the team, so maybe adding a file ignore could also fix this. Please say so that I can make this the right way for you.

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/rfw/test/argument_decoders_test.dart
```diff
@@ -380,7 +380,7 @@ void main() {
       matchesGoldenFile('goldens/argument_decoders_test.containers.png'),
       // TODO(louisehsu): Unskip once golden file is updated. See
       // https://github.com/flutter/flutter/issues/151995
-      skip: !runGoldens || true,
+      skip: !runGoldens,
```
stuartmorgan-g: This change re-enabled a failing test, and needs to be reverted.

The `|| true` pattern here is to not lose the original condition while the test is temporarily skipped.

packages/rfw/test/material_widgets_test.dart
```diff
@@ -213,7 +213,7 @@ void main() {
       matchesGoldenFile('goldens/material_test.scaffold.png'),
       // TODO(louisehsu): Unskip once golden file is updated. See
       // https://github.com/flutter/flutter/issues/151995
-      skip: !runGoldens || true,
+      skip: !runGoldens,
```
stuartmorgan-g: Here as well you have enabled a bunch of broken tests.

packages/url_launcher/url_launcher/lib/src/legacy_api.dart
```diff
@@ -78,7 +78,7 @@ Future<bool> launch(
   final bool isWebURL =
       url != null && (url.scheme == 'http' || url.scheme == 'https');
 
-  if ((forceSafariVC ?? false || forceWebView) && !isWebURL) {
+  if ((forceSafariVC ?? forceWebView) && !isWebURL) {
```
stuartmorgan-g: You have changed the behavior of this code; `forceSafariVC = false` and `forceWebView = true` now evaluates, incorrectly, to `false`.

I don't understand how this tripped the lint in the first place. What is wrong with using `?? <default value>` with a nullable boolean? Is there a bug in the linter?

FMorschel: Not as I can see. The precedence in the operators (as I tipped in the PR description) actually does this.

The first thing that will occur is `false || forceWebView` and only then `forceSafariVC ?? expr`. 

https://github.com/dart-lang/sdk/blob/da5ae4786cc41812eee6f83ed1e9c5c05c15f387/pkg/_fe_analyzer_shared/lib/src/scanner/token.dart#L21-L22

packages/rfw/test/argument_decoders_test.dart
FMorschel: I see. I'll add the ignore for the file then. Thanks for the explanation.

packages/url_launcher/url_launcher/lib/src/legacy_api.dart
stuartmorgan-g: Huh, that's very unintuitive. That means that the code was unfortunately incorrect and we just hadn't noticed. Linter improvement to the rescue!

In that case the fix for the lint here is to add the parens to make the code work as intended, where `forceSafariVC ?? false` is one logical unit, and to add a test [like this one](https://github.com/flutter/packages/blob/e1919381a0cc2a947da87d69e59b8dd14608e580/packages/url_launcher/url_launcher/test/src/legacy_api_test.dart#L218-L230) for the broken case to make sure we don't regress it later.

FMorschel: Sorry for the long delay. I came back to this today. I (fortunately) don't think that any cases are broken.

There would only be a problem if either the `forceWebView` or `forceSafariVC` default was `true`.

When adding the change you suggested on the last comment, no test broke, and I could not think of a test that would break before/after this change. If you can, please say so that I can add it.

FMorschel:
I'm continuing this at https://github.com/flutter/packages/pull/9409 (more explanation there). Sorry for the confusion.

PR #9379, opened 2025-06-05
Roll Flutter master to d8baa77b3

stuartmorgan-g:
Rolls Flutter master version to d8baa77b38461e7061e06e72c6bf50d64d302b8c

Manual version of https://github.com/flutter/packages/pull/9346

Skips a `quick_actions_android` test for now to unblock the roller; see https://github.com/flutter/flutter/issues/170141


stuartmorgan-g:
The quick_actions crash appears to be a real regression, not flake.

stuartmorgan-g:
I'm using this PR rather than the minimal roll PR I narrowed down to because the only failure here was the quick_action_android failure, so landing this one will get us caught up a bit further. (We'll still be really behind, so I'll do more follow-ups after this lands.)

PR #9395, opened 2025-06-06
[pigeon] Create a message call free InstanceManager when running unit tests

bparrishMines:
To support hot restart, `PigeonInstanceManager.instance` makes a call to clear the native `InstanceManager` when it is instantiated.

This was the source of unit test race condition flakes like https://github.com/flutter/flutter/issues/164132. 

The current workaround is to `try/catch` the async `PlatformException` at the top of every test file or pass a test `InstanceManager` to every `ProxyApi` class: 

```dart
final MyNativeClass object = MyNativeClass(
  pigeon_instanceManager: PigeonInstanceManager(
    onWeakReferenceRemoved: (_) {}
  ),
);
```

This PR changes it to [detect a test ran as Flutter unit test](https://api.flutter.dev/flutter/flutter_test/TestWidgetsFlutterBinding/ensureInitialized.html) and sets the default `InstanceManager` with one that doesn't make any message calls.

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/pigeon/platform_tests/shared_test_plugin_code/test/instance_manager_test.dart
```diff
@@ -167,7 +167,7 @@ void main() {
       instanceManager.addDartCreatedInstance(object);
 
       object = null;
-      await forceGC();
+      await forceGC(fullGcCycles: 2);
```
bparrishMines, line 170: This test was failing on windows, so I added another cycle.

tarrinneal (approved):

packages/pigeon/lib/src/dart/templates.dart
```diff
@@ -87,6 +87,9 @@ class $dartInstanceManagerClassName {
   late final void Function(int) onWeakReferenceRemoved;
 
   static $dartInstanceManagerClassName _initInstance() {
+    if (Platform.environment['FLUTTER_TEST'] == 'true') {
```
stuartmorgan-g, line 90: Hm, is `FLUTTER_TEST` only set in unit tests? Or will this break `flutter test` integration test behavior?

bparrishMines: I forgot that integration tests also use `flutter test`. But I just tried it and it looks like this doesn't affect integration tests. It seems the environment variable is only for unit tests.

bparrishMines: This is probably the case because integration tests still need to use the real BinaryMessenger. When `FLUTTER_TEST` is set, it uses a test BinaryMessenger from [TestWidgetsFlutterBinding](https://api.flutter.dev/flutter/flutter_test/TestWidgetsFlutterBinding-class.html).

Edit:
Actually I found a bit of info on it: https://api.flutter.dev/flutter/flutter_test/LiveTestWidgetsFlutterBinding-class.html. It looks like integration tests from `flutter test` is treated the same way as running tests with `flutter run`.

stuartmorgan-g: Sounds good; LGTM then.

I wish we didn't have to embed special test logic in the production code, but it seems like the best option we have at the moment.

bparrishMines: Hmm would it be better to copy https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/foundation/_platform_io.dart#L30 then. And wrap the `Platform.environment` check in an `assert` so that it gets ignored in release mode.

```dart
static PigeonInstanceManager _initInstance() {
    PigeonInstanceManager? instanceManager;
    assert(() {
      if (Platform.environment.containsKey('FLUTTER_TEST')) {
        instanceManager = PigeonInstanceManager(onWeakReferenceRemoved: (_) {});
      }
      return true;
    }());
    if (instanceManager == null) {
      // ...
    }
    return instanceManager!;
  }
```

PR #9402, opened 2025-06-06
Roll Flutter master to 2d9977c14ebe

stuartmorgan-g:
Rolls Flutter master version to 2d9977c14ebe9f181c99b00ffe2b632116f70099

Manual version of https://github.com/flutter/packages/pull/9371

Skips a `quick_actions_android` test for now to unblock the roller; see https://github.com/flutter/flutter/issues/170141

PR #9407, opened 2025-06-09
Roll Flutter master to 8b22f67c85114

stuartmorgan-g:
Rolls the Flutter master pin to 8b22f67c85114def3f090286c386627efdf59614

https://github.com/flutter/flutter/compare/d8baa77b3846...8b22f67c8511


stuartmorgan-g:
This is a subset of https://github.com/flutter/packages/pull/9402/ to narrow down the range where the various failures there started.

stuartmorgan-g:
It looks like this narrow range has the iOS quick actions failure, as expected, and also a Pigeon integration test failure that's also in the larger-range roll PR:

```
00:21 +0 -1: loading /Volumes/Work/s/w/ir/x/w/packages/packages/pigeon/example/app/integration_test/example_app_test.dart [E]                                                                          
  Failed to load "/Volumes/Work/s/w/ir/x/w/packages/packages/pigeon/example/app/integration_test/example_app_test.dart": WebSocketChannelException: SocketException: Connection refused (OS Error: Connection refused, errno = 61), address = 127.0.0.1, port = 55185
```

I'm not sure what would be causing that.

stuartmorgan-g:
From local testing, https://github.com/flutter/flutter/pull/169365 broke the Pigeon example, although the error message is slightly different if I check out that hash (maybe the follow-up to do a runtime warning changed the behavior slightly). That suggests that the automigration is not always safe. /cc @gaaclarke

stuartmorgan-g:
Oh, I see. The example is doing
```swift
let controller = window?.rootViewController as! FlutterViewController
```

so it's crashing. This is flagged by the runtime warning if I `flutter run`.

stuartmorgan-g:
This is going to be a problem. Fixing this within the context of our CI, which runs against both `stable` and `master`, is non-trivial, because the symbols I need (`FlutterPluginRegistrant` and friends) don't existing on `stable`. I'm trying to figure out if there's a solution that doesn't involve re-adding the pre-compilation file rewriting system we had briefly several years ago (IIRC) for a somewhat similar situation on the Android side.

gaaclarke:
> Fixing this within the context of our CI, which runs against both stable and master

You can mimic the behavior of the FlutterPluginRegistrant for now by subclassing the view controller and putting your setup code in the awakeFromNib method.

stuartmorgan-g:
Thanks, I was trying to figure out what the least disruptive way to get a call at the right time was. A custom view controller makes sense!

stuartmorgan-g:
Now investigating the video_player_avfoundation failures, which I can repro locally.

stuartmorgan-g:
```objc
// TODO(stuartmorgan): Convert to using mock registrars instead.
NSObject<FlutterPluginRegistry> *GetPluginRegistry(void) {
#if TARGET_OS_IOS
  return (NSObject<FlutterPluginRegistry> *)[[UIApplication sharedApplication] delegate];
#else
  return (FlutterViewController *)NSApplication.sharedApplication.windows[0].contentViewController;
#endif
}
```

🤦🏻

stuartmorgan-g:
@bparrishMines I believe this will be landable now (modulo flake) as another incremental to continue working toward getting the roller caught up and healthy again.

bparrishMines (approved):
LGTM

tarrinneal (approved):

stuartmorgan-g:
Overrides: the changes to the example app aren't relevant to clients.

stuartmorgan-g:
(Force pushed due to https://github.com/flutter/flutter/issues/170300)

stuartmorgan-g:
This `testBlankVideoBugWithEncryptedVideoStreamAndInvertedAspectRatioBugForSomeVideoStream` failure may be real? It's at least a frequent flake since it showed up in the last run too. It didn't repro locally for me though.

```
Test Case '-[VideoPlayerTests testBlankVideoBugWithEncryptedVideoStreamAndInvertedAspectRatioBugForSomeVideoStream]' started.
2025-06-09 15:55:24.753451-0700 video_player_avfoundation_example[53023:1014858] [aqme]        LoudnessManager.mm:413   PlatformUtilities::CopyHardwareModelFullName() returns unknown value: Macmini9,1, defaulting hw platform key
/Volumes/Work/s/w/ir/x/w/packages/packages/video_player/video_player_avfoundation/darwin/RunnerTests/VideoPlayerTests.m:216: error: -[VideoPlayerTests testBlankVideoBugWithEncryptedVideoStreamAndInvertedAspectRatioBugForSomeVideoStream] : ((player.playerLayer.superlayer) != nil) failed - AVPlayerLayer should be added on screen.
Test Case '-[VideoPlayerTests testBlankVideoBugWithEncryptedVideoStreamAndInvertedAspectRatioBugForSomeVideoStream]' failed (0.061 seconds).
```

stuartmorgan-g:
Maybe a race condition. We are getting what should end up being the superlayer via `UIApplication.sharedApplication.keyWindow.rootViewController`, and with the delegate migration maybe that's not reliably available when the test runs.

stuartmorgan-g:
`video_player_avfoundation` should get a re-review; I changed the production code to allow an injectable view provider rather than iOS hard-coding getting it from the app's root view controller.

PR #9408, opened 2025-06-09
[google_maps_flutter_platform_(web/android/ios)] Add a new zIndexInt param to marker and deprecate zIndex

Hari-07:
This is just the `web`, `android` and `ios` package part of #8012

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
```diff
@@ -730,3 +730,13 @@ gmaps.LatLng _pixelToLatLng(gmaps.Map map, int x, int y) {
 
   return projection.fromPointToLatLng(point)!;
 }
+
+extension on Marker {
```
stuartmorgan-g: Extensions violate Flutter's style guide. There are cases where we have to use them for JS interop, but that doesn't apply here.

Why do we need this at all, vs just using `zIndex` with a comment like in the Android example?

packages/google_maps_flutter/google_maps_flutter_web/pubspec.yaml
```diff
@@ -2,11 +2,11 @@ name: google_maps_flutter_web
 description: Web platform implementation of google_maps_flutter
 repository: https://github.com/flutter/packages/tree/main/packages/google_maps_flutter/google_maps_flutter_web
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+maps%22
-version: 0.5.12
+version: 0.5.13
```
stuartmorgan-g: This should be 0.5.12+1 as it's a very small change.

packages/google_maps_flutter/google_maps_flutter_android/example/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ publish_to: none
 
 environment:
   sdk: ^3.6.0
-  flutter: ">=3.27.0"
+  flutter: '>=3.27.0'
```
stuartmorgan-g: Why does this PR change the style of all the string values in pubspec.yaml files to deviate from our usual style?

Hari-07: I assume there might be something wrong with my linte, will fix that. I had to do it manually for the interface package

packages/google_maps_flutter/google_maps_flutter_web/lib/src/convert.dart
Hari-07: Right. IIRC this was written when 2 different parameters were there initially, now that we don't allow the constructor to pass in both at the same time, this isn't required

packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml
```diff
@@ -21,7 +21,7 @@ dependencies:
   flutter:
     sdk: flutter
   flutter_plugin_android_lifecycle: ^2.0.1
-  google_maps_flutter_platform_interface: ^2.11.0
+  google_maps_flutter_platform_interface: ^2.12.1
```
stuartmorgan-g: Looking at the PR again... do we actually need any of the changes to this file? It looks like the only actual change is to the integration tests and the example pubspec, and those don't require publishing the package.

packages/google_maps_flutter/google_maps_flutter_ios/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: google_maps_flutter_ios
 description: iOS implementation of the google_maps_flutter plugin.
 repository: https://github.com/flutter/packages/tree/main/packages/google_maps_flutter/google_maps_flutter_ios
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+maps%22
-version: 2.15.2
+version: 2.15.3
```
stuartmorgan-g: Similarly, the only change to this package is to a file in the example app that's not the one visible on pub.dev, so this should be exempt from versioning.

packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml
Hari-07: In that case Im gonna close this PR. And since it doesn't need a new version, the main PR's overrides can all be removed.

Hari-07: Doesn't it need to target the updated platform interface package though? Not sure if Im misunderstanding this line from the federated packages section 

> If new functionality is being added to the API surface of the app-facing package, be sure to update the version constraints of the implementation packages in its pubspec.yaml

packages/google_maps_flutter/google_maps_flutter_ios/pubspec.yaml
stuartmorgan-g: Although, why is only the example changing for iOS? Shouldn't it be using `zIndexInt` instead of the deprecated `zIndex` now, since it can't use any fractional values anyway?

packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml
stuartmorgan-g: > Doesn't it need to target the updated platform interface package though?

What is "it" here? If it's `google_maps_flutter_android`, what part of the `google_maps_flutter_android` requires the new APIs added to the platform interface?

If it's the app-facing package (in the other PR), yes, because it's exporting that surface to clients.

> Not sure if Im misunderstanding this line from the federated packages section
> 
> > If new functionality is being added to the API surface of the app-facing package, be sure to update the version constraints of the implementation packages in its pubspec.yaml

See the explanation in the following sentence for the reason this would need to be done.

For example, if the platform interface adds a new `Future<void> foo() { throw UnimplementedError(); }`, and all the platforms are updated to implement `foo()`, and the app-facing package is updated to call `foo()`, we don't want a valid set of resolved packages where clients can get the updated version of the app-facing and platform interface packages, but *not* the implementation packages, such that everything compiles but all calls to `foo()` throw.

packages/google_maps_flutter/google_maps_flutter_ios/pubspec.yaml
Hari-07: You're right, not sure how that ended up getting removed or at what point. I think the reasonable thing to do would be to make ios's `PlatformMarker` expect an integral zIndex, so that this truncation would now be a type error to which we can pass `zIndexInt`

packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml
Hari-07: Ok got it. For android we dont't need a new version. I guess even web doesn't need a new version by the same logic since that's just a comment change?

stuartmorgan-g: > I guess even web doesn't need a new version by the same logic since that's just a comment change?

Correct; IIRC the tooling will do the right thing with a comment-only change, but if not we can override the version check.

packages/google_maps_flutter/google_maps_flutter_ios/pubspec.yaml
stuartmorgan-g: Agreed, the iOS changes look good here.

Based on the CI failure though, it appears that the Obj-C compiler flags 64-bit int assignment to 32-bit int assignment as lossy, but did *not* flag it when it was assigning a double to a 32-bit int. Odd, but you'll need to add an explicit cast [here](https://github.com/flutter/packages/blob/2930e2012d1425514d5c56d19b453e21a66eaed1/packages/google_maps_flutter/google_maps_flutter_ios/ios/Classes/GoogleMapMarkerController.m#L124) to make the analysis check happy.

Hari-07: That's seems reasonable from flags.
`-Wno-float-conversion` looks like it allows implicit float conversion which is what was happening
`-Wshorten-64-to-32` would disallow the long to int cast that's happening at the line you pointed out

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
stuartmorgan-g, line 1: The changes to this file can be reverted as well; we don't need to CHANGELOG an integration test comment.

stuartmorgan-g (approved):
LGTM

stuartmorgan-g:
@vashworth / @cbracken could you give the second approval here since the substantive change is iOS?

PR #9409, opened 2025-06-09
[url_launcher] Fixes new unnecessary boolean operations warnings

FMorschel:
I had opened https://github.com/flutter/packages/pull/9361 and messed up when rebasing the changes. That made the old PR get closed.

CC @stuartmorgan-g for continuing the review.

Description from the old PR:

I'm fixing https://github.com/dart-lang/sdk/issues/60614, which adds some other cases to `no_literal_bool_comparisons`, and these were triggered here.

I'm unsure of the intended behaviour of the `packages\url_launcher\url_launcher\lib\src\legacy_api.dart` part, but I can fix it if the result wasn't intended this way (the operators' priority may be less known here).

About the other cases with `|| true`, I think these might be a pattern from the team, so maybe adding a file ignore could also fix this. Please say so that I can make this the right way for you.

## Pre-Review Checklist

- [x] I signed the [CLA].


FMorschel:
Continuing from my last message:

Sorry for the long delay. I came back to this today. I (fortunately) don't think that any cases are broken.

There would only be a problem if either the `forceWebView` or `forceSafariVC` default was `true`.

When adding the change you suggested on the last comment, no test broke, and I could not think of a test that would break before/after this change. If you can, please say so that I can add it.

stuartmorgan-g:
> I (fortunately) don't think that any cases are broken.
> 
> There would only be a problem if either the `forceWebView` or `forceSafariVC` default was `true`.

Why would a problem only happen with defaults? Clients can pass non-default values.

> When adding the change you suggested on the last comment, no test broke

I'm aware that your PR didn't break any tests; that's why I said we needed to add a test. The lack of test breakage with the previous version indicates missing test coverage.

> and I could not think of a test that would break before/after this change. If you can, please say so that I can add it.

As I said in the previous review:
>> forceSafariVC = false and forceWebView = true

A test asserting that this call throws should pass:
```dart
launch('nonwebscheme://', forceSafariVC: false, forceWebView: true);
```

FMorschel:
Alright, I see where I made the mistake in my consideration. Sorry, I'll add the test. Thanks!

stuartmorgan-g:
> Since it would only affect the last two cases, it would make them the middle cases and the outputs are exactly the same.

That is not correct.

Have you actually run the test I suggested before and after the PR?

FMorschel:
> Have you actually run the test I suggested before and after the PR?

Now I have. I'm really sorry I got confused on my last comment.

I'm adding the new test now. Thanks for your patience!

stuartmorgan-g (changes requested):
url_launcher needs a version bump, per the CI failure, since there are changes to production code.

FMorschel:
I've bumped the version already @stuartmorgan-g.

I've also added `url_launcher` to the PR name since this is the only package that had any significant changes.

Is the version bump I did enough? Am I missing anything still? Thanks!

stuartmorgan-g:
> Am I missing anything still?

Yes; all of [the failures flagged by CI](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8712342398102760305/+/u/Run_package_tests/CHANGELOG_and_version_validation/stdout) need to be resolved.

FMorschel:
I've added the version number on the Changelog, and a new entry about the new case that will throw.

1. One of the tests [timed out](https://ci.chromium.org/ui/p/flutter/builders/try/Linux_android%20android_build_all_packages%20stable/17710/overview).
2. I need an exemption for `in_app_purchase`. The only change there was adding the ignore comment.

stuartmorgan-g:
Version/CHANGELOG override: Although the change is to example/lib/main.dart, and thus usually requires a version change since it's published on pub.dev, the `// ignore` isn't important for the use case of using the code from pub.dev since it's analysis-option specific.

FMorschel:
Hey @Hixie and @bparrishMines can I get some reviews here please? The changes are really minimal. Simply adding an ignore comment.

Thanks a lot!

stuartmorgan-g:
@bparrishMines for secondary review; please add `autosubmit` once it's approved.

bparrishMines (approved):
LGTM

PR #9413, opened 2025-06-10
Roll Flutter master to 7e30df2225

stuartmorgan-g:
Rolls Flutter master pin to 7e30df2225f65141e0227e424986311b79d39505

https://github.com/flutter/flutter/compare/8b22f67c85114...7e30df2225f


stuartmorgan-g:
It looks like all web integration test shards are hanging on the first test, so something is very broken in this range.

stuartmorgan-g:
This is almost certainly https://github.com/flutter/flutter/pull/169174:

> This change avoids enabling hot reload in the flutter drive tests since they rely on
-d web-server which has known startup issues. When https://github.com/dart-lang/sdk/issues/60289 is
resolved it should be safe to enable hot reload by default for the flutter drive
tests.

script/tool/lib/src/drive_examples_command.dart
```diff
@@ -140,6 +140,11 @@ class DriveExamplesCommand extends PackageLoopingCommand {
         platformWeb: <String>[
           '-d',
           'web-server',
+          // TODO(stuartmorgan): Remove once drive works without it. See
+          // https://github.com/dart-lang/sdk/issues/60289 and
+          // https://github.com/flutter/flutter/pull/169174
+          if (platform.environment['CHANNEL']?.toLowerCase() == 'master')
+            '--no-web-experimental-hot-reload',
```
mdebbar: Why the `if`?

mdebbar: Oh, is this to omit the flag when running against `stable`? Did the flag not make it into stable yet?

stuartmorgan-g: I assumed it was recent (like the breakage) and hadn't, but I just checked the source and it looks like it's in 3.32. I'll remove the if.

PR #9416, opened 2025-06-11
Roll Flutter master to 824868f5d1e6c091a5df23d3628407e75149e1da

stuartmorgan-g:
Rolls Flutter master version to 824868f5d1e6c091a5df23d3628407e75149e1da

https://github.com/flutter/flutter/compare/2d9977c14e...824868f5d1e6

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -38,12 +38,15 @@ platform_properties:
       os: Ubuntu
       cores: "8"
       device_type: none
+      # The android_virtual_device here should be as old as is feasible. Currently it's 31
+      # because the older devices available (see link above for options) are all x86, which
+      # is no longer supported by Flutter.
       dependencies: >-
         [
           {"dependency": "android_sdk", "version": "version:33v6"},
           {"dependency": "open_jdk", "version": "version:17"},
           {"dependency": "curl", "version": "version:7.64.0"},
-          {"dependency": "android_virtual_device", "version": "generic_android22.textpb"},
+          {"dependency": "android_virtual_device", "version": "android_31_google_apis_x64.textpb"},
           {"dependency": "avd_cipd_version", "version": "build_id:8733065022087935185"}
         ]
   linux_desktop:


stuartmorgan-g:
Android failures are probably https://github.com/flutter/flutter/pull/169884

```
Package install error: Failure [INSTALL_FAILED_NO_MATCHING_ABIS]
```

stuartmorgan-g:
Trying the emulator version from https://github.com/flutter/flutter/pull/158017, which it looks like at least some flutter/flutter configs are still using.

stuartmorgan-g:
Wait, that's the version we're already using. Now I'm confused.

@reidbaker Any idea why we're getting emulator failures suggesting we're using x86 emulators if the `avd_cipd_version` is the same as in flutter/flutter?

stuartmorgan-g:
Oh, never mind, it's the legacy tests that are failing, so the issue is `{"dependency": "android_virtual_device", "version": "generic_android22.textpb"}`, which is indeed x86.

stuartmorgan-g:
Hm, it looks from https://chromium.googlesource.com/chromium/src.git/+/HEAD/tools/android/avd/proto like API 30 or 31 is the oldest we can now go with the "legacy" tests. We're going to lose a lot of safety about continuing to work on older versions of Android.

stuartmorgan-g:
@reidbaker Do you have any better ideas than us making API 31 our "old" test device?

stuartmorgan-g:
> We're going to lose a lot of safety about continuing to work on older versions of Android.

(The counterpoint is that I don't think we've actually had many cases of things that only failed these tests.)

reidbaker:
Is 31 the oldest device that has a non x86 image? If you haven't checked then I can.

stuartmorgan-g:
From the link above, the only older one I can find is `android_30_google_atd_x64`. I'm not clear what the `atd` vs `api` distinction is though.

reidbaker:
Our only other option would be authoring one of those proto files and getting chromium to accept it. 

https://github.com/chromium/chromium/blob/main/docs%2Fandroid_emulator.md

stuartmorgan-g:
In that case, since the roller is currently over a week behind and this is a blocker, I would like to land this rather than block for an unknown amount of time. Any objection to that?

Should I file a `team-android` bug for investigating configuring and switching to an older x64 device as a follow-up?

reidbaker:
No objection to setting the old tests to api 31

stuartmorgan-g:
Thanks! I filed https://github.com/flutter/flutter/issues/170474 to make sure we don't forget to at least consider adjusting it later.

PR #9417, opened 2025-06-11
Roll Flutter master to 2eec076321fb

stuartmorgan-g:
Rolls Flutter master version to 2eec076321fb4ea99a461fd18e540e7ed110eb2c

https://github.com/flutter/flutter/compare/824868f5d1e6...2eec076321fb


stuartmorgan-g:
We may not end up needing to land this; I turned on the autoroller to see what would happen, and https://github.com/flutter/packages/pull/9419 is looking promising.

PR #9436, opened 2025-06-17
[go_router] ShellRoute will merge GoRouter's observers

huanghui1998hhh:
`ShellRoute` currently does not trigger `observers` defined in `GoRouter`, which is unexpected behavior.

Fixes https://github.com/flutter/flutter/issues/112196

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/go_router/lib/src/route.dart
```diff
@@ -482,8 +483,20 @@ abstract class ShellRouteBase extends RouteBase {
...
+  /// Whether to merge the observers of the shell route's parent with the
+  /// observers of the shell route.
+  ///
+  /// When `true`, the observers of the shell route's parent will be merged with
+  /// the observers of the shell route.
+  ///
+  /// Only effective for `observers` passed into `GoRouter`.
+  ///
+  /// Defaults to `false`.
+  final bool mergeObservers;
```
chunhtai: why make this a per route setting instead of a global setting in goRouter?

chunhtai: also what if it is a shellroute nested inside another shell route, if this is true, should the innershellroute notify the observer of outer shellroute?

packages/go_router/lib/src/route.dart
```diff
@@ -552,9 +565,27 @@ class ShellRouteContext {
...
+      bool mergeObservers,
+      String? restorationScopeId) {
+    final List<NavigatorObserver> effectiveObservers = <NavigatorObserver>[
+      ...?observers
+    ];
+
+    if (mergeObservers) {
+      final List<NavigatorObserver>? rootObservers =
+          // ignore: invalid_use_of_visible_for_testing_member
+          GoRouter.maybeOf(context)?.routerDelegate.builder.observers;
```
chunhtai: should just store the root observer in GoRouter object so that we don't access the observer this way

huanghui1998hhh: fix

packages/go_router/lib/src/route.dart
huanghui1998hhh: Maybe we should use `Navigator.maybeOf(context).widget.observers` to merge observers and keep the `mergeObservers` property for each `ShellRoute`.

This ensures that the case of `ShellRoute` nesting `ShellRoute` is handled correctly, but there is one exception: when `ShellRoute` is nested under a `Navigator` that is not built by `ShellRoute`. Maybe we should ignore this case.

The reason for adding `mergeObservers` property to each `ShellRoute` is to allow users to decide that a specific `ShellRoute` should not merge observers. Defining it as a global setting in `GoRouter` would make maintaining this case difficult.

justinmc:
@huanghui1998hhh FYI you have some conflicts on this PR.

justinmc:
@huanghui1998hhh Some merge conflicts and a failure still on here FYI.

alexobviously:
Just wondering what the status of this is? We have some issues caused by the bug this PR fixes. Happy to help if there's anything I could do to help this get merged.

Piinks:
Since this is a draft we have not been actively reviewing it, @huanghui1998hhh when you are ready for another review, you can mark the PR 'ready for review'. :)

huanghui1998hhh:
Need more advice on the implementation details. Especially in this [case](https://github.com/flutter/packages/pull/9436#discussion_r2156361021).

This feature is highly requested. We need to properly track page navigation within ShellRoute.

packages/go_router/lib/src/route.dart
chunhtai: After thinking a bit, this should probably called something like notifyRootObserver or something since merging observer is implementation detail.

also consider default this to true and do a breaking change major bump

packages/go_router/lib/src/misc/merged_observer.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:flutter/widgets.dart';
```
chunhtai: let's not create a new file but instead make this a private class in route.dart so that we don't expose implementation detail

chunhtai:
Can you add a breaking change doc for this?

1. make a copy of flutter.dev/go/go-router-v16-breaking-changes and update the content
2. share the doc to everyone with view permission, and add a golink entry in here https://github.com/flutter/website/blob/e86734e470b7b9bd588c7ada19647d4c9f4cdca6/firebase.json#L471
3. add the link to readme https://github.com/flutter/packages/blob/main/packages/go_router/README.md#migration-guides

packages/go_router/README.md
```diff
@@ -38,6 +38,7 @@ See the API documentation for details on the following topics:
 - [State restoration](https://pub.dev/documentation/go_router/latest/topics/State%20restoration-topic.html)
 
 ## Migration Guides
+- [Migrating to 17.0.0](https://flutter.dev/go/go-router-v17-breaking-changes).
```
chunhtai, line 41: for the doc content, instead of using word `merge`, we should state its effect instead.

e.g.

instead of
`ShellRoute should merge GoRouter's observers by default.`

we should state this instead

`ShellRoute's navigating changes notify GoRouter's observers by default.`

packages/go_router/lib/src/route.dart
```diff
@@ -498,8 +498,18 @@ abstract class ShellRouteBase extends RouteBase {
     super.redirect,
     required super.routes,
     required super.parentNavigatorKey,
+    this.notifyRootObserver = true,
   }) : super._();
 
+  /// Whether to merge the observers of the GoRouter with the observers
+  /// of the ShellRoute.
+  ///
+  /// When `true`, the observers of the GoRouter will be merged with
```
chunhtai:
```suggestion
  /// When `true`, navigation changes within this shell route will notify the GoRouter's observers.
```

packages/go_router/lib/src/route.dart
```diff
@@ -498,8 +498,18 @@ abstract class ShellRouteBase extends RouteBase {
     super.redirect,
     required super.routes,
     required super.parentNavigatorKey,
+    this.notifyRootObserver = true,
   }) : super._();
 
+  /// Whether to merge the observers of the GoRouter with the observers
```
chunhtai:
```suggestion
  /// Whether navigation changes will notify the GoRouter's observers.
```

packages/go_router/lib/src/route.dart
```diff
@@ -1231,6 +1265,9 @@ class StatefulNavigationShell extends StatefulWidget {
   /// Corresponds to the index in the branches field of [StatefulShellRoute].
   final int currentIndex;
 
+  /// Same to [ShellRoute.notifyRootObserver].
```
chunhtai:
```suggestion
  /// Whether navigation changes will notify the GoRouter's observers.
  ///
  /// Same as [ShellRoute.notifyRootObserver].
```

chunhtai: also, should this be defined in ShellRouteBase?

huanghui1998hhh: Removed. Now gets the value from StatefulNavigationShellState.route.notifyRootObserver

chunhtai (approved):
LGTM

PR #9459, opened 2025-06-20
[local_auth] Convert iOS/macOS to Swift

stuartmorgan-g:
Rewrites the iOS/macOS implementation in Swift. This is an in-place rewrite with minimal changes, to minimize the chances of breakage, and to simplify review. Future refactorings can improve the Swift structure (e.g., fully adopting thread enforcement).

Fixes https://github.com/flutter/flutter/issues/119104

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/SystemWrappers.swift
stuartmorgan-g, line 1: This can mostly be compared side-by-side to FLALocalAuthPlugin_Test.h

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
stuartmorgan-g, line 1: This should be renamed in a follow-up; I didn't rename it here because there are enough changes that git would probably not show it as a diff, making review harder.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import LocalAuthentication
```
stuartmorgan-g, line 5: This should always have been here, but transitive includes in Obj-C masked the missing include.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -16,24 +17,24 @@ import XCTest
 private let timeout: TimeInterval = 30.0
 
 /// A context factory that returns preset contexts.
-final class StubAuthContextFactory: NSObject, FLADAuthContextFactory {
+final class StubAuthContextFactory: AuthContextFactory {
```
stuartmorgan-g, line 20: I removed NSObject from all the test implementations since the protocols aren't Obj-C now, so it was just cruft.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -16,24 +17,24 @@ import XCTest
...
 
-  var contexts: [FLADAuthContext]
-  init(contexts: [FLADAuthContext]) {
+  var contexts: [AuthContext]
+  init(contexts: [AuthContext]) {
```
stuartmorgan-g, line 23: All the Pigeon-generated type names changed because there's no more prefixing.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -52,52 +53,75 @@ final class StubViewProvider: NSObject, FLAViewProvider {
       return NSButton()  // The return value is not used by the plugin.
     }
 
-    func beginSheetModal(for sheetWindow: NSWindow) async -> NSApplication.ModalResponse {
+    func beginSheetModal(
+      for sheetWindow: NSWindow,
+      completionHandler handler: ((NSApplication.ModalResponse) -> Void)? = nil
```
stuartmorgan-g, line 58: The signature change here is because I directly copied the Swift declaration of this method from Apple's docs to make the Swift version of the AuthAlert protocol, whereas before it was using the auto-transform from Obj-C. It seemed like this was necessary in order to make NSAlert implement AuthAlert, but maybe it's just my lack of familiarity with the subtleties of Swift/Obj-C interop.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -52,52 +53,75 @@ final class StubViewProvider: NSObject, FLAViewProvider {
...
+    ) {
       presentingWindow = sheetWindow
+      handler?(NSApplication.ModalResponse.OK)
+    }
+
+    func runModal() -> NSApplication.ModalResponse {
```
stuartmorgan-g, line 64: See implementation comments for the reason for this change.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -52,52 +53,75 @@ final class StubViewProvider: NSObject, FLAViewProvider {
...
     func present(
-      on presentingViewController: UIViewController, animated flag: Bool,
+      on presentingViewController: UIViewController, animated: Bool,
       completion: (() -> Void)? = nil
     ) {
       presented = true
       self.presentingViewController = presentingViewController
+      // The plugin does not use the passed action, so just send a dummy value. If that ever
+      // changes, the test will need to track the action along with the handler.
+      onPresentActionHandler?(UIAlertAction())
```
stuartmorgan-g, line 88: See below for the explanation of this.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -52,52 +53,75 @@ final class StubViewProvider: NSObject, FLAViewProvider {
...
       return self.alertController
     }
+
+    func createAlertAction(
+      title: String?, style: UIAlertAction.Style, handler: ((UIAlertAction) -> Void)?
+    ) -> UIAlertAction {
+      // Configure the fake controller to trigger this button when presented. This is currently an
+      // arbitrary button, just to ensure that the completion handler is triggered so that the
+      // test can wait for the full cycle of async calls to complete.
+      alertController.onPresentActionHandler = handler
```
stuartmorgan-g, line 118: ... and this.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -161,26 +187,31 @@ class FLALocalAuthPluginTests: XCTestCase {
...
+    ) { resultDetails in
       XCTAssertTrue(Thread.isMainThread)
-      XCTAssertEqual(resultDetails?.result, FLADAuthResult.success)
-      XCTAssertNil(error)
+      switch resultDetails {
+      case .success(let successDetails):
+        XCTAssertEqual(successDetails.result, AuthResult.success)
+      case .failure(let error):
+        XCTFail("Unexpected error: \(error)")
+      }
```
stuartmorgan-g, line 196: This change, which is repeated throughout the file, is because the Obj-C Pigeon-generated protocol uses a completion that takes a nullable real result value and a nullable `NSError`, one of which is null, whereas the Swift Pigeon generation uses the simpler and safer `Result` type, so the structure is a little different.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -561,21 +627,25 @@ class FLALocalAuthPluginTests: XCTestCase {
...
+    }
   }
 
-  // Creates an FLADAuthStrings with placeholder values.
-  func createAuthStrings() -> FLADAuthStrings {
-    return FLADAuthStrings.make(
-      withReason: "a reason", lockOut: "locked out", goToSettingsButton: "Go To Settings",
-      goToSettingsDescription: "Settings", cancelButton: "Cancel", localizedFallbackTitle: nil)
+  // Creates an AuthStrings with placeholder values.
+  func createAuthStrings(localizedFallbackTitle: String? = nil) -> AuthStrings {
```
stuartmorgan-g, line 592: This minor change was just so that the return value could be `let`, instead of `var` followed by setting this value.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -304,81 +355,81 @@ class FLALocalAuthPluginTests: XCTestCase {
...
     plugin.authenticate(
-      with: FLADAuthOptions.make(
-        withBiometricOnly: false,
+      options: AuthOptions(
+        biometricOnly: false,
         sticky: false,
         useErrorDialogs: true),
       strings: strings
-    ) { resultDetails, error in
-      // TODO(stuartmorgan): Add a wrapper around UIAction to allow accessing the handler, so
```
stuartmorgan-g, line 347: This is the reason for the changes noted above; I had to implement this TODO to make the test work. Because of the threading/actor checks in Swift, to fix warnings I needed to add an async-dispatch-to-main-queue. Previously this test worked despite the lack of expectation await on iOS because in practice, the callback was being called synchronously (even though the API has an async structure). Once that was no longer true, I needed a way to have this test wait on iOS, so I wired up a way to get the handler and trigger it.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/SystemWrappers.swift
```diff
@@ -0,0 +1,106 @@
...
+  ) -> Bool
+  func evaluatePolicy(
+    _ policy: LAPolicy,
+    localizedReason: String,
+    reply: @escaping @Sendable (Bool, (any Error)?) -> Void
+  )
+}
+
+/// AuthContext is intentionally a direct passthroguh to LAContext.
+extension LAContext: AuthContext {}
```
stuartmorgan-g, line 39: This and the extension below are changes relative to the Obj-C implementation, addressing the TODOs in the Obj-C implementation to remove the manual passthrough implementations in favor of this version.

I could back this out to reduce changes if reviewers prefer, but since this is almost purely code removal that eliminates a possible source of errors (bugs in the non-test passthrough, which by definition isn't tested by unit tests) I went ahead and included it.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/FLALocalAuthPlugin.m
```diff
@@ -1,493 +0,0 @@
...
-#import "./include/local_auth_darwin/FLALocalAuthPlugin.h"
-#import "./include/local_auth_darwin/FLALocalAuthPlugin_Test.h"
-
-#import <LocalAuthentication/LocalAuthentication.h>
-
-typedef void (^FLADAuthCompletion)(FLADAuthResultDetails *_Nullable, FlutterError *_Nullable);
-
-/// A default auth context that wraps LAContext.
-// TODO(stuartmorgan): When converting to Swift, eliminate this class and use an extension to make
-// LAContext declare conformance to FLADAuthContext.
```
stuartmorgan-g, line 13: This class was replaced by an extension, per the TODO.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/FLALocalAuthPlugin.m
```diff
@@ -1,493 +0,0 @@
...
-  return [[FLADefaultAuthContext alloc] initWithContext:[[LAContext alloc] init]];
-}
-@end
-
-#pragma mark -
-
-#if TARGET_OS_OSX
-/// A default alert that wraps NSAlert.
-// TODO(stuartmorgan): When converting to Swift, eliminate this class and use an extension to make
-// NSAlert declare conformance to FLANSAlert.
```
stuartmorgan-g, line 74: Same here.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/FLALocalAuthPlugin.m
```diff
@@ -1,493 +0,0 @@
...
-
-- (void)beginSheetModalForWindow:(NSWindow *)sheetWindow
-               completionHandler:(void (^_Nullable)(NSModalResponse returnCode))handler {
-  [self.alert beginSheetModalForWindow:sheetWindow completionHandler:handler];
-}
-@end
-#elif TARGET_OS_IOS
-/// A default alert controller that wraps UIAlertController.
-// TODO(stuartmorgan): When converting to Swift, eliminate this class and use an extension to make
-// UIAlertController declare conformance to FLAUIAlertController.
```
stuartmorgan-g, line 113: This TODO was incorrect; this was still needed due to `presentOnViewController:animated:completion:` which isn't quite a passthrough (and can't be, as explained in the protocol).

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g, line 1: This can be compared side-by-side to FLALocalAuthPlugin.m

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+    ) -> AuthAlertController {
+      return DefaultAlertController(
+        wrapping:
+          UIAlertController(
+            title: title,
+            message: message,
+            preferredStyle: preferredStyle))
+    }
+
+    func createAlertAction(
```
stuartmorgan-g, line 69: This is new relative to the Obj-C code. This is how the tests save off the `handler` (for which UIAlertAction has no accessor) so that we can trigger it later to simulate the user picking a dialog option in unit tests (see discussion in my comments on the unit test file changes).

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+  let options: AuthOptions
+  let strings: AuthStrings
+  let resultHandler: (Result<AuthResultDetails, Error>) -> Void
+}
+
+// MARK: -
+
+/// A flutter plugin for local authentication.
+// TODO(stuartmorgan): Remove the @unchecked Sendable, and convert to strict thread enforcement.
+// This was added to minimize the changes while converting from Swift to Objective-C.
```
stuartmorgan-g, line 109: This was unfortunate, but it was the best I could come up with to fix all the warnings without significant restructuring. Without this, the places where we dispatch from a potentially-background-thread-callback back to a main thread call to a method on `self` triggers warnings about `self` not being Sendable.

There are ways to fix that with more static methods, passing more state around instead of using self, etc., but they would require refactoring that I didn't want to combine with the language conversion as it would have made side-by-side review nearly impossible.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+            strings: strings,
+            completion: completion)
+        }
+      }
+    } else {
+      if let authError = authError {
+        self.handleError(authError, options: options, strings: strings, completion: completion)
+      } else {
+        // This should not happen according to docs, but if it ever does the plugin should still
+        // fire the completion.
```
stuartmorgan-g, line 182: This branch is new. In Obj-C, we just assumed the docs weren't lying about the error being set if it returned false, but that's not type safe in Swift, and since docs have lied about nullability before I wanted a fallback instead of a crash from force-unwrapping.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+            }
+            alert.addAction(additionalAction)
+          }
+        }
+        alert.present(
+          on: controller,
+          animated: true,
+          completion: nil)
+      } else {
+        // TODO(stuartmorgan): Create a new error code for failure to show UI, and return it here.
```
stuartmorgan-g, line 268: This branch is new. We were failing to handle the case where there is no view before, so I added minimal handling for now.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+      alert.messageText = message
+      alert.addButton(withTitle: dismissButtonTitle)
+      if let window = viewProvider.view?.window {
+        alert.beginSheetModal(
+          for: window
+        ) { code in
+          self.handleResult(succeeded: false, completion: completion)
+        }
+      } else {
+        alert.runModal()
```
stuartmorgan-g, line 261: This branch is new. We were failing to handle the case where there is no view before.

We could fail like on iOS (see below), but since macOS has a simple way to handle showing an alert without a view I went with that option.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+          } else {
+            handleResult(succeeded: false, completion: completion)
+          }
+        default:
+          handleError(error, options: options, strings: strings, completion: completion)
+        }
+      } else {
+        // The Obj-C declaration of evaluatePolicy defines the callback type as NSError*, but the
+        // Swift version is (any Error)?, so provide a fallback in case somehow the type is not
+        // NSError.
```
stuartmorgan-g: This branch is new, for the reason described here.

stuartmorgan-g:
Hm, it looks like `(any Error)` is newer Swift syntax than our version of ~Xcode~ `swift-format`? I'll look into that and fix it.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -106,24 +130,25 @@ final class StubAuthContext: NSObject, FLADAuthContext {
...
-  /// The error to return from evaluatePolicy:error:.
+  /// The error to return from evaluatePolicy.
   var evaluateError: NSError?
 
   // Overridden as read-write to allow stubbing.
   var biometryType: LABiometryType = .none
   var localizedFallbackTitle: String?
 
-  func canEvaluatePolicy(_ policy: LAPolicy) throws {
+  func canEvaluatePolicy(_ policy: LAPolicy, error: NSErrorPointer) -> Bool {
```
hellohuanlin, line 142: curious if this also conforms to the protocol: 

```
func canEvaluatePolicy(_ policy: LAPolicy, error: inout NSError?) -> Bool
```

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -161,26 +187,31 @@ class FLALocalAuthPluginTests: XCTestCase {
...
         useErrorDialogs: false),
       strings: strings
-    ) { resultDetails, error in
+    ) { resultDetails in
       XCTAssertTrue(Thread.isMainThread)
-      XCTAssertEqual(resultDetails?.result, FLADAuthResult.success)
-      XCTAssertNil(error)
+      switch resultDetails {
+      case .success(let successDetails):
+        XCTAssertEqual(successDetails.result, AuthResult.success)
```
hellohuanlin: nit: `XCTAssertEqual(successDetails.result, .success)`

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -478,81 +535,90 @@ class FLALocalAuthPluginTests: XCTestCase {
...
       stubAuthContext.biometryType = .faceID
-
     }
 
-    var error: FlutterError?
-    let result = plugin.getEnrolledBiometricsWithError(&error)
-    XCTAssertEqual(result!.count, 1)
-    XCTAssertEqual(result![0].value, FLADAuthBiometric.face)
-    XCTAssertNil(error)
+    do {
```
hellohuanlin: here and other tests - we can mark the function signature as `throws`, and remove this `do/catch` block

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
```diff
@@ -561,21 +627,25 @@ class FLALocalAuthPluginTests: XCTestCase {
...
-  func createAuthStrings() -> FLADAuthStrings {
-    return FLADAuthStrings.make(
-      withReason: "a reason", lockOut: "locked out", goToSettingsButton: "Go To Settings",
-      goToSettingsDescription: "Settings", cancelButton: "Cancel", localizedFallbackTitle: nil)
+  // Creates an AuthStrings with placeholder values.
+  func createAuthStrings(localizedFallbackTitle: String? = nil) -> AuthStrings {
+    return AuthStrings(
+      reason: "a reason", lockOut: "locked out", goToSettingsButton: "Go To Settings",
+      goToSettingsDescription: "Settings", cancelButton: "Cancel",
+      localizedFallbackTitle: localizedFallbackTitle)
```
hellohuanlin, line 599: nit: separate lines for each items

(https://google.github.io/swift/ -> Line-Wrapping -> "Comma-delimited lists are only laid out in one direction: horizontally or vertically. ")

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+#if os(macOS)
+  import Cocoa
+  import FlutterMacOS
+#elseif os(iOS)
+  import Flutter
+  import UIKit
+#endif
+
+/// A default context factory that wraps standard LAContext allocation.
+struct DefaultAuthContextFactory: AuthContextFactory {
```
hellohuanlin: nit: use `class` unless we need value semantics of `struct` (since `struct` comes with overhead)

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+  func createAuthContext() -> AuthContext {
+    return LAContext()
+  }
+}
+
+// MARK: -
+
+#if os(iOS)
+  /// A default alert controller that wraps UIAlertController.
+  struct DefaultAlertController: AuthAlertController {
```
hellohuanlin: nit: can be `class`

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+    /// The wrapped alert controller.
+    var controller: UIAlertController
```
hellohuanlin: can be `private` and `let`?

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+
+    /// Returns a wrapper for the given UIAlertController.
+    init(wrapping controller: UIAlertController) {
```
hellohuanlin, line 31: maybe easier for the caller to just pass in the parameters (e.g. `init(title: ... message: ...)`)

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
hellohuanlin: and other places

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+    var view: NSView? {
+      return registrar.view
+    }
+  #endif  // os(macOS)
+}
+
+// MARK: -
+
+/// A data container for sticky auth state.
+struct StickyAuthState {
```
hellohuanlin, line 99: what is "sticky"?

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
hellohuanlin: I think let's keep `@unchecked Sendable` for now. I plan to look closer into Swift concurrency later this year

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+  let strings: AuthStrings
+  let resultHandler: (Result<AuthResultDetails, Error>) -> Void
+}
+
+// MARK: -
+
+/// A flutter plugin for local authentication.
+// TODO(stuartmorgan): Remove the @unchecked Sendable, and convert to strict thread enforcement.
+// This was added to minimize the changes while converting from Swift to Objective-C.
+public class LocalAuthPlugin: NSObject, FlutterPlugin, LocalAuthApi, @unchecked Sendable {
```
hellohuanlin: nit: `final`

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+
+  /// Manages the last call state for sticky auth.
+  var lastCallState: StickyAuthState?
+  /// The factory to create LAContexts.
+  var authContextFactory: AuthContextFactory
+  /// The factory to create alerts.
+  var alertFactory: AuthAlertFactory
+  /// The Flutter view provider.
+  var viewProvider: ViewProvider
```
hellohuanlin: nit: can these be `private`?

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+    let policy =
+      options.biometricOnly
+      ? LAPolicy.deviceOwnerAuthenticationWithBiometrics
+      : LAPolicy.deviceOwnerAuthentication
+    var authError: NSError?
+    if context.canEvaluatePolicy(policy, error: &authError) {
+      context.evaluatePolicy(
+        policy,
+        localizedReason: strings.reason
+      ) { (success: Bool, error: Error?) in
```
hellohuanlin: `[weak self]`

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+        self.handleError(authError, options: options, strings: strings, completion: completion)
+      } else {
+        // This should not happen according to docs, but if it ever does the plugin should still
+        // fire the completion.
+        completion(
+          .success(
+            AuthResultDetails(
+              result: AuthResult.failure,
+              errorMessage: "evaluatePolicy failed without an error"
+            )))
```
hellohuanlin, line 188: this indentation is tooooooo much

```
completion(.success(AuthResultDetails(
  result: .failure,
  errorMessage: "..."
)))
```

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+    let context = authContextFactory.createAuthContext()
+    var biometrics: [AuthBiometric] = []
+    if context.canEvaluatePolicy(
+      LAPolicy.deviceOwnerAuthenticationWithBiometrics,
+      error: nil)
+    {
+      if #available(macOS 10.15, iOS 11.0, *) {
+        if context.biometryType == LABiometryType.faceID {
+          biometrics.append(AuthBiometric.face)
+          return biometrics
```
hellohuanlin, line 228: is it intended to return here directly?

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+      if context.biometryType == LABiometryType.touchID {
+        biometrics.append(AuthBiometric.fingerprint)
+      }
+    }
+    return biometrics
+  }
+
+  func isDeviceSupported() throws -> Bool {
+    let context = authContextFactory.createAuthContext()
+    return context.canEvaluatePolicy(LAPolicy.deviceOwnerAuthentication, error: nil)
```
hellohuanlin: nit: can infer `canEvaluatePolicy(.deviceOwnerAuthentication, ...)`

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+    completion: @escaping (Result<AuthResultDetails, Error>) -> Void
+  ) {
+    #if os(macOS)
+      var alert = alertFactory.createAlert()
+      alert.messageText = message
+      alert.addButton(withTitle: dismissButtonTitle)
+      if let window = viewProvider.view?.window {
+        alert.beginSheetModal(
+          for: window
+        ) { code in
```
hellohuanlin: nit: `alert.beginSheetModal(for: window) { code in `

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+      } else {
+        alert.runModal()
+        self.handleResult(succeeded: false, completion: completion)
+      }
+    #elseif os(iOS)
+      // TODO(stuartmorgan): Get the view controller from the view provider once it's possible.
+      // See https://github.com/flutter/flutter/issues/104117.
+      if let controller = UIApplication.shared.delegate?.window??.rootViewController {
+        let alert =
+          alertFactory.createAlertController(
```
hellohuanlin: nit: no need new line

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+            title: "",
+            message: message,
+            preferredStyle: UIAlertController.Style.alert)
+
+        let defaultAction = alertFactory.createAlertAction(
+          title: dismissButtonTitle,
+          style: UIAlertAction.Style.default
```
hellohuanlin: nit: can infer `style: .default`

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+        ) { action in
+          self.handleResult(
+            succeeded: false,
+            completion: completion)
+        }
+
+        alert.addAction(defaultAction)
+        if let openSettingsButtonTitle = openSettingsButtonTitle {
+          let url = URL(string: UIApplication.openSettingsURLString)
+          if let url = url {
```
hellohuanlin: nit: can reduce 1 level of indentation by: 
```
if let openSettingsButtonTitle, 
  let url = URL(String: UIApplication.openSettingsURLString)
{
```

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+                completionHandler: nil)
+              self.handleResult(succeeded: false, completion: completion)
+            }
+            alert.addAction(additionalAction)
+          }
+        }
+        alert.present(
+          on: controller,
+          animated: true,
+          completion: nil)
```
hellohuanlin: nit: does it fit one line?

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+  private func handleAuthReply(
+    success: Bool,
+    error: Error?,
+    options: AuthOptions,
+    strings: AuthStrings,
+    completion: @escaping (Result<AuthResultDetails, Error>) -> Void
+  ) {
+    if success {
+      handleResult(succeeded: true, completion: completion)
+    } else {
```
hellohuanlin: nit: `return` in success to reduce indentation? 
```
if success {
  handle...
  return
}
...
```

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+      if let error = error as? NSError {
+        switch LAError.Code(rawValue: error.code) {
+        case LAError.biometryNotAvailable,
+          LAError.biometryNotEnrolled,
+          LAError.biometryLockout,
+          LAError.userFallback,
+          LAError.passcodeNotSet,
+          LAError.authenticationFailed:
```
hellohuanlin: nit: can infer these (e.g. `case .biometryNotAvailable`)

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,435 @@
...
+    }
+  }
+
+  private func handleResult(
+    succeeded: Bool, completion: @escaping (Result<AuthResultDetails, Error>) -> Void
+  ) {
+    completion(
+      .success(
+        AuthResultDetails(
+          result: succeeded ? AuthResult.success : AuthResult.failure,
```
hellohuanlin: nit: infer type

hellohuanlin: and below

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/SystemWrappers.swift
```diff
@@ -0,0 +1,106 @@
...
+    @MainActor
+    var messageText: String { get set }
+    @MainActor
+    @discardableResult func addButton(withTitle title: String) -> NSButton
+    @MainActor
+    func beginSheetModal(
+      for sheetWindow: NSWindow,
+      completionHandler handler: ((NSApplication.ModalResponse) -> Void)?
+    )
+    @MainActor
```
hellohuanlin, line 72: we can mark the whole protocol `@MainActor` so no need each function. (same below)

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/SystemWrappers.swift
```diff
@@ -0,0 +1,106 @@
...
+  /// AuthAlert is intentionally a direct passthroguh to NSAlert.
+  extension NSAlert: AuthAlert {}
+#endif  // macOS
+
+#if os(iOS)
+  /// Protocol for interacting with UIAlertController instances, abstracted to allow using mock/fake
+  /// instances in unit tests.
+  protocol AuthAlertController {
+    @MainActor
+    func addAction(_ action: UIAlertAction)
```
hellohuanlin, line 86: nit: add a blank line between each members (same below)

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: What's the overhead? I thought it was the opposite; https://developer.apple.com/documentation/swift/choosing-between-structures-and-classes#Overview says to prefer structs by default.

hellohuanlin: > What's the overhead? 

Value types increase both build time and binary size. Though Apple has improved it over the years, so it shouldn't be our key consideration. 

The key consideration is whether we want value or reference semantics (which should be the "control identity" in the link you provided)

For example, factories in general should be reference types, because they often share state, so we don't want multiple factories. This particular factory is stateless, so not the end of the world, but still, semantically we don't want multiple factories when passing it around. 

The next example is more interesting: 
```
struct DefaultAlertController {
    var controller: UIAlertController
}
```
If we use value type like here, imagine this code: 
```
let c1 = DefaultAlertController(...)
let c2 = c1
```
Then we have 2 separate values (`c1` and `c2`), but both pointing to the same underlying `UIAlertController`, which is conceptually misleading (it looks like you have created 2 distinct controllers, but you haven't).

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
stuartmorgan-g: The goal is to match `LAContext`, and what I have here is the signature according to Apple:

https://developer.apple.com/documentation/localauthentication/lacontext/canevaluatepolicy(_:error:)

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
stuartmorgan-g: Done throughout.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
stuartmorgan-g: Done throughout.

packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift
stuartmorgan-g: Done. I forgot `swift-format` annoyingly doesn't handle this.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Makes sense! Changed all of them.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Nothing in this class can be unit tested; the explicit goal of this class is to have the absolutely least possible logic. The only reason this isn't just a protocol extension like the other cases is because the last method reverses the semantics.

Abstracting the fact that this is an extremely thin wrapper around `UIViewController` by doing more in this class is an anti-goal.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: https://pub.dev/packages/local_auth#sticky-auth

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done. Also changed those that are init params to `let`.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: TIL that Swift closures have capture lists. Added to all the blocks that use `self`.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: You'll have to file a `swift-format` issue for that; every other version I tried is turned back to this.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: It doesn't actually matter (since the condition below can't possibly be true), but it's how the Obj-C code was written so I kept it.

I can remove it if you want, but it won't change the behavior.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done for several enums (`LAPolicy`, `LAError`, `AuthResult`) throughout the file.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done for all instances.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Yep!

I did a pass over the PR to manually re-wrap some things now that I've been reminded that `swift-format` won't do most line break handling.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/SystemWrappers.swift
stuartmorgan-g: This is problematic until we annotate both Pigeon-generated methods and the `FlutterPlugin` registration methods as `@MainActor`, because it cascades through other code that ends up in places like the creation of the default implementations in `register(with:)`, which aren't annotated `@MainActor` even though in practice they are.

It's much easier for now to annotate the specific methods.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/SystemWrappers.swift
stuartmorgan-g: Done

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,421 @@
...
+#if os(macOS)
+  import Cocoa
+  import FlutterMacOS
+#elseif os(iOS)
+  import Flutter
+  import UIKit
+#endif
+
+/// A default context factory that wraps standard LAContext allocation.
+class DefaultAuthContextFactory: AuthContextFactory {
```
hellohuanlin: nit: these classes can all be final

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
hellohuanlin: thanks. not straightforward but interesting name

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,421 @@
...
+      options.biometricOnly
+      ? LAPolicy.deviceOwnerAuthenticationWithBiometrics
+      : LAPolicy.deviceOwnerAuthentication
+    var authError: NSError?
+    if context.canEvaluatePolicy(policy, error: &authError) {
+      context.evaluatePolicy(
+        policy,
+        localizedReason: strings.reason
+      ) { (success: Bool, error: Error?) in
+        DispatchQueue.main.async { [weak self] in
```
hellohuanlin: I think you have to put [weak self] in outer closure `context.evaluatePolicy`, otherwise it would remain strong when outer closure is still alive.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,421 @@
...
+          self?.handleResult(succeeded: false, completion: completion)
+        }
+      } else {
+        alert.runModal()
+        self.handleResult(succeeded: false, completion: completion)
+      }
+    #elseif os(iOS)
+      // TODO(stuartmorgan): Get the view controller from the view provider once it's possible.
+      // See https://github.com/flutter/flutter/issues/104117.
+      if let controller = UIApplication.shared.delegate?.window??.rootViewController {
```
hellohuanlin: nit: this is a good use case for `guard let`

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
```diff
@@ -0,0 +1,421 @@
...
+      ))
+  }
+
+  private func handleError(
+    _ authError: NSError,
+    options: AuthOptions,
+    strings: AuthStrings,
+    completion: @escaping (Result<AuthResultDetails, Error>) -> Void
+  ) {
+    var result = AuthResult.errorNotAvailable
```
hellohuanlin: nit: can be `let result: AuthResult` (prefer immutability). This will fail to compile if any of the case below forget to set the result.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/SystemWrappers.swift
```diff
@@ -0,0 +1,111 @@
...
+  import FlutterMacOS
+#elseif os(iOS)
+  import Flutter
+  import UIKit
+#endif
+
+/// Protocol for interacting with LAContext instances, abstracted to allow using mock/fake instances
+/// in unit tests.
+protocol AuthContext {
+  var localizedFallbackTitle: String? { get set }
```
hellohuanlin, line 19: nit: blank line between each members. Also wanna add doc comments?

hellohuanlin (approved):

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Makes sense; done.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Converted to `guard`.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/LocalAuthPlugin.swift
stuartmorgan-g: Done. I didn't know Swift could do this to; I'm a big fan of the pattern in Dart.

packages/local_auth/local_auth_darwin/darwin/local_auth_darwin/Sources/local_auth_darwin/SystemWrappers.swift
stuartmorgan-g: Done.

PR #9460, opened 2025-06-20
[tool] Use `swift-format` from Xcode

stuartmorgan-g:
As of Xcode 16, `swift-format` is part of the Xcode distribution. Since CI now uses Xcode 16, we can just always use that version.

This:
- Removes the `swift-format-path` logic.
- Removes the check for `swift-format` being available
- Defaults Swift formatting to true only on macOS
- Removes the `swift-format` CIPD package from .ci.yaml
- Updates Pigeon in packages using Swift Pigeon generation to pick up a fix for a lint issue picked up by the latest version of `swift-format` (which was already fixed in Pigeon).

Fixes https://github.com/flutter/flutter/issues/153803

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/script/tool/lib/src/format_command.dart
+++ b/script/tool/lib/src/format_command.dart
@@ -58,6 +58,11 @@ class FormatCommand extends PackageLoopingCommand {
     argParser.addFlag(_kotlinArg,
         help: 'Format Kotlin files', defaultsTo: true);
     argParser.addFlag(_javaArg, help: 'Format Java files', defaultsTo: true);
+    // Currently swift-format is run via xcrun, so only works on macOS. If that
+    // ever becomes an issue, the ability to find it in the path and/or allow
+    // providing a path could be restored, to allow developers on Linux or
+    // Windows to build swift-format from source and use that. See
+    // https://github.com/flutter/packages/pull/9460.
     argParser.addFlag(_swiftArg,
         help: 'Format and lint Swift files', defaultsTo: platform.isMacOS);
     argParser.addOption(_clangFormatPathArg,


stuartmorgan-g:
@jmagman / @vashworth Any objection to this approach of just unconditionally using `xcrun`, rather than using the multi-step check originally described in https://github.com/flutter/flutter/issues/153803 ?

We could still do that version, if we wanted to support overriding the `swift-format` version (or to allow running `swift-format` from other platforms), but in practice I think that now that it's part of Xcode the chances that anyone will ever actually want to run a version that's not the Xcode version is very low, so probably not worth the complexity in the tool.

packages/file_selector/file_selector_macos/pigeons/messages.dart
```diff
@@ -45,7 +45,7 @@ class SavePanelOptions {
 /// Options for open panels.
 ///
 /// These correspond to NSOpenPanel properties.
-class OpenPanelOptions extends SavePanelOptions {
```
stuartmorgan-g, line 48: This change preserves the old behavior. When I first wrote this I used inheritance out of habit, but that didn't actually do anything in Pigeon at the time. I fixed it by adding composition (line 62 below), but forgot to remove this `extends` that was being ignored.

In the current Pigeon this is an error, because inheritance does work, but only to inherit from a sealed empty base class to allow type switching.

script/tool/lib/src/format_command.dart
```diff
@@ -59,13 +59,11 @@ class FormatCommand extends PackageLoopingCommand {
         help: 'Format Kotlin files', defaultsTo: true);
     argParser.addFlag(_javaArg, help: 'Format Java files', defaultsTo: true);
     argParser.addFlag(_swiftArg,
-        help: 'Format and lint Swift files', defaultsTo: true);
+        help: 'Format and lint Swift files', defaultsTo: platform.isMacOS);
```
cbracken, line 67: fyi: Nothing to do for this patch, but it looks like swift-format is open source https://github.com/swiftlang/swift-format and tagged for various releases if we one did want to support this on Linux as well.

cbracken (approved):

vashworth:
> @jmagman / @vashworth Any objection to this approach of just unconditionally using `xcrun`, rather than using the multi-step check originally described in [flutter/flutter#153803](https://github.com/flutter/flutter/issues/153803) ?
> 
> We could still do that version, if we wanted to support overriding the `swift-format` version (or to allow running `swift-format` from other platforms), but in practice I think that now that it's part of Xcode the chances that anyone will ever actually want to run a version that's not the Xcode version is very low, so probably not worth the complexity in the tool.

No objection. We should document how to add this back somewhere in case we do want to use it again.

stuartmorgan-g:
> We should document how to add this back somewhere in case we do want to use it again.

I'll add a comment with some context and a link to this PR on the constructor line that makes it default to on for macOS only.

stuartmorgan-g:
CHANGELOG/version override: `webview_flutter_wkwebview` change is whitespace-only.

jmagman (approved):
Thank you!

PR #9473, opened 2025-06-23
[in_app_purchase] Annotate deprecation in test

stuartmorgan-g:
Annotate an intentional use of a deprecated method in an integration test, so that it doesn't show up in deprecation audits.

See https://github.com/flutter/flutter/blob/main/docs/infra/Packages-Gardener-Rotation.md#deprecations


gmackall (approved):
lgtm

PR #9500, opened 2025-06-26
Revert "[camera_android_camerax] Force new `Surface` for each `SurfaceRequest`"

stuartmorgan-g:
Reverts flutter/packages#9360

It uses an API that's not available on `stable`, breaking compilation of the plugin on `stable`.


stuartmorgan-g:
@ash2moon please make sure to watch for tree breakage when you are the Packages Gardener.

PR #9508, opened 2025-06-26
[quick_actions] Restore the appShortcutLaunchActivityAfterStarting test in quick_actions_android

jason-simmons:
This was temporarily disabled due to https://github.com/flutter/flutter/issues/170141, which should be fixed now.


stuartmorgan-g:
Looks like the roller hasn't quite caught up to that commit yet (we've had some tree closure so it's behind).

stuartmorgan-g (approved):
LGTM once the roller catches up so it passes presubmit.

PR #9511, opened 2025-06-27
[video_player] Move Android to per-player-instance Pigeon APIs

stuartmorgan-g:
Rather than having a single API to talk directly to the plugin, and having most of the methods in that API just do a map lookup and dispatch to a player instance, which was necessary when Pigeon didn't support instantiating multiple instances of an API, have each player instance set up an API instance that the Dart code can talk to directly. This follows the pattern used by plugins that migrated to Pigeon more recently (e.g., `google_maps_flutter` has a very similar pattern), reduces boilerplate native code, and moves closer to what an FFI-based implementation would look like if we go that route in the future.

Since the Dart unit tests needed to be significantly reworked anyway, this also moves to the pattern we are using in newer plugin code, where we use `mockito` to mock the Pigeon API surface. The "call log" approach it replaces dates back to pre-Pigeon, when only way to test that the right platform calls were made was to intercept and track method channel calls at the framework level.

Also updates to the latest version of Pigeon.

## Pre-Review Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter].
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ ] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or I have commented below to indicate which [version change exemption] this PR falls under[^1].
- [x] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or I have commented below to indicate which [CHANGELOG exemption] this PR falls under[^1].
- [x] I updated/added any relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or I have commented below to indicate which [test exemption] this PR falls under[^1].
- [x] All existing and new tests are passing.

[^1]: Regular contributors who have demonstrated familiarity with the repository guidelines only need to comment if the PR is not auto-exempted by repo tooling.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[...]

PR #9529, opened 2025-06-30
[video_player] Move iOS/macOS to per-player-instance Pigeon APIs

stuartmorgan-g:
Rather than having a single API to talk directly to the plugin, and having most of the methods in that API just do a map lookup and dispatch to a player instance, which was necessary when Pigeon didn't support instantiating multiple instances of an API, have each player instance set up an API instance that the Dart code can talk to directly. This follows the pattern used by plugins that migrated to Pigeon more recently (e.g., `google_maps_flutter` has a very similar pattern), reduces boilerplate native code, and moves closer to what an FFI-based implementation would look like if we go that route in the future.

Since the Dart unit tests needed to be significantly reworked anyway, this also moves to the pattern we are using in newer plugin code, where we use `mockito` to mock the Pigeon API surface. The "call log" approach it replaces dates back to pre-Pigeon, when only way to test that the right platform calls were made was to intercept and track method channel calls at the framework level.

Also updates to the latest version of Pigeon.

## Pre-Review Checklist

- [x] I signed the [CLA].


stuartmorgan-g:
(c.f. https://github.com/flutter/packages/pull/9511, which is the same structural change for the Android implementation.)

packages/video_player/video_player_avfoundation/darwin/RunnerTests/VideoPlayerTests.m
```diff
@@ -310,29 +312,22 @@ - (void)testSeekToWhilePausedStartsDisplayLinkTemporarily {
...
   // Seeking to a new position should start the display link temporarily.
   OCMVerify([mockDisplayLink setRunning:YES]);
-  FVPTextureBasedVideoPlayer *player =
-      (FVPTextureBasedVideoPlayer *)videoPlayerPlugin.playersByIdentifier[playerIdentifier];
-  // Wait for the player's position to update, it shouldn't take long.
-  XCTestExpectation *positionExpectation =
-      [self expectationForPredicate:[NSPredicate predicateWithFormat:@"position == 1234"]
-                evaluatedWithObject:player
-                            handler:nil];
-  [self waitForExpectations:@[ positionExpectation ] timeout:3.0];
```
stuartmorgan-g, line 319: This expectation was failing for me locally, and I believe it's been flaking occasionally. IIRC this check dates to before we waited for the seek to fully finish (the expectation fulfilled above, whose name I fixed). This test doesn't actually test the logic of seeking itself, only the frame handling around seeking, so expectation above should be all we need to order the test validations correctly.

Same with the change below.

packages/video_player/video_player_avfoundation/darwin/RunnerTests/VideoPlayerTests.m
```diff
@@ -590,59 +583,36 @@ - (void)testBufferingStateFromPlayer {
 }
 
 - (void)testVideoControls {
-  NSObject<FlutterPluginRegistrar> *registrar = OCMProtocolMock(@protocol(FlutterPluginRegistrar));
```
stuartmorgan-g, line 562: The rest of the changes in this file are just cleanup for cases that used to test via the plugin, but were really just testing player functionality so could be easily switched to creating a player instance directly instead of creating and driving the player via the plugin (which after my change was creating the player via the plugin, looking it up, and then driving it via the player anyway). See https://github.com/flutter/packages/pull/9529/commits/1963bb362c96ae7b98d7f1013df819a5b142e299 for that cleanup, isolated as a separate commit.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
```diff
@@ -28,32 +28,28 @@ @interface FVPTextureBasedVideoPlayer ()
...
-@property(nonatomic, copy) void (^onDisposed)(int64_t);
 @end
 
 @implementation FVPTextureBasedVideoPlayer
 - (instancetype)initWithAsset:(NSString *)asset
                  frameUpdater:(FVPFrameUpdater *)frameUpdater
                   displayLink:(FVPDisplayLink *)displayLink
                     avFactory:(id<FVPAVFactory>)avFactory
-                 viewProvider:(NSObject<FVPViewProvider> *)viewProvider
-                   onDisposed:(void (^)(int64_t))onDisposed {
```
stuartmorgan-g, line 40: This changed to being a post-init property assignment because we need the player ID, which is only available just after creation.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPTextureBasedVideoPlayer.m
```diff
@@ -153,12 +146,6 @@ - (void)disposeSansEventChannel {
   _displayLink = nil;
 }
 
-- (void)dispose {
-  [super dispose];
-
-  _onDisposed(self.frameUpdater.textureIdentifier);
-}
```
stuartmorgan-g, line 160: This moved to the superclass since all players have an onDisposed now.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -104,6 +104,25 @@ - (void)dealloc {
   }
 }
 
+/// This method allows you to dispose without touching the event channel. This
+/// is useful for the case where the Engine is in the process of deconstruction
+/// so the channel is going to die or is already dead.
+- (void)disposeSansEventChannel {
```
stuartmorgan-g, line 109: This is moved from below without change, to create a `#pragma mark -` separated section for each protocol implemented.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -104,6 +104,25 @@ - (void)dealloc {
...
+  _disposed = YES;
+  [self removeKeyValueObservers];
+
+  [self.player replaceCurrentItemWithPlayerItem:nil];
+  [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
+
+- (void)dispose {
```
stuartmorgan-g, line 117: This was also moved, for the same reason. The change is the addition of the `_onDisposed` call.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -405,57 +424,56 @@ - (void)setupEventSinkIfReadyToPlay {
...
+- (void)pauseWithError:(FlutterError *_Nullable *_Nonnull)error {
   _isPlaying = NO;
   [self updatePlayingState];
 }
 
-- (int64_t)position {
-  return FVPCMTimeToMillis([_player currentTime]);
-}
-
-- (int64_t)duration {
```
stuartmorgan-g, line 422: This is moved below as it's not part of of the protocol.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -264,54 +271,6 @@ - (void)disposePlayer:(NSInteger)playerIdentifier error:(FlutterError **)error {
...
-        dispatch_async(dispatch_get_main_queue(), ^{
-          completion(nil);
-        });
-      }];
-}
-
-- (void)pausePlayer:(NSInteger)playerIdentifier error:(FlutterError **)error {
-  FVPVideoPlayer *player = self.playersByIdentifier[@(playerIdentifier)];
-  [player pause];
-}
```
stuartmorgan-g, line 320: This is the core of the change: we no longer need boilerplate like this for every player method.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -108,18 +108,32 @@ - (int64_t)onPlayerSetup:(FVPVideoPlayer *)player {
...
   }
 
+  NSObject<FlutterBinaryMessenger> *messenger = self.messenger;
+  NSString *channelSuffix = [NSString stringWithFormat:@"%lld", playerIdentifier];
+  // Set up the player-specific API handler, and its onDispose unregistration.
+  SetUpFVPVideoPlayerInstanceApiWithSuffix(messenger, player, channelSuffix);
+  __weak typeof(self) weakSelf = self;
+  BOOL isTextureBased = textureBasedPlayer != nil;
+  player.onDisposed = ^() {
+    SetUpFVPVideoPlayerInstanceApiWithSuffix(messenger, nil, channelSuffix);
```
stuartmorgan-g, line 131: The addition of `onDisposed` to the base class is to allow easily cleaning up the communication channel, without having to push all the communication channel wire-up into the player class (which would be a bigger change, and is also undesirable for potential conversion to using FFI to talk directly to the player instance).

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -92,45 +112,58 @@ class AVFoundationVideoPlayer extends VideoPlayerPlatform {
...
+    ensureApiInitialized(playerId);
 
     return playerId;
   }
 
+  /// Returns the API instance for [playerId], creating it if it doesn't already
+  /// exist.
+  @visibleForTesting
+  VideoPlayerInstanceApi ensureApiInitialized(int playerId) {
+    VideoPlayerInstanceApi? api = _playerApis[playerId];
```
LongCatIsLooong: nit: `putIfAbsent`?

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -219,6 +252,14 @@ class AVFoundationVideoPlayer extends VideoPlayerPlatform {
     return EventChannel('flutter.io/videoPlayer/videoEvents$playerId');
   }
 
+  VideoPlayerInstanceApi _apiFor(int playerId) {
```
LongCatIsLooong: Question: as a layman reader I find the "api" in these names slightly confusing. Is this a naming convention? I'd go with something like `VideoPlayerInstance _playerWith(playerId)` since these are just players?

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -219,6 +252,14 @@ class AVFoundationVideoPlayer extends VideoPlayerPlatform {
...
+    final VideoPlayerInstanceApi? api = _playerApis[playerId];
+    if (api == null) {
```
LongCatIsLooong: uber nit: 
```dart 
return _playerApis[playerId] ?? (throw StateError('No active player with ID $playerId.'));
```

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
LongCatIsLooong: I'm curious why this would have FFI implications? In my mind the communication channel wire-up / tear-down part are implementation details within the platform, the dart side should be able to tell the plugin to create a player, or dispose an existing player without involving any communication channel concepts?

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
stuartmorgan-g: Done; I forgot that was a thing!

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
stuartmorgan-g: It is a convention, but it probably doesn't make sense to apply like this. I've changed it to player everywhere.

I believe it started because in pre-Pigeon code there would usually be a channel instance called something like "_channel", and so in Pigeon conversions where the channel was replaced by a more abstract API, it became "_api". But we should probably rethink that name in general.

> since these are just players?

Well, they aren't just players is the thing. They are instances of Pigeon APIs that are intended to talk to players. They are not the actual player objects in some pretty important ways (e.g., there is no ownership dynamic across the boundary). But `player` is still probably better shorthand for these instances.

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
stuartmorgan-g: Done.

... I'm not sure how I feel about this construction though. On one hand, it reads nicely. On the other hand, I didn't even know this would compile. I guess it works because the second part's type is `Never`, and that's a bottom type?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
stuartmorgan-g: If the structure of the native code is that all the channel-related details are contained entirely within the main plugin class (this file), and the player classes are just receiving calls, then switching to FFI can be done with almost no changes to the player instance classes.

In contrast, if I had pushed instance channel setup in the player instance classes in this PR, then a future change to FFI would require more changes in those classes to unwind that, instead of mostly changing this plugin-structure-specific class.

PR #9534, opened 2025-06-30
Make sure the `AVPlayerItem` is `.readyToPlay` before emitting an `initialized` event

LongCatIsLooong:
Fixes https://github.com/flutter/flutter/issues/166914. The duration check was first introduced in [here](https://github.com/flutter/plugins/pull/1554) but zero division should probably be handled by the client not the plugin.

Also I suspect that https://github.com/flutter/flutter/issues/91975 wasn't caused by an AVFoundation bug but rather a bug in `video_player_avfoundation`:

The `_isInitialized` flag was roughly equivalent to `.readyToPlay` [a long time ago](https://github.com/flutter/plugins/pull/280/files#diff-f4c7ef029bb5e0be8a6851b6b022c745dc4e99e3739f1c2f7f4944b85bc154a6R174-R178). The flag was later re-purposed to indicate whether an `initialized` message has already been sent (to make sure the message is only sent once), and [that PR](https://github.com/flutter/plugins/pull/1552) also introduced a size check to determine whether the media is loaded (which doesn't seem to be correct, `.readyToPlay` is the public API for this, see Apple's [AVPlayerDemo](https://developer.apple.com/library/archive/samplecode/AVPlayerDemo/Listings/Classes_AVPlayerDemoPlaybackViewController_m.html#//apple_ref/doc/uid/DTS40010101-Classes_AVPlayerDemoPlaybackViewController_m-DontLinkElementID_8)). But that PR didn't update the `onListenWithArguments` implementation so the plugin started sending `initialized` events on first subscription even the AVPlayerItem isn't ready to play.

## Pre-Review Checklist

- [ ] I signed the [CLA].


LongCatIsLooong:
I'm still trying to generate a [sample video with 0 duration](https://github.com/flutter/flutter/issues/166914#issuecomment-3046115365) (which seems to be an AVFoundation bug as the video plays in chrome, but not in quicktime) so I can add a test. But much of this change should already be covered by existing tests added for the workarounds in the past.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -351,58 +325,24 @@ - (void)sendFailedToLoadVideoEvent {
...
+- (void)sendVideoInitializedEvent {
+  AVPlayerItem *currentItem = self.player.currentItem;
+  NSAssert(currentItem.status == AVPlayerItemStatusReadyToPlay,
+           @"sendVideoInitializedEvent was called when the item wasn't ready to play.");
+  NSAssert(_eventSink, @"sendVideoInitializedEvent was called when the event sink was nil.");
+  NSAssert(!_isInitialized, @"sendVideoInitializedEvent should only be called once.");
+  CGSize size = currentItem.presentationSize;
+  CGFloat width = size.width;
+  CGFloat height = size.height;
+
```
LongCatIsLooong, line 370: Removed the duration / size checks since now the caller has to guarantee the item is `.readyToPlay` and I think those checks shouldn't be necessary.

Before this you may enter this `if` block even when the item's status is `.unknown` and I believe that's why we added the workarounds (size / duration checks).

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -474,27 +414,29 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
 #pragma mark - FlutterStreamHandler
 
 - (FlutterError *_Nullable)onCancelWithArguments:(id _Nullable)arguments {
+  NSAssert([NSThread isMainThread], @"event sink must only be accessed from the main thread");
   _eventSink = nil;
+  _isInitialized = NO;
```
LongCatIsLooong: `_isInitialized` is now reset to `NO` on `onListenWithArguments`. According to the documentation:
>The channel implementation may call this method with nil arguments to separate a pair of two consecutive set up requests. Such request pairs may occur during Flutter hot restart.

So this makes sure the new event sink gets an `initialized` event in case it's a hot restart (in which case I think the dart vm will also restart erasing all user states?). But if the engine is allowed to call `onCancelWithArguments` and `onListenWithArguments` whenever it wants then the dart side may receive more than one `initialized` event.

packages/video_player/video_player_avfoundation/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.8.1
+
+* Removes unnecessary workarounds, fixes "initialized" event not firing when the duration of the media is 0.
```
hellohuanlin: > Removes unnecessary workarounds

What workarounds?

LongCatIsLooong: The size / duration checks removed in the `sendInitialized` method.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -245,56 +236,41 @@ - (void)observeValueForKeyPath:(NSString *)path
...
   } else if (context == statusContext) {
     AVPlayerItem *item = (AVPlayerItem *)object;
     switch (item.status) {
-      case AVPlayerItemStatusFailed:
-        [self sendFailedToLoadVideoEvent];
-        break;
       case AVPlayerItemStatusUnknown:
         break;
       case AVPlayerItemStatusReadyToPlay:
         [item addOutput:_videoOutput];
```
misos1: If status is changed to `AVPlayerItemStatusReadyToPlay` before the event sink is set up then it seems this `addOutput` will not be called.

LongCatIsLooong: Oops thanks for catching this!

packages/video_player/video_player_avfoundation/CHANGELOG.md
hellohuanlin: It's unclear from reading the CHANGELOG.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -383,52 +371,15 @@ - (void)sendFailedToLoadVideoEvent {
...
-    if (duration == 0) {
-      return;
-    }
-
-    _isInitialized = YES;
-    [self updatePlayingState];
-
-    [self.eventListener videoPlayerDidInitializeWithDuration:duration size:size];
-  }
+  _isInitialized = YES;
```
LongCatIsLooong, line 371: looks like the flag is still used by the `updatePlayingState` method.

LongCatIsLooong:
Renamed the PR since with recent changes I think we're only calling `reportInitialized` after making sure the status of the item is `.readyToPlay`.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
misos1: If `_isInitialized == true` always implies that player status is ready to play then maybe `updateRate` can be simplified. Maybe `updatePlayingState` could instead check that player status is ready to play (`canPlayFastForward` and `canPlaySlowForward` do not work when status is not ready to play). Btw was not `_isInitialized` set to false in `onCancelWithArguments` before?

stuartmorgan-g:
> The duration check was first introduced [here](https://github.com/flutter/plugins/pull/1554) but zero division should probably be handled by the client not the plugin.

The plugin is a client; the mini-controller code you changed here is a duplicate of [the actual video_player's progress indicator code](https://github.com/flutter/packages/blob/08a9b2cc46ddb8777a719f745c132b216b2b6121/packages/video_player/video_player/lib/video_player.dart#L1120-L1148). That will need to be fixed first, and we'll need to decide if we're okay regressing the div-zero issue in the (unlikely) case of someone updating video_player_avfoundation without having updated video_player.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -383,52 +373,15 @@ - (void)sendFailedToLoadVideoEvent {
 }
 
 - (void)reportInitializedIfReadyToPlay {
-  if (!_isInitialized) {
```
stuartmorgan-g, line 373: Does the `status` API guarantee that it can only transition to `readyToPlay` once? It seems *likely* that it would, but I didn't see any docs in a quick look.

stuartmorgan-g:
From triage: What's the status of this PR?

LongCatIsLooong:
(addressing review comments now). I'll update the `video_player` implementation to handle div by 0.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -863,42 +863,37 @@ class VideoPlayer extends StatefulWidget {
...
     _playerId = widget.controller.playerId;
-    widget.controller.addListener(_listener);
+    widget.controller.addListener(_controllerDidUpdateValue);
   }
 
   @override
-  void deactivate() {
-    super.deactivate();
-    widget.controller.removeListener(_listener);
+  void dispose() {
```
LongCatIsLooong, line 895: drive-by fix: `deactivate` is called either when the state is going to be reparented or disposed.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -1089,58 +1084,52 @@ class VideoProgressIndicator extends StatefulWidget {
...
-
-  late VoidCallback listener;
-
   VideoPlayerController get controller => widget.controller;
 
   VideoProgressColors get colors => widget.colors;
 
+  void _didUpdateControllerValue() {
+    setState(() {
+      // The build method reads from controller.value.
```
LongCatIsLooong, line 1094: The `mounted` check wasn't necessary. The state will stop listening when the element is unmounted.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
LongCatIsLooong: Yeah the documentation doesn't seem to make that guarantee (nor the sample code). I tried to update the code to allow `FVPVideoPlayer` to send "initialized" events more than once but it seems the `video_player` implementation is only expecting that event exactly once:

```dart
          assert(
            !initializingCompleter.isCompleted,
            'VideoPlayerController already initialized. This is typically a '
            'sign that an implementation of the VideoPlayerPlatform '
            '(${_videoPlayerPlatform.runtimeType}) has a bug and is sending '
            'more than one initialized event per instance.',
          );
```

So for now I added an `if` to prevent `reportInitializedIfReadyToPlay` being called more than once. But that means if the state goes from `.readyToPlay` -> `.failed` -> `.readyToPlay` (again this may not be possible but the documentation doesn't say), `video_player` wouldn't know that the platform player has recovered.

LongCatIsLooong:
Ready for review:
1. Added an `if` to prevent "initialized" from being sent for more than once, in case `AVPlayer` can enter `.readyToPlay` state more than once. (However as a result the dart implementation won't be able to recover from a failed state or update size/duration in case `.readyToPlay` can be sent a second time).
2. Fixes widget implementation in the dart code.

packages/video_player/video_player_android/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.8.14
```
stuartmorgan-g, line 16: You should revert the CHANGELOG and version change in `video_player_android`; the changes here are only to an unpublished file.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -370,52 +362,15 @@ - (void)sendFailedToLoadVideoEvent {
 }
 
 - (void)reportInitializedIfReadyToPlay {
```
stuartmorgan-g: This should just be `reportInitialized` now, as there is no more "if" aspect to its behavior.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -370,52 +362,15 @@ - (void)sendFailedToLoadVideoEvent {
...
-      return;
-    }
-
-    _isInitialized = YES;
-    [self updatePlayingState];
-
-    [self.eventListener videoPlayerDidInitializeWithDuration:duration size:size];
-  }
+  _isInitialized = YES;
+  [self updatePlayingState];
```
stuartmorgan-g: Since this isn't related to reporting initialization, and there's only one place calling this method now, I think it would be better to move this to line 292 (just after the `if` block at the call site).

stuartmorgan-g:
From triage: @LongCatIsLooong is this still on your radar to update and land?

PR #9572, opened 2025-07-07
[tool] Remove Android x86 compilation

stuartmorgan-g:
Remvoe x86 from the ABI set when compiling Android integration tests, as x86 is no longer supported for Flutter builds.

See https://github.com/flutter/flutter/pull/170191

### DIFF
--- a/script/tool/lib/src/native_test_command.dart
+++ b/script/tool/lib/src/native_test_command.dart
@@ -400,7 +400,6 @@ this command.
             'android-arm',
             'android-arm64',
             'android-x64',
-            'android-x86'
           ];
           final int exitCode = await project.runCommand(
             'app:connectedAndroidTest',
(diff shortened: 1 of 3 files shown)


stuartmorgan-g:
~I believe this will resolve one source of roll breakage that's accumulated since the last successful roll; I haven't tested that, but this is definitely code we shouldn't need any more.~

I've verified locally that this fixes the

```
* What went wrong:
An exception occurred applying plugin request [id: 'dev.flutter.flutter-gradle-plugin']
> Failed to apply plugin 'dev.flutter.flutter-gradle-plugin'.
   > Invalid platform: android-x86
```

failures that are hitting the roller since https://github.com/flutter/flutter/pull/170191 landed in flutter/flutter.

tarrinneal (approved):

PR #9594, opened 2025-07-11
Replace popularity badges

stuartmorgan-g:
Updates the repo-level README to replace the popularity badges, which no longer work, with monthly download count badges. Also removes the links, since there's no specific page for them to link to; they were linking to the score page, which was probably a copy/paste bug that we never noticed.


bparrishMines (approved):
LGTM

PR #9626, opened 2025-07-14
[go_router_builder] Change mixin name

StevenSorial:
Fixes [#170650](https://github.com/flutter/flutter/issues/170650)

This PR fixes workflows that relied on putting child routes in files different than parent files, which resulted in the (private) mixin being generated in a different file than the route itself.

To avoid releasing a new major version in such a short period, this PR makes the mixin public only if needed. Admittedly makes the behavior somewhat unexpected for the user. @chunhtai @hannah-hyj  I will leave that decision for you, whether that's ok or I should make it always public and release a major version instead.

Since this change just fixes a use-case that already didn't compile in 3.x.x, this PR doesn't bump the major version.

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/go_router_builder/lib/src/route_config.dart
+++ b/packages/go_router_builder/lib/src/route_config.dart
@@ -881,7 +881,7 @@ RouteBase get $_routeGetterName => ${_invokesRouteConstructor()};
 
   String get _className => routeDataClass.displayName;
 
-  String get _mixinName => '_\$$_className';
+  String get _mixinName => '\$$_className';
 
   String get _extensionName => '\$${_className}Extension';
 
--- a/packages/go_router_builder/CHANGELOG.md
+++ b/packages/go_router_builder/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.0.0
+
+- Make Route mixins public.
+
 ## 3.3.1
 
 - Fixes warnings in generated code of iterable parameters.
--- a/packages/go_router_builder/README.md
+++ b/packages/go_router_builder/README.md
@@ -41,6 +41,9 @@ dart run build_runner build
 Read more about using
 [`build_runner` on pub.dev](https://pub.dev/packages/build_runner).
 
+## Migration Guides
+- [Migrating to 4.0.0](https://flutter.dev/go/go-router-builder-v4-breaking-changes).
+
 ## Overview
 
 `go_router` fundamentally relies on the ability to match a string-based location
--- a/packages/go_router_builder/README.md
+++ b/packages/go_router_builder/README.md
@@ -86,7 +86,7 @@ method.
 
 <?code-excerpt "example/lib/readme_excerpts.dart (HomeRoute)"?>
 ```dart
-class HomeRoute extends GoRouteData with _$HomeRoute {
+class HomeRoute extends GoRouteData with $HomeRoute {
   const HomeRoute();
 
   @override
@@ -106,7 +106,7 @@ The tree of routes is defined as an attribute on each of the top-level routes:
     TypedGoRoute<FamilyRoute>(path: 'family/:fid'),
   ],
 )
-class HomeRoute extends GoRouteData with _$HomeRoute {
+class HomeRoute extends GoRouteData with $HomeRoute {
   const HomeRoute();
 
   @override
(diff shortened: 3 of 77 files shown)


ManuelRauber:
Works for me for the moment. :)

StevenSorial:
> change makes sense to me. we should also update the readme

@chunhtai, so are you ok with the approach where the mixin will be public /private based on the routes configuration? or does it seem too inconsistent for you?

If you are ok with the current approach, I will add a note in the readme

packages/go_router_builder/lib/src/route_config.dart
```diff
@@ -729,7 +743,11 @@ RouteBase get $_routeGetterName => ${_invokesRouteConstructor()};
 
   String get _className => routeDataClass.name;
 
-  String get _mixinName => '_\$$_className';
+  String get _mixinName {
+    // If the routeDataClass is in a different file, we need to make the mixin public
+    final Uri routeUri = routeDataClass.source.uri;
+    return routeUri != targetUri ? '\$$_className' : '_\$$_className';
```
chunhtai: thinking about it more, let's just do a breaking change and make this a public class. having dealing with cases in both seems counter intuitive since we are maintaining a versioned package.

chunhtai (commented):
I think let's just make this a breaking change. I see no reason to not make this a public mixin. WDYT?

StevenSorial:
> I think let's just make this a breaking change. I see no reason to not make this a public mixin. WDYT?

@chunhtai  My ideal scenario would be to improve the Generator to be able track the routes in different files without an annotation in every route file

```
// parent_route.dart

@TypedGoRoute<ParentRoute>(
  path: '/',
  routes: [TypedGoRoute<SubRoute>(path: 'sub')],
)
class ParentRoute extends GoRouteData with _$ParentRoute {
}

// sub_route.dart

class SubRoute extends GoRouteData with _$SubRoute {}
```

in my ideal scenario, the generator would be able to generate `parent_route.g.dart` containing `_$ParentRoute` mixin, and `sub_route.g.dart` containing `_$SubRoute` mixin.

I looked into this, and it seems that it will require so much work and possibly getting rid of `SharedPartBuilder`. Is that assumption correct?

mustafamasri1:
any update?

StevenSorial:
@chunhtai could you please answer my last [question](https://github.com/flutter/packages/pull/9626#issuecomment-3166056882)?

mustafamasri1:
> > I think let's just make this a breaking change. I see no reason to not make this a public mixin. WDYT?
> 
> @chunhtai My ideal scenario would be to improve the Generator to be able track the routes in different files without an annotation in every route file
> 
> ```
> // parent_route.dart
> 
> @TypedGoRoute<ParentRoute>(
>   path: '/',
>   routes: [TypedGoRoute<SubRoute>(path: 'sub')],
> )
> class ParentRoute extends GoRouteData with _$ParentRoute {
> }
> 
> // sub_route.dart
> 
> class SubRoute extends GoRouteData with _$SubRoute {}
> ```
> 
> in my ideal scenario, the generator would be able to generate `parent_route.g.dart` containing `_$ParentRoute` mixin, and `sub_route.g.dart` containing `_$SubRoute` mixin.
> 
> I looked into this, and it seems that it will require so much work and possibly getting rid of `SharedPartBuilder`. Is that assumption correct?

This idea is already created, check this [comment](https://github.com/flutter/flutter/issues/122258#issuecomment-3044054364) in [#122258](https://github.com/flutter/flutter/issues/122258)

chunhtai:
My concern is that developer need to be aware of 2 option, `_$Route` and `$Route` and remember when the `_$Route` is allow and when it is not. We should just give them one choice, and it seems $Route for all cases?

> in my ideal scenario, the generator would be able to generate parent_route.g.dart containing _$ParentRoute mixin, and sub_route.g.dart containing _$SubRoute mixin.

That will work as well

> I looked into this, and it seems that it will require so much work and possibly getting rid of SharedPartBuilder. Is that assumption correct?

How so?

mustafamasri1:
any update?

StevenSorial:
> LGTM, can you write a breaking change doc?

Sure. How can I do that?

chunhtai:
1. You can make a doc similar to this one
https://docs.google.com/document/d/1E4B5nMDEvsqQiCo-Ki9LohocLFC88X398-AiLXRQhwU/edit?tab=t.0

2. share with everyone for view only
3. Add an golink entry in https://github.com/flutter/website/blob/main/firebase.json
4. add the breaking change doc to Readme of go_router_builder similar to what we did in go_router

chunhtai (approved):
LGTM

StevenSorial:
@chunhtai The submit label got removed

chunhtai:
waiting for another review

hannah-hyj (approved):
LGTM

PR #9628, opened 2025-07-15
[camera_avfoundation] Fix crash when restarting image stream after recording

t09tanaka:
## Summary

Added a guard in `captureOutput:didOutputSampleBuffer:` to safely handle nil pixel buffers that occur when transitioning from recording to image streaming. This prevents app crashes caused by force unwrapping nil values.

## What

- Added `guard let` check for `CMSampleBufferGetImageBuffer` result in DefaultCamera.swift
- Decrements pending frame count when skipping nil buffers to maintain accurate state
- Added comprehensive test coverage for nil buffer scenarios

## Why

When stopping video recording and immediately restarting the image stream, the app crashes with:
```
Thread 55: Fatal error: Unexpectedly found nil while unwrapping an Optional value
```

This occurs because:
1. After recording stops, the recording format buffers continue to flow for several frames
2. These buffers don't contain actual image data, causing `CMSampleBufferGetImageBuffer` to return NULL
3. The force unwrap (`\!`) on the nil result causes an immediate crash

## Technical Details

The issue happens during the transition period between recording and streaming:
- Recording uses AVAssetWriter which has a specific buffer format
- When transitioning back to streaming, there's a brief period where invalid buffers are received
- These buffers have no image data, resulting in nil from `CMSampleBufferGetImageBuffer`

## Note

Ideally, Flutter should be able to query buffer state before processing. However, this PR focuses solely on preventing the crash by safely handling nil buffers. Future improvements could include exposing buffer state information to the Flutter side for more sophisticated handling.

## Test

Added `StreamingNilBufferTests.swift` with comprehensive tests for:
- Handling nil image buffers during streaming
- Mixed sequences of valid and nil buffers
- Verifying frame count accuracy when buffers are skipped

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
+++ b/packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
@@ -276,7 +276,11 @@ final class DefaultCamera: FLTCam, Camera {
       {
         streamingPendingFramesCount += 1
 
-        let pixelBuffer = CMSampleBufferGetImageBuffer(sampleBuffer)!
+        guard let pixelBuffer = CMSampleBufferGetImageBuffer(sampleBuffer) else {
+          // Skip this frame if pixelBuffer is nil (can happen right after recording stops)
+          streamingPendingFramesCount -= 1
+          return
+        }
         // Must lock base address before accessing the pixel data
         CVPixelBufferLockBaseAddress(pixelBuffer, .readOnly)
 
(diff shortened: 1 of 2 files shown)


stuartmorgan-g:
> 🤖 Generated with [Claude Code](https://claude.ai/code)

While it's fine to use AI tools as part of developing a PR, we still require all PRs to follow our standard PR process. Please file a new PR whose description follows our template (including completing the checklist) rather than a completely AI-generated description that ignores our process, and we'll be happy to take a look. Thanks!

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -276,7 +276,11 @@ final class DefaultCamera: FLTCam, Camera {
       {
         streamingPendingFramesCount += 1
 
-        let pixelBuffer = CMSampleBufferGetImageBuffer(sampleBuffer)!
+        guard let pixelBuffer = CMSampleBufferGetImageBuffer(sampleBuffer) else {
+          // Skip this frame if pixelBuffer is nil (can happen right after recording stops)
+          streamingPendingFramesCount -= 1
+          return
```
RobertOdrowaz, line 282: This early return isn't quite correct. While we should ignore non-image (usually audio) buffers for purposes of streaming, they should be recorded. I would expect that this change would cause videos recording with streaming to have no audio

PR #9632, opened 2025-07-16
[ci] Add Gemini Code Assist review config

stuartmorgan-g:
This adds initial configuration files for [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code) in preparation for turing it on for this repository on a trial basis. These config files are based on [the official
docs](https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github)

It also adds a note to the PR template to provide some notice/guidance to contributors about how to use automated feedback, to clarify the role of the automated reviews and hopefully minimize potential issues in case the bot leaves incorrect feedback.

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -1,0 +1,23 @@
+# Flutter Packages Style Guide
+
+## Introduction
+
+This style guide outlines the coding conventions for contributions to the
+flutter/packages repository.
+
+## Style Guides
+
+Code should follow the relevant style guides, and use the correct
+auto-formatter, for each language, as described in
+[the repository contributing guide's Style section](https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style).
+
+## Best Practices
+
+- Code should follow the guidance and principles described in
+  [the flutter/packages contribution guide](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md).
+- Code should be tested. Changes to plugin packages, which include code written
+  in C, C++, Java, Kotlin, Objective-C, or Swift, should have appropriate tests
+  as described in [the plugin test guidance](https://github.com/flutter/flutter/blob/master/docs/ecosystem/testing/Plugin-Tests.md).
+- PR descriptions should include the Pre-Review Checklist from
+  [the PR template](https://github.com/flutter/packages/blob/main/.github/PULL_REQUEST_TEMPLATE.md),
+  with all of the steps completed.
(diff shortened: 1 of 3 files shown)


.gemini/styleguide.md
stuartmorgan-g, line 1: I have no idea how to know how well this will work. I tried to start with something reasonably concise, but that linked the most important documents, and I figure we can iterate from there if we are seeing patterns in the reviews that could potentially be addressed with additional guidance here.

jtmcdole:
Non-blocking: Do we also want a GEMINI.md in the root folder as well? Or plan for multiple ones in the sub folders?

.gemini/styleguide.md
jtmcdole: +1 to iterate.  Can we have it call out:

```
- Code should be formatted with `dart format`
- Code should pass `dart analyze`
```

stuartmorgan-g: > - Code should be formatted with `dart format`

The file links to https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style which already says that, along with listing all of the other formatters we require. I'm really hoping that reference is enough, because I don't want to maintain two copies of that list forever.

> - Code should pass `dart analyze`

We already have CI that validates that. I don't really want to try to teach it all the CI requirements since:
- That would be really unpleasant to maintain, since it would be a second copy our whole CI config.
- It would presumably either a) not work because it wasn't actually running those commands, or b) be super wasteful because it would be running all the commands that CI is running just to give a second copy of the same feedback.
- My primary goal here is to provide value to Flutter team members doing reviews, and they already know how to go look at the CI failures. (If we want to make that experience better via AI, I would recommend we build cross-repo tooling to teach some agent—maybe this one, maybe not—to understand and summarize all the failures in a set of LUCI tests on a PR, rather than try to teach the review agent what the commands our CI runs are).

stuartmorgan-g:
> Non-blocking: Do we also want a GEMINI.md in the root folder as well? Or plan for multiple ones in the sub folders?

The docs for the code review agent never mention GEMINI.md, so it is not clear to me that they apply to this system. We may want to add it later for other tools though.

stuartmorgan-g:
@jtmcdole We should be good to go for you to [install the app for this repo now](https://developers.google.com/gemini-code-assist/docs/review-github-code#install-gemini-code-assist).

jtmcdole:
App unsuspended and only allowed on flutter/packages

PR #9639, opened 2025-07-17
Fix Gemini note wrapping

stuartmorgan-g:
Apparently line wrapping in this file is preserved in the HTML display, which I wasn't expecting, and makes the note look terrible. This removes it so that it will wrap based on the UI width.


stuartmorgan-g:
Thanks Gemini 🤣

bparrishMines (approved):
LGTM

PR #9667, opened 2025-07-23
[ci] Re-enable wasm and Windows ARM64 tests

stuartmorgan-g:
The issues that caused these tests to be set to bringup have both been resolved, and the tasks have been consistently green recently.

See https://github.com/flutter/flutter/issues/170433
See https://github.com/flutter/flutter/issues/134083
Fixes https://github.com/flutter/flutter/issues/170434


bparrishMines (approved):
LGTM

PR #9674, opened 2025-07-25
[google_maps_flutter] [WIP] Add ability to perform Google Maps SDK warmup

filiph:
The Android Google Maps SDK janks when the first map is shown, completely blocking the app thread. This PR introduces `GoogleMapsFlutterAndroid.warmup()` which lets Flutter developers do what Android developers do to mitigate this: a fake show of a Google Map that forces the SDK to initialize. This is typically done at app startup so that the jank is less visible than during, say, a page transition.

More info: https://github.com/flutter/flutter/issues/28493#issuecomment-2919150669

This PR also introduces injecting `MapsInitializerApi` to the `GoogleMapsFlutterAndroid` constructor, for testing, and adds a basic test for `initializeWithRenderer`, too.

Fixes https://github.com/flutter/flutter/issues/28493

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapInitializer.java
```diff
@@ -41,6 +44,29 @@ public void initializeWithPreferredRenderer(
...
+  public void warmup(@NonNull Messages.VoidResult result) {
+      Log.i(TAG, "Google Maps warmup started.");
+      try {
+        // This creates a fake map view in order to trigger the SDK's
+        // initialization. For context, see
+        // https://github.com/flutter/flutter/issues/28493#issuecomment-2919150669.
+        MapView mv = new MapView(context);
+        mv.onCreate(null);
+        mv.onPause();
+        mv.onDestroy();
```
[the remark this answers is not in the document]
filiph: Thanks for the suggestion, robot.

I tried adding `onResume()` despite the fact it's not in the aforementioned workaround. Unfortunately, it had no effect. It seems most of the work happens in `onCreate()`. When I tried surrounding the `onResume()` call with log messages, they got called basically simultaneously, which means that `onResume()` is, for all intents and purposes, a no-op in this setup (remember, we're not actually showing any map here, just running some thread-blocking Google Maps initialization code).

filiph:
Hey maintainers, I've finalized this PR and removed the `WIP` tag. I see that [one of the test suites failed](https://github.com/flutter/packages/pull/9674/checks?check_run_id=46856116649) with a timeout, but I think maybe that's not my fault.

reidbaker:
> Hey maintainers, I've finalized this PR and removed the `WIP` tag. I see that [one of the test suites failed](https://github.com/flutter/packages/pull/9674/checks?check_run_id=46856116649) with a timeout, but I think maybe that's not my fault.

Yeah I dont think that should prevent review.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapInitializer.java
```diff
@@ -41,6 +44,24 @@ public void initializeWithPreferredRenderer(
...
 
+  @Override
+  public void warmup(@NonNull Messages.VoidResult result) {
+    Log.i(TAG, "Google Maps warmup started.");
+    try {
+      // This creates a fake map view in order to trigger the SDK's
+      // initialization. For context, see
+      // https://github.com/flutter/flutter/issues/28493#issuecomment-2919150669.
+      MapView mv = new MapView(context);
+      mv.onCreate(null);
```
reidbaker, line 55: Reading the implementation on this side I think you should include onResume.

packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: google_maps_flutter_android
 description: Android implementation of the google_maps_flutter plugin.
 repository: https://github.com/flutter/packages/tree/main/packages/google_maps_flutter/google_maps_flutter_android
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+maps%22
-version: 2.17.0
+version: 2.17.1
```
reidbaker: @stuartmorgan-g should this be a minor version bump not a patch version bump? 
I reviewed https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md but didnt see anything more specific than my general understadning of sem-ver.

packages/google_maps_flutter/google_maps_flutter_android/test/google_maps_flutter_android_test.dart
```diff
@@ -32,6 +33,39 @@ void main() {
...
+    final MockMapsInitializerApi initializerApi = MockMapsInitializerApi();
+    final GoogleMapsFlutterAndroid maps = GoogleMapsFlutterAndroid(
+        apiProvider: (_) => api, initializerApi: initializerApi);
+    await maps.initializeWithRenderer(AndroidMapRenderer.latest);
+
+    verify(initializerApi
+        .initializeWithPreferredRenderer(PlatformRendererType.latest));
+  });
+
+  test('warmup forwards the call', () async {
```
reidbaker:
```suggestion
  test('warmup forwards the initialization call', () async {
```

packages/google_maps_flutter/google_maps_flutter_android/test/google_maps_flutter_android_test.dart
```diff
@@ -32,6 +33,39 @@ void main() {
...
+    final GoogleMapsFlutterAndroid maps = GoogleMapsFlutterAndroid(
+        apiProvider: (_) => api, initializerApi: initializerApi);
+    const int mapId = 1;
+    maps.ensureApiInitialized(mapId);
+    await maps.init(1);
+
+    verifyZeroInteractions(initializerApi);
+  });
+
+  test('initializeWithPreferredRenderer forwards the call', () async {
```
reidbaker:
```suggestion
  test('initializeWithPreferredRenderer forwards the initialization call', () async {
```

packages/google_maps_flutter/google_maps_flutter_android/test/google_maps_flutter_android_test.mocks.dart
reidbaker, line 1: This looks like formatting and new content making it harder to review.

reidbaker (commented):
Thank you for your contribution!

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapInitializer.java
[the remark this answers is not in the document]
reidbaker: The api is intended to mirror the android lifecycle. Reading the implementation of the MapView code and looking at the prs that modified this api and the delegates that back it please call onResume after onCreate and before onPause. There is logic paired in onResume with onPause that can behave poorly if not paired. 

Additionally even if the existing implementation works future version of the api may not because of how engineers think about those lifecycle events. 

Googlers here is the implementation code that i found after digging through interfaces and internal apis. 
go/reidbaker-flutter-packages-pull-9674-maps-code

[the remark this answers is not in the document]
filiph: Ah, ok. Makes sense! Thanks for checking on the inside. I'll make the change.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapInitializer.java
filiph: Ack

packages/google_maps_flutter/google_maps_flutter_android/test/google_maps_flutter_android_test.mocks.dart
filiph: Unfortunately, this is a source-generated file and when I reran the source generation, it made all these formatting changes. Is there a way to avoid that?

reidbaker: The way to avoid it would be for us to regenerate the existing checked in code with the new formatter settings in a different pr. That is more work than it is worth. Feel free to ignore this comment.

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_android/README.md
```diff
@@ -61,6 +61,22 @@ the issue in the TLHC mode.
 | Heatmap.maximumZoomIntensity |     x     |
 | HeatmapGradient.colorMapSize |     ✓     |
 
+## Warmup
+
+The first time a map is shown, the Google Maps SDK will run thread-blocking
+initialization that can easily freeze the app for hundreds of milliseconds.
```
stuartmorgan-g: This language seems unnecessarily opinionated, and it also may not be accurate in the future as it's an implementation detail of the underlying SDK, which this plugin does not control.

packages/google_maps_flutter/google_maps_flutter_android/README.md
```diff
@@ -61,6 +61,22 @@ the issue in the TLHC mode.
...
+This is a limitation of the SDK on Android and has nothing to do with Flutter.
```
stuartmorgan-g: This sounds defensive; the previous sentence already says that the SDK is doing it.

packages/google_maps_flutter/google_maps_flutter_android/README.md
```diff
@@ -61,6 +61,22 @@ the issue in the TLHC mode.
...
+
+In order to avoid paying this initialization price at the worst possible time
+(e.g., during a page transition), Android developers have devised a workaround
+where a "fake" map is shown first, during a time at which a thread freeze is less
```
stuartmorgan-g: The history of the workaround isn't something that needs to be explained to clients.

packages/google_maps_flutter/google_maps_flutter_android/README.md
```diff
@@ -61,6 +61,22 @@ the issue in the TLHC mode.
...
+conspicuous. (Typically, this is during app startup.) The initialization of
+an actual map will then be significantly faster.
+
+This plugin surfaces this workaround to Flutter developers via
+`GoogleMapsFlutterAndroid.warmup()`. See this plugin's example code for
+one way of using this API.
+
```
stuartmorgan-g, line 71: Let's replace all of the above with just the following text:

> The first time a map is show, the Google Maps SDK may briefly block the main thread,
> which could cause UI jank. If you prefer to control when this happens, you can call
> `GoogleMapsFlutterAndroid.warmup()` at some point before showing any maps to
> pre-warm the SDK. See this plugin's example code for one way of using this API.

packages/google_maps_flutter/google_maps_flutter_android/README.md
```diff
@@ -61,6 +61,22 @@ the issue in the TLHC mode.
...
+## Warmup
+
+The first time a map is shown, the Google Maps SDK will run thread-blocking
+initialization that can easily freeze the app for hundreds of milliseconds.
+This is a limitation of the SDK on Android and has nothing to do with Flutter.
+
+In order to avoid paying this initialization price at the worst possible time
+(e.g., during a page transition), Android developers have devised a workaround
+where a "fake" map is shown first, during a time at which a thread freeze is less
+conspicuous. (Typically, this is during app startup.) The initialization of
```
stuartmorgan-g: I don't think we have any meaningful data to say when apps "typically" do this. We don't need to comment on this; developers who are using this API are presumably doing so because they have thought about where they would prefer jank to be.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -786,6 +786,15 @@ abstract class MapsInitializerApi {
   @async
   PlatformRendererType initializeWithPreferredRenderer(
       PlatformRendererType? type);
+
+  /// Asks the Google Maps SDK to do the thread-blocking work it normally does
+  /// when a map is shown for the first time.
+  ///
+  /// This gives the developer the option to move that jank to a different
+  /// part of the map (typically, the app startup, where missed frames
+  /// aren't going to be noticed).
```
stuartmorgan-g: This internal implementation API comment does not need to predict when people might choose to call this method.

packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml
stuartmorgan-g: Yes, since it adds new public API semver requires it to be a minor API bump. (That's not a repo policy, it's basic semver.)

packages/google_maps_flutter/google_maps_flutter_android/README.md
filiph: Fixed

packages/google_maps_flutter/google_maps_flutter_android/README.md
filiph: Ok. I think it's still useful to hint to the developer when such warm up makes the most sense. And while we don't have any data, it's a common practice. But let's agree to disagree, in order to land this.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
filiph: Removed.

filiph:
Thanks for the feedback. Implemented.

packages/google_maps_flutter/google_maps_flutter_android/README.md
stuartmorgan-g: > I think it's still useful to hint to the developer when such warm up makes the most sense.

And in the apps I've worked on, what this text was recommending would have been banned practice, because making the app launch quickly was a priority.

My issue here is not "hints are bad", it's that I fundamentally disagree with the practice that was being recommended here, and don't think we should steer developers to do it.

reidbaker: At a minimum we see that it can be a contested topic and developers can decide when it makes sense for their users.

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml
reidbaker: @filiph I think this is still blocking.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.17.1
```
reidbaker: After the pubspec is updated this will also need to be updated.

packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml
filiph: I think this was already addressed in https://github.com/flutter/packages/pull/9674/commits/8b9878efa1e4d8ce996c90c7bcdf300bf6b8d1f8 ? Or at least I see this as addressed both locally and here on github.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
filiph: I think this was addressed with https://github.com/flutter/packages/pull/9674/commits/8b9878efa1e4d8ce996c90c7bcdf300bf6b8d1f8 and is already at 2.18.0.

reidbaker: I am sorry. I see the changes now I must have had an older versions cached in my browser.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -532,16 +536,25 @@ class GoogleMapsFlutterAndroid extends GoogleMapsFlutterPlatform {
...
       PlatformRendererType.legacy => AndroidMapRenderer.legacy,
     };
   }
 
+  /// Asks the Google Maps SDK to do the thread-blocking work it normally does
+  /// when a map is shown for the first time.
+  ///
+  /// This gives the developer the option to move that jank to a different
+  /// part of the map (typically, the app startup, where missed frames
+  /// aren't going to be noticed).
```
stuartmorgan-g: Please remove the parenthetical here, per the discussion in the README.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -532,16 +536,25 @@ class GoogleMapsFlutterAndroid extends GoogleMapsFlutterPlatform {
...
+    final PlatformRendererType initializedRenderer = await _initializerApi
+        .initializeWithPreferredRenderer(preferredRenderer);
 
     return switch (initializedRenderer) {
       PlatformRendererType.latest => AndroidMapRenderer.latest,
       PlatformRendererType.legacy => AndroidMapRenderer.legacy,
     };
   }
 
+  /// Asks the Google Maps SDK to do the thread-blocking work it normally does
```
stuartmorgan-g: How about "Attempts to trigger any thread-blocking work the Google Maps SDK normally does when a map is shown for the first time."

It's not actually asking the SDK to do this; that implies some officially supported API intended for this purpose. Instead it's trying to trigger it via reliance on internal implementation details of the SDK that could change at any time.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.18.0
+
+* Adds GoogleMapsFlutterAndroid.warmup().
```
stuartmorgan-g: Nit: backticks around the code.

Also, this should provide some client-facing context about what this is, rather than someone needing to go read the API docs to understand what this method is for.

filiph: I came up with

```md
* Adds support for warming up the Google Maps SDK 
  via `GoogleMapsFlutterAndroid.warmup()`.
```

I try to balance the laconic style of the changelog with giving some context, but I feel I'm botching it?

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
filiph: I really like this, thanks for the suggestion! Implemented.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
filiph: Done.

packages/google_maps_flutter/google_maps_flutter_android/pigeons/messages.dart
```diff
@@ -786,6 +786,11 @@ abstract class MapsInitializerApi {
   @async
   PlatformRendererType initializeWithPreferredRenderer(
       PlatformRendererType? type);
+
+  /// Attempts to trigger any thread-blocking work
+  /// the Google Maps SDK normally does when a map is shown for the first time.
+  @async
```
stuartmorgan-g: Sorry, I missed this in previous review: why is this marked `@async`? There's no async step in the Java code, it's all synchronous. Without the `@async` annotation the Java code would be simpler, without the possibility of accidentally not calling the `result`.

filiph: Ah! Damn, sorry I missed it. The original code was async but then I found out it wouldn't work so I made it synchronous — and promptly forgot about this being async.

Addressed.

stuartmorgan-g (approved):
LGTM

PR #9685, opened 2025-07-28
[video_player] Move more Obj-C logic to Dart

stuartmorgan-g:
Moves some logic from native code to Dart:
- Asset -> URL lookup.

Also does some minor native cleanup:
- Removes unnecessary state from `FVPVideoPlayerPlugin`; several registrar-vended objects were being stored, and later changes required storing the registrar itself, so now we can just get those from the registrar on demand.
- Removes nil player handling; the player can never return nil now.
- Removes `formatHint` from the Pigeon interface; it was only ever used on Android, so doesn't need to be sent across the language boundary.

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
+++ b/packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
@@ -88,9 +88,7 @@ class AVFoundationVideoPlayer extends VideoPlayerPlatform {
           dataSource.package,
         );
       case DataSourceType.network:
-        uri = dataSource.uri;
       case DataSourceType.file:
-        uri = dataSource.uri;
       case DataSourceType.contentUri:
         uri = dataSource.uri;
     }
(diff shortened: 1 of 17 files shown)


packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -74,30 +74,38 @@ class AVFoundationVideoPlayer extends VideoPlayerPlatform {
...
+          dataSource.package,
+        );
+        if (uri == null) {
+          // Throw a platform exception for compatibility with the previous
+          // implementation, which threw on the native side.
+          throw PlatformException(
+              code: 'video_player',
+              message:
+                  'Asset $asset not found in package ${dataSource.package}.');
+        }
```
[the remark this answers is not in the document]
stuartmorgan-g: This failure to understand Dart 3 has already been reported.

packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -74,30 +74,38 @@ class AVFoundationVideoPlayer extends VideoPlayerPlatform {
     final DataSource dataSource = options.dataSource;
     final VideoViewType viewType = options.viewType;
 
-    String? asset;
-    String? packageName;
     String? uri;
```
LongCatIsLooong, line 77: nit: final?

stuartmorgan-g: That would make the code fail to compile if another type were added to `DataSourceType`, which is from a different package (and thus could be updated separately by clients). And preventing that would require a `default` case, which would mean we wouldn't get a warning about a non-exhaustive cases locally as a flag to update this code. So non-`final` has enough benefits to outweigh the downside.

LongCatIsLooong: Hmm I'm not sure I understand. The switch statement below is exhaustive and for each case in the switch the `url` variable is assigned exactly once so it could be final I think? I thought there's lint for this not sure why it is not complaining.

stuartmorgan-g: > The switch statement below is exhaustive and for each case in the switch the `url` variable is assigned exactly once so it could be final I think?

Yes, *right now*, with `video_player_platform_interface` 6.4.0, it is exhaustive.

So imagine we had made it `final`, and published that as `video_player_avfoundation` 2.8.1. Then next week, we decide to add a new source type (e.g., in-memory bytes). So we make a series of PRs for the packages to add support for it, including:
- `video_player_platform_interface` 6.5.0 adds a new `DataSourceType`.
- `video_player_avfoundation` 2.9.0 adds support for the new `DataSourceType`, including updating this switch statement.

Now consider a client who updates `video_player_platform_interface` without updating `video_player_avfoundation`, which is perfectly valid. At this point the code no longer compiles.

The only way to prevent that outcome would be to make adding a new enum value be a major version change for the platform interface, and we a) try very hard to not make breaking changes to platform interface packages, and b)  don't generally consider adding an enum value to be a breaking change.

So instead we code defensively in our switch statements that use *cross-package* enums. (I actually have a long-standing open feature request for a lint that would tell people not to rely on exhaustive switches for cross-platform enums, to use in libraries, for exactly this reason.)

PR #9687, opened 2025-07-28
[tool] Remove web drive hot reload workaround

stuartmorgan-g:
`flutter drive` for web should no longer need this workaround.

See https://github.com/dart-lang/sdk/issues/60289
See https://github.com/flutter/flutter/pull/170612


stuartmorgan-g:
@nshahan Should this work now (per [this comment](https://github.com/flutter/flutter/pull/170612#discussion_r2155528252))? It looks like we are still getting hangs when removing the flag.

nshahan:
I'm expecting it to work so I'll investigate these hangs to see what is happening. What version of Flutter are these actions configured to run with?

stuartmorgan-g:
Any bot named `... master` is using [this file](https://github.com/flutter/packages/blob/main/.ci/flutter_master.version) for the Flutter version, which is rolled about once a day.

nshahan:
It looks like some of the example apps index.html files are missing `<meta charset="UTF-8" />` in the head element and that is causing the app to crash. It looks like I have it working here except for one remaining failure but I think that is unrelated. https://github.com/flutter/packages/pull/9704

Would you like to bring in my changes in that PR into this one?

stuartmorgan-g:
Odd, I wonder why that's only an issue when removing flag. Thanks for investigating!

We can just use your PR since it's already complete.

nshahan:
> Odd, I wonder why that's only an issue when removing flag.

This is an issue we discovered with our new runtime representation that supports hot reload. The old representation used a deprecated browser API to manually set the encoding of every JavaScript module independently. The current guidance is to use the meta tag and the flutter new app template has already been using that for years. We've seen this issue appear in some terse hand authored index.html files (usually in tests) that didn't set the encoding.

PR #9694, opened 2025-07-29
#167410: _initCalled completed twice

srivats22:
This PR Fixes: https://github.com/flutter/flutter/issues/167410, where _initCalled was being performed twice on the web

Based on the discussion comments I have removed the calles to _initCalled in the google_sign_in_web package

## Pre-Review Checklist

- [x] I signed the [CLA].


srivats22:
Hi,

any idea or pointers on how to fix this:
```
The following TestFailure was thrown running a test (but after
the test had completed):
Expected: throws <Instance of 'StateError'>
  Actual: <Closure: () => Future<Null> from: () => {
                            let t$goto = 0, t$completer =
async._makeAsyncAwaitCompleter(T.Null());
                            var t$36asyncBody =
async._wrapJsFunctionForAsync((t$errorCode, t$result) => {
                              if (t$errorCode === 1) return
async._asyncRethrow(t$result, t$completer);
                              while (true)
                                switch (t$goto) {
                                  case 0:
                                    // Function start
                                    t$goto = 2;
                                    return
async._asyncAwait(t$36$35plugin$35get().disconnect(C[8] ||
CT.C8), t$36asyncBody, t$completer);
                                  case 2:
                                    // returning from await.
                                    // implicit return
                                    return
async._asyncReturn(null, t$completer);
                                }
                            });
                            return
async._asyncStartSync(t$36asyncBody, t$completer);
                          }>

```

its failing for the same reason in 3 of the tests and the repo check I know the issue which I will fix...

stuartmorgan-g:
I'm not sure I understand the question. The expectations that check that calling a method without calling init throws a `StateError` are failing because you removed the code that throws the `StateError` if a method is called without calling init.

srivats22:
Oh so the changes in the PR is incorrect? Or something else needs to change?

stuartmorgan-g:
Tests that expect that the web implementation asserts init completion everywhere are no longer valid if the web implementation no longer asserts init completion. Intentional behavioral changes often require changing tests.

srivats22:
Understood let me look at the code and see where the initCompleted is still be called in the test and remove those... Last I checked wasn't able to find any

stuartmorgan-g:
From triage: Is this ready for review?

srivats22:
Yes its ready for review looks like I might need to fix the conflicts which I will get to

packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart
```diff
@@ -89,28 +86,14 @@ class GoogleSignInPlugin extends GoogleSignInPlatform {
...
   /// A future that resolves when the plugin is fully initialized.
   ///
   /// This ensures that the SDK has been loaded, and that the `init` method
   /// has finished running.
   @visibleForTesting
   Future<void> get initialized {
-    _assertIsInitCalled();
     return Future.wait<void>(<Future<void>>[
       _jsSdkLoadedFuture,
-      _initCalled!.future,
```
mdebbar, line 113: I don't think this is the right fix.

There's a chance `_jsSdkLoadedFuture` completes and app code starts using the plugin before `_gisSdkClient` has been set (in `init()`).

packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart
```diff
@@ -89,28 +86,14 @@ class GoogleSignInPlugin extends GoogleSignInPlatform {
...
     ]);
```
mdebbar: This can become:
```dart
Future<void> get initialized {
  _assertIsInitCalled();
  return _initCalled!.future; // `init()` already awaits `_jsSdkLoadedFuture` so there's no need to wait for it here.
}
```

packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart
```diff
@@ -133,12 +116,8 @@ class GoogleSignInPlugin extends GoogleSignInPlatform {
       ' or pass clientId when initializing GoogleSignIn',
     );
 
-    assert(
-      params.serverClientId == null,
-      'serverClientId is not supported on Web.',
-    );
-
-    _initCalled = Completer<void>();
```
mdebbar, line 141: This may work by putting the completer in a local variable:
```dart
final initCalled = _initCalled = Completer<void>();
```

and later:
```dart
initCalled.complete(); // Use the local completer, not `_initCalled`
```

mdebbar (commented):
Thanks for contributing this fix!

Some thoughts:

When the app calls `init()` twice (potentially with different `params`), it becomes a tricky situation. We already have the first `init()` in flight while the second `init()` is being executed.

I think we should let both `init()` calls continue to completion. Any calls to `await initialized;` should await the latest `init()` call.

```dart
init(InitParameters(clientId: 111));
init(InitParameters(clientId: 222));
await initialized;
// At this point, the plugin should be ready with clientId 222.
```

@stuartmorgan-g what do other platforms do in this case?

srivats22:
I made the changes based on what was put in the issue... What I got from it was since the logic has been moved outside the package the code for init seems redundant... I might have miss understood let me take a look based on the comments and try and incorporate it

stuartmorgan-g:
> @stuartmorgan-g what do other platforms do in this case?

Nothing; in the new version of `google_sign_in`, clients explicitly call an `initialize` method (rather than it being an internal, implicit thing on most platforms, but semi-exposed on web, as was the case before), and the docs say they must do so exactly once. It's a programming error by the client to write the code you've shown there.

If we think clients need explicit `Error`s when doing the wrong thing, that should be added to the app-facing package, rather than handled in each implementation. This code is a legacy of the exposed-for-web nature of the previous init code.

(Also, clients should basically never call the platform interface methods directly. That's not how most federated plugins are designed.)

packages/google_sign_in/google_sign_in_web/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.1
+
+* Fixes Bad state: Future already completed on the web.
```
stuartmorgan-g: This sounds like there was a bug in the plugin that was fixed, which is not the case. This should say something like "Removes initialization checks, since initialization handling has moved to the app-facing package."

packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart
stuartmorgan-g: As of `google_sign_in` 7.x [it is a programming error at the app-facing package level](https://github.com/flutter/packages/blob/dc48f5cf75af83d5798317868f55e2268493e12a/packages/google_sign_in/google_sign_in/lib/google_sign_in.dart#L260-L261) for app to code to fail to [(indirectly) await this layer's init()](https://github.com/flutter/packages/blob/dc48f5cf75af83d5798317868f55e2268493e12a/packages/google_sign_in/google_sign_in/lib/google_sign_in.dart#L289-L302).

packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart
```diff
@@ -89,28 +86,14 @@ class GoogleSignInPlugin extends GoogleSignInPlatform {
...
-      );
-    }
-  }
-
   /// A future that resolves when the plugin is fully initialized.
   ///
   /// This ensures that the SDK has been loaded, and that the `init` method
   /// has finished running.
   @visibleForTesting
   Future<void> get initialized {
```
stuartmorgan-g: This method should just be removed.

srivats22: Will work on it and push the updated changes by today if possible

srivats22: Removing the function completely won't work cause the renderButton uses the initialized function as the future

stuartmorgan-g: That code is wrong now though; `initialized` is now equivalent to just `_jsSdkLoadedFuture` in this PR, but the code in `renderButton` that is being guarded requires `_gisSdkClient` to be non-null, and there's no order guarantee between those two things.

If we want to keep `renderButton` callable before initialization (I'm skeptical that that's actually important with the new API structure), then the internals need to be restructured to make that actually work. But that can be done with an internal future that doesn't have a getter wrapping it.

srivats22: Understood so how should I proceed??

stuartmorgan-g: @mdebbar Do you want `renderButton` to be callable before the plugin client has run `initialize` at the app-facing package level?

stuartmorgan-g: From triage: Ping @mdebbar on the question above.

mdebbar: I looked at iOS and Android implementations and I didn't see explicit errors being thrown if `init()` is called twice.

The documentation says: `or calling this method more than once, will result in undefined behavior`. I would say throwing an error qualifies as "undefined behavior" :)

Here's my suggestion:
1. Make `initialized` a private property `_initialized` to be used by methods in this file, and returned from `init()` to be used by users.
2. Throw if `init()` is called more than once.
3. `_gisSdkClient` should become `late final GisSdkClient _gisSdkClient;` so that it can only be set once, and it throws if accessed before initialization.

srivats22:
Thank you, I will work on it with the directions provided and update the PR

packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart
```diff
@@ -125,6 +108,13 @@ class GoogleSignInPlugin extends GoogleSignInPlatform {
 
   @override
   Future<void> init(InitParameters params) async {
+    // Throw if init() is called more than once
+    if (_initCalled.isCompleted) {
+      throw StateError(
+        'init() has already been called. Calling init() more than once results in undefined behavior.',
+      );
+    }
```
mdebbar: This will prevent:
```dart
await plugin.init(...);
await plugin.init(...); // This will throw.
```
But will not prevent this:
```dart
plugin.init(...);
plugin.init(...); // This will NOT throw.
```

One way to solve it is to introduce a new boolean:

```suggestion
    if (_isInitCalled) {
      throw StateError(
        'init() has already been called. Calling init() more than once results in undefined behavior.',
      );
    }
    
    _isInitCalled = true;
```

packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart
```diff
@@ -68,49 +72,28 @@ class GoogleSignInPlugin extends GoogleSignInPlatform {
...
+  late final GisSdkClient _gisSdkClient;
 
   /// A future that resolves when the plugin is fully initialized.
   ///
   /// This ensures that the SDK has been loaded, and that the `init` method
   /// has finished running.
-  @visibleForTesting
-  Future<void> get initialized {
-    _assertIsInitCalled();
+  Future<void> get _initialized {
```
mdebbar: To avoid the unnecessary creation of new futures:
```dart
Future<void> get _initialized => _initCalled!.future;
```

packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart
```diff
@@ -138,27 +128,28 @@ class GoogleSignInPlugin extends GoogleSignInPlatform {
...
       hostedDomain: params.hostedDomain,
       authenticationController: _authenticationController,
       loggingEnabled: kDebugMode,
     );
 
-    _initCalled!.complete(); // Signal that `init` is fully done.
+    _initCalled.complete();
+
+    // Return the initialized future for external users
+    return _initialized;
```
mdebbar: Returning this isn't necessary actually. `async` functions automatically create a future that resolves when the function's execution is complete.

srivats22: Thank you for the comments, i will implement these, fix the test failur and push them again

srivats22:
Is it ok if I close this PR and open a new one?

stuartmorgan-g:
@srivats22 What would the goal of doing that be? Making it harder to find all the review context makes continuing the review much harder.

srivats22:
> @srivats22 What would the goal of doing that be? Making it harder to find all the review context makes continuing the review much harder.

Oh ok no worries I will just push to this... Was just thinking the number of commits were increasing hence...

stuartmorgan-g:
From triage: @srivats22 Are you still planning on updating this PR based on the review feedback?

srivats22:
> From triage: @srivats22 Are you still planning on updating this PR based on the review feedback?

Hi,

yes I am still working on it... I had posted a question:
Quick question on this, should _isInitCalled be a local boolean? or how should I procceed on it? Once I get the answer I will try and finalize it by this week

stuartmorgan-g:
> I had posted a question: Quick question on this, should _isInitCalled be a local boolean?

I don't see anywhere in the discussion above where you asked that; the only comment since the last review was when you asked about creating a new PR.

You'll need to provide more context for the question, since I'm not sure what you mean. I don't see how a local variable would be relevant to the previous discussion.

srivats22:
> > I had posted a question: Quick question on this, should _isInitCalled be a local boolean?
> 
> I don't see anywhere in the discussion above where you asked that; the only comment since the last review was when you asked about creating a new PR.
> 
> You'll need to provide more context for the question, since I'm not sure what you mean. I don't see how a local variable would be relevant to the previous discussion.
Screenshot 2025-12-03 at 9 48 14 AM
Sorry I had accidently put it as a review comment have attached a screenshot of the same

mdebbar:
@srivats22 yes `_isInitCalled` should be private and serve a single purpose: detect when `init()` is called more than once.

packages/google_sign_in/google_sign_in_web/lib/google_sign_in_web.dart
```diff
@@ -68,51 +72,31 @@ class GoogleSignInPlugin extends GoogleSignInPlatform {
 
   // A future that completes when the JS loader is done.
   late Future<void> _jsSdkLoadedFuture;
-  // A future that completes when the `init` call is done.
-  Completer<void>? _initCalled;
+
+  /// A completer used to track whether [init] has been called.
+  ///
+  /// Completes when [init] finishes, and is used to prevent multiple
+  /// initialization calls.
```
mdebbar:
```suggestion
  /// A completer used to track whether [init] has finished.
```

packages/google_sign_in/google_sign_in_web/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 * Updates minimum supported SDK version to Flutter 3.32/Dart 3.8.
 
+## 1.1.1
+
+* Fixes Bad state: Future already completed on the web.
+
```
mdebbar, line 5: 1. This CHANGELOG entry is out of place. It should be combined with the `NEXT` above.
2. With all the recent changes in this PR, I don't think this entry message is accurate anymore. The PR doesn't "fix" the bad state. Instead, it throws a more clear and actionable error when the user makes a mistake (i.e. calls `init` more than once).

srivats22:
How do I fix the build failures?

packages/google_sign_in/google_sign_in_web/example/integration_test/google_sign_in_web_test.dart
```diff
@@ -65,8 +65,20 @@ void main() {
...
+      );
 
-      expect(plugin.initialized, completes);
+      // Calling init() a second time should throw state error
+      expect(
+        () => plugin.init(
+          const InitParameters(clientId: 'some-non-null-client-id'),
+        ),
+        throwsStateError,
+      );
```
mdebbar, line 81: Please add another test for multiple synchronous calls:
```dart
    testWidgets('throws if init is called twice synchronously', (_) async {
      final firstInit = plugin.init(
        const InitParameters(clientId: 'some-non-null-client-id'),
      );

      // Calling init() a second time synchronously should throw state error
      expect(
        () => plugin.init(
          const InitParameters(clientId: 'some-non-null-client-id'),
        ),
        throwsStateError,
      );
      
      await firstInit;
    });
```

packages/google_sign_in/google_sign_in_web/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.1.1
+
+* Throws a more actionable error when init is called more than once
```
mdebbar: This needs to be combined with the items from `NEXT` section:

```
## 1.1.1

* Throws a more actionable error when init is called more than once
* Updates minimum supported SDK version to Flutter 3.32/Dart 3.8.
```

stuartmorgan-g:
From triage: What is the status of this PR? It looks like there are still CI failures that need to be addressed.

srivats22:
> From triage: What is the status of this PR? It looks like there are still CI failures that need to be addressed.

I tried fixing it based on the failures but it kept comming back not sure what needs to change

stuartmorgan-g:
Have you run the integration test with the debugger to find out which expectation is failing?

srivats22:
I did let me try it again and see

srivats22:
I looked at the errors again

The Linux repo checks failed cause the change log was not correct but I addressed that

the others are failing cause of this:
Unhandled exception:
DriverError: Error while reading FlutterDriver result for command: window.$flutterDriver('{"command":"request_data","timeout":"1200000"}')
Original error: Expected: not null
  Actual: <null>
  
  not sure how to proceed

stuartmorgan-g:
> not sure how to proceed

Have you tried running that failing integration test under the debugger, with breakpoints, to find out where exactly the test expectation is failing, as I suggested in my previous comment?

srivats22:
> > not sure how to proceed
> 
> Have you tried running that failing integration test under the debugger, with breakpoints, to find out where exactly the test expectation is failing, as I suggested in my previous comment?

I have been pasting the below command into my terminal and been running it
```
dart run script/tool/bin/flutter_plugin_tools.dart drive-examples --packages=google_sign_in/google_sign_in_web --web
```

and then it runs but gets stuck after sometime

Am i suppose to be running it differently?

srivats22:
All tests have passed sorry for the delay...

mdebbar:
From web triage: @stuartmorgan-g would you be able to take a final look?

PR #9704, opened 2025-07-30
[tool] Remove web drive hot reload workaround

nshahan:
flutter drive for web should no longer need this workaround.

See dart-lang/sdk#60289
See flutter/flutter#170612

Add charset to some index.html files which is needed in the new DDC module system.


stuartmorgan-g:
CHANGELOG override: the minor example app config change is not client-relevant.

stuartmorgan-g (approved):
LGTM

PR #9705, opened 2025-07-30
[Local_Auth] Differentiate iOS authentication errors - 

okorohelijah:
On iOS, the plugin previously returned a generic `NotAvailable`
PlatformException for several distinct failure conditions. This made
it impossible for developers to handle specific cases e.g user
cancellation.

This change maps the native LAError codes to more specific
PlatformExceptions:
- `LAError.userCancel` is now mapped to `UserCancelled`.
- `LAError.biometryLockout` is now mapped to `LockedOut`.
- `LAError.userFallback` is now mapped to `UserFallback`.

Fixes #148942

## Pre-Review Checklist

- [X] I signed the [CLA].
- [] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or I have commented below to indicate which [version change exemption] this PR falls under[^1].


LouiseHsu (changes requested):
I took a quick skim and it looks pretty good, but I think this PR might need tests added to https://github.com/flutter/packages/blob/main/packages/local_auth/local_auth_darwin/darwin/Tests/FLALocalAuthPluginTests.swift ?

stuartmorgan-g:
Changing the codes of existing exceptions is potentially client-breaking. Is there a compelling reason this needs to be done as a semi-breaking change just for iOS, still using the anti-pattern of requiring clients to do string matching against magic, undocumented string values, rather than doing a full breaking change to [correctly fix this across all implemenations](https://github.com/flutter/flutter/issues/113687#issuecomment-1284010735)?

okorohelijah:
@stuartmorgan-g Please can you clarify what you mean ? I don't think i altered any original exceptions - I only added more exceptions

stuartmorgan-g:
> I don't think i altered any original exceptions

How so? The PR description lists three cases where a specific case used to have one exception code, and now has a different code. Anyone who has code trying to handle those cases by matching the exception code string (e.g., by providing a list of possible problems to the user) will now not have that code triggered.

It's true that the old situation wasn't good and this allows for better handler targeting going forward, but this change could still require people to update their code, and it seems less than ideal to do that twice (once with this change, and then again later to use consistent, structured exceptions) rather than going directly to the final state.

okorohelijah:
@stuartmorgan-g I appreciate you taking the time to clarify. I apologize my PR description was not clear (i have updated it).

My approach is that since so many different errors were all funneled into the generic NotAvailable code, I am simply adding new, more specific exceptions(there were no previous implementations of these exceptions)

However, I now understand your point that this is a semi-breaking change in the sense that developers who were previously handling a user cancellation by checking for e.code == 'NotAvailable' would find that their logic no longer triggers after this change, as they would now need to check for the new, more specific code.

packages/local_auth/local_auth_darwin/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 1.5.1
+
+*  Provides more specific error codes on iOS for authentication failures.
+*  Previously, several distinct errors (like user cancellation or biometric lockout) would return a generic `NotAvailable` exception.
+*  These failures now return specific codes. You may need to update your error handling logic to check for these new codes:
+    * `LockedOut` is now returned for biometric lockout.
+    * `UserCancelled` is now returned when the user cancels the prompt.
+    * `UserFallback` is now returned when the user selects the fallback option.
+
```
LouiseHsu, line 7: @stuartmorgan-g do you think we require a breaking change warning here?

LouiseHsu (approved):
left a minor comment on the changelog, but otherwise lgtm!

packages/local_auth/local_auth_darwin/CHANGELOG.md
stuartmorgan-g: I think we can reasonably consider this a non-breaking change. The set of possible code strings is not actually documented, so it's very unlikely that anyone is writing error handlers that assume that they have exhaustive checks. And since the old error was so generic, it would be hard for handlers targeting that code to provide much feedback.

Given the low chance of actually breaking anyone (and no possibility of compile-level breakage), I'm fine with non-breaking.

packages/local_auth/local_auth_darwin/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 1.5.1
```
stuartmorgan-g: This should be a minor version, rather than a bugfix version, as it's effectively new API.

packages/local_auth/local_auth_darwin/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
...
+
+*  Provides more specific error codes on iOS for authentication failures.
+*  Previously, several distinct errors (like user cancellation or biometric lockout) would return a generic `NotAvailable` exception.
+*  These failures now return specific codes. You may need to update your error handling logic to check for these new codes:
```
stuartmorgan-g: All three of these are describing the same change, so should not be three separate bullet points.

stuartmorgan-g: Let's remove "You may need to update your error handling logic to check for these new codes", as in practice I don't think people are going to *need* to update their logic (if we did think that, we'd version this as a breaking change).

packages/local_auth/local_auth_darwin/example/ios/Runner.xcodeproj/project.pbxproj
stuartmorgan-g, line 1: The changes to this file should be reverted. Running examples on `master` will auto-migrated some settings that we don't want checked in, as these should continue to support everything `stable` supports.

packages/local_auth/local_auth_darwin/CHANGELOG.md
okorohelijah: Fixed.

packages/local_auth/local_auth_darwin/CHANGELOG.md
okorohelijah: Fixed.

packages/local_auth/local_auth_darwin/CHANGELOG.md
okorohelijah: Fixed

packages/local_auth/local_auth_darwin/example/ios/Runner.xcodeproj/project.pbxproj
okorohelijah: Makes sense! Fixed.

PR #9718, opened 2025-07-31
[video_player] Improve KVO handling on iOS

stuartmorgan-g:
This improves the KVO handling in `FVPVideoPlayer` in several ways:
- Eliminates the use of methods to do KVO registration and unregistration, since that involved calling `self` during init and dealloc, which is a dangerous anti-pattern.
- Eliminates `NSKeyValueObservingOptionInitial` from the registration, as this was non-obviously creating a call to self, within init, for every registration. In practice, none of them were necessary from what I could determine:
  - The handlers for most keys just collect information to send to the event listener, but there is no event listener during init (it's set just after), so those were expensive no-ops i.
  - The handlers related to initialization require the item to be in the `AVPlayerItemStatusReadyToPlay` state, and items start in `AVPlayerItemStatusUnknown`, so the initial state call won't do anything useful.
- Stores a dictionary of registrations, which is used for unregistration, so that it's impossible to add a new key to the observation set but forget to unregister it. Previously the two lists of strings had to be kept in sync. New static functions (to avoid `self` calls) process those lists.
- Clears the list after unregistering to avoid double-unregistration locally to the class, instead of requiring a subclass to contain a workaround to avoid problems in the superclass code.

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
+++ b/packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
@@ -18,21 +18,19 @@ static void *rateContext = &rateContext;
 
 /// Registers KVO observers on 'object' for each entry in 'observations', which must be a
 /// dictionary mapping KVO keys to NSValue-wrapped context pointers.
+///
+/// This does not call any methods on 'observer', so is safe to call from 'observer's init.
 static void FVPRegisterKeyValueObservers(NSObject *observer,
                                          NSDictionary<NSString *, NSValue *> *observations,
                                          NSObject *target) {
-  // TODO(stuartmorgan): Investigate why NSKeyValueObservingOptionInitial is being used here. This
-  // code is called from the player's init method, whith self as the observer, and
-  // NSKeyValueObservingOptionInitial is documented to synchronously call the observer during
-  // the registration, so this is causing invocations to self during init, which is dangerous.
-  // Most of the change handlers are just calling a listener that's not even set up yet, so many
-  // of the calls will not do anything useful anyway. Ideally NSKeyValueObservingOptionInitial
-  // should be removed, and if the initialization code potentially needs to be triggered that should
-  // be done explicitly.
+  // It is important not to use NSKeyValueObservingOptionInitial here, because that will cause
+  // synchronous calls to 'observer', violating the requirement that this method does not call its
+  // methods. If there are use cases for specific pieces of initial state, those should be handled
+  // explicitly by the caller, rather than by adding initial-state KVO notifications here.
   for (NSString *key in observations) {
     [target addObserver:observer
              forKeyPath:key
-                options:NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionNew
+                options:NSKeyValueObservingOptionNew
                 context:observations[key].pointerValue];
   }
 }
--- a/packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
+++ b/packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
@@ -16,7 +16,51 @@ static void *durationContext = &durationContext;
 static void *playbackLikelyToKeepUpContext = &playbackLikelyToKeepUpContext;
 static void *rateContext = &rateContext;
 
-@implementation FVPVideoPlayer
+/// Registers KVO observers on 'object' for each entry in 'observations', which must be a
+/// dictionary mapping KVO keys to NSValue-wrapped context pointers.
+static void FVPRegisterKeyValueObservers(NSObject *observer,
+                                         NSDictionary<NSString *, NSValue *> *observations,
+                                         NSObject *target) {
+  // TODO(stuartmorgan): Investigate why NSKeyValueObservingOptionInitial is being used here. This
+  // code is called from the player's init method, whith self as the observer, and
+  // NSKeyValueObservingOptionInitial is documented to synchronously call the observer during
+  // the registration, so this is causing invocations to self during init, which is dangerous.
+  // Most of the change handlers are just calling a listener that's not even set up yet, so many
+  // of the calls will not do anything useful anyway. Ideally NSKeyValueObservingOptionInitial
+  // should be removed, and if the initialization code potentially needs to be triggered that should
+  // be done explicitly.
+  for (NSString *key in observations) {
+    [target addObserver:observer
+             forKeyPath:key
+                options:NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionNew
+                context:observations[key].pointerValue];
+  }
+}
+
+/// Registers KVO observers on 'object' for each entry in 'observations', which must be a
+/// dictionary mapping KVO keys to NSValue-wrapped context pointers.
+///
+/// This should only be called to balance calls to FVPRegisterKeyValueObservers, as it is an
+/// error to try to remove observers that are not currently set.
+///
+/// This does not call any methods on 'observer', so is safe to call from 'obsever's dealloc.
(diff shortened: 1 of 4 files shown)


stuartmorgan-g:

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -82,68 +124,55 @@ - (instancetype)initWithPlayerItem:(AVPlayerItem *)item
...
   _disposed = YES;
-  [self removeKeyValueObservers];
 
-  [self.player replaceCurrentItemWithPlayerItem:nil];
   [[NSNotificationCenter defaultCenter] removeObserver:self];
+  FVPRemoveKeyValueObservers(self, _playerItemObservations, self.player.currentItem);
+  FVPRemoveKeyValueObservers(self, _playerObservations, self.player);
+  // Clear the list of observations, so that dealloc (or potential duplicate dispose calls, in the
+  // case of hot restart) don't try to re-remove them, which would be an error.
+  _playerItemObservations = @{};
```
LongCatIsLooong: Is this ivar essentially serving the same purpose as `_disposed`? If so does it make sense to make the two dictionaries static and then just use `_dispose` instead as the source of the truth?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -82,68 +124,55 @@ - (instancetype)initWithPlayerItem:(AVPlayerItem *)item
...
-  }
+  // In case dispose was never called for some reason, remove any remaining observers to prevent
+  // crashes.
+  FVPRemoveKeyValueObservers(self, _playerItemObservations, _player.currentItem);
+  FVPRemoveKeyValueObservers(self, _playerObservations, _player);
 }
 
 - (void)dispose {
   _disposed = YES;
-  [self removeKeyValueObservers];
```
LongCatIsLooong, line 161: not directly related to this PR: can we return early if `_dispose` is already YES, or this method shouldn't be a no-op if called a second time?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -82,68 +124,55 @@ - (instancetype)initWithPlayerItem:(AVPlayerItem *)item
...
+    @"status" : [NSValue valueWithPointer:statusContext],
+    @"presentationSize" : [NSValue valueWithPointer:presentationSizeContext],
+    @"duration" : [NSValue valueWithPointer:durationContext],
+    @"playbackLikelyToKeepUp" : [NSValue valueWithPointer:playbackLikelyToKeepUpContext],
+  };
+  _playerObservations = @{
+    @"rate" : [NSValue valueWithPointer:rateContext],
+  };
+  FVPRegisterKeyValueObservers(self, _playerItemObservations, item);
+  FVPRegisterKeyValueObservers(self, _playerObservations, _player);
```
LongCatIsLooong: It looks like the original implementation specified `NSKeyValueObservingOptionInitial`, but here I don't see how the initial values are going to be reported?

LongCatIsLooong: In the PR description it says:

> The handlers for most keys just collect information to send to the event listener, but there is no event listener during init (it's set just after), so those were expensive no-ops.
> The handlers related to initialization require the item to be in the AVPlayerItemStatusReadyToPlay state, and items start in AVPlayerItemStatusUnknown, so the initial state call won't do anything useful.

So this means the item is always in `.unknown` state when the event listener is set? Is it possible to get an item that's already `.failed`?

stuartmorgan-g: Not according to [the docs](https://developer.apple.com/documentation/avfoundation/avplayeritem/status-swift.property):

> When a player item is created, its [status](https://developer.apple.com/documentation/avfoundation/avplayeritem/status-swift.property) is [AVPlayerItem.Status.unknown](https://developer.apple.com/documentation/avfoundation/avplayeritem/status-swift.enum/unknown), meaning its media hasn’t been loaded and has not yet been enqueued for playback. Associating a player item with an [AVPlayer](https://developer.apple.com/documentation/avfoundation/avplayer) immediately begins enqueuing the item’s media and preparing it for playback.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
stuartmorgan-g: Ideally it shouldn't ever be called multiple times, it's just a think that can happen with hot reload the way things are wired up currently. Returning early should be fine; I'll change it.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
stuartmorgan-g: That should work. Originally I was going to get rid of `_disposed` after adding this, but it turned out to be used elsewhere to bail out of async callbacks.

stuartmorgan-g: Done. (I didn't actually make the dictionaries themselves static since that would add a lot of complexity for no real benefit, just the accessors.)

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
LongCatIsLooong: Hmm I feel that doesn't guarantee that `AVPlayer.status` can't change synchronously especially when the KVO registration happens after `_player = [avFactory playerWithPlayerItem:item];` (which I assume is when the player item is associated with an `AVPlayer`). On the other hand I agree that it probably doesn't make too much sense for `AVFoundation` to change the `status` synchronously (maybe if the `AVPlayer` implementation has logic that does simple sanity checks as soon as a new item is assigned?). Should we add an assert before the registration to make sure the status is `.unknown`?

stuartmorgan-g: > especially when the KVO registration happens after `_player = [avFactory playerWithPlayerItem:item];` (which I assume is when the player item is associated with an `AVPlayer`)

That's a great point; I wasn't considering the order of operations there if the internals make synchronous status changes. Rather than relying on undocumented behavior and asserting, I've instead moved the KVO registration to `setEventListener:`. That makes things much easier to reason about, because we're no longer doing any of this from init, so we can call whatever we want, and unlike the assert means clients won't get incorrect runtime behavior if AVPlayer ever starts actually doing that.

Rather than add back `NSKeyValueObservingOptionInitial` (which would cause a bunch of initial state event updates in Dart that didn't used to exist, and in a random order), I've added an explicit call to check the state and, if it has already changed, send an initialization event (which is now safe because this isn't calling methods from `init`) before registering the listeners.

(The listener is set right after init, so in practice this will have almost no effect on when any of this happens, but I think this structure is much cleaner.)

stuartmorgan-g:
> LGTM (I think you mentioned that the player instances will be re-created on hot-restart right?)

Yes, there's Dart code in the plugin that re-initializes the native code if a hot restart happens.

PR #9762, opened 2025-08-06
Path provider foundation ffi

stuartmorgan-g:
Converts `path_provider_foundation` from a Flutter plugin to a Dart-only package using `ffigen` to call into Foundation directly.

Most of the Dart unit tests had to be converted to integration tests, although they are still conceptually unit tests (i.e., they still inject mocks for the platform calls). This is because unlike Pigeon, where the API layer being mocked uses Dart types, the FFI calls that we now mock use Obj-C types, and those don't exist outside of an actual app runtime, so the integration test driver is the only place they can be hosted. This is very similar to how most of our web plugin implementations are unit tested.

Fixes https://github.com/flutter/flutter/issues/111033

## Pre-Review Checklist

- [x] I signed the [CLA].


stuartmorgan-g:
This is blocked on https://github.com/dart-lang/native/issues/2477 (or a policy decision not to wait for that change), but I wanted to go ahead with a review and discussion in the meantime.

Unlike many of our other plugins, adopting FFI here isn't blocked on non-optional thread merging, since per [Apple docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html), `NSBundle` and `NSFileManager` are threadsafe, and `NSURL` can be presumed thread-safe as an immutable data class ("Mutable Versus Immutable" on that page).

packages/path_provider/path_provider_foundation/darwin/path_provider_foundation/Sources/path_provider_foundation/Resources/PrivacyInfo.xcprivacy
```diff
@@ -1,14 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
-<plist version="1.0">
-<dict>
```
vashworth, line 4: Do we know how Privacy Manifests are going to work for FFI? If I remember correctly on how the tooling works for this, I believe a dylib gets built into a `.framework`. Technically `.framework` isn't listed as being needed in [Apple's documentation on Privacy Manifests](https://developer.apple.com/documentation/bundleresources/privacy-manifest-files?language=objc), but `.xcframework` is listed, and `.xcframework` gets thinned to `.framework` when put into the app bundle. 

Basically, I think we likely still need a Privacy Manifest.

Looking at ffigen [docs](https://pub.dev/packages/ffigen#:~:text=ffi%2Dnative%3A%0A%20%20asset%2Did%3A), perhaps an asset can be added to the config?

```yaml
ffi-native:
  asset-id: 'package:some_pkg/asset' # Optional, was assetId in previous versions
```

packages/path_provider/path_provider_foundation/example/build.yaml
```diff
@@ -0,0 +1,13 @@
+targets:
```
vashworth, line 1: What is this file for? Is it specific to packages CI or is it needed for ffi?

packages/path_provider/path_provider_foundation/ffigen_config.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - 'NSURL'
+  member-filter:
+    NSBundle:
+      include:
+        - "bundleIdentifier"
+        - "mainBundle"
+    NSFileManager:
+      include:
+        - "containerURLForSecurityApplicationGroupIdentifier:"
+        - "defaultManager"
```
vashworth: Why doesn't `NSSearchPathDirectory` need to be included?

packages/path_provider/path_provider_foundation/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.5.0
```
vashworth, line 1: Should this be a breaking change? Developers who are importing the "path_provider_foundation" module directly will get a build failure since it's not longer available. Here's an example I found:

https://github.com/MixinNetwork/flutter-app/blob/1a0fd1691c3c5fd9a4d86097d73efeb438cd9ca5/macos/Runner/AppDelegate.swift#L5

packages/path_provider/path_provider_foundation/lib/path_provider_foundation.dart
```diff
@@ -2,106 +2,8 @@
...
-@visibleForTesting
-class PathProviderPlatformProvider {
-  /// Specifies whether the current platform is iOS.
-  bool get isIOS => Platform.isIOS;
-}
+// path_provider_foundation is implemented using FFI; export a stub for
+// platforms that don't support FFI (e.g., web) to avoid having transitive
+// dependencies break web compilation.
+export 'src/path_provider_foundation_stub.dart'
+    if (dart.library.ffi) 'src/path_provider_foundation_real.dart';
```
vashworth, line 9: What are your thoughts on remove `_real`? Adding `_stub` to the placeholder one seems sufficient to me

```suggestion
    if (dart.library.ffi) 'src/path_provider_foundation.dart';
```

packages/path_provider/path_provider_foundation/darwin/path_provider_foundation/Sources/path_provider_foundation/Resources/PrivacyInfo.xcprivacy
stuartmorgan-g: > Do we know how Privacy Manifests are going to work for FFI?

[Not really](https://github.com/flutter/flutter/issues/147186).

> Basically, I think we likely still need a Privacy Manifest.

I'll set up a meeting to discuss; I don't think this version of the package would still need a manifest.

packages/path_provider/path_provider_foundation/example/build.yaml
stuartmorgan-g: It's to make mockito generate mocks from annotations on files in `integration_test`; by default it will ignore them (because normally unit-style tests, where you use mocks, don't go in `integration_test`).

packages/path_provider/path_provider_foundation/ffigen_config.yaml
stuartmorgan-g: Types that are needed by an API are included transitively, so `NSSearchPathForDirectoriesInDomains` would be bringing it in.

packages/path_provider/path_provider_foundation/CHANGELOG.md
stuartmorgan-g: We don't generally consider native code changes to be breaking changes for the plugin unless the native code is *explicitly* intended for direct native use (e.g., [this file](https://github.com/flutter/packages/blob/main/packages/webview_flutter/webview_flutter_wkwebview/darwin/webview_flutter_wkwebview/Sources/webview_flutter_wkwebview/WebViewFlutterWKWebViewExternalAPI.swift)).

I can't think of a good reason to directly import this native package. Notably, that code... doesn't appear to make sense? The import is there to create an instance of the plugin that is never used and never wired up to anything. Also, even if we considered this a breaking change for `path_provider_foundation`, we wouldn't consider it a breaking change for `path_provider`, and that app is not directly relying on `path_provider_foundation` so would be broken anyway.

Ultimately, there's a large grey area of things that can break unexpected use cases, but that we don't consider breaking (e.g., adding a parameter to a method of a Dart class that's not intended to be subclassed). I would definitely include this change in that area. We have not considered it breaking in the past to change the name of the plugin class, which is very similar.

packages/path_provider/path_provider_foundation/lib/path_provider_foundation.dart
stuartmorgan-g: This is how `path_provider_windows` is set up, so I mirrored it. IIRC we did this to avoid having two files with the same name in the same package? I don't have strong feelings about it though.

stuartmorgan-g:
Yep, that's https://github.com/dart-lang/native/issues/2482

dcharkes:
FYI @liamappelbe. Maybe take a look if there's anything in here that could be made easier on the FFIgen side or you have any review comments to this PR.

MichealReed:
Why not have a simple C library that provides the paths and use a native_assets build hook to build dynamic libs for all platforms? Windows is broken now as well with a MissingPluginException. The platform differences can be accommodated with preprocessor directives on the native side. 

Edit: This would also support a wasm path in the future.

stuartmorgan-g:
> Why not have a simple C library that provides the paths

If we wanted to keep maintaining an intermediate native code layer between the system and Dart, we would just keep the implementation we already have.

Adding a C library adds complexity relative to this PR, and does not achieve the goals described in the linked issue.

> and use a native_assets build hook to build dynamic libs for all platforms?

This also just replaces what we already have with a slightly different version of the same thing. Flutter's plugin system already has hooks to build dynamic libraries for the implementations that need one.

> Windows is broken now as well with a MissingPluginException.

I don't see how this PR could possibly affect Windows, given that the Windows implementation is in another package.

> The platform differences can be accommodated with preprocessor directives on the native side.

That does not solve any problems that we have, given that the `path_provider` plugin implementations are fully federated.

MichealReed:
It would still meet the main goal as you will still need to shim the C with an ObjC header (that you could test from dart with the interop) on the library if implemented in C. Native assets seems much more reliable than the flutter plugin system, I have migrated several projects now and am only using plugins for web (have to manually invoke build on the file and distribute js and wasm binaries/loader).

While this fixes iOS/Mac, path_provider has issues on other platforms too. Is it too out of bounds to include a fix for a break with another platform?

Package path_provider:windows references path_provider_windows:windows as the default plugin, but the
package does not exist, or is not a plugin package.
Ask the maintainers of path_provider to either avoid referencing a default implementation via
`platforms: windows: default_package: path_provider_windows` or create a plugin named
path_provider_windows.

Migrating to native assets would fix all platforms simultaneously, it'd be quite easy to find the paths in C on the various platforms. You can see my work in miniav to shim multiple capture apis from native using the native assets build hooks, heavily tested from dart with native assets. I have another library that does similar from c++ for webgpu and it runs successfully on all flutter supported platforms. The sooner flutter plugins start migrating to this, the better in my opinion.

https://github.com/Practicalxr/miniAV/tree/master/miniav_ffi

dcharkes:
Let me push back a little here. If there OS-specific APIs available on Java/Kotlin/Swift/Objective-C that are not available in C in a cross-platform manner, then Federated plugins are likely the better option. As you can see in the PR from the `ffigen_config.yaml` it's binding to Objective-C APIs, specific to MacOS/iOS.

(I don't know enough about the feature-set by path provider to know if there's a ready-made C library that can provide that complete feature set on all targets. And even if that is available, that is a larger conceptual migration as @stuartmorgan-g points out. If you know of such a C library @MichealReed, then please feel free to open an issue with a proposal to refactor the path provider plugins to use that instead.)

stuartmorgan-g:
> It would still meet the main goal as you will still need to shim the C with an ObjC header (that you could test from dart with the interop) on the library if implemented in C.

As I said above:

>> If we wanted to keep maintaining an intermediate native code layer between the system and Dart, we would just keep the implementation we already have.

> Native assets seems much more reliable than the flutter plugin system

If you have reliability issues with the Flutter plugin system, please file specific issues with details.

> While this fixes iOS/Mac path_provider has issues on other platforms too.

This doesn't fix anything, it just restructures internals. If you have issues with `path_provider`, you should file issues with details.

> Is it too out of bounds to include a fix for a break with another platform?

Fixing issues with `path_provider` on some unrelated platform is out of scope for an internal refactoring of `path_provider_foundation`, yes.

> Package path_provider:windows references path_provider_windows:windows [...]

This sound like you are trying to report an issue with `path_provider_windows`. Please use the issue tracker for that, not a PR in an unrelated package.

stuartmorgan-g:
> If you know of such a C library @MichealReed, then please feel free to open an issue with a proposal to refactor the path provider plugins to use that instead.

I would wontfix an issue to add third-party code to intermediate a very simple plugin that we already have implementations of.

stuartmorgan-g:
If you want to continue discussing a proposal for a completely new architecture for all of `path_provider`, please file a new issue with a clear explanation of what benefits it provides. It's not on topic for this PR.

Piinks:

stuartmorgan-g:
I'm currently most of the way through updating this PR to use the recently-released updated versions of `objective_c` and `ffigen`, which should unblock releasing it.

stuartmorgan-g:
New warnings with `objective_c` 9.0; I'll file them upstream.

Edit: https://github.com/dart-lang/native/issues/2776

stuartmorgan-g:
From triage: The issue is fixed upstream; blocked on that fix being published to pub.dev.

PR #9771, opened 2025-08-07
[video_player] Move Android buffer updates to Dart

stuartmorgan-g:
Since ExoPlayer does not generate events for changes to the buffer position, the `video_player_android` synthesizes them via polling. This moves the logic to generate these events from Java to Dart, to reduce the amount of logic on the Java side.

This does not attempt to fix the structural problem that the workaround was piggy-backed on the fact that the app-facing package polls frequently for the playback position, generating buffer update events when that polling happens rather than via any independent internal polling system. This approach is not only fragile, as it depends on logic defined in another package, but already doesn't work in the case of a paused video. However, this refactoring will make it possible to change this behavior in Dart code in a follow-up focused on that behavioral change. This change should be behavior-neutral, except that it adds the simple improvement of not creating events if the buffer position hasn't changed, so that clients aren't getting a stream of useless buffer position events during playback.

This version will also be more efficient from a platform channel standpoint, since instead of sending a separate message with the buffer position, it's returned in the same call as the playback position.

To support synthesizing events on the Dart side, this adds a custom stream controller in Dart, instead of creating it directly from the EventChannel stream; this change will make future changes to a non-method-channel approach (e.g., FFI callbacks) straightforward. As this adds non-trivial additional per-player-instance state on the Dart side, this adds a Dart player instance class, and consolidates existing state into that class, including passthrough methods for the API calls, making it clearer which state is plugin level vs player level.

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java
+++ b/packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java
@@ -127,17 +127,6 @@ public final class VideoPlayerTest {
   }
 
   @Test
-  public void sendsBufferingUpdatesOnDemand() {
-    VideoPlayer videoPlayer = createVideoPlayer();
-
-    when(mockExoPlayer.getBufferedPosition()).thenReturn(10L);
-    videoPlayer.sendBufferingUpdate();
-    verify(mockEvents).onBufferingUpdate(10L);
-
-    videoPlayer.dispose();
-  }
-
-  @Test
   public void togglesLoopingEnablesAndDisablesRepeatMode() {
     VideoPlayer videoPlayer = createVideoPlayer();
 
@@ -177,14 +166,29 @@ public final class VideoPlayerTest {
   }
 
   @Test
-  public void seekAndGetPosition() {
+  public void seekTo() {
     VideoPlayer videoPlayer = createVideoPlayer();
 
     videoPlayer.seekTo(10L);
     verify(mockExoPlayer).seekTo(10);
 
-    when(mockExoPlayer.getCurrentPosition()).thenReturn(20L);
-    assertEquals(20L, videoPlayer.getPosition().longValue());
+    videoPlayer.dispose();
+  }
+
+  @Test
+  public void GetPlaybackState() {
+    VideoPlayer videoPlayer = createVideoPlayer();
+
+    final long playbackPosition = 20L;
+    final long bufferedPosition = 10L;
+    when(mockExoPlayer.getCurrentPosition()).thenReturn(playbackPosition);
+    when(mockExoPlayer.getBufferedPosition()).thenReturn(bufferedPosition);
+
+    final Messages.PlaybackState state = videoPlayer.getPlaybackState();
+    assertEquals(playbackPosition, state.getPlayPosition().longValue());
+    assertEquals(bufferedPosition, state.getBufferPosition().longValue());
+
+    videoPlayer.dispose();
   }
 
   @Test
--- a/packages/video_player/video_player_android/lib/src/android_video_player.dart
+++ b/packages/video_player/video_player_android/lib/src/android_video_player.dart
@@ -46,8 +46,9 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
 
   @override
   Future<void> dispose(int playerId) async {
+    final _PlayerInstance? player = _players.remove(playerId);
     await _api.dispose(playerId);
-    _players.remove(playerId);
+    await player?.dispose();
   }
 
   @override
@@ -132,7 +133,13 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
         ),
         VideoViewType.platformView => const _VideoPlayerPlatformViewState(),
       };
-      return _PlayerInstance(_playerProvider(playerId), viewState);
+      final String eventChannelName =
+          'flutter.io/videoPlayer/videoEvents$playerId';
+      return _PlayerInstance(
+        _playerProvider(playerId),
+        viewState,
+        eventChannelName: eventChannelName,
+      );
     });
   }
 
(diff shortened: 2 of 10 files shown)


packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -172,52 +172,17 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
...
-          eventType: VideoEventType.bufferingStart,
-        ),
-        'bufferingEnd' => VideoEvent(eventType: VideoEventType.bufferingEnd),
-        'isPlayingStateUpdate' => VideoEvent(
-          eventType: VideoEventType.isPlayingStateUpdate,
-          isPlaying: map['isPlaying'] as bool,
-        ),
-        _ => VideoEvent(eventType: VideoEventType.unknown),
-      };
-    });
```
stuartmorgan-g, line 220: This switch is moved almost unchanged to the player instance class below.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -280,6 +237,130 @@ PlatformVideoViewType _platformVideoViewTypeFromVideoViewType(
...
+        eventType: VideoEventType.isPlayingStateUpdate,
+        isPlaying: map['isPlaying'] as bool,
+      ),
+      _ => VideoEvent(eventType: VideoEventType.unknown),
+    });
+  }
+
+  // Turns a single buffer position, which is what ExoPlayer reports, into the
+  // DurationRange array expected by [VideoEventType.bufferingUpdate].
+  List<DurationRange> _bufferRangeForPosition(int milliseconds) {
```
stuartmorgan-g, line 362: This was extracted from the switch to allow it to be used in `_updateBufferingState` as well.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -280,6 +237,130 @@ PlatformVideoViewType _platformVideoViewTypeFromVideoViewType(
...
+  Future<Duration> getPosition() async {
+    final PlaybackState state = await _api.getPlaybackState();
+    // TODO(stuartmorgan): Move this logic. This is a workaround for the fact
+    //  that ExoPlayer doesn't have any way to observe buffer position
+    //  changes, so polling is required. To minimize platform channel overhead,
+    //  that's combined with getting the position, but this relies on the fact
+    //  that the app-facing package polls getPosition frequently, which makes
+    //  this fragile (for instance, as of writing, this won't be called while
+    //  the video is paused). It should instead be called on its own timer,
+    //  independent of higher-level package logic.
```
stuartmorgan-g, line 306: I've expanded on the explanation relative to the version removed from VideoPlayer.java, but it's still fundamentally the same hack as the Java version was for now (to minimize behavioral changes in this PR).

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -280,6 +237,130 @@ PlatformVideoViewType _platformVideoViewTypeFromVideoViewType(
...
+          (map['width'] as num?)?.toDouble() ?? 0.0,
+          (map['height'] as num?)?.toDouble() ?? 0.0,
+        ),
+        rotationCorrection: map['rotationCorrection'] as int? ?? 0,
+      ),
+      'completed' => VideoEvent(eventType: VideoEventType.completed),
+      'bufferingUpdate' => VideoEvent(
+        eventType: VideoEventType.bufferingUpdate,
+        buffered: _bufferRangeForPosition(map['position'] as int),
+      ),
```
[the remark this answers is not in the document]
stuartmorgan-g: This is not correct; the Java code still sends this when the player enters `STATE_BUFFERING`.

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/VideoPlayerTest.java
```diff
@@ -177,14 +166,29 @@ public void setPlaybackSpeedSetsPlaybackParametersWithValue() {
...
     videoPlayer.seekTo(10L);
     verify(mockExoPlayer).seekTo(10);
 
-    when(mockExoPlayer.getCurrentPosition()).thenReturn(20L);
-    assertEquals(20L, videoPlayer.getPosition().longValue());
+    videoPlayer.dispose();
+  }
+
+  @Test
+  public void GetPlaybackState() {
```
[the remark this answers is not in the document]
stuartmorgan-g: Fixed.

packages/video_player/video_player_android/pigeons/messages.dart
```diff
@@ -50,7 +60,12 @@ abstract class VideoPlayerInstanceApi {
   void setVolume(double volume);
   void setPlaybackSpeed(double speed);
   void play();
-  int getPosition();
   void seekTo(int position);
```
ash2moon: Can we specify that the position is also in milliseconds here?

packages/video_player/video_player_android/pigeons/messages.dart
```diff
@@ -50,7 +60,12 @@ abstract class VideoPlayerInstanceApi {
   void setVolume(double volume);
   void setPlaybackSpeed(double speed);
```
ash2moon: I think it might help to provide what that max/min bounds of these values would be or at least specify the unit.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -134,12 +121,25 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
...
+    _players.putIfAbsent(playerId, () {
+      final _VideoPlayerViewState viewState = switch (viewType) {
+        // playerId is also the textureId when using texture view.
+        VideoViewType.textureView => _VideoPlayerTextureViewState(
+          textureId: playerId,
+        ),
+        VideoViewType.platformView => const _VideoPlayerPlatformViewState(),
+      };
+      final String eventChannelName =
+          'flutter.io/videoPlayer/videoEvents$playerId';
```
ash2moon: Can we make the magic string a constant outside of this method?

ash2moon (approved):

packages/video_player/video_player_android/pigeons/messages.dart
stuartmorgan-g: Good call; added docs to all of these methods.

packages/video_player/video_player_android/pigeons/messages.dart
stuartmorgan-g: Added.

packages/video_player/video_player_android/lib/src/android_video_player.dart
stuartmorgan-g: Extracted, and added a comment about the value it has to match on the other side.

PR #9782, opened 2025-08-11
[video_player] Add audio track metadata support (bitrate, sample rate, channels, codec)

nateshmbhat:
## Description

This PR adds comprehensive audio track metadata support to the video_player package, enabling developers to access detailed information about available audio tracks including bitrate, sample rate, channel count, and codec information.

### Changes Made
- **Enhanced VideoAudioTrack model** with new metadata fields: `bitrate`, `sampleRate`, `channelCount`, `codec`
- **Added qualityDescription getter** that formats metadata into human-readable strings (e.g., "128kbps • Stereo • AAC")
- **Updated platform implementations**:
  - **Android**: Real metadata extraction from ExoPlayer Format objects
  - **iOS**: Metadata extraction from AVFoundation with HLS stream simulation
- **Updated Pigeon interfaces** for both Android and iOS platforms
- **Added comprehensive test coverage** with 96 tests passing
- **Updated example app** to display quality information in the UI

## Related Issues : 
 [#59437](https://github.com/flutter/flutter/issues/59437) - Add audio track metadata support to video_player

### Testing
- Added unit tests for all new functionality
- Tested with HLS streams on both platforms
- Verified backward compatibility

### Breaking Changes
None - all changes are additive and backward compatible.

## Pre-Review Checklist

- [x] I signed the [CLA].


nateshmbhat:
since in my code i've used the packages of video_player from the same branch's code instead of published ones using "path" option, i think some of these checks are failing. What should i do in this case ? until its published, the code would have to use "path" option right ?

stuartmorgan-g:
Thanks for the contribution! A few high-level comments before this can be reviewed:

> since in my code i've used the packages of video_player from the same branch's code instead of published ones using "path" option, i think some of these checks are failing. What should i do in this case ?

Please see [the instructions in our contributor docs](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins).

> * **Added comprehensive test coverage** with 96 tests passing

Please check that you've uploaded all of your changes; the PR as it is here does not have comprehensive test coverage. The only testing here appears to be of the Dart data class, with no testing of any of the new platform logic.

> * **Added qualityDescription getter** that formats metadata into human-readable strings (e.g., "128kbps • Stereo • AAC")

This is out of scope for the plugin, and should be removed. It's up to plugin clients how/if they want to format any data.

> ### Use Cases
> * **Adaptive streaming**: Select tracks based on network conditions
> * **Quality preferences**: Let users choose between high/low quality audio

This PR does not appear to have any support for these use cases. Is the PR description inaccurate, or is the PR still a work in progress?

nateshmbhat:
@stuartmorgan-g 

1. Removed qualityDescription formatting from library.
2. removed the use cases section in PR as the PR is currently just to fetch the audio tracks
3. went through Changing Federated plugin section doc and updated pubspec with overrides as changed by the script tool.

nateshmbhat:
@stuartmorgan-g have addressed your high level concerns. anything else that i need to do ?

stuartmorgan-g:
> @stuartmorgan-g have addressed your high level concerns. anything else that i need to do ?

There are some CI failures that still need to be addressed, but those don't need to block review. I won't be the primary reviewer here, but some things I can flag from a quick skim while this is waiting for a full review:
- This will need more testing. Looking at iOS for instance, there are almost 400 lines of tests of [this trivial passthrough statement](https://github.com/flutter/packages/pull/9782/files#diff-1a110fd39c95303a6bd97033203f739866aab30518994ac55bf134e7d9f60e03R217-R228), but no testing at all of the ~300 lines of complex branching Obj-C code. The native Android code is much simpler, but also entirely untested. Future changes to this package could completely break this feature, up to the point of crashing the app on any attempt to get track info, without any tests failing. The PR will need integration tests and/or native unit tests. (I would expect integration tests to be straightforward here since there's no UI interaction involved.)
- It's not clear that the iOS code is ready for review. [This code](https://github.com/flutter/packages/pull/9782/files#diff-ed00b25c9f04f87670b54e140ca5d91a0a1074c30f35e6f282256f7d1dfcd47cR573-R591) has comments describing logic that doesn't exist, and unused variable declarations that appear to be placeholders.
- The iOS code has comments saying that HLS needs to be handled differently and then appears to switch critical logic on an unreliable guess based on the URL; it's not clear why that guesswork is necessary vs actually inspecting the data you are looking for to see if it's there. If that's actually required, the code will need to explain why.
- The iOS code currently does all of the complex adaptation between the platform interface and the underlying native objects in Objective-C. I would strongly recommend doing most of that work in Dart instead:
  - It will be easier to unit test.
  - It will likely be easier to iterate on review feedback, since it looks like you are more familiar with Dart than Objective-C.
  - This package in particular is moving toward doing as much as possible in Dart rather than native code.
  
  You could define two Pigeon data classes that would closely mirror the underlying data from the two codepaths, make them both inherit from an empty `sealed` base class, and then return one of those from the Pigeon `getAudioTracks` depending on the case, then do all the transforms, switching, etc. to make it match the platform interface format in Dart.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -479,6 +479,284 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+  AVMediaSelectionGroup *audioGroup = nil;
+
+  // Check if this is an HLS stream and if we should use media selection options
+  if ([asset isKindOfClass:[AVURLAsset class]]) {
+    AVURLAsset *urlAsset = (AVURLAsset *)asset;
+    NSString *urlString = urlAsset.URL.absoluteString;
+
+    // Check if this is an HLS stream
+    if ([urlString containsString:@".m3u8"] ||
+        [urlString containsString:@"application/x-mpegURL"]) {
```
stuartmorgan-g: In what circumstances would a MIME type appear in a URL?

nateshmbhat:
@stuartmorgan-g have updated ios implementation and tests based on your feedback.

nateshmbhat:
i used claude for generating native tests as i'm not used to writing tests on native side. i see that there are still some build issues in test code. 
Let me know if this is a blocker for review. I was thinking be good to fix the test code after the rest of the source code gets reviewed.  
cc : @stuartmorgan-g

nateshmbhat:
@stuartmorgan-g instead of fetching the list of audio tracks from native android/ios, do you think it would be better to get the track metadata for hls directly from dart side by making a http get call and parsing the hls manifest ?

nateshmbhat:
@stuartmorgan-g 
any comment or suggestion for me as to where i need to focus on next ? i'm willing see it through to the end that this PR gets ready for merging.

ash2moon:
Hi @nateshmbhat! Thank you your efforts on this PR and appreciate the willingness to see the PR to get ready to be merged.

Looks like this PR has some merge conflicts and needs to be merged to the latest code on main. Also I noticed some of the linux failing tests.
For example:
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8706656615145803985/+/u/Run_package_tests/analyze/stdout?format=raw

Could you please take a look at these issues?

stuartmorgan-g:
> @stuartmorgan-g instead of fetching the list of audio tracks from native android/ios, do you think it would be better to get the track metadata for hls directly from dart side by making a http get call and parsing the hls manifest ?

Sorry, I missed this question. We should not re-fetch information that the native player has already fetched and parsed unless there is some extremely compelling reason to do so.

stuartmorgan-g:
FYI you can ignore the `Dart unit tests - pathified` failures in `Linux dart_unit_test_shard_1 *`; that's due to a repo tooling bug, and the failure will go away once the first sub-PR lands.

nateshmbhat:
i'm raising a separate PR which has both getAudioTracks and selectAudioTrack methods to get and select different tracks. will close this PR for now.

nateshmbhat:
#9925 is the NEW PR

nateshmbhat:
Closing this one as the new PR is here #9925

PR #9800, opened 2025-08-13
[video_player] Simplify native iOS code

stuartmorgan-g:
Refactors and simplifies some of the native code in `video_player_avfoundation`:
- Separates `create` into two different entry points for the two display types, instead of having one entry point and then branching several times based on the display type.
- Moves `dispose` to the player instance, with the other instance-specific calls, by making plugin-level tracking cleanup part of `onDispose`.
- Moves the URL->`AVPlayerItem` logic out of the two player classes and into a central helper in the plugin class, reducing duplication and simplifying the player initialization logic.
- Addresses the TODO to separate texture IDs from player IDs.
- Makes `expectFrame` a private implementation detail rather than making a client call it.
- Removes an ARC check; we have used ARC everywhere in this repo for a long time.

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
+++ b/packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
@@ -191,8 +191,9 @@ static void upgradeAudioSessionCategory(AVAudioSessionCategory requestedCategory
   }
 }
 
-- (nullable NSNumber *)createTexturePlayerWithOptions:(nonnull FVPCreationOptions *)options
-                                                error:(FlutterError **)error {
+- (nullable FVPTexturePlayerIds *)createTexturePlayerWithOptions:
+                                      (nonnull FVPCreationOptions *)options
+                                                           error:(FlutterError **)error {
   @try {
     AVPlayerItem *item = [self playerItemWithCreationOptions:options];
     FVPFrameUpdater *frameUpdater =


packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -186,55 +165,64 @@ - (void)initialize:(FlutterError *__autoreleasing *)error {
   upgradeAudioSessionCategory(AVAudioSessionCategoryPlayback, 0, 0);
 #endif
 
-  [self.playersByIdentifier.allValues makeObjectsPerformSelector:@selector(dispose)];
+  FlutterError *disposeError;
```
LongCatIsLooong, line 168: This local variable seems write-only?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
```diff
@@ -84,49 +79,39 @@ - (instancetype)initWithAVFactory:(id<FVPAVFactory>)avFactory
...
-  FVPTextureBasedVideoPlayer *textureBasedPlayer =
-      [player isKindOfClass:[FVPTextureBasedVideoPlayer class]]
-          ? (FVPTextureBasedVideoPlayer *)player
-          : nil;
-
-  int64_t playerIdentifier;
-  if (textureBasedPlayer) {
-    playerIdentifier = [self.registrar.textures registerTexture:textureBasedPlayer];
-    [textureBasedPlayer setTextureIdentifier:playerIdentifier];
-  } else {
```
LongCatIsLooong, line 115: Out of curiosity, is there a reason for generating `playerIdentifier` differently for texture based and platform view based players? Looks like it's trying to use texture identifier as player identifier?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
stuartmorgan-g: Yeah, Pigeon generates its methods with non-null error pointers since it should always be calling them with a value, so they don't accidentally get dropped.

The alternative here is that I could make a public `dispose` and make `disposeWithError:` call that, and then have these call points call `dispose`. I couldn't decide at the time if that was worth the extra method, so whichever you prefer.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
stuartmorgan-g: Historically, there was only the texture ID, and that was used throughout the plugin as the identifier for the player instance. When the platform view variant was added recently, that broke down; we still needed an identifier for each player, but not all players had texture IDs.

The PRs that added platform view support renamed all the platform interface parameters from `textureId` to `playerId` since that was simple, but to minimize churn in already-large PRs they didn't disentangle the texture ID from the player ID in the platform implementation internals. This is paying off that tech debt by switching to having a single way to manage player IDs, and reporting texture IDs separately in the texture case.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayerPlugin.m
LongCatIsLooong: Ah thanks for the explanation. I'm onboard with keeping the local variable then.

PR #9805, opened 2025-08-14
Add review agent style guidelines to .gemini/styleguide.md

stuartmorgan-g:
Initial attempt at adding guidance for GCAfGH to try cut down on the behavior observed so far that it:
- Almost always includes generic praise for the PR, which is unhelpful since the agent has no context for evaluating what PRs we actually want (see [this particularly problematic example](https://github.com/flutter/packages/pull/9795#pullrequestreview-3113026764))
- Appears to trust PR descriptions. E.g., recently praising a PR that was missing tests of most of the PR, but had a PR description saying it included extensive testing, as being extremely well tested.

This text is based on some iteration with Gemini, under the theory that the AI model has far more exposure to examples of good prompts than I do. 🤷🏻

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -21,3 +21,13 @@ auto-formatter, for each language, as described in
 - PR descriptions should include the Pre-Review Checklist from
   [the PR template](https://github.com/flutter/packages/blob/main/.github/PULL_REQUEST_TEMPLATE.md),
   with all of the steps completed.
+
+## Review Agent Guidelines
+
+When providing a summary, the review agent must adhere to the following principles:
+- **Be Objective:** Focus on a neutral, descriptive summary of the changes. Avoid subjective value judgments
+  like "good," "bad," "positive," or "negative." The goal is to report what the code does, not to evaluate it.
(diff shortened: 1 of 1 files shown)


.gemini/styleguide.md
```diff
@@ -21,3 +21,13 @@ auto-formatter, for each language, as described in
 - PR descriptions should include the Pre-Review Checklist from
   [the PR template](https://github.com/flutter/packages/blob/main/.github/PULL_REQUEST_TEMPLATE.md),
   with all of the steps completed.
+
+## Review Agent Guidelines
+
+When providing a summary, the review agent must adhere to the following principles:
+- **Be Objective:** Focus on a neutral, descriptive summary of the changes. Avoid subjective value judgments
+  like "good," "bad," "positive," or "negative." The goal is to report what the code does, not to evaluate it.
```
stuartmorgan-g, line 29: The "not to evaluate it" may end up swinging too far the other way; we'll have to see if it stops pointing out high-level issues (e.g., insufficient testing).

jtmcdole (approved):

PR #9813, opened 2025-08-14
[vector_graphics_compiler] Set the m4_10 (Z scale) value to 1 when constructing an AffineMatrix from an SVG matrix

jason-simmons:
The 4x4 matrix produced by the SVG matrix() function should have its 10th element set to 1. (see https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix)

The SVG matrix parser constructs an AffineMatrix from the SVG input.  The AffineMatrix is then converted into a list that is given to Flutter engine APIs.

Before https://github.com/flutter/engine/commit/62cb2c9ee7e88163344d38dd8e5ad3a260658d9f, the Flutter engine was implicitly overriding the 10th element of the matrix with a 1 value.  After that commit, the engine uses the matrix contents as is and does not override any values.

This PR changes the SVG matrix parser to explicitly set the AffineMatrix m4_10 parameter to 1.  This reflects the SVG spec and produces output that matches the original engine behavior.

See https://github.com/flutter/flutter/issues/171854


jason-simmons:
> LGTM.
> 
> Should we also file a bug against the package that other (implicit?) uses of `m4_10` and how it is maintained through various operations are likely suspect as well?

Your understanding of this is better than mine.  Can you file an issue in https://github.com/flutter/flutter/issues and tag it with `[vector_graphics]`?

jtmcdole (approved):

flar:
> Your understanding of this is better than mine. Can you file an issue in https://github.com/flutter/flutter/issues and tag it with `[vector_graphics]`?

https://github.com/flutter/flutter/issues/173817

PR #9815, opened 2025-08-15
[url_launcher] fix: Link widget Tab traversal

pedromassango:
Tries to fix the TAB traversal issue on web by ensuring the semantics of the `Link` are merged with its children. The buttons are treated as LINKS only and this prevents semantic conflicts that would create extra focusable DOM nodes.

With `MergeSemantics`, the semantics tree was not generated properly, e.g `linkUrl` value gets removed from the final semantics tree. This is being fixed in the engine, see https://github.com/flutter/flutter/pull/174473

Fixes https://github.com/flutter/flutter/issues/157689

Required engine fix PR:
https://github.com/flutter/flutter/pull/174473

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/url_launcher/url_launcher_web/lib/src/link.dart
```diff
@@ -121,6 +121,7 @@ class WebLinkDelegateState extends State<WebLinkDelegate> {
   Widget _buildChild(BuildContext context) {
     return Semantics(
       link: true,
+      excludeSemantics: true,
```
pedromassango: Note that this seems to drastically change the semantics tree, printing:

```dart
        print(tester
            .getSemantics(find.byType(WebLinkDelegate))
            .debugListChildrenInOrder(DebugSemanticsDumpOrder.traversalOrder));
```

Gives the following output:

### With `excludeSemantics: true`
```console
SemanticsNode#3
 │ Rect.fromLTRB(0.0, 0.0, 1200.0, 846.0)
 │ flags: scopesRoute
 │
 └─SemanticsNode#4
     Rect.fromLTRB(0.0, 0.0, 121.5, 32.0)
     flags: isLink
     identifier: "link-0"
```

Only one node gets focused when pressing tab, works as expected.

### With `excludeSemantics: false`
```console
SemanticsNode#3
 │ Rect.fromLTRB(0.0, 0.0, 1200.0, 846.0)
 │ flags: scopesRoute
 │
 └─SemanticsNode#4
   │ Rect.fromLTRB(0.0, 0.0, 121.5, 32.0)
   │ flags: isLink
   │ identifier: "link-0"
   │
   └─SemanticsNode#5
       Rect.fromLTRB(0.0, 0.0, 121.5, 32.0)
       actions: focus, tap
       flags: isButton, hasEnabledState, isEnabled, isFocusable
       label: "First Button"
       textDirection: ltr
```

We have two Nodes, when pressing tab, Flutter focus `SemanticsNode#4` and then `SemanticsNode#5`, Requiring double TAB to move the focus to the next focusable widget in the tree.

I ran out of ideas in how to fix this so feel free to propose a different approach

pedromassango: @ditman can you help review this PR?

mdebbar:
Thanks for tackling this issue!

Unfortunately, `excludeSemantics: true` won't work because it removes all other info provided by the child (e.g. label).

@chunhtai do you have suggestions on how to make this work? Maybe making the child `focusable: false` or something?

pedromassango:
> @chunhtai do you have suggestions on how to make this work? Maybe making the child `focusable: false` or something?

Isn't `focusable` false by default?

chunhtai:
I don't think the original use case makes sense. If their end goal is to create a link that looks like a button, they should style it themselves instead of wrapping button with a link. 

The original issue is like doing
```html
<a href="https://www.w3schools.com"><button type="button" onclick="alert('Hello world!')">Click Me!</button></a>
```

pedromassango:
> they should style it themselves instead of wrapping button with a link.

Can you give a POC showcasing how this is possible with the current implementation of the `Link package? You're missing the focus of the issue: when focusing a widget wrapped with the `Link` package, some widget deep within the `Link` widget gets focused as well, requiring double TAB.

All `Link` customers want is to be able to provide this native right-click behavior to any widget. See the image below where you can have the same behavior on a Button in GitHub.
Screenshot 2025-08-20 at 12 44 25 AM

chunhtai:
> All Link customers want is to be able to provide this native right-click behavior to any widget.

on I forgot it doesn't register ontap. yeah in this case, Link is not a full fledged `a` tag equivalent.

The problem is elevated button itself will create a semantics container. You mentioned mergesemantics doesn't work, do you know which part it is not working?

mdebbar:
> I don't think the original use case makes sense. If their end goal is to create a link that looks like a button, they should style it themselves instead of wrapping button with a link.
> 
> The original issue is like doing
> 
> ```
> <a href="https://www.w3schools.com"><button type="button" onclick="alert('Hello world!')">Click Me!</button></a>
> ```

This is actually exactly the use case that the `Link` widget was designed for 🙂

In the past, people did:
```dart
SomeButton(
  onPressed: () {
    launchUrl("https://www.w3schools.com");
  },
  child: Text("Click here"),
)
```
and complained that they didn't get a proper web link behavior (cmd+click, context menu, etc).

So we wanted people to do this instead:
```dart
Link(
  uri: Uri.parse("https://www.w3schools.com"),
  builder: (ctx, onPressed) => SomeButton(
    onPressed: onPressed,
    child: Text("Click here"),
  ),
)
```
which lets them use all existing buttons and widgets, and makes them behave like real web links.

chunhtai:
If you do this on the web
```
<a href="https://www.w3schools.com"><button type="button" onclick="alert('Hello world!')">Click Me!</button></a>
```

you will also need two tabs

pedromassango:
> You mentioned mergesemantics doesn't work, do you know which part it is not working?

Yes, the problem is that when you merge them, the semantics's Uri of the link gets lost.

It looks like the engine is removing it. Likely because the child of the Link widget has its semantics button and is is taking precedence over the link's semantics properties. Interesting enough, the Link's semantics `link: true` is keeped properly.

Will share the semantics tree diff here in a few hours

chunhtai:
I can think of several way.
1. use MergeSemantics and adjust web engine to let `link` take priority (bandage fix)
2. expose isSemanticsButton in elevatedButton, use Mergesemantics, and ask developer to set ElevateButton(isSemanticsButton: false)
3. Convert link:true to SemanticsRole (possibly button too but may be a bigger breaking change), in this case mergeSemantics will prioritize the role of parent widget

pedromassango:
> I can think of several way.
> 
> 1. use MergeSemantics and adjust web engine to let `link` take priority (bandage fix)
> 2. expose isSemanticsButton in elevatedButton, use Mergesemantics, and ask developer to set ElevateButton(isSemanticsButton: false)
> 3. Convert link:true to SemanticsRole (possibly button too but may be a bigger breaking change), in this case mergeSemantics will prioritize the role of parent widget

I like the first suggestion, tho I don't think it is a "bandage" fix. "Natively", buttons with a link (aka have a `href`) are treated as a link and even VoiceOver reads them as Link. Take GitHub for instance, any component with a href regradless of their appearance, is treated as link, as long as it have a `href`.

## Back button
Looks like a button but is treated like a link and VoiceOver reads it as link
Screenshot 2025-08-20 at 11 42 59 PM

## Pull requests button
Looks like a an icon button but is treated like a link and VoiceOver reads it as a link
Screenshot 2025-08-20 at 11 38 51 PM

## What looks like a primary button...
Is treated like a link as well, because it have a `href`

Screenshot 2025-08-20 at 11 51 45 PM


That said, I will modify this PR to fix it by modifying the engine

pedromassango:
@chunhtai any way you could help me with the fix in Flutter's framework/engine? Initially I though by adding the missing `_linkUrl ??= child._linkUrl;` in the `void absorb(...)` would ge me closer to the fix.

I would say, I'm now bakc to square zero as I ran out of ideas, would appreciate some hints on where to start

https://github.com/flutter/flutter/blob/9dd1ccb7010a7441b6d44fc5fc1be6842cfd1361/packages/flutter/lib/src/semantics/semantics.dart#L5994-L6007

chunhtai:
Where are you stuck? it does look like we missed the absorb for the link url.

pedromassango:
> Where are you stuck?

In your comment `use MergeSemantics and adjust web engine to let link take priority (bandage fix)` I can't seem to find the correct place to `... adjust web engine to let link take priority`

mdebbar:
> In your comment `use MergeSemantics and adjust web engine to let link take priority (bandage fix)` I can't seem to find the correct place to `... adjust web engine to let link take priority`

I believe that would be here:
https://github.com/flutter/flutter/blob/84f5d2625d9dfa2f3d5f79af06c7219120013743/engine/src/flutter/lib/web_ui/lib/src/engine/semantics/semantics.dart#L2058-L2065

pedromassango:
Hmm, I actualy did tried changing that already, couldn't notice any difference on the semantic tree 🤔 perhaps my testing env wasn't using the correct engine path... I will try again, thanks!

mdebbar:
@pedromassango I'm not sure how familiar you are with engine development, so I may be stating the obvious 🙂

After you make changes in engine sources, you have to rebuild the engine using `felt build`. And when you run your app, you have to use `--local-web-sdk=wasm_release` in order to pick the locally built engine.

pedromassango:
> After you make changes in engine sources, you have to rebuild the engine using `felt build`. And when you run your app, you have to use `--local-web-sdk=wasm_release` in order to pick the locally built engine.

Yeah, I totally missed this, let me try again.
Will also share a draft PR so you can have a look

btw, I am mostly familiar with the framework & pretty new to the engine

pedromassango:
Ok, this is ready for review, I still would love the Link widget to merge the semantics itself, if you allow for it. Reason being without this, the framework/engine fix is somewhat useless and we would still need to educate customers in how to "workaround the issue".

By letting `Link` merge semantics, we ensure there will always be only one semantic node.

pedromassango:
Finally have a draft PR for the engine: https://github.com/flutter/flutter/pull/174473. Pleas have a look!

Turns out I was missing the commands https://github.com/flutter/packages/pull/9815#issuecomment-3225176161 to very my fixes 🫠

packages/url_launcher/url_launcher_web/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## NEXT
```
chunhtai: 2.4.3

pedromassango: I've bumped already, see changes on lines 3-6

pedromassango: ah, to fix the CI... so you mean NEXT should be below the new version bump?

pedromassango:
@chunhtai any hints on why some tests are failing? 🤔 It is curious because:

1. The one I added works fine
2. The other tests are failing even after removing my changes which means this PR don't cause them to break 

Is it ok to merge with the unrelated failing tests?

stuartmorgan-g:
> Is it ok to merge with the unrelated failing tests?

No, absolutely not. If you believe there are unrelated failures, escalate to someone, don't just force-land your PR. If you are wrong, then you will break the tree, and if you are right, then you are effectively landing a PR into a broken tree for a purpose other than fixing the tree, which isn't allowed either.

stuartmorgan-g:
> 2\. The other tests are failing even after removing my changes which means this PR don't cause them to break

All presubmits are passing in the revert PR.

pedromassango:
> No, absolutely not. If you believe there are unrelated failures, escalate to someone, don't just force-land your PR. If you are wrong, then you will break the tree, and if you are right, then you are effectively landing a PR into a broken tree for a purpose other than fixing the tree, which isn't allowed either.

My bad, glad you reverted it.

I will give it another try in a separate PR, see https://github.com/flutter/packages/pull/10020

PR #9816, opened 2025-08-15
Update repo for 3.35 stable release

stuartmorgan-g:
Does the steps listed at https://github.com/flutter/flutter/blob/master/docs/ecosystem/release/Updating-Packages-repo-for-a-stable-release.md for the 3.35 release.

This makes the minimum supported version for all packages Dart 3.7, which means it picks up the new formatter. As a result, there are massive amounts of code changes; all of it is just the result of running `dart format` after changing the min SDK version.


stuartmorgan-g:
Version override: Per documented policy, we don't version for dropping SDK support, and autoformat changes should not affect package clients.
CHANGELOG override: CHANGELOGs were added for SDK changes as usual; the override is for packages that contained only autoformat changes.

Many packages do have README updates as a result of code excerpt format changes, but I don't think it's worth doing releases just to update code format in a README.

stuartmorgan-g:
I've split out https://github.com/flutter/packages/pull/9817 for the go_router test failure.

Given the need to do that, and that there's already been a collision, I think I'll split out some sub PRs that pre-update the min SDK of some groups of packages to make this easier to land.

PR #9822, opened 2025-08-15
[camera] Updates min SDK to 3.29

stuartmorgan-g:
For the `camera` packages:
- Updates the min SDK version to 3.29
- Runs the autoformatter with the new format
- Update code excerpts

This does not update versions because pushing format changes (even in READMEs) isn't worth doing a release that drops an SDK version (which we don't normally version)

Prep for https://github.com/flutter/packages/pull/9816

### DIFF
--- a/packages/camera/camera/CHANGELOG.md
+++ b/packages/camera/camera/CHANGELOG.md
@@ -1,3 +1,7 @@
+## NEXT
+
+* Updates minimum supported SDK version to Flutter 3.29/Dart 3.7.
+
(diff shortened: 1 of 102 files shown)


stuartmorgan-g:
> I've also left a couple of comments on duplicated lines in the changelog files that could be cleaned up.

Oops, forgot that step on this PR. Fixed. Thanks Gemini!

stuartmorgan-g:
Obsoleted by the combination PR.

PR #9847, opened 2025-08-18
[flutter_migrate] Remove source

stuartmorgan-g:
The package has been discontinued on pub.dev.

Context for removal:
- It was never fully launched (e.g., the version on pub didn't follow the standard release process).
- It has not been actively developed in more than a year, and has been subject to bitrot ([example](https://github.com/flutter/flutter/issues/121257)).
- There are currently no plans to re-staff the project in the foreseeable future in order to get it to a stable state.

Since it's a use-as-needed tool rather than something that applications actively depend on, and it has <100 downloads per month, we are not doing a deprecation period.


frederikstonge:
@stuartmorgan-g Am I allowed to take the source and maintain it? What are the rules that I have to follow? Keep the copyright on files and the license? Thank you

stuartmorgan-g:
> Am I allowed to take the source and maintain it?

Anyone can use the source in any way that the package's license allows.

> What are the rules that I have to follow? Keep the copyright on files and the license?

If you have questions about software licensing, you would need to consult a lawyer; the Flutter team cannot provide you with legal advice.

PR #9925, opened 2025-08-29
[video_player] Adds audio track metadata fetching and audio track selection feature 

nateshmbhat:
## Description

This PR adds comprehensive audio track retrieval and selection support to the video_player package, enabling developers to access detailed information about available audio tracks and switch between them during playback.

Breakout PRs: 
1. Platform interface : #10171 
2. Andrioid : #10312 
3. ios : #10313 

#### Core Features
- **Added `VideoAudioTrack` model** with comprehensive metadata fields: `id`, `label`, `language`, `isSelected`, `bitrate`, `sampleRate`, `channelCount`, `codec`
- **Added [getAudioTracks()]() method** to retrieve all available audio tracks with real metadata
- **Added [selectAudioTrack()]() method** to switch between audio tracks during playback
- **Updated `VideoPlayerController`** to expose audio track functionality

#### Platform Implementations
- **Android**: 
  - Real metadata extraction using ExoPlayer's `getCurrentTracks()` API
  - Robust track selection using `TrackSelectionOverride` with proper error handling
  - Support for multiple audio formats (AAC, AC3, EAC3, MP3, etc.)
- **iOS**: 
  - Metadata extraction from AVFoundation using `AVAssetTrack` for regular videos
  - HLS stream support using `AVMediaSelectionGroup` for adaptive streams
  - Proper track selection for both asset tracks and media selection options

#### Technical Infrastructure
- **Updated Pigeon interfaces** for both Android and iOS with new data structures:
  - `AudioTrackMessage`, `ExoPlayerAudioTrackData`, `AssetAudioTrackData`, `MediaSelectionAudioTrackData`, `NativeAudioTrackData`
- **Enhanced platform interface** with new methods and data classes
- **Has native unit tests** for both Android and iOS platforms
- **Created demo screen** showcasing audio track functionality with interactive UI

### Demo Features
- Interactive video player with audio track selection
- Real-time metadata display (bitrate, sample rate, channels, codec)
- Support for multiple video sources including HLS streams
- Visual indicators for currently selected tracks

## Related Issues
Fixes https://github.com/flutter/flutter/issues/59437

### Testing
- Added native unit tests for both Android and iOS
- Tested with various video formats and HLS streams
- Verified backward compatibility with existing functionality
- Demo screen widget for manual testing and validation

### Breaking Changes
None - all changes are additive and backward compatible.

## Pre-Review Checklist

- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[video_player]`


packages/video_player/video_player_avfoundation/lib/src/avfoundation_video_player.dart
```diff
@@ -212,6 +212,59 @@ class AVFoundationVideoPlayer extends VideoPlayerPlatform {
...
+    if (nativeData.mediaSelectionTracks != null) {
+      for (final MediaSelectionAudioTrackData track
+          in nativeData.mediaSelectionTracks!) {
+        final String trackId = 'media_selection_${track.index}';
+        final String label = track.commonMetadataTitle ?? track.displayName!;
+        tracks.add(
+          VideoAudioTrack(
+            id: trackId,
+            label: label,
+            language: track.languageCode!,
```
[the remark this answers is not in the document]
nateshmbhat: fixed

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -213,6 +213,38 @@ class AndroidVideoPlayer extends VideoPlayerPlatform {
...
+    final List<VideoAudioTrack> tracks = <VideoAudioTrack>[];
+
+    // Convert ExoPlayer tracks to VideoAudioTrack
+    if (nativeData.exoPlayerTracks != null) {
+      for (final ExoPlayerAudioTrackData track in nativeData.exoPlayerTracks!) {
+        tracks.add(
+          VideoAudioTrack(
+            id: track.trackId!,
+            label: track.label!,
+            language: track.language!,
```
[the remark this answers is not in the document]
nateshmbhat: fixed

packages/video_player/video_player_android/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.0-milestone-1-bin.zip
```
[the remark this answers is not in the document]
nateshmbhat: fixed

packages/video_player/video_player_avfoundation/darwin/RunnerTests/AudioTracksTests.m
```diff
@@ -0,0 +1,278 @@
...
+  FVPNativeAudioTrackData *result = [self.player getRawAudioTrackData:&error];
+
+  // Verify results - should handle nil option gracefully
+  XCTAssertNil(error);
+  XCTAssertNotNil(result);
+  XCTAssertNotNil(result.mediaSelectionTracks);
+  XCTAssertEqual(result.mediaSelectionTracks.count, 0);  // Should skip nil options
+}
+
+@end
```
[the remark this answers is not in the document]
nateshmbhat: fixed

packages/video_player/video_player_android/android/src/test/java/io/flutter/plugins/videoplayer/AudioTracksTest.java
```diff
@@ -0,0 +1,328 @@
...
+    // Verify results
+    assertNotNull(result);
+    assertEquals(numGroups, result.size());
+
+    // Should complete within reasonable time (1 second for 50 tracks)
+    assertTrue(
+        "getAudioTracks took too long: " + (endTime - startTime) + "ms",
+        (endTime - startTime) < 1000);
+  }
+}
```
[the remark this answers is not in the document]
nateshmbhat: fixed

nateshmbhat:
I updated native andorid tests and ran the tests and verified that all native android tests are passing.

nateshmbhat:
@stuartmorgan-g @ash2moon can you please comment on this PR and lemme know if there are any concerns ?

packages/video_player/video_player_avfoundation/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## NEXT
 
+* Implements `getAudioTracks()` and `selectAudioTrack()` methods for iOS/macOS using AVFoundation.
```
LongCatIsLooong: uber nit: the package is called `video_player_avfoundation` so I think the ios/macOS and AVFoundation part seems redundant.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+
+class _AudioTracksDemoState extends State<AudioTracksDemo> {
+  VideoPlayerController? _controller;
+  List<VideoAudioTrack> _audioTracks = <VideoAudioTrack>[];
+  bool _isLoading = false;
+  String? _error;
+
+  // Sample video URLs with multiple audio tracks
+  final List<String> _sampleVideos = <String>[
+    'https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4',
```
LongCatIsLooong: Would it be possible to use existing videos hosted in the flutter/samples repo (see the other demos), or do they not have tracks? I'm not sure what licenses these samples use.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+  Future<void> _initializeVideo() async {
+    setState(() {
+      _isLoading = true;
+      _error = null;
+    });
+
+    try {
+      await _controller?.dispose();
+
+      _controller = VideoPlayerController.networkUrl(
```
LongCatIsLooong: nit: here and everywhere (especially in  `_loadAudioTracks`): consider using a local var so you don't have to use `!` to force unwrap the controller every time.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+    if (_controller == null) {
+      return;
+    }
+
+    try {
+      await _controller!.selectAudioTrack(trackId);
+
+      // Add a small delay to allow ExoPlayer to process the track selection change
+      // This is needed because ExoPlayer's track selection update is asynchronous
+      await Future<void>.delayed(const Duration(milliseconds: 100));
```
LongCatIsLooong: Hmm this could be a major source of flakiness.  Shouldn't `selectAudioTrack` completes the future only after the new tracks becomes ready?

LongCatIsLooong: Also, a major source of user errors if this isn't documented (still I would prefer that we hide the ugliness within our implementation so the user doesn't have to worry about that).

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+
+      // Reload tracks to update selection status
+      await _loadAudioTracks();
+
+      if (!mounted) {
```
LongCatIsLooong, line 121: This check is also needed whenever you want to call `setState` I think?

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+          Padding(
+            padding: const EdgeInsets.all(16.0),
+            child: DropdownButtonFormField<int>(
+              value: _selectedVideoIndex,
+              decoration: const InputDecoration(
+                labelText: 'Select Video',
+                border: OutlineInputBorder(),
+              ),
+              items:
+                  _sampleVideos.asMap().entries.map((MapEntry<int, String> entry) {
```
LongCatIsLooong: nit: I'd use https://main-api.flutter.dev/flutter/dart-collection/IterableExtensions/indexed.html instead of `asMap()`. That would require `dart:collection` but iirc it's already a transitive dependency of the material library.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
LongCatIsLooong: This can be a `const` right?

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+      appBar: AppBar(
+        title: const Text('Audio Tracks Demo'),
+        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
+      ),
+      body: Column(
+        children: <Widget>[
+          // Video selection dropdown
+          Padding(
+            padding: const EdgeInsets.all(16.0),
+            child: DropdownButtonFormField<int>(
```
LongCatIsLooong: Consider using https://api.flutter.dev/flutter/material/DropdownMenuFormField-class.html instead, IIRC the Dropdown* widgets are going to be replaced.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+              ),
+              items:
+                  _sampleVideos.asMap().entries.map((MapEntry<int, String> entry) {
+                    return DropdownMenuItem<int>(
+                      value: entry.key,
+                      child: Text('Video ${entry.key + 1}'),
+                    );
+                  }).toList(),
+              onChanged: (int? value) {
+                if (value != null && value != _selectedVideoIndex) {
```
LongCatIsLooong, line 199: I read the material DropdownMenu documentation and it doesn't really tell me when the callback would give me a  `null` value. Does it mean no selection was made?

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+      ),
+    );
+  }
+
+  Widget _buildVideoPlayer() {
+    if (_isLoading) {
+      return const Center(child: CircularProgressIndicator());
+    }
+
+    if (_error != null) {
```
LongCatIsLooong: nit: `final error case _error?`

nateshmbhat:
any update on this PR ?

stuartmorgan-g:
From triage: ping @tarrinneal @ash2moon @LongCatIsLooong @hellohuanlin for review

@nateshmbhat Apologies for the delay; things have been unusually busy recently. A couple of high-level notes while you are waiting for full reviews:
- The `video_player_web` changes can all be reverted; it looks like you probably ran the auto-formatter in the package without running `pub get`, so applied the wrong formatter version.
- It appears that one of the new tests is crashing on macOS.
- If this feature isn't going to be implemented on all platforms, it needs to have [a support query](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#api-support-queries) so clients know if they can call it.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+      child: IconButton(
+        iconSize: 48,
+        color: Colors.white,
+        onPressed: () {
+          if (_controller!.value.isPlaying) {
+            _controller!.pause();
+          } else {
+            _controller!.play();
+          }
+          setState(() {});
```
LongCatIsLooong: why is this setState needed here? Can we listen to the controller?

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,326 @@
...
+              Text('Sample Rate: ${track.sampleRate} Hz'),
+            if (track.channelCount != null)
+              Text('Channels: ${track.channelCount}'),
+          ],
+        ),
+        trailing:
+            track.isSelected
+                ? const Icon(Icons.radio_button_checked, color: Colors.green)
+                : const Icon(Icons.radio_button_unchecked),
+        onTap: track.isSelected ? null : () => _selectAudioTrack(track.id),
```
LongCatIsLooong, line 375: The ListTile is going to look like it's disabled when the track is selected. Is that intended?

nateshmbhat:
> From triage: ping @tarrinneal @ash2moon @LongCatIsLooong @hellohuanlin for review

Pining @tarrinneal @ash2moon @LongCatIsLooong @hellohuanlin for review

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -466,6 +466,216 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+      currentSelection = [currentItem selectedMediaOptionInMediaSelectionGroup:audioGroup];
+#pragma clang diagnostic pop
+    }
+
+    for (NSInteger i = 0; i < audioGroup.options.count; i++) {
+      AVMediaSelectionOption *option = audioGroup.options[i];
+
+      // Skip nil options
+      if (!option || [option isKindOfClass:[NSNull class]]) {
```
LongCatIsLooong: nit: the first `nil` check doesn't seem to be needed?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -466,6 +466,216 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+        continue;
+      }
+
+      NSString *displayName = option.displayName;
+      if (!displayName || displayName.length == 0) {
+        displayName = [NSString stringWithFormat:@"Audio Track %ld", (long)(i + 1)];
```
LongCatIsLooong: Why does the implementation have to provide a fallback name?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -466,6 +466,216 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+      }
+
+      NSString *languageCode = @"und";
+      if (option.locale) {
+        languageCode = option.locale.languageCode ?: @"und";
```
LongCatIsLooong: nit: `NSString* languageCode = option.locale.languageCode ?: @"und";`, without the if?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -466,6 +466,216 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+      }
+
+      NSString *commonMetadataTitle = nil;
+      for (AVMetadataItem *item in option.commonMetadata) {
```
LongCatIsLooong: the API documentation recommends filtering using a different method:  https://developer.apple.com/documentation/avfoundation/avmediaselectionoption/commonmetadata?language=objc#Discussion

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -466,6 +466,216 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+
+    // Extract format information
+    NSNumber *bitrate = nil;
+    NSNumber *sampleRate = nil;
+    NSNumber *channelCount = nil;
+    NSString *codec = nil;
+
+    // Only attempt format description parsing in production (non-test) environments
+    // Skip entirely if we detect any mock objects or test environment indicators
+    NSString *trackClassName = NSStringFromClass([track class]);
```
LongCatIsLooong: Why is the behavior different in production and tests? I think that defeats the purpose of testing if the logic is different for tests?

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: > Would it be possible to use existing videos hosted in the flutter/samples repo (see the other demos), or do they not have tracks? I'm not sure what licenses these samples use.

I tried to find if such urls are there in flutter/samples but coudn't find it.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: will add it after async calls whereever it's needed

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: will move the delay inside the selectAudioTrack dart method itself. that should take care of this.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: ok

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: yes we can , will update

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: the tile is looking the same whether i pass null or empty function. so, should be fine.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: didn't understand this.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
nateshmbhat: will update

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
nateshmbhat: will update

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
nateshmbhat: The fallback name "Audio Track %ld" is provided because AVMediaSelectionOption's displayName property can be nil or empty in real-world scenarios. Even when not nil, it can be an empty string "".

Without a fallback, users would see:
Empty labels in audio track selection UI
Confusing blank entries
No way to distinguish between tracks

nateshmbhat:
@LongCatIsLooong  Have updated code based on review comments and added api support query check to return `false` in web as suggested by @stuartmorgan-g

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
stuartmorgan-g: > Without a fallback, users would see:
> Empty labels in audio track selection UI
> Confusing blank entries
> No way to distinguish between tracks

`video_player` doesn't show any UI; everything you are describing would happen only if the plugin client chose to not provide fallback when displaying the underlying data to the user.

Putting a fallback here makes it impossible for the client to pick their own fallback, localize the fallback, etc.

nateshmbhat: I understand, what do you suggest I do here? Remove the "Audio Track" prefix?

stuartmorgan-g: I would suggest returning whatever the underlying SDK returns.

nateshmbhat: done

stuartmorgan-g:
The new files here will require a minor change to the license header when you merge in `main`; you can re-copy from an existing file post-merge.

nateshmbhat:
> The new files here will require a minor change to the license header when you merge in `main`; you can re-copy from an existing file post-merge.

i'm not understanding this, can u please elaborate ?

stuartmorgan-g:
> i'm not understanding this, can u please elaborate ?

[This line](https://github.com/flutter/packages/pull/9925/files#diff-123d5ef4afc3d73b1b493f2b8c26b619394d8612acf5e3dff5b7b19717e42647R1) will fail the license check after your next merge from `main`, because the repo has been updated to longer include "All rights reserved". Once you merge, replace that line with a copy from any other existing (post-merge) file.

nateshmbhat:
> [This line](https://github.com/flutter/packages/pull/9925/files#diff-123d5ef4afc3d73b1b493f2b8c26b619394d8612acf5e3dff5b7b19717e42647R1) will fail the license check after your next merge from `main`, because the repo has been updated to longer include "All rights reserved". Once you merge, replace that line with a copy from any other existing (post-merge) file.

Pushed the change. pls check @stuartmorgan-g

nateshmbhat:
any update on this PR ? what should i do next to get this merged ?

stuartmorgan-g:
> Pushed the change. pls check @stuartmorgan-g

Our CI checks the license conformance.

> any update on this PR ? what should i do next to get this merged ?

It still needs review and approval from @tarrinneal, @ash2moon, and @LongCatIsLooong

nateshmbhat:
@tarrinneal, @ash2moon, @LongCatIsLooong
please help with the review

stuartmorgan-g:
@nateshmbhat Our triage process will follow up with people regularly; pinging everyone two days after I did reminder pings is not productive.

nateshmbhat:
> @nateshmbhat Our triage process will follow up with people regularly; pinging everyone two days after I did reminder pings is not productive.

Got it, thanks for clarifying!

packages/video_player/video_player/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## NEXT
```
tarrinneal: Assuming no breaking changes (This is the first thing I'm reading), this should be 2.11.0

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -819,6 +819,65 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+  Future<void> selectAudioTrack(String trackId) async {
+    if (_isDisposedOrNotInitialized) {
+      throw Exception('VideoPlayerController is disposed or not initialized');
+    }
+    await _videoPlayerPlatform.selectAudioTrack(_playerId, trackId);
+
+    if (!kIsWeb && Platform.isAndroid) {
+      // Add a small delay to allow ExoPlayer to process the track selection change
+      // This is needed because ExoPlayer's track selection update is asynchronous
+      await Future<void>.delayed(const Duration(milliseconds: 100));
```
tarrinneal: This scares me. Is this number arbitrary? Could it be less? Is it enough to work consistently?

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -819,6 +819,65 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+  ///
+  /// Example usage:
+  /// ```dart
+  /// if (await controller.isAudioTrackSupportAvailable()) {
+  ///   final tracks = await controller.getAudioTracks();
+  ///   // Show audio track selection UI
+  /// } else {
+  ///   // Hide audio track selection UI or show unsupported message
+  /// }
+  /// ```
```
tarrinneal, line 986: I don't know that this example is necessary, but removing it isn't necessary.

packages/video_player/video_player_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## NEXT
```
tarrinneal: 2.9.0

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -120,6 +136,91 @@ public ExoPlayer getExoPlayer() {
     return exoPlayer;
   }
 
+  @UnstableApi
```
tarrinneal, line 147: What's the purpose of labeling these unstable

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -120,6 +136,91 @@ public ExoPlayer getExoPlayer() {
...
+      // Get the track group and create a selection override
+      TrackGroup trackGroup = group.getMediaTrackGroup();
+      TrackSelectionOverride override = new TrackSelectionOverride(trackGroup, trackIndex);
+
+      // Apply the track selection override
+      trackSelector.setParameters(
+          trackSelector.buildUponParameters().setOverrideForType(override).build());
+
+    } catch (NumberFormatException | ArrayIndexOutOfBoundsException e) {
+      // Invalid trackId format, ignore
```
tarrinneal: Is there a reason not to allow the error to be handled by the user here?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/platformview/PlatformViewVideoPlayer.java
```diff
@@ -22,6 +23,7 @@
  * displaying the video in the app.
  */
 public class PlatformViewVideoPlayer extends VideoPlayer {
+  @UnstableApi
```
tarrinneal, line 27: Why?

packages/video_player/video_player_avfoundation/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## NEXT
```
tarrinneal: 2.9.0

packages/video_player/video_player_platform_interface/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## NEXT
```
tarrinneal, line 1: 6.5.0

packages/video_player/video_player_web/lib/video_player_web.dart
```diff
@@ -170,4 +170,22 @@ class VideoPlayerPlugin extends VideoPlayerPlatform {
   /// Sets the audio mode to mix with other sources (ignored).
   @override
   Future<void> setMixWithOthers(bool mixWithOthers) => Future<void>.value();
+
+  @override
+  Future<List<VideoAudioTrack>> getAudioTracks(int playerId) async {
+    // Web platform does not support audio track selection
```
tarrinneal: Is this a limitation of web, or just due to it not being implemented?

tarrinneal (commented):
A few nits and questions. I think a break off pr for the interface is good to go from here.

packages/video_player/video_player_web/lib/video_player_web.dart
nateshmbhat: haven't explored the web apis for this. for now, i mean that its not supported for web yet. will update the comment

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: i was getting compilation issues without this.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: here's a brief : 

1. Defensive Programming for Invalid Input
The method expects a specific trackId format ("groupIndex_trackIndex"), and the exceptions occur when:

NumberFormatException: The trackId contains non-numeric parts
ArrayIndexOutOfBoundsException: The parsed indices are out of bounds

2. Graceful Degradation
The method already has multiple validation layers that return early for invalid conditions:

Checks if trackSelector is null
Validates trackId format (must have exactly 2 parts)
Validates group index bounds
Validates track type and track index bounds
The catch block serves as a final safety net for edge cases that might slip through.

3. Platform Interface Contract
since selectAudioTrack returns Future<void>, not Future<bool> or a result type. The API is designed to be "fire and forget" - either it works or it doesn't, but it shouldn't crash the app.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: also, there's already some methods in existing video_player_android code which are using this. And there's a related issue here : https://github.com/flutter/flutter/issues/147039

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: `Could it be less` : probably yes
`Is it enough to work consistently` : Based on my thorough testing on different devices both low end and high end phones, i didn't face any issue even once and this delay was enough.

nateshmbhat:
> A few nits and questions. I think a break off pr for the interface is good to go from here.

raised the break off pr for the interface here :  https://github.com/flutter/packages/pull/10171

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,379 @@
...
+    if (_isLoading) {
+      return const Center(child: CircularProgressIndicator());
+    }
+
+    if (_error != null) {
+      return Center(
+        child: Column(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: <Widget>[
+            Icon(Icons.error, size: 48, color: Colors.red[300]),
```
ash2moon, line 237: nit: Can we use the theme based colors throughout this example? Including where we use Colors.white etc?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -120,6 +136,91 @@ public ExoPlayer getExoPlayer() {
...
+  public void selectAudioTrack(@NonNull String trackId) {
+    if (trackSelector == null) {
+      return;
+    }
+
+    try {
+      // Parse the trackId (format: "groupIndex_trackIndex")
+      String[] parts = trackId.split("_");
+      if (parts.length != 2) {
+        return;
```
ash2moon: Can we at least log a warning on these early returns to let the developer know why the tracks could not be slelected?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
ash2moon: What part exactly is unstable? Is there a way to do this without making them unstable? I don't feel comfortable adding in more unstable API usage if we can avoid them.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -120,6 +136,91 @@ public ExoPlayer getExoPlayer() {
...
+      int trackIndex = Integer.parseInt(parts[1]);
+
+      // Get current tracks
+      Tracks tracks = exoPlayer.getCurrentTracks();
+
+      if (groupIndex >= tracks.getGroups().size()) {
+        return;
+      }
+
+      Tracks.Group group = tracks.getGroups().get(groupIndex);
```
ash2moon: Let's move all the encoding/decoding logic into separate methods. It would be better if we could just publish the group and track index to the dart side so there's no need to do this.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
ash2moon: Can we at least log in this event?

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -819,6 +819,65 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+    if (_isDisposed) {
+      throw Exception('VideoPlayerController is disposed');
+    }
+    if (!value.isInitialized) {
+      return <VideoAudioTrack>[];
+    }
+    return _videoPlayerPlatform.getAudioTracks(_playerId);
+  }
+
+  /// Selects an audio track by its ID.
```
ash2moon: Can you clarify what "selects" mean here? According to the exoplayer documentation it is:

"When a media item contains multiple tracks, track selection is the process that determines which of them are chosen for playback" https://developer.android.com/media/media3/exoplayer/track-selection

ash2moon (changes requested):

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: Can you suggest an appropriate doc comment? I'm not able to figure one out based on that exoplayer description

ezamagni-tg: I'll give a try:
"Selects which audio track is chosen for playback by indicating its [trackId]"

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: done

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: updated

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: can u elaborate more on this. also, is this a must have ?

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -16,6 +15,7 @@ export 'package:video_player_platform_interface/video_player_platform_interface.
     show
         DataSourceType,
         DurationRange,
+        VideoAudioTrack,
```
stuartmorgan-g: Related to some comments I left in the interface PR: this shouldn't be re-exported; instead there should be a similar type defined at this layer, and there should be an internal method that converts from the platform interface type to the public type.

Forcing the platform interface types and the public types to be the same has bitten us in the past, and so we try to avoid it for any new non-trivial types.

nateshmbhat: updated

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
ash2moon: It seems that it would be unnecessary to have to have a parser and then also encode two values into a string just to read back the values. @mboetger for second pair of eyes

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
LongCatIsLooong: Example: 
https://github.com/flutter/packages/blob/169d55502eca9ca5d72da960cd23e20bbcb8a20e/packages/video_player/video_player_avfoundation/example/lib/main.dart#L183-L185

LongCatIsLooong: Or you can upload one with a proper license to https://github.com/flutter/assets-for-api-docs/tree/main/assets/videos

nateshmbhat:
Any update on this PR needed from my side?

stuartmorgan-g:
@nateshmbhat You can split out a new PR with the platform implementations (and with the overrides removed in favor of requiring the newly published interface package version), and then the rest of the platform implementation reviews can happen there.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,379 @@
...
+
+      final VideoPlayerController controller = VideoPlayerController.networkUrl(
+        Uri.parse(_sampleVideos[_selectedVideoIndex]),
+      );
+      _controller = controller;
+
+      await controller.initialize();
+
+      // Add listener for video player state changes
+      _controller!.addListener(_onVideoPlayerValueChanged);
```
LongCatIsLooong: nit: here and below, use the local controller to avoid `!`.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
```diff
@@ -0,0 +1,379 @@
...
+        return;
+      }
+      ScaffoldMessenger.of(context).showSnackBar(
+        SnackBar(content: Text('Failed to select audio track: $e')),
+      );
+    }
+  }
+
+  void _onVideoPlayerValueChanged() {
+    if (!mounted || _controller == null) {
```
LongCatIsLooong: uber nit: the mounted check is not needed if this method is unsubscribed at unmount.

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
LongCatIsLooong:
```suggestion
    if (_error case final String error?) {
```
so you don't need to use `!`.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -24,6 +25,112 @@ export 'package:video_player_platform_interface/video_player_platform_interface.
...
+///
+/// Normalizes null values from the platform to provide a consistent API:
+/// - null label becomes 'Unknown'
+/// - null language becomes 'und' (undefined)
+VideoAudioTrack _convertPlatformAudioTrack(
+  platform_interface.VideoAudioTrack platformTrack,
+) {
+  return VideoAudioTrack(
+    id: platformTrack.id,
+    label: platformTrack.label ?? 'Unknown',
```
LongCatIsLooong: nit: consider not using these hard coded default values for localization. (same for `language`).

packages/video_player/video_player/lib/video_player.dart
LongCatIsLooong: This is still scary even with thorough testing. A future version of exoPlayer can break this hardcoded delay timing (in case the selection takes longer to complete in that version).  

I'm not familiar with the API of exoPlayer, but not having an API to tell whether the audio track selection is completed just feels a bit strange.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -466,6 +466,221 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+
+      // Skip nil options
+      if (!option || [option isKindOfClass:[NSNull class]]) {
+        continue;
+      }
+
+      NSString *displayName = option.displayName;
+
+      NSString *languageCode = nil;
+      if (option.locale) {
```
LongCatIsLooong, line 470: nit: the if doesn't seem to be necessary?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -466,6 +466,221 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
   [self updatePlayingState];
 }
 
+- (nullable FVPNativeAudioTrackData *)getAudioTracks:(FlutterError *_Nullable *_Nonnull)error {
+  AVPlayerItem *currentItem = _player.currentItem;
+  if (!currentItem || !currentItem.asset) {
+    return [FVPNativeAudioTrackData makeWithAssetTracks:nil mediaSelectionTracks:nil];
+  }
+
+  AVAsset *asset = currentItem.asset;
```
LongCatIsLooong, line 451: The documentation says this returns an empty array when tracks haven't been loaded yet. Can this method be called when the tracks are not yet loaded?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -466,6 +466,221 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+        id formatDescObj = track.formatDescriptions[0];
+
+        // Validate that we have a valid format description object
+        if (formatDescObj && [formatDescObj respondsToSelector:@selector(self)]) {
+          NSString *className = NSStringFromClass([formatDescObj class]);
+
+          // Only process objects that are clearly Core Media format descriptions
+          // This works for both real CMFormatDescription objects and properly configured mock
+          // objects
+          if ([className hasPrefix:@"CMAudioFormatDescription"] ||
```
LongCatIsLooong: I don't think we should include logic that pertains to testing / mocking.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -466,6 +466,221 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+
+  AVAsset *asset = currentItem.asset;
+
+  // Check if this is a media selection track (for HLS streams)
+  if ([trackId hasPrefix:@"media_selection_"]) {
+    AVMediaSelectionGroup *audioGroup =
+        [asset mediaSelectionGroupForMediaCharacteristic:AVMediaCharacteristicAudible];
+    if (audioGroup && audioGroup.options.count > 0) {
+      // Parse the track ID to get the index
+      NSString *indexString = [trackId substringFromIndex:[@"media_selection_" length]];
```
LongCatIsLooong: This looks very fragile. Could you find an alternative to map trackID to the track?

nateshmbhat:
> @nateshmbhat You can split out a new PR with the platform implementations (and with the overrides removed in favor of requiring the newly published interface package version), and then the rest of the platform implementation reviews can happen there.

@stuartmorgan-g Raised #10313 and #10312

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
nateshmbhat: updated with different approach, pls check

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
nateshmbhat: updated

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
nateshmbhat: yes,

When getAudioTracks is called before the async loading completes:

- tracksWithMediaType:AVMediaTypeAudio (line 544) returns an empty array per Apple's documentation
- mediaSelectionGroupForMediaCharacteristic: (line 483-484) also returns nil or an empty group
- The method gracefully returns an empty track list via the early return at line 477 or empty arrays

nateshmbhat:
@LongCatIsLooong have updated the waiting for exoplayer to select the track approach (as per this comment https://github.com/flutter/packages/pull/9925#discussion_r2466556034 )

please check

nateshmbhat:
Have addressed all the major concerns in this PR , can we please have the platform PRs merged ?  If any furthur changes are needed, pls let me know at the earliest so that i can quickly push the changes and we can take this forward sooner. 

cc: @stuartmorgan-g @ash2moon @LongCatIsLooong

stuartmorgan-g:
> Have addressed all the major concerns in this PR , can we please have the platform PRs merged ? If any furthur changes are needed, pls let me know at the earliest so that i can quickly push the changes and we can take this forward sooner.

I'm not sure why you are pinging people here. This PR can't land until the sub-PRs have landed, and the sub PRs are still under review. Those PRs will be landed when they have been reviewed and approved.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -819,6 +926,63 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+  /// Gets the available audio tracks for the video.
+  ///
+  /// Returns a list of [VideoAudioTrack] objects containing metadata about
+  /// each available audio track. The list may be empty if no audio tracks
+  /// are available or if the video is not initialized.
+  ///
+  /// Throws an exception if the video player is disposed.
+  Future<List<VideoAudioTrack>> getAudioTracks() async {
+    if (_isDisposed) {
+      throw Exception('VideoPlayerController is disposed');
```
nateshmbhat: should we be throwing an exception if the video is already disposed ? or should be just return like how its just returned in many of the other methods ?

camsim99:
@nateshmbhat I'll take a look at the Android changes again once you port them over from the split out PR. Thanks!

nateshmbhat:
@camsim99 i have ported the android changes.

stuartmorgan-g:
That's presumably a merge mistake; there shouldn't be Android changes in this PR now that the Android sub-PR has landed. The Android changes here should be reverted.

Also, the web changes can be reverted since the overrides don't do anything meaningfully different than the base class implementation that will be inherited.

gmackall:
> That's presumably a merge mistake; there shouldn't be Android changes in this PR now that the Android sub-PR has landed. The Android changes here should be reverted.
> 
> Also, the web changes can be reverted since the overrides don't do anything meaningfully different than the base class implementation that will be inherited.

Removing triage-android label based on this

nateshmbhat:
native PRs are merged. I have removed the dependency overrides with the new version of the packages. anything else needed from my side ?

packages/video_player/video_player/CHANGELOG.md
nateshmbhat: done

isra60:
Is there any plans to support also subtitle (text) selection? And maybe styles of them?

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
stuartmorgan-g, line 1: There are a number of unaddressed comments in this file.

packages/video_player/video_player/lib/video_player.dart
stuartmorgan-g: This comment has not been addressed. I don't think this is just a nit; the fact that this is making up placeholder values rather than returning null when nothing is available would be a breaking change to fix if shipped this way.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -819,6 +927,63 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
```
stuartmorgan-g: Why is this an Exception rather than a StateError?

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -819,6 +927,63 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+  /// Selects which audio track is chosen for playback from its [trackId]
+  ///
+  /// The [trackId] should match the ID of one of the tracks returned by
+  /// [getAudioTracks]. If the track ID is not found or invalid, the
+  /// platform may ignore the request or throw an exception.
+  ///
+  /// Throws an exception if the video player is disposed or not initialized.
+  Future<void> selectAudioTrack(String trackId) async {
+    if (_isDisposedOrNotInitialized) {
+      throw Exception('VideoPlayerController is disposed or not initialized');
```
stuartmorgan-g: Same.

packages/video_player/video_player_web/example/pubspec.yaml
stuartmorgan-g, line 1: Why are there dependency changes to the `video_player_web` package when there is no new code requiring updated dependencies?

nateshmbhat: will revert it.

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: updated

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: updated

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: addressed them

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: updated

packages/video_player/video_player/example/lib/audio_tracks_demo.dart
nateshmbhat: The DropdownMenu.onSelected callback receives null when the menu is dismissed without making a selection (e.g., tapping outside the menu). Added a comment to clarify this behavior.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -10,6 +10,8 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:video_player_platform_interface/video_player_platform_interface.dart';
+import 'package:video_player_platform_interface/video_player_platform_interface.dart'
```
stuartmorgan-g, line 12: There should not be two imports of the same file with different prefixes. If we need prefixing, the other import needs to be removed.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -25,6 +27,110 @@ export 'package:video_player_platform_interface/video_player_platform_interface.
...
+    this.sampleRate,
+    this.channelCount,
+    this.codec,
+  });
+
+  /// Unique identifier for the audio track.
+  final String id;
+
+  /// Human-readable label for the track.
+  /// May be null if not available from the platform.
```
stuartmorgan-g, line 49: This needs a blank comment line above it, per Effective Dart style. Applies throughout this class.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -819,6 +925,63 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
     }
   }
 
+  /// Gets the available audio tracks for the video.
+  ///
+  /// Returns a list of [VideoAudioTrack] objects containing metadata about
+  /// each available audio track. The list may be empty if no audio tracks
+  /// are available or if the video is not initialized.
+  ///
+  /// Throws an exception if the video player is disposed.
```
stuartmorgan-g: It's an error, not an exception.

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: updated

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: updated

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: i think option here is to either replace every usage of all the classes that are used from platform interface with platform_interface. prefix or to rename the VideoAudioTrack to something else in video_player.dart to avoid the naming conflict. what would you prefer ? here's a summary of options we have.


## Options for handling duplicate imports

### Option 1: Keep both imports (current approach)
```dart
import 'package:video_player_platform_interface/video_player_platform_interface.dart';
import 'package:video_player_platform_interface/video_player_platform_interface.dart'
    as platform_interface;
```

Pros: Minimal code changes - use unprefixed types everywhere except where disambiguation is needed (platform_interface.VideoAudioTrack)
Cons: Two imports of the same file looks redundant
Option 2: Single prefixed import with show clause

```dart
import 'package:video_player_platform_interface/video_player_platform_interface.dart'
    as platform_interface show VideoAudioTrack, DataSource, ...;
```


Option 3: Rename the local VideoAudioTrack class
Rename to something like VideoAudioTrackInfo or AudioTrack to avoid the naming collision.

Pros: No duplicate imports; cleaner code
Cons: API change





Option 4: Use hide + prefixed show
dart
import 'package:video_player_platform_interface/video_player_platform_interface.dart'
    hide VideoAudioTrack;
import 'package:video_player_platform_interface/video_player_platform_interface.dart'
    as platform_interface show VideoAudioTrack;

stuartmorgan-g: > Option 4: Use hide + prefixed show
> dart
> import 'package:video_player_platform_interface/video_player_platform_interface.dart'
> hide VideoAudioTrack;
> import 'package:video_player_platform_interface/video_player_platform_interface.dart'
> as platform_interface show VideoAudioTrack;

If switching everything to using the prefix (I'm not sure why that wasn't any of the four options) is too verbose, this is the correct way to handle it.

nateshmbhat: updated

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -9,7 +9,8 @@ import 'dart:math' as math show max;
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
-import 'package:video_player_platform_interface/video_player_platform_interface.dart';
+import 'package:video_player_platform_interface/video_player_platform_interface.dart'
+    as platform_interface;
```
tarrinneal, line 13: Maybe I missed it, but is there a reason for this change? Is there a naming collision?

nateshmbhat: Yes, there's a naming collision for the audio track model class.

nateshmbhat: Discussed different options to handle this with Stuart and this is the agreed upon approach.

nateshmbhat:
> autosubmit label was removed for flutter/packages/9925, because - The status or check suite [Mac_arm64 macos_platform_tests master - packages](https://cr-buildbucket.appspot.com/build/8690242530619916353) has failed. Please fix the issues identified (or deflake) before re-applying this label.
> 

Not sure I understand what's the build timeout error. Do I need to do something about this? Can someone guide me?

stuartmorgan-g:
I thought it was flake since this PR doesn't change any integration tests, but it happened again on a re-run. Does this test pass for you locally?

nateshmbhat:
need some help with the CI checks. can someone help ?

nateshmbhat:
These tests are passing locally for me. 🤔 

The failing checks are all pre-existing/flaky and unrelated to this PR's  changes:
                                                    
  1. Linux repo_checks & Windows_x64 repo_tools_tests
          
  Both fail with the same trailing \n\n\n newline mismatch in
  build_examples_command_test.dart and analyze_command_test.dart. These tests
  check pubspec restoration after SwiftPM flag toggling via YamlEditor (in
  https://github.com/flutter/packages/blob/main/script/tool/lib/src/common/plugin_utils.dart#L98-L144).

  This is a pre-existing failure on main:
  - https://github.com/flutter/packages/runs/63643572548

  This PR did not touch any of the code involved in these failures. The only
  changes to analyze_command.dart / analyze_command_test.dart were https://githu
  b.com/flutter/packages/pull/9925/files#diff-d5e0a5c0c5e8a2b8e8e5e5e5e5e5e5e5 —
   build_examples_command.dart, build_examples_command_test.dart, and
  plugin_utils.dart were not modified at all.

  - PR failing log (Linux): https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8689746820334650065/+/u/Run_package_tests/tool_unit_tests/stdout
  - PR failing log (Windows):
  https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8689746820133950641/+/u/Run_package_tests/tool_unit_tests/stdout

  2. Mac_arm64 macos_platform_tests

  The integration test "network videos reports buffering status" timed out. This
   test is already known to be flaky and is skipped on other platforms for the
  same hanging behavior:
  - Skipped on Web: https://github.com/flutter/flutter/issues/169219
  - Skipped on Android: https://github.com/flutter/flutter/issues/160797

  The test waits for buffering events that may never arrive in CI network
  conditions. This PR didn't modify the integration test or the buffering logic.

  - PR failing log (macOS): https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8689746820127214769/+/u/Run_package_tests/drive_examples/stdout
  - Rerun also timed out: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8690242530619916353/+/u/Run_package_tests/drive_examples/stdout

  Could we get a re-run on the macOS test, or should I file a separate issue to
  skip/deflake it on macOS as well?

camsim99:
Updating your branch to pick up https://github.com/flutter/packages/pull/11014

PR #9980, opened 2025-09-09
Revert "[url_launcher] fix: Link widget Tab traversal"

stuartmorgan-g:
Reverts flutter/packages#9815

The tree is now broken in exactly the way the presubmits for that PR were broken; it should not have landed.


bparrishMines (approved):
LGTM

PR #10058, opened 2025-09-23
[google_fonts] Make a map keyed on FontWeight non-const

jason-simmons:
A proposed engine PR will allow construction of arbitrary FontWeight instances and thus needs to overload the FontWeight equality operator (see https://github.com/flutter/flutter/pull/175771)

With that change it will no longer be possible to use FontWeight as the key in a const map.

### DIFF
--- a/packages/google_fonts/CHANGELOG.md
+++ b/packages/google_fonts/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 6.3.2
 
-- Make a map keyed on FontWeight non-const for compatibility with proposed changes to the engine's implementation of FontWeight.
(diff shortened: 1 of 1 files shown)


packages/google_fonts/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 6.3.2
+
+- Make a map keyed on FontWeight non-const for compatibility with proposed changes to the engine's implementation of FontWeight.
```
stuartmorgan-g: Nit: Makes

https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style

jason-simmons: Done

stuartmorgan-g (approved):

PR #10066, opened 2025-09-23
Remove "All right reserved" from all files

stuartmorgan-g:
Per guidance from legal on current preferred practice, remove "All rights reserved" from all LICENSE files and header blocks, and update the repo tool check accordingly.


stuartmorgan-g:
@bparrishMines Could you review this one? Everyone else, please feel free to ignore.

PR #10148, opened 2025-09-30
[google_sign_in] Remove references to dart:js_util

fishythefish:
Flutter has already migrated from dart:js_util to dart:js_interop (which is unsupported by dart2wasm). This migrates a remaining reference to js_util which will be broken when dart2wasm drops the library entirely.


fishythefish:
This is a trivial migration that shouldn't require a CHANGELOG entry, so if a reviewer could add the "override: no changelog needed" label, that'd be lovely.

stuartmorgan-g:

stuartmorgan-g (approved):
LGTM

PR #10186, opened 2025-10-07
[many] Update compileOptions to use Java 17

reidbaker:
Fixes #flutter/flutter/issues/176027

Reviewers be sure to checkout `script/tool/lib/src/gradle_check_command.dart` which is a tools change. 

- **update tool**
- **update compile options to java 17**
- **Update min flutter version**
- **Update changelog and yamls**

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/pigeon/lib/src/generator_tools.dart
+++ b/packages/pigeon/lib/src/generator_tools.dart
@@ -15,7 +15,7 @@ import 'generator.dart';
 /// The current version of pigeon.
 ///
 /// This must match the version in pubspec.yaml.
-const String pigeonVersion = '26.0.1';
+const String pigeonVersion = '26.1.0';
 
 /// Read all the content from [stdin] to a String.
 String readStdin() {


reidbaker:
Will get an approval then handle merge conflicts.

jesswrd (approved):
Thanks again for agreeing to let this land first: https://github.com/flutter/packages/pull/10187. Sorry for raining on your parade :0.

packages/animations/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## NEXT
+## 2.1.0
 
-* Updates minimum supported SDK version to Flutter 3.29/Dart 3.7.
+* Updates Java compatibility version to 17.
+* If required, Updates minimum supported SDK version to Flutter 3.35/Dart 3.9.
```
stuartmorgan-g, line 4: The "If required" is kind of confusing for a client-facing message. If you don't want to manually trim the line out of the packages that don't need it, I would just say "Sets minimum supported SDK version to Flutter 3.35/Dart 3.9."

packages/camera/camera_android/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Android implementation of the camera plugin.
 repository: https://github.com/flutter/packages/tree/main/packages/camera/camera_android
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+camera%22
 
-version: 0.10.10+8
+version: 0.10.11
```
stuartmorgan-g, line 6: 0.10.10+1? I would expect this to be a bugfix version bump for everything, is that not the case?

packages/file_selector/file_selector/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.1.0
```
stuartmorgan-g, line 1: This definitely shouldn't be a minor version bump; it's only changing example and test code.

stuartmorgan-g: (This is probably more manual tweaking than you want to do, but policywise it would be fine to skip changelog and version updates on the packages where the only change is to example and test code. The tool will think you need to version it because `lib/main.dart` is changing, but updating the autoformat version on pub.dev is not actually something that matters.)

packages/file_selector/file_selector/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+
+* Updates Java compatibility version to 17.
```
stuartmorgan-g, line 3: Optional since it's manual work, but ideally this line should be removed from every package except the Android plugin implementation packages, since changing the Java version that example app uses is not a package-client-affecting change.

packages/file_selector/file_selector/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+* If required, Updates minimum supported SDK version to Flutter 3.35/Dart 3.9.
```
stuartmorgan-g, line 4: Actually, here's something we could consider: I'd be fine with changing the Java version used in just the example app *without* changing the min SDK version. Technically this means someone *could* fail to build an example app that it looks like they should be able to but this would only happen:
- if someone is checking out the repo and building the example app in the first place, which I believe to be quite rare outside of contributors, and
- they are using an old version of Flutter, which is very unlikely for contributors, and
- they are using an old version of Java.

So an option here would be reverting the changes in this PR for all packages where the only gradle change is in `example/`, and then doing a second PR for just the examples that *only* changes the Java version, and has no versioning, no changelog, no min SDK updates (and thus reformatting), etc.

reidbaker: I dont have a strong opinion on the order these things happen I just want to move to java 17 in the gradle files. 

I would really like if I could use the same changelog entry for all packages and if I could ship a new version for all packages but I will do whatever you want. 

Based on the comment above and @jesswrd pr that updated most packages for gradle 9 I will start this process over from scratch but invert the order. Start with the examples then try to land a pr that has tooling enforcement and the formatting and version bump changes.

packages/file_selector/file_selector/CHANGELOG.md
reidbaker: When making package wide changes I would only do that if the tooling had a way to figure that out for me.

packages/animations/CHANGELOG.md
reidbaker: If required was so that I could have the same message in all changelogs and handled the case where some plugins had already bumped to 3.35.

packages/file_selector/file_selector/CHANGELOG.md
stuartmorgan-g: We expect all changes in this PR to be a no-op for clients thought, right? I would version all of them as a bugfix, not minor, if that's the case; no package switching required.

PR #10688, opened 2025-12-27
[video_player] : Add video track selection support for Android and iOS

nateshmbhat:
This PR adds video track selection (quality switching) support to the video_player plugin for Android and iOS platforms. This allows users to programmatically get available video quality variants and select a specific quality track, similar to the existing audio track selection feature.

Addresses issue : https://github.com/flutter/flutter/issues/58854

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/video_player/video_player/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
   versions of the endorsed platform implementations.
   * Applications built with older versions of Flutter will continue to
     use compatible versions of the platform implementations.
+* Adds video track (quality) selection support via `getVideoTracks()`, `selectVideoTrack()`, and `isVideoTrackSupportAvailable()` methods.
```
hellohuanlin: it's hard to understadn what "video track (quality) selection" is 
also this should bump version otherwise it won't get released

packages/video_player/video_player/example/ios/Podfile
```diff
@@ -1,5 +1,5 @@
 # Uncomment this line to define a global platform for your project
-# platform :ios, '12.0'
+# platform :ios, '13.0'
```
hellohuanlin: is this change expected? cc @vashworth

packages/video_player/video_player/example/lib/main.dart
```diff
@@ -11,6 +11,8 @@ library;
 import 'package:flutter/material.dart';
 import 'package:video_player/video_player.dart';
```
hellohuanlin, line 13: remove blank line

packages/video_player/video_player/example/macos/Podfile
```diff
@@ -1,4 +1,4 @@
-platform :osx, '10.14'
+platform :osx, '10.15'
```
hellohuanlin: also this change

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -820,6 +823,73 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   }
 
   bool get _isDisposedOrNotInitialized => _isDisposed || !value.isInitialized;
+
+  /// Gets the available video tracks (quality variants) for the video.
+  ///
+  /// Returns a list of [VideoTrack] objects representing the available
+  /// video quality variants. For HLS/DASH streams, this returns the different
+  /// quality levels available. For regular videos, this may return a single
+  /// track or an empty list.
```
hellohuanlin: 1. what is "regular videos"? do you mean non-stream videos?
2. when will it return single track vs empty list

packages/video_player/video_player_avfoundation/darwin/RunnerTests/VideoPlayerTests.m
```diff
@@ -1051,4 +1051,226 @@ - (nonnull AVPlayerItem *)playerItemWithURL:(NSURL *)url {
   return [AVPlayerItem playerItemWithAsset:[AVURLAsset URLAssetWithURL:url options:nil]];
 }
 
+#pragma mark - Video Track Tests
+
+// Tests getVideoTracks with a regular MP4 video file using real AVFoundation.
```
hellohuanlin: what is "real AVFoundation"?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -421,6 +422,130 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
   [self updatePlayingState];
 }
 
+- (void)getVideoTracks:(void (^)(FVPNativeVideoTrackData *_Nullable,
+                                 FlutterError *_Nullable))completion {
+  NSMutableArray<FVPMediaSelectionVideoTrackData *> *mediaSelectionTracks = [NSMutableArray array];
+
+  AVPlayerItem *currentItem = _player.currentItem;
+  if (!currentItem) {
+    completion([[FVPNativeVideoTrackData alloc] init], nil);
```
hellohuanlin: why doesn't it complete with `nil`?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -421,6 +422,130 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+  AVURLAsset *urlAsset = (AVURLAsset *)currentItem.asset;
+  if (![urlAsset isKindOfClass:[AVURLAsset class]]) {
+    completion([[FVPNativeVideoTrackData alloc] init], nil);
+    return;
+  }
+
+  // Use AVAssetVariant API for iOS 15+ to get HLS variants
+  if (@available(iOS 15.0, macOS 12.0, *)) {
+    [urlAsset
+        loadValuesAsynchronouslyForKeys:@[ @"variants" ]
```
hellohuanlin: does apple provide a constant for this key? if not, it should be our own constant, with a comment saying it's not provided by apple.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -421,6 +422,130 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+                                  FourCharCode codecType = [codecTypes[0] unsignedIntValue];
+                                  codec = [self codecStringFromFourCharCode:codecType];
+                                }
+                              }
+
+                              // Determine if this variant is selected (approximate match by
+                              // bitrate)
+                              BOOL isSelected =
+                                  (currentBitrate > 0 &&
+                                   fabs(peakBitRate - currentBitrate) < peakBitRate * 0.1);
```
hellohuanlin, line 518: if i read it right, it's checking bit rate within 10% range? can you explain it in the comment?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -421,6 +422,130 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+                                                     label:label
+                                                   bitrate:peakBitRate > 0
+                                                               ? @((NSInteger)peakBitRate)
+                                                               : nil
+                                                     width:videoSize.width > 0
+                                                               ? @((NSInteger)videoSize.width)
+                                                               : nil
+                                                    height:videoSize.height > 0
+                                                               ? @((NSInteger)videoSize.height)
+                                                               : nil
```
hellohuanlin: the indentation of the completion handler is a bit out of hand and hard to read. can you create 1-2 helper methods?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -421,6 +422,130 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+                        });
+                      }];
+  } else {
+    // For iOS < 15, return empty list as AVAssetVariant is not available
+    completion([[FVPNativeVideoTrackData alloc] init], nil);
+  }
+}
+
+- (NSString *)codecStringFromFourCharCode:(FourCharCode)code {
+  // Convert common video codec FourCharCodes to readable strings
```
hellohuanlin: > readable strings
what's the strings for? as error output?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -421,6 +422,130 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+                        dispatch_async(dispatch_get_main_queue(), ^{
+                          NSError *error = nil;
+                          AVKeyValueStatus status = [urlAsset statusOfValueForKey:@"variants"
+                                                                            error:&error];
+
+                          if (status == AVKeyValueStatusLoaded) {
+                            NSArray<AVAssetVariant *> *variants = urlAsset.variants;
+                            double currentBitrate = currentItem.preferredPeakBitRate > 0
+                                                        ? currentItem.preferredPeakBitRate
+                                                        : 0;
```
hellohuanlin: can be just `max(..., 0)`

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -421,6 +422,130 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+
+                            for (NSInteger i = 0; i < variants.count; i++) {
```
hellohuanlin: nit: for variant in variants

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -421,6 +422,130 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+                              }
+
+                              // Determine if this variant is selected (approximate match by
+                              // bitrate)
+                              BOOL isSelected =
+                                  (currentBitrate > 0 &&
+                                   fabs(peakBitRate - currentBitrate) < peakBitRate * 0.1);
+
+                              // Generate label from resolution
+                              NSString *label = nil;
```
hellohuanlin: nit: need more descriptive than just `label`

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
```diff
@@ -421,6 +422,130 @@ - (void)setPlaybackSpeed:(double)speed error:(FlutterError *_Nullable *_Nonnull)
...
+                                                isSelected:isSelected];
+                              [mediaSelectionTracks addObject:trackData];
+                            }
+                          }
+
+                          FVPNativeVideoTrackData *result = [FVPNativeVideoTrackData
+                               makeWithAssetTracks:nil
+                              mediaSelectionTracks:mediaSelectionTracks.count > 0
+                                                       ? mediaSelectionTracks
+                                                       : nil];
```
hellohuanlin: why not empty array?

hellohuanlin (commented):
thanks for the contribution. this seems like a useful feature. left some comments on iOS part.

packages/video_player/video_player/example/lib/main.dart
nateshmbhat: done

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: updated comments

packages/video_player/video_player/CHANGELOG.md
nateshmbhat: updated

nateshmbhat:
@hellohuanlin  Thanks for the review. have addressed your comments 🙏🏻

packages/video_player/video_player/pubspec.yaml
```diff
@@ -25,10 +25,14 @@ dependencies:
...
-  video_player_platform_interface: ^6.3.0
-  video_player_web: ^2.1.0
+  video_player_android:
+    path: ../video_player_android
+  video_player_avfoundation:
+    path: ../video_player_avfoundation
+  video_player_platform_interface:
+    path: ../video_player_platform_interface
+  video_player_web:
+    path: ../video_player_web
```
LouiseHsu: I don't think these changes are correct. We can't pin dependencies to a local path in pubspec.yaml - they must be pinned to published versions. Otherwise, pub publish will reject the package.

packages/video_player/video_player_android/example/pubspec.yaml
```diff
@@ -18,7 +18,8 @@ dependencies:
     # The example app is bundled with the plugin so we use a path dependency on
     # the parent directory to use the current plugin's version.
     path: ../
-  video_player_platform_interface: ^6.6.0
+  video_player_platform_interface:
```
LouiseHsu: same comment as my above pubspec comment

packages/video_player/video_player_android/pubspec.yaml
```diff
@@ -20,7 +20,8 @@ flutter:
 dependencies:
   flutter:
     sdk: flutter
-  video_player_platform_interface: ^6.6.0
+  video_player_platform_interface:
```
LouiseHsu: here too

packages/video_player/video_player_avfoundation/example/pubspec.yaml
```diff
@@ -16,7 +16,8 @@ dependencies:
     # The example app is bundled with the plugin so we use a path dependency on
     # the parent directory to use the current plugin's version.
     path: ../
-  video_player_platform_interface: ^6.3.0
+  video_player_platform_interface:
```
LouiseHsu: here too

packages/video_player/video_player_avfoundation/pubspec.yaml
```diff
@@ -24,7 +24,8 @@ flutter:
 dependencies:
   flutter:
     sdk: flutter
-  video_player_platform_interface: ^6.3.0
+  video_player_platform_interface:
```
LouiseHsu: here too

packages/video_player/video_player_web/example/pubspec.yaml
```diff
@@ -8,7 +8,8 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  video_player_platform_interface: ^6.3.0
+  video_player_platform_interface:
```
LouiseHsu: here too

packages/video_player/video_player_web/pubspec.yaml
```diff
@@ -21,7 +21,8 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  video_player_platform_interface: ^6.4.0
+  video_player_platform_interface:
```
LouiseHsu: here too

packages/video_player/video_player/pubspec.yaml
LouiseHsu: btw to test stuff locally, you can try using a pubspec_overrides.yaml which is gitignored, but will still let you use your local changes.

https://dart.dev/tools/pub/dependencies#pubspec-overrides

LouiseHsu (changes requested):

packages/video_player/video_player/pubspec.yaml
nateshmbhat: if i don't add path based override won't it cause compile errors and break the CI checks ?  

These paths are done by running the make-deps-path based command as documented here :

https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins

Once the review is done and the interfaces are published from the breakout PRs then this PR will be updated to use the newer versions of the dependencies.

I used the same approach for #9925 PR also. @stuartmorgan-g pointed me to refer to the above doc initially.

nateshmbhat: These paths are done by running the make-deps-path based command as documented here :

https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins

Once the review is done and the interfaces are published then it will be updated.

Pls check #9925 PR as an example

nateshmbhat: These paths are done by running the make-deps-path based command as documented here :

https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins

Once the review is done and the interfaces are published then it will be updated.

Pls check #9925 PR as an example

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/FVPVideoPlayer.m
hellohuanlin: Circling back on this. Is `[[FVPNativeVideoTrackData alloc] init]` a valid track data? why not use `nil` to indicate no data?

hellohuanlin (approved):
just one question and then lgtm after address @LouiseHsu's feedback

packages/video_player/video_player/pubspec.yaml
stuartmorgan-g: > These paths are done by running the make-deps-path based command

This change was definitely not made by the repo tooling. The repo tooling creates a dependency overrides section, with a clear comment explain why it's there. It doesn't delete anything from the actual dependencies.

This looks like a manual attempt to create path-based dependencies in a way that doesn't match our standard process. Not following our standard process leads to reviewer confusion, as evidenced here.

nateshmbhat: Oops terribly sorry about the confusion that I created. I had mistaken that I had done this change through tooling command itself as I had initially used path based dependencies and then ran the make-path-based command after that. 

sorry about this. Have pushed the fix.

LouiseHsu (approved):

packages/video_player/video_player_android/pigeons/messages.dart
```diff
@@ -192,6 +235,13 @@ abstract class VideoPlayerInstanceApi {
 
   /// Selects which audio track is chosen for playback from its [groupIndex] and [trackIndex]
   void selectAudioTrack(int groupIndex, int trackIndex);
+
+  /// Gets the available video tracks for the video.
+  NativeVideoTrackData getVideoTracks();
+
+  /// Selects which video track is chosen for playback from its [groupIndex] and [trackIndex].
+  /// Pass -1 for both indices to enable auto quality selection.
+  void selectVideoTrack(int groupIndex, int trackIndex);
```
camsim99, line 243: Nit: have you considered implementing a separate method for auto quality selection? Might be more readable.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -233,6 +233,171 @@ public void selectAudioTrack(long groupIndex, long trackIndex) {
...
+          boolean isSelected = group.isTrackSelected(trackIndex);
+
+          // Create video track data with metadata
+          ExoPlayerVideoTrackData videoTrack =
+              new ExoPlayerVideoTrackData(
+                  (long) groupIndex,
+                  (long) trackIndex,
+                  format.label,
+                  isSelected,
+                  format.bitrate != Format.NO_VALUE ? (long) format.bitrate : null,
```
camsim99: Nit: For readability, you could implement a `getFormatValue(formatValue)` type of helper method

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -233,6 +233,171 @@ public void selectAudioTrack(long groupIndex, long trackIndex) {
...
+    // a fresh renderer properly configured for the new dimensions.
+    //
+    // Why is this necessary?
+    // When switching between video tracks with different resolutions (e.g., 720p to 1080p),
+    // the existing video surface and MediaCodec decoder may not properly reconfigure for the
+    // new dimensions. This can cause visual glitches where the video appears in the wrong
+    // position (e.g., top-left corner) or the old surface remains partially visible.
+    // By disabling the video track type, we force ExoPlayer to completely release the
+    // current renderer and decoder, ensuring a clean slate for the new resolution.
+    if (dimensionsChanged) {
```
camsim99, line 399: This disabling/re-enabling hack make sense thanks to your explanation. However, I just want to make sure there's no more direct way to force a re-render. Is there some documentation that we can point to as to why we are handling it this way?

camsim99 (commented):
Android overall LGTM just have one question about how we force a video player re-render after selecting a new track.

packages/video_player/video_player_android/pigeons/messages.dart
nateshmbhat: done

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: updated

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: updated with relevant details

camsim99: Those links don't clearly state why we need this approach. I did a little digging and found this Exoplayer issue that does seem relevant though: https://github.com/androidx/media/issues/1631. My read on those issues is that the seamless changing of tracks with different formats is currently not supported and their suggested workaround is to disallow that, but please correct me if I'm wrong.

Unless we can find evidence for this solution of disabling/re-enabling the video track type with a delay being reliable, I would suggest that we move in that direction versus implementing a hack. For now, then, we could file an issue for supporting this in the future and follow up in the Exoplayer bug.

nateshmbhat: The thing is, during my testing I was actually facing the visual glitches and that's why I had implement this approach after digging around for solutions.

stuartmorgan-g:
@tarrinneal I believe this still needs your review for the cross-platform parts.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
camsim99: I think instead of landing this initially with a hack to support this on Android, we should not allow the switching of media formats for now and file an issue to follow up on that. Then a fix for the issue could potentially land as a follow up PR. I'm open to pushback, but I think we should invest in finding a reliable solution for this issue if possible. We can follow up in the ExoPlayer bug I linked, too.

nateshmbhat: Thanks, this is a fair concern. I agree we should avoid shipping an Android-specific workaround unless it is addressing a real failure mode and is scoped narrowly.

In this case, the reset path is only used when the selected track changes video dimensions. Same-dimension switches still use the normal ExoPlayer override path.

This also came from an actual issue I hit during manual testing: without the renderer reset, switching between HLS variants with different resolutions could leave the video visually corrupted. With the reset, the switch completed correctly.

I also took a closer look at the Media3/ExoPlayer issue you linked. My read is that it is discussing a different problem: avoiding playback reset/rebuffering when TrackSelector parameters are changed and applied lazily to future chunk selection. The issue here is different: an explicit user-triggered switch where playback continues, but the renderer/surface can end up in a broken visual state after a dimension change.

My hesitation with disallowing format changes for now is that this removes most of the practical value of manual quality selection on Android, since the common HLS variants users actually want to switch between usually differ in resolution/bitrate.

I’m happy to make the intent/scope clearer in code comments and file a follow-up issue to revisit this if/when there is a cleaner upstream-supported solution, but given the known rendering failure in a common scenario, I think the guarded workaround is the better tradeoff for now.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -1394,3 +1462,116 @@ class ClosedCaption extends StatelessWidget {
...
+  /// - iOS: `"variant_{bitrate}"` for HLS adaptive streams
+  final String id;
+
+  /// Whether this track is currently selected.
+  final bool isSelected;
+
+  /// Human-readable label for the track (e.g., "1080p", "720p").
+  ///
+  /// Defaults to an empty string if not available from the platform.
+  final String label;
```
tarrinneal: Curious why you chose to have this one field non-nullable with a default instead of allowing null like the rest of the fields on the class?

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
tarrinneal: Should leave a TODO: for these at least. Hopefully this workaround won't be necessary forever.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -989,6 +989,74 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
 
   bool get _isDisposedOrNotInitialized => _isDisposed || !value.isInitialized;
+
+  /// Gets the available video tracks (quality variants) for the video.
+  ///
+  /// Returns a list of [VideoTrack] objects representing the available
+  /// video quality variants. For HLS/DASH adaptive streams, this returns the
+  /// different quality levels (resolution/bitrate variants) available.
+  /// For non-adaptive videos (MP4, MOV, etc.), this returns an empty list
+  /// as they have a single fixed quality that cannot be switched.
```
tarrinneal: Is this always true? Could a non-adaptive video send a single track? I think it's possible in the platform implementations

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -153,6 +153,46 @@ abstract class VideoPlayerPlatform extends PlatformInterface {
...
+
+  /// Gets the available video tracks (quality variants) for the video.
+  ///
+  /// Returns a list of [VideoTrack] objects representing the available
+  /// video quality variants. For HLS/DASH streams, this returns the different
+  /// quality levels available. For regular videos, this may return a single
+  /// track or an empty list.
+  ///
+  /// Note: On iOS 13-14, this returns an empty list as the AVAssetVariant API
+  /// requires iOS 15+.
```
tarrinneal: Probably don't need to mention iOS specifically here. The user will either use the main package or the Foundation package so this won't be useful necessarily.

tarrinneal: I was going to delete this, but I'll just let you decide if you agree or not :)

tarrinneal (commented):
Seems like the non-platform code is pretty solid. I didn't review the platform code in depth, but I didn't see anything glaringly wrong there either.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: done

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: Good catch. I double-checked this and the previous wording was too narrow, so I updated it.

My read from the platform docs is that for non-adaptive video we can’t safely assume only “a single track or an empty list”:

On Android/Media3, a TrackGroup represents tracks with the same content in different formats, and the track selection docs/API also allow for multiple video groups such as alternate video feeds/camera angles:
[Track selection guide](https://developer.android.com/media/media3/exoplayer/track-selection)
[TrackGroup API](https://developer.android.com/reference/androidx/media3/common/TrackGroup)
On Apple’s AVFoundation side, their docs note that a simple asset may have one video track, but more complex assets/compositions can have multiple video tracks:
[Time and Media Representations](https://developer.apple.com/library/archive/documentation/AudioVideo/Conceptual/AVFoundationPG/Articles/06_MediaRepresentations.html)
[About AVFoundation](https://developer.apple.com/library/archive/documentation/AudioVideo/Conceptual/AVFoundationPG/Articles/00_Introduction.html)
So I changed the doc to say that for non-adaptive videos, platform implementations may return one or more tracks, or an empty list, depending on the media and the metadata available.

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
nateshmbhat: updated : )

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: Good point. I changed the app-facing VideoTrack.label to stay nullable instead of defaulting to '', so it now matches the rest of the metadata fields and preserves the platform value more faithfully.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
camsim99: @nateshmbhat Thanks for the explanation! Ultimately, your logic as to keeping the workaround in makes sense to me and I'm okay moving forward.

If you don't mind, can you file an issue for the todo and link it here?

nateshmbhat: done , added here https://github.com/flutter/flutter/issues/183824

camsim99: Thank you! And sorry for being unclear but can you link it in the todo?

nateshmbhat: Done :) ✨

camsim99 (approved):
Android LGTM!

nateshmbhat:
@stuartmorgan-g are we good to create breakout PRs ?

priyankj-simform:
Hi, any update on when these changes will be merged?

stuartmorgan-g:
> @stuartmorgan-g are we good to create breakout PRs ?

It needs a re-review from @tarrinneal to make sure he's happy with all the cross-platform code.

Before that can happen you will need to address the fact that you've pushed a commit that doesn't pass the CLA, so the PR is currently unreviewable.

nateshmbhat:
Fixed it. Requested review from all reviewers by mistake 😅 sorry

reidbaker:
I removed everyone that was not on the ecosystem, android or ios team in an effort to minimize their notification spam.

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -1028,6 +1028,72 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
   }
 
   bool get _isDisposedOrNotInitialized => _isDisposed || !value.isInitialized;
+
+  /// Gets the available video tracks (quality variants) for the video.
```
tarrinneal: Is quality the only thing that tracks are used for?

packages/video_player/video_player/lib/video_player.dart
```diff
@@ -1028,6 +1028,72 @@ class VideoPlayerController extends ValueNotifier<VideoPlayerValue> {
...
+            width: track.width,
+            height: track.height,
+            frameRate: track.frameRate,
+            codec: track.codec,
+          )
+        : null;
+    await _videoPlayerPlatform.selectVideoTrack(_playerId, platformTrack);
+  }
+
+  /// Returns whether video track selection is supported on this platform.
```
tarrinneal: nit: remove returns

tarrinneal:
This should be ready to break out into sub prs

nateshmbhat:
Added breakout PRs for the same  @stuartmorgan-g

almezead:
any news about this ?

stuartmorgan-g:
From triage: the next round of sub-PRs is out for review.

nateshmbhat:
good to merge this main pr ? @tarrinneal

stuartmorgan-g:
> good to merge this main pr ? @tarrinneal

There are two unresolved comments from his review above.

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: Good point — video tracks aren'\''t only quality. I'\''ve reworded the docs (getVideoTracks/selectVideoTrack/VideoTrack) so they no longer describe tracks solely as quality variants; adaptive bitrate ladders are now just the common example. The API is intentionally scoped to selecting among the available tracks (mirroring the audio track API) for this first version.

packages/video_player/video_player/lib/video_player.dart
nateshmbhat: Done — dropped the "Returns" lead-in on isVideoTrackSupportAvailable (and the getVideoTracks doc).

nateshmbhat:
Resolved @stuartmorgan-g @tarrinneal

tarrinneal (approved):

nateshmbhat:
@stuartmorgan-g can you pls review and add cicd label ? 

we're good to merge this PR

stuartmorgan-g:
I'm not a reviewer for this PR. In the future, please reach out to a reviewer for help getting a PR landed once it's been reviewed.

PR #10714, opened 2026-01-05
[google_maps_flutter] Support Google Maps onMapLoaded

### LINKED ISSUE #99610: [google_maps_flutter] Expose an "onTilesLoaded" to signal when all tiles have been fully rendered.
## Use case

The flutter plugin exposes an `onMapCreated()` function that states when the map is ready.
However, until all tiles are fully loaded it shows a grid with grey lines, and any effect or any interaction with the map seems useless until tiles are rendered.

These are the use cases I wish to accomplish:
- Ensure all tiles are rendered before taking a snapshot.
Running the method `takeSnapshot()` after `onMapCreated()` leads to an image with either, partial tiles rendered or none at all.
- Zoom-in to a position once the map is first loaded.
The zoom effect is kind of weird if done after `onMapCreated()` is fired, since it will zoom a grid of grey lines.

## Proposal

Each platform already contains a functionality to determine when tiles are fully rendered:
- Android - `GoogleMap.OnMapLoadedCallback`
- iOS - `GMSMapViewDelegate.mapViewDidFinishTileRendering`
- Web - `Map.tilesloaded`

However, AFAIK, there is no such functionality in the flutter plugin.

hearsilent:
Implements the GoogleMap.OnMapLoadedCallback interface and triggers the onMapLoaded event via platform channels on Android and iOS, with corresponding handling on Web. Updates the generated Pigeon message files and platform interface to support the new event, enabling Flutter apps to listen for when the map has fully loaded across all supported platforms.

Fixes https://github.com/flutter/flutter/issues/99610

## Pre-Review Checklist

- [x] I signed the [CLA].


stuartmorgan-g:
Thanks for the contribution!

Two initial high-level notes while this waits for review:
- Please follow the instructions at https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins so that CI can run successfully.
- The PR needs to address web in some way, either by adding web support, or by explaining in the PR why it can't be, and adding [a support query](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#api-support-queries). Currently clients who put important code in this callback would be silently broken on web.

hearsilent:
Thanks for the notes!

Both points have been fully addressed:
- The federated plugin change instructions have been followed, and CI should now run successfully.
- Web is now handled appropriately, so clients relying on this callback will not be silently broken.

Please let me know if there’s anything else to adjust while this is under review.

stuartmorgan-g:
The CI failures all look like real failures now, which will need to be resolved (other than the expected `federated safety check`).

hearsilent:
Thanks for the heads up. All CI failures have now been resolved, except for the expected federated safety check.

vashworth (approved):
iOS side LGTM

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
```diff
@@ -173,6 +173,16 @@ class MethodChannelGoogleMapsFlutter extends GoogleMapsFlutterPlatform {
     return _events(mapId).whereType<ClusterTapEvent>();
   }
 
+  @override
+  Stream<GroundOverlayTapEvent> onGroundOverlayTap({required int mapId}) {
+    return _events(mapId).whereType<GroundOverlayTapEvent>();
+  }
+
+  @override
+  Stream<MapLoadedEvent> onMapLoaded({required int mapId}) {
```
LongCatIsLooong, line 182: I'm not familiar with the API conventions of this package, but each event in the `Stream` basically carries no information so it might as well be `Void`. It looks to me that the `MapLoadedEvent` type only exists so that this implementation can use it to filter the event stream and extract the relevant events, and `MapLoadedEvent` is not useful to the caller of this API at all. If that's indeed the case, would it be possible to hide the `MapLoadedEvent` type? Or is it made public for subclassing?

LongCatIsLooong: I assume with the current interface, to provide a "start loading" event stream in the future would require adding a new `onMapLoading` method to this interface. I would expect a lot of use cases of this API would also involve listening to the "Map started loading" event stream and combining the two streams (e.g., I want to perform asynchronous work on map load, and cancel that work if the map starts loading again).

IMO it might be a bit more ergonomic to return a `Stream` that gives you both start and finish events (so something like `Stream<bool>`, or the boolean flag can be added to `MapLoadEvent`), or even a `ValueNotifier<bool>` (but now the caller would have to manage the lifespan of the `ValueNotifier` so I'm not sure about that).

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -290,6 +291,13 @@ class GoogleMap extends StatefulWidget {
   /// Called every time a [GoogleMap] is long pressed.
   final ArgumentCallback<LatLng>? onLongPress;
 
+  /// Called when the tiles have been loaded.
+  ///
+  /// This implementation maps to [Map.tilesloaded](https://developers.google.com/maps/documentation/javascript/reference/map#Map.tilesloaded) on Web,
+  /// [GoogleMap.OnMapLoadedCallback](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.OnMapLoadedCallback) on Android,
+  /// and [GMSMapViewDelegate.mapViewDidFinishTileRendering](https://developers.google.com/maps/documentation/ios-sdk/reference/protocol_g_m_s_map_view_delegate-p#mapViewDidFinishTileRendering:) on iOS.
```
LongCatIsLooong: fyi this link doesn't work when I pasted it into the address bar. The link below seems to work
```
https://developers.google.com/maps/documentation/ios-sdk/reference/objc/Protocols/GMSMapViewDelegate#-mapviewdidfinishtilerendering:
```

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/method_channel/method_channel_google_maps_flutter.dart
LongCatIsLooong: Ah it looks like this is a low-level interface and most app developers don't have to use this directly. Feel free to ignore the comments then.

LongCatIsLooong (approved):
The iOS implementation lgtm (modulo a potentially broken link).

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
hearsilent: fixed

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
```diff
@@ -294,6 +299,12 @@ class GoogleMapController {
         _streamController.add(CameraIdleEvent(_mapId));
       }
     });
+    _onMapLoadedSubscription = map.onTilesloaded.listen((void _) {
+      _tilesLoaded = true;
+      if (!_streamController.isClosed) {
+        _streamController.add(MapLoadedEvent(_mapId));
+      }
+    });
```
mdebbar, line 306: Can we combine this with the `map.onTilesLoaded.first` on line 257 so there's only one listener on that stream?

Something like this:
```dart
    _onMapLoadedSubscription = map.onTilesloaded.listen((void _) {
      final bool isFirstTilesLoaded = !_tilesLoaded;
      _tilesLoaded = true;
      if (!_streamController.isClosed) {
        if (isFirstTilesLoaded) {
          // Report the map as ready to go the first time the tiles load
          _streamController.add(WebMapReadyEvent(_mapId));
        }
        _streamController.add(MapLoadedEvent(_mapId));
      }
    });
```

mdebbar (approved):
LGTM with a small suggestion!

packages/google_maps_flutter/google_maps_flutter_web/lib/src/google_maps_controller.dart
hearsilent: Yes — I’ve updated it to combine with `map.onTilesLoaded.first` (line 257) so we only register a single listener on that stream now. This avoids multiple subscriptions and keeps the load flow cleaner.

packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.14.2
```
stuartmorgan-g: New public API is a minor version change in semver, not a bugfix version.

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -290,6 +291,13 @@ class GoogleMap extends StatefulWidget {
   /// Called every time a [GoogleMap] is long pressed.
   final ArgumentCallback<LatLng>? onLongPress;
 
+  /// Called when the tiles have been loaded.
```
stuartmorgan-g: Looking at the docs, it seems like it's specifically that all visible tiles have been loaded and rendered. If that's correct, the docs should include that level of detail.

packages/google_maps_flutter/google_maps_flutter/lib/src/google_map.dart
```diff
@@ -290,6 +291,13 @@ class GoogleMap extends StatefulWidget {
...
+  ///
+  /// This implementation maps to [Map.tilesloaded](https://developers.google.com/maps/documentation/javascript/reference/map#Map.tilesloaded) on Web,
+  /// [GoogleMap.OnMapLoadedCallback](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap.OnMapLoadedCallback) on Android,
+  /// and [GMSMapViewDelegate.mapViewDidFinishTileRendering](https://developers.google.com/maps/documentation/ios-sdk/reference/objc/Protocols/GMSMapViewDelegate#-mapviewdidfinishtilerendering:) on iOS.
```
stuartmorgan-g: The app-facing package does not control implementations, so it should not make claims about what the implementation is. This sentence should be removed.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
```diff
@@ -239,6 +240,7 @@ public void onMapReady(@NonNull GoogleMap googleMap) {
       updateMapStyle(initialMapStyle);
       initialMapStyle = null;
     }
+    googleMap.setOnMapLoadedCallback(this);
```
stuartmorgan-g, line 243: This only ever results in one callback. Is that the desired behavior? Does it match the implementation on the other platforms?

packages/google_maps_flutter/google_maps_flutter_android/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: google_maps_flutter_android
 description: Android implementation of the google_maps_flutter plugin.
 repository: https://github.com/flutter/packages/tree/main/packages/google_maps_flutter/google_maps_flutter_android
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+maps%22
-version: 2.18.8
+version: 2.18.9
```
stuartmorgan-g: Same here and throughout the PR; all packages have new public API.

stuartmorgan-g (commented):
Mostly looks good, but one important question about the API behavior.

stuartmorgan-g (changes requested):
(That was meant to be a "request changes" review.)

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
hearsilent: Yes, this is the expected behavior.
OnMapLoadedCallback is invoked once per registration when the map has finished rendering all required tiles and labels. It is not a continuous callback and won’t fire again unless re-registered.

This matches the documented behavior and is consistent with other platforms.
https://developers.google.com/maps/documentation/navigation/android-sdk/reference/com/google/android/gms/maps/GoogleMap.OnMapLoadedCallback

hearsilent:
@stuartmorgan-g 
Got it, thanks for clarifying.
The requested changes have been addressed.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
stuartmorgan-g: > Yes, this is the expected behavior.

Expected by whom? The Android docs explicitly mention how to get more callbacks (suggesting they think people may want that), and the web API appears to have the opposite behavior. Is there evidence suggesting that this variant is the one people want?

I'm also confused by the platform interface API if this is the expected behavior. Why is a stream being returned for something that can only happen one time?

> OnMapLoadedCallback is invoked once per registration when the map has finished rendering all required tiles and labels. It is not a continuous callback and won’t fire again unless re-registered.
> 
> This matches the documented behavior and is consistent with other platforms. https://developers.google.com/maps/documentation/navigation/android-sdk/reference/com/google/android/gms/maps/GoogleMap.OnMapLoadedCallback

I know that's what the Android API says; that's why I asked in the first place. The iOS API, on the other hand, doesn't say (and the existence of `mapViewDidStartTileRendering:` suggests the opposite since I'm not sure why it would be needed for a one-time call during map initialization). What scenarios have you tested on iOS to ensure that it doesn't retrigger? Large moves? Tile changes?

Also, nothing in the public Dart APIs added in this PR say what the intended behavior is.

hearsilent: Sorry about that — I was mistaken earlier. Thanks for the detailed follow-up.

I’ve now updated the API documentation to clearly state that this is a one-time callback per registration, not a continuous or repeatable event. The Dart-facing docs no longer imply stream-like behavior.

Appreciate you calling this out and helping catch the inconsistency.

stuartmorgan-g: > I’ve now updated the API documentation to clearly state that this is a one-time callback per registration, not a continuous or repeatable event.

That is not at all what the docs say in the updated version. Instead it says it's just exposing all the platform differences directly to the client, but without any of the associated platform behavior like iOS's `mapViewDidStartTileRendering:`.

At this point, it seems like there's clearly a non-trivial question of what exactly the desired semantics of a coherent cross-platform API are, and how that could be implemented on each platform, so before continuing to iterate on the PR the next step would be a short [design document](https://github.com/flutter/flutter/blob/master/docs/contributing/Design-Documents.md) discussing the options and their tradeoffs, and we can get alignment on that at a high level before proceeding here.

camsim99:
@hearsilent Do you mind taking a look at the merge conflicts so it will be easier to review this PR? Thank you!

stuartmorgan-g:
@camsim99 The Android conflicts are just the usual version/changelog conflicts, which we generally don't ask people to resolve during incremental review due to the frequency of conflicts there.

camsim99:
> @camsim99 The Android conflicts are just the usual version/changelog conflicts, which we generally don't ask people to resolve during incremental review due to the frequency of conflicts there.

Oh I missed that, sorry!

stuartmorgan-g:
From triage: @hearsilent Are you planning on following up with a design document as [requested above](https://github.com/flutter/packages/pull/10714#discussion_r2828527528)?

hearsilent:
@stuartmorgan-g I have this on my radar! I'm a bit busy with other tasks at the moment, but I’ll follow up with the design doc over the next few days.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
hearsilent: ## Problem Statement

The `google_maps_flutter` plugin needs an `onMapLoaded` callback to notify developers when the map has finished rendering its tiles. However, the underlying native Google Maps SDKs have **fundamentally different semantics** for their "map loaded" events:

| Platform | Native API | Fires Once? | Fires After Camera Move? |
|----------|-----------|-------------|--------------------------|
| **Android** | `GoogleMap.OnMapLoadedCallback` | Yes (per registration) | No (must re-register) |
| **iOS** | `mapViewDidFinishTileRendering:` | No | Yes |
| **Web** | `tilesloaded` event | No | Yes |

This creates a design question: **what should the cross-platform Flutter API promise to developers?**

## Native API Details

### Android: `GoogleMap.OnMapLoadedCallback`

- Called **once** when all visible tiles have been loaded and all labeling is complete.
- This is a **one-shot callback**: after it fires, it is automatically unregistered.
- To receive subsequent notifications, developers must call `setOnMapLoadedCallback()` again.
- **Will not fire** if the map never reaches a fully loaded state (e.g., connectivity issues, continuous interaction).
- Runs on the Android UI thread.

### iOS: `mapViewDidFinishTileRendering:` (GMSMapViewDelegate)

- Called whenever **tile rendering completes**, which can happen many times: after initial load, after panning, zooming, or map data updates.
- This is a **continuous delegate method**: it fires every time the map finishes rendering a set of tiles.
- iOS also offers `mapViewSnapshotReady:`, which fires when the map is fully stable (tiles loaded + labels rendered + camera idle + overlays rendered). This is a stricter "everything is done" signal.

### Web (JavaScript API): `tilesloaded` event

- Fires every time visible tiles finish loading — initial load, pan, zoom, map type changes.
- This is a **continuous event listener**.
- Will not fire if no new tiles need to be fetched (e.g., all tiles already cached).
- The `idle` event is a separate concept that fires when the map becomes completely idle.

## Design Options

### Option A: One-Time Callback ✅ (Recommended)

**Guarantee**: `onMapLoaded` fires **exactly once** after the initial map tiles finish rendering.

| Platform | Approach |
|----------|----------|
| Android | Use `OnMapLoadedCallback` directly (native one-shot behavior). |
| iOS | Listen to `mapViewDidFinishTileRendering:`, fire once, ignore subsequent calls. |
| Web | Listen to `tilesloaded`, fire once (using a flag), ignore subsequent events. |

**Pros:**
- Simple, predictable contract — developers know the callback fires once.
- Covers the most common use case: hiding a loading spinner, enabling UI, or taking a screenshot.
- Consistent with the existing `onMapCreated` callback pattern.
- Non-breaking path to Option D if repeatable events are needed later.

**Cons:**
- Cannot detect subsequent tile loads (but `onCameraIdle` covers most of that need).

---

### Option B: Repeatable Callback

**Guarantee**: `onMapLoaded` fires **every time** the map finishes rendering tiles.

| Platform | Approach |
|----------|----------|
| Android | Re-register `OnMapLoadedCallback` after each invocation. |
| iOS | Use `mapViewDidFinishTileRendering:` directly. |
| Web | Use `tilesloaded` event directly. |

**Pros:**
- Exposes full native capability; supports advanced use cases (e.g., showing a loading indicator during pans/zooms).

**Cons:**
- Android re-registration is non-standard; potential edge cases (gap between callback firing and re-registration during rapid interactions, undocumented behavior differences).
- Performance concerns: frequent callbacks could cause issues if developers do expensive work.
- Even with re-registration, subtle timing/frequency differences across platforms may remain.

---

### Option C: Document Platform Differences

**Guarantee**: None — each platform exposes its native behavior, documented in the API.

This is the approach taken by the current PR. Android fires once; iOS/Web fire repeatedly.

**Pros:**
- Simplest implementation.

**Cons:**
- Inconsistent API contract violates Flutter's cross-platform philosophy.
- Shifts the burden of handling differences to every developer.

---

### Option D: Dual API

Provide both semantics through different API surfaces:
- `onMapLoaded` (`VoidCallback?`): Fires **once** after initial rendering.
- `onTilesRendered` (stream or separate callback): Fires **every time** tiles finish rendering.

**Pros:**
- Best of both worlds; clean separation of concerns.

**Cons:**
- Larger API surface.
- Same Android re-registration concerns for the repeatable stream.
- Could be added later on top of Option A without breaking changes.

## Recommendation: Option A

**Option A (One-Time Callback)** is recommended because:

1. **Strongest cross-platform consistency**: All platforms guarantee exactly one callback.
2. **Addresses the primary use case**: Knowing when the map is initially ready.
3. **Simplest and safest**: No Android re-registration edge cases, no performance concerns.
4. **Forward-compatible**: A repeatable stream (`onTilesRendered`) can be added later via Option D without modifying `onMapLoaded`.
5. **Follows existing patterns**: Consistent with `onMapCreated` (also fires once).

### Implementation Changes

- **Android**: No changes — native `OnMapLoadedCallback` already fires once.
- **iOS**: Add a boolean flag to `GoogleMapController.m` to fire `didFinishRenderingMap` only once.
- **Web**: Only emit `MapLoadedEvent` on the first `tilesloaded` event (already tracks `_tilesLoaded` flag).
- **API docs**: Update `onMapLoaded` doc comment to describe consistent one-time behavior.

## `onMapLoaded` vs. `onCameraIdle`

These are complementary signals:
- `onCameraIdle` fires when camera **animations complete** — tiles may still be loading.
- `onMapLoaded` fires when **visual content is rendered** — the map is visually ready.

Developers needing ongoing state change detection should use `onCameraIdle`, which fires on all platforms after camera movements complete.

stuartmorgan-g: @hearsilent Please follow the design doc process I linked to. GitHub's linear commenting system is not suitable for design doc feedback, which is why we have a separate system for design docs.

gmackall:
From android triage: removing triage label based on discussion that a design document is in the works

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/GoogleMapController.java
hearsilent: @stuartmorgan-g Got it… I’ll go through the design doc process you shared and try to get everything set up properly. This might take me a bit, but I’ll work through it.

hearsilent: @stuartmorgan-g I’ve already submitted the design doc for this proposal:
https://flutter.dev/go/onmaploaded

Please let me know if there’s anything specific you’d like me to update or clarify in the document.

stuartmorgan-g:
From triage: This is waiting on comments on the design doc to be addressed.

stuartmorgan-g:
From triage: @hearsilent It looks like the design doc has never been updated. Is this still something you are planning on moving forward?

Marking as a draft pending design doc approval.

Piinks:
Thank you for your contribution. I'm going to close this PR for now since there are outstanding comments, just to get this off our PR review queue. Please don't hesitate to re-open or submit a new PR if you have the time to address the review comments. Thanks!

PR #11143, opened 2026-02-28
[go_router] Fix key collisions when sibling ShellRoutes coexist during push

Gorniv:
This PR fixes key collisions in `go_router` when `push()` navigates from one `ShellRoute` branch to a sibling `ShellRoute` branch under the same parent.

In that scenario, both shell branches can temporarily coexist in the widget tree. Before this change, that could cause framework assertions such as `Multiple widgets used the same GlobalKey` because:
- `_CustomNavigator` was keyed with `GlobalObjectKey(navigatorKey.hashCode)`, which could collide when two navigator keys had the same `hashCode`
- `ShellRouteMatch` reused the configuration-time `ShellRoute.navigatorKey`
- `ShellRouteMatch.pageKey` was derived only from `route.hashCode`, so multiple matches for the same `ShellRoute` could share page identity

This PR fixes the problem for regular `ShellRoute` by:
- creating a fresh `GlobalKey<NavigatorState>` for each `ShellRouteMatch`
- deriving a unique `pageKey` from that per-match navigator key
- keying `_CustomNavigator` with `GlobalObjectKey(navigatorKey)` instead of `GlobalObjectKey(navigatorKey.hashCode)`

The change is intentionally scoped to `ShellRoute` only. `StatefulShellRoute` keeps its existing stable `pageKey` behavior because applying per-match page identity there can create duplicate `StatefulNavigationShell` instances with the same internal `GlobalKey`.

This PR also adds a regression test that reproduces the old crash by forcing a navigator key hash collision, verifies that pushing to a sibling shell route no longer crashes, and confirms that the previous shell route state is preserved after `pop()`.

No screenshots are included since this is a routing/runtime fix rather than a visual UI change.

- https://github.com/flutter/flutter/issues/148712
- https://github.com/flutter/flutter/issues/140586

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/go_router/lib/src/builder.dart
+++ b/packages/go_router/lib/src/builder.dart
@@ -108,7 +108,7 @@ class RouteBuilder {
       context,
       _CustomNavigator(
         // The state needs to persist across rebuild.
-        key: GlobalObjectKey(configuration.navigatorKey.hashCode),
+        key: GlobalObjectKey(configuration.navigatorKey),
         navigatorKey: configuration.navigatorKey,
         observers: observers,
         navigatorRestorationId: restorationScopeId,
@@ -300,7 +300,7 @@ class _CustomNavigatorState extends State<_CustomNavigator> {
               canPop: match.matches.length == 1,
               child: _CustomNavigator(
                 // The state needs to persist across rebuild.
-                key: GlobalObjectKey(navigatorKey.hashCode),
+                key: GlobalObjectKey(navigatorKey),
                 navigatorRestorationId: restorationScopeId,
                 navigatorKey: navigatorKey,
                 matches: match.matches,
--- a/packages/go_router/lib/src/match.dart
+++ b/packages/go_router/lib/src/match.dart
@@ -175,14 +175,22 @@ abstract class RouteMatchBase with Diagnosticable {
     if (subRouteMatches?.isEmpty ?? true) {
       return _empty;
     }
+    final matchNavigatorKey = route is ShellRoute
+        ? GlobalKey<NavigatorState>(debugLabel: navigatorKeyUsed.toString())
+        : navigatorKeyUsed;
+    final pageKey = route is ShellRoute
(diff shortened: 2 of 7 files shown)


packages/go_router/lib/src/match.dart
```diff
@@ -175,14 +175,22 @@ abstract class RouteMatchBase with Diagnosticable {
     if (subRouteMatches?.isEmpty ?? true) {
       return _empty;
     }
+    final matchNavigatorKey = route is ShellRoute
```
chunhtai, line 178: does this mean the shell route navigator won't have stable key anymore? this may cause the navigator to keep rebuild from scratch every time the route is reparsed

packages/go_router/lib/src/builder.dart
```diff
@@ -108,7 +108,7 @@ class RouteBuilder {
       context,
       _CustomNavigator(
         // The state needs to persist across rebuild.
-        key: GlobalObjectKey(configuration.navigatorKey.hashCode),
```
chunhtai, line 111: just want to double check, this is purely clean up right? I would be surprised if this causes different behavior

packages/go_router/test/go_router_test.dart
```diff
@@ -5579,7 +5579,7 @@ void main() {
                     return Scaffold(
                       body: TextButton(
                         onPressed: () async {
-                          shellNavigatorKey.currentState!.push(
```
chunhtai, line 5582: This doesn't feel right. maybe we want a different way to create the persistent globalkey?

stuartmorgan-g:
From triage: @Gorniv Are you planning on updating this PR based on the feedback above?

justinmc:
@Gorniv I'm going to go ahead and close this PR as inactive, but if you end up having time to return to it, please let us know and we'll be happy to help you move forward with it.

PR #11150, opened 2026-03-02
[go_router] Allow users to specify onExit as optional

hantrungkien:
Related to https://github.com/flutter/flutter/issues/183099

Allow users to specify onExit as optional

## Pre-Review Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter].
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I followed [the version and CHANGELOG instructions], using [semantic versioning] and the [repository CHANGELOG style], or I have commented below to indicate which documented exception this PR falls under[^1].
- [x] I updated/added any relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or I have commented below to indicate which [test exemption] this PR falls under[^1].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[^1]: Regular contributors who have demonstrated familiarity with the repository guidelines only need to comment if the PR is not auto-exempted by repo tooling.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[the auto-formatter]: https://github.com/flutter/packages/blob/main/script/tool/README.md#format-code
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[linked to at least one issue that this PR fixes]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[...]

PR #11151, opened 2026-03-02
[go_router_builder] Allow users to specify onExit as optional

hantrungkien:
Related to https://github.com/flutter/flutter/issues/183099

Allow users to specify onExit as optional

Needs https://github.com/flutter/packages/pull/11150 to get merged first

## Pre-Review Checklist

- [x] I signed the [CLA].


stuartmorgan-g:
Thanks for the contribution! You’ve checked boxes in the PR checklist above that are not reflected in this PR, so I’m assuming this is a work in progress and am marking it as a Draft. Please review the checklist, updating the PR as appropriate, and when the state of the PR as posted reflects the checklist please feel free to mark it as ready for review.

> * [x]  I added new tests to check the change I am making, or I have commented below to indicate which [test exemption](https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests) this PR falls under[1](#user-content-fn-1-84d1bd3a1f3ec8428c269eae2d348fcb).
> * [x]  All existing and new tests are passing.

hantrungkien:
@stuartmorgan-g Thank you for your quick response. I’ve added the missing items. Please take a look and review them for me!

Just a reminder that https://github.com/flutter/packages/pull/11150 needs to be merged first.

packages/go_router_builder/example/lib/on_exit_example.dart
```diff
@@ -24,7 +24,7 @@ class App extends StatelessWidget {
 @TypedGoRoute<HomeRoute>(
   path: '/',
   routes: <TypedGoRoute<GoRouteData>>[
-    TypedGoRoute<SubRoute>(path: 'sub-route'),
+    TypedGoRoute<SubRoute>(path: 'sub-route', overrideOnExit: true),
```
chunhtai: This api is not ideal since the customer may not know why the flag is here if they don't have the context.

I have several questions

1. looking at the code I am not sure why an onExit that always returns true will cause the bug in the original github issue, do you know why?
2. if we really want to make such change, we should consider make it default that it only conditionally pass in onExit if user does override instead of controlling by a flag that may be confusing for the customer

hantrungkien: @chunhtai Thank you for your response.
1. I think the issue is caused by this:
https://github.com/flutter/packages/blob/main/packages/go_router/lib/src/delegate.dart#L69

In my example [here](https://github.com/hantrungkien/go_router_builder_on_exit_issue/blob/main/lib/page/details_1_page.dart#L28C17-L32C37):


If I do it like this, I can return to HomePage:

```dart
final result = await Details2RouteData(
  $extra: Details2PageExtra(data: 'Test data 2'),
).push<String?>(context);
if (!context.mounted) return;
await WidgetsBinding.instance.endOfFrame;
if (!context.mounted) return;
context.pop(result);
```

However, the user will briefly see Details1Page before returning to Home. Therefore, I think a better solution would be to allow the default onExit parameter to be null, instead of always returning true.

2. `we should consider make it default that it only conditionally pass in onExit if user does override`

What do you think if I handle it like this?

[go_router_builder](https://github.com/flutter/packages/blob/main/packages/go_router_builder/lib/src/route_config.dart#L720)

```dart
        final bool hasOverriddenOnExit = classElement.methods.any(
          (method) => method.name == 'onExit',
        );
```

hantrungkien: @chunhtai I have updated the new solution. Please review it for me!

chunhtai (approved):
LGTM

stuartmorgan-g:
From triage: @chunhtai Please add a secondary reviewer when signing off on a community PR that doesn't already have a second review request.

Piinks:
@hannah-hyj can you take a look?

packages/go_router_builder/test_inputs/has_overridden_on_exit.dart.expect
hannah-hyj, line 1: in packages/go_router_builder/test_inputs/, can you add both tests cases for `has_overridden_on_exit.dart`  and `has_not_overridden_on_exit`

packages/go_router_builder/lib/src/route_config.dart
```diff
@@ -582,6 +603,7 @@ mixin $_mixinName on $routeDataClassName {
   String get routeConstructorParameters =>
       'path: ${escapeDartString(path)},'
       '${caseSensitive ? '' : 'caseSensitive: $caseSensitive,'}'
+      '${hasOverriddenOnExit ? 'hasOverriddenOnExit: $hasOverriddenOnExit,' : ''}'
```
hannah-hyj: if we only add "hasOverriddenOnExit: true," and omit "hasOverriddenOnExit: false,", 
it seems like it has conflict  with  https://github.com/flutter/packages/pull/11150/ which null means true?

packages/go_router_builder/test_inputs/has_overridden_on_exit.dart.expect
hantrungkien: @hannah-hyj I've updated. Please help me review.

packages/go_router_builder/lib/src/route_config.dart
hantrungkien: @hannah-hyj The null value in #1150 is only for backward-compatibility purposes (for ex, when using a newer version of `go_router` with an older version of `go_router_builder`, the behavior remains unchanged). But hasOverriddenOnExit here (#1151) cannot be null.

hannah-hyj: I understand hasOverriddenOnExit here is non-null boolean,
but 
1. when it's true, "hasOverriddenOnExit: true" is in routeConstructorParameters, 
so it generates 
```
     GoRouteData.$route(
          ...
      hasOverriddenOnExit: true,
    );
```
2.
when it's false , "hasOverriddenOnExit: false" is not in  routeConstructorParameters. 
so it generates 
```
     GoRouteData.$route(
          ...
    );
```
Correct me if I'm wrong but my question is that in the second case , according to  #1150, when `hasOverriddenOnExit` is omitted, it's null in go_router and it's treated as `true` instead of `false`.  so I think in second case in go_router_builder  it should generate 
```
     GoRouteData.$route(
          ...
      hasOverriddenOnExit: false,
    );
```

hantrungkien: @hannah-hyj You're right. My issue occurs when the project pins a go_router version that doesn't include #11150, but uses a go_router_builder version that includes #11151. This causes an error because routeConstructorParameters lacks the 'hasOverriddenOnExit' field. I'm looking for a workaround - do you have any suggestions?

hannah-hyj: > My issue occurs when the project pins a go_router version that doesn't include https://github.com/flutter/packages/pull/11150, but uses a go_router_builder version that includes https://github.com/flutter/packages/pull/11151.


I think if the pubspec.yaml in this PR is updated that go_router_builder that includes the changes from https://github.com/flutter/packages/pull/11151 will depend on  a go_router version that includes the changes in https://github.com/flutter/packages/pull/11150, you will be fine.  And we're doing batch release now so that go_router version should be the next version which is 17.3.0.  @chunhtai can you confirm that?

hantrungkien: @chunhtai please help me check.

packages/go_router_builder/pubspec.yaml
```diff
@@ -27,7 +27,7 @@ dev_dependencies:
   dart_style: ">=2.3.7 <4.0.0"
   flutter:
     sdk: flutter
-  go_router: ^17.1.0
+  go_router: ^17.2.0
```
hannah-hyj: This should be 17.3.0

hannah-hyj (approved):
LGTM except 1. there's conflict in CHANGELOG.md 2.  go_router version number in pubsub.ymal should be 17.3.0 instead of 17.2.0

hannah-hyj:
Hey @chunhtai , it looks like a significant number of files were changed after your initial approval. can you take another look at the latest commits to make sure you still approve this?

packages/go_router_builder/example/lib/has_not_overridden_on_exit_example.dart
```diff
@@ -0,0 +1,112 @@
+// Copyright 2013 The Flutter Authors
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// ignore_for_file: public_member_api_docs, unreachable_from_main
+
+import 'package:flutter/material.dart';
```
chunhtai: I don't think this example is helpful, the test cases should be good enough.

Can we remove this example?

packages/go_router_builder/lib/src/route_config.dart
```diff
@@ -717,10 +735,20 @@ abstract class RouteBaseConfig {
         }
         final ConstantReader nameValue = reader.read('name');
         final ConstantReader caseSensitiveValue = reader.read('caseSensitive');
+        final bool hasHasOverriddenOnExitParam =
+            classElement.supertype?.element.methods
+                .firstWhereOrNull((m) => m.name == r'$route')
+                ?.formalParameters
+                .any((p) => p.name == 'hasOverriddenOnExit') ??
```
chunhtai: can you remind me why we check this?

hantrungkien: @chunhtai 
@hannah-hyj's comment will probably be simpler than my explanation
https://github.com/flutter/packages/pull/11151#discussion_r3237178390

chunhtai: This is for backward compatibility ? is it still needed if we update the go_router dependency?

packages/go_router_builder/pubspec.yaml
hantrungkien: @chunhtai @hannah-hyj I've bumped.

packages/go_router_builder/example/lib/has_not_overridden_on_exit_example.dart
hantrungkien: @chunhtai I've removed.

packages/go_router_builder/lib/src/route_config.dart
hantrungkien: @chunhtai It is needed because `hasOverriddenOnExit` in go_router can be null || true || false.

chunhtai: I still don't get it

`hasHasOverriddenOnExitParam` keeps track whether `$route` has a named parameter `hasOverriddenOnExit`

 I assume this is to keep track whether the go_router support this named parameter? i.e. before 17.3.0.

If we pin ^17.3.0 we can assume this is always true I think

hantrungkien: @chunhtai  `If we pin ^17.3.0 we can assume this is always true I think` you are right. I've removed!

packages/go_router_builder/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 * Updates minimum supported SDK version to Flutter 3.38/Dart 3.10.
 
+## 4.4.0
```
chunhtai, line 1: please remove `## NEXT` section and move the item in 4.4.0 change log

```
* Updates minimum supported SDK version to Flutter 3.38/Dart 3.10.
```

chunhtai (commented):
besides the changelog comment, the rest lgtm

packages/go_router_builder/CHANGELOG.md
hantrungkien: @chunhtai I've updated

chunhtai (approved):
LGTM

hantrungkien:
@chunhtai please help me to re-run CICD

PR #11155, opened 2026-03-02
[camera] Add setImageQuality for JPEG compression control

Bolling88:
Adds `setJpegImageQuality(int quality)` (1–100) to the app-facing `camera` package, forwarding to the platform interface method.

This PR is now scoped down to `camera` only. The platform interface and implementations already landed and published separately:

| Package | Version | Landed in |
|---------|---------|-----------|
| `camera_platform_interface` | 2.13.0 | #11155 (split out) |
| `camera_platform_interface` | 2.13.1 | #12123 — default implementation made a no-op |
| `camera_android_camerax` | 0.7.4 | #11616 |
| `camera_avfoundation` | 0.10.2 | #11616 |

### This PR

| Package | Version |
|---------|---------|
| `camera` | 0.12.0+2 → 0.12.1 |

* Adds `CameraController.setJpegImageQuality`, with 1–100 range validation and `CameraException` translation.
* Bumps constraints to the published versions that implement it: `camera_platform_interface: ^2.13.1`, `camera_android_camerax: ^0.7.4`, `camera_avfoundation: ^0.10.2`. No `dependency_overrides`.
* Per the review discussion, the platform interface default is a no-op rather than an `UnimplementedError`, so calling this is always safe. The dartdoc says the setting is best-effort and links to https://github.com/flutter/flutter/issues/191790 for the current state of platform support (web and Windows are not implemented yet).

Fixes https://github.com/flutter/flutter/issues/183229

## Pre-Review Checklist

- [x] I signed the [CLA].


stuartmorgan-g:
Thanks for the contribution!

> ### Pre-Review Checklist

In the future, please use our checklist rather than replacing it with one you have invented. Our checklist is standardized for a reason.

> ## Related issues
> Fixes [flutter/flutter#60397](https://github.com/flutter/flutter/issues/60397)

Please provide the actual issue this is related to. That's not even a `camera` issue.

Bolling88:
> 
> > ### Pre-Review Checklist
> 
> In the future, please use our checklist rather than replacing it with one you have invented. Our checklist is standardized for a reason.
> 
> > ## Related issues
> > Fixes [flutter/flutter#60397](https://github.com/flutter/flutter/issues/60397)
> 
> Please provide the actual issue this is related to. That's not even a `camera` issue.

Sorry for that! My first contribution, and still learning!

Regarding the issue, I just took a placeholder and forgot to change it. This is rather a feature improvement than resolving an issue, but I found and linked a closed issue where one of the comments requested the possibility to set image quality. If that is not enough, can I just create a new issue?

stuartmorgan-g:
> This is rather a feature improvement than resolving an issue

As with almost every project, we track bug reports and feature requests in the same issue tracker.

> but I found and linked a closed issue where one of the comments requested the possibility to set image quality. If that is not enough, can I just create a new issue?

An issue that was fixed seven years ago does not provide an explanation of what current use case needs new APIs. [If there isn't an issue covering the use case you are trying to address](https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview), then you should file one.

Bolling88:
> > This is rather a feature improvement than resolving an issue
> 
> As with almost every project, we track bug reports and feature requests in the same issue tracker.
> 
> > but I found and linked a closed issue where one of the comments requested the possibility to set image quality. If that is not enough, can I just create a new issue?
> 
> An issue that was fixed seven years ago does not provide an explanation of what current use case needs new APIs. [If there isn't an issue covering the use case you are trying to address](https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview), then you should file one.

Thank you for your patience. I have created a new issue ticket that clearly explains the problem/feature.

packages/camera/camera/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.12.1
+
+* Adds `setImageQuality` for controlling JPEG compression quality.
```
hellohuanlin: is this function JPEG only? if so, better to put it in the function name

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -739,6 +740,7 @@ final class DefaultCamera: NSObject, Camera {
     let savePhotoDelegate = SavePhotoDelegate(
       path: path,
       ioQueue: photoIOQueue,
+      imageQuality: fileExtension == "jpg" && imageQuality < 100 ? imageQuality : nil,
```
hellohuanlin: what if file extension is `jpeg`? 
does `nil` quality mean 100% quality? 
can quality be a floating point number?

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
```diff
@@ -78,8 +87,49 @@ class SavePhotoDelegate: NSObject, AVCapturePhotoCaptureDelegate {
     didFinishProcessingPhoto photo: AVCapturePhoto,
     error: Error?
   ) {
-    handlePhotoCaptureResult(error: error) {
-      photo.fileDataRepresentation()
+    handlePhotoCaptureResult(error: error) { [weak self] in
+      guard let quality = self?.imageQuality, quality < 100 else {
+        return photo.fileDataRepresentation()
+      }
+      guard let originalData = photo.fileDataRepresentation() else {
```
hellohuanlin: it seems `fileDataRepresentation` is used in both case. can reduce duplicate code by putting it in the guard statement above

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
```diff
@@ -78,8 +87,49 @@ class SavePhotoDelegate: NSObject, AVCapturePhotoCaptureDelegate {
...
+        return nil
+      }
+      return Self.reencodeJPEG(data: originalData, quality: quality)
```
hellohuanlin: is it guaranteed to be JPEG format at this point? it is not clear from reading this chunk of code, without looking at the caller site.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
```diff
@@ -78,8 +87,49 @@ class SavePhotoDelegate: NSObject, AVCapturePhotoCaptureDelegate {
...
+    }
+
+    let metadata =
+      CGImageSourceCopyPropertiesAtIndex(source, 0, nil) as? [CFString: Any] ?? [:]
+
+    let mutableData = NSMutableData()
+    guard
+      let destination = CGImageDestinationCreateWithData(
+        mutableData as CFMutableData,
+        "public.jpeg" as CFString,
```
hellohuanlin: why hardcode filename? and also wanna double check `jpg` vs `jpeg` since you used both in this PR

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
```diff
@@ -78,8 +87,49 @@ class SavePhotoDelegate: NSObject, AVCapturePhotoCaptureDelegate {
...
+        1,
+        nil)
```
hellohuanlin: can you add comment on what these params are

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
```diff
@@ -78,8 +87,49 @@ class SavePhotoDelegate: NSObject, AVCapturePhotoCaptureDelegate {
...
+    else {
+      return data
     }
+
+    var properties = metadata
+    properties[kCGImageDestinationLossyCompressionQuality] = CGFloat(quality) / 100.0
```
hellohuanlin: it looks like quality can be a random floating point, and it doesn't have to be a multiple of 0.01? If so, a better API would be just to pass in a float, just like this Apple API.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
```diff
@@ -78,8 +87,49 @@ class SavePhotoDelegate: NSObject, AVCapturePhotoCaptureDelegate {
...
+
+    CGImageDestinationAddImage(destination, cgImage, properties as CFDictionary)
+
+    guard CGImageDestinationFinalize(destination) else {
+      return data
+    }
+
+    return mutableData as Data
```
hellohuanlin: Any reason why we can't use simpler APIs such as https://developer.apple.com/documentation/uikit/uiimage/jpegdata(compressionquality:)?

hellohuanlin (changes requested):

packages/camera/camera/CHANGELOG.md
Bolling88: The API is intentionally named `setImageQuality` rather than `setJpegQuality` for a couple of reasons:

1. It's consistent with Flutter's existing `image_picker` plugin which uses `imageQuality` with the same 1–100 int scale.
2. It's more future-proof — if quality control is ever extended to other formats (e.g. HEIF), the API name still makes sense.
3. From the app developer's perspective, they're setting the quality of the captured *image*, regardless of the underlying format.

The current implementation applies to JPEG only (HEIF passes through unmodified), but the abstraction level feels right for a cross-platform API.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
Bolling88: Good questions!

- **File extension**: The extension is always set by us in `captureToFile` (either `"heif"` or `"jpg"`), so `"jpeg"` won't occur. But I've changed the check to `fileExtension != "heif"` which is more robust and format-oriented.
- **nil quality**: `nil` means no quality override was requested — the original photo data is written as-is without re-encoding. This is the default path for users who never call `setImageQuality`.
- **Float**: The int 1–100 scale is consistent with Flutter's `image_picker` plugin (`imageQuality` parameter) and with Android's native `CaptureRequest.JPEG_QUALITY` (byte 0–100) and CameraX's `ImageCapture.Builder.setJpegQuality(int)`. We convert to float internally on iOS with `CGFloat(quality) / 100.0`. A cross-platform API should use a single consistent type, and int 1–100 is the natural fit for the Android side.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: Good catch! Refactored to call `fileDataRepresentation()` once at the top, then conditionally re-encode. Done in the latest commit.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: Yes — the caller (`DefaultCamera.captureToFile`) only passes a non-nil `imageQuality` when the format is JPEG (checked via `fileExtension != "heif"`). I've added an inline comment in the latest commit to make this guarantee visible without needing to check the caller.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: The \`\"public.jpeg\"\` string was the UTI for the output format passed to \`CGImageDestinationCreateWithData\`. I've replaced it with \`UTType.jpeg.identifier\` in the latest commit which is cleaner. 

Regarding \`jpg\` vs \`jpeg\` — the only place we use \`\"jpg\"\` is the file extension (set by us in \`captureToFile\`), and \`\"public.jpeg\"\` / \`UTType.jpeg\` is the standard UTI for the JPEG format. These are different things (file extension vs format identifier) so the difference is expected.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: Added doc comments to the `reencodeJPEG` method with `- Parameters:` section explaining both `data` and `quality`. Done in the latest commit.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: The int 1–100 scale was chosen to be consistent across all platforms:

- Android Camera2: `CaptureRequest.JPEG_QUALITY` uses a byte (0–100)
- Android CameraX: `ImageCapture.Builder.setJpegQuality(int)` uses int 1–100
- Flutter's own `image_picker` plugin: `imageQuality` uses int 0–100

Using a float would match the iOS-native API but would be inconsistent with the Android side and existing Flutter conventions. The conversion `CGFloat(quality) / 100.0` is straightforward and lossless for all int values in the 1–100 range.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: `UIImage.jpegData(compressionQuality:)` strips all EXIF metadata (GPS coordinates, camera info, orientation, etc.). For a camera plugin, preserving this metadata is important — users expect captured photos to retain location data, orientation, and other properties.

`CGImageDestination` lets us re-encode the pixels at a lower quality while copying all original metadata properties through `CGImageSourceCopyPropertiesAtIndex`. I've added a doc comment to `reencodeJPEG` explaining this rationale.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -739,6 +740,7 @@ final class DefaultCamera: NSObject, Camera {
     let savePhotoDelegate = SavePhotoDelegate(
       path: path,
       ioQueue: photoIOQueue,
+      imageQuality: fileExtension != "heif" && imageQuality < 100 ? imageQuality : nil,
```
LouiseHsu: hmm,  I dislike that `imageQuality` is doing any arbitrary filtering on the file extension at this layer. Could we maybe it handle it in the init?

LouiseHsu: also, can we guarantee its a .heif file at this point? how about .heic files?

LouiseHsu (changes requested):

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
Bolling88: Good points, both of you!

I've changed the check to use the `fileFormat` enum directly (`fileFormat == .jpeg`) instead of comparing extension strings. This avoids string-matching and makes the intent clear — and also addresses the `.heic` concern since we're no longer comparing file extensions at all.

The filtering stays at the caller level rather than moving into `SavePhotoDelegate`'s init because the delegate shouldn't need to know about file formats — it just receives optional quality and re-encodes if set. This keeps `SavePhotoDelegate` focused on the save/encode responsibility while the caller decides *when* re-encoding is appropriate.

Regarding `.heic` — the format decision is made earlier in this method via the `fileFormat` enum (line 718). The extension is always set by us: either `"heif"` or `"jpg"`. There are no `.heic` files in this flow.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/features/jpegquality/JpegQualityFeature.java
```diff
@@ -0,0 +1,53 @@
...
+
+import android.annotation.SuppressLint;
+import android.hardware.camera2.CaptureRequest;
+import androidx.annotation.NonNull;
+import io.flutter.plugins.camera.CameraProperties;
+import io.flutter.plugins.camera.features.CameraFeature;
+
+/** Controls the JPEG compression quality on the {@link android.hardware.camera2} API. */
+public class JpegQualityFeature extends CameraFeature<Integer> {
+  private int currentSetting = 100;
```
camsim99, line 15: Could `currentSetting` reflect the actual current setting by making the default `requestBuilder.get(CaptureRequest.JPEG_QUALITY)` instead of 100?

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/features/jpegquality/JpegQualityFeature.java
camsim99, line 1: Can you add a test for this feature? Something like [one of these tests](https://github.com/flutter/packages/tree/main/packages/camera/camera_android/android/src/test/java/io/flutter/plugins/camera/features).

packages/camera/camera_android_camerax/android/src/test/java/io/flutter/plugins/camerax/ImageCaptureTest.java
camsim99, line 1: Can you modify this test to make sure that the quality gets set as expected?

camsim99 (commented):
Reviewed `camera_android` and `camera_android_camerax`. Both LGTM overall, thanks for implementing this feature! Just left a couple of requests concerning a default value and some tests.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/features/jpegquality/JpegQualityFeature.java
Bolling88: Good question! The feature doesn't have access to a `CaptureRequest.Builder` at construction time — the builder is only passed to `updateBuilder()` during capture. Also, the feature is lazily created only when `setImageQuality` is first called, at which point `setValue()` is called immediately with the user's requested quality, so the default of 100 is never actually applied to the builder.

That said, 100 as the default makes sense semantically — it means "maximum quality" which is the expected behavior when quality hasn't been explicitly set.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/features/jpegquality/JpegQualityFeature.java
Bolling88: Done! Added `JpegQualityFeatureTest` with tests for:
- `getDebugName` returns correct name
- `getValue` returns 100 as default
- `getValue` echoes set value
- `checkIsSupported` returns true
- `updateBuilder` sets `JPEG_QUALITY` on the builder with the correct value
- `updateBuilder` uses default quality when not explicitly set

packages/camera/camera_android_camerax/android/src/test/java/io/flutter/plugins/camerax/ImageCaptureTest.java
Bolling88: Done! Added a new test `pigeon_defaultConstructor_setsJpegQualityWhenProvided` that creates an `ImageCapture` with `jpegQuality = 75` and verifies `imageCapture.getJpegQuality()` returns 75.

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/features/jpegquality/JpegQualityFeature.java
camsim99: SGTM, thanks for the clarity :)

camsim99 (approved):
Android implementations LGTM!

LouiseHsu (approved):
lgtm on the ios side! Thanks for making those changes!

LouiseHsu:
@Bolling88 looks like you still have some temporary dependency_overrides in the pubspec yamls - once you remove those we can probably merge :)

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
hellohuanlin: This seems to be quite brittle and could very well surprise us down the road. The `imageQuality` naming suggest that it could be extended to other formats such as HEIF, as you said previously, but in code we are making the assumption that it must be JPEG.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
hellohuanlin: Sorry i meant params of `CGImageDestinationCreateWithData` (e.g. what does `nil` mean?)

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
hellohuanlin: This is worth documenting, in case of future cleanup that "simplifies" it into a bad state. Could you add a comment?

hellohuanlin (changes requested):

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: Good point about brittleness. I've added a runtime JPEG format guard inside `reencodeJPEG` itself — it now checks `CGImageSourceGetType` and returns the original data unchanged if the source isn't JPEG. This way the method is self-contained and safe even if it's called with non-JPEG data in the future.

Also addressed the other two comments in the same commit:
- Added inline comments for all `CGImageDestinationCreateWithData` parameters (the `nil` is "no additional options")
- Added a comment on the `mutableData as Data` return explaining it contains the finalized JPEG after `CGImageDestinationFinalize`

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: Added inline comments for each parameter. The `nil` is documented as "no additional options" — Apple's API accepts an optional dictionary for type-specific options, but none are needed for basic JPEG encoding.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: Good call. Added a comment explaining that `mutableData` contains the finalized JPEG after `CGImageDestinationFinalize` and the cast to `Data` is intentional.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: Ah sorry, I misunderstood the first time! Added inline comments for all four `CGImageDestinationCreateWithData` parameters in the latest commit:
- `mutableData as CFMutableData` — output buffer for the encoded image
- `"public.jpeg" as CFString` — UTI for JPEG format
- `1` — imageCount: single image
- `nil` — no additional type-specific options (Apple's API accepts an optional dictionary here, but none are needed for basic JPEG encoding)

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: Done! Added a comment above the return explaining that `mutableData` contains the finalized JPEG after `CGImageDestinationFinalize`, so future readers won't be tempted to "simplify" it away.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
hellohuanlin: I'd recommend to move the branching logic to the caller, so that the helper method reencodeJPEG is more focused on just JPEG. 

```
switch type {
  case jpeg: 
    helper that compresses JPEG
   case _:
    return original image
}
```

Bolling88: Makes sense! Removed the format check from `reencodeJPEG` — it now focuses purely on JPEG re-encoding. The format branching is already handled by the caller (`DefaultCamera.captureToFile` only passes `imageQuality` for JPEG captures via the `fileFormat == .jpeg` check).

hellohuanlin: What you did here is in the exact same state discussed in my previous concern (https://github.com/flutter/packages/pull/11155/changes#r2962038676). I was talking about the immediate caller of reencodeJPEG.

Bolling88: Sorry about that! I misunderstood — you meant the immediate caller of `reencodeJPEG`, not the higher-level caller. 

Fixed: `photoOutput` now explicitly checks the data format using `CGImageSourceGetType` before calling `reencodeJPEG`. Non-JPEG data returns unchanged without entering the re-encoding path at all.

hellohuanlin (approved):

Bolling88:
Hi team, thanks for all the reviews and approvals! 🙏

Now that we have the needed approvals, what are the next steps for landing this? As I understand the federated plugin process:

1. `camera_platform_interface` (2.13.0) needs to be published to pub.dev first (currently 2.12.0)
2. Once published, I can remove all the `dependency_overrides` from the 8 pubspec.yaml files
3. Then the rest of the PR can be merged

Should I split the platform interface changes into a separate PR for landing first, or is there a preferred way to handle this? Happy to do whatever works best for the team.

stuartmorgan-g:
> Now that we have the needed approvals

The overall design and cross-platform code still need review by @bparrishMines before this proceeds.

> what are the next steps for landing this?

Once that happens, the next steps are to follow [the documented process](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changing-federated-plugins).

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
bparrishMines: I'm not sure if this is mentioned anywhere else, but is there a reason we cant set `AVCapturePhotoSettings`  with a jpeg quality. Something like:

```swift
let settings = AVCapturePhotoSettings(format: [
    AVVideoCodecKey: AVVideoCodecType.jpeg,
    AVVideoCompressionPropertiesKey: [
        AVVideoQualityKey: 0.9 // quality 0-1
    ]
])
```

packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
```diff
@@ -347,4 +347,11 @@ abstract class CameraPlatform extends PlatformInterface {
   Future<void> setImageFileFormat(int cameraId, ImageFileFormat format) {
     throw UnimplementedError('setImageFileFormat() is not implemented.');
   }
+
+  /// Sets the JPEG compression quality for still image capture.
+  ///
+  /// The [quality] must be between 1 (lowest) and 100 (highest).
+  Future<void> setImageQuality(int cameraId, int quality) {
```
bparrishMines: I think this would be a better fit in `createCameraWithSettings` because we already set video settings there. But it looks like we added `setImageFileFormat` as a method so this should probably follow the convention.

I worry this method clashes with `setImageFileFormat`. What happens if another file format is set when this method is called? Would this value be ignored without an error? It looks like all platforms use  jpeg and only iOS supports `setImageFileFormat` currently, but this could change.

I think this could be renamed to `setJpegImageQuality` and update the doc comments that this does nothing if the file format is not set to jpeg. Same for the method in the app-facing `CameraController`.

Bolling88: Great points, thank you for the thoughtful review!

**On naming:** Agreed — I've renamed the method from `setImageQuality` to `setJpegImageQuality` across all packages. This makes the contract explicit: the quality setting only applies to JPEG output. The updated doc comments now also state that this has no effect if the file format is not set to JPEG.

**On `createCameraWithSettings`:** I agree it would be a natural fit there, but since the existing `setImageFileFormat` was added as a standalone method, I followed the same pattern for consistency. Happy to move it if you feel strongly about it.

**On the clash with `setImageFileFormat`:** With the rename to `setJpegImageQuality`, this concern is largely resolved — the name itself signals that it's JPEG-specific. If the format is set to something other than JPEG, the quality value is simply not applied (documented in the method's doc comments).

stuartmorgan-g: @Bolling88 Please don't use AI to generate your review responses. More verbose is not better in review discussions.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
Bolling88: The re-encoding approach preserves EXIF metadata and works regardless of the capture pipeline format. AVCapturePhotoSettings quality only works when explicitly capturing as JPEG.

packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
Bolling88: Good call — renamed to setJpegImageQuality across all packages. Followed the setImageFileFormat pattern of using a standalone method rather than createCameraWithSettings.

stuartmorgan-g: @Bolling88 Deleting comments that people have already replied to and then posting new variants of them is very disruptive for anyone trying to follow the review. Also, comments can be edited, or at least they can by humans.

Are you using an agent to post review responses directly?

Bolling88: Ah sorry! I will keep that in mind. I formulated the responses myself, with the help of AI to fix them up, since I am not very articulate, and asked it to edit my existing comments above, but it seems it failed at that. I will make sure not to rely on it when posting new replies or at least editing existing ones.

stuartmorgan-g: > and asked it to edit my existing comments above, but it seems it failed at that

Please take a moment to review our new [AI contribution guidelines](https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#ai-contribution-guidelines) (which were added since this PR was opened). There should always be a human directly in the loop of any AI interaction with a PR.

> with the help of AI to fix them up, since I am not very articulate

FWIW, I do not know a single person on the Flutter team who would rather read AI-generated or AI-edited review comments than actual human thoughts as expressed by a human.

packages/camera/camera_avfoundation/ios/camera_avfoundation/Sources/camera_avfoundation/SavePhotoDelegate.swift
bparrishMines: Yes, but isn't the the re-encoding approach unnecessary if the output is already set to jpeg? If someone is setting the jpeg image quality, then I would assume they would want to output a jpeg image anyways.

Bolling88: Yeah, I guess we can safely assume that now since we also changed the method naming. Will fix!

packages/camera/camera/test/camera_test.dart
```diff
@@ -885,6 +885,83 @@ void main() {
       },
     );
 
+    test('setJpegImageQuality() calls $CameraPlatform', () async {
```
bparrishMines: nit: We don't include `$` in test names because it messed up analysis on some IDEs. I believe that this is still true for VS Code, but may be wrong.

```suggestion
    test('setJpegImageQuality() calls CameraPlatform', () async {
```

packages/camera/camera/test/camera_test.dart
```diff
@@ -885,6 +885,83 @@ void main() {
...
+
+      await cameraController.setJpegImageQuality(50);
+
+      verify(
+        CameraPlatform.instance.setJpegImageQuality(cameraController.cameraId, 50),
+      ).called(1);
+    });
+
+    test(
+      'setJpegImageQuality() throws $CameraException on $PlatformException',
```
bparrishMines: Same here.

```suggestion
      'setJpegImageQuality() throws CameraException on PlatformException',
```

bparrishMines (approved):
LGTM with just a few nits

I'll add the CICD label

Bolling88:
camera_platform_interface 2.13.0 is now published, so I've dropped the path override for
   it across all camera pubspecs and bumped the constraints to `^2.13.0`. The implementation package overrides      
  (camera_android_camerax, camera_avfoundation) stay in place until those new versions publish. Please have another look!

stuartmorgan-g:
Nobody can re-review this PR while it is failing the CLA check.

Also, could you please explain why you posted and deleted three almost identical versions of that comment, given [this discussion](https://github.com/flutter/packages/pull/11155#discussion_r3023763723)? Are you still using an agent to directly post comments here even after explicitly agreeing not to?

Bolling88:
> Nobody can re-review this PR while it is failing the CLA check.
> 
> Also, could you please explain why you posted and deleted three almost identical versions of that comment, given [this discussion](https://github.com/flutter/packages/pull/11155#discussion_r3023763723)? Are you still using an agent to directly post comments here even after explicitly agreeing not to?

CLA check fixed!

I doubt AI would fail that miserably 😅 I tried to post my comment via the command line, and it gave me a false negative 2 times, hence 3 comments!

bparrishMines:
@Bolling88 Since this was already reviewed. You should now create a new PR with only the platform implementation changes.

stuartmorgan-g:
From triage: This is just waiting on https://github.com/flutter/packages/pull/11616 to get final updates so that it can land, then this can be reconciled and landed.

stuartmorgan-g:
Once https://github.com/flutter/packages/pull/11616 autopublishes this can be reconciled down to just the app-facing package without dependency overrides, for final approval and landing.

Bolling88:
Reconciled down to just the app-facing camera package now that the platform implementations have been published. I also removed the dependency overrides and set the constraints to the versions that implement it

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -970,6 +970,20 @@ class CameraController extends ValueNotifier<CameraValue> {
...
 
+  /// Sets the JPEG compression quality for still image capture.
+  ///
+  /// The [quality] must be between 1 (lowest) and 100 (highest).
+  Future<void> setJpegImageQuality(int quality) async {
+    if (quality < 1 || quality > 100) {
+      throw ArgumentError.value(quality, 'quality', 'Must be between 1 and 100.');
+    }
+    try {
+      await CameraPlatform.instance.setJpegImageQuality(_cameraId, quality);
```
stuartmorgan-g, line 986: @bparrishMines I just looked at the platform interface PR and the default implementation throws an unimplemented error, but there's no support query API. That's not a valid combination for new APIs, because there's no way for a client to know if it's safe to call this method.

Either the default needs to be a no-op, or there needs to be a support query that's plumbed through this layer, with clear docs about checking before calling this.

stuartmorgan-g (changes requested):

packages/camera/camera/lib/src/camera_controller.dart
Bolling88: Went with the no-op default. Opened #[12123](https://github.com/flutter/packages/pull/12123) to change it in the platform interface. Once that lands and publishes I'll bump the constraint here to ^2.13.1 and add a doc note that it's best-effort and ignored where it's not supported. Sounds good?

bparrishMines: > @bparrishMines I just looked at the platform interface PR and the default implementation throws an unimplemented error, but there's no support query API. That's not a valid combination for new APIs, because there's no way for a client to know if it's safe to call this method.

I agree. I didn't want to add an API to query for it because I expected this to be a feature that could be implemented on each platform. (e.g. like adding an API query for `takePicture`). But I suppose this is the downside of not implementing it on all platforms initially. 

Assuming we don't have plans to implement on windows or web in the near future, an unimplemented error or a noop are neither ideal solutions. `UnimplementedError` would be annoying for windows/web users, while a noop could potentially lead to confusion.

I would prefer that we go with the noop and create an issue to add support for web and windows. Then add the link to the issue in the app-facing plugin so that users can check when support has been added. Therefore the documentation won't cause confusion if we forget to update it after adding support.

Piinks:
Hey @Bolling88 are you still planning to work on this change?

Bolling88:
> Hey @Bolling88 are you still planning to work on this change?

Yes, still on it.

I've rebased onto current main, so the merge conflict is gone, and scoped this PR down to just the app-facing `camera` package now that the platform interface and the implementations have published.

The two open review points are handled:

- The platform interface default is a no-op as of `camera_platform_interface` 2.13.1 (#12123), so calling this is always safe. Constraint here is now `^2.13.1`.
- I filed https://github.com/flutter/flutter/issues/191790 to track web and Windows support and linked it from the `setJpegImageQuality` dartdoc, as @bparrishMines asked for.

The diff is now 5 files in `packages/camera/camera` only, no dependency overrides.

This needs a re-review from @stuartmorgan-g to clear the changes-requested, and the CICD label for full presubmit.

stuartmorgan-g (approved):

stuartmorgan-g:
The `camera` min Dart and Flutter SDK versions will need to be updated to match the versions required by the latest `camera_android_camerax` to resolve the CI failure.

Bolling88:
Rebased onto latest main, so the branch is up to date now. The min SDK bump to Flutter 3.44/Dart 3.12 is in.

Bolling88:
The force-push dropped the `cicd` label, so presubmit hasn't run on 0550e16 yet — could someone re-add it? @bparrishMines

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -51,6 +51,9 @@ class CameraValue {
     this.isPreviewPaused = false,
     this.previewPauseOrientation,
     this.videoStabilizationMode = VideoStabilizationMode.off,
+    // An initializing formal isn't possible here: named parameters can't be
+    // private, and `isRecordingPaused` is a derived getter, not a stored field.
+    // ignore: prefer_initializing_formals
   }) : _isRecordingPaused = isRecordingPaused;
```
Piinks: Flyby comment while responding to CICD request: With private named parameters it can be, depends on what version of Dart this is on.

packages/camera/camera/pubspec.yaml
```diff
@@ -4,11 +4,11 @@ description: A Flutter plugin for controlling the camera. Supports previewing
   Dart.
 repository: https://github.com/flutter/packages/tree/main/packages/camera/camera
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+camera%22
-version: 0.12.0+2
+version: 0.12.1
 
 environment:
-  sdk: ^3.10.0
-  flutter: ">=3.38.0"
+  sdk: ^3.12.0
```
Piinks, line 10: Looks like yes!

packages/camera/camera/lib/src/camera_controller.dart
Bolling88: You're right, thanks. Switched to `required this._isRecordingPaused`, so the ignore is gone.

Rebased on main too, which dropped the CICD label again. Mind re-adding it? :)

Bolling88:
All green now. Ready for autosubmit whenever someone can add the label.

PR #11399, opened 2026-03-30
[ci] refactor: Cleanup tooling for handling arch build output <= Flutter v3.13.x

Gustl22:
There remained some code, which was left in to support build output for Flutter <= v3.16.x (see:  https://github.com/flutter/flutter/commit/792e26df9540cf2cf510ea0c11a53ba10a243424)
As the minimum supported Flutter version is 3.35.x as of https://github.com/flutter/packages/blob/eab1265529a0fe75e1f800ecf96c5012be155942/.ci/targets/repo_checks.yaml#L31
we can remove the special handling for versions prior v3.16.x.

Contributes to flutter/flutter#129807

## Pre-Review Checklist

- [x] I signed the [CLA].


Gustl22:
I suspect its

test-exempt: Remove code without adding functionality.

stuartmorgan-g:
test-exempt: is a test

(This test code isn't in a standard location that the bot recognizes as a test.)

> test-exempt: Remove code without adding functionality.

It does change functionality, and it's not just removing code, so that's not the case here.

stuartmorgan-g (changes requested):
> There remained some code, which was left in to support build output for Flutter <= v3.13.x.

What commit(s) exactly is this referring to? The only code I'm aware of that meets that description (paths without an arch) was already removed. https://github.com/flutter/flutter/pull/141930/ appears to be the earliest that this code could possibly be correct, and that a) didn't land until 3.19, and b) wasn't on `stable` at that point.

Based on the discussion in https://github.com/flutter/flutter/issues/176603 it doesn't appear that this new logic is currently correct for anything but `master` still. Am I missing something?

Gustl22:
> What commit(s) exactly is this referring to?

I was relating to the commit in this comment (whose issue I mentioned in the issue description):
https://github.com/flutter/flutter/issues/129807#issuecomment-1701911299

So the actual change was made here: https://github.com/flutter/flutter/commit/792e26df9540cf2cf510ea0c11a53ba10a243424, which was merged to stable for `3.16.0`

`flutter build windows` since then outputs `build\windows\x64\runner\Release\example.exe`, so there's no need any more to respect outputs without the `x64` arch dir.

stuartmorgan-g:
> I was relating to the commit in this comment (whose issue I mentioned in the issue description): [flutter/flutter#129807 (comment)](https://github.com/flutter/flutter/issues/129807#issuecomment-1701911299)

I'm confused; there is no commit in that comment.

> So the actual change was made here: [flutter/flutter@792e26d](https://github.com/flutter/flutter/commit/792e26df9540cf2cf510ea0c11a53ba10a243424), which was merged to stable for `3.16.0`
> 
> `flutter build windows` since then outputs `build\windows\x64\runner\Release\example.exe`, so there's no need any more to respect outputs without the `x64` arch dir.

And it already doesn't. That's what I was referring to when I said:

>> The only code I'm aware of that meets that description (paths without an arch) was already removed.

What part of this PR do you think is removing support for paths without an arch subdirectory?

Gustl22:
> I'm confused; there is no commit in that comment.

There are linked changes to a PR which was merged in the commit https://github.com/flutter/flutter/commit/792e26df9540cf2cf510ea0c11a53ba10a243424. Sorry, I thought that is obvious.

> What part of this PR do you think is removing support for paths without an arch subdirectory?

https://github.com/flutter/packages/pull/11399/changes#diff-bb0d4d38717be1c6c0cae870a81c95d238c7106871b1437a083a5c1debe0d469L715

Are the TODOs you made in the code back then still valid? What should I do differently?

Gustl22:
I must admit, I couldn't make it work locally to test it. Google testing is still relying on CMake <3.5. I couldn't reinstall VS 2019 (only 6 GB left). I tried upgrading to a newer googletest version: https://github.com/google/googletest/releases/download/v1.15.2/googletest-1.15.2.tar.gz but then a pointer issue. So I was hoping to simply test it via `CICD` label.

stuartmorgan-g:
> > What part of this PR do you think is removing support for paths without an arch subdirectory?
> 
> https://github.com/flutter/packages/pull/11399/changes#diff-bb0d4d38717be1c6c0cae870a81c95d238c7106871b1437a083a5c1debe0d469L715

Thanks, I missed that part among all the other changes. That and https://github.com/flutter/packages/pull/11399/changes#diff-0a053995b2f55ca2d7ada1ab35a6cf03b0f9e1f9c7a300aa4707c71e55195171L36 are indeed cruft that we missed when removing the other code that handled that case.

> Are the TODOs you made in the code back then still valid? What should I do differently?

The TODOs about 3.13 are obsolete and that code can be removed. As far as I can tell the TODOs about https://github.com/flutter/flutter/issues/129807 are still all correct, because that issue is still open, and the other issue I linked to suggests that's at least partially correct.

If you build a Flutter app on Windows on arm64 right now on stable, will it build an arm64 app? It doesn't sound like it, and if it's not, then the parts of this PR that make the tooling unconditionally assume that will happen ([example](https://github.com/flutter/packages/pull/11399/changes#diff-b525444e626a654538177afff3418a5da77038d43a58e3b52118523f87e9b7bfL449-R479)) are not correct, which is why I'm trying to understand what those changes are based on.

> So I was hoping to simply test it via `CICD` label.

We don't run this codepath on Windows arm64 in CI, so if you break the tooling for arm64 Windows the CI won't catch it.

Gustl22:
> If you build a Flutter app on Windows on arm64 right now on stable, will it build an arm64 app? It doesn't sound like it

Absolutely, I didn't think of that!

>  why I'm trying to understand what those changes are based on

I mistakenly mixed the removal of both TODOs, because they were added in the same commit https://github.com/flutter/packages/commit/b8b84b2304f00a3f93ce585cc7a30e1235bde7a0

> We don't run this codepath on Windows arm64 in CI, so if you break the tooling for arm64 Windows the CI won't catch it.

I see. Thanks for the hint!


I updated the code now, and added an additional TODO to refactor the Linux also after it supports arm. If anything should be in a separate PR, let me know.

script/tool/lib/src/native_test_command.dart
```diff
@@ -678,50 +678,35 @@ this command.
...
     String? arch;
-    const x64DirName = 'x64';
-    const arm64DirName = 'arm64';
     if (platform.isWindows) {
-      arch = _abi == Abi.windowsX64 ? x64DirName : arm64DirName;
+      arch = switch (_abi) {
+        Abi.windowsX64 => 'x64',
+        Abi.windowsArm64 => 'arm64',
+        _ => null,
+      };
```
Gustl22, line 688: I used this check, to be more explicit and perhaps its easier in the future to unify the handling for Linux and Windows in the same switch statement.

script/tool/lib/src/native_test_command.dart
```diff
@@ -678,50 +678,35 @@ this command.
...
       // CI configuration for the repository.
+      // See: https://github.com/flutter/flutter/issues/114349
       arch = 'x64';
     }
+
+    if (arch == null) {
+      return _PlatformResult(
+        RunState.failed,
+        error:
+            'Testing on platform and architecture "$_abi" as CMakeProjects is not supported.',
```
stuartmorgan-g, line 698: This should be `$platformName` rather than `platform`.

stuartmorgan-g, line 698: I'm not sure what "as CMakeProjects" means exactly. This should probably just say "Google Test testing is not supported on $platformName $_abi."

script/tool/lib/src/native_test_command.dart
```diff
@@ -678,50 +678,35 @@ this command.
...
+      final project = CMakeProject(
         example.directory,
         buildMode: buildMode,
         processRunner: processRunner,
         platform: platform,
         arch: arch,
       );
-      if (platform.isWindows) {
-        if (arch == arm64DirName && !project.isConfigured()) {
-          // Check for x64, to handle builds newer than 3.13, but that don't yet
```
stuartmorgan-g, line 700: Per the previous discussion, AFAICT removing this particular block is still incorrect. Only the next block where it checks for the pre-3.13 structure is obsolete.

stuartmorgan-g (changes requested):

script/tool/lib/src/native_test_command.dart
Gustl22: That was by purpose: `_abi.toString()` already includes the "Platform" name. This would just duplicate it (?): E.g. `Abi.windowsX64.toString()` gives `windows_x64`. There's no build in method to just get the architecture.

Gustl22: I changed it but see https://github.com/flutter/packages/pull/11399#discussion_r3143891616

script/tool/lib/src/native_test_command.dart
Gustl22: Thanks! I missed that it states x64 arch there, and that the comment was referring to the other issue!

Gustl22:
I removed the legacy build output tests because they were dependent on the architecture in `CMakeProject` and I guess they are obsolete also.
https://github.com/flutter/packages/pull/11399/changes#diff-9b83e4ab1a6685b63bf348b9211108901a496493ceea8d0d3d8063412187293dL2314-L2354

stuartmorgan-g (approved):
LGTM

stuartmorgan-g:
@robert-ancell for secondary review

robert-ancell (approved):
Looks like this will have no impact on Linux, so fine by me!

stuartmorgan-g:
@Gustl22 This needs to have the autoformatter run on the tool, then we can land it.

Gustl22:
Thank you! Sometimes I am cautious to enable the format on commit option, as it has made me sweat in the past. Also IntelliJ format and dart format sometimes also is off, but that's another story.

Thanks for reviewing, ready to land now @stuartmorgan-g

PR #11401, opened 2026-03-31
[Animations] Add support for custom shadows to OpenContainer (#62475)

### LINKED ISSUE #62475: OpenContainer add customize shadow support
For OpenContainer in animations, the contains can only set elevation, it's ugly in most cases, if custom shadow can be set would be better:
The grey is too dark

image

moepanda:
## Summary
This PR adds support for custom `BoxShadow` to the `OpenContainer` widget. This allows developers to specify `closedShadows` and `openShadows` to create more advanced and visually appealing transitions that go beyond the standard material `elevation`.
* Fixes a legacy layout overflow issue in `_ExampleSingleTile` by replacing the fixed height with responsive constraints, ensuring the example remains stable across different system font sizes and shadow distributions.

## Motivation
Currently, `OpenContainer` only supports a numeric `elevation` property, which limits modern UI designs requiring colored or multi-layered shadows. This PR also refactors the internal rendering hierarchy to move the shadow decoration outside the clipping surface, fixing a legacy issue where custom shadows were clipped by `Material`.

## Issues
Fixes flutter/flutter#62475

## Pre-Review Checklist
- [x] I read the [AI contribution guidelines] and understand my responsibilities.
- [x] I signed the [CLA]. (Reminder: Please ensure you've signed it!)
- [x] The title of the PR starts with `[animations]`
- [x] I linked to at least one issue that this PR fixes in the description above.
- [x] I followed [the version and CHANGELOG instructions].
- [x] I added new tests to check the change I am making.

### DIFF
--- a/packages/animations/lib/src/open_container.dart
+++ b/packages/animations/lib/src/open_container.dart
@@ -137,6 +137,8 @@ class OpenContainer<T extends Object?> extends StatefulWidget {
     this.useRootNavigator = false,
     this.routeSettings,
     this.clipBehavior = Clip.antiAlias,
+    this.closedShadows,
+    this.openShadows,
   });
 
   /// Background color of the container while it is closed.
@@ -299,6 +301,16 @@ class OpenContainer<T extends Object?> extends StatefulWidget {
   ///  * [Material.clipBehavior], which is used to implement this property.
   final Clip clipBehavior;
 
+  /// Custom shadows of the container while it is closed.
+  ///
+  /// If this is provided, [closedElevation] will be ignored.
+  final List<BoxShadow>? closedShadows;
+
+  /// Custom shadows of the container while it is open.
+  ///
+  /// If this is provided, [openElevation] will be ignored.
+  final List<BoxShadow>? openShadows;
+
   @override
   State<OpenContainer<T?>> createState() => OpenContainerState<T>();
 }
@@ -349,6 +361,8 @@ class OpenContainerState<T> extends State<OpenContainer<T?>> {
             transitionType: widget.transitionType,
             useRootNavigator: widget.useRootNavigator,
             routeSettings: widget.routeSettings,
+            closedShadows: widget.closedShadows,
+            openShadows: widget.openShadows,
           ),
         );
     if (widget.onClosed != null) {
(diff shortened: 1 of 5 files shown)


packages/animations/lib/src/open_container.dart
```diff
@@ -362,16 +376,22 @@ class OpenContainerState<T> extends State<OpenContainer<T?>> {
...
-          child: Builder(
-            key: _closedBuilderKey,
-            builder: (BuildContext context) {
-              return widget.closedBuilder(context, openContainer);
-            },
+        child: DecoratedBox(
+          decoration: ShapeDecoration(
+            shape: widget.closedShape,
+            shadows: widget.closedShadows,
+          ),
```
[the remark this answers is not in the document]
moepanda: Done. By defining local variables, we avoided redundant declarations of the Material widget and standardized the code format.

hannah-hyj (approved):
LGTM, thank you for fixing the layout overflow issue too!

moepanda:
> LGTM, thank you for fixing the layout overflow issue too!

Thank you! Glad to help.

stuartmorgan-g:
From triage: @hannah-hyj Please add a second reviewer familiar with the package, so this can get the secondary review for landing.

packages/animations/lib/src/open_container.dart
```diff
@@ -358,22 +372,32 @@ class OpenContainerState<T> extends State<OpenContainer<T?>> {
 
   @override
   Widget build(BuildContext context) {
+    final Widget material = Material(
+      clipBehavior: widget.clipBehavior,
+      color: widget.closedColor,
+      elevation: widget.closedShadows == null ? widget.closedElevation : 0.0,
```
QuncCccccc, line 378: Just for my learning, why do we need to set elevation to 0.0 when closedShadows is non-null.

packages/animations/lib/src/open_container.dart
```diff
@@ -822,6 +872,64 @@ class _OpenContainerRoute<T> extends ModalRoute<T> {
           assert(scrimTween != null);
 
           final Rect rect = _rectTween.evaluate(curvedAnimation)!;
+          final Widget material = Material(
+            clipBehavior: Clip.antiAlias,
+            animationDuration: Duration.zero,
+            color: colorTween!.evaluate(animation),
+            shape: _shapeTween.evaluate(curvedAnimation),
+            elevation: _elevationTween.evaluate(curvedAnimation),
```
QuncCccccc, line 880: do we need to check currentShadows here like above?

packages/animations/test/open_container_test.dart
```diff
@@ -355,6 +355,106 @@ void main() {
     expect(dataClosed.rect, dataTransitionDone.rect);
   });
 
+  testWidgets('Custom shadows work', (WidgetTester tester) async {
```
QuncCccccc, line 358: Can we add a test case for when one of the shadow properties is `null` and the other is provided? This would ensure the fallback to elevation behavior is verified and doesn't break in the future.

QuncCccccc (commented):
Thank you for your contribution! Overall looks good! Just left one comment for adding an additional unit test.

packages/animations/lib/src/open_container.dart
moepanda: We set elevation to 0.0 so the shadow comes from a single source when custom shadows are provided.
In this case the visual shadow is drawn by the wrapping DecoratedBox, so keeping Material.elevation would stack the framework shadow on top of the custom BoxShadows.
The same rule is applied in the fully-open state as well.

packages/animations/lib/src/open_container.dart
moepanda: This is intentional. The elevation tween already accounts for whether custom shadows are provided on either side, so mixed cases can smoothly transition between Material elevation and custom shadows. If we zeroed elevation whenever `currentShadows` is non-null, the transition would become abrupt in those mixed cases.

packages/animations/test/open_container_test.dart
moepanda: Good point, I added mixed null/non-null shadow test coverage for both directions to lock in the elevation fallback behavior during the transition.

packages/animations/lib/src/open_container.dart
QuncCccccc: I see, it makes sense! Thanks for the explanation.

QuncCccccc (approved):
LGTM. Thank you!

QuncCccccc:
Seems there are some format issues that need to fix.

Screenshot 2026-04-22 at 5 18 38 PM

moepanda:
> Seems there are some format issues that need to fix.
>

PTAL

QuncCccccc (approved):
LGTM.

PR #11433, opened 2026-04-04
[google_fonts] Add google_fonts_lite file to allow tree-shaking of the other huge files

TheCarpetMerchant:
Adds a `google_fonts_lite.dart` generated file, which contains a map of `fontFamily` to fonts variants and a `getFont` function that simply forwards the call to the internal `googleFontsTextStyle` function which does the downloading and loading of the font.

https://github.com/flutter/flutter/issues/184337

I have not done documentation, testing etc because this PR is for discussion purposes. I'd like approval of the concept before doing the paperwork.

The idea is to provide a _lite_ version of the `GoogleFonts` class. This lite version does not depend on any of the other generated code, which is absolutely huge and takes up a massive amount of space (8MB on android for v8.0.0). By using this lite version only, the compiler is able to tree-shake all that additionnal code while you retain the ability to invoke any font from the package via `getFont`.

For me this brings the package's footprint from 8MB/target platform to less than a 1MB/target when building a release apk.


Piinks (commented):
@guidezpl, correct me if I am wrong, the generator script here would need to integrate with the internal dashboard where we pull these fonts from? Thanks in advance. :)


I appreciate the contribution @TheCarpetMerchant! We need to confirm if this will work with the existing workflows we have for this package.

guidezpl:
> @guidezpl, correct me if I am wrong, the generator script here would need to integrate with the internal dashboard where we pull these fonts from? Thanks in advance. :)
> 
> I appreciate the contribution @TheCarpetMerchant! We need to confirm if this will work with the existing workflows we have for this package.

Before this PR, the generator finds the font directory to generate the Dart code, and this PR seems to do nothing different in that regard, so from that point of view, this PR is okay

Piinks (commented):
Thanks for your patience while I reviewed this very carefully. This is an excellent addition. Bringing the package footprint down from 8MB to <1MB is a massive win for the ecosystem, especially for web and size-constrained mobile applications. 🎊 

While the current results are impressive, I noticed an architectural 'leak' in the dependency graph that we should address to prevent future regressions.

Currently, `google_fonts_lite.dart` imports `google_fonts_base.dart`, which in turn imports `../google_fonts.dart`. The main `google_fonts.dart` file exports `google_fonts_all_parts.dart`, where the 'heavy' `GoogleFonts` class and its `asMap()` method live.

The likely reason you are seeing size savings now is that the compiler is smart enough to see that `GoogleFonts.asMap()` is never called when using the Lite path. However, this is a fragile dependency. Because `google_fonts_base.dart` still has a direct import path to the heavy library (just to access the `Config` instance), any future change to the `GoogleFonts` class, such as a static initializer, a new field, or a change in how `Config` is accessed, could accidentally 'trap' the compiler into pulling all 8MB of font parts back into the binary without warning.

To make this optimization robust and permanent, we should decouple the configuration from the font metadata:

1.  Move the `Config` class and the `GoogleFonts.config` static instance into a new, standalone internal file (e.g., `lib/src/google_fonts_config.dart`).
2.  Update `google_fonts_base.dart` to import this new config file instead of the heavy `google_fonts.dart` entry point.
3.  Ensure the generator is updated to reflect this new structure.

Given that the `Config` refactor is a foundational change, it might be cleanest to handle the extraction of `Config` in a separate, small PR first. Once the configuration is decoupled, this PR can be rebased to provide a truly 'Lite' path that is cryptographically sealed off from the heavy parts files.

WDYT @TheCarpetMerchant? :)

TheCarpetMerchant:
> The likely reason you are seeing size savings now is that the compiler is smart enough to see that `GoogleFonts.asMap()` is never called when using the Lite path. However, this is a fragile dependency. Because `google_fonts_base.dart` still has a direct import path to the heavy library (just to access the `Config` instance), any future change to the `GoogleFonts` class, such as a static initializer, a new field, or a change in how `Config` is accessed, could accidentally 'trap' the compiler into pulling all 8MB of font parts back into the binary without warning.
> 
> To make this optimization robust and permanent, we should decouple the configuration from the font metadata:
> 
>     1. Move the `Config` class and the `GoogleFonts.config` static instance into a new, standalone internal file (e.g., `lib/src/google_fonts_config.dart`).
> 
>     2. Update `google_fonts_base.dart` to import this new config file instead of the heavy `google_fonts.dart` entry point.
> 
>     3. Ensure the generator is updated to reflect this new structure.
> 

Sure, makes sense to make it more robust. Here's a PR where each part of the library (the full part and the lite part) has its own instance of the config stored as a static variable within their own file (although the lite version doesn't exist yet lol) : https://github.com/flutter/packages/pull/11602

Edit : I didn't keep a single static instance because that would have meant breaking existing user code by moving the instance outside of the GoogleFonts class and into the GoogleFontsConfig class. This would be a huge annoyance for users for no real benefit as users of the lite version will only use that one, and users of the main one won't see the difference.

Piinks:
Following up in #11602 👍

stuartmorgan-g:
From triage: This is blocked on the PR referenced above, right?

Piinks:
> From triage: This is blocked on the PR referenced above, right?

Yup! After that lands we can move this forward.

Piinks:
The change to extract the config class has landed, if you would rebase here, we can finally move this forward! Thanks!

TheCarpetMerchant:
> The change to extract the config class has landed, if you would rebase here, we can finally move this forward! Thanks!

Alright, I've rebased and updated the changelog, readme + bump to 8.3.0. Squashed it to one commit.
The last thing to do would be to add a test for this, but since it requires running a full build I'm unsure of how that could be put in place with regards to CI.

packages/google_fonts/generator/google_fonts_lite.tmpl
```diff
@@ -0,0 +1,26 @@
...
+
+import 'google_fonts_base.dart';
+import 'google_fonts_descriptor.dart';
+import 'google_fonts_variant.dart';
+
+// Hosts a Map<String, Map<GoogleFontsVariant, GoogleFontsFile>> where the String is the font's name.
+// That is used to allow tree-shaking to remove all of the _parts files.
+// If you only call GoogleFontsLite.fontsMap or GoogleFontsLite.loadFont(),
+// the code in the GoogleFonts class and its parts classes is never called.
+// WHat is called is googleFontsTextStyle, which attempts to load a font if not already loaded.
```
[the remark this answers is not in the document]
Piinks: ++

packages/google_fonts/generator/google_fonts_lite.tmpl
```diff
@@ -0,0 +1,26 @@
...
+class GoogleFontsLite {
+  static Map<String, Map<GoogleFontsVariant, GoogleFontsFile>> fontsMap = {
+    {{#fontEntries}}
+      '{{fontFamily}}': {
+      {{#fontUrls}}
+      const GoogleFontsVariant(fontWeight: FontWeight.w{{variantWeight}}, fontStyle: FontStyle.{{variantStyle}},) : const GoogleFontsFile('{{hash}}', {{length}},),
+      {{/fontUrls}}
+      },
+    {{/fontEntries}}
+  };
```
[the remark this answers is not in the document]
Piinks: ++ This is a good suggestion.

packages/google_fonts/generator/google_fonts_lite.tmpl
```diff
@@ -0,0 +1,26 @@
...
+  
+  static TextStyle getFont(String fontFamily) {
+    return googleFontsTextStyle(fontFamily: fontFamily, fonts: fontsMap[fontFamily]!);
+  }
```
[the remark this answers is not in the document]
Piinks: This is also a good suggestion.

Separate from that `GoogleFontsLite.getFont` currently only accepts `String fontFamily`, omitting `textStyle`, `fontSize`, `color`, `fontWeight`, `fontStyle`, etc. We should accept and forward all optional parameters to `googleFontsTextStyle` so `GoogleFontsLite.getFont` shares the same signature as `GoogleFonts.getFont`.

Piinks (commented):
Hi @TheCarpetMerchant, thanks for rebasing this after #11602 landed!

To answer your question regarding testing:

You do not need to set up a full `flutter build apk` release build in CI. Package unit tests in this repository run under `flutter test`. 

To satisfy the testing requirement, please add a unit test file `test/google_fonts_lite_test.dart` that tests `GoogleFontsLite` in scenarios like:
- `GoogleFontsLite.getFont('Lato')` returns a `TextStyle` with the expected font family.
- Styling options (`fontSize`, `color`, `fontWeight`, etc.) are forwarded correctly to the returned `TextStyle`.
- Requesting an unknown font family throws a descriptive error.
- `GoogleFontsLite.fontsMap` is populated.

Thanks!

TheCarpetMerchant:
> Hi @TheCarpetMerchant, thanks for rebasing this after #11602 landed!
> 
> To answer your question regarding testing:
> 
> You do not need to set up a full `flutter build apk` release build in CI. Package unit tests in this repository run under `flutter test`.
> 
> To satisfy the testing requirement, please add a unit test file `test/google_fonts_lite_test.dart` that tests `GoogleFontsLite` in scenarios like:
> 
>     * `GoogleFontsLite.getFont('Lato')` returns a `TextStyle` with the expected font family.
> 
>     * Styling options (`fontSize`, `color`, `fontWeight`, etc.) are forwarded correctly to the returned `TextStyle`.
> 
>     * Requesting an unknown font family throws a descriptive error.
> 
>     * `GoogleFontsLite.fontsMap` is populated.
> 
> 
> Thanks!

Done. I've also added the LICENSE mention at the top of the file. If everything looks good I'll make a squash and I suggest you run the generator and subsequently make the release, as I've noticed there's new fonts available.
If you don't plan on updating the fonts I can make the pubspec change. Cheers

packages/google_fonts/generator/google_fonts_lite.tmpl
```diff
@@ -0,0 +1,86 @@
...
+class GoogleFontsLite {
+  static Map<String, Map<GoogleFontsVariant, GoogleFontsFile>> fontsMap = {
+    {{#fontEntries}}
+      '{{fontFamily}}': {
+      {{#fontUrls}}
+      const GoogleFontsVariant(fontWeight: FontWeight.w{{variantWeight}}, fontStyle: FontStyle.{{variantStyle}},) : const GoogleFontsFile('{{hash}}', {{length}},),
+      {{/fontUrls}}
+      },
+    {{/fontEntries}}
+  };
```
Piinks: A few more improvements for the class declaration and map:
1. `abstract final class`: Declaring `GoogleFontsLite` as an `abstract final class` prevents instantiation or subclassing of this utility namespace.
2. `static const fontsMap`: Making `fontsMap` a compile-time `static const` prevents allocating a 55,000-line mutable map at runtime and ensures optimal memory usage and compiler constant folding.
3. `fontFamilyDisplay` key: Keying the map by `{{fontFamilyDisplay}}` (e.g. `'Open Sans'`) rather than `{{fontFamily}}` (`'OpenSans'`) ensures font lookup matches the official Google Fonts family names and matches `GoogleFonts.getFont(...)`.
4. Also, `//` comments should be converted to `///` dartdoc comments so they appear on pub.dev, and adding breadcrumbs for references like `loadFont()` to `getFont()`.

```suggestion
/// Provides a lightweight entry point to [TextStyle]s from Google Fonts.
///
/// Unlike the main `GoogleFonts` class which declares dedicated methods for
/// each font, [GoogleFontsLite] provides dynamic font lookup via [getFont]
/// and [fontsMap], allowing unused font methods to be tree-shaken by the compiler.
abstract final class GoogleFontsLite {
  /// Map of all available Google Fonts families to their variant file descriptors.
  static const Map<String, Map<GoogleFontsVariant, GoogleFontsFile>> fontsMap = {
    {{#fontEntries}}
      '{{fontFamilyDisplay}}': {
      {{#fontUrls}}
      const GoogleFontsVariant(fontWeight: FontWeight.w{{variantWeight}}, fontStyle: FontStyle.{{variantStyle}},) : const GoogleFontsFile('{{hash}}', {{length}},),
      {{/fontUrls}}
      },
    {{/fontEntries}}
  };
```

packages/google_fonts/generator/google_fonts_lite.tmpl
```diff
@@ -0,0 +1,86 @@
...
+    double? decorationThickness,
+  }) {
+    if (!fontsMap.containsKey(fontFamily)) {
+      throw Exception("No font family by name '$fontFamily' was found.");
+    }
+
+    return googleFontsTextStyle(
+      fontFamily: fontFamily,
+      fonts: fontsMap[fontFamily]!,
+      textStyle: textStyle,
```
Piinks, line 68: When lookups are done using the display name (e.g. `'Open Sans'`), `fontsMap[fontFamily]!` finds the variant map. However, `googleFontsTextStyle` expects `fontFamily` without spaces (e.g. `'OpenSans'`) so that asset path prefixing (`toApiFilenamePrefix()`) and fallback font family naming match the behavior of `GoogleFonts.openSans()`.

Passing `fontFamily: fontFamily.replaceAll(' ', '')` to `googleFontsTextStyle` ensures full parity with `GoogleFonts.getFont`.

```suggestion
    if (!fontsMap.containsKey(fontFamily)) {
      throw Exception("No font family by name '$fontFamily' was found.");
    }

    return googleFontsTextStyle(
      fontFamily: fontFamily.replaceAll(' ', ''),
      fonts: fontsMap[fontFamily]!,
      textStyle: textStyle,
```

packages/google_fonts/lib/google_fonts.dart
```diff
@@ -4,3 +4,4 @@
```
Piinks, line 4: Similar to this file, consider adding a dedicated entry point `lib/google_fonts_lite.dart`:

```dart
// found in the LICENSE file.

export 'src/google_fonts_config.dart';
export 'src/google_fonts_lite.dart';
```

This gives developers aiming for minimal binary size a clean public library to import (`import 'package:google_fonts/google_fonts_lite.dart';`) without having to import `google_fonts.dart` (which exports `google_fonts_all_parts.dart`) or importing from `src/`.

packages/google_fonts/test/google_fonts_lite_test.dart
```diff
@@ -0,0 +1,92 @@
+// Copyright 2013 The Flutter Authors
```
Piinks, line 1: Here are a few updates for the test file to align with repository standards:
- Import `package:google_fonts/google_fonts.dart` rather than importing implementation files from `src/`.
- Remove unused `package:flutter/cupertino.dart`.
- Clean up dead commented-out `foreground` code.
- Use `expect(..., throwsA(...))` instead of a manual `try/catch` block.
- Add a test verifying lookups for fonts with spaces in their family name (like `'Open Sans'`).
- Verify that `GoogleFontsLite.fontsMap.keys` matches `GoogleFonts.asMap().keys` exactly.

Some additional tests to include:

<details>

```dart
// found in the LICENSE file.

import 'package:google_fonts/google_fonts.dart';

void main() {
  testWidgets('GoogleFontsLite getFont returns the correct font with the given parameters', (WidgetTester tester) async {
    const textStyle = TextStyle(
      color: Color(0xAABBCCDD),
      fontSize: 20,
      letterSpacing: 20,
      wordSpacing: 20,
      height: 20,
      decorationThickness: 20,
      fontWeight: FontWeight.w300,
      fontStyle: FontStyle.italic,
      textBaseline: TextBaseline.alphabetic,
      locale: Locale('fr'),
      background: Paint()..color = const Color(0xAABBCCDD),
      shadows: <Shadow>[Shadow(blurRadius: 1)],
      fontFeatures: <FontFeature>[FontFeature.slashedZero()],
      decoration: TextDecoration.lineThrough,
      decorationColor: Color(0xAABBCCDD),
      decorationStyle: TextDecorationStyle.dashed,
    );

    void verifyTextStyle(TextStyle res) {
      expect(res.fontFamily, equals('Lato_300italic'));
      expect(res.color, equals(textStyle.color));
      expect(res.fontSize, equals(textStyle.fontSize));
      expect(res.letterSpacing, equals(textStyle.letterSpacing));
      expect(res.wordSpacing, equals(textStyle.wordSpacing));
      expect(res.decorationThickness, equals(textStyle.decorationThickness));
      expect(res.fontWeight, equals(textStyle.fontWeight));
      expect(res.fontStyle, equals(textStyle.fontStyle));
      expect(res.textBaseline, equals(textStyle.textBaseline));
      expect(res.locale, equals(textStyle.locale));
      expect(res.background, equals(textStyle.background));
      expect(res.shadows, equals(textStyle.shadows));
      expect(res.fontFeatures, equals(textStyle.fontFeatures));
      expect(res.decoration, equals(textStyle.decoration));
      expect(res.decorationColor, equals(textStyle.decorationColor));
      expect(res.decorationStyle, equals(textStyle.decorationStyle));
    }

    verifyTextStyle(GoogleFontsLite.getFont(
      'Lato',
      textStyle: textStyle,
    ));

    verifyTextStyle(GoogleFontsLite.getFont(
      'Lato',
      color: textStyle.color,
      fontSize: textStyle.fontSize,
      letterSpacing: textStyle.letterSpacing,
      wordSpacing: textStyle.wordSpacing,
      height: textStyle.height,
      decorationThickness: textStyle.decorationThickness,
      fontWeight: textStyle.fontWeight,
      fontStyle: textStyle.fontStyle,
      textBaseline: textStyle.textBaseline,
      locale: textStyle.locale,
      background: textStyle.background,
      shadows: textStyle.shadows,
      fontFeatures: textStyle.fontFeatures,
      decoration: textStyle.decoration,
      decorationColor: textStyle.decorationColor,
      decorationStyle: textStyle.decorationStyle,
    ));
  });

  testWidgets('GoogleFontsLite getFont supports fonts with spaces in their family name', (WidgetTester tester) async {
    final TextStyle style = GoogleFontsLite.getFont('Open Sans');
    expect(style.fontFamily, equals('OpenSans_regular'));
    expect(style.fontFamilyFallback, equals(<String>['OpenSans']));
  });

  testWidgets('GoogleFontsLite getFont throws an exception when the fontFamily is unknown', (WidgetTester tester) async {
    expect(
      () => GoogleFontsLite.getFont('ZZZ_NON_EXISTENT_FONT_ZZZ'),
      throwsA(
        isA<Exception>().having(
          (Exception e) => e.toString(),
          'message',
          contains("No font family by name 'ZZZ_NON_EXISTENT_FONT_ZZZ' was found."),
        ),
      ),
    );
  });

  test('GoogleFontsLite.fontsMap keys match GoogleFonts.asMap keys exactly', () {
    expect(GoogleFontsLite.fontsMap.keys, equals(GoogleFonts.asMap().keys));
  });
}
```

packages/google_fonts/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 8.3.0
+
+- Add a `GoogleFontsLite` class. Using only this class and not the regular `GoogleFonts` allows the compiler to tree-shake most of the package's code.
```
Piinks: Following the [repository CHANGELOG guidelines](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style) (third-person present tense verb, full sentence):

```suggestion
## 8.3.0

* Adds the `GoogleFontsLite` class to allow tree-shaking unused font code.
```

Piinks (commented):
Thanks for continuing to iterate on this, @TheCarpetMerchant! This is going to be a substantial improvement for app binary size.

Before landing, there are a few important issues to address, commented inline below.

You should be able to run the generator from packages/google_fonts/, running:
`dart generator/generator.dart`

TheCarpetMerchant:
The map can't be const because GoogleFontsVariant overrides '==' and hashCode so we can't make the map const. I've made it final though.
Rest has been added and squashed, I ran the generator & updated the changelog as well.

Piinks:
> The map can't be const because GoogleFontsVariant overrides '==' and hashCode so we can't make the map const. I've made it final though. Rest has been added and squashed, I ran the generator & updated the changelog as well.

The `const` issue causing static analysis to fail is in `generator/google_fonts_part.tmpl`. Because `GoogleFontsFile`'s constructor was made `const`, the template for the 26 part files needs `const` added before `GoogleFontsFile`:

    ```diff
    // packages/google_fonts/generator/google_fonts_part.tmpl:44
    - const GoogleFontsVariant(fontWeight: FontWeight.w{{variantWeight}}, fontStyle:
  FontStyle.{{variantStyle}},): GoogleFontsFile('{{hash}}', {{length}},),
    + const GoogleFontsVariant(fontWeight: FontWeight.w{{variantWeight}}, fontStyle:
  FontStyle.{{variantStyle}},): const GoogleFontsFile('{{hash}}', {{length}},),
```

And adding const to the test fixtures in:

- test/google_fonts_text_style_test.dart (line 36)
- test/load_font_if_necessary_test.dart (lines 71, 75)
- test/load_font_if_necessary_with_local_fonts_test.dart (line 58)

After re-running dart generator/generator.dart, dart run script/tool/bin/flutter_plugin_tools.dart analyze --packages google_fonts will pass cleanly.

**Comment by TheCarpetMerchant** [comment]
I purposefully didn't want to touch the other files, but it makes sense to want to keep the analysis at 100% green. I'll keep that in mind next time.
I can't run the analysis cmd, but the VSCode plugin isn't reporting anything now and I suppose it's the same tool running that.

File: packages/google_fonts/CHANGELOG.md [hunk 9]
```diff
@@ -1,3 +1,46 @@
+## 8.3.0
```
Piinks: Removing fonts is a breaking change, so this will need to be a bump to 9.0.0. 
That being said, there is another PR migrating this to material_ui (#12489), which is also a breaking change. Rather than release 2 major versions, we'll aim to batch these changes together for one 9.0.0 release.
Let's make this "Next", and in pubspec.yaml, set `publish_to: none`.

TheCarpetMerchant: Done

packages/google_fonts/pubspec.yaml
```diff
@@ -3,6 +3,7 @@ description: A Flutter package to use fonts from fonts.google.com. Supports HTTP
 repository: https://github.com/flutter/packages/tree/main/packages/google_fonts
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+google_fonts%22
 version: 8.2.1
+publish_to: none
```
Piinks, line 6: Perfect, thank you

Piinks (commented):
It looks like the formatter is failing here, but otherwise is nearly ready to land. Can you run formatting?

TheCarpetMerchant:
> It looks like the formatter is failing here, but otherwise is nearly ready to land. Can you run formatting?

Right, didn't run it on the test file. And there was a line in the generator file. Done

Piinks (commented):
This LGTM, let me send it out for secondary review

Piinks (approved):

packages/google_fonts/generator/google_fonts_lite.tmpl
```diff
@@ -0,0 +1,87 @@
...
+    Paint? background,
+    List<ui.Shadow>? shadows,
+    List<ui.FontFeature>? fontFeatures,
+    TextDecoration? decoration,
+    Color? decorationColor,
+    TextDecorationStyle? decorationStyle,
+    double? decorationThickness,
+  }) {
+    if (!fontsMap.containsKey(fontFamily)) {
+      throw Exception("No font family by name '$fontFamily' was found.");
```
elliette, line 60: I think this error could be shortened to simply "Font family $fontFamily not found."

packages/google_fonts/generator/google_fonts_lite.tmpl
elliette: Follow-up to this: could we add a comment above `fontFamily.replaceAll(' ', '')` to document this? (e.g. `// googleFontsTextStyle expects the font family name to not include spaces.`

packages/google_fonts/generator/google_fonts_lite.tmpl
TheCarpetMerchant: This was already the format used in other places in the package. I wouldn't change it in case people are treating this exception specifically (since it's there's no type they would have to rely on the content).

elliette (approved):
LGTM, thank you!

PR #9770, opened 2025-08-07
[path_provider_android] Changes internal implementation to use JNI

bparrishMines:
https://developer.android.com/guide/components/processes-and-threads
## Pre-Review Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] page, which explains my responsibilities.
- [ ] I read and followed the [relevant style guides] and ran [the auto-formatter].
- [ ] I signed the [CLA].
- [ ] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [ ] I [linked to at least one issue that this PR fixes] in the description above.
- [ ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or I have commented below to indicate which [version change exemption] this PR falls under[^1].
- [ ] I updated `CHANGELOG.md` to add a description of the change, [following repository CHANGELOG style], or I have commented below to indicate which [CHANGELOG exemption] this PR falls under[^1].
- [ ] I updated/added any relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or I have commented below to indicate which [test exemption] this PR falls under[^1].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[^1]: Regular contributors who have demonstrated familiarity with the repository guidelines only need to comment if the PR is not auto-exempted by repo tooling.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[the auto-formatter]: https://github.com/flutter/packages/blob/main/script/tool/README.md#format-code
[CLA]: https://cla.developers.google.com/
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md
[...]

PR #11453, opened 2026-04-06
Revert "[path_provider_android] Changes internal implementation to use JNI"

Renzo-Olivares:
Reverts flutter/packages#9770

Reason: Fix flutter/flutter tree 

Tracking issue: https://github.com/flutter/flutter/issues/184689

Crash below:

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8685239933882993057/+/u/run_test.dart_for_tool_tests_commands_shard_and_subshard_1_1/stdout

```
'package:flutter_tools/src/platform_plugins.dart': Failed assertion: line 444 pos 15: 'pluginClass != null || dartPluginClass != null || defaultPackage != null': is not true.
```


hellohuanlin (approved):

hellohuanlin:
I wonder how the versioning works? Was the previous PR published already (it has a version bump). 

I'm also curious if the original PR will cause the same crash for plugin users.

jmagman:
> I wonder how the versioning works? Was the previous PR published already (it has a version bump).

This won't unpublish the package. You would need to bump the version. Technically you can unpublish a pub package but I'm not sure the process for that. @stuartmorgan-g would know. Generally we don't do it.

bparrishMines:
cc @liamappelbe

bparrishMines:
If possible the flutter/flutter repo may need to set `path_provider_android: 2.2.23` temporarily. I'm assuming Flutter is using the latest version from `pub.dev`.

jmagman:
> If possible the flutter/flutter repo may need to set `path_provider_android: 2.2.23` temporarily. I'm assuming Flutter is using the latest version from `pub.dev`.

I was trying to do that here, but I'm still hitting test failures. I wasn't able to reproduce them locally. https://github.com/flutter/flutter/pull/184688

Renzo-Olivares:
I'm also trying updating the assert we are hitting, it looks like it might just be an oversight that we are hitting this crash but I don't have enough context https://github.com/flutter/flutter/pull/184695.

the ffiPlugin flag was added in https://github.com/flutter/flutter/pull/96225 , but it was not added to assert of `WindowsPlugin` for some reason (it was added to the assert of `LinuxPlugin`).

stuartmorgan-g:
Since so far the issues are specific to our CI I would rather we avoid a full revert cycle if possible. I have unpublished 2.3.0 from pub.dev temporarily while we work on fixes to our systems; it looks like flutter/flutter already had a real fix landed, and I should be able to fix flutter/packages today, and then undo the unpublshing to check that everything is good without doing full revert/reland cycles.

(If we do revert, we should not revert the changes outside of the package itself, as it causes a ton more CI runs and potential flake)

stuartmorgan-g:
Closing since this was resolved without a revert.

PR #11474, opened 2026-04-09
[video_player_platform_interface] Add video track selection support

nateshmbhat:
## Summary
Platform interface breakout PR for #10688.

- Adds `VideoTrack` class with metadata fields: `trackId`, `size` (width/height), and `bitrate`
- Adds `getVideoTracks()` method to retrieve available video quality variants
- Adds `selectVideoTrack()` method to switch between quality tracks
- Adds `isVideoTrackSupportAvailable()` method to check platform support

## Dependency Chain
This PR is **first** in a series of breakout PRs:
1. `video_player_platform_interface` (this PR)
2. `video_player_android` (pending)
3. `video_player_avfoundation` (pending)
4. `video_player` + `video_player_web` (pending - original PR #10688 updated)

## Test Plan
- [x] Added unit tests for `VideoTrack` class (constructor, equality, hashCode, toString)
- [x] Added unit tests for default implementations of new platform interface methods

PR #11475, opened 2026-04-09
[video_player_android] Add video track selection support

nateshmbhat:
## Summary
Android breakout PR for #10688.

- Implements `getVideoTracks()` and `selectVideoTrack()` methods using ExoPlayer's `TrackSelectionOverride`
- Adds `onVideoTrackChanged` event callback for track change notifications
- Adds comprehensive Java and Dart unit tests

## Dependency Chain
This PR is **second** in a series of breakout PRs:
1. `video_player_platform_interface` (#11474) - pending
2. `video_player_android` (this PR)
3. `video_player_avfoundation` (pending)
4. `video_player` + `video_player_web` (pending - original PR #10688 updated)

**Note:** This PR depends on `video_player_platform_interface` 6.7.0 being published first.

## Test Plan
- [x] Java unit tests for `getVideoTracks()` and `selectVideoTrack()` in `VideoPlayerTest.java`
- [x] Java unit tests for `onVideoTrackChanged` callback in `VideoPlayerEventCallbacksTest.java`
- [x] Dart unit tests for `AndroidVideoPlayer` overrides


stuartmorgan-g:
This PR requires https://github.com/flutter/packages/pull/11474 to land and be published.

nateshmbhat:
@stuartmorgan-g PR unblocked. good to merge ?

stuartmorgan-g:
> @stuartmorgan-g PR unblocked. good to merge ?

No PRs can me merged without review. This needs final re-review by the folks who looked at the original combination PR.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -233,6 +255,188 @@ public void selectAudioTrack(long groupIndex, long trackIndex) {
...
+                        .setOverrideForType(override)
+                        .build());
+
+                // Restore playback state
+                exoPlayer.seekTo(currentPosition);
+                if (wasPlaying) {
+                  exoPlayer.play();
+                }
+              },
+              150);
```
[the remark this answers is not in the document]
mboetger: Please address this.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -384,6 +433,63 @@ class _PlayerInstance {
...
+        const Duration(seconds: 5),
+        onTimeout: () {
+          // If we timeout, just continue - the track may still have been selected
+        },
+      );
+    } finally {
+      _videoTrackSelectionCompleter = null;
+      _expectedVideoTrackId = null;
+    }
+  }
```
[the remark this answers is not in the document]
mboetger: And this

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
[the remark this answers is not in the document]
nateshmbhat: Addressed in 7bdc0327259544c4df4f0df1523e0f39eaeee905. Switched to a member `Handler` plus an `isDisposed` flag; `dispose()` now calls `removeCallbacksAndMessages(null)` and sets the flag before releasing the player, so the 150 ms dimension-change callback cannot run on a released `ExoPlayer`. Added a Robolectric regression test (`testSelectVideoTrack_disposeDuringDimensionChangeDelayDoesNotCrash`) that uses `ShadowLooper.idleFor(200ms)` after `dispose()` and asserts no `seekTo`/`play` is invoked.

packages/video_player/video_player_android/lib/src/android_video_player.dart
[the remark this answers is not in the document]
nateshmbhat: Addressed in 7bdc0327259544c4df4f0df1523e0f39eaeee905. Captured the completer and expected id in locals (`completer`, `expectedId`) and guarded the `finally` cleanup with `identical(_videoTrackSelectionCompleter, completer)` / `_expectedVideoTrackId == expectedId`, so an older overlapping `selectVideoTrack` call can no longer null out a newer call's state. Also merged the null/non-null track branches to remove the duplicated wait/cleanup logic. Added a regression test (`concurrent selectVideoTrack calls do not clobber each other`) that fails fast on call 1 and verifies call 2 still completes on its matching `VideoTrackChangedEvent`.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -384,6 +433,56 @@ class _PlayerInstance {
...
+        // Extract groupIndex and trackIndex from the track id
+        final List<String> parts = track.id.split('_');
+        if (parts.length != 2) {
+          throw ArgumentError(
+            'Invalid track id format: "${track.id}". Expected format: "groupIndex_trackIndex"',
+          );
+        }
+
+        final int groupIndex = int.parse(parts[0]);
+        final int trackIndex = int.parse(parts[1]);
```
tarrinneal: Is it possible these could end up not being ints? int.parse will crash if so.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -384,6 +433,56 @@ class _PlayerInstance {
...
+
+        await _api.selectVideoTrack(groupIndex, trackIndex);
+      }
+
+      // Wait for the onTracksChanged event from ExoPlayer with a timeout
+      await completer.future.timeout(
+        const Duration(seconds: 5),
+        onTimeout: () {
+          // If we timeout, just continue - the track may still have been selected
```
tarrinneal: might be worth logging something for debugging purposes.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
```diff
@@ -121,4 +125,27 @@ private String findSelectedAudioTrackId(@NonNull Tracks tracks) {
...
+   */
+  @Nullable
+  private String findSelectedVideoTrackId(@NonNull Tracks tracks) {
+    int groupIndex = 0;
+    for (Tracks.Group group : tracks.getGroups()) {
+      if (group.getType() == C.TRACK_TYPE_VIDEO && group.isSelected()) {
+        // Find the selected track within this group
+        for (int i = 0; i < group.length; i++) {
+          if (group.isTrackSelected(i)) {
+            return groupIndex + "_" + i;
```
tarrinneal, line 145: since this is mirrored on the dart side, we should probably add a comment on both to ensure that if this changes it's changed in both locations

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -235,7 +262,192 @@ public void selectAudioTrack(long groupIndex, long trackIndex) {
...
+    // Get the track group and create a selection override
+    TrackGroup trackGroup = group.getMediaTrackGroup();
+    TrackSelectionOverride override = new TrackSelectionOverride(trackGroup, (int) trackIndex);
+
+    // Check if the new track has different dimensions than the current track
+    Format currentFormat = exoPlayer.getVideoFormat();
+    Format newFormat = trackGroup.getFormat((int) trackIndex);
+    boolean dimensionsChanged =
+        currentFormat != null
+            && (currentFormat.width != newFormat.width || currentFormat.height != newFormat.height);
```
tarrinneal: If `currentFormat` evaluates to `null` because the player hasn't fully negotiated or initialized the stream format yet, `dimensionsChanged` will resolve to `false`. This bypasses this workaround. if `current == null` and `new != null` should we consider that "dimensionsChanged"?

packages/video_player/video_player_android/lib/src/android_video_player.dart
nateshmbhat: No not possible

packages/video_player/video_player_android/lib/src/android_video_player.dart
nateshmbhat: Added debug logging on the 5-second timeout path so we emit whether the stalled selection was for a specific track or auto quality selection. There is also a test that advances the timeout and asserts that the log is produced.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: Updated the dimension-change check so currentFormat == null also takes the renderer-reset workaround path. That keeps the safe behavior when the current stream format has not been negotiated yet, and there is now a Robolectric regression test covering that branch.

packages/video_player/video_player_android/lib/src/android_video_player.dart
nateshmbhat: Updated this path to parse Android track IDs with int.tryParse and throw ArgumentError if the numeric contract is violated, rather than letting a FormatException bubble out. Added a regression test for non-numeric track IDs as well.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/ExoPlayerEventListener.java
nateshmbhat: Added sync comments on both sides. ExoPlayerEventListener.findSelectedAudioTrackId and findSelectedVideoTrackId now point at android_video_player.dart::_parseAndroidTrackId, and the Dart parser notes that the groupIndex_trackIndex contract is shared with the Java emitters.

packages/video_player/video_player_android/lib/src/android_video_player.dart
```diff
@@ -487,6 +608,19 @@ class _PlayerInstance {
             !_audioTrackSelectionCompleter!.isCompleted) {
           _audioTrackSelectionCompleter!.complete();
         }
+      case VideoTrackChangedEvent _:
```
mboetger, line 582: I feel like the race case described above will happens if you call selectVideoTrack(track) and then immediately call selectVideoTrack(null)

mboetger (changes requested):

packages/video_player/video_player_android/lib/src/android_video_player.dart
nateshmbhat: Good catch — you're right, that interleave was broken. The root cause was that the auto/`null` branch completed on *any* `VideoTrackChangedEvent` (`_expectedVideoTrackId == null || …`). Since the native side reports a concrete track id even for adaptive playback, there was no value to match against, so a `VideoTrackChangedEvent` still in flight from the earlier `selectVideoTrack(track)` call would satisfy and complete the `selectVideoTrack(null)` future on the wrong event (and orphan the first call until its 5s timeout).

Fixed in 30fd84da2: `selectVideoTrack(null)` no longer registers a completer or waits for a track event — clearing the override just resolves on its own, since adaptive selection has no single deterministic "selected track" to wait for. The event handler now matches the reported id exactly (`event.selectedTrackId == _expectedVideoTrackId`), so an unrelated/earlier selection can't complete an in-flight one. Added two regression tests, including the exact `selectVideoTrack(track)` → `selectVideoTrack(null)` interleave you described.

nateshmbhat:
@mboetger have addressed your concern and pushed fix.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -235,7 +262,194 @@ public void selectAudioTrack(long groupIndex, long trackIndex) {
...
+      // Re-enable video with the new track selection after allowing renderer to release.
+      //
+      // Why 150ms delay?
+      // This delay is necessary to allow the MediaCodec decoder and video renderer to fully
+      // release their resources before we attempt to create new ones. Without this delay,
+      // the new decoder may be initialized before the old one is completely released, leading
+      // to resource conflicts and rendering artifacts. The 150ms value was determined through
+      // empirical testing across various Android devices and provides a reliable balance
+      // between responsiveness and ensuring complete resource cleanup. Shorter delays (e.g.,
+      // 50-100ms) were found to still cause glitches on some devices, while longer delays
```
mboetger, line 417: My biggest concern with the PR is this delay.  It seems to be device dependent too - so on lower end devices, this probably just won't work.  That being said, I recognize this is new functionality and their is no exoplayer api to alleviate this.  I'll give it approval, but not a big fan.

mboetger (approved):

nateshmbhat:
@tarrinneal good to merge?

stuartmorgan-g:
@nateshmbhat The PR needs two approvals to land, so no. I'm not a reviewer for this PR, so I'm not sure why you are pinging me.

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
```diff
@@ -385,7 +386,8 @@ public void selectVideoTrack(long groupIndex, long trackIndex) {
     // - ExoPlayer TrackSelection documentation:
     //   https://developer.android.com/media/media3/exoplayer/track-selection
     // - DefaultTrackSelector.setParameters() for track type disabling:
-    //   https://developer.android.com/reference/androidx/media3/exoplayer/trackselection/DefaultTrackSelector.Parameters.Builder#setTrackTypeDisabled(int,boolean)
+    //
+    // https://developer.android.com/reference/androidx/media3/exoplayer/trackselection/DefaultTrackSelector.Parameters.Builder#setTrackTypeDisabled(int,boolean)
```
tarrinneal: This seems unintentional

packages/video_player/video_player_android/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 2.10.0
+* Implements `getVideoTracks()` and `selectVideoTrack()` methods for video track (quality) selection using ExoPlayer.
```
tarrinneal, line 3: need a new line between these

tarrinneal (changes requested):
lg two nits

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
nateshmbhat: not sure what you mean by this here ? are you saying the doc link to the exoplayer api setTrackTypeDisabled is not needed ?

packages/video_player/video_player_android/CHANGELOG.md
nateshmbhat: updated, pls check

packages/video_player/video_player_android/android/src/main/java/io/flutter/plugins/videoplayer/VideoPlayer.java
tarrinneal: no, just that the formatting change seems wrong to me, but I could be wrong

nateshmbhat: I reformatted and verified that it's correct 🙏🏻

tarrinneal (approved):

stuartmorgan-g:
@tarrinneal Anything missing for this to be autosubmitted?

tarrinneal:
> @tarrinneal Anything missing for this to be autosubmitted?

formatting mostly

nateshmbhat:
formatting fixed @stuartmorgan-g @tarrinneal

PR #11479, opened 2026-04-09
[file_selector] Switch to Kotlin Pigeon

stuartmorgan-g:
Replaces the Java Pigeon generator with the Kotlin Pigeon generator, and adjusts the project accordingly:
- Adds Kotlin build setings to Gradle.
- Updates API signatures for Kotlin/Java differences.
- Adds generic Java/Kotlin compat shim to create Result objects from Java, since those haven't been added to the Pigeon generator yet.
- Updates tests to use constructors instead of builders, since the Kotlin generator doesn't create builders.
- Updates tests to use a Java/Kotlin compat shim to read Kotlin Result values, instead of mocking the Java Pigeon response object.

This does not attempt to fix the pre-existing invalid use of null paths described in https://github.com/flutter/flutter/issues/159568, but the behavior does change slightly from an IllegalStateException (which came from the Java builder) to a NullPointerException (from trying to assign to a non-null Kotlin field).

Part of https://github.com/flutter/flutter/issues/158287

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/file_selector/file_selector_android/android/src/test/java/dev/flutter/packages/file_selector_android/FileSelectorAndroidPluginTest.java
```diff
@@ -286,7 +284,7 @@ public void openFilesReturnsSuccessfully() throws FileNotFoundException {
       when(resultMockIntent.getClipData()).thenReturn(mockClipData);
 
       assertThrows(
-          IllegalStateException.class,
+          NullPointerException.class,
```
mboetger, line 287: Isn't this a breaking change?  How do we usually handle things like that?

packages/file_selector/file_selector_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.5.2+6
+
+* Updates internal implementation to use Kotlin Pigeon.
```
mboetger, line 3: Should we mention the NullPointer change?

packages/file_selector/file_selector_android/android/src/test/java/dev/flutter/packages/file_selector_android/FileSelectorAndroidPluginTest.java
stuartmorgan-g: This is a grey area; either way it's a `PlatformException` on the Dart side, so it only breaks people doing string matching on the exact exception. When it's something that we've previously had as an intentional user-facing error that we expect them to handle, we have generally treated changing the exception internals as breaking and avoided doing it unless we are overhauling the plugin to use structure errors.

In this case, where it's an unintentional plugin bug (in an edge case), I don't think we need to assume people are string-matching the exception details and trying to handle it. Once once we fix the bug this will go away regardless.

packages/file_selector/file_selector_android/CHANGELOG.md
stuartmorgan-g: Since it's not an intentional behavior, I didn't think there was much to usefully document. It would have to say something like "The string details of the `PlatformException` reported in some rare failure cases have changed from referencing `IllegalStateException` to `NullPointerException`."

reidbaker (approved):

PR #11496, opened 2026-04-13
[go_router] Batch release

fluttergithubbot:
This PR was created automatically to batch release the `go_router`.


chunhtai:
Looks like this is not going to work due to squash commit that has different sha then the previous release. will need to think of a better way

chunhtai:
I think for now we can continue with manually resolve the conflict by hand until we have a better way

chunhtai (approved):
LGTM

PR #11508, opened 2026-04-14
[tool] Update java formatter

stuartmorgan-g:
Updates the Java formatter from 1.3 to 1.28, which is the latest version of the formatter that can be run with Java 17. Version 1.3 doesn't understand newer Java syntax that we could be using in plugins now that they all require Java 17, such as pattern expressions and switch expressions.

The changes outside of script/tool/ ware made by:
- Running the repo tool `format` command with the new version.
- Asking Gemini to fix up some bad comment wrapping in the autoformat changes (committed as a separate commit, for ease of reviewing). The new formatter enforces line length limits on comments, but doesn't re-wrap later lines after adding breaks, resulting in some comments like:
  ```java
  // Some long comment whose first line was slightly too
  // long
  // and then a bit more.
  ```


stuartmorgan-g:
CHANGELOG/version override: These are just format changes in native code, so there is value to clients in releasing them.

reidbaker (approved):

PR #11526, opened 2026-04-17
Decoupling Material's ProgressIndicator

justinmc:
I'm attempting to move a minimal widget from Material to material_ui in order verify that everything is working. I've chosen ProgressIndicator because it makes only a few other imports from the Material library, but it still has `examples/` that we can copy over as well.

## Not viable without copying over nearly everything?
Wow there are a lot of transitive dependencies. Two big offenders:

 * [debug.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/debug.dart), which imports Scaffold and all of its dependencies.
 * [theme_data.dart](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/theme_data.dart), which imports every *_theme.dart file.

I'm going to start with another PR copying just a leaf file (not a full widget).

  ## TODO

 - [ ] Copy over referenced files.
 - [ ] Export from material_ui.dart.
 - [ ] Try installing and using the package locally.
 - [ ] Any other files needed for the examples or anything else you're forgetting?


justinmc:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

I think this bot comment is because one or more of the committers in the history that I copied in has not signed the CLA, even though I have.

stuartmorgan-g:
> I think this bot comment is because one or more of the committers in the history that I copied in has not signed the CLA, even though I have.

Yes, every commit's author is checked. The standard way we do this for imports is to use the built-in cla/google override mechanism once we are ready to land a PR, with an explanation that it's an import for a repo that was already enforcing the CLA.

justinmc:
From discussion on https://github.com/flutter/packages/pull/11543, it seems that we'll have to do everything in a single mega PR to avoid duplicate commits. So I'll close this PR.

I'll make a note on the new mega PR about overriding the CLA check, thanks for the explanation.

PR #11225, opened 2026-03-11
[video_player] Implement screen auto-lock control for video playback

shrabanti722:
## Description

Adds support for screen auto-lock during video playback on iOS.

By default, the video player keeps the screen awake during playback. Some apps (e.g. login screens with background videos, ambient content) want the screen to be able to auto-lock while video plays. This PR adds an option to allow that behavior.

### Changes

* Add `preventsDisplaySleepDuringVideoPlayback` to `VideoPlayerOptions` (default: `true` to preserve existing behavior)
* When `false`, the screen can auto-lock during playback
* Implemented on iOS/macOS via `AVPlayer.preventsDisplaySleepDuringVideoPlayback`
* Platform interface: `setPreventsDisplaySleepDuringVideoPlayback(int playerId, bool preventsDisplaySleepDuringVideoPlayback)`

### Usage

```dart
VideoPlayerController.networkUrl(
  url,
  videoPlayerOptions: VideoPlayerOptions(preventsDisplaySleepDuringVideoPlayback: false),
);
```

## Breakout PRs

1. Platform interface: #11546
2. AVFoundation (iOS/macOS): #11547

## Issues fixed

* Fixes: [flutter/flutter#183520](https://github.com/flutter/flutter/issues/183520) (How to disable wakelock on iOS side?)
* Addresses the need to allow screen auto-lock during video playback on iOS (e.g. for ambient/background videos)

## Pre-Review Checklist

- [x] I read the [Contributor Guide](https://github.com/flutter/packages/blob/main/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene](https://github.com/flutter/flutter/wiki/Tree-hygiene) page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides](https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style) and ran the [auto-formatter](https://github.com/flutter/packages/blob/main/script/tool/README.md#format-code).
- [x] I signed the [CLA](https://cla.developers.google.com/).
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[video_player]`
- [x] I [linked to at least one issue that this PR fixes](https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview) in the description above.
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning).
[...]

PR #11546, opened 2026-04-21
[video_player_platform_interface] Add preventsDisplaySleepDuringVideoPlayback option

### LINKED ISSUE #183520: Allow screen auto-lock during video playback on iOS
### Use case

The video_player plugin keeps the screen awake during playback. Some apps (e.g. login screens with background videos, ambient content) want the screen to be able to auto-lock while video plays. There is currently no way to allow this on iOS.

### Proposal

Add an option to VideoPlayerOptions (e.g. allowScreenAutoLock) that lets apps opt in to allowing screen auto-lock during playback. When enabled, the screen can auto-lock as usual while video continues playing.

Using wakelock_plus – does not reliably control screen auto-lock during video playback on iOS.
Using other video packages (e.g. Chewie, Better Player) – they have similar limitations; see [fluttercommunity/chewie#808](https://github.com/fluttercommunity/chewie/issues/808).

Additional context

On iOS, AVPlayer has preventsDisplaySleepDuringVideoPlayback, which is enabled by default and keeps the screen awake. Setting it to false when the app opts in would allow the screen to auto-lock during playback.

shrabanti722:
## Description

Adds the `preventsDisplaySleepDuringVideoPlayback` field to `VideoPlayerOptions` and a `setPreventsDisplaySleepDuringVideoPlayback` method to `VideoPlayerPlatform`, allowing platform implementations to control whether the display sleeps during video playback.

Platform interface breakout PR for flutter/packages#11225.

## Changes

- Adds `preventsDisplaySleepDuringVideoPlayback` field to `VideoPlayerOptions` (defaults to `true` to preserve existing behavior).
- Adds `setPreventsDisplaySleepDuringVideoPlayback(int playerId, bool preventsDisplaySleepDuringVideoPlayback)` method to `VideoPlayerPlatform` with a default no-op implementation so existing platform implementations continue to compile without changes.
- Updates CHANGELOG (6.7.0) and pubspec version.
- Adds unit test covering the default value of `preventsDisplaySleepDuringVideoPlayback`.

## Related

- Parent PR: flutter/packages#11225
- Fixes: flutter/flutter#183520


tarrinneal (approved):
@stuartmorgan-g 2nd rev

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -449,6 +460,13 @@ class VideoPlayerOptions {
   /// currently no way to implement this feature in this platform).
   final bool mixWithOthers;
 
+  /// Whether the screen is prevented from sleeping during video playback.
+  ///
+  /// Defaults to `true`.
+  ///
+  /// This option is currently only supported on iOS and macOS.
```
stuartmorgan-g: This comment should not be here; a platform interface cannot make claims about implementations, because they are part of another package that this doesn't control the version of.

It's fine to say that something may not be supported on all platforms, because that's a generic statement that won't become inaccurate if other packages change, but not to say specific things about which ones do and don't.

stuartmorgan-g (changes requested):

shrabanti722:
@stuartmorgan-g are there any more changes needed from my side ?

stuartmorgan-g:
@shrabanti722 I can't review the PR in its current state, as you have uploaded commits with a co-author that doesn't have a signed CLA entry. We can only review PRs that pass the CLA check.

shrabanti722:
@stuartmorgan-g fixed it.

stuartmorgan-g (approved):
LGTM

stuartmorgan-g:
Unfortunately, due to the delay the code in this PR will need to be re-autoformatted, as the line length for the repo changed. Sorry about that.

Once it's updated we can land it.

shrabanti722:
@stuartmorgan-g reformatted.

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
```diff
@@ -122,6 +122,16 @@ abstract class VideoPlayerPlatform extends PlatformInterface {
     throw UnimplementedError('setAllowBackgroundPlayback() has not been implemented.');
   }
 
+  /// Sets whether the screen is prevented from sleeping during video playback.
+  Future<void> setPreventsDisplaySleepDuringVideoPlayback(
+    int playerId,
+    bool preventsDisplaySleepDuringVideoPlayback,
+  ) {
+    throw UnimplementedError(
```
stuartmorgan-g: Sorry, I missed this in the previous review. It looks like the main PR plans to only implement this for some platforms.

A feature cannot both throw an error when not implemented, and not have [a support query](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#api-support-queries), because otherwise there is no way for a client to know whether or not it is safe to call the method.

Is the desired behavior for this feature that it silently no-ops if not implemented, or do clients need to know whether it is going to work?

/cc @tarrinneal

stuartmorgan-g (changes requested):

packages/video_player/video_player_platform_interface/lib/video_player_platform_interface.dart
shrabanti722: switched the default to a silent no-op.

shrabanti722:
@stuartmorgan-g anything else needs to be done ?

stuartmorgan-g (approved):
LGTM

PR #11549, opened 2026-04-21
Decoupling Material

justinmc:
This is the mega PR that attempts to import all of Material into material_ui.

To create this PR I followed [this process](https://github.com/justinmc/flutter-skills/blob/main/flutter-clone/SKILL.md) using the following command:

```
git filter-repo --path packages/flutter/lib/src/material --path examples/api/lib/material
```

### TODOs

 * [ ] Export m3 from material_ui.dart.
 * [ ] Export everything from m3.dart.
 * [ ] Try installing and using the package locally.
 * [ ] Verify examples work.
 * [ ] Verify that `git blame` appears the same in both repos on various files.

### Open questions

 * Should we actually have a combined material_ui and cupertino_ui mega mega PR so that the commit SHAs align between both of them?

### Resources


stuartmorgan-g:
> * Should we actually have a combined material_ui and cupertino_ui mega mega PR so that the commit SHAs align between both of them?

Yes, I would highly, *highly* recommend doing it that way.

To clarify, it's not just the SHAs aligning, it's whether there is one commit that actually has all the changes, or two as-far-as-git-is-concerned-completely-unrelated commits, each with part of the changes. For example, `git show` will not work as expected if it's not all imported at once.

justinmc:
This escalated quickly. Super mega PR coming soon.

PR #11550, opened 2026-04-21
[tool] Remove --against-pub flag

stuartmorgan-g:
We haven't used `--against-pub` in CI in a very long time, and we don't expect to need it locally. We report pub versions as part of the publish step, but the source of truth for validation is the repo, not pub. This removes the flag to simplify the tool.


stuartmorgan-g:
I didn't realize this code was still here until I went to refactor this command into a separate validator class!

bparrishMines (approved):
LGTM

PR #11559, opened 2026-04-21
[in_app_purchase_android] Bump gradle plugin to 9.1.1

camsim99:
Bumps com.android.tools.build:gradle from 8.13.1 to 9.1.1 in packages/in_app_purchase/in_app_purchase_android.


reidbaker:
From @jesswrd, "I would rather this be AGP 9.0.1 and not AGP 9.1.1"

PR #11570, opened 2026-04-23
[file_selector] Update Windows example for deprecations

stuartmorgan-g:
The `file_selector_windows` example was using some methods that have been deprecated in favor of param-object versions. This updates to use those new versions.


stuartmorgan-g:
text-exempt: refactor with no semantic change

stuartmorgan-g:
I'm not sure how I missed this when I made the same fix to the Linux version back in https://github.com/flutter/packages/pull/10542

tarrinneal (approved):
beautiful

PR #11572, opened 2026-04-23
[google_maps_flutter] Replace use of zIndex in examples and tests

stuartmorgan-g:
Replaces uses of the deprecated `zIndex` in examples and tests with the preferred `zIndexInt`.

This doesn't change the production use of `zIndex` in google_maps_flutter_web, as that could change behavior, so should stay for now.


bparrishMines (approved):
LGTM

PR #11602, opened 2026-04-29
[google_fonts] Extract the config class to its own file in prevision of a GoogleFontsLite version

TheCarpetMerchant:
See https://github.com/flutter/packages/pull/11433 and the comment https://github.com/flutter/packages/pull/11433#pullrequestreview-4184165033

### DIFF
--- a/packages/google_fonts/CHANGELOG.md
+++ b/packages/google_fonts/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 8.2.0
+
+- Extract the class `Config` to its own file and rename it `GoogleFontsConfig`. The `Config` name will be deprecated in a future release.
+
 ## 8.1.0
 
 - Adds the ability to supply a custom HTTP client to `GoogleFonts.config`.
--- a/packages/google_fonts/lib/src/google_fonts_config.dart
+++ b/packages/google_fonts/lib/src/google_fonts_config.dart
@@ -18,3 +18,6 @@ class GoogleFontsConfig {
   /// If you supply a client, you are responsible for closing it.
   http.Client? httpClient;
 }
+
+@Deprecated('Use GoogleFontsConfig instead')
+typedef Config = GoogleFontsConfig;
--- a/packages/google_fonts/lib/src/google_fonts_config.dart
+++ b/packages/google_fonts/lib/src/google_fonts_config.dart
@@ -1,0 +1,16 @@
+import 'package:http/http.dart' as http;
+
+/// A collection of properties used to specify custom behavior of the
+/// GoogleFonts library.
+class GoogleFontsConfig {
+  /// Whether or not the GoogleFonts library can make requests to
+  /// [fonts.google.com](https://fonts.google.com/) to retrieve font files.
+  bool allowRuntimeFetching = true;
+
+  /// The HTTP client used to fetch fonts.
+  ///
+  /// If this is null, a shared default [http.Client] will be used.
+  ///
+  /// If you supply a client, you are responsible for closing it.
+  http.Client? httpClient;
+}
--- a/packages/google_fonts/lib/src/google_fonts_config.dart
+++ b/packages/google_fonts/lib/src/google_fonts_config.dart
@@ -1,3 +1,7 @@
+// Copyright 2013 The Flutter Authors
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
 import 'package:http/http.dart' as http;
 
 /// A collection of properties used to specify custom behavior of the
--- a/packages/google_fonts/generator/google_fonts.tmpl
+++ b/packages/google_fonts/generator/google_fonts.tmpl
@@ -9,27 +9,12 @@
 import 'dart:ui' as ui;
 
 import 'package:flutter/material.dart';
-import 'package:http/http.dart' as http;
 
 import 'google_fonts_base.dart';
+import 'google_fonts_config.dart';
 {{#allParts}}
 import '{{partFilePath}}';
 {{/allParts}}
-
-/// A collection of properties used to specify custom behavior of the
-/// GoogleFonts library.
-class Config {
-  /// Whether or not the GoogleFonts library can make requests to
-  /// [fonts.google.com](https://fonts.google.com/) to retrieve font files.
-  bool allowRuntimeFetching = true;
-
-  /// The HTTP client used to fetch fonts.
-  ///
-  /// If this is null, a shared default [http.Client] will be used.
-  ///
-  /// If you supply a client, you are responsible for closing it.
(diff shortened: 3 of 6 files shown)


packages/google_fonts/lib/src/google_fonts_config.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:http/http.dart' as http;
```
Piinks, line 5: This is missing the necessary license header

packages/google_fonts/lib/src/google_fonts_config.dart
```diff
@@ -0,0 +1,16 @@
...
+  /// [fonts.google.com](https://fonts.google.com/) to retrieve font files.
+  bool allowRuntimeFetching = true;
+
+  /// The HTTP client used to fetch fonts.
+  ///
+  /// If this is null, a shared default [http.Client] will be used.
+  ///
+  /// If you supply a client, you are responsible for closing it.
+  http.Client? httpClient;
+}
```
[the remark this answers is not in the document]
Piinks: These are good points. We should try to avoid a breaking change unless absolutely necessary, and in order to allow users to access this class, it should be exported.

Piinks (changes requested):

Piinks:
Thank you for putting this together to make https://github.com/flutter/packages/pull/11433 more robust!

packages/google_fonts/lib/src/google_fonts_config.dart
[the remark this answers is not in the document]
TheCarpetMerchant: This just clutters the namespace for nothing, especially with such a common name as "Config".
The only purpose of the class was to be a clear separation between configuration fields and the rest of the GoogleFonts class, there is no utility to developers having their own instance for other purposes (again, especially with this unfortunate "Config" name). As a dev I would rather not have this in my namespace and I see removing it as an improvement.

Still, if you'd rather not have a breaking change I'll export the google_fonts_config file. It really does seem like a negative to me though.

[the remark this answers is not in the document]
Piinks: Thanks for the explanation! I agree that  Config  is a very generic name and renaming it to GoogleFontsConfig  is a solid improvement to avoid cluttering the namespace.

However, we still need to export  GoogleFontsConfig  because otherwise, developers cannot reference the type directly (e.g., they won't be able to write  `GoogleFontsConfig cfg = GoogleFonts.config;`  or pass it to configuration helper functions).

Since we recently bumped the major version, using a deprecation path is best to avoid churn on the ecosystem. We can expose a deprecated typedef in  google_fonts_config.dart  and export it in lib/google_fonts.dart. We can then clean up and remove this deprecation in a future major version bump.

```dart
@Deprecated('Use GoogleFontsConfig instead')
typedef Config = GoogleFontsConfig;
```

What do you think about using this deprecation path?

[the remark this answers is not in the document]
TheCarpetMerchant: Done

packages/google_fonts/lib/src/google_fonts_config.dart
```diff
@@ -0,0 +1,23 @@
...
+
+@Deprecated('Use GoogleFontsConfig instead')
+typedef Config = GoogleFontsConfig;
```
Piinks, line 23: This will still need a public doc comment on it to pass analysis:
```
/// Deprecated. Use [GoogleFontsConfig] instead.
```

Piinks (commented):
Thanks for the updates! This needs the pubspec and changelog updated in order to merge.

TheCarpetMerchant:
> Thanks for the updates! This needs the pubspec and changelog updated in order to merge.

I can update the changelog, but I'm unsure of what you would want changed in the pubspec ?

Piinks:
You'll have to bump the package version in the pubspec so we can release it.

TheCarpetMerchant:
> You'll have to bump the package version in the pubspec so we can release it.

Ah, I was thinking of waiting for the GoogleFontsLite commit to make a release. Making one for just a change in which file hosts a particular class doesn't seem very useful ?

Piinks:
> > You'll have to bump the package version in the pubspec so we can release it.
> 
> Ah, I was thinking of waiting for the GoogleFontsLite commit to make a release. Making one for just a change in which file hosts a particular class doesn't seem very useful ?

It does. This changes the public API. We don't batch changes for releases of this package.

TheCarpetMerchant:
> > > You'll have to bump the package version in the pubspec so we can release it.
> > 
> > 
> > Ah, I was thinking of waiting for the GoogleFontsLite commit to make a release. Making one for just a change in which file hosts a particular class doesn't seem very useful ?
> 
> It does. This changes the public API. We don't batch changes for releases of this package.

Done

packages/google_fonts/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 8.2.0
+
+- Extract the class `Config` to its own file and rename it `GoogleFontsConfig`. The `Config` name will be deprecated in a future release.
```
Piinks, line 3: It is deprecated though in this change, isn't it?

packages/google_fonts/lib/src/google_fonts_config.dart
Piinks: This documentation is still needed.

Piinks (commented):
Hi @TheCarpetMerchant thanks for updating the pubspec and the changelog. This isn't quite ready to merge yet. Would you like me to finish up the small bits that remain so we can get this merged and unblock the other PR? I want to respect your time and feel like we have been going back and forth over some small final things to make this ready. Let me know, happy to get this in finally.

TheCarpetMerchant:
> Hi @TheCarpetMerchant thanks for updating the pubspec and the changelog. This isn't quite ready to merge yet. Would you like me to finish up the small bits that remain so we can get this merged and unblock the other PR? I want to respect your time and feel like we have been going back and forth over some small final things to make this ready. Let me know, happy to get this in finally.

Sure go ahead. I'll take a look to know what to do next time.
Feel free to squash this to a single commit as well

packages/google_fonts/CHANGELOG.md
Piinks:
```suggestion
- Extract the class `Config` to its own file and rename it `GoogleFontsConfig`. The `Config` class is now deprecated.
```

Piinks:
It looks like I do not have permissions to push to the PR, so I am going to close it to re-open as a fresh PR and get this in to unblock the other change: https://github.com/flutter/packages/pull/12202

PR #11610, opened 2026-04-29
[tool] Extract most conformance checks to separate validator classes

stuartmorgan-g:
Extracts the logic for many of the lightweight, repo-practice commands into separate validator classes, leaving just a minimal amount of logic in the command, in preparation for combining all of these into a single command. Doing so will make running conformance checks locally much easier since there will be fewer commands to run, and will make documenting both the tool itself, and the commands we want run against PRs (e.g., for agent instructions), easier as well.

This is a mostly no-op code move, so that tests can change as little as possible in this PR. Then the next PR will then combine the commands, which will require a lot of test changes, but almost all of the non-test code will be able to stay unchanged because it's in the separate validator classes, which the combined command can call into as they are.

All the new files were created by moving code from the corresponding command file, with very minimal changes (e.g., sometimes context that was a non-private getter from the base command class is a private field in the validator class, so underscores were added.) This is reflected in the fact that there are almost no test changes.


script/tool/test/dependabot_check_command_test.dart
```diff
@@ -68,7 +68,7 @@ $gradleEntries
 ''');
   }
 
-  test('skips with no supported ecosystems', () async {
+  test('passes with no supported ecosystems', () async {
```
stuartmorgan-g, line 71: Skipping won't make sense in the combined command (no package skips *all* of the checks that will be in the combined command), so I didn't restructure the common validator pattern to allow it just for this command, only to remove it again in the next PR, and instead changed this behavior now.

script/tool/test/gradle_check_command_test.dart
stuartmorgan-g, line 1: All the changes in this file are just because the constants moved.

script/tool/test/license_check_command_test.dart
```diff
@@ -74,9 +75,8 @@ void main() {
           .listSync(recursive: true, followLinks: false)
           .whereType<File>()
           .map(
-            (File f) => p.posix.joinAll(
-              p.split(p.relative(f.absolute.path, from: root.path)),
-            ),
+            (File f) =>
+                relativePosixPath(f, from: root, platformContext: p.posix),
```
stuartmorgan-g, line 79: Since I was extracting a helper for use in the validators (to replace the one in the looping command), I changed this to use the new helper that has the same logic.

script/tool/test/version_check_command_test.dart
```diff
@@ -371,8 +372,8 @@ void main() {
           output,
           containsAllInOrder(<Matcher>[
             contains(
-              'Allowing breaking change to plugin_platform_interface due '
-              'to --ignore-platform-interface-breaks',
+              'Ignoring breaking change to plugin_platform_interface due '
+              'to command configuration',
```
stuartmorgan-g, line 376: I didn't want to pass in the flag name to the validator just to keep this message the same. I don't think having the message be less descriptive is going to matter for a flag that's only run in CI, and in a message that's just for auditing if something weird ever happens. We'll still be able to trace the message back to this line.

bparrishMines (approved):
LGTM

PR #11619, opened 2026-04-30
[vector_graphics_compiler] Fix HSL color parsing for decimal percentage components

### LINKED ISSUE #185833: [flutter_svg] HSL to RGB conversion incorrect due to percentage parsing
### What package does this bug report belong to?

flutter_svg

Description
---
There's a bug in the SVG color parsing code that incorrectly scales percentage components that contain decimal points. When an HSL/HSLA component string contains a decimal (for example "76.2745098039%"), the code multiplies the parsed value by 2.55 (parseDouble(raw) * 2.55) and rounds it. That conversion is appropriate for RGB percentage inputs (to convert percent → 0–255), but it is incorrect for HSL saturation and lightness components. HSL percent values should be interpreted in the 0–100 range (or normalized to 0.0–1.0), not converted to 0–255. This leads to incorrect color conversions.

Offending code location
---
The problematic logic is in parser.dart around this line:
https://github.com/flutter/packages/blob/c3360ac8ae748c270cabe3a2f867b4ac9d6d36b1/packages/vector_graphics_compiler/lib/src/svg/parser.dart#L1436

Specifically, the branch that does parseDouble(rawColor)! * 2.55 is applied to any numeric token that contains a decimal point, regardless of whether the token is part of an rgb(...) or hsl(...) input.

Reproduction
---
Input:
- color string: "hsl(270, 100%, 76.2745098039%)"

Observed (current) result:
- Converted color: "#efdeff"

Expected result:
- Converted color: "#c286ff"

Image

Root cause
---
When a percentage token contains a decimal point, the parser multiplies the parsed value by 2.55 and rounds it. For HSL this turns, e.g., 76.2745098039 → ~194 (instead of using ~76.2745 as a percent), so subsequent HSL→RGB math is performed with wrong saturation/ lightness values.

Suggested fix
---
- Do not multiply decimal percentage tokens by 2.55 when parsing HSL/HSLA inputs.
- Distinguish the parsing behavior for rgb()/rgba() versus hsl()/hsla():
  - For rgb()/rgba(), percent components may be converted to 0–255 via *2.55 as currently done.
  - For hsl()/hsla(), percent components should be parsed as floats in the 0–100 range (or directly normalized to 0.0–1.0). Remove the *2.55 scaling for HSL parsing.
- Concretely, modify the parsing branch so that decimal percentages are returned as floats when the current color function is hsl/hsla, while retaining the *2.55 behavior for rgb/rgba percent inputs.

Minimal pseudocode patch suggestion
---
Before (problematic):
```dart
if (rawColor.contains('.')) {
  return (parseDouble(rawColor)! * 2.55).round();
}
```

After (fix — conceptually):
```dart
if (rawColor.contains('.')) {
  if (parsingRgbPercentage) {
    // rgb percentage -> 0..255
    return (parseDouble(rawColor)! * 2.55).round();
  } else {
    // hsl percentage -> keep as percentage float (0..100) to be normalized later
    return parseDouble(rawColor)!;
  }
}
```

Notes
---
- The project already has shared parsing logic intended to handle a variety of color formats; the fix should be careful to preserve existing rgb percentage behavior.
- I verified locally that parsing HSL percentage decimals as floats (0–100) and normalizing to 0.0–1.0 before the HSL→RGB math produces the expected result for the example ("hsl(270, 100%, 76.2745098039%)" → "#c286ff").

Offer
---
I can open a PR with a concrete patch and unit tests covering:
- hsl with integer percentages
- hsl with decimal percentages
- rgb with percentage components (ensuring existing behavior is preserved)
- hsla / rgba alpha parsing

Please let me know if you prefer the fix to be implemented by splitting the parsing flow for rgb/hsl or by adding a flag/parameter to the existing token parser.

fondoger:
## Description

Fixes a bug in the HSL color parser where decimal percentage components
(e.g. `76.2745098039%`) were incorrectly multiplied by `2.55` — the
conversion factor used for **rgb** percentages (percent → 0–255).  For
**hsl**, saturation and lightness must stay in the 0–100 range (later
divided by 100) so the multiplication produced wildly wrong values
(e.g. `76.27 → 194`, then `194 / 100 = 1.94`).

The same branch also mis-handled the `hsla` alpha component: a unitless
decimal like `0.5` should be converted to 0–255 via `× 255`, not `× 2.55`.

### Root cause

`parser.dart` checked `rawColor.contains('.')` and unconditionally
applied `* 2.55`, regardless of whether the current function was
`rgb()`/`rgba()` (correct) or `hsl()`/`hsla()` (incorrect).

### Fix

Track whether each token had a `%` suffix before stripping it.
For HSL percentage components, return the raw `double` (0–100 range).
For unitless alpha decimals (0–1 range), multiply by 255.

### Reproduction (from issue #185833)

```
hsl(270, 100%, 76.2745098039%)
```

| Before fix | After fix |
|---|---|
| `#efdeff` ❌ | `#c286ff` ✓ |

## Tests

Four new tests added to `parsers_test.dart`:

- `hsl` with integer percentages (regression)
- `hsl` with decimal lightness percentage (the reported bug)
- `hsla` with integer percentages + decimal alpha (regression)
- `hsla` with decimal lightness + decimal alpha (combined case)

## Related Issues

Fixes https://github.com/flutter/flutter/issues/185833

### DIFF
--- a/packages/vector_graphics_compiler/lib/src/svg/parser.dart
+++ b/packages/vector_graphics_compiler/lib/src/svg/parser.dart
@@ -1434,33 +1434,29 @@ class SvgParser {
 
     // Conversion code from: https://github.com/MichaelFenwick/Color, thanks :)
     if (colorString.toLowerCase().startsWith('hsl')) {
-      final List<num> values = colorString
+      final List<String> values = colorString
           .substring(colorString.indexOf('(') + 1, colorString.indexOf(')'))
           .split(',')
-          .map((String rawColor) {
-            rawColor = rawColor.trim();
-
-            final bool isPercentage = rawColor.endsWith('%');
-            if (isPercentage) {
-              rawColor = rawColor.substring(0, rawColor.length - 1);
-            }
-
-            if (rawColor.contains('.')) {
-              final double v = parseDouble(rawColor)!;
-              // HSL percentage components (saturation, lightness) stay in
-              // the 0–100 range so they can be divided by 100 below.
-              // Non-percentage decimals are alpha values in 0–1; convert to
-              // 0–255 here to match the int-based storage used below.
-              return isPercentage ? v : (v * 255).round();
-            }
-
-            return int.parse(rawColor);
-          })
+          .map((String rawColor) => rawColor.trim())
           .toList();
-      final double hue = values[0] / 360 % 1;
-      final double saturation = values[1] / 100;
-      final double luminance = values[2] / 100;
-      final int alpha = values.length > 3 ? values[3].toInt() : 255;
+      double parseHslValue(String rawColor) {
+        if (rawColor.endsWith('%')) {
+          rawColor = rawColor.substring(0, rawColor.length - 1);
+        }
+        return parseDouble(rawColor)!;
+      }
+
+      int parseHslAlpha(String rawAlpha) {
+        if (rawAlpha.endsWith('%')) {
+          return (parseHslValue(rawAlpha).clamp(0, 100) * 2.55).round();
+        }
+        return (parseDouble(rawAlpha)!.clamp(0, 1) * 255).round();
+      }
+
+      final double hue = parseHslValue(values[0]) / 360 % 1;
+      final double saturation = parseHslValue(values[1]) / 100;
+      final double luminance = parseHslValue(values[2]) / 100;
+      final int alpha = values.length > 3 ? parseHslAlpha(values[3]) : 255;
       var rgb = <double>[0, 0, 0];
 
       if (hue < 1 / 6) {
--- a/packages/vector_graphics_compiler/lib/src/svg/parser.dart
+++ b/packages/vector_graphics_compiler/lib/src/svg/parser.dart
@@ -1434,18 +1434,24 @@ class SvgParser {
 
     // Conversion code from: https://github.com/MichaelFenwick/Color, thanks :)
     if (colorString.toLowerCase().startsWith('hsl')) {
-      final List<int> values = colorString
+      final List<num> values = colorString
           .substring(colorString.indexOf('(') + 1, colorString.indexOf(')'))
           .split(',')
           .map((String rawColor) {
             rawColor = rawColor.trim();
 
-            if (rawColor.endsWith('%')) {
+            final bool isPercentage = rawColor.endsWith('%');
+            if (isPercentage) {
               rawColor = rawColor.substring(0, rawColor.length - 1);
             }
 
             if (rawColor.contains('.')) {
-              return (parseDouble(rawColor)! * 2.55).round();
+              final double v = parseDouble(rawColor)!;
+              // HSL percentage components (saturation, lightness) stay in
+              // the 0–100 range so they can be divided by 100 below.
+              // Non-percentage decimals are alpha values in 0–1; convert to
+              // 0–255 here to match the int-based storage used below.
+              return isPercentage ? v : (v * 255).round();
             }
 
             return int.parse(rawColor);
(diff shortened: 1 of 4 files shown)


gaaclarke:
Here's a validation of the math in the description using apple's colorsync utility:

Screenshot 2026-05-05 at 9 50 40 AM

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -1434,18 +1434,24 @@ class SvgParser {
 
     // Conversion code from: https://github.com/MichaelFenwick/Color, thanks :)
     if (colorString.toLowerCase().startsWith('hsl')) {
-      final List<int> values = colorString
+      final List<num> values = colorString
```
gaaclarke: The following test fails with this code.  Let's avoid using `List<num>`, `num` is a pain to work with and is slow.

```dart
    test('hsla with integer percentages and alpha 1', () {
      expect(
        parser.parseColor(
          'hsla(270, 100%, 76%, 1)',
          attributeName: 'fill',
          id: null,
        ),
        const Color.fromARGB(255, 194, 133, 255),
      );
    });
```

gaaclarke (changes requested):
The math and premise of the change sounds good, there is just bugs in the implementation.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
fondoger: Thanks for catching this. I updated the implementation to avoid `List<num>` by keeping the parsed HSL function arguments as `List<String>` and parsing hue/saturation/lightness/alpha separately.

For alpha, unitless values now clamp to `0..1` and convert with `* 255`, so `hsla(270, 100%, 76%, 1)` correctly produces full opacity. I also added the regression test you suggested, and `flutter test test/parsers_test.dart` passes locally.

Could you please take another look?

gaaclarke:
@fondoger I think you accidentally closed this PR.  We're still interested in it.

fondoger:
> @fondoger I think you accidentally closed this PR. We're still interested in it.

Hi @gaaclarke , sorry I don't know why it is closed. I just reopened it now. Please review it. Thank you!

gaaclarke (approved):
LGTM! Thanks.

gaaclarke:
Hey @andywolff can you be my second reviewer here, please?

packages/vector_graphics_compiler/lib/src/svg/parser.dart
andywolff, line 1432: Optional, doesn't need to be part of this PR, but an idea for another potential improvement:

If we move the hsl code to a parseHslFunction in colors.dart similar to this parseRgbFunction, that could enable us to support both legacy and modern css style color strings "hsla(H, S, L, A)" and "hsla(H S L A)" the way we do for rgb in colors.dart.

packages/vector_graphics_compiler/lib/src/svg/parser.dart
```diff
@@ -1434,27 +1434,29 @@ class SvgParser {
...
+        }
+        return parseDouble(rawColor)!;
+      }
 
-            if (rawColor.contains('.')) {
-              return (parseDouble(rawColor)! * 2.55).round();
-            }
+      int parseHslAlpha(String rawAlpha) {
+        if (rawAlpha.endsWith('%')) {
+          return (parseHslValue(rawAlpha).clamp(0, 100) * 2.55).round();
```
andywolff, line 1451: I like that you've added clamps. Please consider adding tests for values which would be clamped

andywolff (approved):
LGTM, thanks

PR #11352, opened 2026-03-25
[pigeon] Add FFI and JNI support to Swift and Kotlin

tarrinneal:
This PR introduces optional Native Interop to Pigeon, enabling direct communication between Dart and native code without the overhead of traditional MethodChannel serialization. It leverages FFI (Foreign Function Interface) for Swift (iOS/macOS) and JNI (Java Native Interface) for Kotlin (Android).

This represents a significant architectural shift, moving from message-based passing to direct memory sharing and function calls. It also updates the concurrency model for asynchronous methods, moving from completion handlers/callbacks to modern language features: async/await in Swift and Coroutines in Kotlin.

Generators Covered

Swift Generator: Updated to support FFI bindings and async/await for asynchronous methods.
Kotlin Generator: Updated to support JNI bindings and Kotlin Coroutines for asynchronous methods.
Dart Generator: Updated to handle the generated interop bindings on the Dart side.

What's In Scope

1. Infrastructure: Added core support for useFfi and useJni options.
2. Automation: Implemented multi-step generation flows that automatically invoke jnigen and ffigen to produce final bindings.
3. Config Generation: Added jnigen_config_generator.dart and ffigen_config_generator.dart to generate the necessary configuration files for the external tools.
4. Documentation: Added a detailed native_interop_guide.md explaining prerequisites, setup, and usage.
Tests: Added ni_tests.dart and associated generated files and integration tests to verify the feature.

What's Out of Scope

1. Other Languages: This PR specifically targets Swift and Kotlin for the Native Interop feature. Support for Objective-C, C++, and GObject is not included in this interop implementation, and may not be in the future.
2. Performance Optimization for Complex Classes: As noted in the guide, there is a known performance regression when transferring complex classes with many fields compared to MethodChannel Pigeon. This PR delivers the functional infrastructure, but optimizing this specific case is left for follow-up work.
3. Non-instant released data. Currently all data that is sent over host or flutter api surfaces is converted to the correct shape and type for the language it is moving toward and the data created in the other language is then discarded. This presents some inefficiencies and potential workflows that are not yet available.

work toward https://github.com/flutter/flutter/issues/182230
design doc https://github.com/flutter/flutter/issues/181430

PR #11625, opened 2026-04-30
[pigeon] updates toString and isNullish methods

tarrinneal:
prequel pr to https://github.com/flutter/packages/pull/11352 to land non NI changes to simplify pr.


packages/pigeon/lib/src/cpp/cpp_generator.dart
```diff
@@ -1133,6 +1140,43 @@ class CppSourceGenerator extends StructuredGenerator<InternalCppOptions> {
...
+              }
+            });
+            indent.nest(1, () {
+              indent.writeln('ss << "null";');
+            });
+          } else {
+            if (field.type.isClass) {
+              indent.writeln('ss << $name.ToString();');
+            } else {
+              indent.writeln('ss << $name;');
```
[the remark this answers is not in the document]
stuartmorgan-g: Gemini is correct here. You'll also need custom handling of maps, and maybe lists.

packages/pigeon/lib/src/cpp/cpp_generator.dart
```diff
@@ -1133,6 +1140,43 @@ class CppSourceGenerator extends StructuredGenerator<InternalCppOptions> {
       },
     );
 
+    _writeFunctionDefinition(
+      indent,
+      'ToString',
```
stuartmorgan-g: The idiomatic C++ way to do this is to override `operator<<`; see [this discussion](https://learn.microsoft.com/en-us/cpp/standard-library/overloading-the-output-operator-for-your-own-classes) for instance.

packages/pigeon/lib/src/dart/dart_generator.dart
```diff
@@ -398,6 +398,19 @@ class DartGenerator extends StructuredGenerator<InternalDartOptions> {
...
+    indent.newln();
+    indent.writeln('@override');
+    indent.writeScoped('String toString() {', '}', () {
+      final Iterable<NamedType> fields = getFieldsInSerializationOrder(
+        classDefinition,
+      );
+      final Iterable<String> fieldStrings = fields.map((NamedType field) {
+        return '${field.name}: \$${field.name}';
+      });
+      final String fieldsConcat = fieldStrings.join(', ');
```
stuartmorgan-g: Nit: Consider just inlining this as `${fieldStrings.join(', ')}` since it's barely longer than the variable name.

packages/pigeon/lib/src/gobject/gobject_generator.dart
```diff
@@ -1346,6 +1358,89 @@ class GObjectSourceGenerator
...
+            indent.writeln(
+              'g_string_append_printf(str, "\\"%s\\"", self->$fieldName);',
+            );
+          });
+          indent.writeScoped('else {', '}', () {
+            indent.writeln('g_string_append(str, "null");');
+          });
+        } else if (_isNumericListType(field.type)) {
+          indent.writeln(
+            'g_string_append_printf(str, "[...], length: %zu", self->${fieldName}_length);',
```
stuartmorgan-g: Why is the representation of a number list `[...]` rather than the list of the numbers?

packages/pigeon/lib/src/objc/objc_generator.dart
```diff
@@ -686,6 +687,39 @@ class ObjcSourceGenerator extends StructuredGenerator<InternalObjcOptions> {
...
+      ) {
+        return '${field.name}: %@';
+      });
+      final formatString = '$className(${fieldLabels.join(', ')})';
+
+      final Iterable<String> fieldValues = classDefinition.fields.map((
+        NamedType field,
+      ) {
+        if (_usesPrimitive(field.type)) {
+          return '@(self.${field.name})';
```
stuartmorgan-g: FWIW it would be conceptually cleaner to use the right format string for each primitive type, rather than boxing them just to print them as objects. But since this is just a nice-to-have utility method, and Obj-C isn't a development focus, it's fine to leave as-is.

packages/pigeon/platform_tests/test_plugin/linux/pigeon/core_tests.gen.cc
```diff
@@ -909,6 +919,74 @@ guint core_tests_pigeon_test_all_types_hash(CoreTestsPigeonTestAllTypes* self) {
...
+  g_string_append(str, ", int_map: ");
+  g_string_append(str, "...");
+  g_string_append(str, ", enum_map: ");
+  g_string_append(str, "...");
+  g_string_append(str, ", object_map: ");
+  g_string_append(str, "...");
+  g_string_append(str, ", list_map: ");
+  g_string_append(str, "...");
+  g_string_append(str, ", map_map: ");
+  g_string_append(str, "...");
```
stuartmorgan-g: These should also not all be `...`s.

stuartmorgan-g (changes requested):
The native unit tests should have a test that calls the to-string method for a test object and ensure that the expected string results, so that we are validating that the methods do what we expect (and also have a clear snapshot of what the expected output actually is).

packages/pigeon/lib/src/objc/objc_generator.dart
tarrinneal: lmk if my change is what you had in mind.

packages/pigeon/lib/src/objc/objc_generator.dart
```diff
@@ -686,6 +687,62 @@ class ObjcSourceGenerator extends StructuredGenerator<InternalObjcOptions> {
...
+      ) {
+        if (_usesPrimitive(field.type)) {
+          if (field.type.isEnum) {
+            return '${field.name}: %ld';
+          }
+          switch (field.type.baseName) {
+            case 'bool':
+              return '${field.name}: %@';
+            case 'int':
+              return '${field.name}: %ld';
```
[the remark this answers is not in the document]
tarrinneal: My other Gemini says this is wrong

packages/pigeon/example/app/ios/Runner/Messages.g.swift
```diff
@@ -188,7 +198,7 @@ enum Code: Int {
 }
 
 /// Generated class from Pigeon that represents data sent in messages.
-struct MessageData: Hashable {
+struct MessageData: Hashable, CustomStringConvertible {
   var name: String? = nil
   var description: String? = nil
```
stuartmorgan-g: You'll need to rename this in the Pigeon definition file to fix the compilation error.

It's probably worth calling out in the CHANGELOG as a possible issue, but I don't think we need to version it as breaking (since normally we don't consider additions to be breaking in Dart packages even though they could always cause name collisions in standard usage).

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
```diff
@@ -14,8 +14,18 @@ import Foundation
   #error("Unsupported platform.")
 #endif
 
-private func isNullish(_ value: Any?) -> Bool {
-  return value is NSNull || value == nil
+enum EventChannelMessagesPigeonInternal {
```
stuartmorgan-g, line 17: Why is this changing from a static function to an enum containing only a static function?

packages/pigeon/lib/src/gobject/gobject_generator.dart
```diff
@@ -1346,6 +1359,150 @@ class GObjectSourceGenerator
...
+        } else if (field.type.baseName == 'double') {
+          if (field.type.isNullable) {
+            indent.writeScoped('if (self->$fieldName != nullptr) {', '}', () {
+              indent.writeln(
+                'g_string_append_printf(str, "%g", *self->$fieldName);',
+              );
+            });
+            indent.writeScoped('else {', '}', () {
+              indent.writeln('g_string_append(str, "null");');
+            });
```
stuartmorgan-g: There are enough repetitions of this that it seems worth extracting an `appendValueOrNullString` helper function that takes `({required String fieldName, required String nonNullOutput})` and does:
```dart
indent.writeScoped('if (self->$fieldName != nullptr) {', '}', () {
  indent.writeln(nonNullOutput);
});
indent.writeScoped('else {', '}', () {
  indent.writeln('g_string_append(str, "null");');
});
```

packages/pigeon/lib/src/gobject/gobject_generator.dart
```diff
@@ -3096,3 +3253,126 @@ void _writeDeepHash(Indent indent) {
     },
   );
 }
+
+void _writeDeepToString(Indent indent) {
+  indent.writeScoped(
```
stuartmorgan-g: Wouldn't this be much simpler as a single `indent.format` call?

stuartmorgan-g (changes requested):
Re-review in progress; here are the partial notes. Also, there are some real CI failures.

packages/pigeon/lib/src/kotlin/kotlin_generator.dart
```diff
@@ -390,6 +397,38 @@ class KotlinGenerator extends StructuredGenerator<InternalKotlinOptions> {
...
+        }
+        return '$name=\$$name';
+      });
+      if (fieldStrings.isEmpty) {
+        indent.writeln('return "${classDefinition.name}()"');
+      } else {
+        indent.writeln(
+          'return "${classDefinition.name}(${fieldStrings.join(', ')})"',
+        );
+      }
```
stuartmorgan-g: Why do we need this branching? `[].join(', ')` is an empty string.

(Applies to at least the Java generator as well, possibly others but I didn't go back and check them.)

packages/pigeon/lib/src/objc/objc_generator.dart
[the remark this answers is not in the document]
stuartmorgan-g: If we still supported 32-bit iOS, this would be correct. Since we don't, `long` is fine.

packages/pigeon/lib/src/objc/objc_generator.dart
```diff
@@ -686,6 +687,62 @@ class ObjcSourceGenerator extends StructuredGenerator<InternalObjcOptions> {
...
+            case 'int':
+              return '(long)self.${field.name}';
+            case 'double':
+              return 'self.${field.name}';
+          }
+        }
+        return 'self.${field.name}';
+      });
+
+      final args = fieldValues.isEmpty ? '' : ', ${fieldValues.join(', ')}';
```
stuartmorgan-g: This conditional shouldn't be needed.

stuartmorgan-g (commented):
Remaining comments below. Generally looks good!

packages/pigeon/example/app/ios/Runner/EventChannelMessages.g.swift
tarrinneal: to avoid collisions

packages/pigeon/lib/src/objc/objc_generator.dart
tarrinneal: the comma isn't needed? (or isn't needed to be not there)

tarrinneal:
@stuartmorgan-g I think I've covered all your comments (waiting for ci to tell me if I missed anything build related), but I don't really remember since I did most of this work 2 weeks ago.

packages/pigeon/lib/src/objc/objc_generator.dart
stuartmorgan-g: Oh, I missed the comma. Ignore me then.

packages/pigeon/lib/src/gobject/gobject_generator.dart
```diff
@@ -1346,6 +1359,162 @@ class GObjectSourceGenerator
...
+              final String elementTypeName = _getType(
+                module,
+                field.type,
+                isElementType: true,
+              );
+              indent.writeln(
+                'const $elementTypeName* data = self->$fieldName;',
+              );
+              indent.writeScoped('for (size_t i = 0; i < len; i++) {', '}', () {
+                indent.writeln('if (i > 0) g_string_append(str, ", ");');
```
stuartmorgan-g: The `if` body should have braces, as in the C++ generator.

packages/pigeon/lib/src/gobject/gobject_generator.dart
```diff
@@ -3096,3 +3265,109 @@ void _writeDeepHash(Indent indent) {
...
+    case FL_VALUE_TYPE_FLOAT:
+      return g_strdup_printf("%g", fl_value_get_float(value));
+    case FL_VALUE_TYPE_STRING:
+      return g_strdup_printf("\"%s\"", fl_value_get_string(value));
+    case FL_VALUE_TYPE_UINT8_LIST: {
+      GString* str = g_string_new("[");
+      size_t len = fl_value_get_length(value);
+      const uint8_t* data = fl_value_get_uint8_list(value);
+      for (size_t i = 0; i < len; i++) {
+        if (i > 0) g_string_append(str, ", ");
```
stuartmorgan-g: Same here and all the instances below.

packages/pigeon/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
-## NEXT
+## 26.4.0
```
stuartmorgan-g: Breaking change because of the `description` thing?

stuartmorgan-g (commented):
LGTM with nits.

packages/pigeon/CHANGELOG.md
tarrinneal: You suggested not to, but might as well

stuartmorgan-g: 🤡

stuartmorgan-g (approved):
Changes LGTM, but it looks like a project file is still referring to AllDatatypesTest.m‎ instead of DataClassMethodsTest.m.

tarrinneal:
@stuartmorgan-g this is done, you can submit or review.

PR #11637, opened 2026-05-02
[vector_graphics][vector_graphics_compiler] Fix text-anchor and tspan gap with adjacent <tspan>s

### LINKED ISSUE #185927: [vector_graphics] text-anchor on <text> with multiple <tspan> children does not anchor the chunk as a whole
### Steps to reproduce

Render the following two SVGs side by side with `vector_graphics` (e.g. via `VectorGraphic` + `RawBytesLoader` after `encodeSvg`). The two SVGs are equivalent per the SVG spec — both should produce identical output (a single centered `ABCDEFGABCDEFG` string).

```xml
<!-- svg1: single tspan -->
<svg xmlns="http://www.w3.org/2000/svg" width="200" height="100" fill="none">
  <text x="100" y="50" text-anchor="middle" fill="#FF0000">
    <tspan>ABCDEFGABCDEFG</tspan>
  </text>
</svg>
```

```xml
<!-- svg2: two tspans, same visible text -->
<svg xmlns="http://www.w3.org/2000/svg" width="200" height="100" fill="none">
  <text x="100" y="50" text-anchor="middle" fill="#FF0000">
    <tspan>ABCDEFG</tspan><tspan>ABCDEFG</tspan>
  </text>
</svg>
```

### Expected results

Both SVGs render the same centered text, matching how every browser / Inkscape / Chrome devtools render them.

### Actual results

- svg1 renders correctly (centered).
- svg2 renders **shifted right by half the chunk width**: the first `ABCDEFG` ends up centered around `x=100`, and the second `ABCDEFG` is drawn flush against its right edge, overflowing the container.
- Additionally, there is a small visible **gap between the two tspans** because each tspan is laid out as an independent `Paragraph`, so cross-tspan advance/kerning is lost.

Image

### Root cause

In `packages/vector_graphics/lib/src/listener.dart` (`FlutterVectorGraphicsListener.onDrawText`), each `<tspan>` becomes its own `ui.Paragraph` and is independently offset by `dx - paragraphWidth * xAnchorMultiplier`. Per the SVG spec, `text-anchor` applies to the entire **anchored chunk** (the contiguous sequence of glyphs whose start is established by an explicit `x`/`y` on the parent `<text>` or `<tspan>`), not to each tspan in isolation.

Concretely, for two equal-width (`w`) tspans at `x=100` with `text-anchor=middle`:

| | first tspan dx | second tspan dx |
|---|---|---|
| Spec / browsers (chunk-anchored) | `100 - w` | `100` |
| Current code (per-tspan anchored) | `100 - w/2` | `100 + w/2` |

i.e. the whole chunk is shifted right by `w/2`.

A reproducing unit test (in `packages/vector_graphics/test/listener_test.dart`) drives the listener with two `onDrawText` calls sharing `xAnchorMultiplier = 0.5` and a single `onTextPosition(x: 100, ...)`; the second draw's `dx` lands at `100 + w/2` instead of `100`.

### Code sample

```dart
import 'dart:convert';
import 'dart:typed_data';

import 'package:vector_graphics/vector_graphics.dart';
import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';

const String svg1 = '''
<svg xmlns="http://www.w3.org/2000/svg" width="200" height="100" fill="none">
  <text x="100" y="50" text-anchor="middle" fill="#FF0000"><tspan>ABCDEFGABCDEFG</tspan></text>
</svg>
''';

const String svg2 = '''
<svg xmlns="http://www.w3.org/2000/svg" width="200" height="100" fill="none">
  <text x="100" y="50" text-anchor="middle" fill="#FF0000"><tspan>ABCDEFG</tspan><tspan>ABCDEFG</tspan></text>
</svg>
''';

ByteData _compile(String s) => encodeSvg(
      xml: s,
      debugName: '<string>',
      enableClippingOptimizer: false,
      enableMaskingOptimizer: false,
      enableOverdrawOptimizer: false,
    ).buffer.asByteData();

void main() => runApp(MaterialApp(
      home: Scaffold(
        body: Center(
          child: Column(mainAxisSize: MainAxisSize.min, children: [
            Container(
              decoration: BoxDecoration(border: Border.all(color: Colors.black26)),
              child: VectorGraphic(loader: RawBytesLoader(_compile(svg1))),
            ),
            const SizedBox(height: 8),
            Container(
              decoration: BoxDecoration(border: Border.all(color: Colors.black26)),
              child: VectorGraphic(loader: RawBytesLoader(_compile(svg2))),
            ),
          ]),
        ),
      ),
    ));

class RawBytesLoader extends BytesLoader {
  const RawBytesLoader(this.data);
  final ByteData data;
  @override
  Future<ByteData> loadBytes(BuildContext? context) async => data;
  @override
  int get hashCode => data.hashCode;
  @override
  bool operator ==(Object other) => other is RawBytesLoader && other.data == data;
}
```

### Logs

N/A — purely a rendering/layout issue.

### Flutter Doctor output

<details><summary>flutter doctor -v</summary>

(to be filled in by the issue author when attaching screenshots)

</details>

fondoger:
## Summary

`<text x="..." text-anchor="middle"><tspan>ABCDEFG</tspan><tspan>ABCDEFG</tspan></text>` rendered noticeably differently from the equivalent single-tspan `<tspan>ABCDEFGABCDEFG</tspan>`: the two-tspan version was shifted right past the anchor *and* had a visible space between the two tspans. Two independent root causes, fixed in two separate places.

**1. `vector_graphics` — anchor applies per-chunk, not per-tspan.** The listener was independently offsetting each tspan's paragraph by `dx - paragraphWidth * xAnchorMultiplier`. Per the SVG spec, `text-anchor` applies to the entire anchored chunk (the contiguous run of glyphs whose start is established by an explicit `x`/`y`), not to each tspan in isolation. The listener now buffers per-tspan paragraphs within a chunk, flushes on the next explicit position update or on `toPicture()`, and applies the anchor offset to the chunk total. `text-anchor="start"` behavior is unchanged. A new TextPosition with `x` (or `reset=true`) starts a new chunk; bare per-tspan TextPositions emitted by the parser do not.

**2. `vector_graphics_compiler` — parser injected a spurious space between adjacent tspans.** `_appendText` had a "previous element was a tspan ⇒ prepend a space" rule that fired regardless of whether the source actually had whitespace at the boundary. So `<tspan>A</tspan><tspan>B</tspan>` parsed to `['A', ' B']` (the visible gap), even though every browser parses it to `'AB'`. Replaced with a rule that only prepends when whitespace truly exists at the boundary — either as a leading-whitespace prefix on the current text, or as an earlier whitespace-only text event we've now flagged.

Tests added on both sides:
- `vector_graphics/test/listener_test.dart`: `Text anchor middle centers the entire chunk across tspans` — drives the listener with a two-tspan middle-anchored chunk (including the parser's bare per-tspan TextPosition between them) and asserts the second tspan starts at the original x.
- `vector_graphics_compiler/test/parser_test.dart`: `adjacent tspans without whitespace are not separated by a space` and `adjacent tspans with whitespace between still get a space` — locks down both branches of the new parser rule.

Both packages bumped to `1.2.1` with CHANGELOG entries.

Fixes flutter/flutter#185927

## Before / After

| SVG | Before (buggy) | After (this PR) |
| --- | --- | --- |
| **svg1** (single tspan)<br>`<svg xmlns="http://www.w3.org/2000/svg" width="200" height="100" fill="none">`<br>&nbsp;&nbsp;`<text x="100" y="50" text-anchor="middle" fill="#FF0000">`<br>&nbsp;&nbsp;&nbsp;&nbsp;`<tspan>ABCDEFGABCDEFG</tspan>`<br>&nbsp;&nbsp;`</text>`<br>`</svg>`<br><br>**svg2** (two tspans)<br>`<svg xmlns="http://www.w3.org/2000/svg" width="200" height="100" fill="none">`<br>&nbsp;&nbsp;`<text x="100" y="50" text-anchor="middle" fill="#FF0000">`<br>&nbsp;&nbsp;&nbsp;&nbsp;`<tspan>ABCDEFG</tspan><tspan>ABCDEFG</tspan>`<br>&nbsp;&nbsp;`</text>`<br>`</svg>` | image | image |

## Test plan

- [x] `flutter test` in `packages/vector_graphics` (all 64 tests pass, including the new `Text anchor middle centers the entire chunk across tspans`).
- [x] `dart analyze lib test` — no issues.
- [x] `dart format` — clean.
- [x] Manually verified rendering of the reproducer SVG (single-tspan and two-tspan equivalents now render identically).

## Pre-launch Checklist

- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I updated `CHANGELOG.md` to add a description of the change.

[Flutter Style Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md
[exempt from version changes]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version-and-changelog-updates
[test-exempt]: https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/vector_graphics/lib/src/listener.dart
```diff
@@ -652,6 +667,9 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
   @override
   void onUpdateTextPosition(int textPositionId) {
     final _TextPosition position = _textPositions[textPositionId];
+    // Any explicit absolute position (or reset) starts a new SVG anchored
+    // chunk, so flush whatever we've queued for the previous one.
+    _flushPendingTextChunk();
```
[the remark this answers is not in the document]
fondoger: Good catch — addressed in 26978d9. Now flushes on `position.reset || position.x != null || position.y != null` (per the SVG spec, both `x` and `y` start a new anchored chunk; relative `dx`/`dy` correctly continue the chunk and only move the pen).

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -685,9 +703,15 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
     final double dy = _textPositionY;
-    double paragraphWidth = 0;
 
-    void draw(int paintId) {
+    if (_pendingChunk.isEmpty) {
+      _chunkOriginX = dx;
+      _chunkAnchorMultiplier = textConfig.xAnchorMultiplier;
+      _chunkAdvance = 0;
+    }
+    final double offsetWithinChunk = _chunkAdvance;
```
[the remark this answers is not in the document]
fondoger: Both points addressed in 26978d9: (a) a change in `xAnchorMultiplier` between segments now flushes the chunk before starting the next; (b) when continuing a chunk, `_chunkAdvance` is recomputed as `dx - _chunkOriginX!` so any in-chunk `dx="..."` movements applied via `onUpdateTextPosition` are reflected in the segment's offset within the chunk. Thanks for the careful review!

gaaclarke (approved):
This looks good to me.  The code is good style, the tests are adequate and the premise for the change matches my understanding of the SVG spec.

> The ‘text-anchor’ property is applied to each individual text chunk within a given ‘text’ element.

https://www.w3.org/TR/SVG11/text.html#TextAnchorProperty

gaaclarke:
@walley892 please be a second review for this PR.

fondoger:
Hi @walley892, you were added as a required reviewer for this PR. Could you please review it? Thank you!

gaaclarke:
> Hi @walley892, you were added as a required reviewer for this PR. Could you please review it? Thank you!

Hey @fondoger, English may be your second language so I wanted to let you know that this was a rude comment.  A more appropriate message would have been "Friendly ping wally :)", or saying something like "Hey wally, just making sure you saw this.  Thanks!"

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -707,37 +750,60 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
+        _PendingTextDraw(p, offsetWithinChunk, dy, _textTransform),
+      );
+    }
+    if (strokeId != null) {
+      final Paragraph p = buildParagraph(strokeId);
+      paragraphWidth = p.maxIntrinsicWidth;
+      _pendingChunk.add(
+        _PendingTextDraw(p, offsetWithinChunk, dy, _textTransform),
+      );
+    }
```
walley892, line 767: Can neither of these be true? Or is strokeId || fillId always true?

If so, these blocks can probably be unified with `p = fillId != null : buildParagraph(fillId) ? buildParagraph(strokeId)` followed by the rest

packages/vector_graphics/lib/src/listener.dart
```diff
@@ -685,9 +709,28 @@ class FlutterVectorGraphicsListener extends VectorGraphicsCodecListener {
...
+      _chunkAnchorMultiplier = textConfig.xAnchorMultiplier;
+      _chunkAdvance = 0;
+    } else {
+      // Continuing the chunk: take the live pen position so any in-chunk
+      // relative `dx="..."` movements applied via onUpdateTextPosition
+      // since the last segment are accounted for in the segment's offset
+      // within the chunk.
+      _chunkAdvance = dx - _chunkOriginX!;
+    }
+    final double offsetWithinChunk = _chunkAdvance;
```
walley892: Can you move this closer to where it's used below?

walley892: Actually, if you can unify the blocks below this variable is probably unnecessary

walley892 (changes requested):
A few small comments, mostly looks good!

packages/vector_graphics/lib/src/listener.dart
fondoger: `fillId` and `strokeId` can both be non-null for text that has both fill and stroke, so we can't collapse this to a single ternary without dropping one of the draws.

I agree there is a little duplication here, but since the two cases intentionally both run and the duplicated code is local, I think keeping the two explicit branches is clearer than adding a helper/nullable state just to share these few lines.

packages/vector_graphics/lib/src/listener.dart
fondoger: Removed `offsetWithinChunk` and now pass `_chunkAdvance` directly when queueing each `_PendingTextDraw`.

I kept the fill/stroke branches separate for the reason mentioned in the other thread: both can be non-null and need separate draw entries.

packages/vector_graphics/lib/src/listener.dart
walley892: Thanks for the clarification!

walley892 (approved):
Thanks for the contribution!

gaaclarke (approved):

walley892 (approved):

gaaclarke (approved):

gaaclarke (approved):

PR #11641, opened 2026-05-04
[tool] Unify most validation commands

stuartmorgan-g:
Replaces all of the following repo tooling commands with a new unified `validate` command:
- `dependabot-check`
- `gradle-check`
- `pubspec-check`
- `readme-check`
- `repo-package-info-check`
- `version-check`

It also makes some simplifications in using these:
- Temporarily disabling the code excerpt check is now done via package-level ci_config.yaml files. This means that we don't have to exempt a package from the entire command just for this, and also means that fixing a package and removing its exemption will no longer trigger whole-repo testing as it used to.
- Setting the min Flutter version allowed for the repo is now a repo-level config in a new root level `tool_config/` directory. This avoids the need to pass a flag every time for something that isn't actually intended to be variable.
- The allowed dependencies are now also in hard-coded locations in the `tool_config/` directory, instead of having to pass them in all the time, to make the commands easier to call. As with min SDK version, this isn't something callers actually need to be able to change on the fly to the flags add more complexity (in exchange for making it extremely clear at the CI call site what was happening, but I no longer think that tradeoff is the right one based on experience using the tool).

To avoid huge amounts of test file churn, and massively complicating test setup, there is a test-only mechanism to select individual validators to run in tests, rather than having to test the entire command at once.

Part of https://github.com/flutter/flutter/issues/173413


.ci/targets/repo_checks.yaml
```diff
@@ -22,46 +22,12 @@ tasks:
     script: .ci/scripts/tool_runner.sh
     args: ["license-check"]
```
stuartmorgan-g, line 27: The reason this isn't combined is that it's a repo-wide command, not package-level (because we don't want files to accidentally not get license blocks when they aren't in a package).

It is annoying for local validation that it's not part of the same command, but I haven't thought of a good way to combine it yet without making it more likely to miss files.

script/tool/lib/src/validators/readme_validator.dart
```diff
@@ -30,16 +30,13 @@ class ReadmeValidator {
     required path.Context path,
     required String indentation,
     required void Function(String) warningLogger,
-    required bool requireCodeExcerpts,
```
stuartmorgan-g, line 33: The changes in this file are to switch to the new package-local config.

script/tool/lib/src/validators/version_and_changelog_validator.dart
```diff
@@ -120,7 +120,15 @@ class VersionAndChangelogValidator {
     required bool checkForMissingChanges,
     required bool ignorePlatformInterfaceBreaks,
   }) async {
-    final Pubspec pubspec = package.parsePubspec();
+    final Pubspec? pubspec = _tryParsePubspec(package);
+    if (pubspec == null) {
```
stuartmorgan-g, line 124: I moved this logic from the command to the validator, since we no longer need to support "skip" for publish-to none, just pass.

script/tool/lib/src/validate_command.dart
```diff
@@ -0,0 +1,384 @@
...
+
+    return errors.isEmpty
+        ? PackageResult.success()
+        : PackageResult.fail(errors);
+  }
+
+  bool _shouldRun(Validator validator) =>
+      targetedValidators?.contains(validator) ?? true;
+
+  /// Runs repo-level checks.
```
stuartmorgan-g, line 190: Essentially all of the code from here on in this file is moved with minimal changes from the deleted command files.

stuartmorgan-g:
@bparrishMines @tarrinneal Ping on this review; this is blocking work to make flutter/core-packages usable.

script/tool/lib/src/common/ci_config.dart
```diff
@@ -26,18 +26,33 @@ class CIConfig {
...
     }
 
-    return CIConfig._(isBatchRelease);
+    // Any package that hasn't been explicitly exempted is assumed to require
+    // excerpts.
+    var requiresExcerpts = true;
+    final Object? exemptFromExcerpts = loaded['exempt_from_excerpts'];
+    if (exemptFromExcerpts is bool) {
+      requiresExcerpts = !exemptFromExcerpts;
+    }
```
tarrinneal: probably just this way to be very obvious about behaviors, but it could be shorter:
`final requiresExcerpts = loaded['exempt_from_excerpts'] != true;`

tarrinneal (commented):
loogs good ta meh

tarrinneal (approved):

script/tool/lib/src/common/ci_config.dart
stuartmorgan-g: It's probably this way because the more complicated code above it that needed to read a sub-map did that, and it didn't occur to me to simplify it 🙂  Done!

PR #11649, opened 2026-05-05
[material_ui, cupertino_ui, infra] Support skipping golden files for the design migration

Piinks:
This branches off of https://github.com/flutter/packages/pull/10753 and strips it down to just the skipping comparator. This means when golden file testing is used with this version of flutter_goldens, `matchesGoldenFile` will return true in all environments without executing image comparison.
The goal of this PR is to join the material_ui and cupertino_ui mega-migration PR (cc @justinmc ) to disable golden file testing for now in material_ui and cupertino_ui so we can get a clearer signal from CI about any remaining blockers.
Full golden file testing support will likely land in #10753 as a separate step given the enormity of these PRs.

cupertino_ui and material_ui are set `publish_to: none` here, which is intentional. This allows us to split up these changes, and wait to publish until everything lands and is working as we want it to.

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/material_ui/pubspec.yaml
```diff
@@ -1,6 +1,7 @@
 name: material_ui
 description: The official Flutter Material UI Library, implementing Google's Material Design design system.
-version: 0.0.1
+version: 0.0.2
+publish_to: none
```
justinmc, line 4: Good call for now. I'll plan to remove this when we are ready to attempt to publish the first pre-release version.

justinmc (approved):
LGTM. After merging I can update my PR https://github.com/flutter/packages/pull/11568 and then make sure it correctly skips the golden tests.

packages/material_ui/pubspec.yaml
Piinks: Ok thanks. Let me double check with Stuart before merging this.

script/flutter_goldens/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
...
+  crypto: any
+  file: any
+  flutter:
+    sdk: flutter
+  flutter_test:
+    sdk: flutter
+  path: any
+  platform: any
+  process: any
+  yaml: any
```
stuartmorgan-g: Why are all of these `any`? That's almost guaranteed to cause breakage in the future.

packages/cupertino_ui/test/goldens/goldens_test.dart
stuartmorgan-g, line 1: Are any of these file paths that exist in the real packages? Collisions are likely to make the import more difficult, so ideally any placeholders should use different names than real files.

script/flutter_goldens/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
+name: flutter_goldens
+publish_to: none
+
+environment:
+  sdk: ^3.9.0-0
```
stuartmorgan-g: Why is this using a prerelease version format?

Piinks: Probably because I copied most of this over from flutter/flutter for consistency. 😅 
Let me fix this, thank you!

script/flutter_goldens/pubspec.yaml
Piinks: So this is the same as in flutter/flutter, and AFAIK it has not caused breakages. What should it be? The latest?

packages/cupertino_ui/test/goldens/goldens_test.dart
Piinks: Yup, that is intentional. These are not existing file paths so they should not create collisions so that it is easier to merge with the existing mega-migration PR.

script/flutter_goldens/pubspec.yaml
Piinks: Wait, yeah I thought I had checked this.. other packages here use the same format:

https://github.com/flutter/packages/blob/67ec0d36293eb963590e0606c8bede379be4d142/packages/two_dimensional_scrollables/pubspec.yaml#L8

https://github.com/flutter/packages/blob/67ec0d36293eb963590e0606c8bede379be4d142/packages/go_router/pubspec.yaml#L9

What should it be? I can send a separate PR to fix others if they are wrong too.

Piinks: Oh, unless it's the `*-0`?

stuartmorgan-g: Yes, `X-0` means "any pre-release version of X". I would not expect us to need our tooling to support old prereleases.

script/flutter_goldens/pubspec.yaml
stuartmorgan-g: > So this is the same as in flutter/flutter, and AFAIK it has not caused breakages.

flutter/flutter uses a repo-level workspace and checks in a lock file for it, so that every version of everything is exactly controlled in CI. This repo just uses standard pub resolution on a per-package basis, and (per Dart standards) doesn't check in lock files for packages.

So `any` in flutter/flutter is effectively "don't get in the way of the script that updates the lock file, which is where everything is actually controlled", whereas `any` here has the typical pub meaning of "Give me literally any version (which will in practice be the latest resolvable version every time it runs), and I don't care if it's a major version with breaking changes".

> What should it be? The latest?

Typically we want packages depending on a `^` version for the latest major version, but I'm not sure what exactly this particular package needs. If you want to mostly match flutter/flutter's current behavior you could use a `^` range starting with whatever the flutter/flutter pubspec.lock version is for each package. If you want to exactly match flutter/flutter you could exactly pin every version here, but since we don't have a pub dependency roller for this repo that would mean a lot more manual work managing dependency versions over time.

.ci/targets/repo_tools_tests.yaml
```diff
@@ -4,3 +4,5 @@ tasks:
     infra_step: true # Note infra steps failing prevents "always" from running.
   - name: tool unit tests
     script: .ci/scripts/plugin_tools_tests.sh
+  - name: flutter_goldens unit tests
```
stuartmorgan-g, line 7: If this package isn't intended to be run from Windows, you could just not add this here rather than fix the failures.

script/flutter_goldens/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
...
+  crypto: 3.0.7
+  file: 7.0.1
+  flutter:
+    sdk: flutter
+  flutter_test:
+    sdk: flutter
+  path: 1.9.1
+  platform: 3.1.6
+  process: 5.0.5
+  yaml: 3.1.3
```
stuartmorgan-g: What is the plan for keeping these updated?

Also, since this is non-standard for this repo, it would be good to explain in a comment in this file why it's important that these be pinned instead of using the typical `^` dependencies.

stuartmorgan-g (approved):

script/flutter_goldens/pubspec.yaml
Piinks: > What is the plan for keeping these updated?

I am not sure yet given they are `any` in flutter/flutter we haven't had to consider that. I can take up as an action item for the full golden infra PR to resolve ongoing maintenance.

> instead of using the typical ^ dependencies.

I don't think they need to be pinned, I just referred to the lock file in flutter/flutter to determine the right versions. If they weren't pinned, would ongoing maintenance be easier?

stuartmorgan-g: > If they weren't pinned, would ongoing maintenance be easier?

Yes; if they use a standard ^ range then they only need to be touched when there is a new major version, or someone is adding code that uses a newer feature that needs a higher minimum version (both of which are likely to be rare), rather than every single time we want to get a bugfix to any dependency.

Piinks: Ok cool. All green here, I am will add a section to the doc about maintaining this going forward. We need a better strategy for that in flutter/flutter anyways, recently we bumped the goldctl version that had not been updated in... maybe years.

PR #11653, opened 2026-05-06
[two_dimensional_scrollables] Activate leak testing and fix memory leaks

ValentinVignal:
The package manipulates disposable objects. This PR activates leak testing to make sure disposable objects are correctly disposed.

It also fixes the memory leak warnings from the tests

See the documentation: https://github.com/dart-lang/leak_tracker/blob/main/doc%2Fleak_tracking%2FDETECT.md

## Pre-Review Checklist

- [x] I signed the [CLA].


ValentinVignal:
cc @polina-c

packages/two_dimensional_scrollables/lib/src/table_view/table.dart
```diff
@@ -129,34 +129,34 @@ class TableView extends TwoDimensionalScrollView {
...
-    super.horizontalDetails,
-    super.verticalDetails,
-    super.cacheExtent,
-    super.diagonalDragBehavior = DiagonalDragBehavior.none,
-    super.dragStartBehavior,
-    super.keyboardDismissBehavior,
-    super.clipBehavior,
+  static Widget builder({
+    Key? key,
+    bool? primary,
```
ValentinVignal: I'm not sure whether that is okay to change from a constructor to a static method. Would you have a better design in mind?

packages/two_dimensional_scrollables/test/table_view/table_test.dart
```diff
@@ -2030,106 +2038,118 @@ void main() {
...
-              },
-              cellBuilder: (_, TableVicinity vicinity) {
-                // Merged column
-                if (mergedColumns.contains(vicinity.column) &&
-                    vicinity.row == 0) {
+      testWidgets(
+        'merged column that exceeds metrics will assert',
+        // The build throws an assertion error which prevents the table from
+        // properly disposing the elements.
+        experimentalLeakTesting: LeakTesting.settings.withIgnoredAll(),
```
ValentinVignal, line 1812: Is it acceptable to have memory leaks when the build asserts?

packages/two_dimensional_scrollables/example/lib/table_view/simple_table.dart
```diff
@@ -51,6 +51,7 @@ class _TableExampleState extends State<TableExample> {
                 ),
               )
             : TableView.builder(
+                key: ValueKey(_selectionMode),
```
ValentinVignal: I'm doing this to force the widget to rebuild (and re-run `_buildCell`). Because now, when we toggle from "Single-cell" to "Disable" selection, the delegate doesn't change, and `_buildCell` is not rebuilt. That makes the test `Selection SegmentedButton control works` in `packages/two_dimensional_scrollables/example/test/table_view/simple_table_test.dart:60` fail.

Ideally, I think we'd need some kind of notifier that the widget built in `_buildCell` would listen to. But since it is only an example, I'd thought this would be enough?

What do you think?

Piinks:
Hey @ValentinVignal thanks for this PR! Sorry I have not finished reviewing it yet. It looks like several checks are failing, can you take a look?

Piinks (commented):
Hi @ValentinVignal, thanks again for working on this! Fixing these memory leaks is super important, but I have some concerns about the breaking changes introduced in this PR:

Changing the constructors `TableView.builder` and `TableView.list` to static methods returning `Widget` (instead of `TableView`) is a breaking change. It breaks code that explicitly types variables as `TableView` (e.g., `TableView table = TableView.builder(...)`) and widget tests querying `TableView` by key.

Instead, I think we can make `TableView` itself a `StatefulWidget` that wraps a private stateless `_TableViewWidget` extending `TwoDimensionalScrollView`. The state class can manage the lifecycle of internally-created delegates, avoiding any breaking changes or type changes.

Since `TreeView` is already a `StatefulWidget`, we can manage the `TreeRowBuilderDelegate` directly inside the existing `_TreeViewState` instead of introducing the new private stateful `_TreeView` widget wrapper.

What do you think about going with this alternative? It allows us to land leak safety without breaking changes or class bloat.

ValentinVignal:
@Piinks I updated the PR as per your suggestion. Let me know what you think about it :)

Piinks (changes requested):
Thanks for the updates @ValentinVignal! It looks like two things still need to be addressed:
1. Revert `TableView.builder` and `TableView.list` back to class constructors/factories returning `TableView` by making `TableView` itself a `StatefulWidget`.
2. Move `TreeRowBuilderDelegate` lifecycle management directly into `_TreeViewState`.

CI appears to be a bit unhappy as well.

ValentinVignal:
> 1. Revert TableView.builder and TableView.list back to class constructors/factories returning TableView by making TableView itself a StatefulWidget.

Isn't it what I did already? 

Screenshot 2026-06-24 at 11 45 56 AM

> 2. Move TreeRowBuilderDelegate lifecycle management directly into _TreeViewState.

Oops, sorry about that, I missed that one. This should be good not

> CI appears to be a bit unhappy as well.

It seems like it is missing some python packages when trying to install the android emulators. Are those failure coming from the code changes of this PR ?

```
  Downloading https://us-python.pkg.dev/chrome-python-ar/chrome-python-ar/pyfakefs/pyfakefs-3.7.2-py3-none-any.whl (177 kB)
ERROR: Could not find a version that satisfies the requirement pylsqpack==0.3.12
ERROR: No matching distribution found for pylsqpack==0.3.12

If you see this, it means some python packages are missing from Artifact Registry or installation failed.
Please report this issue at https://crbug.com/492362903
```

packages/two_dimensional_scrollables/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * Fixes hit testing for `TreeView` row content and gestures after horizontal scrolling.
 * Updates minimum supported SDK version to Flutter 3.38/Dart 3.10.
+* Fixes memory leaks.
```
Piinks: 0.5.3 has already been released, so this will need a new entry

packages/two_dimensional_scrollables/example/lib/table_view/simple_table.dart
Piinks: Will other users have to apply this change to their existing code? This makes me wonder if this is a breaking change.. might there be performance concerns as well?

packages/two_dimensional_scrollables/lib/src/table_view/table.dart
```diff
@@ -222,18 +235,308 @@ class TableView extends TwoDimensionalScrollView {
...
+class _TableViewState extends State<TableView> {
+  late TableCellDelegateMixin _delegate;
+
+  TableCellDelegateMixin _buildDelegate() {
+    if (widget.delegate != null) {
+      return widget.delegate!;
+    }
+    return switch (widget._buildDelegateParameters) {
+      _TableCellBuilderDelegateParameters(
+        :final int? columnCount,
```
Piinks: This is interesting syntax, maybe I am just unfamiliar, but it feels difficult to read and follow along

packages/two_dimensional_scrollables/example/lib/table_view/simple_table.dart
Piinks: I think I have an idea to mitigate this - instead of caching delegate parameters in  TableView  and conditionally updating the delegate in  TreeView , we should simplify the lifecycle management to avoid correctness bugs (e.g., method tear-offs evaluating to  ==  and blocking cell updates):

1. Constructors: Instantiate the delegates directly within the  TableView constructors and track if they are internally managed with a  final bool _isInternalDelegate  flag.
2. Disposal: In  _TableViewState.didUpdateWidget  and  dispose , simply check if _isInternalDelegate  is true and dispose of the delegate.
3. Rebuilding: Remove the parameter caching classes (_TableCellDelegateParameters  and subclasses) completely. Always recreate the delegates on rebuilds (including in  _TreeViewState.didUpdateWidget ) to ensure parent state changes are correctly propagated to the viewport.
4. Revert the  key: ValueKey(_selectionMode)  workaround in the example application since parent updates will now propagate naturally.

WDYT?

packages/two_dimensional_scrollables/lib/src/table_view/table.dart
ValentinVignal: That was a way to make the rest of the code less verbose. Another way to write it would be (and I was trying to avoid it):

```dart
      _TableCellBuilderDelegateParameters() => TableCellBuilderDelegate(
        columnCount:
            (widget._buildDelegateParameters! as _TableCellBuilderDelegateParameters).columnCount,
        rowCount:
            (widget._buildDelegateParameters! as _TableCellBuilderDelegateParameters).rowCount,
        pinnedColumnCount: (widget._buildDelegateParameters! as _TableCellBuilderDelegateParameters)
            .pinnedColumnCount,
        pinnedRowCount: (widget._buildDelegateParameters! as _TableCellBuilderDelegateParameters)
            .pinnedRowCount,
        trailingPinnedColumnCount:
            (widget._buildDelegateParameters! as _TableCellBuilderDelegateParameters)
                .trailingPinnedColumnCount,
        trailingPinnedRowCount:
            (widget._buildDelegateParameters! as _TableCellBuilderDelegateParameters)
                .trailingPinnedRowCount,
        cellBuilder:
            (widget._buildDelegateParameters! as _TableCellBuilderDelegateParameters).cellBuilder,
        columnBuilder:
            (widget._buildDelegateParameters! as _TableCellBuilderDelegateParameters).columnBuilder,
        rowBuilder:
            (widget._buildDelegateParameters! as _TableCellBuilderDelegateParameters).rowBuilder,
      ),
```

Having said that, do you prefer the current code, or the more verbose version? Or maybe there is a better way to write it you have in mind? :)

ValentinVignal: Ultimately, that got removed with [Do not cache deleguate](https://github.com/flutter/packages/pull/11653/commits/c2ec712d22fbd6661ff1d5bfd5696c8a7054e36f)

packages/two_dimensional_scrollables/CHANGELOG.md
ValentinVignal: This should be fixed in [Fix version number](https://github.com/flutter/packages/pull/11653/commits/a8c891c7935549e4ecd275f0bd59e92d22aee0ae)

packages/two_dimensional_scrollables/example/lib/table_view/simple_table.dart
ValentinVignal: I'm good with that. What do you think of [Do not cache deleguate](https://github.com/flutter/packages/pull/11653/commits/c2ec712d22fbd6661ff1d5bfd5696c8a7054e36f) ? :)

packages/two_dimensional_scrollables/lib/src/tree_view/tree.dart
```diff
@@ -705,6 +737,14 @@ class _TreeViewState<T> extends State<TreeView<T>>
     assert(_treeController != null);
     assert(_treeController!._state != null);
     _unpackActiveNodes();
+    if (oldWidget.addAutomaticKeepAlives != widget.addAutomaticKeepAlives ||
```
Piinks: Thanks for updating this @ValentinVignal! Looking closer at `_TreeViewState.didUpdateWidget`, adding this `if` guard introduces a stale-rendering regression when `widget.tree` changes.

If a parent widget rebuilds `TreeView` with a new `tree` that happens to have the **same number of active nodes** as the previous tree (e.g. swapping 2 nodes for 2 different nodes):
1. The `if` condition evaluates to `false` because `treeNodeBuilder`, `treeRowBuilder`, etc. haven't changed.
2. `_unpackActiveNodes()` updates `_activeNodes` and sets `_treeRowBuilderDelegate.rowCount = _activeNodes.length`.
3. In `TwoDimensionalChildBuilderDelegate`, setting `rowCount` (`maxYIndex`) to the same value (e.g. `1 -> 1`) is a no-op and does **not** call `notifyListeners()`.
4. `_treeRowBuilderDelegate` is not recreated or notified, so `TwoDimensionalViewport` skips building the new cells and silently displays stale widgets from the old tree.

To fix this, as discussed in point 3 earlier, we should **always recreate the delegate on `didUpdateWidget`** (or dispose and recreate `_treeRowBuilderDelegate` unconditionally whenever `didUpdateWidget` runs):

```dart
@override
void didUpdateWidget(TreeView<T> oldWidget) {
  super.didUpdateWidget(oldWidget);
  ...
  _unpackActiveNodes();
  _treeRowBuilderDelegate.dispose();
  _setTreeRowBuilderDelegate();
}
```

Could you update `didUpdateWidget` to recreate `_treeRowBuilderDelegate` unconditionally and add a regression test verifying that updating `widget.tree` to new nodes with the same active node count re-renders the new content?

ValentinVignal: Good catch! I removed the `if` in [Always recreate _treeRowBuilderDeleguate](https://github.com/flutter/packages/pull/11653/commits/f55b81fd9d3e39ebdafefc6c96c69b5e9110cf89) and added a regression test in [Add a regression test](https://github.com/flutter/packages/pull/11653/commits/ee8da12527799e57fbef65cd8ddb6d745bbeca26)

packages/two_dimensional_scrollables/lib/src/table_view/table.dart
```diff
@@ -222,17 +235,138 @@ class TableView extends TwoDimensionalScrollView {
...
+  /// {@macro flutter.rendering.RenderViewportBase.cacheExtent}
+  @Deprecated(
+    'Use scrollCacheExtent instead. '
+    'This feature was deprecated after v3.41.0-0.0.pre.',
+  )
+  final double? cacheExtent;
+
+  /// The alignment of the table within the viewport when there is extra space.
+  ///
+  /// Defaults to [Alignment.topLeft].
```
Piinks: This looks like a copy paste extra. :)

Piinks (commented):
LGTM with just a doc nit below!

packages/two_dimensional_scrollables/lib/src/table_view/table.dart
ValentinVignal: Oops, sorry about that, I removed it in [Delete copy paste extra](https://github.com/flutter/packages/pull/11653/commits/a0dd25f00edb6f9338e75b80602f682e462afcaa)

Piinks (approved):
LGTM

Piinks (approved):
LGTM

Piinks:
@ValentinVignal can you rebase this? That should fix CI so we can get this in! 🥳

ValentinVignal:
@Piinks There are some merge conflicts if I rebase; I'm trying a merge first to see if it is enough :)

Piinks:
> @Piinks There are some merge conflicts if I rebase; I'm trying a merge first to see if it is enough :)

I understand. Unfortunately we need to rebase this because the base commit is old, which would prevent it from landing.

ValentinVignal:
@Piinks I have rebased the branch :)

Piinks (approved):
Thank you so much! So excited for this to land!

PR #11658, opened 2026-05-06
[video_player] Convert top-level classes to Swift

stuartmorgan-g:
Converts the top-level plugin class, along with the native view factory that is used for platform view management, from Objective-C to Swift. The Pigeon interface for the plugin class has been split out into a separate definition file so that it can use the Swift Pigeon generator while the player-instance-level API can continue to use Obj-C for now.

Since this is the first use of Swift in the plugin, this includes separating the Swift Package into more modules; now instead of:
video_player_avfoundation (Obj-C) -> video_player_avfoundation_ios and _macos (both Obj-C)
it's
video_player_avfoundation (Swift) -> video_player_avfoundation_objc (Obj-C) -> video_player_avfoundation_ios and _macos (both Obj-C)

The Swift conversions were done via Gemini as a starting point, but I did a side-by-side comparison to the Obj-C to ensure that everything was actually preserved, and fixed issues in its conversion manually.

The PR also includes adding some NONNULL region annotations to Obj-C headers that were accidentally missing it, since I ran into one of the omissions once I was calling the API from Swift, and then I audited all of the headers.

Part of https://github.com/flutter/flutter/issues/119105

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/video_player/video_player_avfoundation/darwin/RunnerTests/VideoPlayerTests.swift
```diff
@@ -531,16 +497,15 @@ private let hlsAudioTestURI =
...
     let realObjectFactory = FVPDefaultAVFactory()
-    let videoPlayerPlugin = createInitializedPlugin(avFactory: realObjectFactory)
+    let videoPlayerPlugin = try createInitializedPlugin(avFactory: realObjectFactory)
 
-    var error: FlutterError?
-    let identifiers = videoPlayerPlugin.createTexturePlayer(
-      with: FVPCreationOptions.make(withUri: "", httpHeaders: [:]), error: &error)
-    #expect(error == nil)
-    let player = videoPlayerPlugin.playersByIdentifier[identifiers!.playerId] as! FVPVideoPlayer
+    // Provide a URI that is a valid URI, but not a video, to trigger a failure inside AVPlayer.
```
stuartmorgan-g, line 505: Unlike the other changes in this file, this isn't just a rote structural change for the switch from testing a Obj-C-Pigeon-generated API to a Swift-Pigeon-generated API. The old test used `""` as the URI, but that actually ends up creating a nil `NSURL`, which fails earlier than what this test needs now because we aren't internally passing along a nil `NSURL` on accident.

The goal of this test is to test an AVPlayer that fails to load a video, and for that we need a valid URI.

stuartmorgan-g:
> Iterating directly over `playersByIdentifier.values` while calling `disposeWithError` is unsafe because `disposeWithError` triggers the `onDisposed` callback, which removes the player from the dictionary. Mutating a dictionary during iteration causes a runtime crash in Swift.

I researched this when I was locally reviewing the Gemini-created migration, because I was surprised that it had removed the copy, and from what I found [it is safe](https://forums.swift.org/t/is-it-safe-to-remove-values-from-a-dictionary-while-iterating-over-its-keys/84600/5). The Gemini comment here is misleading; based on what I found, the code here is *not* actually mutating the dictionary during iteration *of the dictionary*, but instead during iteration of an independent view.

packages/video_player/video_player_avfoundation/darwin/RunnerTests/VideoPlayerTests.swift
```diff
@@ -41,32 +42,25 @@ private let hlsAudioTestURI =
...
+    let identifiers = try videoPlayerPlugin.createTexturePlayer(
+      options: CreationOptions(uri: mp4TestURI, httpHeaders: [:]))
     let player =
       videoPlayerPlugin.playersByIdentifier[identifiers.playerId] as! FVPTextureBasedVideoPlayer
 
     #expect(player.playerLayer.superlayer == view.layer)
   }
 
-  @Test func playerForPlatformViewDoesNotRegisterTexture() {
+  @Test func playerForPlatformViewDoesNotRegisterTexture() throws {
```
hellohuanlin, line 55: nice!

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/NativeVideoViewFactory.swift
```diff
@@ -0,0 +1,66 @@
...
+  #error("Unsupported platform.")
+#endif
+
+#if canImport(video_player_avfoundation_objc)
+  import video_player_avfoundation_objc
+#endif
+
+/// A factory class responsible for creating native video views that can be embedded in a
+/// Flutter app.
+class NativeVideoViewFactory: NSObject, FlutterPlatformViewFactory {
```
hellohuanlin: nit: `final`
is NSObject required?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/NativeVideoViewFactory.swift
```diff
@@ -0,0 +1,66 @@
...
+    playerByIdentifierProvider: @escaping (Int64) -> FVPVideoPlayer?
+  ) {
+    self.messenger = messenger
+    self.playerByIdentifierProvider = playerByIdentifierProvider
+    super.init()
+  }
+
+  #if os(macOS)
+    func create(
+      withViewIdentifier viewId: Int64,
```
hellohuanlin: nit: `viewID`

hellohuanlin: should we be consistent with either using `identifier` or `ID`?

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/NativeVideoViewFactory.swift
```diff
@@ -0,0 +1,66 @@
...
+    func create(
+      withFrame frame: CGRect,
+      viewIdentifier viewId: Int64,
+      arguments args: Any?
+    ) -> FlutterPlatformView {
+      return createNativeVideoView(arguments: args as! PlatformVideoViewCreationParams)
+    }
+  #endif
+
+  func createArgsCodec() -> FlutterMessageCodec & NSObjectProtocol {
```
hellohuanlin, line 53: maybe just `argsCodec` as it doesn't actually create

hellohuanlin (approved):
LGTM

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/NativeVideoViewFactory.swift
stuartmorgan-g: I think many of the engine's Obj-C APIs are poorly named, but changing them is difficult. If the iOS team is willing to maintain multiple versions of a bunch of APIs for a long time horizon, so that we can eventually move toward better names, I'd definitely be on board with that, it's just not been something I've had a chance to prioritize.

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/NativeVideoViewFactory.swift
stuartmorgan-g: > nit: `final`

Done.

> is NSObject required?

Yes; you can't have a Swift class that implements an Obj-C protocol without also inheriting from NSObject (it's a compiler error).

packages/video_player/video_player_avfoundation/darwin/video_player_avfoundation/Sources/video_player_avfoundation/NativeVideoViewFactory.swift
stuartmorgan-g: I changed it to `viewIdentifier`. Sadly this poor Obj-C variable naming was [straight from our header](https://github.com/flutter/flutter/blob/cb6eaf798aee0b908c76fe5bcaf8888a5afe677c/engine/src/flutter/shell/platform/darwin/ios/framework/Headers/FlutterPlatformViews.h#L42).

PR #11669, opened 2026-05-07
Decoupling Material and Cupertino (with dart fixes)

justinmc:
Superseded by https://github.com/flutter/packages/pull/11888

This is a single super mega PR attempting to migrate all Material and Cupertino code from flutter/flutter to material_ui and cupertino_ui.

To create this PR I used:

```
git filter-repo --path packages/flutter/lib/src/material --path packages/flutter/test/material --path examples/api/lib/material --path examples/api/test/material --path packages/flutter/lib/src/cupertino --path packages/flutter/test/cupertino --path examples/api/lib/cupertino --path examples/api/test/cupertino --path packages/flutter/lib/fix_data/fix_material --path packages/flutter/test_fixes/material --path packages/flutter/lib/fix_data/fix_cupertino.yaml --path packages/flutter/test_fixes/cupertino 
cd ../packages
git remote add source-origin ../flutter
git fetch source-origin
git merge source-origin/master --allow-unrelated-histories
```

I then cherry picked everything I could from the previous PR (https://github.com/flutter/packages/pull/11568).

For both Material and Cupertino, I'm importing:

 * The main source code (packages/flutter/lib/src/<material/cupertino>).
 * The widget/unit tests (packages/flutter/test/<material/cupertino>).
 * The examples (examples/api/lib/<material/cupertino>).
 * The example tests (examples/api/test/<material/cupertino>).
 * The dart fixes (packages/flutter/lib/fix_data).
 * The dart fix tests (packages/flutter/test_fixes).

### TODOs

 * [x] Do exports for Material and Cupertino.
 * [x] Confirm the directory structure (current plan: lib/src/ for now, reorganize after 1.0.0).
 * [x] Try installing and using the packages locally.
 * [x] Try running the tests locally (works except for tests with cross imports).
 * [x] Try running the example tests locally (work after migrating `package:flutter_api_samples` import).
 * [x] Verify examples work.
 * [x] Verify that `git blame` appears the same in both repos on various files. (Yes, but the SHAs are different, as expected.)
 * [x] Add auxiliary files for dart fixes and tests.
 * [x] Port imports of material/cupertino in places like examples/test to use the new packages.
 * [x] ~~Fix~~ Skip all test cross imports.
 * [x] Make sure the analyzer is working in the same exact way as it does in flutter/flutter. (https://github.com/flutter/packages/pull/11692)

### Open questions

   * Confirmed with @dkwingsmt and @QuncCccccc.
 * Do we like the examples setup? Separate package at material_ui_examples that points to material_ui via the pubspec `workspace`. Similar to how flutter/flutter's flutter_api_samples works.
 * Are we cool with skipping cross imports tests and fixing them later? This approach should not fracture commits because the commits were made in flutter/flutter after the code was copied.

### How to try using these packages

Add the local packages to your project:

```
flutter pub add cupertino_ui --path ../../packages/packages/cupertino_ui
flutter pub add material_ui --path ../../packages/packages/material_ui  
```

Then import the packages instead of the libraries in the SDK:

```diff
- import 'package:flutter/material.dart';
- import 'package:flutter/cupertino.dart';
+ import 'package:material_ui/material_ui.dart';
+ import 'package:cupertino_ui/cupertino_ui.dart';
```

### How to review this PR

 * Do read the PR description at the top, including Open Questions.
 * Don't read all 600k+ lines changed! Ignore the main source code and test files. The only thing I changed in there is to change imports from `flutter/material.dart` to `material_ui/material_ui.dart` and the equivalent for Cupertino.
 * Do look closely at the file tree browser on the left side of the "Files Changed" tab, though. Check the directory structure of material_ui and cupertino_ui and make sure it looks good to you.
 * Do take a look at the small config files like pubspecs, analysis_options, etc.
 * See the examples and their tests now located in material_ui/material_ui_examples and cupertino_ui/cupertino_ui_examples.
 * See the data driven fixes in lib/fix_data and their tests in test_fixes.

And let me know if there is any code that's not in this PR that you expected to be. It will be best to get everything we can in this one PR rather than following up later in order to prevent fragmenting commits.

### Resources

 * Closed PR where i tried to import everything but forgot dart fixes.
    - https://github.com/flutter/packages/pull/11568
 * Close PR where I tried to import all of Material, before realizing that I would create different SHAs between that PR and a future Cupertino PR.
   * https://github.com/flutter/packages/pull/11549


justinmc:
@Piinks I've created a new PR to include the dart fixes, which was easy enough. I spent some time cleaning up the dart fixes so they use the new imports. As of right now all the tests pass:

``` bash
cd packages/material_ui/test_fixes # Or packages/cupertino_ui/test_fixes
dart fix --compare-to-golden
```

However I did remove some fixes that seemed to be related exclusively to the Widgets library. Those are still covered in flutter/flutter.

stuartmorgan-g:
You may need to push a new commit here? The CICD label check seems to be stuck.

justinmc:
How to review this PR:

 * Do read the PR description at the top, including Open Questions.
 * Don't read all 600k+ lines changed! Ignore the main source code and test files. The only thing I changed in there is to change imports from `flutter/material.dart` to `material_ui/material_ui.dart` and the equivalent for Cupertino.
 * Do look closely at the file tree browser on the left side of the "Files Changed" tab, though. Check the directory structure of material_ui and cupertino_ui and make sure it looks good to you.
 * Do take a look at the small config files like pubspecs, analysis_options, etc.
 * See the examples and their tests now located in material_ui/material_ui_examples and cupertino_ui/cupertino_ui_examples.
 * See the data driven fixes in lib/fix_data and their tests in test_fixes.

And let me know if there is any code that's not in this PR that you expected to be. It will be best to get everything we can in this one PR rather than following up later in order to prevent fragmenting commits.

*Edit*: I added this to the top PR description as well since this PR tends to get collapsed.

justinmc:
Looks like the CICD label check is stuck. I'm going to try a rebase. Hoping that it doesn't mess with the history too much here.

justinmc:
No luck rebasing in GitHub or on the command line due to confusing conflicts. I'll try a merge commit (but I've tried that already above), and otherwise look for any other options.

packages/material_ui/pubspec.yaml
```diff
@@ -4,12 +4,19 @@ version: 0.0.2
 publish_to: none
 repository: https://github.com/flutter/packages/tree/main/packages/material_ui
 issue_tracker:  https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A%20material%20design%22
+publish_to: 'none'
```
stuartmorgan-g: There are now two copies of this line.

packages/cupertino_ui/example/analysis_options.yaml
stuartmorgan-g, line 1: You'll need to add the new packages to script/configs/custom_analysis.yaml so CI doesn't fail on these local files.

Piinks:
Looks like the cupertino examples app needs the same test config set up for flutter_goldens to use the one here instead of the one from the framework:

```
Because cupertino_ui_examples depends on flutter_goldens from sdk and cupertino_ui depends on flutter_goldens from path, version solving failed.
Failed to update packages.
```

chunhtai (commented):
I am not sure if the following is required for merging the first pr, so feel free to ignore them and do it later.

1. The example should have a lib/main.dart which should be the minimum example to get started, and will be show cased in pub like this one https://pub.dev/packages/go_router/example

2. both packages are lacking a readme and changelog.md

3. For dartfix test, you can create a script similar to https://github.com/flutter/packages/blob/main/packages/go_router/tool/run_tests.dart

`tool/run_tests.dart` is a special file that will be picked up by the ci

chunhtai:
@stuartmorgan-g  for this one https://github.com/flutter/packages/blob/main/packages/go_router/tool/run_tests.dart it seems may be used by multiple packages, is there a place to put the share code?

QuncCccccc:
> Are we sure about the directory structure?

The `material_ui` directory structure looks good to me:)!

stuartmorgan-g:
> @stuartmorgan-g for this one https://github.com/flutter/packages/blob/main/packages/go_router/tool/run_tests.dart it seems may be used by multiple packages, is there a place to put the share code?

There's very little shared code in the various custom test scripts (and some of it could be eliminated in favor of using `ProcessStartMode.detachedWithStdio` instead of manually setting up streaming), so we just duplicate it.

packages/cupertino_ui/example/lib/activity_indicator/cupertino_activity_indicator.0.dart
```diff
@@ -0,0 +1,75 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
```
stuartmorgan-g: All of the files will need to change the first line to
```
```

(different year, no period or "All right reserved.") to match the license standardization of this repo.

packages/cupertino_ui/example/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+The examples can be run individually by just specifying the path to the example
+on the command line (or in the run configuration of an IDE).
+
+For example (no pun intended!), to run the first example from the
+`CupertinoActivityIndicator` class in Chrome, you would run it like so from the
+[api](.)
+directory:
+
+```
```
stuartmorgan-g: All the READMEs need language annotations on all the blocks to pass CI.

You'll probably also need to temporarily add the two new packages to https://github.com/flutter/packages/blob/main/script/configs/temp_exclude_excerpt.yaml

packages/cupertino_ui/example/lib/activity_indicator/cupertino_activity_indicator.0.dart
justinmc: Done for material in https://github.com/flutter/packages/pull/11669/changes/1d32f6dd02c7899e35888fc53d4d754ab394b5d8 and cupertino in https://github.com/flutter/packages/pull/11669/changes/eeefb93d4ef21b5a778d0bfa1ee6be49c3d0a2b5

justinmc:
@chunhtai Replying to your comments:

 > The example should have a lib/main.dart which should be the minimum example to get started, and will be show cased in pub like this one https://pub.dev/packages/go_router/example

I've added this to be tracked in https://github.com/flutter/flutter/issues/182568 and will do it after this PR lands, in order to avoid bloating this very large PR.

 > both packages are lacking a readme and changelog.md

They both have them, but they were added prior to this PR. I'll update the READMEs before publishing.

 > For dartfix test, you can create a script similar to https://github.com/flutter/packages/blob/main/packages/go_router/tool/run_tests.dart

Thanks for the explanation. I may also delay this until after this PR for the same reasons.

packages/cupertino_ui/example/README.md
justinmc: @stuartmorgan-g I get a 404 for the file you linked to, could you double check it?

stuartmorgan-g: I landed a tooling change since that comment; you now need a file like https://github.com/stuartmorgan-g/packages/blob/d2491e72e7db649e682aaeb5ff205f7f70f9f49b/packages/in_app_purchase/in_app_purchase/ci_config.yaml in each package.

Piinks:
Looks like vector_math might be one of the last snags here:

```
Running command: "flutter pub get" in /b/s/w/ir/x/w/packages/packages/cupertino_ui
Resolving dependencies...
Note: vector_math is pinned to version 2.2.0 by flutter_test from the flutter SDK.
See https://dart.dev/go/sdk-version-pinning for details.


Because every version of material_ui from path depends on vector_math ^2.3.0 and every version of flutter_test from sdk depends on vector_math 2.2.0, material_ui from path is incompatible with flutter_test from sdk.
So, because cupertino_ui depends on both flutter_test from sdk and material_ui from path, version solving failed.
Failed to update packages.
```

justinmc:
So close! The package was rolled from 2.2.0 to 2.3.0 here: https://github.com/flutter/flutter/pull/185274. That was after the code freeze, so I should have used 2.2.0 in this PR instead of grabbing the version I saw on master like I did. At least that's my best guess, let's see :crossed_fingers:

justinmc:
I've merged in the updated line length here, but I'm currently waiting on the rest of the analysis options updates to land in https://github.com/flutter/packages/pull/11823.

Piinks:
Analysis looks clear! 🎉  Looks like timeouts (maybe the shard time needs to be increased since we are adding so many tests?), and potentially the reduced-test-set tag in test files with golden file tests. In flutter/flutter, the reduced-test-set tag is used to filter out hermetic tests we do not need to run on all platforms. I think we can remove it in the context of flutter/packages?

justinmc:
Wow last commit of the day and it finally passed the analyzer 🤩 . Though it does look like there are still a few things to format in the repo_checks failure. I will check with Stuart on Monday about these infra questions.

stuartmorgan-g:
> Looks like timeouts (maybe the shard time needs to be increased since we are adding so many tests?)

How were these tests managed in flutter/flutter? In this repo the way the sharding works is that each shard runs a subset of the packages, but in these runs essentially the entire run was just each of the two packages (`cupertino_ui` had <2 minutes of other tests run before it, `material_ui` had none). Both packages' unit tests took basically the whole hour and didn't finish, so if that's an expected amount of unit test time for these packages to take, that's a significant problem.

The current runs don't look like that's the case though. Both appear to have burned almost all of their time with 10-minute time jumps in the logged timestamps, with these failure messages:
```
  TimeoutException after 0:10:00.000000: Test timed out after 10 minutes.
  dart:isolate  _RawReceivePort._handleMessage
```

justinmc:
At some point during all of the analyzer fixes I made, I ended up adding `await` in a bunch of places in the tests. I wonder if some of those are never completing and causing the timeouts. I'll look into it.

justinmc:
Indeed these tests are timing out locally, too. I will investigate a fix that solves the required `await` lint but doesn't cause the tests to hang.

Some (maybe all) of the commits that added `await`s are:

https://github.com/flutter/packages/pull/11669/changes/dfeb5798accb2f7fc75cc77957765ebd12e51040
https://github.com/flutter/packages/pull/11669/changes/96e2712220650065438530e43647c26b6bbff211
https://github.com/flutter/packages/pull/11669/changes/2ae12c53087e6b419cec48cf0b1a15d2967f6d3f
https://github.com/flutter/packages/pull/11669/changes/1b799417b67c322f5c5d176598022c3f815950b5

stuartmorgan-g:
> I will investigate a fix that solves the required `await` lint but doesn't cause the tests to hang.

The behavior-preserving change that addresses the lint is to use `unawaited` rather than to add `await`. If you want to quickly address the analyzer warning while avoiding the possibility of regressions, you should revert any `await`s you added, use `unawaited` everywhere, and if you're not sure if some of them are correct, file an issue to audit them after the fact.

stuartmorgan-g:
Actually, the *really* easy way to land now would be to revert the `await`s, and temporarily turn that specific lint off in the local analysis options for these packages (ideally with a TODO linking to an issue to undo that once there's time to audit).

justinmc:
> Actually, the _really_ easy way to land now would be to revert the `await`s, and temporarily turn that specific lint off in the local analysis options for these packages (ideally with a TODO linking to an issue to undo that once there's time to audit).

I didn't see this reply because I was heads down porting all of my `await`s to `unawaited`s haha. Hopefully I got them all. If not and it's a pain to fix then I'll consider this.

justinmc:
Currently waiting for https://github.com/flutter/packages/pull/11874 to land, which should fix the analyzer failure.

justinmc:
@stuartmorgan-g It looks like the test command fails on web because of my skipped and ignored files. So some of the test files import a non-existent file (for example see material_ui/test/text_button_test.dart which imports ../widgets/semantics_tester.dart). I have skipped these tests and ignored them in analysis_options.yaml. This works fine for `flutter test`. But for `flutter test --platform=chrome`, it still reads those skipped files and so fails to compile.

Any ideas? I do plan to fix all of those skipped tests, but I would prefer to do it after this PR lands as it's a non-trivial amount of work (see https://github.com/flutter/flutter/issues/177028).

stuartmorgan-g:
I know very little about the internals of `flutter test`, or how `--platform=chrome` changes the implementation logic.

Landing files in test/ with invalid code seems questionable in general though. I assume the reason you don't want to just drop them from the PR and add them later is to preserve history for them; could you do something like move those files from `test/` to `temporarily_disabled_test/`, so that they import with history, but are not active at all, and then move them later once they are fixed up?

stuartmorgan-g:
Also, I forget whether I mentioned this before or not, but when you are happy with the state of the PR you should do local interactive rebasing to get it down to:
- the merge commit
- a very small number of follow-up commits from you (possibly even one) with very clear commit messages describing what was done

and then force-push that back to the PR.

Because this will land as a merge commit, not a squash, every commit in the PR will become a permanent part of the commit history of the repo, so we don't want a lot of incremental cleanup commits, especially those that were reverted in a later commit, or were testing something in CI.

justinmc:
> move those files from test/ to temporarily_disabled_test/

Good idea, I'll do that. Indeed I want to keep these files' history.

>  when you are happy with the state of the PR you should do local interactive rebasing

Will do, I forgot that all of my messy commits will go directly into the repo history :grin:

justinmc:
I've created a new branch to preserve my original commits from this PR, in case I need to remember what I did: https://github.com/flutter/packages/tree/decoupling-material-and-cupertino-and-fixes-preserve-commits

When CI goes green I will do the rebase and squash all/most of my commits.

justinmc:
There are a few more examples tests that are failing when run on the web, I'm working on fixing them.

justinmc:
Alright we're passing CI, time to squash all of my commits.

*Update*: This is harder than I thought because I've got a few merge commits in the middle of my commits.

justinmc:
@stuartmorgan-g I've managed to reduce my history to a squashed commit plus a bunch of merge commits. I assume this is not what you had in mind, and we need to have just a single squashed commit on top of all of the flutter/flutter history. This is hard because I had several merge commits in the course of working on this PR, and I can't squash my changes with those without losing the ability to merge with main.

I'm also unable to rebase the PR with main, because it ends up with conflicts in the many flutter/flutter commits that appear infeasible to resolve manually.

I think I will have to recreate this PR on the latest main and cherry pick in my changes, then squash them. That's my plan for tomorrow.

stuartmorgan-g:
Hm, incremental merge commits would make that tricky.

Rather than cherry-picking and squashing, which may or may not have a bunch of incremental conflicts, can you just `git diff 487f0f06402b652af21fd65302bd3b455df1604b packages/*_ui > changes.patch`, then apply that patch onto a fresh merge?

justinmc:
The problem with that approach is it will include changes that came in from the merge commits, which will then cause problems when we try to merge the PR.

But there should only be a few changes coming from the merge commits if I scope the diff to the *_ui packages... I'm going to try manually excluding those things from the diff. Then I'll need to do one final merge commit and hopefully we'll be good.

chunhtai:
Can you use rebase instead of merge to update the base? It will change the sha for the old commit pull from the flutter/flutter, but I think it is already different from the commit in flutter/flutter when you do filter-repo

stuartmorgan-g:
> But there should only be a few changes coming from the merge commits if I scope the diff to the *_ui packages

I forgot there had been some changes to those packages in the main repo; I was thinking scoping to those directories would make it apply cleanly.

stuartmorgan-g:
> Can you use rebase instead of merge to update the base?

In my experience trying to rebase *across* existing merge commits gets very messy and confusing.

justinmc:
> Can you use rebase instead of merge to update the base? It will change the sha for the old commit pull from the flutter/flutter, but I think it is already different from the commit in flutter/flutter when you do filter-repo

I'm hesitant to rebase across the 7k flutter/flutter commits. I tried it before and it gave me tons of conflicts that I had to manually resolve. My conclusion was that I'm not going to be able to complete that rebase in a reasonable amount of time. It might be faster to just re-copy all of those commits from flutter/flutter fresh. But if I get desperate I might try a rebase again.

justinmc:
Closing in favor of a recreate here: https://github.com/flutter/packages/pull/11888

PR #11692, opened 2026-05-11
Updated analysis_options line width from flutter/flutter

justinmc:
It seems like we're supposed to keep analysis_options.dart in sync with the one in flutter/flutter, but we haven't done it since 2023. This PR updates it. The most disruptive change appears to be the increased page width.

I'm inspired to update this after working on https://github.com/flutter/packages/pull/11669, which imports flutter/flutter code that is formatted according to the latest rules. Better to update the rules than to format the new code back to the old rules.

*Update*: This only updates the line width. Other changes will follow.


analysis_options.yaml
```diff
@@ -158,8 +158,7 @@ linter:
     # - prefer_constructors_over_static_methods # far too many false positives
     - prefer_contains
     # - prefer_double_quotes # opposite of prefer_single_quotes
-    # - prefer_expression_function_bodies # conflicts with ./docs/contributing/Style-guide-for-Flutter-repo.md#consider-using--for-short-functions-and-methods
-    - prefer_final_fields
+    # - prefer_expression_function_bodies # conflicts with https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#consider-using--for-short-functions-and-methods    - prefer_final_fields
```
stuartmorgan-g: Why was prefer_final_fields pulled into this line and thus removed?

analysis_options.yaml
```diff
@@ -128,12 +137,12 @@ linter:
     - noop_primitive_operations
     - null_check_on_nullable_type_parameter
     - null_closures
-    # - omit_local_variable_types # conflicts with specify_nonobvious_local_variable_types
-    - omit_obvious_local_variable_types
+    # - omit_local_variable_types # superset of omit_obvious_local_variable_types
+    - omit_obvious_local_variable_types # not yet tested
```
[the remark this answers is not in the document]
stuartmorgan-g: Let's fix this.

analysis_options.yaml
```diff
@@ -190,12 +198,14 @@ linter:
...
     - specify_nonobvious_local_variable_types
     - specify_nonobvious_property_types
+    - strict_top_level_inference
     - test_types_in_equals
     - throw_in_finally
     - tighten_type_of_initializing_formals
     - type_annotate_public_apis
     - type_init_formals
     - type_literal_in_constant_pattern
+    # - unawaited_futures # too many false positives, especially with the way AnimationController works
```
stuartmorgan-g: This is on, it's just in the wrong place. It's two lines down.

stuartmorgan-g (changes requested):

stuartmorgan-g:
I'm not sure what's going on here, but most of the tests never ran. If you are changing the line length, you're going to have to reformat the entire repo in this PR otherwise CI will (when it actually runs) fail.

analysis_options.yaml
justinmc: Typo, thanks for catching that!

justinmc:
@stuartmorgan-g Is that just a `dart format .`? I've now done that in https://github.com/flutter/packages/pull/11692/commits/4210b407a98fa4aa9425365e80e4fe311db116df.

The reason why I'm doing this is that the code imported in https://github.com/flutter/packages/pull/11669 is formatted according to the latest flutter/flutter analysis_options.yaml

stuartmorgan-g:
Changelog/version override: autoformat doesn't affect clients.

stuartmorgan-g:
The analysis options look good.

What is going on with LUCI tests today ☹️  Did you run the repo tool's `analyze` command locally to make sure everything passes with the new options, since we don't have CI output for some reason?

stuartmorgan-g (approved):
analysis_options.yaml changes LGTM.

You'll need to revert the autoformat commit, sync main, and then re-apply it. Most PRs will cause conflicts so you'll want to do it at a time when there aren't PRs landing (e.g., evening or early morning); we don't have a way to artificially close this tree AFAIK.

No need to get a re-review on format changes.

stuartmorgan-g:
@justinmc Do you want me to try to get this updated, through tests, and landed sometime morning my time when there's not much tree activity, to minimize the chance of conflicts?

justinmc:
@stuartmorgan-g Yes please if you have time! I keep forgetting about this.

justinmc:
Actually I'm going to go ahead and attempt this right now. If I fail I'll reach out for help!

stuartmorgan-g:
I was actually just looking at this; a locally analysis run is showing a bunch of failures.

I think we should split this into two PRs: one to pull in the new options other than line width, fixing violations, and one that just does the reformat.

stuartmorgan-g:
I'll work on analysis changes in a new PR, and you can downscope this to just the autoformat so we can try to get that landed this morning before anything else comes in.

packages/rfw/lib/src/flutter/runtime.dart
```diff
@@ -410,7 +418,9 @@ class Runtime extends ChangeNotifier {
         return _cachedConstructors[fullName] = _ResolvedConstructor(fullName, constructor);
       }
     } else {
-      assert(library is Null); // ignore: prefer_void_to_null, type_check_with_null, https://github.com/dart-lang/sdk/issues/47017#issuecomment-907562014
+      assert(
+        library is Null,
+      ); // ignore: prefer_void_to_null, type_check_with_null, https://github.com/dart-lang/sdk/issues/47017#issuecomment-907562014
```
stuartmorgan-g: The reformat broke this `// ignore`; you'll need to move it manually.

justinmc: Thanks!

Piinks:
Looks like go_router_builder/example needs formatting here.

justinmc:
It looks like CI is not picking up the line length change that I added to the root analysis_options.yaml for some reason. For example

```
-    GoRouteData.$route(path: 'int-route/:requiredIntField', factory: $IntExtensionRoute._fromState),
-    GoRouteData.$route(path: 'num-route/:requiredNumField', factory: $NumExtensionRoute._fromState),
+    GoRouteData.$route(
+      path: 'int-route/:requiredIntField',
+      factory: $IntExtensionRoute._fromState,
+    ),
+    GoRouteData.$route(
+      path: 'num-route/:requiredNumField',
+      factory: $NumExtensionRoute._fromState,
+    ),
```

It's telling me to go back to a shorter line length. Maybe it's because those are generated files and they should be ignored by the formatter?

stuartmorgan-g:
Note: Normally we would publish updates for README changes (which `update-excerpts` triggered), but we've skipped that for mass reformats since the code is correct either way, so there's no user-driven need to publish the changes in a timely manner.

stuartmorgan-g:
`rfw` has some files that had opted out of autoformat enforcement. I wasn't going to worry about that since I think we should revisit whether we support that at all now that flutter/flutter autoformats (the context here was that Hixie didn't want the formatter enforced on `rfw` back when the framework still didn't autoformat), but it turns out that the package has some custom coverage checking in its package-specific tests, and it has special logic based on comment annotations that the autoformatter moves. So I'm reverting the autoformat changes in the opted-out files, and I'll file an issue (edit: https://github.com/flutter/flutter/issues/187508) to eventually consider removing the opt-out support (and then changing the coverage calculation logic).

stuartmorgan-g:
`go_router_builder` is a conflict between what the autoformatter wants, and what the generator generates; there's a unit test that runs the generation and makes sure nothing changed, so currently either that test fails or the format check files. I'm sure we hit this with the last big reformat, and I'm trying to figure out how we handled it then.

justinmc:
Thank you for the help here! I'll plan to wait for what you find out for go_router_builder. If you want me to apply the diffs from the failure or anything else in the meantime let me know.

stuartmorgan-g:
It turns out this case was different from the previous case, but chunhtai pointed me to the actual problem and solution. For posterity (it's in comments in the PR diffs, but a comment here may be more discoverable for a future person searching old PRs like I was): the issue this time was that `go_router_builder` is using the `dart_style` package, and it uses code-based configuration rather than reading local analysis options. So `go_router_builder` makes code that uses the default of 80 columns, while we are expecting files in the repo to be formatted with 100 now.

So what was happening was:
- The repo format check would fail if `go_router_builder/example` used 80-column format
- `expectBuildClean` in `ensure_build_test.dart` would fail if `go_router_builder/example` used 100-column format, because all the checked-in, `go_router_builder`-generated files are the goldens for that test.

The simple solution chunhtai pointed out was that we can give `go_router_builder/example` custom analysis settings that set just that sub-project to 80 columns.

PR #11701, opened 2026-05-13
[flutter_hook_config] Add the flutter_hook_config package

bdero:
Towards flutter/flutter#186299

This came out of an exchange with @dcharkes here: https://github.com/flutter/flutter/pull/186300#issuecomment-4430396023

Adds `flutter_hook_config`, a `package:hooks` protocol extension owned by Flutter.

The intent: `flutter_tools` constructs a `FlutterExtension` and passes it to every build and link hook invocation, populating it with engine host-tool paths resolved through `Artifacts.getHostArtifact` (so the `--local-engine` override applies, same as for flutter_tools' own build steps). Build hooks then read those paths through a typed accessor instead of re-deriving them by walking the SDK cache directory, which is what packages like `flutter_gpu_shaders` do today (and which misses `--local-engine`).

What a Flutter build hook looks like with it:

```dart
import 'package:flutter_hook_config/flutter_hook_config.dart';
import 'package:hooks/hooks.dart';

void main(List<String> args) async {
  await build(args, (input, output) async {
    if (input.config.buildForFlutter) {
      final impellerc = input.config.flutter.impellerc; // absolute Uri
      // ... invoke impellerc to compile a shader bundle ...
    } else {
      // Plain `dart` build, or a Flutter SDK without this extension:
      // fall back to the hook's own tool discovery.
    }
  });
}
```

v1 exposes `impellerc` and `libtessellator`; more host artifacts can be added later as null-checked fields. Pure Dart, depends on `hooks: ^1.0.0`, starts at `0.1.0` while the shape stabilizes. The `flutter_tools` side lands in a separate flutter/flutter PR and depends on this being published first.

## Pre-Review Checklist

- [x] I signed the [CLA].


.github/labeler.yml
```diff
@@ -38,6 +38,11 @@
     - any-glob-to-any-file:
       - packages/file_selector/**/*
 
+'p: flutter_hook_config':
```
bdero, line 41: Note: label needs to be created

bdero:
Note for shipping this: `flutter_hook_config` is a brand-new package, so on its first publish the `release` workflow will publish it under the bot publisher account rather than the `flutter.dev` verified publisher. Someone on the Flutter team with publisher-account access will need to do the one-time transfer from the package's Admin tab on pub.dev afterward (per the [packages release process docs](https://github.com/flutter/flutter/blob/master/docs/ecosystem/release/README.md#new-packages); access is tracked internally at b/191674407). Happy to coordinate in `#hackers-ecosystem`.

Also flagging: the top-level `README.md` row and the `issue_tracker` URL reference `pub.dev/packages/flutter_hook_config` and the `flutter/flutter` label `p: flutter_hook_config`, neither of which exists yet, so the badges/links will 404 until the package publishes and the label is created.

packages/flutter_hook_config/AUTHORS
```diff
@@ -0,0 +1,6 @@
+# Below is a list of people and organizations that have contributed
+# to the Flutter project. Names should be added to the list like so:
+#
+#   Name/Organization <email address>
+
+Google Inc.
```
bdero, line 6: Added this following the template of other new packages. I assume this is correct but not 100% sure. I'm CLA signed of course, but I no longer work at Google.

packages/flutter_hook_config/lib/src/config.dart
```diff
@@ -0,0 +1,106 @@
...
+///
+/// This is intended for use by the Flutter SDK (`flutter_tools`) via
+/// [FlutterExtension]; ordinary hook authors do not call this directly.
+extension HookConfigBuilderFlutterConfig on HookConfigBuilder {
+  /// Records the Flutter-specific configuration on a build or link hook input.
+  ///
+  /// [impellerc] and [libtessellator] must be absolute file URIs.
+  void setupFlutter({required Uri impellerc, required Uri libtessellator}) {
+    assert(impellerc.isAbsolute, 'impellerc must be an absolute URI');
+    assert(libtessellator.isAbsolute, 'libtessellator must be an absolute URI');
```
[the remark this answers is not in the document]
bdero: improved in 097ace08eb.

packages/flutter_hook_config/lib/src/config.dart
```diff
@@ -0,0 +1,112 @@
...
+        'No Flutter-specific hook configuration is available. The hook is not '
+        'being invoked by a Flutter SDK, or the Flutter SDK predates '
+        'flutter_hook_config support. Check `config.buildForFlutter` first.',
+      );
+    }
+    return FlutterConfig._(extensionJson);
+  }
+
+  Map<String, Object?>? get _flutterExtensionJson {
+    final Object? extensions = json['extensions'];
```
dcharkes: The json keys should probably all be consts.

(On dart-lang/native we actually use a json schema as the source of truth and generate a Dart parser based off it, but that json_syntax_generator package is not published so we can't use it here.)

packages/flutter_hook_config/lib/src/config.dart
```diff
@@ -0,0 +1,112 @@
...
+  FlutterConfig._(this._json);
+
+  final Map<String, Object?> _json;
+
+  /// The absolute path to the `impellerc` offline shader compiler that ships
+  /// with the engine artifacts of the invoking Flutter SDK.
+  ///
+  /// When the SDK is invoked with `--local-engine`, this points at the locally
+  /// built `impellerc`, matching the engine the app is being built against.
+  Uri get impellerc => _filePath(_impellercKey);
```
dcharkes, line 49: Are these two related? Do they change together? If yes they should maybe be nested in an object together.

Also, does _every_ Flutter have impeller? E.g. With third-party forks of Flutter is there ever a possibility there's no impeller? If yes, then that wrapper object should be nullable. And if it's not null both fields are non-nullably provided.

Do we need a some kind of version number for this? e.g. if any of these files change or the underlying formats change, you'd want the hook to be rerun.

packages/flutter_hook_config/pubspec.yaml
```diff
@@ -0,0 +1,21 @@
...
+  paths (such as impellerc) to build and link hooks.
+version: 0.1.0
+repository: https://github.com/flutter/packages/tree/main/packages/flutter_hook_config
+issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A%20flutter_hook_config%22
+
+environment:
+  sdk: ^3.9.0
+
+dependencies:
+  hooks: ^1.0.0
```
dcharkes, line 13: Side note: I'm publishing a 2.0.0 soon, because we need to add a method to the `ProtocolExtension` (https://github.com/dart-lang/native/pull/3358).

packages/flutter_hook_config/lib/src/config.dart
bdero: > Also, does every Flutter have impeller? With third-party forks of Flutter is there ever a possibility there's no impeller? If yes, then that wrapper object should be nullable. And if it's not null both fields are non-nullably provided.

`impellerc` is used for both Impeller and Skia. Even if one were to do the heavy work of ripping out Impeller from the engine (since we don't even ship Skia on iOS), features like the Fragment Program API/RuntimeEffect use `impellerc` to compile shaders regardless of the graphics backend (`impellerc` transpiles to SkSL for Skia).

The `impellerc` dependency is baked elsewhere in `flutter_tools` already.

packages/flutter_hook_config/lib/src/config.dart
bdero: Alrighty, done in be4ada98c1.

packages/flutter_hook_config/lib/src/config.dart
bdero: > Do we need a some kind of version number for this? e.g. if any of these files change or the underlying formats change, you'd want the hook to be rerun.

Hmm yes that's a good point and we should think about this... Since the artifacts are per-engine build, maybe the right thing to do would be to add a `config.flutter.engineVersion`, so that the cache key gets invalidated whenever the engine artifacts change. I can just grab the engine revision from `Cache.engineRevision` in `flutter_tools`.

Does that seem right?

bdero: Added `engineVersion`.

stuartmorgan-g:
> The `flutter_tools` side lands in a separate flutter/flutter PR and depends on this being published first.

Circular repo dependencies are an anti-pattern that we don't want to make more of. I'm in the process of standing up CI for flutter/core-packages, so if this package is something that is used by flutter/flutter rather than depending on the Flutter SDK, it should wait until that repo is ready and then land there instead.

packages/flutter_hook_config/lib/src/config.dart
dcharkes: Yes that makes sense to me. For code assets we also pass in the target macos ios and android API versions, for exactly the same reason. 👍

dcharkes:
> > The `flutter_tools` side lands in a separate flutter/flutter PR and depends on this being published first.
> 
> Circular repo dependencies are an anti-pattern that we don't want to make more of. I'm in the process of standing up CI for flutter/core-packages, so if this package is something that is used by flutter/flutter rather than depending on the Flutter SDK, it should wait until that repo is ready and then land there instead.

That makes sense. For the hooks protocol and code assets they needed to be out of repo because they are shared across SDKs (Dart standalone, Flutter, and whatever other SDK is built on top of Dart). This package is tied to Flutter specifically, so co-locating it with the SDK itself SGTM!

(For my information: What is the reasoning behind a separate packages repo? In Dart we have both: packages in the SDK that are published to pub and packages on other repos that are pulled in via DEPS. And it's a per-package decision what's easier. E.g. third-party contributions are easier out-of-SDK. But if rolls can break the SDK that's harder again.)

stuartmorgan-g:
> This package is tied to Flutter specifically, so co-locating it with the SDK itself SGTM!

Can you elaborate on what you mean by "tied to Flutter"? I didn't see a Flutter dependency in the package.

flutter/core-packages isn't co-located with the SDK. It's a new repo that follows the structure of this one, but will contain packages used by flutter/flutter instead of packages that depend on the Flutter SDK.

> What is the reasoning behind a separate packages repo?

I don't know the original reason since it predates me, but I think that we get a lot of benefit from this structure. Eating our own dogfood and supporting `stable` channel as a first-class priority, for instance, both of which would be much harder (or at the very least much more confusing) in the same repo.

dcharkes:
> > This package is tied to Flutter specifically, so co-locating it with the SDK itself SGTM!
> 
> Can you elaborate on what you mean by "tied to Flutter"? I didn't see a Flutter dependency in the package.

The flutter_hook_config will be always not-provided for non-Flutter-SDKs that are built on top of Dart.

> flutter/core-packages isn't co-located with the SDK. It's a new repo that follows the structure of this one, but will contain packages used by flutter/flutter instead of packages that depend on the Flutter SDK.

Ah right! Yes that makes sense. Then it will function very very similarly to dart-lang/native rolling into both flutter/flutter and dart-lang/sdk. 👍 

> > What is the reasoning behind a separate packages repo?
> 
> I don't know the original reason since it predates me, but I think that we get a lot of benefit from this structure. Eating our own dogfood and supporting `stable` channel as a first-class priority, for instance, both of which would be much harder (or at the very least much more confusing) in the same repo.

👍

bdero:
Alrighty, just lemme know when the new repo is open for business and I'll be happy to move this over. :)

stuartmorgan-g:
flutter/core-packages is open for business now.

bdero:
Alright, moved this over to https://github.com/flutter/core-packages/pull/19!

PR #11736, opened 2026-05-19
[google_maps_flutter] Add MapColorScheme support to Android and iOS

### LINKED ISSUE #186737: [google_maps_flutter] colorScheme not wired up on Android and iOS
## Problem

The `colorScheme` parameter was added to the `GoogleMap` widget in `google_maps_flutter` 2.16.0 (PR flutter/packages#11280), but it only works on web. The Android and iOS platform plugins silently ignore it.

This is a follow-up to #176445 and #184064 (the latter was closed after I drafted an implementation; this issue tracks the actual PR).

## Native SDK support

Both mobile SDKs already support this:

- **Android**: `GoogleMap.setMapColorScheme(MapColorScheme.DARK)` — available in `play-services-maps:20.0.0+`
- **iOS**: `GMSMapView.overrideUserInterfaceStyle = .dark` — the documented color-scheme API per the Google Maps SDK for iOS

## Expected behavior

Setting `colorScheme: MapColorScheme.dark` on a `GoogleMap` widget with a cloud-based `mapId` should apply the dark color scheme on Android and iOS, matching the web behavior.

## Current behavior

The `colorScheme` parameter is accepted by the widget but has no effect on Android or iOS. The map always renders in the default (light) scheme.

## Proposed implementation

PR open at flutter/packages#11736.

snaeji:
## Description

Wires up `MapColorScheme` (light/dark/followSystem) to native APIs on Android and iOS. The platform interface and app-facing `google_maps_flutter` package already exposed this parameter (added in 2.16.0 for web), but the Android and iOS plugins silently ignored it.

- **Android**: `GoogleMap.setMapColorScheme()` is invoked with the matching `MapColorScheme` constant. Both creation-time and runtime updates (via `interpretMapConfiguration`) are supported.
- **iOS**: `GMSMapView.overrideUserInterfaceStyle` is set to `UIUserInterfaceStyleLight` / `Dark` / `Unspecified` — the documented API per the Google Maps SDK for iOS.

The change is mirrored across `google_maps_flutter_ios`, `google_maps_flutter_ios_sdk9`, `google_maps_flutter_ios_sdk10`, and `google_maps_flutter_ios_shared_code` via `dart tool/sync_shared_files.dart`.

The `colorScheme` doc comment in `google_maps_flutter` and `google_maps_flutter_platform_interface` is updated to reflect the new platform support (previously documented as "Web only").

Tests added:
- **Android**: `Convert.toMapColorScheme` JUnit tests for all three enum values; `interpretMapConfiguration_handlesColorScheme` dispatcher test mirroring `_handlesStyle`; Dart-side creation-params test asserting `colorScheme` flows through to the platform view.
- **iOS**: Parametrized Dart-side creation-params tests covering `light`, `dark`, `followSystem`, and `null`; matching tests propagated to sdk9 and sdk10 via the shared-code sync.

Fixes flutter/flutter#186737

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.20.0
+
+* Adds `colorScheme` support for cloud-based maps styling brightness.
+
 ## 2.19.12
 
 * Bumps the androidx group across 10 directories with 1 update.
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
@@ -577,6 +577,18 @@ class Convert {
     if (style != null) {
       sink.setMapStyle(style);
     }
+    final PlatformMapColorScheme colorScheme = config.getColorScheme();
+    if (colorScheme != null) {
+      sink.setMapColorScheme(toMapColorScheme(colorScheme));
+    }
+  }
+
+  static int toMapColorScheme(@NonNull PlatformMapColorScheme colorScheme) {
+    return switch (colorScheme) {
+      case LIGHT -> com.google.android.gms.maps.model.MapColorScheme.LIGHT;
+      case DARK -> com.google.android.gms.maps.model.MapColorScheme.DARK;
+      case FOLLOW_SYSTEM -> com.google.android.gms.maps.model.MapColorScheme.FOLLOW_SYSTEM;
+    };
   }
 
   /** Set the options in the given object to marker options sink. */
--- a/packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ConvertTest.java
+++ b/packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ConvertTest.java
@@ -522,6 +522,28 @@ public class ConvertTest {
   }
 
   @Test
+  public void toMapColorScheme_mapsAllPlatformValues() {
+    for (PlatformMapColorScheme value : PlatformMapColorScheme.values()) {
+      final int expected =
+          switch (value) {
+            case LIGHT -> com.google.android.gms.maps.model.MapColorScheme.LIGHT;
+            case DARK -> com.google.android.gms.maps.model.MapColorScheme.DARK;
+            case FOLLOW_SYSTEM -> com.google.android.gms.maps.model.MapColorScheme.FOLLOW_SYSTEM;
+          };
+      Assert.assertEquals(expected, Convert.toMapColorScheme(value));
+    }
+  }
+
+  @Test
+  public void interpretMapConfiguration_handlesColorScheme() {
+    final PlatformMapConfiguration config =
+        getMinimalConfigurationBuilder().setColorScheme(PlatformMapColorScheme.DARK).build();
+    Convert.interpretMapConfiguration(config, optionsSink);
+    verify(optionsSink, times(1))
+        .setMapColorScheme(com.google.android.gms.maps.model.MapColorScheme.DARK);
+  }
+
+  @Test
   public void interpretMapConfiguration_handlesUnboundedCameraTargetBounds() {
     final PlatformMapConfiguration config =
         getMinimalConfigurationBuilder()
@@ -859,6 +881,7 @@ public class ConvertTest {
     private @Nullable PlatformMarkerType markerType;
     private @Nullable String mapId;
     private @Nullable String style;
+    private @Nullable PlatformMapColorScheme colorScheme;
 
     public @NonNull PlatformMapConfigurationBuilder setCompassEnabled(@Nullable Boolean setterArg) {
       this.compassEnabled = setterArg;
@@ -982,6 +1005,12 @@ public class ConvertTest {
       return this;
     }
 
+    public @NonNull PlatformMapConfigurationBuilder setColorScheme(
+        @Nullable PlatformMapColorScheme setterArg) {
+      this.colorScheme = setterArg;
+      return this;
+    }
+
     public @NonNull PlatformMapConfiguration build() {
       return new PlatformMapConfiguration(
           compassEnabled,
@@ -1004,7 +1033,8 @@ public class ConvertTest {
           liteModeEnabled,
           Objects.requireNonNull(markerType),
           mapId,
-          style);
+          style,
+          colorScheme);
     }
   }
 }
--- a/packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
@@ -1242,6 +1242,26 @@ PlatformMarkerType _platformMarkerTypeFromMarkerType(MarkerType markerType) {
   };
 }
 
+PlatformMapColorScheme? _platformMapColorSchemeFromMapColorScheme(MapColorScheme? colorScheme) {
+  if (colorScheme == null) {
+    return null;
+  }
+  switch (colorScheme) {
+    case MapColorScheme.light:
+      return PlatformMapColorScheme.light;
+    case MapColorScheme.dark:
+      return PlatformMapColorScheme.dark;
+    case MapColorScheme.followSystem:
+      return PlatformMapColorScheme.followSystem;
+    // The enum comes from a different package, which could get a new value at
+    // any time, so provide a fallback that ensures this won't break when used
+    // with a version that contains new values.
+    // ignore: no_default_cases, unreachable_switch_default
+    default:
+      throw UnimplementedError('MapColorScheme "$colorScheme" has not been implemented');
+  }
+}
+
 PlatformMapConfiguration _platformMapConfigurationFromMapConfiguration(MapConfiguration config) {
   return PlatformMapConfiguration(
     compassEnabled: config.compassEnabled,
@@ -1267,6 +1287,7 @@ PlatformMapConfiguration _platformMapConfigurationFromMapConfiguration(MapConfig
     markerType: _platformMarkerTypeFromMarkerType(config.markerType ?? MarkerType.marker),
     mapId: config.mapId,
     style: config.style,
+    colorScheme: _platformMapColorSchemeFromMapColorScheme(config.colorScheme),
   );
 }
 
--- a/packages/google_maps_flutter/google_maps_flutter_ios/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter_ios/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.19.0
+
+* Adds `colorScheme` support for cloud-based maps styling brightness.
+
 ## 2.18.4
 
 * Fixes a potential compilation issue in tile downscaling.
--- a/packages/google_maps_flutter/google_maps_flutter_ios/example/ios/RunnerTests/GoogleMapsTests.m
+++ b/packages/google_maps_flutter/google_maps_flutter_ios/example/ios/RunnerTests/GoogleMapsTests.m
@@ -222,6 +222,56 @@
   XCTAssertEqual(cameraPosition.zoom, initialCameraPosition.zoom);
 }
 
+- (void)testSetMapColorScheme {
+  CGRect frame = CGRectMake(0, 0, 100, 100);
+  GMSMapViewOptions *options = [[GMSMapViewOptions alloc] init];
+  options.frame = frame;
+  options.camera = [[GMSCameraPosition alloc] initWithLatitude:0 longitude:0 zoom:0];
+  PartiallyMockedMapView *mapView = [[PartiallyMockedMapView alloc] initWithOptions:options];
+  FGMGoogleMapController *controller =
+      [[FGMGoogleMapController alloc] initWithMapView:mapView
+                                       viewIdentifier:0
+                                   creationParameters:[self emptyCreationParameters]
+                                        assetProvider:[[TestAssetProvider alloc] init]
+                                      binaryMessenger:[[StubBinaryMessenger alloc] init]];
+  FlutterError *error = nil;
+
+  // Light. Start from a different style so the assertion confirms a real change.
+  mapView.overrideUserInterfaceStyle = UIUserInterfaceStyleDark;
+  FGMPlatformMapConfiguration *lightConfig = [self emptyCreationParameters].mapConfiguration;
+  lightConfig.colorScheme =
+      [[FGMPlatformMapColorSchemeBox alloc] initWithValue:FGMPlatformMapColorSchemeLight];
+  [controller.callHandler updateWithMapConfiguration:lightConfig error:&error];
+  XCTAssertNil(error);
+  XCTAssertEqual(mapView.overrideUserInterfaceStyle, UIUserInterfaceStyleLight);
+
+  // Dark.
+  mapView.overrideUserInterfaceStyle = UIUserInterfaceStyleLight;
+  FGMPlatformMapConfiguration *darkConfig = [self emptyCreationParameters].mapConfiguration;
+  darkConfig.colorScheme =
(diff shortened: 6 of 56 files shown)


packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
```diff
@@ -1373,6 +1374,32 @@ PlatformMarkerType _platformMarkerTypeFromMarkerType(MarkerType markerType) {
...
+  switch (colorScheme) {
+    case MapColorScheme.light:
+      return PlatformMapColorScheme.light;
+    case MapColorScheme.dark:
+      return PlatformMapColorScheme.dark;
+    case MapColorScheme.followSystem:
+      return PlatformMapColorScheme.followSystem;
+  }
+
+  // The enum comes from a different package, which could get a new value at
```
LongCatIsLooong: I believe switch expressions / statements have to be exhaustive or it is gong to be a compile time error. The code isn't going to compile if there's an unhandled case. IOW the runtime error will never be reachable, unless you add a default case. But adding a new case to an existing enum should require a major version bump I think? So we don't really have to handle that at runtime.

LongCatIsLooong: Also this could just be an expression:
```dart
return switch (colorScheme) { 
  null => null,
  .light => .light,
  .dart => .dart,
  .followSystem => .followSystem,
};
```

LongCatIsLooong: Actually, why do we need 2 enums that represent the same thing? Can we just re-export one / making one of the an alias of the other? Then we don't need this since there's only one type under the hood.

LongCatIsLooong (approved):
The iOS implementation LGTM modulo https://github.com/flutter/packages/pull/11736/changes#r3270153752

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
stuartmorgan-g: > I believe switch expressions / statements have to be exhaustive or it is gong to be a compile time error.

Switch expressions do, not arbitrary switch statements.

> But adding a new case to an existing enum should require a major version bump I think?

We have not traditionally considered that to be a major version change. We may need to revisit that at some point as switch expressions become more common, but for now we don't (at least for platform interfaces).

> So we don't really have to handle that at runtime.

We do, for the specific goal of *not* requiring breaking the platform interface all the time.

stuartmorgan-g: > Actually, why do we need 2 enums that represent the same thing?

Because Pigeon definitions currently aren't allowed to import things from other files. And doing so in the case of an enum from another package would be extremely dangerous since it would allow for Dart code to change without the native side changing.

snaeji: Thanks both!

**_platformMapColorSchemeFromMapColorScheme**
The structure here mirrors _platformMarkerTypeFromMarkerType immediately above (line 1360).

I gather from the discussion above that the alternatives suggested (switch expression / collapsing the two enums) wouldn't fit here — the runtime fallback is intentional so new platform-interface enum values don't force a breaking change, and the Pigeon enum has to stay separate since Pigeon definitions can't import across files.

If there is something needed on my part or some restructuring needed please let me know.

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
```diff
@@ -577,6 +577,18 @@ static void interpretMapConfiguration(
     if (style != null) {
       sink.setMapStyle(style);
     }
+    final PlatformMapColorScheme colorScheme = config.getColorScheme();
+    if (colorScheme != null) {
+      sink.setMapColorScheme(toMapColorScheme(colorScheme));
+    }
+  }
+
+  static int toMapColorScheme(@NonNull PlatformMapColorScheme colorScheme) {
```
reidbaker, line 586: Should this method handle unknown. What do you want the unknown behavior to be an why.

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ConvertTest.java
```diff
@@ -521,6 +521,36 @@ public void interpretMapConfiguration_handlesStyle() {
     verify(optionsSink, times(1)).setMapStyle("foo");
   }
 
+  @Test
+  public void toMapColorScheme_light() {
```
reidbaker: I think instead of tests that verify the 3 conversions I think the tests should be higher value like looping over all of one type and making sure there is a value that is of the appropriate other type. Handling unknown values etc.

reidbaker (commented):
some comments added but directionally the android code looks ok

packages/google_maps_flutter/google_maps_flutter_android/android/src/main/java/io/flutter/plugins/googlemaps/Convert.java
snaeji: I just followed patterns that are already established in the code so I am pretty confident about this.

```
// Line 370
static int toMapType(@NonNull PlatformMapType type) {
  return switch (type) {
    case NONE -> MAP_TYPE_NONE;
    case NORMAL -> MAP_TYPE_NORMAL;
    case SATELLITE -> MAP_TYPE_SATELLITE;
    case TERRAIN -> MAP_TYPE_TERRAIN;
    case HYBRID -> MAP_TYPE_HYBRID;
  };
}

// Line 388
static MapsInitializer.Renderer toMapRendererType(@Nullable PlatformRendererType type) {
  // null-check, then…
  return switch (type) {
    case LATEST -> MapsInitializer.Renderer.LATEST;
    case LEGACY -> MapsInitializer.Renderer.LEGACY;
  };
}
```

packages/google_maps_flutter/google_maps_flutter_android/android/src/test/java/io/flutter/plugins/googlemaps/ConvertTest.java
snaeji: Good point. I replaced the tests with a single loop over PlatformMapColorScheme.values(). The expected mapping is defined via an exhaustive switch expression, so a new enum value forces a compile error in both the production conversion and this test.

snaeji:
Force-pushed to drop a Co-Authored-By trailer that was failing the Google CLA check. Diff is identical — only the commit message changed.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
stuartmorgan-g: > > I believe switch expressions / statements have to be exhaustive or it is gong to be a compile time error.
> 
> Switch expressions do, not arbitrary switch statements.

It turns out I was wrong here; *some* switch statements *do* have to be exhaustive now, including enums. The behavior changed from warning to error for enums several years ago and I'd never hit it, so we'll need to audit all the plugins for this construction and adjust it.

snaeji:
@stuartmorgan-g rebased on main to clear the conflicts, no behavior changes. I believe the open threads are addressed and it's ready to land when you are. Or let me know if there is something outstanding work on my end 🙏

reidbaker (approved):

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -1242,6 +1242,17 @@ PlatformMarkerType _platformMarkerTypeFromMarkerType(MarkerType markerType) {
...
 }
 
+PlatformMapColorScheme? _platformMapColorSchemeFromMapColorScheme(MapColorScheme? colorScheme) {
+  if (colorScheme == null) {
+    return null;
+  }
+  return switch (colorScheme) {
+    MapColorScheme.light => PlatformMapColorScheme.light,
+    MapColorScheme.dark => PlatformMapColorScheme.dark,
+    MapColorScheme.followSystem => PlatformMapColorScheme.followSystem,
```
stuartmorgan-g: This needs a `default` case (with an ignore to suppress the analyzer warnings, and a comment similar to what's on like 1214, minus the part about not using `default`) to avoid potential future breakage (per [this thread](https://github.com/flutter/packages/pull/11736#discussion_r3270085095)).

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.19.13
```
stuartmorgan-g: We generally consider new functionality to be a minor version bump.

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios/Sources/google_maps_flutter_ios/FGMGoogleMapController.m
stuartmorgan-g, line 1: Where are the tests for the changes in this file?

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
stuartmorgan-g: > If there is something needed on my part or some restructuring needed please let me know.

Given what I found and described in the comment above, this should become a `default` case (with the `ignore` updated accordingly, and the part of the comment about not using `default` removed) so that it works as intended.

packages/google_maps_flutter/google_maps_flutter_ios/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.18.4
```
stuartmorgan-g: Same versioning note for the iOS packages.

stuartmorgan-g (changes requested):

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
snaeji: Done — added the `default:` case with `// ignore: no_default_cases, unreachable_switch_default` and a comment matching `_platformMapTypeFromMapType` (~line 1214), minus the part about not using `default`.

packages/google_maps_flutter/google_maps_flutter_android/CHANGELOG.md
snaeji: Bumped to a minor version, 2.20.0.

packages/google_maps_flutter/google_maps_flutter_ios/CHANGELOG.md
snaeji: Bumped the iOS packages to minor — ios/sdk9/sdk10 are all 2.19.0.

packages/google_maps_flutter/google_maps_flutter_ios/ios/google_maps_flutter_ios/Sources/google_maps_flutter_ios/FGMGoogleMapController.m
snaeji: Added `testSetMapColorScheme` in RunnerTests (synced to all four iOS packages). It drives `updateWithMapConfiguration:` → `interpretMapConfiguration:` → `setColorScheme:` and asserts `overrideUserInterfaceStyle` for light/dark/followSystem plus the nil case. Passes on a simulator.

packages/google_maps_flutter/google_maps_flutter_ios/lib/src/google_maps_flutter_ios.dart
snaeji: Done — switched to a `default:` case with the same ignore, and dropped the "deliberately outside the switch" note. Synced across all four iOS packages.

snaeji:
Rebased onto main — single clean commit now, and the conflict is cleared. All the review points are addressed (replied inline on each thread):

- `default:` case on both the Android and iOS converters, with the `ignore` updated and the "not using default" note removed.
- Minor version bumps for the new functionality — google_maps_flutter_android 2.20.0, google_maps_flutter_ios / _sdk9 / _sdk10 2.19.0.
- Added a native `testSetMapColorScheme` covering the iOS color-scheme path (light/dark/followSystem + the nil-unchanged case).

One thing worth flagging since it wasn't in the threads: the platform-interface and app-facing changes are doc-comment only, so I left google_maps_flutter_platform_interface under the existing `## NEXT` (no version bump) and gave google_maps_flutter a 2.17.2 patch.

Verified locally before pushing: flutter analyze and dart format clean across all packages, Dart unit tests pass, and the new iOS XCTest passes on a simulator. Ready for another look — thanks!

stuartmorgan-g:
> the platform-interface and app-facing changes are doc-comment only, so I left google_maps_flutter_platform_interface under the existing `## NEXT` (no version bump)

Could you point me to what part of the docs linked from the PR checklist item gave the impression that doc comments were a documented exemption to our versioning policy?

snaeji:
Nothing — I relied too heavily on an LLM that pushed for this and didn't verify it myself. Note taken — bumped it to 2.15.1.

stuartmorgan-g:
> I relied too heavily on an LLM that pushed for this and didn't verify it myself

Please take a few minutes to re-read the AI contribution guidelines that you agreed to in the PR checklist.

packages/google_maps_flutter/google_maps_flutter_ios/example/ios/RunnerTests/GoogleMapsTests.m
```diff
@@ -222,6 +222,56 @@ - (void)testInspectorAPICameraPosition {
...
+  // Light. Start from a different style so the assertion confirms a real change.
+  mapView.overrideUserInterfaceStyle = UIUserInterfaceStyleDark;
+  FGMPlatformMapConfiguration *lightConfig = [self emptyCreationParameters].mapConfiguration;
+  lightConfig.colorScheme =
+      [[FGMPlatformMapColorSchemeBox alloc] initWithValue:FGMPlatformMapColorSchemeLight];
+  [controller.callHandler updateWithMapConfiguration:lightConfig error:&error];
+  XCTAssertNil(error);
+  XCTAssertEqual(mapView.overrideUserInterfaceStyle, UIUserInterfaceStyleLight);
+
+  // Dark.
```
stuartmorgan-g, line 248: Four distinct tests should be four distinct test functions, not a constant cycle of setup/assert/setup/assert in a single test.

stuartmorgan-g (changes requested):
The remaining comment is just about the iOS tests, so you can create the sub-PR to land the platform interface change now in parallel with updating the test.

OlaiyaO:
I validated the Android side locally: the Dart package tests and the changed native `ConvertTest` pass, and `MapColorScheme.dark` loads the expected cloud dark style in an Android app.

One lifecycle concern: the Maps SDK documentation for `GoogleMap.setMapColorScheme` says that it does not save state after the map is destroyed and must be reinitialized via `GoogleMapOptions.mapColorScheme(int)` or another `setMapColorScheme(int)` call: https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap#setMapColorScheme(int)

The current implementation clears `initialMapColorScheme` after the first `onMapReady`, so a later destroy/recreate of the underlying map on the same controller may fall back to light. Would it be safer for `GoogleMapBuilder.setMapColorScheme` to configure `options.mapColorScheme(colorScheme)` for the initial value, while retaining `GoogleMap.setMapColorScheme` in `GoogleMapController` for runtime updates? That also ensures the initial scheme is part of map construction rather than applied after readiness.

stuartmorgan-g:
> The remaining comment is just about the iOS tests, so you can create the sub-PR to land the platform interface change now in parallel with updating the test.

From triage: are you still planning on moving forward with breaking this into sub PRs to land?

PR #11748, opened 2026-05-20
[path_provider_linux] use soname for libgio

### LINKED ISSUE #186834: [path_provider_linux] libgio load should use soname instead of link name
### What package does this bug report belong to?

path_provider

### What target platforms are you seeing this bug on?

Linux

### Have you already upgraded your packages?

Yes

### Dependency versions

_No response_

### Steps to reproduce

1. ensure that the development package supplying the libgio link name is *not* installed
2. create flutter project on linux
3. add path_provider to dependencies
4. use path_provider to store a file in the data path
5. run the app, and check the newly created folder path

### Expected results

the data subdirectory name should be the application ID, but it will be the appname instead

### Actual results

the data subdirectory name is the appname instead

### Code sample

n/a: i already have a patch to fix. creating this issue because one is required for the pr

### Screenshots or Videos

_No response_

### Logs

_No response_

### Flutter Doctor output

n/a

gfongT5:
This was using the link name instead of the soname (see [1]) for dynamically loading libgio, resulting in a failed load if the relevant development package providing the link name symlink was not installed. The fallback behavior if the load fails is to use the executable name instead of the expected application ID, which means this would result in a different path being provided depending on whether the dev package is installed, even though the required library is actually present.

For example, on Ubuntu 22.04, with an executable named "testapp", with application ID set to "com.example.testapp":
- if `libglib2.0-dev` is installed, `getApplicationSupportPath()` returns `$XDG_DATA_HOME/com.example.testapp`
- if `libglib2.0-dev` is *not* installed, `getApplicationSupportPath()` returns `$XDG_DATA_HOME/testapp`

Change to use the soname instead, ensuring that the behavior is consistent whether or not the dev package that provides libgio is installed.

Also, replicate the fallback logic to use the appname for application cache path as well if an appname directory name exists, since it may have been unintentionally used prior to this change. Updated tests for the new behavior, and also changed to use a temporary test directory so that the test runner users' actual directories aren't used for the tests.

[1] https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html#AEN46

Fixes https://github.com/flutter/flutter/issues/186834

----

## Pre-Review Checklist

- [x] I signed the [CLA].
- [x] I added new tests to check the change I am making, or I have commented below to indicate which [test exemption] this PR falls under.

### DIFF
--- a/packages/path_provider/path_provider_linux/CHANGELOG.md
+++ b/packages/path_provider/path_provider_linux/CHANGELOG.md
@@ -1,5 +1,6 @@
-## NEXT
+## 2.2.2
 
+* Fixes getApplicationId() behavior on Linux when GLib development packages are not installed.
 * Updates minimum supported SDK version to Flutter 3.38/Dart 3.10.
 
 ## 2.2.1
--- a/packages/path_provider/path_provider_linux/lib/src/get_application_id_real.dart
+++ b/packages/path_provider/path_provider_linux/lib/src/get_application_id_real.dart
@@ -20,7 +20,7 @@ class GioUtils {
   /// Creates a default instance that uses the real libgio.
   GioUtils() {
     try {
-      _gio = DynamicLibrary.open('libgio-2.0.so');
+      _gio = DynamicLibrary.open('libgio-2.0.so.0');
     } on ArgumentError {
       _gio = null;
     }
--- a/packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart
+++ b/packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart
@@ -70,9 +70,21 @@ class PathProviderLinux extends PathProviderPlatform {
   @override
   Future<String?> getApplicationCachePath() async {
     final directory = Directory(path.join(xdg.cacheHome.path, await _getId()));
-    if (!directory.existsSync()) {
-      await directory.create(recursive: true);
+    if (directory.existsSync()) {
+      return directory.path;
     }
+
+    // Unlike for the application support path, the plugin didn't intentionally
+    // use the executable name as a directory for cache path in the past.
+    // However, it *unintentionally* did use it depending on which packages are
+    // installed, so fall back to it anyway.
+    // See: https://github.com/flutter/flutter/issues/186834
+    final legacyDirectory = Directory(path.join(xdg.cacheHome.path, await _getExecutableName()));
+    if (legacyDirectory.existsSync()) {
+      return legacyDirectory.path;
+    }
+
+    await directory.create(recursive: true);
     return directory.path;
   }
 
--- a/packages/path_provider/path_provider_linux/pubspec.yaml
+++ b/packages/path_provider/path_provider_linux/pubspec.yaml
@@ -2,7 +2,7 @@ name: path_provider_linux
 description: Linux implementation of the path_provider plugin
 repository: https://github.com/flutter/packages/tree/main/packages/path_provider/path_provider_linux
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+path_provider%22
-version: 2.2.1
+version: 2.2.2
 
 environment:
   sdk: ^3.10.0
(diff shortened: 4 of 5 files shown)


stuartmorgan-g:
Thanks for the contribution!

> [x] I followed [the version and CHANGELOG instructions](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version-and-changelog-updates), using [semantic versioning](https://dart.dev/tools/pub/versioning#semantic-versions) and the [repository CHANGELOG style](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style), or I have commented below to indicate which documented exception this PR falls under.

Where did you do this step?

stuartmorgan-g:
According to the description here and in the issue, this will change the storage location on a subset of machines. That's effectively a complete data-loss event for any existing application using `path_provider` on such a machine, which isn't something we can just ship near-silently.

Changing the behavior would need to include automatic detection and migration of data at the previous location.

gfongT5:
thanks @stuartmorgan-g for review! responses follow.

> > [x] I followed [the version and CHANGELOG instructions](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#version-and-changelog-updates), using [semantic versioning](https://dart.dev/tools/pub/versioning#semantic-versions) and the [repository CHANGELOG style](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style), or I have commented below to indicate which documented exception this PR falls under.
> 
> Where did you do this step?

I initially thought I did by updating the changelog NEXT entry as was done previously, but reading more carefully, I misunderstood how versioning works for this project. Updated using the tool.

> According to the description here and in the issue, this will change the storage location on a subset of machines. That's effectively a complete data-loss event for any existing application using `path_provider` on such a machine, which isn't something we can just ship near-silently.
> 
> Changing the behavior would need to include automatic detection and migration of data at the previous location.

It will not result in data loss because there is fallback behavior to prioritizing using the appname folder if that exists instead of creating a new folder using the app id, which was done for backward compatibility with previous behavior. Relevant excerpt from `packages/path_provider/path_provider_linux/lib/src/path_provider_linux.dart` function `getApplicationSupportPath()`:

```dart
@override
  Future<String?> getApplicationSupportPath() async {
    final directory = Directory(path.join(xdg.dataHome.path, await _getId()));
    if (directory.existsSync()) {
      return directory.path;
    }

    // This plugin originally used the executable name as a directory.
    // Use that if it exists for backwards compatibility.
    final legacyDirectory = Directory(
      path.join(xdg.dataHome.path, await _getExecutableName()),
    );
    if (legacyDirectory.existsSync()) {
      return legacyDirectory.path;
    }

    // Create the directory, because mobile implementations assume the directory exists.
    await directory.create(recursive: true);
    return directory.path;
  }
```

Whether that is good unconditional fallback behavior is open to question, since it makes this code into a name collision hazard (e.g., if two different programs with the appname `cool_app` using the ids `com.example.cool_app` and `org.example.cool_app` exist, and there's a `cool_app` subdirectory already, they'll *both* try to use that existing appname-based path as fallback instead of using their app ids to create new paths), but that tradeoff has already been made and shipped. At any rate, no change needed here for existing installs to continue working.

Automatically migrating data from the appname based path to an app ID path if possible might make sense to help avoid that name collision hazard, but I don't think it belongs within the scope of this pr or issue.

stuartmorgan-g:
Ah right, I forget we'd done that. That's only for `getApplicationSupportPath` though; `getApplicationCachePath` also uses the ID, and doesn't have that logic (since it's newer than the introduction of ID). So to avoid losing cache data, you'll need to add similar logic there.

gfongT5:
> Ah right, I forget we'd done that. That's only for `getApplicationSupportPath` though; `getApplicationCachePath` also uses the ID, and doesn't have that logic (since it's newer than the introduction of ID). So to avoid losing cache data, you'll need to add similar logic there.

Done, and added tests for both the updated behavior and the fallback behavior that already existed but had no test. I guess that means this doesn't need an test exception anymore (the main change still isn't easily testable because it involves adjusting the installed system packages, but this will make the checker happy :woman_shrugging:). Updating the description accordingly.

gfongT5:
Rebased and resolved a minor merge conflict. If you have a chance to review @robert-ancell or @stuartmorgan-g , it'd be much appreciated!

robert-ancell (approved):
Looks good, thanks!

stuartmorgan-g:
Overall the changes look good, but this will need to be reformatted with a recent version of Flutter `master`, and the analyzer issues will need to be fixed.

stuartmorgan-g (changes requested):
(That was meant to be a request-change review comment; fixing review status.)

stuartmorgan-g (approved):

PR #11632, opened 2026-05-01
 [camera] Add support for RGBA_8888 image streaming format

Mairramer:
This PR adds support for the rgba8888 image format group for both Android (`camera_android_camerax`) and iOS (`camera_avfoundation`).

Closes https://github.com/flutter/flutter/issues/151193

## Pre-Review Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [AI contribution guidelines] and understand my responsibilities, or I am not using AI tools.
- [x] I read the [Tree Hygiene] page, which explains my responsibilities.
- [x] I read and followed the [relevant style guides] and ran [the auto-formatter].
- [x] I signed the [CLA].
- [x] The title of the PR starts with the name of the package surrounded by square brackets, e.g. `[shared_preferences]`
- [x] I [linked to at least one issue that this PR fixes] in the description above.
- [x] I followed [the version and CHANGELOG instructions], using [semantic versioning] and the [repository CHANGELOG style], or I have commented below to indicate which documented exception this PR falls under[^1].
- [x] I updated/added any relevant documentation (doc comments with `///`).
- [x] I added new tests to check the change I am making, or I have commented below to indicate which [test exemption] this PR falls under[^1].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

**Note**: The Flutter team is currently trialing the use of [Gemini Code Assist for GitHub](https://developers.google.com/gemini-code-assist/docs/review-github-code). Comments from the `gemini-code-assist` bot should not be taken as authoritative feedback from the Flutter team. If you find its comments useful you can update your code accordingly, but if you are unsure or disagree with the feedback, please feel free to wait for a Flutter team member's review for guidance on which automated comments should be addressed.

[^1]: Regular contributors who have demonstrated familiarity with the repository guidelines only need to comment if the PR is not auto-exempted by repo tooling.

[Contributor Guide]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md
[AI contribution guidelines]: https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#ai-contribution-guidelines
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[relevant style guides]: https://github.com/flutter/packages/blob/main/CONTRIBUTING.md#style
[...]

PR #11765, opened 2026-05-22
[camera_platform_interface] Add rgba8888 format to ImageFormatGroup

Mairramer:
Part of https://github.com/flutter/packages/pull/11632

## Pre-Review Checklist

- [x] I signed the [CLA].


bparrishMines (approved):
LGTM

@stuartmorgan-g or @tarrinneal  for secondary review

stuartmorgan-g:
The addition looks good, but the two switch analyzer errors will need temporary `// ignore`s, which the follow-up PR should clean up.

Mairramer:
> The addition looks good, but the two switch analyzer errors will need temporary `// ignore`s, which the follow-up PR should clean up.

Done.

stuartmorgan-g:
You added case handling, not ignores. That will fail our safety check against modifying code across dependency layers in a federated plugin.

You need to ignore the warning in this PR, and add the cases in the next PR.

Mairramer:
> You added case handling, not ignores. That will fail our safety check against modifying code across dependency layers in a federated plugin.
> 
> You need to ignore the warning in this PR, and add the cases in the next PR.

Yes, I messed up. I'll fix it right away.

Mairramer:
@stuartmorgan-g I believe it should pass now, hopefully.

stuartmorgan-g:
I'm very confused by the compile error here; I thought this was only an analyzer warning as long as it wasn't a switch expression (e.g., `return switch ...`). Did the behavior change?

stuartmorgan-g:
Wow, apparently [it changed several years ago for enums](https://github.com/dart-lang/language/commit/ccfb7cfba9b8b7cd80ebe8afbdb85c8de0d83f68) and we hadn't hit it until now.

So there will need to be a prequel PR to change all of those constructions to use `default` instead of the current return-outside-switch approach. Which unfortunately means that we won't get warnings about places we need to update any more, and it'll just have to be feature-test driven.

Mairramer:
> Wow, apparently [it changed several years ago for enums](https://github.com/dart-lang/language/commit/ccfb7cfba9b8b7cd80ebe8afbdb85c8de0d83f68) and we hadn't hit it until now.
> 
> So there will need to be a prequel PR to change all of those constructions to use `default` instead of the current return-outside-switch approach. Which unfortunately means that we won't get warnings about places we need to update any more, and it'll just have to be feature-test driven.

How to proceed in this case?

stuartmorgan-g:
Specifically, all the code in those util files that does:
```dart
switch (foo) {
  ...
}
// The enum comes from a different package, which could get a new value at
// any time, so provide a fallback that ensures this won't break when used
// with a version that contains new values. This is deliberately outside
// the switch rather than a `default` so that the linter will flag the
// switch as needing an update.
// ignore: dead_code
return bar;
```

Will need to become:
```dart
return switch(foo) {
 ...
  // The enum comes from a different package, which could get a new value at
  // any time, so provide a fallback that ensures this won't break when used
  // with a version that contains new values. 
  // ignore: no_default_cases, unreachable_switch_default
  default:
    bar;
}
```

And that will need to land in a separate PR before this one can land.

stuartmorgan-g:
Oh, but that's not enough, because it would still break the existing versions. So despite our intentions, adding an enum value *is* breaking for our existing implementation code.

I think we'll need to do a one-time breaking change to the platform interface, as much as I hate to do that. We can add a private enum value to each of the enums that we want to be able to extend in that breaking change, to force clients not to rely on exhaustive matching. @bparrishMines Can you think of any better options?

(Alternately we could ship the implementation change, wait a while, assume that it's very unlikely that anyone would be on an old implementation package version but get a new platform interface version, and then change it without a major version bump, but I don't think we should do that since it violates the intent of using semver.)

bparrishMines:
> We can add a private enum value to each of the enums that we want to be able to extend in that breaking change, to force clients not to rely on exhaustive matching.

@stuartmorgan-g I'm not sure what you mean by this? As in adding a private field or an enum value that is private.

I think we originally added `unknown` for this exact reason before exhaustive enums became an error and not a warning. If we want enums to be extendable we may be better off switching them to constants or extension types. Assuming we take the path for creating a breaking change in the interface.

stuartmorgan-g:
> @stuartmorgan-g I'm not sure what you mean by this? As in adding a private field or an enum value that is private.

There's a pattern I saw mentioned in a Dart issue or discussion (I forget where exactly) where you do something like:

```dart
enum Foo { a, b, c, _requireDefaultHandler }
```

Since clients can't reference `_requireDefaultHandler`, they have to have a `default` case. IIRC the error message isn't great though; we should check that.

Mairramer:
> > @stuartmorgan-g I'm not sure what you mean by this? As in adding a private field or an enum value that is private.
> 
> There's a pattern I saw mentioned in a Dart issue or discussion (I forget where exactly) where you do something like:
> 
> ```dart
> enum Foo { a, b, c, _requireDefaultHandler }
> ```
> 
> Since clients can't reference `_requireDefaultHandler`, they have to have a `default` case. IIRC the error message isn't great though; we should check that.

I've taken some time to locally test both approaches discussed here: migrating to an extension type and adding a private _requireDefaultHandler to the existing enum.

While the private enum value (_requireDefaultHandler) is functional and successfully forces a fallback, it feels a bit unidiomatic. A major downside I noticed during testing is that enforcing the default: clauses across the plugin ecosystem forces us to litter the code with // ignore: no_default_cases, unreachable_switch_default, or alternatively, remove these strict rules from the analysis_options.yaml. 

On the other hand, the extension type approach worked flawlessly out of the box. Since it's not strictly exhaustive by nature, it behaves exactly how we want an open-ended set of formats to behave without requiring any linter suppression or hacky sentinel values.

stuartmorgan-g:
> forces us to litter the code with // ignore: no_default_cases, unreachable_switch_default

Are you sure `unreachable_switch_default` is triggered for that? I would not expect that, since handlers won't be able to include all the cases, so the default should be reachable.

Mairramer:
> > forces us to litter the code with // ignore: no_default_cases, unreachable_switch_default
> 
> Are you sure `unreachable_switch_default` is triggered for that? I would not expect that, since handlers won't be able to include all the cases, so the default should be reachable.

You're right, it shouldn't trigger. That was actually left over from an earlier experiment I did without `_requireDefaultHandler`.

bparrishMines:
@stuartmorgan-g Since the private enum causes the `ino default cases` lint message. Should we use class types instead. They don't allow exhaustive switch statements and would allow platform implementations to create platform specific values? e.g.

```dart
interface class ImageFormatGroup {
  const ImageFormatGroup._();

  static const ImageFormatGroup jpg = ImageFormatGroup._();
  static const ImageFormatGroup rgba8888 = ImageFormatGroup._();
}

// In a future Android implementation:
interface class AndroidImageFormat implements ImageFormatGroup {
  const AndroidImageFormat._();

  static const AndroidImageFormat nv12 = AndroidImageFormat._();
}
```

Mairramer:
@stuartmorgan-g @bparrishMines

How would you like to proceed in this case?

Depending on the direction you'd like to take, I can put together an implementation that aligns with the requirements and constraints we expect to have.

stuartmorgan-g:
@bparrishMines Could you write up a short design doc that lays out the options (enum with private value, extension types, class types) with pros and cons for our general use case of enums in federated plugins? Since this will be a breaking change to platform interfaces we want to get it right, so we should have a clear record of our reasoning, and also run it by Dart folks for feedback.

Mairramer:
@stuartmorgan-g I took the liberty of creating a design doc for this, although I'm not sure it was necessary. I included some of the proposals discussed here and added two more.

I also created an [issue](https://github.com/flutter/flutter/issues/189696) for it  and added a direct link to the document while we validate the proposal. Later, if it makes sense to move forward, I'll create a PR for the website.

stuartmorgan-g:
From triage: this is blocked on the design doc, which is under review.

PR #11767, opened 2026-05-23
Migrate shared_preferences_android to built-in Kotlin

faisalansari0367:
Removes explicit Kotlin Gradle Plugin dependency and migrates to Flutter's built-in Kotlin support per the breaking-changes migration guide.


stuartmorgan-g:
Thanks for the contribution, but we aren't doing ad-hoc migrations of individual plugins. We will be applying a consistent migration process, including repository tooling.

PR #11774, opened 2026-05-25
[camera] add custom path ouput to recording

### LINKED ISSUE #91680: [Camera]:  Proposal to add set path file on startVideoRecording
## Use case
Need set path file on start videorecording

Mairramer:
This PR introduces the ability to save videos to custom locations, taking into account the limitations and behavior of each platform.

Due to platform-specific security restrictions and limitations:

* On Android, I was able to save videos directly to `/storage/emulated/0/Download/video.mp4` without issues.
* On iOS, I was able to save videos to `On My iPhone -> Camera Example -> Movies`.
* On Windows, the behavior is similar to Android.

Closes https://github.com/flutter/flutter/issues/91680

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/camera/camera_web/lib/src/camera_web.dart
```diff
@@ -461,7 +461,11 @@ class CameraPlugin extends CameraPlatform {
...
 
   @override
-  Future<void> startVideoRecording(int cameraId, {Duration? maxVideoDuration}) {
+  Future<void> startVideoRecording(
+    int cameraId, {
+    Duration? maxVideoDuration,
+    String? videoOutputPath,
+  }) {
     // Ignore maxVideoDuration, as it is deprecated.
     return startVideoCapturing(VideoCaptureOptions(cameraId));
```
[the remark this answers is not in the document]
Mairramer: In web implementation, this doesn't matter.

bparrishMines (commented):
@Mairramer What's the use case you want to solve by adding this feature? From my understanding, it should currently be possible for the app to rename the file to change it's location without copying.

The issue linked mentions being able to save to an external SD card and finding the file if the app crashes during recording, but I dont think this PR would fix these.

Mairramer:
> @Mairramer What's the use case you want to solve by adding this feature? From my understanding, it should currently be possible for the app to rename the file to change it's location without copying.
> 
> The issue linked mentions being able to save to an external SD card and finding the file if the app crashes during recording, but I dont think this PR would fix these.

Thanks for the review! 
I get your point about `rename`, but the main issue here is cross-filesystem moves.

If we record a video to the internal cache and then try to `rename()` it to an external SD card, the OS can't just move the file pointer. It's forced to do a full byte-by-byte copy and delete. For large videos, this causes a huge I/O bottleneck and temporarily doubles the storage space needed (which can fail if internal storage is low). Writing directly to the target path avoids this completely.

Regarding the SD card: apps still have direct write access to their app-specific folders on external SD cards (for example, the second directory returned by `getExternalStorageDirectories()` in `path_provider`). By using `videoOutputPath`, we can record straight to the SD card from the start and bypass internal storage limits altogether.

Regarding the crash recovery scenario mentioned in https://github.com/flutter/flutter/issues/91680, you're absolutely right that the MP4 may end up corrupted after a crash, especially if the final `moov` atom was never written.

The main issue today is that the native layer generates a random temporary filename, and Dart only receives that path after `stopVideoRecording` completes successfully. If the app crashes before that happens, the temporary file becomes orphaned and the application has no reliable way to discover or identify it afterward.

By providing the output path upfront, the application always knows where the recording is being written. After a restart, developers can at least locate the interrupted file, attempt MP4 recovery if possible, or simply clean up the orphaned file to avoid unnecessary storage leaks.

Let me know if this makes sense or if I should adjust anything in the implementation.

Mairramer:
Regarding crash recovery:

On iOS, we could use [movieFragmentInterval](https://developer.apple.com/documentation/avfoundation/avassetwriter/moviefragmentinterval), which allows the movie file to be fragmented periodically and can help reduce data loss in crash scenarios.

On Android, as far as I know, there is no equivalent native solution available. We would probably need a custom implementation using [MediaMuxer](https://developer.android.com/reference/android/media/MediaMuxer) or something similar to periodically fragment or remux the recording stream, but that’s not something I’d really like to introduce at the moment.

packages/camera/camera/example/lib/video_recording_example.dart
bparrishMines, line 1: I don't think we need a full example just for video recording. The `main.dart` already has recording in it. I think you could change this to be a small snippet showing how to set an output path when recording a video. It doesn't need to be a fully runnable app.

packages/camera/camera/README.md
```diff
@@ -92,6 +92,58 @@ Here is a list of all permission error codes that can be thrown:
...
+```dart
+// Always ensure the path ends with the .mp4 extension
+await controller.startVideoRecording(
+  videoOutputPath: '/path/to/your/custom_video.mp4',
+);
+```
+
+#### Platform-Specific Considerations
+
+**Android**
```
bparrishMines: This information should be moved to the `README` of their respective platforms. And this should just link to their pub.dev pages.

bparrishMines (changes requested):
@Mairramer Thanks for the explanation! I think it makes sense to add this feature. The platform interface and app-facing interface look fine to me just a couple comments about the README and example.

I will add reviewers for platform implementations.

stuartmorgan-g:
From triage: This is still waiting for sign-off from iOS and Android reviewers.

okorohelijah:
from triage cc @hellohuanlin

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -1317,6 +1324,23 @@ void prepareRecording() {
     setFpsCameraFeatureForRecording(cameraProperties);
   }
 
+  private void validateOutputPath(String path) {
+    File file = new File(path);
+    if (file.isDirectory()) {
+      throw new Messages.FlutterError("IOError", "The output path is a directory: " + path, null);
```
camsim99, line 1331: Can this error message mention that a file path should be provided instead?

packages/camera/camera_android/android/src/main/java/io/flutter/plugins/camera/Camera.java
```diff
@@ -1317,6 +1324,23 @@ void prepareRecording() {
...
+    }
+    File parent = file.getParentFile();
+    if (parent != null && !parent.exists()) {
+      throw new Messages.FlutterError(
+          "IOError", "The parent directory does not exist: " + parent.getAbsolutePath(), null);
+    }
+
+    String lowerPath = path.toLowerCase(Locale.ROOT);
+    if (!lowerPath.endsWith(".mp4")) {
+      throw new Messages.FlutterError("IOError", "Invalid video extension. Supported: .mp4", null);
```
camsim99: Nit for clarity:

```suggestion
      throw new Messages.FlutterError("IOError", "Invalid video extension. Supported extensions: .mp4", null);
```

packages/camera/camera_android/android/src/test/java/io/flutter/plugins/camera/CameraTest.java
```diff
@@ -711,6 +720,113 @@ public void resumeVideoRecording_shouldCallPauseWhenRecording() {
...
+          assertThrows(
+              Messages.FlutterError.class,
+              () -> {
+                camera.startVideoRecording(null, invalidPath);
+              });
+      // The old test asserted "Invalid path" message. Wait, validateOutputPath will throw first!
+      // Actually, if validateOutputPath throws, it will throw IOError before building.
+      // So this test as originally written was flawed because validateOutputPath catches the
+      // invalidPath before build().
+      // Let's just assert the IOError from validateOutputPath for this test.
```
camsim99: Nit: This comment can just clarify the assertion below vs what the old test did.

packages/camera/camera_android/example/lib/camera_controller.dart
```diff
@@ -315,9 +315,23 @@ class CameraController extends ValueNotifier<CameraValue> {
...
   /// Throws a [CameraException] if the capture fails.
-  Future<void> startVideoRecording({void Function(CameraImageData image)? streamCallback}) async {
+  Future<void> startVideoRecording({
+    void Function(CameraImageData image)? streamCallback,
+    String? videoOutputPath,
+  }) async {
+    if (videoOutputPath != null) {
+      final String lowerPath = videoOutputPath.toLowerCase();
+      if (!lowerPath.endsWith('.mp4')) {
+        throw CameraException('InvalidFilePath', 'Invalid video extension. Supported: .mp4');
```
camsim99: Nit
```suggestion
        throw CameraException('InvalidFilePath', 'Invalid video extension. Supported extensions: .mp4');
```

packages/camera/camera_android/lib/src/android_camera.dart
```diff
@@ -186,14 +186,21 @@ class AndroidCamera extends CameraPlatform {
...
+    return startVideoCapturing(VideoCaptureOptions(cameraId, videoOutputPath: videoOutputPath));
   }
 
   @override
   Future<void> startVideoCapturing(VideoCaptureOptions options) async {
-    await _hostApi.startVideoRecording(options.streamCallback != null);
+    await _hostApi.startVideoRecording(
+      options.streamCallback != null,
+      videoOutputPath: options.videoOutputPath,
+    );
```
camsim99, line 203: Can you add a test for this feature in `packages/camera/camera_android/test/android_camera_test.dart`?

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/RecorderProxyApi.java
```diff
@@ -70,12 +71,31 @@ public PendingRecording prepareRecording(Recorder pigeonInstance, @NonNull Strin
     return pendingRecording;
   }
 
+  private void validateOutputPath(@NonNull String path) {
+    File file = new File(path);
+    if (file.isDirectory()) {
+      throw new GeneratedCameraXLibrary.FlutterError(
```
camsim99: `GeneratedCameraXLibrary` is actually deprecated and should be removed. You can instead directly throw `IOException`s here!

packages/camera/camera_android_camerax/lib/src/android_camera_camerax.dart
```diff
@@ -1085,9 +1086,13 @@ class AndroidCameraCameraX extends CameraPlatform {
   ///
   /// This method is deprecated in favour of [startVideoCapturing].
   @override
-  Future<void> startVideoRecording(int cameraId, {Duration? maxVideoDuration}) async {
+  Future<void> startVideoRecording(
```
camsim99, line 1134: Can you update the dart doc to clarify what happens when `videoOutputPath` is/is not specified?

packages/camera/camera_android_camerax/README.md
```diff
@@ -55,10 +55,28 @@ in the merged Android manifest of your app, then take the following steps to rem
 
 ### Notes on video capture
 
+#### Custom Video Recording Path
+
+Although it is possible to use an absolute path like `/storage/emulated/0/Download/video.mp4`, this is a fragile practice and may fail on many devices or Android versions due to **Scoped Storage** restrictions.
```
camsim99: Nit: For clarity, can you start this note off with the context that this is an optional thing you can do? Something like "If you wish to record a video to a custom file path, you may use blah blah method."

camsim99 (commented):
Overall LGTM :) just have some thoughts on testing and docs

packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/RecorderProxyApi.java
camsim99: @bparrishMines Actually to confirm, now that the package is migrated to proxy APIs, can we delete `GeneratedCameraXLibrary`? I see [one usage](https://github.com/flutter/packages/blob/97438f1c083368f44d0201ad52ebe7f7ca924602/packages/camera/camera_android_camerax/android/src/test/java/io/flutter/plugins/camerax/SystemServicesTest.java#L96) of it, but I can easily migrate that too delete it if so.

LongCatIsLooong:
from: @Mairramer could you resolve the merge conflicts?

camsim99 (approved):
Android LGTM!

LouiseHsu:
@hellohuanlin from triage :)

packages/camera/camera_avfoundation/darwin/camera_avfoundation/Sources/camera_avfoundation/CameraPlugin.swift
```diff
@@ -379,11 +379,13 @@ extension CameraPlugin: CameraApi {
   }
 
   func startVideoRecording(
-    enableStream: Bool, completion: @escaping (Result<Void, any Error>) -> Void
+    enableStream: Bool, videoOutputPath: String?,
```
hellohuanlin, line 382: nit: separate lines for these 2

packages/camera/camera_avfoundation/darwin/camera_avfoundation/Sources/camera_avfoundation/DefaultCamera.swift
```diff
@@ -520,27 +521,38 @@ final class DefaultCamera: NSObject, Camera {
...
+        videoRecordingPath = try getTemporaryFilePath(
+          withExtension: "mp4",
+          subfolder: "videos",
+          prefix: "REC_")
+      } catch let error as NSError {
+        completion(.failure(DefaultCamera.pigeonErrorFromNSError(error)))
+        return
+      }
     }
+    self.videoRecordingPath = videoRecordingPath
```
hellohuanlin: should we remove this line? 

> self.videoRecordingPath = videoRecordingPath

hellohuanlin (approved):

hellohuanlin:
@Mairramer can you resolve the merge conflict? thanks

vashworth:
@bparrishMines Can you re-review since you requested changes?

packages/camera/camera_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.14.0
+
+* Adds `videoOutputPath` support to `MethodChannelCamera.startVideoRecording`.
```
bparrishMines: nit: This adds the feature to the VideoCaptionOptions, not just the the `MethodChannelCamera` implementation.

This should say something like: "Adds support for setting the video output path. See \`VideoCaptureOptions.videoOutputPath\`".

packages/camera/camera_platform_interface/lib/src/platform_interface/camera_platform.dart
```diff
@@ -143,6 +143,7 @@ abstract class CameraPlatform extends PlatformInterface {
     int cameraId, {
     @Deprecated('This parameter is unused, and will be ignored on all platforms')
     Duration? maxVideoDuration,
+    String? videoOutputPath,
```
bparrishMines: Can this be added without breaking the platform implementations? They extend this class, so wouldn't the missing parameter make this a breaking change?

I think this can only be supported in `startVideoCapturing(VideoCaptureOptions options)`. Making the change below uneccessary.

bparrishMines (changes requested):

packages/camera/camera/lib/src/camera_controller.dart
```diff
@@ -577,6 +578,13 @@ class CameraController extends ValueNotifier<CameraValue> {
       );
     }
 
+    if (videoOutputPath != null) {
+      final String lowerPath = videoOutputPath.toLowerCase();
+      if (!lowerPath.endsWith('.mp4')) {
+        throw CameraException('InvalidFilePath', 'Invalid video extension. Supported: .mp4');
+      }
+    }
```
bparrishMines: This is something that should be handled by the platform implementations. If one platform started supporting another format, this would block them from supporting it.

packages/camera/camera/example/ios/Runner/Info.plist
bparrishMines, line 1: Were the changes to this file intended to be added?

bparrishMines (changes requested):

packages/camera/camera/lib/src/camera_controller.dart
Mairramer: Ah, yeah, I forgot to remove that.

packages/camera/camera/example/ios/Runner/Info.plist
Mairramer: We don't need it anymore. There used to be an example here, but I documented it in [packages/camera/camera_avfoundation/README.md](https://github.com/flutter/packages/pull/11774/changes/BASE..bc3dccfb145435652e0b9807ebdf5076f45bcc5a#diff-eec6f14195592918d65e451eee950ce2b14c30b2e7ed3fd3a78e10f689bb5254), so I think we can remove it now.

okorohelijah:
from triage: @Mairramer can you resolve the test failure and merge conflicts?

bparrishMines (approved):
LGTM

Platform interface and app facing look good to me. It looks like there isn't a `camera_web` implementation, but I'm assuming this is because you can't access the file system with web.

However, to keep platform parity and prevent web from ignoring the value silently, can you update the file creation name at:

https://github.com/flutter/packages/blob/main/packages/camera/camera_web/lib/src/camera.dart#L470

And then also do a check that it is only a filename and doesn't include directories (e.g. `myVideo.mp4` and not `some/folder/myVideo.mp4`.

The windows implementation looks fine to me as well, but I'm not that familiar with the platform. @stuartmorgan-g Is there someone that usually handles that review or is it up to me?

Mairramer:
> LGTM
> 
> Platform interface and app facing look good to me. It looks like there isn't a `camera_web` implementation, but I'm assuming this is because you can't access the file system with web.
> 
> However, to keep platform parity and prevent web from ignoring the value silently, can you update the file creation name at:
> 
> https://github.com/flutter/packages/blob/main/packages/camera/camera_web/lib/src/camera.dart#L470
> 
> And then also do a check that it is only a filename and doesn't include directories (e.g. `myVideo.mp4` and not `some/folder/myVideo.mp4`.
> 
> The windows implementation looks fine to me as well, but I'm not that familiar with the platform. @stuartmorgan-g Is there someone that usually handles that review or is it up to me?

For now, I’ve avoided changing the web implementation. But I’ll take a closer look at it separately.
Edit:
If needed, I can create a separate issue for the web implementation and close the current one, since this will be supported on three platforms and only the web, with its specific quirks, would be left out for now.

stuartmorgan-g:
> @stuartmorgan-g Is there someone that usually handles that review or is it up to me?

Windows reviews have mostly been me recently, but @Piinks is following up on ownership for Windows plugin implementations.

Piinks:
Touched base with @loic-sharma, added @flutter/windows-reviewers and tagged `triage-windows` for review.

PR #11784, opened 2026-05-26
[tool] Add config options needed by core-packages

stuartmorgan-g:
Adds some optional behaviors needed to use the tool in flutter/core-packages:
- Allow setting a min Dart version rather than a min Flutter version at the repo level
- Don't require .ci.yaml to be present when batch config isn't used (since that repo currently doesn't use LUCI)


script/tool/lib/src/common/tool_config.dart
```diff
@@ -63,6 +63,19 @@ String? getMinFlutterVersion(Directory repoRoot) {
   return yaml;
 }
 
+/// Returns the minimum Dart version allowed.
+String? getMinDartVersion(Directory repoRoot) {
+  final Object? yaml = _getToolConfig(repoRoot)['min_dart'];
+  if (yaml == null) {
+    return null;
```
chunhtai, line 70: why not throw?

script/tool/lib/src/validate_command.dart
```diff
@@ -149,7 +156,22 @@ class ValidateCommand extends PackageLoopingCommand {
     }
     if (_shouldRun(Validator.pubspec)) {
       await _loadAllowedDependencies();
-      _minMinFlutterVersion = _loadMinMinFlutterVersion();
+      final ({Version? flutter, Version? dart}) minSdkVersions =
```
chunhtai: Can just use destruction
```
final ({flutter: Version? minFlutter, dart: Version? minDart}) = _loadMinMinSdkVersions();
```

script/tool/lib/src/validators/repo_info_validator.dart
```diff
@@ -428,6 +428,17 @@ class RepoInfoValidator {
   }) {
     final errors = <String>[];
     final File ciYamlFile = repoRoot.childFile('.ci.yaml');
+    if (!ciYamlFile.existsSync()) {
```
chunhtai, line 431: why do we bar batch release based on .ci.yaml? IIRC, batch release infra are mostly based on GA and didn't rely on ci.yaml. Or is this a indirect way to disable the batch release for core-package?

script/tool/lib/src/common/tool_config.dart
stuartmorgan-g: Because only one of `min_flutter` and `min_dart` is required. The minimum Dart version is inferred from the Flutter version in this repo still.

script/tool/lib/src/validators/repo_info_validator.dart
stuartmorgan-g: Because the next non-new line in this file unconditionally reads `ciYamlFile`, and that will explode if, as is currently the case in core-packages, there is no such file.

Rather than speculatively implement an alternate codepath that does the same validation for a non-LUCI-based configuration that we've never built and may never need, I'm just describing the actual state of the tooling.

script/tool/lib/src/common/tool_config.dart
chunhtai: oh ok, makes sense

script/tool/lib/src/validators/repo_info_validator.dart
chunhtai: oh ok, I think we can just return without the print error then. 

The reason behind this check for batch release is the we have to add a enable_branch in ci.yaml to make sure the luci ci runs.

If it doesn't exist, then we don't need to worry about having the enable_branch check.

stuartmorgan-g: Why wouldn't the same basic requirement apply to a repo using GA to run CI, given that GA also has branch restrictions in configurations?

chunhtai: yeah the same will apply if we use GA to run ci, but this helper method is specific for ci.yaml.

chunhtai: oh I see the reasoning, so until we implement the GA restriction, batch release is not supported. yeah I think that makes sense too.

script/tool/lib/src/validate_command.dart
stuartmorgan-g: Done. I tried that before but got tripped on having forgotten to remove the `{}` (the same mistake you've made there) and thought I was misremembering what was possible with destructuring.

chunhtai (approved):
LGTM

PR #11792, opened 2026-05-26
[google_maps_flutter] Disable marker move animation on iOS

Gibbo97:
Disable the marker move animation google_maps_flutter on iOS. Currently when you update the position it takes some time to move. However, on other platforms; android, web in my testing, the marker move is not animated. Additionally other features like lines being moved is not animated.

Marker move has forced animation on google_maps_flutter iOS.
Issue: https://github.com/flutter/flutter/issues/129826

Behaviour previously, showing lines and markers behaving differently:
https://github.com/user-attachments/assets/d129bbb6-af93-4825-9b05-10befed4fd87

Behaviour now showing markers and lines behaving the same:
https://github.com/user-attachments/assets/3437caf9-628c-489e-ac5d-f0fb55897a6e

## Pre-Review Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
@@ -9,7 +9,8 @@ import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/services.dart';
-import 'package:google_maps_flutter_platform_interface/google_maps_flutter_platform_interface.dart';
+import 'package:google_maps_flutter_platform_interface/google_maps_flutter_platform_interface.dart'
+    hide PlatformMapConfiguration;
 import 'package:stream_transform/stream_transform.dart';
 
 import 'google_map_inspector_android.dart';
--- a/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/platform_map_configuration.dart
+++ b/packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/platform_map_configuration.dart
@@ -1,0 +1,15 @@
+// Copyright 2013 The Flutter Authors
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+
+/// Base class for platform-specific map configuration.
+///
+/// Platform implementations can add platform-specific configuration options
+/// by extending this class.
+@immutable
+class PlatformMapConfiguration {
+  /// Creates a platform-specific map configuration.
+  const PlatformMapConfiguration();
+}
--- a/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
+++ b/packages/google_maps_flutter/google_maps_flutter/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.19.0
+
+* Adds `GoogleMapController.setPlatformConfiguration` for configuring
+  platform-specific map options.
+
 ## 2.18.0
 
 * Adds support for mapTypeControlEnabled, fullscreenControlEnabled, and streetViewControlEnabled on web.
--- a/packages/google_maps_flutter/google_maps_flutter/lib/google_maps_flutter.dart
+++ b/packages/google_maps_flutter/google_maps_flutter/lib/google_maps_flutter.dart
@@ -54,6 +54,7 @@ export 'package:google_maps_flutter_platform_interface/google_maps_flutter_platf
         MinMaxZoomPreference,
         PatternItem,
         PinConfig,
+        PlatformMapConfiguration,
         Polygon,
         PolygonId,
         Polyline,
--- a/packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
+++ b/packages/google_maps_flutter/google_maps_flutter/lib/src/controller.dart
@@ -128,6 +128,15 @@ class GoogleMapController {
     return GoogleMapsFlutterPlatform.instance.updateMapConfiguration(update, mapId: mapId);
   }
 
+  /// Sets platform-specific configuration for this map.
+  ///
+  /// Platform-specific configuration classes are provided by the platform
+  /// implementation packages.
+  Future<void> setPlatformConfiguration(PlatformMapConfiguration configuration) {
+    _checkWidgetMountedOrThrow();
+    return GoogleMapsFlutterPlatform.instance.setPlatformConfiguration(configuration, mapId: mapId);
+  }
+
   /// Updates marker configuration.
   ///
   /// Change listeners are notified once the update has been made on the
--- a/packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
+++ b/packages/google_maps_flutter/google_maps_flutter/pubspec.yaml
@@ -2,7 +2,7 @@ name: google_maps_flutter
 description: A Flutter plugin for integrating Google Maps in iOS and Android applications.
 repository: https://github.com/flutter/packages/tree/main/packages/google_maps_flutter/google_maps_flutter
 issue_tracker: https://github.com/flutter/flutter/issues?q=is%3Aissue+is%3Aopen+label%3A%22p%3A+maps%22
-version: 2.18.0
+version: 2.19.0
 
 environment:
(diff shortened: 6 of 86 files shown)


stuartmorgan-g:
This is unconditionally making `google_maps_flutter` on iOS behave differently from the default behavior of Google Maps SDK on iOS, which is generally not the approach we take. Adding a map-level option to disable implicit animation for markers is fine, but unconditionally changing it may regress other clients.

Gibbo97:
@stuartmorgan-g Thanks for the prompt reply. 

Just so I can get something in which you would be more keen on: are you picturing a new field on `PlatformMapConfiguration` (defaulting to animated, matching the SDK's native behavior) that, when set to disabled, wraps the `marker.position` assignment in a `CATransaction` with actions disabled? So default to the current behaviour, opt in to the standardised behaviour.

Or are you thinking of a bigger change?

A couple other things I'd like to confirm before I sink an afternoon into this:
1. Should the flag be exposed on the cross-platform `GoogleMap` widget (iOS-only effect, no-op elsewhere) or kept as an iOS-specific option?
2. Any preference on naming/polarity (e.g. `markerPositionAnimationsEnabled` default-on)?
3. Scope it to `position` only, or also cover other implicitly animated properties like `rotation`?

stuartmorgan-g:
> 1. Should the flag be exposed on the cross-platform `GoogleMap` widget (iOS-only effect, no-op elsewhere) or kept as an iOS-specific option?

Do other platforms have APIs that allow toggling animation for these properties? If so we would want to abstract them, if not it should be an iOS-specific option. (I'm not familiar with all of the differences of the Google Maps API surface and behaviors on each platform offhand.)

> 2. Any preference on naming/polarity (e.g. `markerPositionAnimationsEnabled` default-on)?

APIs in this package should follow [Flutter style guidance](https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-double-negatives-in-apis).

> 3. Scope it to `position` only, or also cover other implicitly animated properties like `rotation`?

If there are other APIs that this could obviously apply to, it should be designed in such a way that it would be easily extensible to other properties. E.g., a simple data class of animation toggles for various objects.

stuartmorgan-g:
From triage: marking as draft pending the update to the approach described above.

Gibbo97:
Hi @stuartmorgan-g I have had a chance to come back to this, hope this is what you were thinking.

Some general implementation notes:
This is now opt-in and iOS-specific, rather than changing marker updates globally. You weren't keen on the earlier unconditional change could regress clients relying on the Google Maps iOS SDK’s implicit marker animation. The new configuration defaults to enabled animations, so existing behavior is preserved unless you explicitly opt out.

The API is a configuration object rather than a single `disableMarkerPositionAnimation` flag which leaves room for other animated marker properties. Rotation is included because iOS implicitly animates marker rotation too, and position/rotation can be toggled independently.

stuartmorgan-g:
We don't want clients bypassing the map controller and calling platform interface methods directly; that's a dangerous pattern that we don't want to encourage.

Because this package unfortunately doesn't have map instance objects at the platform interface level, we can't do this using the `in_app_purchase` addition system approach, so I think the best option under the current design (and without reworking the creation flow to make this a widget-level option) is to do something more like we do with [webview configuration objects](https://github.com/flutter/packages/blob/main/packages/webview_flutter/webview_flutter_platform_interface/lib/src/types/platform_webview_controller_creation_params.dart):
- Add an empty `PlatformMapConfiguration` object in the platform interface, and also a (no-op) platform interface method that takes that and a map ID to update the platform-specific configuration.
- Add a `setPlatformConfiguration(PlatformMapConfiguration configuration)` method to the app-facing package
- Add an iOS-package-specific subclass of `PlatformMapConfiguration` that includes the new options, and override the platform interface method to handle its option subclass.

It's not ideal in that it's inconsistent with the existing widget-config-based pattern, but I think plumbing platform options into that would require a non-trivial redesign of the plugin, and I don't want to block this on existing technical debt.

stuartmorgan-g (changes requested):
(Fixing review state; that was meant to be a review comment.)

stuartmorgan-g:
From triage: Are you still planning on updating this PR based on the feedback above?

Gibbo97:
Hey @stuartmorgan-g 
Yes. I have done half the rework last week.

Sorry its taking so long!

Gibbo97:
Hi @stuartmorgan-g, I’ve made the requested changes, and this is ready for review.

The PR now introduces a platform-specific map configuration API rather than adding iOS-only options to the cross-platform Marker API. On iOS, callers can use GoogleMapsFlutterIOSMapConfiguration to independently disable implicit marker position and rotation animations while preserving the existing animated behavior by default.

For context, the underlying native approach remains the same as in the earlier revision. The configuration is passed through the platform interface and Pigeon to the native marker controller. When an animation is disabled, the corresponding GMSMarker.position or rotation update is performed inside a CATransaction with actions disabled, causing the update to take effect immediately. Position and rotation use separate transactions so they can be controlled independently.

Gibbo97:
~~I'm actually going to mark this as draft and having a second go at getting Google to fix the underlying SDK bug fixed as it would be a lot cleaner.~~

~~Happy to resolve the conflicts and update this branch if you do just want it to go in @stuartmorgan-g~~

Edit: crossed out my earlier plan to draft and submit a bug.

Gibbo97:
Did some more digging. We do need this change and the SDK team are extremely unlike to make an API change.

The marker move is explicitly animated in the google maps iOS SDK documentation.
https://developers.google.com/maps/documentation/ios-sdk/reference/objc/Classes/GMSMarker
> Marker position. Animated

And they document the exact opt-out mechanism we are using in this code.
https://developers.google.com/maps/documentation/ios-sdk/release-notes#version_15_-_september_2013

> Changes to the GMSMarker.position property are now animated. You can disable this animation with the following sample code.
> ```
> [CATransaction begin];
> [CATransaction setDisableActions:YES];
> marker.position = ...;
> [CATransaction commit];
> ```

stuartmorgan-g:
Structurally the Dart flow here looks good to me, and I agree that this is something we should allow controls for in the plugin; unfortunately the Google Maps SDK has a number of things in the API surface that are inconsistent across platforms. Adding @bparrishMines for a second look at the cross-platform API structure in case he has feedback given his experience with these kinds of APIs (see https://github.com/flutter/packages/pull/11792#issuecomment-4905929356 for context on why I suggested this approach, but input welcome).

FYI https://github.com/flutter/packages/pull/12768 will cause a little more churn, in case you want to wait to resolve until that lands, but the impact of that one should be minimal.

packages/google_maps_flutter/google_maps_flutter_platform_interface/lib/src/types/platform_map_configuration.dart
```diff
@@ -0,0 +1,15 @@
...
+// found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+
+/// Base class for platform-specific map configuration.
+///
+/// Platform implementations can add platform-specific configuration options
+/// by extending this class.
+@immutable
+class PlatformMapConfiguration {
```
bparrishMines, line 12: This should use `base` to ensure its extended only.

packages/google_maps_flutter/google_maps_flutter_android/lib/src/google_maps_flutter_android.dart
```diff
@@ -9,7 +9,8 @@ import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/services.dart';
-import 'package:google_maps_flutter_platform_interface/google_maps_flutter_platform_interface.dart';
+import 'package:google_maps_flutter_platform_interface/google_maps_flutter_platform_interface.dart'
```
bparrishMines, line 12: Does this need to be hidden?

bparrishMines (commented):
Since this seems like an iOS only feature, are the `google_maps_flutter_android` changes needed?

bparrishMines (commented):
The cross platform API looks fine to me. It seems like a good way to add platform-specific functionality without a breaking change.

stuartmorgan-g:
> FYI #12768 will cause a little more churn, in case you want to wait to resolve until that lands, but the impact of that one should be minimal.

All of the iOS churn PRs have landed, so it's safe to update this now so that it can be reviewed by platform teams.

PR #11797, opened 2026-05-27
[tool] Make more commands work without Flutter

stuartmorgan-g:
- Updates some cases where 'flutter' was used unconditionally, now that we need to support a repository without Flutter installed.
  - Removes some outdated comments and plumbing for forcing `flutter pub` over `dart pub`, as the reason we used to do it no longer applies now that `dart pub` also fetches examples by default.
- Adds `--skip-if-not-supporting-dart-version` to support N-1 and N-2 testing in core-packages.

### DIFF
--- a/script/tool/lib/src/analyze_command.dart
+++ b/script/tool/lib/src/analyze_command.dart
@@ -357,8 +357,11 @@ class AnalyzeCommand extends PackageLoopingCommand {
   }
 
   Future<bool> _runPubCommand(RepositoryPackage package, String command) async {
+    final String pubCommand = package.requiresFlutter()
+        ? flutterCommand
+        : 'dart';
     final int exitCode = await processRunner.runAndStream(
-      flutterCommand,
+      pubCommand,
       <String>['pub', command],
       workingDir: package.directory,
     );
--- a/script/tool/lib/src/analyze_command.dart
+++ b/script/tool/lib/src/analyze_command.dart
@@ -13,6 +13,7 @@ import 'common/gradle.dart';
 import 'common/output_utils.dart';
 import 'common/package_looping_command.dart';
 import 'common/plugin_utils.dart';
+import 'common/pub_utils.dart';
 import 'common/repository_package.dart';
 import 'common/xcode.dart';
 
@@ -357,15 +358,13 @@ class AnalyzeCommand extends PackageLoopingCommand {
   }
 
   Future<bool> _runPubCommand(RepositoryPackage package, String command) async {
-    final String pubCommand = package.requiresFlutter()
-        ? flutterCommand
-        : _dartBinaryPath;
-    final int exitCode = await processRunner.runAndStream(
-      pubCommand,
-      <String>['pub', command],
-      workingDir: package.directory,
+    return runPubCommand(
+      <String>[command],
+      package,
+      processRunner,
+      platform,
+      dartSdkPathOverride: _dartBinaryPath,
     );
-    return exitCode == 0;
   }
 
   /// Runs Gradle lint analysis for the given package, and returns the result
--- a/script/tool/lib/src/analyze_command.dart
+++ b/script/tool/lib/src/analyze_command.dart
@@ -359,7 +359,7 @@ class AnalyzeCommand extends PackageLoopingCommand {
   Future<bool> _runPubCommand(RepositoryPackage package, String command) async {
     final String pubCommand = package.requiresFlutter()
         ? flutterCommand
-        : 'dart';
+        : _dartBinaryPath;
     final int exitCode = await processRunner.runAndStream(
       pubCommand,
       <String>['pub', command],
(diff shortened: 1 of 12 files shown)


script/tool/lib/src/analyze_command.dart
```diff
@@ -357,8 +357,11 @@ class AnalyzeCommand extends PackageLoopingCommand {
   }
 
   Future<bool> _runPubCommand(RepositoryPackage package, String command) async {
+    final String pubCommand = package.requiresFlutter()
+        ? flutterCommand
+        : 'dart';
```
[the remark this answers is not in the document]
stuartmorgan-g: I forgot this command had a custom binary path available; fixed.

script/tool/lib/src/common/pub_utils.dart
```diff
@@ -12,22 +12,17 @@ import 'repository_package.dart';
...
   Platform platform, {
-  bool alwaysUseFlutter = false,
   bool streamOutput = true,
 }) async {
   // Running `dart pub get` on a Flutter package can fail if a non-Flutter Dart
   // is first in the path, so use `flutter pub get` for any Flutter package.
-  final bool useFlutter = alwaysUseFlutter || package.requiresFlutter();
+  final bool useFlutter = package.requiresFlutter();
   final command = useFlutter
       ? (platform.isWindows ? 'flutter.bat' : 'flutter')
```
chunhtai: This can use flutterCommand from package_command.dart

script/tool/lib/src/analyze_command.dart
```diff
@@ -357,8 +357,11 @@ class AnalyzeCommand extends PackageLoopingCommand {
   }
 
   Future<bool> _runPubCommand(RepositoryPackage package, String command) async {
```
chunhtai, line 360: should the be refactored out to pub_utils.dart and let the runPubGet to call this method instead?

chunhtai: will need to add a dart sdk override to the method if we move it

script/tool/lib/src/publish_check_command.dart
```diff
@@ -163,8 +163,11 @@ class PublishCheckCommand extends PackageLoopingCommand {
     await _fetchExampleDeps(package);
 
     print('Running pub publish --dry-run:');
+    final String pubCommand = package.requiresFlutter()
```
chunhtai: it feels like this could also be deuplicated if we move the _runPubCommand to the utils method

script/tool/lib/src/publish_command.dart
```diff
@@ -477,8 +477,11 @@ Safe to ignore if the package is deleted in this commit.
       _ensureValidPubCredential();
     }
 
+    final String pubCommand = package.requiresFlutter()
```
chunhtai: same

script/tool/lib/src/common/pub_utils.dart
```diff
@@ -12,22 +12,17 @@ import 'repository_package.dart';
...
-/// This can be useful, for instance, to get the `flutter`-default behavior
-/// of fetching example packages as well.
-///
 /// If [streamOutput] is false, output will only be printed if the command
 /// fails.
 Future<bool> runPubGet(
   RepositoryPackage package,
   ProcessRunner processRunner,
   Platform platform, {
-  bool alwaysUseFlutter = false,
```
chunhtai, line 25: do you know why we have this in the first place?

chunhtai (commented):
mostly looks good, just some suggestion that we should probably refactor some common logic out since they are used in multiple places

script/tool/lib/src/common/pub_utils.dart
stuartmorgan-g: This is a stand-alone utility; `flutterCommand` uses the `platform` passed to the command instance. The only way this could use `flutterCommand` is if it were bound to, or passed, a command (and I didn't want to do that just for this one line of logic, which is why this is separate in the first place).

script/tool/lib/src/common/pub_utils.dart
chunhtai: nvm, saw the pr description

script/tool/lib/src/analyze_command.dart
stuartmorgan-g: Done. It required enough extra plumbing that I'm not sure it's actually simpler than the duplication, but it does consolidate logic so is probably still a win.

script/tool/lib/src/publish_check_command.dart
stuartmorgan-g: Done, although it actually required a new, different `startPubCommand` because it needs to return a Process. But it does put everything in one file, sharing the dart/flutter selection logic.

script/tool/lib/src/publish_command.dart
stuartmorgan-g: Now uses the new startPubCommand.

chunhtai (approved):
LGTM

PR #11801, opened 2026-05-29
[material_ui] Add contrastLevel for M3 ColorScheme

QuncCccccc:
Fixes https://github.com/flutter/flutter/issues/184951

This PR is to add medium- and high-contrast M3 ColorScheme baselines. I also added a `double? contrastLevel` to `ThemeData`.

By default, `ThemeData()` shows standard M3 ColorScheme. To choose medium/high contrast, use
```
ThemeData(
  // brightness: Brightness.dark shows dark version of the baselines.
  contrastLevel: ContrastLevel.medium, // ContrastLevel.high
)
```

This change will not cause breaking changes because the standard-contrast m3 color scheme baseline stays the same.

Screenshot 2026-07-31 at 4 35 36 PM

TODO: 
Add an example or update an existing example to reflect the new feature.

## Pre-Review Checklist

- [x] I signed the [CLA].


packages/material_ui/tool/gen_defaults/templates/template.dart
```diff
@@ -177,9 +184,9 @@ abstract class TokenTemplate {
     if (verbose) {
       stdout.writeln('Generating contents...');
     }
-    final String contents = generateContents(_className);
+    final String contents = generateContents(requiresGeneratedClass ? _className : '');
```
QuncCccccc: We don't need a class for ColorScheme, so I added a `requiresGeneratedClass` field.

packages/material_ui/lib/src/theme_data.dart
```diff
@@ -199,6 +201,34 @@ enum StyleVariant {
...
+/// Named contrast levels for Material 3 color schemes.
+///
+/// Material 3 baseline color schemes provide standard, medium, and high
+/// contrast levels. When [ThemeData.colorSchemeSeed] is null, use [standard],
+/// [medium], or [high] to select a token-backed baseline color scheme with that
+/// contrast level.
+///
+/// A custom `double` between -1.0 and 1.0 can also be passed to
+/// [ThemeData.contrastLevel] when [ThemeData.colorSchemeSeed] is provided, in
+/// which case the value is passed to [ColorScheme.fromSeed].
```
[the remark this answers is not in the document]
QuncCccccc: Updated documentation.

packages/material_ui/lib/src/theme_data.dart
```diff
@@ -304,6 +347,7 @@ class ThemeData with Diagnosticable {
     ColorScheme? colorScheme,
     Brightness? brightness,
     Color? colorSchemeSeed,
+    double? contrastLevel,
```
chunhtai, line 350: how does this work with `MaterialApp. highContrastTheme` and `MaterialApp . highContrastDarkTheme`? it feels like we are inventing another system.

Android also has contrast setting that can set contrast from -1 and 1 which felt like where the material design is based on. is there a good way a developer can set this value based on Android settings?

chunhtai: See https://github.com/flutter/flutter/issues/182863

chunhtai (commented):
just some questions

packages/material_ui/lib/src/theme_data.dart
QuncCccccc: This PR is to add medium and high contrast `ColorScheme` baselines. See the M3 update section: https://m3.material.io/styles/color/system/overview#07f3a1d9-6c54-49f6-ae79-e52a6ee3e192

> how does this work with MaterialApp. highContrastTheme and MaterialApp . highContrastDarkTheme?

```
MaterialApp(
  highContrastTheme: ThemeData(contrastLevel: ContrastLevel.high)
  highContrastTheme: ThemeData(brightness: Brightness.dart, contrastLevel: ContrastLevel.high)
)
```
will provide default M3 colorSchemes with high contrast.

> Android also has contrast setting that can set contrast from -1 and 1 which felt like where the material design is based on. is there a good way a developer can set this value based on Android settings?

We do have an api `ColorScheme.contrastLevel` to set -1.0 - 1.0 and build a color scheme. But I don't think it is wired up to Android contrast setting. Are there an Android API available so we can set it up? If so, maybe we can so it in a separate PR.

packages/material_ui/lib/src/theme_data.dart
```diff
@@ -199,6 +201,35 @@ enum StyleVariant {
...
+/// `colorSchemeSeed` is provided, in which case the value is passed to
+/// [ColorScheme.fromSeed].
+///
+/// See also:
+///
+///  * [ColorScheme.fromSeed], which generates a color scheme from a seed color,
+///    brightness, and contrast level.
+///  * <https://m3.material.io/styles/color/system/overview>, the Material 3
+///    color system overview.
+abstract final class ContrastLevel {
```
elliette, line 222: Disregard if this is not a normal Flutter pattern, but I wonder if we could make this an enhanced enum instead?

`ThemeData` could then take the enum as a parameter, whereas `ColorScheme.fromSeed` would take a double (and we could document that there are preset values in the `ContrastLevel` enum.) 

```
ThemeData(contrastLevel: ContrastLevel.medium)
```

vs

```
ColorScheme.fromSeed(contrastLevel: ContrastLevel.medium.value)
```

This would disallow any non-supported levels in `ThemeData`, so we wouldn't need the warning comment about them being dropped.

packages/material_ui/lib/src/theme_data.dart
chunhtai: > Are there an Android API available so we can set it up?

we don't have yet, but i think sooner or later we need to do so. Just want to make sure the api you propose here can be wired up in the future.

Is the intention to deprecate the highContrastTheme and other variant and solely relies on developer to create their own theme class by reading highcontrast double and brightness from mediaquery?

packages/material_ui/lib/src/theme_data.dart
QuncCccccc: Thanks for your review! I considered this, but I think keeping `ThemeData.contrastLevel` as a `double` better fits here.

  `ThemeData` uses `contrastLevel` in two paths:
  1. when `colorSchemeSeed` is provided, it forwards the value directly to `ColorScheme.fromSeed`, which supports the full `-1.0..1.0` range;
  2. when `colorSchemeSeed` is null, it selects among the token-backed baseline schemes, where only `standard`, `medium`, and `high` have generated tokens.

  Using an enum would make the token-backed path stricter, but it would prevent `ThemeData(colorSchemeSeed: Colors.red contrastLevel: 0.6)`, even though that is valid for `ColorScheme.fromSeed`. It would also make future system contrast integration
  harder, since platform APIs may expose a continuous contrast value rather than only the three named presets.

  So I’d prefer to keep the parameter as `double?`, with `ContrastLevel.standard/medium/high` as named constants for the token-backed baselines, and document that custom values only apply when a seed is provided. WDYT😄?

packages/material_ui/lib/src/theme_data.dart
QuncCccccc: No, the intention is not to deprecate `highContrastTheme` or `highContrastDarkTheme` in this PR.

This PR is just to let `ThemeData` generate/select the right M3 **ColorScheme** for standard, medium, and high contrast. `MaterialApp.highContrastTheme` still has a broader role because it lets developers provide a complete high-contrast `ThemeData`, not only a different color scheme. Does this make sense?

packages/material_ui/tool/gen_defaults/templates/template.dart
```diff
@@ -93,6 +93,18 @@ abstract class TokenTemplate {
...
+  String get outputFileName {
+    final String snakeName = name.toLowerCase().replaceAll(' ', '_');
+    return switch (_version) {
+      _MaterialVersion.material3 => '${snakeName}_defaults_m3.g.dart',
+      _MaterialVersion.material3Expressive => '${snakeName}_defaults_m3e.g.dart',
+    };
+  }
+
+  /// Whether the generated contents must declare the generated defaults class.
+  bool get requiresGeneratedClass => true;
```
elliette: Instead of adding `requiresGeneratedClass`, what if we simply override the `className` (like how `outputFileName` is being overridden?)

elliette (approved):
LGTM with one suggestion!

packages/material_ui/tool/gen_defaults/templates/template.dart
QuncCccccc: That makes sense. I removed `requiresGeneratedClass` and give `className` an empty string! Thanks for the suggestion:)!

packages/material_ui/tool/gen_defaults/templates/template.dart
```diff
@@ -76,6 +76,9 @@ abstract class TokenTemplate {
 
   /// The name of the class that will be generated (e.g. `_IconButtonDefaultsM3`
   /// or `_IconButtonDefaultsM3E`).
+  ///
+  /// Return an empty string for templates that generate top-level declarations
+  /// instead of a class.
```
elliette: Phrasing suggestion to make this a little clearer: "Templates that generate top-level declarations instead of a class should override and return an empty string"

packages/material_ui/tool/gen_defaults/templates/template.dart
```diff
@@ -218,10 +219,11 @@ abstract class TokenTemplate {
...
       stdout.writeln('Generating contents...');
     }
-    final String contents = generateContents(requiresGeneratedClass ? className : '');
+    final String generatedClassName = className;
+    final String contents = generateContents(generatedClassName);
     assert(
-      !requiresGeneratedClass || contents.contains(_classRegExp),
-      'The generated contents for "$name" must define the class "$className". '
+      generatedClassName.isEmpty || contents.contains(_classRegExp),
+      'The generated contents for "$name" must define the class "$generatedClassName". '
```
elliette, line 226: Maybe add: "or override className to be empty"

elliette (approved):

packages/material_ui/tool/gen_defaults/templates/template.dart
QuncCccccc: Updated both! Thank you!

QuncCccccc:
Porting it over to https://github.com/flutter/packages/pull/12743. Closing this one.

